PR #1802, opened 2022-10-08
Fix native types2

rajyan:
closes https://github.com/phpstan/phpstan-src/pull/1535
closes https://github.com/phpstan/phpstan-src/pull/1627

and a lot of native type related issues!

Native types should always be specified along phpdoc types.
Also, there should be only a single `treatPhpDocTypesAsCertain` handling to switch between `variableTypes` and `nativeTypeExpression` 
https://github.com/phpstan/phpstan-src/blob/b2bf589eface47075bb83d7e75f582f8aa218d72/src/Analyser/MutatingScope.php#L506-L511

With this pull request, native types and phpdoc types are completely separated by `treatPhpDocTypesAsCertain`.

I hope I've understood
https://github.com/phpstan/phpstan-src/pull/1372#issuecomment-1143182394
> So - if you're interested in $treatPhpDocTypesAsCertain=true, call Scope::getType(), if false, call Scope::getNativeType(). Can you imagine this working? Thanks :)

this correctly...

PR #1805, opened 2022-10-09
One solution

rajyan:
tmp pr. just want to know the results of the integration tests.


rajyan:
This is one possible solution I can think of.
1. `treatPhpDocTypesAsCertain` is kept in `MutatingScope`, but can only be switched from rules
2. `treatPhpDocTypesAsCertain: false` makes `getType` to return native types
3. `nativeTypes` uses https://github.com/phpstan/phpstan-src/pull/1805/files#diff-dc817f2bab8672057a95375a542e68599164f6ab2380e4cccd1b50583e295d85R2116 which should be mostly correct and least an improvement, because the current `getNativeType` is falling back to full types.

rajyan:
The failing test in pocketmine for example
https://github.com/phpstan/phpstan-src/actions/runs/3214922744/jobs/5255662194
is showing that the native types and specification that wasn't working before, is working with this pull request.

ondrejmirtes:
This isn't the direction I want to go for. It breaks some of the ground rules I set in comments of the other PR, and is less useful for my goals.

For example - making getType return only native type when the setting is set to false, it completely breaks user expectation. Basically it says "don't use PHPDocs". Which isn't what the setting is about. We still want to infer types using PHPDocs, we just don't want to complain if the user double-checks the passed type in function arguments.

src/Analyser/MutatingScope.php
```diff
@@ -506,6 +506,13 @@ public function hasVariableType(string $variableName): TrinaryLogic
 	/** @api */
 	public function getVariableType(string $variableName): Type
 	{
+		if (!$this->treatPhpDocTypesAsCertain) {
+			if ($variableName === 'this' && $this->hasVariableType('this')->yes()) {
```
rajyan, line 510: Not exactly correct. 'this' should be duplicated for nativeTypes where assigned to variableType 'this'

src/Analyser/MutatingScope.php
```diff
@@ -2130,24 +2096,7 @@ private function createFirstClassCallable(array $variants): Type
...
-				$this->getTypeFromArrayDimFetch(
-					$expr,
-					$this->getNativeType($expr->dim),
-					$this->getNativeType($expr->var),
-				),
-			);
-		}
-
-		return $this->getType($expr);
+		return $this->doNotTreatPhpDocTypesAsCertain()->getType($expr);
```
rajyan, line 2099: This was the goal of my approach

src/Analyser/LazyScopeFactory.php
```diff
@@ -101,7 +98,7 @@ public function create(
 			$currentlyAllowedUndefinedExpressions,
 			$nativeExpressionTypes,
 			$inFunctionCallsStack,
-			$this->treatPhpDocTypesAsCertain,
+			true,
```
rajyan, line 101: `treatPhpDocTypesAsCertain` of MutatingScope would only be changed by `doNotTreatPhpDocTypesAsCertain`

rajyan: so not
> Basically it says "don't use PHPDocs". Which isn't what the setting is about.

The meaning of the setting is kept and when to use doNotTreatPhpDocTypesAsCertain is handled by the rule side.

ondrejmirtes:
I've got some ideas about this, I'll trynto explore them soon 😊

ondrejmirtes:
Can you please tell me which place in NodeScopeResolver has to care about getType() vs getNativeType()?

rajyan:
@ondrejmirtes 

> Can you please tell me which place in NodeScopeResolver has to care about getType() vs getNativeType()?

For example,
The test failure of https://github.com/phpstan/phpstan-src/pull/1802
https://github.com/phpstan/phpstan-src/actions/runs/3227859202/jobs/5283202887
> 4) PHPStan\Rules\Comparison\UnreachableIfBranchesRuleTest::testDoNotReportPhpDoc
--- Expected
+++ Actual
@@ @@
 '18: Elseif branch is unreachable because previous condition is always true.
 28: Else branch is unreachable because previous condition is always true.
 38: Elseif branch is unreachable because previous condition is always true.
+54: Else branch is unreachable because previous condition is always true.
+64: Elseif branch is unreachable because previous condition is always true.

Is caused by `getType` in this line
https://github.com/phpstan/phpstan-src/blob/e42952ecc182f1cff8c17761d1efba5598555b24/src/Analyser/NodeScopeResolver.php#L704

`$alwaysTerminating` is determined by the `$conditionType` which uses full types and causing the above error.

ondrejmirtes:
About UnreachableIfBranchesRuleTest - I don't think so because `UnreachableIfBranchesRule` already asks correctly about it and the code in NodeScopeResolver doesn't influence that:

```php
$conditionType = $this->treatPhpDocTypesAsCertain ? $scope->getType($condition) : $scope->getNativeType($condition);
$conditionBooleanType = $conditionType->toBoolean();
```

I think the problem is in some mixup of TypeSpecifier/SpecifiedTypes.

rajyan:
I was wrong with my explanation. The problem is caused by `isAlwaysTrue`
The scope merge is performed based on `isAlwaysTrue` of full types, which is causing wrong type for phpdoc types
simple example
https://phpstan.org/r/a8479355-de29-4e8c-9a08-7a2612ee94e1

rajyan:
https://github.com/phpstan/phpstan-src/pull/1805#issuecomment-1276968300
I'll try if this example works with this solution tonight, adding as test.

rajyan:
I now realized that this solution is not correct.
I think we need a separate two scopes for each full type and native type?

ondrejmirtes:
What's the purpose of this PR when compared to https://github.com/phpstan/phpstan-src/pull/1802 and in context of my plan at https://github.com/phpstan/phpstan/issues/8191? Thank you :)

ondrejmirtes:
If there's something in here that still makes sense, please post it as a new PR with this base: https://github.com/phpstan/phpstan-src/pull/1943 Thank you!

PR #1806, opened 2022-10-10
Fix return type of array_search() with constant array type haystack

takaram:
Return type of `array_search()` may be false unless we are sure that `$haystack` includes a value identical to `$needle`.

Closes phpstan/phpstan#3789

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-3789.php
+++ b/tests/PHPStan/Analyser/data/bug-3789.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Bug3789;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @param array{string, string, string} $haystack
+ */
+function doFoo(string $needle, array $haystack): void {
+	assertType('0|1|2|false', array_search($needle, $haystack, true));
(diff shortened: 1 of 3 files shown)


staabm:
//cc @herndlm this will need manual conflict resolving with https://github.com/phpstan/phpstan-src/pull/1874

tests/PHPStan/Analyser/data/bug-3789.php
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @param array{string, string, string} $haystack
+ */
+function doFoo(string $needle, array $haystack): void {
+	assertType('0|1|2|false', array_search($needle, $haystack, true));
+	assertType('0|1|2|false', array_search('foo', $haystack, true));
+
+	$haystack[1] = 'foo';
+	assertType('0|1|2|false', array_search($needle, $haystack, true));
+	assertType('0|1|2', array_search('foo', $haystack, true));
```
staabm: please also add test-cases for the non-strict variant (leaving the 3rd param to `false`)

takaram: I added test cases.

ondrejmirtes:
Thank you!

PR #1807, opened 2022-10-10
Fix passing named arguments to by-ref parameter

rvanvelzen:
Fix phpstan/phpstan#8142


ondrejmirtes (changes requested):
Nice fix :) Thank you!

ondrejmirtes (approved):

PR #1796, opened 2022-10-05
Fix tests

herndlm:
related to 392f7bc78e28da6e0e3badff9bfe61afcce87561

PR #1809, opened 2022-10-10
Fix inferring variadic argument for anonymous function as list

rvanvelzen:


herndlm:
This is what I described at https://github.com/phpstan/phpstan-src/pull/1796#discussion_r987729877 and the counter part to https://github.com/phpstan/phpstan-src/commit/392f7bc78e28da6e0e3badff9bfe61afcce87561, right? nice

ondrejmirtes:
It's not exactly the same. Because PHPDocs above closures are not yet read, there's no support for `@no-named-arguments` there. PHP 8.0+ users will always have `int|string` as the array key there.

ondrejmirtes:
Thank you!

PR #1801, opened 2022-10-07
Optimize `NonexistentOffsetInArrayDimFetchCheck`

herndlm:
This special handling for constant arrays should not be needed and the nested dim loop can break the outer type loop already if it found something to report ($report will not get any more true than true :)).

PR #1811, opened 2022-10-10
Add `Type::getLastIterableValueType()`

herndlm:
just playing around, kind of POC for moving more array-related methods to `Type` which I mentioned in https://github.com/phpstan/phpstan-src/pull/1801#issuecomment-1272401520. I'm not sure about this or if this should be the direction to move forward for other array return type extensions too.. 

What I like
- no special constant array handling and union type building anymore in `ArrayPopFunctionReturnTypeExtension`. I actually **love** that

What I dislike
- it's basically just for constant array and `array_pop` in particular
- boiler plate code in some type classes, but this is already minimized thanks to the new array traits
- naming-wise it's maybe hard to understand in some cases, e.g. how would `flip()` be called on Type and what do people expect it to do ..
- not much related to this PR, but there's also `MaybeIterableTypeTrait` and `NonIterableTypeTrait` which are kind of related to the new array type traits IMO

Most likely this is the best-case example and others will be more complicated to move over.

This PR is really mostly just for discussing this to not do this in other PRs that are not related :)

### DIFF
--- a/src/Type/Php/ArrayPopFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayPopFunctionReturnTypeExtension.php
@@ -5,12 +5,10 @@ namespace PHPStan\Type\Php;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\NullType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
-use function count;
 
 class ArrayPopFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
 {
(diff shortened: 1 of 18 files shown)


src/Type/Php/ArrayPopFunctionReturnTypeExtension.php
```diff
@@ -32,29 +30,12 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-		}
-
-		$itemType = $argType->getIterableValueType();
+		$itemType = $argType->getLastIterableValueType();
 		if ($iterableAtLeastOnce->yes()) {
 			return $itemType;
 		}
 
-		return TypeCombinator::union($itemType, new NullType());
+		return TypeCombinator::addNull($itemType);
```
ondrejmirtes: This isn't strictly the same, if you didn't have a reason to change it, please revert it. It's problematic when a generic bound is nullable.

ondrejmirtes (changes requested):
This totally makes sense to me. Maybe we could tackle first value at the same time? Thanks 😊

src/Type/Php/ArrayPopFunctionReturnTypeExtension.php
herndlm: oh, I wasn't even thinking about that and thought it's basically the same 🤯

ondrejmirtes: The difference is that TypeCombinator::containsNull() and NullType->isSuperTypeOf($type) are different things.

ondrejmirtes:
BTW a random thought - don't feel obligated to get the design right the first time, it doesn't have to be perfect 😊 We'll eventually deprecate some of the new methods anyway, as we come up with better alternatives 😊

herndlm:
> BTW a random thought - don't feel obligated to get the design right the first time, it doesn't have to be perfect 😊 We'll eventually deprecate some of the new methods anyway, as we come up with better alternatives 😊

Yes. But you know, it is _very_ important that they are sorted in the _right_ order 😅

herndlm:
I saw a couple more cases like this one, expect more PRs :) I'll look at the complex ones at the end

herndlm:
this is ready and I'm eager to add a couple more 😊

ondrejmirtes (approved):

ondrejmirtes:
Thank you :)

PR #1812, opened 2022-10-10
Add `Type::getFirstIterableKeyType()` and `Type::getLastIterableKeyType()`

herndlm:
pretty sure this will conflict with my other PR in at least one file, but I'll just fix that when the time comes

### DIFF
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -152,6 +152,16 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 		return IntegerRangeType::fromInterval(0, null);
 	}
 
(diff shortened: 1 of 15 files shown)


ondrejmirtes:
This one should be rebased now. In the future it'd make sense to only send one PR at a time - you'll always apply learnings from the earlier PRs in the later PRs. It's my preferred workflow.

herndlm:
> This one should be rebased now. In the future it'd make sense to only send one PR at a time - you'll always apply learnings from the earlier PRs in the later PRs. It's my preferred workflow.

yeah, I definitely agree and also normally work like that. in this case the motivation took over. but I stopped after the first one since I knew it would end in rebase hell :D

ondrejmirtes:
Thank you!

PR #1814, opened 2022-10-11
Slightly optimize TypeCombinator::reduceArrays() even more

rvanvelzen:
Keep track of the actual overlapping keys and only try merging when the overlapping keys match completely


ondrejmirtes:
Nice, thank you!

PR #1815, opened 2022-10-11
Fix resolving class types in inherited docblock

rvanvelzen:
Fix phpstan/phpstan#7839


ondrejmirtes (changes requested):
Additional NodeScopeResolverTest with assertType() would be nice to verify what it's actually resolved to.

ondrejmirtes:
Thank you!

mvorisek:
Thank you!

PR #1817, opened 2022-10-11
Infer constant array type from specifying count() on a list type

rvanvelzen:
Fix phpstan/phpstan#3499


ondrejmirtes:
I'm drooling over here :) Thank you very much.

herndlm:
a bit of drooling here too :D

PR #1818, opened 2022-10-11
Improve array_search type specifier performance

rvanvelzen:
Mostly resolves phpstan/phpstan#8147. The added integration test case went down from a ~minute to ~10 seconds locally.


herndlm:
haha, nice one

src/Type/Php/ArraySearchFunctionTypeSpecifyingExtension.php
```diff
@@ -45,7 +42,7 @@ public function specifyTypes(
 
 		return $this->typeSpecifier->create(
 			$arrayArg,
-			TypeCombinator::intersect(TypeCombinator::intersect(new ArrayType(new MixedType(), new MixedType()), $scope->getType($arrayArg)), new NonEmptyArrayType()),
+			new NonEmptyArrayType(),
```
ondrejmirtes: Can we return empty SpecifiedTypes if `!$arrayArg->isArray()->yes()`? What I want to avoid is to have `NonEmptyArrayType` on its own (without an intersection) which would happen for MixedType.

ondrejmirtes (changes requested):

ondrejmirtes:
Something's fishy here. Without the change in `src`, the test on my machine runs in 6 seconds. With the change, it runs in less than 2 seconds.

```
$ vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php --filter testBug8147
PHPUnit 9.5.23 #StandWithUkraine

Time: 00:01.874, Memory: 76.00 MB

OK (1 test, 1 assertion)
```

Sure, 3x improvement is nice, but the numbers are nowhere near "~minute to ~10 seconds".

Maybe something change meanwhile that also optimized this?

ondrejmirtes:
Thank you.

rvanvelzen:
I did reduce the test case a bit afterwards, which obviously reduces the runtime a bit. My machine was also quite busy which could have skewed the numbers a bit.

ondrejmirtes:
Makes sense 😊

PR #1820, opened 2022-10-11
added bug 7692 regression test

staabm:
the main problem of the initial report was the error

> +'42: PHPDoc type class-string<InheritingNamespace7692\EntityBaseInterface> of property DeepInheritingNamespace7692\TheChildService::$entityClass is not the same as PHPDoc type class-string<BaseNamespace7692\Entity\EntityBaseInterface> of overridden property BaseNamespace7692\TheBaseService::$entityClass.

which is beeing coverd by this test.

closes https://github.com/phpstan/phpstan/issues/7692


ondrejmirtes:
Thank you.

PR #1821, opened 2022-10-11
Move iterable methods from array to iterable traits

herndlm:
this makes more sense to me, having this in the array traits felt a bit weird already in https://github.com/phpstan/phpstan-src/pull/1811

Also fixes the new iterable methods via `IterableType` and `ObjectType`

### DIFF
--- a/src/Type/BooleanType.php
+++ b/src/Type/BooleanType.php
@@ -7,9 +7,11 @@ use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantFloatType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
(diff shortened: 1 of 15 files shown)


herndlm:
~~needed a couple more nudges than I thought because of the usage of `JustNullableTypeTrait`~~
UPDATE: I think I was misusing `JustNullableTypeTrait` by using other traits inside it which is a bit weird maybe and could block future types somehow potentially. I moved use statements out of it and rather use `NonArrayTypeTrait` and `NonIterableTypeTrait` explicitly which feels "safer". The couple of lines saved was not worth it.

Now I'm happy with this :)

herndlm:
fyi the failure is unrelated, there seems to be a new baseline update needed for composer on 1.9.x

ondrejmirtes:
Thank you.

PR #1822, opened 2022-10-11
Add `Type::isConstantArray()`

herndlm:
Does not get rid of many `instanceof` type checks yet unfortunately though. in order to achieve that more methods on type are needed. I hope I can do a bit more soon, but there are definitely limits to what makes sense right now.

### DIFF
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -472,7 +472,7 @@ class MixedType implements CompoundType, SubtractableType
 
 	public function isConstantArray(): TrinaryLogic
 	{
-		return TrinaryLogic::createMaybe();
+		return $this->isArray();
 	}
 
 	public function isOversizedArray(): TrinaryLogic
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -470,6 +470,11 @@ class MixedType implements CompoundType, SubtractableType
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isConstantArray(): TrinaryLogic
+	{
+		return TrinaryLogic::createMaybe();
+	}
+
 	public function isOversizedArray(): TrinaryLogic
 	{
 		if ($this->subtractedType !== null) {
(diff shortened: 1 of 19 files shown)


herndlm:
failures seem to be unrelated packagist timeouts, I'll leave this alone for now..

src/Type/MixedType.php
```diff
@@ -470,6 +470,17 @@ public function isArray(): TrinaryLogic
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isConstantArray(): TrinaryLogic
+	{
+		if ($this->subtractedType !== null) {
+			if ($this->subtractedType->isSuperTypeOf(new ConstantArrayType([], []))->yes()) {
+				return TrinaryLogic::createNo();
+			}
+		}
```
staabm: supporteded substraction is a new feature which should be covered by a test

herndlm: thx, I had a bad feeling about this and should have been more careful. trying to add tests made me realize that I have no clue how to implement this correctly for constant arrays, so I got rid of it and it returns always maybe now

staabm: hmm just guesswork, but maybe something like

```php
function doFoo($mixed) {
  $arr = ['a' => 1];
  if ($mixed != $arr) {
     assertType(...);
  }
}
```

or do you mean you don't know how the implementation should look like? :)

herndlm: more the latter. I think my initial implementation with the empty constant array is not correct

src/Type/MixedType.php
```diff
@@ -470,6 +470,11 @@ public function isArray(): TrinaryLogic
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isConstantArray(): TrinaryLogic
+	{
+		return TrinaryLogic::createMaybe();
```
ondrejmirtes: Should say no if `isArray()` says no (it's subtracted).

ondrejmirtes (changes requested):

src/Type/MixedType.php
herndlm: pushed a commit. just like that? I did not fully understand array subtraction here

ondrejmirtes: Yeah, make sense 😊

ondrejmirtes:
Thank you!

PR #1824, opened 2022-10-12
added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/5333

### DIFF
--- a/tests/PHPStan/Rules/Classes/data/bug-5333.php
+++ b/tests/PHPStan/Rules/Classes/data/bug-5333.php
@@ -1,0 +1,122 @@
+<?php declare(strict_types = 1);
+
+namespace Bug5333;
+
+use function PHPStan\Testing\assertNativeType;
+use function PHPStan\Testing\assertType;
+
+class Route {}
+final class FinalRoute {}
+
+class HelloWorld
+{
+	/**
+	 * @param Route|Route[] $foo
+	 *
+	 * @return Route
+	 **/
+	public function sayHello($foo): Route
+	{
+		if (\is_array($foo)) {
+			$res = $foo[0];
+			assertType('Bug5333\Route', $res);
+			assertNativeType('mixed', $res);
+
+			if (!$res instanceof Route) {
+				throw new \Exception();
+			}
+
+			assertType('array<Bug5333\Route>', $foo);
+			assertNativeType('array', $foo);
+
+			assertType('Bug5333\Route', $res);
+			assertNativeType('Bug5333\Route', $res);
+
+			return $res;
+		}
+
+		return $foo;
+	}
+}
+
+class HelloWorld2
+{
+	/**
+	 * @var Route|callable():Route
+	 **/
+	private $foo;
+
+	/**
+	 * @param Route|callable():Route $foo
+	 **/
+	public function setFoo($foo): void
+	{
+		assertType('Bug5333\Route|(callable(): Bug5333\Route)', $foo);
+		assertNativeType('mixed', $foo);
+
+		$this->foo = $foo;
+	}
+
+	public function getFoo(): Route
+	{
+		assertType('Bug5333\Route|(callable(): Bug5333\Route)', $this->foo);
+		assertNativeType('Bug5333\Route|(callable(): Bug5333\Route)', $this->foo);
+
+		if (\is_callable($this->foo)) {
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Classes/data/bug-5333.php
```diff
@@ -0,0 +1,34 @@
...
+	{
+		assertNativeType('mixed', $foo);
+		if (\is_array($foo)) {
+			assertNativeType('array', $foo);
+			$res = $foo[0];
+			if (!$res instanceof Route) {
+				throw new \Exception();
+			}
+			assertNativeType('array', $foo);
+			assertNativeType('Bug5333\Route', $res);
```
ondrejmirtes, line 33: Please test assertType for each expr too, so we can see the difference.

tests/PHPStan/Rules/Classes/data/bug-5333.php
```diff
@@ -0,0 +1,34 @@
...
+	 *
+	 * @return Route
+	 **/
+	public function sayHello($foo): Route
+	{
+		assertNativeType('mixed', $foo);
+		if (\is_array($foo)) {
+			assertNativeType('array', $foo);
+			$res = $foo[0];
+			if (!$res instanceof Route) {
```
ondrejmirtes, line 25: The most interesting thing to me is the native type of `$res` before this condition.

ondrejmirtes (changes requested):
Please also test https://phpstan.org/r/fa35b3cc-f0f1-4fd9-9a76-70f24bdcdc0b in the same manner.

tests/PHPStan/Rules/Classes/data/bug-5333.php
```diff
@@ -0,0 +1,79 @@
...
+	{
+		assertType('Bug5333\Route|(callable(): Bug5333\Route)', $this->foo);
+		assertNativeType('Bug5333\Route|(callable(): Bug5333\Route)', $this->foo);
+
+		if (\is_callable($this->foo)) {
+			assertType('(callable(): Bug5333\Route)', $this->foo);
+			assertNativeType('(Bug5333\Route&callable(): mixed)|(callable(): Bug5333\Route)', $this->foo);
+
+			$res = ($this->foo)();
+			assertType('array', $res);
```
staabm: shouldn't this be `Route`?

tests/PHPStan/Rules/Classes/data/bug-5333.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @var Route|callable():Route
+	 **/
+	private $foo;
+
+	/**
+	 * @param Route|callable():Route $foo
+	 **/
+	public function setFoo($foo): void
+	{
+		assertType('array', $foo);
```
staabm: this looks wrong

ondrejmirtes: You're not testing anything, the function isn't used.

staabm: good catch🤦‍♂️

tests/PHPStan/Rules/Classes/data/bug-5333.php
```diff
@@ -0,0 +1,80 @@
...
+		assertType('Bug5333\Route|(callable(): Bug5333\Route)', $this->foo);
+		assertNativeType('Bug5333\Route|(callable(): Bug5333\Route)', $this->foo);
+
+		if (\is_callable($this->foo)) {
+			assertType('(Bug5333\Route&callable(): mixed)|(callable(): Bug5333\Route)', $this->foo);
+			assertNativeType('(Bug5333\Route&callable(): mixed)|(callable(): Bug5333\Route)', $this->foo);
+
+			$res = ($this->foo)();
+			assertType('mixed', $res);
+			assertNativeType('mixed', $res);
```
staabm, line 71: can we be more precise here?

```suggestion
			assertType('callable(): Bug5333\Route', $this->foo);
			assertNativeType('(Bug5333\Route&callable(): mixed)|(callable(): Bug5333\Route)', $this->foo);

			$res = ($this->foo)();
			assertType('Bug5333\Route', $res);
			assertNativeType('mixed', $res);
```

ondrejmirtes: Please add a third example and use FinalRoute (final class) instead of Route. You'll see it's gonna be more precise.

ondrejmirtes:
Thank you.

PR #1828, opened 2022-10-12
Add `Type::getIterableCount()`

herndlm:


ondrejmirtes:
I'm blown away, this is really nice!

What it allows us to do in the future is to track precise array size:

```php
if (count($a) >= 3) {
    array_shift($a);
    // now we now it's >= 2
}
```

So there should be something like `AccessoryArraySizeType` in the future. What's really nice with the "everything is a method on Type" approach is that we can nicely implement the array modifications with methods on `Type` like `push`, `pop`, `shift`, `unshift`. No more weird `TypeUtils::getAccessoryTypes()` juggling!

The precise array size will allow us to tell that when we have a `list` of size `3`, `$list[3]` will definitely not exist.

What's hard about this is that we need to solve all interactions between AccessoryArraySizeType and NonEmptyArrayType. Unions/intersections need to behave deterministically. We should probably deprecate NonEmptyArrayType.

Another thing to think about is type description. Maybe AccessoryArraySizeType should stay hidden similarly to HasPropertyType and HasMethodType because in most situations, it will be irrelevant to the user that we know "this array has size 14". Maybe we should keep saying "non-empty-array" if we have AccessoryArraySizeType of `> 0`.

Anyway, thanks for this PR, it's great.

ondrejmirtes:
BTW the traits situation currently is a bit messy. These are the warnings I get in PhpStorm:

Screenshot 2022-10-13 at 9 04 16

I think using a trait only makes sense if:

1) Type traits themselves have disjoint sets of methods. Meaning that there cannot be multiple traits that define the same method. We can verify this by iterating all used traits by a class (recursively) and see if there are any duplicate methods.
2) If the Type implementation class doesn't override any of the methods used from traits.

So if we want to continue using traits in types, we should write internal PHPStan rules to check these usages in PHPStan :)

herndlm:
Interesting! Yeah, moving the array modifications to Type is what I wanted to do next.
And regarding traits: I see. Yes, this should be adressed sooner than later I guess :/

herndlm:
but regarding the trait example you mentioned with `HasPropertyType` - I didn't see where e.g. `isOversizedArray()` clashed. can you? is PHPStorm maybe confused by our traits? :D I went through all traits and the class itself twice and only saw i once in `NonArrayType`. I know other classes where such clashes are though

ondrejmirtes:
I also didn't find it, but we should have the rules to make sure it's clean.

PR #1829, opened 2022-10-12
Allow string args in signature of RedisCluster::del

ostrolucky:


herndlm:
I think you can target 1.8.x for bug fixes and function signature updates to get them faster released :)

ondrejmirtes:
Thank you.

PR #1830, opened 2022-10-13
Performance improvements

rvanvelzen:
Some perf improvements (mostly related to constant arrays) implemented while investigating phpstan/phpstan#8146.


herndlm:
with the many type related cleanups I do on 1.9.x we need to be careful to not cause conflicts if bigger type changes are made on 1.8.x still :)

rvanvelzen:
I can just point this at 1.9.x of course, though the merge looked relatively clean locally :)

ondrejmirtes:
I rebased this, but the performance issue in the new test doesn't go away :/

ondrejmirtes:
This is probably a dead end :) We already optimized 8146, mainly by dumbing down the arrays again, and be a bit more forgiving with oversized arrays in type checking.

PR #1831, opened 2022-10-13
Update `iterator_to_array` for PHP 8.2

taka-oyama:
`iterator_to_array` was changed to accept `iterable`.

### DIFF
--- a/resources/functionMap_php82delta.php
+++ b/resources/functionMap_php82delta.php
@@ -7,12 +7,12 @@
  */
 
 /**
- * This contains the information needed to convert the function signatures for php 8.0 to php 7.4 (and vice versa)
+ * This contains the information needed to convert the function signatures for php 8.2 to php 8.1 (and vice versa)
  *
  * This has two sections.
- * The 'new' section contains function/method names from FunctionSignatureMap (And alternates, if applicable) that do not exist in php7.4 or have different signatures in php 8.0.
+ * The 'new' section contains function/method names from FunctionSignatureMap (And alternates, if applicable) that do not exist in php 8.1 or have different signatures in php 8.2.
  *   If they were just updated, the function/method will be present in the 'added' signatures.
- * The 'old' signatures contains the signatures that are different in php 7.4.
+ * The 'old' signatures contains the signatures that are different in php 8.1.
  *   Functions are expected to be removed only in major releases of php.
  *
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
PHPStan should already knows this thanks to this: https://github.com/phpstan/php-8-stubs/blob/7294241c2a2a2dcc50c40c756b2dd2adb270cb5d/stubs/ext/spl/iterator_to_array.php#L7-L10

It's worth investigating why it's not the case, in https://github.com/phpstan/phpstan-src/blob/1.9.x/src/Reflection/SignatureMap/Php8SignatureMapProvider.php.

rvanvelzen:
`iterable` from the stubs is narrowed to `Traversable` from the function map: https://github.com/phpstan/phpstan-src/blob/27277582ca08e8b0c1211dcd9641cc6f37144a41/src/Reflection/SignatureMap/Php8SignatureMapProvider.php#L239-L242

ondrejmirtes:
We should have some code to keep the native union type.

curry684:
This PR is important, also for older PHP versions. Consider this code:
```php
    public function foo(): iterable
    {
        yield 'foo';
    }

    public function bar(): void
    {
        dump(iterator_to_array($this->foo()));
    }
```
PHP [explicitly defines](https://www.php.net/manual/en/language.types.iterable.php) `iterable` as a pseudotype (since PHP 7.1) for objects and arrays "implementing the [Traversable](https://www.php.net/manual/en/class.traversable.php) interface". Returning a non-`Traversable` is even a hard error:

> If a value is not an array or instance of [Traversable](https://www.php.net/manual/en/class.traversable.php), a [TypeError](https://www.php.net/manual/en/class.typeerror.php) will be thrown.

`iterable` is therefore a valid input for [`iterator_to_array`](https://www.php.net/manual/en/function.iterator-to-array):
```
iterator_to_array([Traversable](https://www.php.net/manual/en/class.traversable.php)|array $iterator, bool $preserve_keys = true): array
```
The example code above fails however at level 7:
```
Parameter #1 $iterator of function iterator_to_array expects Traversable, iterable given.
```
This is plain wrong.

ondrejmirtes:
@curry684 If you have a problem with a different PHP version, please open an issue about it with a reproducing snippet. Thank you.

PR #1832, opened 2022-10-13
Add `Type::flip()`

herndlm:
This would be the first new array modification function moved to Type. Most likely one of the simplest :)
As you can see and know anyways: it is a very specific thing, but it allows us to get rid of conditional code and `getConstantArrays()` calls, that I don't like much and are also messy with intersections/unions.

What about the name? There will be more to come, e.g. `pop`, `shift`, ... Does a common prefix like `array` for all of them make sense maybe? I'd definitely group them together everywhere, my worry is just that all of them are only meant for arrays and some could suggest that e.g. strings can be flipped or something like that.

### DIFF
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -139,7 +139,7 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 
 	public function flipArray(): Type
 	{
-		return $this;
+		return new MixedType();
 	}
 
 	public function isIterable(): TrinaryLogic
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -137,6 +137,11 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 		return new ErrorType();
 	}
 
+	public function flipArray(): Type
+	{
+		return $this;
+	}
+
 	public function isIterable(): TrinaryLogic
 	{
 		return TrinaryLogic::createYes();
--- a/src/Type/Accessory/NonEmptyArrayType.php
+++ b/src/Type/Accessory/NonEmptyArrayType.php
@@ -127,6 +127,11 @@ class NonEmptyArrayType implements CompoundType, AccessoryType
 		return new ErrorType();
 	}
 
+	public function flipArray(): Type
+	{
+		return $this;
+	}
+
 	public function isIterable(): TrinaryLogic
 	{
 		return TrinaryLogic::createYes();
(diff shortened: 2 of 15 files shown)


ondrejmirtes:
A name that makes sense in English would be best.

So `flipArray` ? 😊

BTW I want to rename getIterableCount to getArraySize. Makes more sense to me 😊

herndlm:
> BTW I want to rename getIterableCount to getArraySize. Makes more sense to me 😊

makes sense. I was also unsure about naming there and found iterable-related things I just kept using. maybe the maybe- and non- implementations should then be moved from the iterable traits to the array traits. not sure if counting iterables makes much sense anyways 🤔 in that case it doesn't even change anything. I'll open a PR in a bit 😊

herndlm:
I want to add another test case

herndlm:
now it's ok

src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -137,6 +137,11 @@ public function unsetOffset(Type $offsetType): Type
 		return new ErrorType();
 	}
 
+	public function flipArray(): Type
+	{
+		return $this;
```
rvanvelzen: Shouldn't flipping a list degrade to a regular array?

herndlm: oh right, thx! I'll add a test and adapt it

herndlm: adapted

src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -139,7 +139,7 @@ public function unsetOffset(Type $offsetType): Type
 
 	public function flipArray(): Type
 	{
-		return $this;
+		return new MixedType();
```
staabm, line 142: shouldn't this be more precise, like `ArrayType($valueType, 0+$positiveInt)`?

herndlm: I think that shouldn't be necessary because this type should come intersected with an array and intersecting with mixed then just removes it basically. the simpler the better, I think, but I'm also not sure and @rvanvelzen just recently improved performance by not being too precise in type specification. maybe he can confirm/explain it better :)

staabm: if it passes a non-constant-list test like

```
/**
 * @param list<string> $array
 */
function foo7($array)
{
	$flip = array_flip($array);
	assertType('array<string int<0, max>>', $flip);
}
```

we are good to go :)

ondrejmirtes:
> makes sense. I was also unsure about naming there

It depends... For example `isIterableAtLeastOnce()` makes sense to me because it's about iteration, about whether foreach runs at least once.

ondrejmirtes:
Beautiful, thank you :)

PR #1833, opened 2022-10-13
Rename getIterableCount to getArraySize and move it to appropriate type trait

herndlm:
ups, I undid the trait move, because `ObjectType` needs this logic, in fact it has to stay in the iterable traits because iterables are countable

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -104,7 +104,7 @@ class FunctionCallParametersCheck
 				if (count($arrays) > 0) {
 					$minKeys = null;
 					foreach ($arrays as $array) {
-						$countType = $array->getIterableCount();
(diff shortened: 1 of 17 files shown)


mad-briller:
so is `ErrorType` used as a sort of "this should not happen" type?

how does the phpstan react to the presence of this type, does it always report an error during the analysis regardless of where it's found? is it the source of the "X will result in an error" messages i've seen from time to time?

i appreciate this isn't really relevant to this PR but it would help me understand more :)

herndlm:
phew, who would have thought that renaming is so hard.. double-checked with full-text-search now, that should be it I hope

herndlm:
OK looks like a new error showed up in the mean time. I see it both here and it the other PR and guess it's unrelated to both then

ondrejmirtes:
Thank you!

PR #1804, opened 2022-10-08
Add `@param-out` support

staabm:
requires https://github.com/phpstan/phpdoc-parser/pull/150
refs https://github.com/phpstan/phpstan/issues/6426#issuecomment-1073791565

closes https://github.com/phpstan/phpstan/issues/7231
closes https://github.com/phpstan/phpstan/issues/6871
closes https://github.com/phpstan/phpstan/issues/6186
closes https://github.com/phpstan/phpstan/issues/4372

see https://psalm.dev/docs/annotating_code/supported_annotations/#param-out-psalm-param-out

just put togehter the basic building blocks and tests:
- [x] wire ResolvedPhpDocBlock
- [x] wire ResolvedDocNodeResolver
- [x] validate param-out phpdoc tags and tests
- [x] initial NodeScopeResolverTests
- [x] implement param-out type specification

I had the idea of adding param-out-type specification of the expressions right before the return-type of `FuntionLike`s is handled - I have not yet an idea where this place is though and whether thats a good idea :)

would be great to get a hint, where/how the actual param-out logic should be implemented.

PR #1834, opened 2022-10-13
Implement FunctionReflection->getDocComment()

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/1804#discussion_r994404893

### DIFF
--- a/src/PhpDoc/ResolvedPhpDocBlock.php
+++ b/src/PhpDoc/ResolvedPhpDocBlock.php
@@ -36,7 +36,7 @@ use function substr;
 class ResolvedPhpDocBlock
 {
 
-	private const EMPTY_DOC_STRING = '/** */';
+	public const EMPTY_DOC_STRING = '/** */';
 
 	private PhpDocNode $phpDocNode;
 
--- a/src/PhpDoc/ResolvedPhpDocBlock.php
+++ b/src/PhpDoc/ResolvedPhpDocBlock.php
@@ -198,7 +198,6 @@ class ResolvedPhpDocBlock
 		$result = new self();
 		// we will resolve everything on $this here so these properties don't have to be populated
 		// skip $result->phpDocNode
-		// skip $result->phpDocString - just for stubs
 		$phpDocNodes = $this->phpDocNodes;
 		$acceptsNamedArguments = $this->acceptsNamedArguments();
 		foreach ($parents as $parent) {
--- a/src/PhpDoc/ResolvedPhpDocBlock.php
+++ b/src/PhpDoc/ResolvedPhpDocBlock.php
@@ -36,6 +36,8 @@ use function substr;
 class ResolvedPhpDocBlock
 {
 
+	private const EMPTY_DOC_STRING = '/** */';
+
 	private PhpDocNode $phpDocNode;
 
 	/** @var PhpDocNode[] */
@@ -149,7 +151,7 @@ class ResolvedPhpDocBlock
 	{
 		// new property also needs to be added to merge()
 		$self = new self();
-		$self->phpDocString = '/** */';
+		$self->phpDocString = self::EMPTY_DOC_STRING;
 		$self->phpDocNodes = [];
 		$self->filename = null;
 		$self->templateTypeMap = TemplateTypeMap::createEmpty();
@@ -206,6 +208,7 @@ class ResolvedPhpDocBlock
 			}
 		}
 		$result->phpDocNodes = $phpDocNodes;
+		$result->phpDocString = $this->phpDocString;
 		$result->filename = $this->filename;
 		// skip $result->nameScope
 		$result->templateTypeMap = $this->templateTypeMap;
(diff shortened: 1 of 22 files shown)


staabm:
the errors of the failled job also appear in other currently open PRs, therefore seem unrelated

src/Reflection/Native/NativeFunctionReflection.php
```diff
@@ -102,4 +102,9 @@ public function getAsserts(): Assertions
 		return $this->assertions;
 	}
 
+	public function getDocComment(): ?string
+	{
+		return null;
```
ondrejmirtes: It's not true that this is always null. Look at where the instance is created, there can be docblock coming from phpstorm-stubs, or PHPStan stub files.

src/Reflection/Php/PhpFunctionReflection.php
```diff
@@ -256,4 +256,13 @@ public function getAsserts(): Assertions
 		return $this->asserts;
 	}
 
+	public function getDocComment(): ?string
+	{
+		$docComment = $this->reflection->getDocComment();
```
ondrejmirtes: This isn't always correct, look where functionReflectionFactory is called. It can either be the method's PHPDoc, or the one in stub files.

src/Reflection/Php/PhpMethodFromParserNodeReflection.php
```diff
@@ -130,7 +130,11 @@ public function isPublic(): bool
 
 	public function getDocComment(): ?string
 	{
-		return null;
+		$phpDoc = $this->getClassMethod()->getDocComment();
+		if ($phpDoc === null) {
+			return null;
+		}
+		return $phpDoc->getText();
```
ondrejmirtes: Technically you could delete this, parent method already does this correctly.

ondrejmirtes (changes requested):

tests/PHPStan/Reflection/FunctionReflectionTest.php
```diff
@@ -0,0 +1,87 @@
...
+			'/** some method phpdoc */',
+		];
+		yield [
+			'noDocMethod',
+			null,
+		];
+		yield [
+			'docViaStub',
+			'/** method doc via stub */',
+		];
```
staabm: this test is failling atm since phpstan currently does not pickup method phpdocs from stubs.

should it be also implemented with this PR, or should I drop the test for now?

staabm: fixed via https://github.com/phpstan/phpstan-src/pull/1834/commits/5f8e978545c39fdcbb8a14e5d7956337694adb03 in a separate commit.

src/PhpDoc/ResolvedPhpDocBlock.php
```diff
@@ -149,7 +151,7 @@ public static function createEmpty(): self
 	{
 		// new property also needs to be added to merge()
 		$self = new self();
-		$self->phpDocString = '/** */';
```
staabm, line 152: because I don't want to change this magic-value for BC reasons, I added a new `hasPhpDocString` to determine whether a phpdoc is user-defined or not

src/PhpDoc/ResolvedPhpDocBlock.php
```diff
@@ -202,6 +204,7 @@ public function merge(array $parents, array $parentPhpDocBlocks): self
 			}
 		}
 		$result->phpDocNodes = $phpDocNodes;
+		$result->phpDocString = $this->phpDocString;
```
staabm, line 210: without this line I got errors like 
>Error: Typed property PHPStan\PhpDoc\ResolvedPhpDocBlock::$phpDocString must not be accessed before initialization

now I see line 197  mentioning `// skip $result->phpDocString - just for stubs` ..hmm?

staabm:
just applied and covered the cases mentioned in https://github.com/phpstan/phpstan-src/pull/1804#issuecomment-1277556907

ondrejmirtes:
I just spent the morning working on this PR because I particularly didn't like the `hasPhpDocString()` approach... But I painfully realized that it has to work like this because of some mess in PhpDocInheritanceResolver / PhpDocBlock.

I wanted for `docBlockToResolvedDocBlock()` method to return null for non-existent docComments but it cannot work this way.

Turns out we have to call `$oneResolvedDockBlock->merge` even on an empty docblock because otherwise generics aren't correctly resolved...

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

ondrejmirtes:
Now you can get to the `@param-out` PR :)

staabm:
Thanks for finishing it. 

I didn't like the hasPhpDoc-method either, but couldn't come up with something better which worked ;-)

staabm:
> Now you can get to the @param-out PR 

Yep, now I have a better idea on the cases beeing involved

ondrejmirtes:
BTW looks like this PR broke several cases:

* https://github.com/phpstan/phpstan/issues/7519#issuecomment-1279925958
* https://github.com/phpstan/phpstan/issues/7317#issuecomment-1279925995
* https://github.com/phpstan/phpstan/issues/6829#issuecomment-1279926099

Can you please reproduce that and investigate that? Thaks!

staabm:
opened a separate issue to handle the regressions with https://github.com/phpstan/phpstan/issues/8175

PR #1835, opened 2022-10-13
Implement template default types

rvanvelzen:
Fixes phpstan/phpstan#4801

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-4801.php
+++ b/tests/PHPStan/Analyser/data/bug-4801.php
@@ -1,0 +1,25 @@
+<?php declare(strict_types = 1);
+
+namespace Bug4801;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @template T
+ */
+interface I {
+	/**
+	 * @template TResult = T
+	 * @param (callable(T): TResult)|null $a
+	 * @return I<TResult>
+	 */
+  public function work(callable|null $a): I;
+}
+
+/**
+ * @param I<string> $i
+ */
+function x(I $i) {
+	assertType('Bug4801\\I<string>', $i->work(null));
+	assertType('Bug4801\\I<int>', $i->work(fn(string $a) => (int) $a));
+}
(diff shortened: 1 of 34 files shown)


ondrejmirtes:
Feel free to rebase your PR without any Composer changes, I updated the parser on 1.9.x :)

tests/PHPStan/Analyser/data/template-default.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+namespace TemplateDefault;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @template T1 = true
+ * @template T2 = true
```
b1rdex, line 9: I wonder what syntax is for bound templates? Something like `@template T of bool = true`?

rvanvelzen: Yes

ruudk:
@rvanvelzen Are you still interested in solving this? Having default values would be extremely handy!

rvanvelzen:
@ruudk I won't have time to work on this in the foreseeable future, but I do intend to solve this.

That being said, anyone should feel free to take this over :)

ondrejmirtes:
@rvanvelzen I need to tell you that I really miss your contributions :) Not just this specific one, but in general - a PR from you was always something to look forward to :)

tests/PHPStan/Analyser/data/bug-4801.php
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types = 1);
+
+namespace Bug4801;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @template T
```
mvorisek, line 8: `@template T of Xxx` test should be added like in https://github.com/phpstan/phpstan/issues/4801#issuecomment-1925329868.

When `of Xxx` is specified, it should be used as implicit default type.

VincentLanglet:
Since https://github.com/phpstan/phpstan-src/pull/1835 is resolved ; I assume this PR can be closed @rvanvelzen

PR #1836, opened 2022-10-13
Fix a couple CURLOPT_* parameter types

jlherren:
Fixes phpstan/phpstan#8152
Fixes phpstan/phpstan#8157


ondrejmirtes:
Thank you.

PR #1841, opened 2022-10-14
Replace ArrayType::castToArrayKey() with Type::toArrayKey()

rvanvelzen:


ondrejmirtes:
All aboard the Type refactoring train! 🚆  I can't wait to get rid of inheritance in Type classes.

src/Type/StringType.php
```diff
@@ -137,6 +137,11 @@ public function toArray(): Type
 		);
 	}
 
+	public function toArrayKey(): Type
+	{
+		return $this;
```
herndlm, line 142: This needs to return most likely `int|string` I believe?

ondrejmirtes: It'd be correct but annoying.

rvanvelzen: It didn't do so in the previous implementation, and though technically that is correct I don't think changing it now is wise.

herndlm: ah right, I thought the failing tests were related, but looks like they aren't :)

rvanvelzen:
The improvement actually causes some tests to fail legitimately, which is why I kept is as a draft. I'll improve those separately first so this can land cleanly :)

herndlm:
ah right, I introduced usage of the old method again in the meantime in `HasOffsetValueType`, sorry 😊

src/Type/ArrayType.php
```diff
@@ -432,55 +429,15 @@ public function toArray(): Type
...
 
-	/** @deprecated Use getArraySize() instead */
-	public function count(): Type
+	public function toArrayKey(): Type
 	{
-		return $this->getArraySize();
+		return new ErrorType();
 	}
 
-	public static function castToArrayKeyType(Type $offsetType): Type
```
ondrejmirtes, line 441: Removing the method is a BC break, we should keep it but simply call `toArrayKey` on the type. And mark it as deprecated.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #1847, opened 2022-10-14
Add `Type::popArray()` and `Type::shiftArray()`

herndlm:
This is a boring one, but at least I could get rid of the `TypeTraverser` usage. I should have never added that..

I wasn't sure at first, but indeed, `array_shift` re-indexes the array and preserves the list.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -134,7 +134,6 @@ use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\Generic\TemplateTypeHelper;
 use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\IntegerType;
-use PHPStan\Type\IntersectionType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\NullType;
@@ -950,6 +949,7 @@ class NodeScopeResolver
 
 			do {
 				$prevScope = $bodyScope;
+				$bodyScope = $bodyScope->mergeWith($scope);
 				$bodyScopeResult = $this->processStmtNodes($stmt, $stmt->stmts, $bodyScope, static function (): void {
 				})->filterOutLoopExitPoints();
 				$alwaysTerminating = $bodyScopeResult->isAlwaysTerminating();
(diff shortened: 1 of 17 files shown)


herndlm:
oh, this is more interesting after all..

here are 2 failures
- ~~https://github.com/phpstan/phpstan-src/blob/1.8.9/tests/PHPStan/Analyser/data/bug-3875.php#L13 is `mixed` instead of `'foo'` because `$queue` becomes `NeverType` in the second iteration of the loop after https://github.com/phpstan/phpstan-src/blob/1.8.9/src/Analyser/NodeScopeResolver.php#L959. This is not the case with other types of loops, so it must be a do-while problem. That type was just passed through via array_pop handling before, now `NeverType::popArray()` results in `mixed`.~~ fixed by adapting do-while handling
- ~~https://github.com/phpstan/phpstan-src/blob/1.8.9/tests/PHPStan/Analyser/data/bug-3993.php#L20 is loosing the subtracted type because mixed is not passed through any more. I need to think about what should happen with subtracted types I guess.~~ irrelevant because of new MixedType handling

src/Analyser/NodeScopeResolver.php
```diff
@@ -950,6 +949,7 @@ private function processStmtNode(
 
 			do {
 				$prevScope = $bodyScope;
+				$bodyScope = $bodyScope->mergeWith($scope);
```
herndlm, line 952: this fixes the first failure, see also https://github.com/phpstan/phpstan-src/pull/1849
at least here removing it causes a test failure. but as mentioned there - I did not fully understand it tbh, it was do-while specific though

src/Type/MixedType.php
```diff
@@ -149,6 +149,16 @@ public function flipArray(): Type
 		return new self($this->isExplicitMixed);
 	}
 
+	public function popArray(): Type
```
staabm, line 152: These 2 cases are not tested

herndlm: added :)

src/Analyser/NodeScopeResolver.php
```diff
@@ -1836,25 +1836,10 @@ function (MutatingScope $scope) use ($expr, $nodeCallback, $context): Expression
...
-					}
-					return $type;
-				});
-
-				$scope = $scope->assignExpression(
+				$scope = $scope->invalidateExpression($arrayArg)->assignExpression(
 					$arrayArg,
-					$arrayArgType,
+					$functionReflection->getName() === 'array_pop' ? $arrayArgType->popArray() : $arrayArgType->shiftArray(),
 					$scope->getNativeType($arrayArg),
```
ondrejmirtes, line 1843: I feel like we should get the native array type and call the same method on it too.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Yeah, I guess it makes sense, the same thing is done (`$bodyScope = $bodyScope->mergeWith($scope);`) before the final analysis on line 978.

ondrejmirtes:
Thank you.

staabm:
@herndlm did you intentiomally not use the new Type methods in the array-pop and array-shift return type extension?

herndlm:
Yes, because those return the first or last element. I mixed this up recently as well :)

PR #1848, opened 2022-10-14
Abort loop processing if condition type gets never

herndlm:
This is a weird one. In https://github.com/phpstan/phpstan-src/pull/1847 I noticed that the loop of https://github.com/phpstan/phpstan-src/blob/1.8.9/tests/PHPStan/Analyser/data/bug-3875.php is executed twice and in the second iteration `$queue` becomes `never`. The reason is that the condition evaluates to false after the first iteration and in the bodies truthy scope this makes it never. Basically, the loop should only be executed once.

I wonder -  is this so special or is there a better way to improve this? I also didn't understand yet how this condition is different to others and why e.g. normal counter variables don't make a problem?


herndlm:
ah ok, this seems to be a do-while loop problem only

ondrejmirtes (commented):
We need a test here 😊

PR #1849, opened 2022-10-14
Add missing scope merge in do-while

herndlm:
Adds the scope merge that all other loops have as well. Fixes the first failure mentioned in https://github.com/phpstan/phpstan-src/pull/1847#issuecomment-1279540898

I didn't understand it fully though tbh

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -945,6 +945,7 @@ class NodeScopeResolver
 
 			do {
 				$prevScope = $bodyScope;
+				$bodyScope = $bodyScope->mergeWith($scope);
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Without a test this doesn't tell us anything 😊

herndlm:
Yeah I didn't succeed in creating one unfortunately. But in the linked PR I added the same thing and there it's needed. This is a hard one to argue about :/

herndlm:
I think the problem was that the never only happens inside the loop iteration and if I try to dump the type I only see the merged scopes where the never is eliminated. But in the other PR it is problematic because it is used via array_shift/pop and actually widens the type via the new methods. At least that's what I think so far.
But I agree, here without a test this is useless..

ondrejmirtes:
So it's linked to the other PR. If you can't make its tests pass without this, it belongs there.

PR #1850, opened 2022-10-15
Preserve subtracted type on `MixedType::flipArray()`

herndlm:
does that make sense? I'm still not good with subtracts, but it felt weird that that info is lost on flip, shouldn't hurt to preserve it I guess

### DIFF
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -146,7 +146,7 @@ class MixedType implements CompoundType, SubtractableType
 
 	public function flipArray(): Type
 	{
-		return new ArrayType(new MixedType(), new MixedType());
+		return new ArrayType(new MixedType($this->isExplicitMixed), new MixedType($this->isExplicitMixed));
 	}
 
 	public function isCallable(): TrinaryLogic
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -146,7 +146,7 @@ class MixedType implements CompoundType, SubtractableType
 
 	public function flipArray(): Type
 	{
-		return new self($this->isExplicitMixed);
+		return new self($this->isExplicitMixed, $this->subtractedType);
 	}
 
 	public function isCallable(): TrinaryLogic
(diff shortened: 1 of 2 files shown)


src/Type/MixedType.php
```diff
@@ -146,7 +146,7 @@ public function unsetOffset(Type $offsetType): Type
 
 	public function flipArray(): Type
 	{
-		return new self($this->isExplicitMixed);
```
ondrejmirtes, line 149: 1) You could just return $this?
2) Also, you could return array<mixed, mixed>...

ondrejmirtes (changes requested):

src/Type/MixedType.php
herndlm: oh right, the latter makes sense I guess. perfect chance to ask another question about the method above: why is the empty array subtracted via `unsetOffset()`? I don't get that. if an offset is unset it meant the array was not empty, but it can be empty afterwards, right?

ondrejmirtes: Typically if you ask why some code exists and what it does, you can try removing it and see what tests fail. You can also take a look at the commit in git blame which will give you more context.

But in this case - if the subtracted type says the mixed type cannot be an empty array, the empty array is removed from the subtracted type because the mixed type can be an empty array after unset.

herndlm: ah of course, that makes total sense. I had it mixed up because of the kind of double negation/removal here .. 😅

src/Type/MixedType.php
```diff
@@ -146,7 +146,7 @@ public function unsetOffset(Type $offsetType): Type
...
+		return new ArrayType(new MixedType(), new MixedType());
```
ondrejmirtes: I think the key and value should keep the explicitness of this MixedType value.

ondrejmirtes (changes requested):

src/Type/MixedType.php
herndlm: good point, yeah, let's try that 😊
adapted

ondrejmirtes:
Thank you!

PR #1851, opened 2022-10-15
Add `Type::getKeysArray()` and `Type::getValuesArray()`

herndlm:
low-hanging fruit cleanup / improvement :)

- fixes `ArrayType::getKeysArray()` (by using `getIterableKeyType()` mixed keys are correctly transformed into benevolent unions)
- simplifies list type creation for `array_is_list`
- gives us support for constant array unions for free (`instanceof ConstantArrayType` be gone! one of the main reasons for all this refactoring)

### DIFF
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -144,6 +144,16 @@ class MixedType implements CompoundType, SubtractableType
 		return $this;
 	}
 
+	public function getKeysArray(): Type
+	{
+		return AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new UnionType([new IntegerType(), new StringType()])));
+	}
+
+	public function getValuesArray(): Type
+	{
+		return AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new MixedType($this->isExplicitMixed)));
+	}
+
 	public function flipArray(): Type
 	{
 		return new ArrayType(new MixedType($this->isExplicitMixed), new MixedType($this->isExplicitMixed));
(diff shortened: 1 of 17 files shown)


src/Type/MixedType.php
```diff
@@ -144,6 +144,16 @@ public function unsetOffset(Type $offsetType): Type
...
 
+	public function getKeysArray(): Type
+	{
+		return AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new UnionType([new IntegerType(), new StringType()])));
+	}
+
+	public function getValuesArray(): Type
+	{
+		return AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new MixedType($this->isExplicitMixed)));
+	}
```
staabm, line 155: I think ~~these 2~~ all these array methods could support a subtracted array type

ondrejmirtes:
Thank you!

src/Type/MixedType.php
herndlm: They are very forgiving and follow https://github.com/phpstan/phpstan-src/pull/1850#discussion_r996297966 now by returning an array already which is preciser than mixed ~array

staabm: I think, taking `subtracted` into account it should get `*NEVER*` in this case
https://phpstan.org/r/918a37d5-92cf-4010-9dd8-d4ebbc315274

herndlm: Ah, I misunderstood you. That indeed makes sense

herndlm: isArray()->no() should be the check we need. Feel free to adapt it if you want, if not I can check it out in the evening or tomorrow 😊
Update: I think the new methods should all error return error than as the traits do

staabm: I will give it a try in https://github.com/phpstan/phpstan-src/pull/1852

PR #1852, opened 2022-10-16
MixedType: support subtracted in array-type methods

staabm:


src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
```diff
@@ -29,13 +29,12 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 		$arrayArg = $functionCall->getArgs()[0]->value ?? null;
 		if ($arrayArg !== null) {
```
staabm: I thought we might be able to reduce `array_keys` and `array_values` return-type-extension to something similar to

https://github.com/phpstan/phpstan-src/blob/949a7eaae5bec315663b74b239ddc1ba9c1c770b/src/Type/Php/ArrayFlipFunctionReturnTypeExtension.php#L20-L27

did not found a place yet, where to put the remaing non-empty-list logic though

src/Type/MixedType.php
```diff
@@ -146,16 +146,28 @@ public function unsetOffset(Type $offsetType): Type
 
 	public function getKeysArray(): Type
 	{
+		if ($this->isArray()->no()) {
+			return new NeverType();
```
herndlm: I had the same thought and next I asked myself why we don't use that in the traits as well. Maybe because it would be ignored in a union, but actually I'm not sure. I was just wondering if we should have the same in both or not 🤔 but here it's safer to use it I guess

src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
```diff
@@ -29,13 +29,12 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-				if ($valueType->isIterableAtLeastOnce()->yes()) {
-					$array = TypeCombinator::intersect($array, new NonEmptyArrayType());
-				}
-				return $array;
+
+			$array = $valueType->getKeysArray();
+			if ($valueType->isArray()->yes() && $valueType->isIterableAtLeastOnce()->yes()) {
+				$array = TypeCombinator::intersect($array, new NonEmptyArrayType());
 			}
+			return $array;
```
ondrejmirtes: This should already work just by returning `$valueType->getKeysArray();`.

Because `getKeysArray` looks correctly implemented in all ArrayType, IntersectionType, NonEmptyArrayType.

src/Type/Php/ArrayValuesFunctionDynamicReturnTypeExtension.php
```diff
@@ -28,13 +28,12 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-			if ($valueType->isArray()->yes()) {
-				$array = $valueType->getValuesArray();
-				if ($valueType->isIterableAtLeastOnce()->yes()) {
-					$array = TypeCombinator::intersect($array, new NonEmptyArrayType());
-				}
-				return $array;
+
+			$array = $valueType->getValuesArray();
+			if ($valueType->isArray()->yes() && $valueType->isIterableAtLeastOnce()->yes()) {
+				$array = TypeCombinator::intersect($array, new NonEmptyArrayType());
```
ondrejmirtes: dtto

src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
herndlm: I think it's still needed to handle the HasOffset* accessories. If they are set we have non-empty keys/values. So it could be added there, but I didn't do it yet, because I wasn't sure if it's really the right place

tests/PHPStan/Analyser/data/array_values.php
```diff
@@ -0,0 +1,31 @@
...
+
+class HelloWorld
+{
+	public function foo1($mixed): void
+	{
+		if(is_array($mixed)) {
+			assertType('list<mixed>', array_values($mixed));
+		} else {
+			assertType('mixed~array', $mixed);
+			assertType('*NEVER*', array_values($mixed));
```
ondrejmirtes: This Never/Error inconsistency is annoying, I'd rather return ErrorType.

ondrejmirtes (changes requested):

src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
herndlm: It needs still special handling in the HasOffset* types AFAIK, see https://github.com/phpstan/phpstan-src/pull/1852#discussion_r996412603

Basically they'd need to return non-empty-array and I wasn't sure if that really makes sense. But then it should work 😊

ondrejmirtes: Yeah, I think it does make sense :)

herndlm:
I like how we iteratively improve things with those array functions. At first I wasn't sure if moving all those "special function related things" to Type really is a good idea. But now with each change I see improvements or it unblocks other cleanups 😊

ondrejmirtes:
This is the way!

herndlm:
sorry @staabm, can you adapt popArray() and shiftArray() here too after a rebase? they were just merged :)

src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
staabm: fixed

staabm:
> can you adapt popArray() and shiftArray() here too after a rebase

I tried, but the situation is a bit different it seems.

this is for now good to go - as is.

ondrejmirtes:
THank you.

PR #1853, opened 2022-10-16
Add `Type::shuffleArray()`

herndlm:


src/Analyser/NodeScopeResolver.php
```diff
@@ -1955,12 +1955,7 @@ static function (?Type $offsetType, Type $valueType, bool $optional) use (&$arra
...
 			if (isset($functionReflection) && $functionReflection->getName() === 'shuffle') {
 				$arrayArg = $expr->getArgs()[0]->value;
-				$arrayArgType = $scope->getType($arrayArg);
-
-				if ($arrayArgType instanceof ConstantArrayType) {
-					$arrayArgType = $arrayArgType->getValuesArray()->generalizeToArray();
-				}
-
+				$arrayArgType = $scope->getType($arrayArg)->shuffleArray();
 				$scope = $scope->assignExpression($arrayArg, $arrayArgType, $arrayArgType);
```
herndlm: based on your last comment in the popArray() PR - not sure what should be done with the native type here. also getting it and apply the method? definitely more tests for that are needed :)

src/Type/Accessory/HasOffsetType.php
```diff
@@ -136,6 +136,11 @@ public function unsetOffset(Type $offsetType): Type
 		return $this;
 	}
 
+	public function shuffleArray(): Type
+	{
+		return $this;
+	}
```
herndlm: missed this in flipArray() for HasOffsetValueType, will open PR for it later. it is time to add those trait rules you mentioned :/

herndlm: But here it's actually wrong like this :) was in a hurry and will fix it later

src/Analyser/NodeScopeResolver.php
ondrejmirtes: The last arg should be `$scope->getNativeType($arrayArg)->shuffleArray()`.

You can test it with assertNativeType.

It isn't always less precise, there are situations when we know ConstantArrayType as native type too, for example after `$a = [...];`

But assertNativeType should be different when the argument type comes from PHPDoc (array shape).

tests/PHPStan/Analyser/data/shuffle.php
```diff
@@ -12,69 +13,115 @@ public function normalArrays(array $arr): void
...
 		shuffle($arr);
 		assertType('array{}', $arr);
+		assertNativeType('array{}', $arr);
 		assertType('array{}', array_keys($arr));
 		assertType('array{}', array_values($arr));
 
 		/** @var array{0?: 1, 1?: 2, 2?: 3} $arr */
 		shuffle($arr);
 		assertType('array<0|1|2, 1|2|3>&list', $arr);
+		assertNativeType('array', $arr);
```
ondrejmirtes: That's it! 🎉

ondrejmirtes (approved):

tests/PHPStan/Analyser/data/shuffle.php
herndlm: Ups, linting error..

src/Type/ArrayType.php
```diff
@@ -379,6 +379,11 @@ public function shiftArray(): Type
 		return $this;
 	}
 
+	public function shuffleArray(): Type
+	{
+		return $this;
```
ondrejmirtes: I just checked what shuffle() does in PHP documentation. Sounds like it's always gonna make a list - reindexes the keys. Can you check and fix this method in ArrayType? Thanks.

src/Type/MixedType.php
```diff
@@ -169,6 +169,15 @@ public function shiftArray(): Type
 		return new ArrayType(new MixedType($this->isExplicitMixed), new MixedType($this->isExplicitMixed));
 	}
 
+	public function shuffleArray(): Type
+	{
+		if ($this->isArray()->no()) {
+			return new ErrorType();
+		}
+
+		return new ArrayType(new MixedType($this->isExplicitMixed), new MixedType($this->isExplicitMixed));
```
ondrejmirtes: Maybe it's gonna be a list?

ondrejmirtes (changes requested):

src/Type/ArrayType.php
herndlm: yes it does.. thx for checking this, I really shouldn't have done this in such a hurry :)

herndlm:
I'm gonna wait for @staabm's PR first to get merged and rebase and check this one then 😊

src/Analyser/NodeScopeResolver.php
```diff
@@ -1955,13 +1955,11 @@ static function (?Type $offsetType, Type $valueType, bool $optional) use (&$arra
 
 			if (isset($functionReflection) && $functionReflection->getName() === 'shuffle') {
 				$arrayArg = $expr->getArgs()[0]->value;
-				$arrayArgType = $scope->getType($arrayArg);
-
-				if ($arrayArgType instanceof ConstantArrayType) {
-					$arrayArgType = $arrayArgType->getValuesArray()->generalizeToArray();
```
herndlm, line 1961: fyi this was the one and only public usage of `ConstantArrayType::generalizeToArray()`. I introduced it via 940b386a to address `shuffle()`.

ondrejmirtes:
You can rebase now, if you meant the PR I just merged :)

herndlm:
ok thx, should be ready now. sorry, I couldn't resist moving and grouping the 3 array HasOffsetValueType methods to the same place were they are everywhere else basically.

ondrejmirtes:
Thank you!

The generalizeToArray() method should probably be deprecated. In order to not increase the baseline, feel free to copy the logic to the one place where the method is called inside ConstantArrayType, so the deprecated method can simply be removed for 2.0.

PR #1856, opened 2022-10-17
Deprecate `ConstantArrayType::generalizeToArray()`

herndlm:
as requested in https://github.com/phpstan/phpstan-src/pull/1853#issuecomment-1280414716


ondrejmirtes:
Thank you :)

PR #1857, opened 2022-10-17
Improve impossible type checker for void-returning functions

rvanvelzen:
Fix phpstan/phpstan#8169

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -51,6 +51,7 @@ parameters:
 		curlSetOptTypes: false
 		listType: false
 		missingMagicSerializationRule: false
+		nullContextForVoidReturningFunctions: false
 	fileExtensions:
 		- php
 	checkAdvancedIsset: false
@@ -261,6 +262,7 @@ parametersSchema:
 		curlSetOptTypes: bool()
 		listType: bool()
 		missingMagicSerializationRule: bool()
+		nullContextForVoidReturningFunctions: bool()
 	])
 	fileExtensions: listOf(string())
 	checkAdvancedIsset: bool()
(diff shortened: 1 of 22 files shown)


ondrejmirtes (changes requested):
Super-interesting! I'd probably put this behind a bleedingEdge toggle.

The `@phpstan-assert` functionality is enabled for everyone because it can be considered just a smarter type inference, but this bit feels a little bit extra and can result in some extra errors, so bleedingEdge would feel safer for me.

conf/config.neon
```diff
@@ -51,6 +51,7 @@ parameters:
 		curlSetOptTypes: false
 		listType: false
 		missingMagicSerializationRule: false
+		nullContextForVoidReturningFunctions: true
```
ondrejmirtes: false needs to be here

ondrejmirtes (changes requested):

herndlm:
wow you were fast 🎉
but is this safe? e.g. can it be assumed that `@phpstan-assert string $var` internally basically does `is_string()`? I'm just afraid that it could cause false-positives when people use such an assert for something like `assertEmail()` basically.

ondrejmirtes:
@herndlm Yes, Richard expressed this concern earlier but I said to try it. That's why it's in bleedingEdge now only.

ondrejmirtes:
@rvanvelzen I think this one is ready to merge. I really appreciate all the work you do, but I feel like you're overusing the Draft feature a bit :) I respect it and I wouldn't merge a PR in draft state, but sometimes I'm eager to merge something because to my eyes it seems totally fine, so don't hesitate to mark things as ready sooner. We can always fix this in the future.

Another problem is that GitHub doesn't notify me when something's marked as ready so I added a workflow that adds a bot comment to such PRs: https://github.com/phpstan/phpstan-src/commit/e3b97716b77a7b04319c19cda258b7f4f30b7304

ondrejmirtes:
Thank you!

PR #1858, opened 2022-10-17
Improve `HasOffsetValueType::flipArray()`

herndlm:
If value is a constant string or constant int we can flip it with offset :)

### DIFF
--- a/src/Type/Accessory/HasOffsetValueType.php
+++ b/src/Type/Accessory/HasOffsetValueType.php
@@ -4,6 +4,7 @@ namespace PHPStan\Type\Accessory;
 
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\CompoundType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
@@ -163,6 +164,16 @@ class HasOffsetValueType implements CompoundType, AccessoryType
 	public function getValuesArray(): Type
 	{
 		return new NonEmptyArrayType();
+	}
+
+	public function flipArray(): Type
+	{
+		$valueType = ArrayType::castToArrayKeyType($this->valueType);
+		if ($valueType instanceof ConstantIntegerType || $valueType instanceof ConstantStringType) {
+			return new self($valueType, $this->offsetType);
+		}
+
+		return new MixedType();
 	}
 
 	public function shuffleArray(): Type
--- a/tests/PHPStan/Analyser/data/array-flip.php
+++ b/tests/PHPStan/Analyser/data/array-flip.php
@@ -76,3 +76,30 @@ function foo9($mixed)
 		assertType('*ERROR*', array_flip($mixed));
 	}
 }
+
+/** @param array<string, int> $array */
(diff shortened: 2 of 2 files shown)


src/Type/Accessory/HasOffsetValueType.php
```diff
@@ -165,6 +165,15 @@ public function getValuesArray(): Type
 		return new NonEmptyArrayType();
 	}
 
+	public function flipArray(): Type
+	{
+		if ($this->valueType instanceof ConstantIntegerType || $this->valueType instanceof ConstantStringType) {
```
ondrejmirtes: Do `ArrayType::castToArrayKeyType()` first :)

ondrejmirtes (changes requested):

src/Type/Accessory/HasOffsetValueType.php
herndlm: adapted in a force-push. I also found another special case where the same value is existing twice that does not behave correct at them moment. I'm not sure which code's responsibility is it to figure that out or if it's a big problem, but I documented it in the test. I'm talking about https://3v4l.org/1TAFk

herndlm: maybe that's a bug in `TypeCombinator::intersect()` with `HasOffsetValueType`

herndlm: hmm no, the intersection is right, this is NEVER. it can't know that PHP's flip does something weird like that hmm

herndlm: I think this could be implemented by doing `$type->uniqueArray($type->reverseArray())->flipArray()` :D see https://3v4l.org/EbUU1

but both unique and reverse are not existing yet, so I could look into this later maybe, or we keep it open for now and I adapt it when the time comes. it's a bit special for sure though

see also docs https://www.php.net/manual/en/function.array-flip.php
>  If a value has several occurrences, the latest key will be used as its value, and all others will be lost.

herndlm: reverse is the wrong thing then, this will mess with order obviously. so it'd need to done differently. since `ConstantArray::flipArray()` is implemented correctly I'm really not sure how to do this with multiple `HasOffsetValueType`. maybe it's not feasible

ondrejmirtes: A bug like that is fine for me now :)

ondrejmirtes:
Thank you!

staabm:
@herndlm can/should we do the same for `array_keys` und `array_values`?

herndlm:
> @herndlm can/should we do the same for `array_keys` und `array_values`?

I didn't think this through yet. I'm not sure if something similar can be done for those.
E.g. if we have `array<mixed>&hasOffsetValue('foo', 17)` and
- we call `array_keys()` we know that it's a non-empty-list and there is the value 'foo' somewhere inside that list (as value!)
- we call `array_values()` we know that it's a non-empty-list and there is the value 17 somewhere inside that list (as value!)


right? and that `HasValueType` thing is not existing _yet_ :)

it could be done for a list in theory, but that list is then a `ConstantArrayType` I guess and there is no `HasOffsetValueType`. But feel free to explore this further and if you find something please go ahead 😊

PR #1862, opened 2022-10-17
Fix FunctionReflection::getDocCmments() regressions

staabm:
as pointed out by Richard in https://github.com/phpstan/phpstan/issues/8175#issuecomment-1280728368

closes https://github.com/phpstan/phpstan/issues/8175

### DIFF
--- a/tests/PHPStan/Rules/Methods/data/bug-7519.php
+++ b/tests/PHPStan/Rules/Methods/data/bug-7519.php
@@ -1,0 +1,41 @@
+<?php declare(strict_types=1);
+
+namespace Bug7519;
+
+use function PHPStan\Testing\assertType;
+use FilterIterator;
+
+/**
+ * @template TKey
+ * @template T
+ * @template TIterator as Iterator<TKey, T>
+ *
+ * @extends FilterIterator<TKey, T, TIterator>
+ */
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Tests please. Judging from those issues, it should be enough to test that "has invalid return type" message is not produced.

tests/PHPStan/Rules/Methods/data/bug-7519.php
```diff
@@ -0,0 +1,41 @@
...
+	public function accept(): bool
+	{
+		return true;
+	}
+
+	public function key()
+	{
+		$key = parent::key();
+
+		assertType('TKey (class A, argument)', $key);
```
ondrejmirtes: This assertType isn't going to be checked. Because you're not referencing this file from NodeScopeResolverTest.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Methods/data/bug-7519.php
staabm: sorry, shouldn't do this things in a hurry. pushed a fix.

ondrejmirtes:
Thank you.

PR #1864, opened 2022-10-17
Explicitly specify native type for `array_pop`/`array_shift` arg

herndlm:
As it was felt in https://github.com/phpstan/phpstan-src/pull/1847#discussion_r996431484 :) and we also did it in https://github.com/phpstan/phpstan-src/pull/1853 in the mean time


ondrejmirtes:
Thank you.

PR #1485, opened 2022-06-30
Update `FilterIterator` stub and tests.

drupol:
This PR:

* Is related to https://github.com/JetBrains/phpstorm-stubs/pull/1399
* Is related to https://github.com/phpstan/phpstan/issues/7519

PR #1865, opened 2022-10-17
Regression tests

staabm:
closes https://github.com/phpstan/phpstan/issues/7519
closes https://github.com/phpstan/phpstan-src/pull/1485

took the changes from https://github.com/phpstan/phpstan-src/pull/1485, so kudos to @drupol

### DIFF
--- a/tests/PHPStan/Analyser/NodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/NodeScopeResolverTest.php
@@ -1083,6 +1083,7 @@ class NodeScopeResolverTest extends TypeInferenceTestCase
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/array_keys.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/../Rules/Methods/data/bug-8174.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/../Rules/Comparison/data/bug-8169.php');
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-7519.php');
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


drupol:
Thanks !

ondrejmirtes (changes requested):
This issue is about false positive when returning from a method. So this needs ReturnTypeRuleTest case: https://github.com/phpstan/phpstan/issues/7519#issuecomment-1187671983

tests/PHPStan/Rules/Functions/ReturnTypeRuleTest.php
```diff
@@ -181,4 +181,11 @@ public function testListWithNullablesUnchecked(): void
 		$this->analyse([__DIR__ . '/data/return-list-nullables.php'], []);
 	}
 
+	public function testBug7519(): void
+	{
+		$this->checkExplicitMixed = true;
+		$this->checkNullables = true;
+		$this->analyse([__DIR__ . '/data/bug-7519.php'], []);
```
ondrejmirtes: functions vs. methods :P Please be less sloppy.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Functions/ReturnTypeRuleTest.php
staabm: sorry. will take a break now.

ondrejmirtes:
Thank you :) I don't trust a test I haven't seen fail, so part of writing a regression test is trying it out with a reverted commit that fixed the problem.

staabm:
> so part of writing a regression test is trying it out with a reverted commit that fixed the problem.

I run the testfiles usually using an older phpstan release to verify they fail - and I did so here.
but did not see the Function vs. Method difference

ondrejmirtes:
It couldn't fail for you if you tested method return type in `ReturnTypeRuleTest` for functions.

PR #1866, opened 2022-10-17
Add support for mb_convert_case() and mb_convert_kana()

zonuexe:
Type [`mb_convert_case()`](https://php.net/mb_convert_case) and [`mb_convert_kana()`](https://php.net/mb_convert_case) with `StrCaseFunctionsReturnTypeExtension`.

Strictly speaking, Japanese [kana characters](https://en.wikipedia.org/wiki/Kana) are not casing, but many of the treatments are common.

PR #1867, opened 2022-10-17
mb_check_encoding() accepts array<string> in PHP 7.2+

zonuexe:
refs https://github.com/phpstan/phpstan-src/pull/1866#discussion_r997124625

https://3v4l.org/GlVUu


ondrejmirtes:
Thank you!

PR #1869, opened 2022-10-18
Add `Type::fillKeysArray()`

herndlm:
I did not change behaviour on purpose, but was wondering why the handling of `ArrayType` and `ConstantArrayType` is different for keys that cannot be casted to string. And also why we return a `NeverType` instead of a `ErrorType` 🤔
e.g. things that cannot be casted to a string will make it fail hard https://3v4l.org/gBHIk

### DIFF
--- a/src/Type/ArrayType.php
+++ b/src/Type/ArrayType.php
@@ -367,6 +367,17 @@ class ArrayType implements Type
 		return $this;
 	}
 
+	public function fillKeysArray(Type $valueType): Type
+	{
+		$itemType = $this->getIterableValueType();
+
+		if ((new IntegerType())->isSuperTypeOf($itemType)->no() && !$itemType->toString() instanceof ErrorType) {
+			return new ArrayType($itemType->toString(), $valueType);
+		}
+
+		return new ArrayType($itemType, $valueType);
+	}
+
 	public function flipArray(): Type
 	{
 		return new self($this->getIterableValueType()->toArrayKey(), $this->getIterableKeyType());
--- a/src/Type/ArrayType.php
+++ b/src/Type/ArrayType.php
@@ -369,10 +369,14 @@ class ArrayType implements Type
 
 	public function fillKeysArray(Type $valueType): Type
 	{
-		$itemType = $this->getIterableValueType();
+		$itemType = $this->getItemType();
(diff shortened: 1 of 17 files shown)


src/Type/ArrayType.php
```diff
@@ -367,6 +367,17 @@ public function unsetOffset(Type $offsetType): Type
 		return $this;
 	}
 
+	public function fillKeysArray(Type $valueType): Type
+	{
+		$itemType = $this->getIterableValueType();
+
+		if ((new IntegerType())->isSuperTypeOf($itemType)->no() && !$itemType->toString() instanceof ErrorType) {
+			return new ArrayType($itemType->toString(), $valueType);
+		}
```
herndlm: not sure how heavy `->toString()` is, but we could assign it to a var here and in `ConstantArrayType` maybe to avoid calling it twice

src/Type/MixedType.php
```diff
@@ -162,6 +162,15 @@ public function getValuesArray(): Type
 		return AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new MixedType($this->isExplicitMixed)));
 	}
 
+	public function fillKeysArray(Type $valueType): Type
+	{
+		if ($this->isArray()->no()) {
+			return new ErrorType();
```
herndlm, line 168: Maybe an invalid keytype should be handled as error here too?

herndlm: No, that is irrelevant I think because we don't know anything about the key = it is mixed

src/Type/ArrayType.php
herndlm: adapted in the second commit

herndlm:
added a second commit that makes it return `*ERROR*` consistently

herndlm:
hmm, weird the tests were a bit flaky here and I had to force-push multiple times. saw random-looking timeouts, but also once the following error in-between for example
> Ignored error pattern #^Call to an undefined method TestResultCache1\\Foo\:\:doFoo\(\)\.$# in path /home/runner/work/phpstan-src/phpstan-src/e2e/result-cache-1/src/Baz.php was not matched in reported errors.

ondrejmirtes:
Yes, it's a mystery for me too: https://github.com/orgs/community/discussions/36656

ondrejmirtes:
Thank you!

PR #1550, opened 2022-07-25
Use Type::isArray() in InvalidKeyInArrayDimFetchRule

herndlm:
Small cleanup since the same information should be already existing on the type itself.

PR #1870, opened 2022-10-18
Get rid of of `Type::getArrays()` usage in `InvalidKeyInArrayDimFetchRule`

herndlm:


herndlm:
come on, I tried to do the same thing already in https://github.com/phpstan/phpstan-src/pull/1550 :D

PR #1872, opened 2022-10-19
Get rid of of `Type::getArrays()` usage in `InvalidKeyInArrayDimFetchRule`

herndlm:
this thing just keeps annoying me and I want to do something about it. of course I dislike the `instanceof MixedType`, but maybe someone has a better idea?

### DIFF
--- a/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
+++ b/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
@@ -8,6 +8,7 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
 use function sprintf;
@@ -37,6 +38,10 @@ class InvalidKeyInArrayDimFetchRule implements Rule
 		}
 
 		$dimensionType = $scope->getType($node->dim);
+		if ($dimensionType instanceof MixedType) {
+			return [];
+		}
+
 		$varType = $this->ruleLevelHelper->findTypeToCheck(
 			$scope,
 			$node->var,
--- a/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
+++ b/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
@@ -8,7 +8,6 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\VerbosityLevel;
-use function count;
 use function sprintf;
 
 /**
@@ -33,7 +32,7 @@ class InvalidKeyInArrayDimFetchRule implements Rule
 		}
 
 		$varType = $scope->getType($node->var);
-		if (count($varType->getArrays()) === 0) {
+		if ($varType->isArray()->no() || $varType instanceof MixedType) {
 			return [];
 		}
 
--- a/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
+++ b/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
@@ -8,7 +8,6 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\Type\ErrorType;
-use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
 use function sprintf;
@@ -50,21 +49,15 @@ class InvalidKeyInArrayDimFetchRule implements Rule
 		}
 
 		$isSuperType = AllowedArrayKeysTypes::getType()->isSuperTypeOf($dimensionType);
-		if ($isSuperType->no()) {
-			return [
-				RuleErrorBuilder::message(
-					sprintf('Invalid array key type %s.', $dimensionType->describe(VerbosityLevel::typeOnly())),
-				)->build(),
-			];
-		} elseif ($this->reportMaybes && $isSuperType->maybe() && !$dimensionType instanceof MixedType) {
-			return [
-				RuleErrorBuilder::message(
-					sprintf('Possibly invalid array key type %s.', $dimensionType->describe(VerbosityLevel::typeOnly())),
-				)->build(),
-			];
+		if ($isSuperType->yes() || ($isSuperType->maybe() && !$this->reportMaybes)) {
+			return [];
 		}
 
-		return [];
+		return [
+			RuleErrorBuilder::message(
+				sprintf('%s array key type %s.', $isSuperType->no() ? 'Invalid' : 'Possibly invalid', $dimensionType->describe(VerbosityLevel::typeOnly())),
+			)->build(),
+		];
 	}
 
 }
--- a/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
+++ b/src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
@@ -6,7 +6,10 @@ use PhpParser\Node;
 use PHPStan\Analyser\Scope;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Rules\RuleLevelHelper;
+use PHPStan\Type\ErrorType;
 use PHPStan\Type\MixedType;
+use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
 use function sprintf;
 
@@ -16,7 +19,10 @@ use function sprintf;
 class InvalidKeyInArrayDimFetchRule implements Rule
 {
 
-	public function __construct(private bool $reportMaybes)
+	public function __construct(
+		private RuleLevelHelper $ruleLevelHelper,
+		private bool $reportMaybes,
+	)
 	{
 	}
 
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Rules/Arrays/data/invalid-key-array-dim-fetch.php
```diff
@@ -35,3 +35,14 @@
...
+/** @var int|object $maybeInt */
+$maybeInt = null;
+
+/** @var string|array $maybeString */
+$maybeString = [];
+echo $maybeString[$maybeInt];
+
+/** @var mixed $mixed */
+$mixed = null;
+echo $mixed[$maybeInt];
```
herndlm: the first one was reported, the second one not (before and after this PR I mean). and the `instanceof MixedType` makes sure to keep that behaviour..

ondrejmirtes: This one should be reported on level 9 though. We can fix that.

herndlm: right, there is another rule complaining but not this one. https://phpstan.org/r/bc4b663d-fa12-4037-aed2-e0ed6d2bb53c

just using ->no() and removing the instanceof is not the fix is it? is level 9 basically explicit mixed, right? strictmixed was something else I suppose

ondrejmirtes: Alright, so the problem is that array dim fetch dim type other than `int|string` might be valid for things that aren't arrays. I think that simply `$varType->isArray()->no()` is correct here. If it might be an array, let's complain about the possibly invalid key.

The example in https://phpstan.org/r/bc4b663d-fa12-4037-aed2-e0ed6d2bb53c is about the `$mixed` part in `$mixed[$maybeInt]`, that rule is gonna complain about `$mixed[$int]` too.

I think that in case of `$mixed[$maybeInt]`, both rules should complain.

So the `$varType instanceof MixedType` is totally unnecessary and also possibly incorrect for subtractions. As we're learning again and again, `instanceof *Type` is always wrong :)

What the problem here is right now is that InvalidKeyInArrayDimFetchRule doesn't use RuleLevelHelper so it's gonna complain about everything right from the start. We should employ `RuleLevelHelper::findTypeToCheck()` here too.

herndlm: ok, that makes sense! I added another commit that introduces usage of it, but I have no experience with it. So I'm not sure if e.g. the `unionTypeCriteriaCallback` is right and also not sure if more tests need to be added with different RuleLevelHelper configuration for example

ondrejmirtes: The usage isn't right, and you can test that :) Some individual level 9 bugs are usually tested in tests that have `private bool $checExplicitMixed` property, but if I implement something from scratch, it belongs to LevelsIntegrationTest - just write the `.php` file, add it to `dataTopics` and run the test, it will save the current JSONs for you. And you should probably comment out the other topics meanwhile otherwise it's gonna run for a long time :)

herndlm: it looks like the `stringOffsetAccess` `LevelsIntegrationTest` still complains on level 7. I must be missing something then hmm

ondrejmirtes: The pattern for using `findTypeToCheck is`:

```
		$typeResult = $this->ruleLevelHelper->findTypeToCheck(
			$scope,
			$node->expr,
			'', // unknown class error message pattern
			static function (Type $type) use ($castTypeCallback): bool {
				// check something on $type, return false if it's an error
			},
		);
		$type = $typeResult->getType();
		if ($type instanceof ErrorType) { // this part is always the same, some rules report the unknown class reports here
			return [];
		}

		// check the same thing on $type here, report errors if it's false
```

You can check for example InvalidCastRule.

herndlm: ah I think this is related to `findTypeToCheck()` returning `ErrorType`. I'll look closely later

herndlm: we wrote almost at the same time and github didn't show me your comments, so it looks like we kind of ignored each other, weird.
anyways, I adapted it to handle the ErrorType and kept the code that aborts if it isn't an array. it now behaves exactly like on 1.9.x, meaning it ignores mixed basically. I couldn't figure out easily how to let it report for StrictMixed, but I'm not sure if it makes sense tbh. It looks like this rule should only work on things that can be arrays (except implicit mixed basically) and explicit mixed is of course no array.

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -32,12 +37,21 @@ public function processNode(Node $node, Scope $scope): array
...
-		$varType = $scope->getType($node->var);
-		if (count($varType->getArrays()) === 0) {
+		$dimensionType = $scope->getType($node->dim);
+		$varType = $this->ruleLevelHelper->findTypeToCheck(
+			$scope,
+			$node->var,
+			'',
+			static function (Type $varType) use ($dimensionType): bool {
+				$arrayDimType = $varType->setOffsetValueType($dimensionType, new MixedType());
+				return !($arrayDimType instanceof ErrorType);
```
ondrejmirtes: You're not doing this correctly. Either the code below has to also use $varType->setOffsetValueType, or the code in this callback has to use isArray(). The thing that's done in the callback and that's done below the RuleLevelHelper call has to match.

Also - if you're introducing RuleLevelHelper, you need to test the behaviour on various levels to see if it's correct :) I told you how in a comment above.

ondrejmirtes (changes requested):

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
herndlm: ah ok, sorry, I'll check this again. there is already the `stringOffsetAccess` levelstest, I thought this is testing this stuff already (because it was the one failing and this is also where I get the additional test cases for this rule from)

herndlm: I think I'm starting to understand how this works, but here it's hard for me to add more useful tests because on level 7 both checkUnionTypes and reportMaybes are enabled. anyways, I think the existing [`stringOffsetAccess`](https://github.com/phpstan/phpstan-src/blob/1.9.x/tests/PHPStan/Levels/data/stringOffsetAccess.php) levels test tests this already, what do you think?

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -32,12 +37,18 @@ public function processNode(Node $node, Scope $scope): array
...
+			$node->var,
+			'',
+			static fn (Type $varType): bool => $varType->isArray()->no() || AllowedArrayKeysTypes::getType()->isSuperTypeOf($dimensionType)->yes(),
+		)->getType();
+
+		if ($varType instanceof ErrorType || $varType->isArray()->no()) {
 			return [];
 		}
 
-		$dimensionType = $scope->getType($node->dim);
```
ondrejmirtes, line 40: The rest of this method should be:

```php
if ($isSuperType-yes()) {
    return [];
}

return [
	RuleErrorBuilder::message(
		sprintf('%s array key type %s.', $isSuperType->no() ? 'Invalid' : 'Possibly invalid', $dimensionType->describe(VerbosityLevel::typeOnly())),
	)->build(),
];

ondrejmirtes (changes requested):

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
herndlm: that looks better! adapted. but I had to keep the maybe-early-exit in to avoid the rule reporting maybes on level 3 already (the newly added levelstest case started failing)

herndlm: oh and I forgot to run make phpstan before pushing. need to check what's going on with mixed there which should be ignored still I guess

ondrejmirtes: It's a bit weird because RuleLevelHelper should have already taken care of the maybe case on lower levels.

herndlm: also had a feeling that the closure is maybe still wrong, but RuleLevelHelper doesn't seem to be dealing with checkMaybe? it deals with e.g. checkUnionTypes, but in my case the union is irrelevant, the int dim on a constantarray is a maybe :/

ondrejmirtes: Oh, you're right, the rule code needs to be a bit more complicated after all. Not sure what's with the mixed in `make phpstan`, PHPStan is running on level 8...

herndlm: the rule was explicitly ignoring mixed dims before as well when checking for maybe/potential invalid keys. maybe it can't be simplified after all. I was also looking forward to put all checks on top and into the closure

herndlm: the mixed dims that were reported seemed mostly legit, maybe this could even be turned on with bleedingEdge. but it's not related to this PR I think, I let the rule ignore them again. at least it's a bit simpler and makes use of RuleLevelHelper now hmm

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #1874, opened 2022-10-19
Add `Type::searchArray()`

herndlm:
Split out of the extension, but with a couple of small differences
- instead of returning `null` (happened on PHP < 8 with a warning, see https://3v4l.org/lNrqs) it returns `*ERROR*`. not sure if that's OK?
- support for `HasOffsetValueType` which wasn't understood before, see https://phpstan.org/r/e20727f4-7809-42fc-825e-85df0f5ef635
- support for optional array keys, this was incorrect before, see https://phpstan.org/r/96b559ec-55a2-4feb-b3a9-a1e4dff2a84b

not sure if `$strictType` should be moved into the method signature as well or not 🤔

### DIFF
--- a/src/Php/PhpVersion.php
+++ b/src/Php/PhpVersion.php
@@ -206,4 +206,9 @@ class PhpVersion
 		return $this->versionId >= 80100;
 	}
 
+	public function arrayFunctionsReturnNullWithNonArray(): bool
+	{
+		return $this->versionId < 80000;
+	}
+
 }
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -162,6 +162,11 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 		return $this;
 	}
 
+	public function searchArray(Type $needleType): Type
+	{
+		return new MixedType();
+	}
+
 	public function shiftArray(): Type
 	{
 		return $this;
--- a/src/Type/Accessory/HasOffsetValueType.php
+++ b/src/Type/Accessory/HasOffsetValueType.php
@@ -180,6 +180,18 @@ class HasOffsetValueType implements CompoundType, AccessoryType
 		return new MixedType();
 	}
 
+	public function searchArray(Type $needleType): Type
+	{
+		if (
+			$needleType instanceof ConstantScalarType && $this->valueType instanceof ConstantScalarType
+			&& $needleType->getValue() === $this->valueType->getValue()
+		) {
+			return $this->offsetType;
+		}
+
+		return new MixedType();
+	}
+
 	public function shuffleArray(): Type
 	{
 		return new NonEmptyArrayType();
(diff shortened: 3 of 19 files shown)


ondrejmirtes:
The null behaviour should be kept - if we're aware of these regressions we should fix them :) Actually it should be `null` before PHP 8 and `NeverType` on PHP 8+, right?

herndlm:
> The null behaviour should be kept - if we're aware of these regressions we should fix them :) Actually it should be `null` before PHP 8 and `NeverType` on PHP 8+, right?

yeah, adding the php-version check makes sense. regarding `NeverType` - I'm not sure if it should be `NeverType` or `ErrorType`, this would be relevant for more array functions most likely :/ e.g. `array_flip()`: https://3v4l.org/XnrUq which also shows the null behaviour again hmm
UDPATE: ah and never/error obviously is relevant when passing something like `false|arary` to `array_flip()` for example. feels like this needs some discussion / cleaning up again

ondrejmirtes:
The ErrorType doesn't do anything useful, it's basically a MixedType.

There's additional downside to ErrorType with arrays - because of some logic in NodeScopeResolver, if you add a new offset to ErrorType, a new array shape (ConstantArrayType) is created, leading to most likely wrong analysis...

NeverType at least signals that no code after this point is executed because it crashes/throws an exception.

herndlm:
I wonder if it makes sense to keep the error types in the traits/types and let only the extensions return null/never depending on php version in case is array()->no(). It still feels inconsistent, but maybe I just need to play around with it..

herndlm:
well, I think this is looking better, but it feels more correct to let the NonArrayTypeTrait return *NEVER* too, but that would cancel it out in a union I guess. on the other hand I'm not sure if *ERROR* is the right type there either, see e.g LegacyNodeScopeResolverTest diff. also: I can't check for the PHP version in types, can I? 🤔

herndlm:
Another idea: the nonarray trait could return a nulltype, which is what the array functions would do without enforcing the array arg type. The extension checks php version and isArray() and either returns nulltype/nevertype directly, does something different with maybe  or passes the type through for a yes(). That could improve the type over all php versions. Not sure yet what should happen with e.g. array|false on php 8+ though. Maybe errortype

ondrejmirtes:
I wouldn't pass PHP array functions weirdnesses into the Type system layer. Most likely it would lead to weird results in union/intersection types, and also it'd be misinformative for other uses...

herndlm:
What do you think of the current state then or any other ideas?

I'm not a 100% happy with it, unsure about unions mostly. But in order to keep the type for pre php 8 I'd need to modify the maybearraytypetrait somehow I guess. Which would mess with the type on php 8+. Except we change that and handle it maybe differently in the extension

herndlm:
Nevermind, I think I know how to improve it in the extension with a bit of extra code for old php. That shouldn't mess with the type system either. I'll try later :)

herndlm:
I give up, my super smart plan was to handle the `isArray()->maybe()` for pre PHP 8 and return the iterable key type union null, but since iterable key type is already `*ERROR*` the outcome is the same 🙃 I think `*ERROR*`  is just fine than. running out of ideas :)
it correctly deals with `*NEVER*` vs `null` for non-arrays, maybe that's as good as it can get. I'd do the same for the other array functions as well afterwards

ondrejmirtes:
Really nice, thank you 😊

PR #1878, opened 2022-10-20
emit warnings when xdebug is not used according to cli-flags

staabm:
as discussed in https://github.com/phpstan/phpstan/discussions/8173

we don't have tests with/without xdebug. does this PR require tests?

I tested it manually
```
$ php bin/phpstan analyze test.php --debug  --level 9
Note: The xdebug php-extension is active, but "--xdebug" is not used.
      This may slow down performance and the process will not halt at breakpoints.
Note: Using configuration file C:\dvl\Workspace\phpstan-src-staabm\phpstan.neon.dist.
C:\dvl\Workspace\phpstan-src-staabm\test.php


$ php bin/phpstan analyze test.php --debug  --level 9   --xdebug
Note: You are running with "--xdebug" enabled, but the xdebug php-extension is not active.
      The process will not halt at breakpoints.
Note: Using configuration file C:\dvl\Workspace\phpstan-src-staabm\phpstan.neon.dist.
C:\dvl\Workspace\phpstan-src-staabm\test.php
```

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -91,13 +91,6 @@ class CommandHelper
 		bool $cleanupContainerCache = true,
 	): InceptionResult
 	{
-		if (!$allowXdebug) {
-			$xdebug = new XdebugHandler('phpstan');
-			$xdebug->setPersistent();
-			$xdebug->check();
-			unset($xdebug);
-		}
-
 		$stdOutput = new SymfonyOutput($output, new SymfonyStyle(new ErrorsConsoleStyle($input, $output)));
 
 		/** @var Output $errorOutput */
@@ -105,6 +98,20 @@ class CommandHelper
 			$symfonyErrorOutput = $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output;
 			return new SymfonyOutput($symfonyErrorOutput, new SymfonyStyle(new ErrorsConsoleStyle($input, $symfonyErrorOutput)));
 		})();
+
+		if ($allowXdebug && !XdebugHandler::isXdebugActive()) {
+			$errorOutput->getStyle()->note('You are running with "--xdebug" enabled, but the Xdebug PHP extension is not active. The process will not halt at breakpoints.');
+		} elseif (!$allowXdebug && XdebugHandler::isXdebugActive()) {
+			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. This may slow down performance and the process will not halt at breakpoints.');
+		}
+
+		if (!$allowXdebug) {
+			$xdebug = new XdebugHandler('phpstan');
+			$xdebug->setPersistent();
+			$xdebug->check();
+			unset($xdebug);
+		}
+
 		if ($memoryLimit !== null) {
 			if (Strings::match($memoryLimit, '#^-?\d+[kMG]?$#i') === null) {
 				$errorOutput->writeLineFormatted(sprintf('Invalid memory limit format "%s".', $memoryLimit));


staabm:
(sorry for the typo in the PR title)

src/Command/CommandHelper.php
```diff
@@ -91,20 +91,27 @@ public static function begin(
...
 		$stdOutput = new SymfonyOutput($output, new SymfonyStyle(new ErrorsConsoleStyle($input, $output)));
 
 		/** @var Output $errorOutput */
 		$errorOutput = (static function () use ($input, $output): Output {
 			$symfonyErrorOutput = $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output;
 			return new SymfonyOutput($symfonyErrorOutput, new SymfonyStyle(new ErrorsConsoleStyle($input, $symfonyErrorOutput)));
 		})();
+
+		if ($allowXdebug && !XdebugHandler::isXdebugActive()) {
+			$errorOutput->writeLineFormatted(sprintf('Note: You are running with "--xdebug" enabled, but the xdebug php-extension is not active.' . "\n" . '      The process will not halt at breakpoints.'));
```
ondrejmirtes: 1) Please use `$errorOutput->getStyle()->note()` instead.
2) Correct capitalization of `Xdebug` please: https://xdebug.org/
3) `php-extension` should be `PHP extension`

ondrejmirtes (changes requested):

src/Command/CommandHelper.php
staabm: fixed thanks.

screenshot of the cli error for preview

grafik

ondrejmirtes:
Purrrrfect!

ondrejmirtes:
Thank you.

staabm:
off-topic: you may also notice, that my commits from the work pc, now also use the same git user as those from my personal computer

ondrejmirtes:
👏

InvisibleSmiley:
The problem with the current warning message is that it can be interpreted as "please add the --xdebug flag to improve performance" which is wrong of course.

See for example https://youtrack.jetbrains.com/issue/WI-69996

Please improve the message to help people understand that:
- if you want to improve performance, disable Xdebug altogether
- debugging PHPStan itself, use --xdebug with Xdebug enabled
- do not use --xdebug with Xdebug enabled for regular use since it really hurts performance

staabm:
@InvisibleSmiley feel free to suggest a pull request with a concrete message.

jorismak:
Shouldn't it be possible to disable this warning in the configuration file or something?

It now breaks integrations with phpstan because the warning is _always_ displayed. Setting 'start_with_request' to trigger still doesn't quiet the warning.

So on a development machine where you want Xdebug to be available, you now always get this warning? Seems like it now breaks more than it helps? (The performance difference with XDEBUG_MODE=off vs XDEBUG_MODE=debug is very, _very_ small).

ondrejmirtes:
What does it break? It's printed to stderr. And there have always been messages printed to stderr, even when combined with JSON error format for example.

jorismak:
I get a popup in phpstorm and vscode every time it tries to run phpstan, that there is some error running the process (and then shows the xdebug warning).

ondrejmirtes:
1) The same integration should fail if there's a note in the output about auto-discovered phpstan.neon or phpstan.neon.dist. Is that correct?
2) Seems like you can disable this message with an environment variable XDEBUG_MODE=off. Is that correct?

jorismak:
I guess these ide plugins have ignored certain output to not complain about
other stderr output. Since i almost always need to specify a config file i
haven't encountered that situation.

In composer.json you can use ***@***.***` in a script definition to set the
XDEBUG_MODE variable I've discovered.

And in phpstorm you can add commandline options to the php interpreter used
for things like phpstan and php-cs-fixer.
The option `xdebug.mode=off` has the same effect as the env variable
(phpstorm uses this with a -d commandline parameter).

So I'm able to work around it for now.
Still seems weird to have to do this, as phpstan is often used during
development, where xdebug is almost always enabled.

Quite easy to see situations where you have little control over the command
used to invoke phpstan when it's inside a plugin or something like that .
Having config options to quiet all warnings still sound like something that
might eventually be necessary.

Balkoth:
I guess this change breaks extensions like [this](https://github.com/breeze2/vscode-phpstan/issues/35)?

ondrejmirtes:
Extension that gets broken when something meant for humans to read is printed to `stderr` needs to be fixed :)

ondrejmirtes:
Anyway, if you have any questions, please open new issues or discussions in phpstan/phpstan. Thank you.

PR #1881, opened 2022-10-20
Remove is_bool, is_countable and is_null type specifying extensions.

mad-briller:
https://github.com/phpstan/phpstan/issues/8186


ondrejmirtes:
Thank you!

PR #1882, opened 2022-10-20
Improve wording of the NonexistentOffset, BooleanAndConstantConditionRule and BooleanOrConstantConditionRule

VincentLanglet:
Related to https://github.com/phpstan/phpstan/issues/8187#issuecomment-1286169876

I feel like 
```
Offset 'c' does not exist on array{c?: int}.
```
is confusing and 
```
Offset 'c' might not exist on array{c?: int}.
```
is better.

### DIFF
--- a/conf/config.level4.neon
+++ b/conf/config.level4.neon
@@ -45,6 +45,7 @@ services:
 		class: PHPStan\Rules\Comparison\BooleanAndConstantConditionRule
 		arguments:
 			treatPhpDocTypesAsCertain: %treatPhpDocTypesAsCertain%
+			bleedingEdge: %featureToggles.bleedingEdge%
 		tags:
 			- phpstan.rules.rule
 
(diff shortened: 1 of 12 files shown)


ondrejmirtes (changes requested):
We can't do this because it'd break existing baselines. We can do this only with a new bleedingEdge toggle.

A similar thing could be fixed in BooleanAndConstantConditionRule and BooleanOrConstantConditionRule where the message still uses `&&` and `||` even for logical `and` and `or` operators.

herndlm:
new bleeding edge toggle kind of means that it will get into the next major automatically, right?
sounds like a good solution then, otherwise it would need to be remembered / noted down :)

ondrejmirtes:
Exactly, and no rebase nightmare either.

VincentLanglet:
Yeah, I'll update the PR

VincentLanglet:
> We can't do this because it'd break existing baselines. We can do this only with a new bleedingEdge toggle.

Done

> A similar thing could be fixed in BooleanAndConstantConditionRule and BooleanOrConstantConditionRule where the message still uses `&&` and `||` even for logical `and` and `or` operators.

Done

ondrejmirtes:
Thank you.

PR #1884, opened 2022-10-21
Replace is_(int,object,resource,scalar) extensions with stubs

BackEndTea:
These were the ones i found that didn't contain special logic besides just specifying the types.


BackEndTea:
Looks like the test failures arent unique to this check.

I added another failing test case for `is_string`. 
Looks like for `is_int($value) === true` does not correcly set the `assert-if-false` data

rvanvelzen:
I'll be fixing the negation logic soon, which should resolve those failures.

rvanvelzen:
You can fix the issue by applying this patch:

```diff
diff --git a/src/Analyser/TypeSpecifier.php b/src/Analyser/TypeSpecifier.php
index c70b847a5..f04522ca5 100644
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1270,9 +1270,9 @@ class TypeSpecifier
        {
                if ($context->null()) {
                        $asserts = $assertions->getAsserts();
-               } elseif ($context->truthy()) {
+               } elseif ($context->true()) {
                        $asserts = $assertions->getAssertsIfTrue();
-               } elseif ($context->falsey()) {
+               } elseif ($context->false()) {
                        $asserts = $assertions->getAssertsIfFalse();
                } else {
                        throw new ShouldNotHappenException();
```

ondrejmirtes:
Thank you!

rvanvelzen:
The new stubs don't need the double assertion anymore though :)

ondrejmirtes:
Cleaned up :) https://github.com/phpstan/phpstan-src/commit/00d7cdbd8cdc7829f1be6e699f0b11b7d98cbb21

PR #1888, opened 2022-10-21
Implement equality docblock assertion syntax

rvanvelzen:


rvanvelzen:
Once this lands I can finish negating `assert-if-true`/`assert-if-false` properly :)

ondrejmirtes:
Thank you!

PR #1889, opened 2022-10-21
Infer negated context from assert-if-true/assert-if-false

rvanvelzen:
Per https://github.com/phpstan/phpstan/issues/8186#issuecomment-1285702156


ondrejmirtes:
Thank you!

PR #1890, opened 2022-10-21
fixed TypeCombinator->containsNull() with late resolvable type

staabm:
given a conditional type like
`@return ($subject is array ? array<string>|null : string|null)`

`TypeCombinator->containsNull()` was not able to identify the containing `NullType`.

I targeted 1.8.x since I consider it a bugfix.

### DIFF
--- a/tests/PHPStan/Type/TypeCombinatorTest.php
+++ b/tests/PHPStan/Type/TypeCombinatorTest.php
@@ -4248,4 +4248,22 @@ class TypeCombinatorTest extends PHPStanTestCase
 		$this->assertSame('array{0: string, 1?: string, 2?: string, 3?: string, 4?: string, test?: string}', $resultType->describe(VerbosityLevel::precise()));
 	}
 
+	/**
+	 * @dataProvider dataContainsNull
+	 */
+	public function testContainsNull(
+		Type $type,
+		bool $expectedResult,
+	): void
+	{
+		$this->assertSame($expectedResult, TypeCombinator::containsNull($type));
+	}
+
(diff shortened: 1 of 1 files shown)


staabm:
for context: I stumbled over this problem, since adding a stub for `preg_replace` like
```
/**
 * @param string|string[] $pattern
 * @param string|array<string|int|float> $replacement
 * @param string|array<string|int|float> $subject
 * @param int $count
 * @param-out 0|positive-int $count
 * @return ($subject is array ? array<string>|null : string|null)
 */
function preg_replace($pattern, $replacement, $subject, int $limit = -1, &$count = null) {}
```

lead to problems,  as the `null` values was no longer added here

https://github.com/phpstan/phpstan-src/blob/a4a5a3e8dc0314a7bd62aaad6116dbffb916d4f5/src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php#L58-L60

rvanvelzen:
@ondrejmirtes I tend to agree and I have thought about it before, but haven't come up with a clean solution yet.

staabm:
Shouldn't we instead just have `Type->containsNull()` to fix it?

ondrejmirtes:
I'd accept if the resolve method was called in the replace extension before passing the type to containsNull.

staabm:
reduced this PR to testing the current feature set only.

> I'd accept if the resolve method was called in the replace extension before passing the type to containsNull.

I guess this was meant for https://github.com/phpstan/phpstan-src/pull/1891, which I will adjust also

ondrejmirtes:
Thank you.

PR #1891, opened 2022-10-21
declare `@param-out` on preg_* and mb_parse_str()

staabm:
this PR currently depends on https://github.com/phpstan/phpstan-src/pull/1890 (or a similar fix for the underlying problem)

### DIFF
--- a/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
@@ -7392,11 +7392,11 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 				'$anotherExpectedArray',
 			],
 			[
-				'array|string',
+				'array<string>|string',
 				'$expectedArrayOrString',
 			],
 			[
-				'(array|string)',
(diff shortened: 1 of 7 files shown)


tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -9052,7 +9052,7 @@ public function dataGeneralizeScope(): array
 	{
 		return [
 			[
-				'array<string, non-empty-array<int|string, array{saveCount: int<0, max>, removeCount: int<0, max>, loadCount: int<0, max>, hitCount: int<0, max>}>>',
+				'array<non-empty-array<int|string, array{saveCount: int<0, max>, removeCount: int<0, max>, loadCount: int<0, max>, hitCount: int<0, max>}>>',
```
staabm, line 9055: I am not sure about this one regarding generalization

ondrejmirtes (changes requested):
Basically with all signatures here we should be able to say more about the array keys. Seems like the failures are about that too.

ondrejmirtes:
Thank you!

PR #1892, opened 2022-10-21
Replace ArrayIsListFunctionTypeSpecifyingExtension with stubs

rvanvelzen:
This also adds a test for calling `unset()` on a list, which is now easily verifiable.


ondrejmirtes:
Really nice, thank you :)

PR #1893, opened 2022-10-21
Test list return type covariance

rvanvelzen:
Per https://github.com/phpstan/phpstan/issues/8185


ondrejmirtes:
Nice, so it works correctly :) Thank you.

PR #1898, opened 2022-10-22
added is_callable() param-out

staabm:
I have never used this parameter, but stills.. thats the last function with a by-ref param from the php top-used 100 functions list

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -298,6 +298,6 @@ function headers_sent(string &$filename = null, int &$line = null): bool {}
 function mb_parse_str(string $string, &$result = []): bool {}
 
 /**
- * @param-out callable|false $callable_name
+ * @param-out callable-string|false $callable_name
  */
 function is_callable(mixed $value, bool $syntax_only = false, string &$callable_name = null): bool {}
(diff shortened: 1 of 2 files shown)


stubs/core.stub
```diff
@@ -296,3 +296,8 @@ function headers_sent(string &$filename = null, int &$line = null): bool {}
  * @param-out array<string, string|array<string>> $result
  */
 function mb_parse_str(string $string, &$result = []): bool {}
+
+/**
+ * @param-out callable|false $callable_name
```
ondrejmirtes: Should rather be callable-string. And have you verified what's set in case of non-callable being passed in?

ondrejmirtes (changes requested):

stubs/core.stub
staabm: yep, I did.

https://3v4l.org/2UpJB

ondrejmirtes: So it's never false?

What about if you pass in a nonexistent method via the array syntax?

staabm: Ahh right: it doesn't get false

https://3v4l.org/HctjS

ondrejmirtes:
Thank you.

PR #1900, opened 2022-10-23
refactor pow() return type extension to re-use BinaryOp\Pow

staabm:
to bring both implementations in sync, and therefore on feature parity.

next step, in a followup PR, will be to add IntegerRangeType support to `pow`

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -1122,6 +1122,25 @@ class InitializerExprTypeResolver
 		$leftType = $getTypeCallback($left);
 		$rightType = $getTypeCallback($right);
 
+		if ($leftType instanceof MixedType || $rightType instanceof MixedType) {
+			return new BenevolentUnionType([
+				new FloatType(),
+				new IntegerType(),
+			]);
+		}
+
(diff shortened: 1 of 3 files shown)


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1122,6 +1122,25 @@ public function getPowType(Expr $left, Expr $right, callable $getTypeCallback):
...
+			return new BenevolentUnionType([
+				new FloatType(),
+				new IntegerType(),
+			]);
+		}
+
+		$object = new ObjectWithoutClassType();
+		if (
+			!$leftType instanceof NeverType &&
+			!$rightType instanceof NeverType &&
```
staabm, line 1135: I added these 2 NeverType guard conditions, to not regress existing integer-range-math tests in https://github.com/phpstan/phpstan-src/blob/ff0d36e26667cac4a18f26576b0cd2fe96d7faf6/tests/PHPStan/Analyser/data/math.php#L157-L158

ondrejmirtes:
Thank you!

VincentLanglet:
I think it solve https://github.com/phpstan/phpstan/issues/2009 @ondrejmirtes @staabm, no ?

PR #1901, opened 2022-10-23
#8086 Add colon as prefix of line number in table report, when running in VSCode terminal

LeoVie:
This solves https://github.com/phpstan/phpstan/issues/8086

The diff of the test class is a bit bigger that required. I've added keys to the data provider arrays, so it's easier to understand, what the data values do, without having to scroll to the test method. I hope, it's okay, I did this in this PR, even if it's a bit out-of-scope.


LeoVie:
Seems like the failing checks are not really related to the changes in this PR.

ondrejmirtes:
Thank you! BTW If you want to follow up this PR, I think the test is going to fail when PHPStan test suite is running inside VSCode terminal 😂

PR #1902, opened 2022-10-24
Consider `get_object_var()` as private property read

vovochka404:
fixes issue https://github.com/phpstan/phpstan/issues/3654 (treat private properties as read if there is a get_object_vars invocation inside class)

### DIFF
--- a/src/Node/ClassStatementsGatherer.php
+++ b/src/Node/ClassStatementsGatherer.php
@@ -17,10 +17,17 @@ use PHPStan\Node\Property\PropertyRead;
 use PHPStan\Node\Property\PropertyWrite;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\ThisType;
 use function count;
+use function in_array;
 
 class ClassStatementsGatherer
 {
+
+	private const PROPERTY_ENUMERATING_FUNCTIONS = [
+		'get_object_vars',
+		'array_walk',
+	];
 
 	/** @var callable(Node $node, Scope $scope): void */
 	private $nodeCallback;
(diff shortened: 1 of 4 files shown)


src/Node/ClassStatementsGatherer.php
```diff
@@ -196,4 +211,31 @@ private function gatherNodes(Node $node, Scope $scope): void
...
+	{
+		$args = $node->getArgs();
+		if (count($args) === 0) {
+			return;
+		}
+
+		$firstArgValue = $args[0]->value;
+		if (
+			!$firstArgValue instanceof Expr\Variable
+			|| $firstArgValue->name !== 'this'
```
ondrejmirtes: Instead of this code, you should check that `$scope->getType($firstArgValue)` is instanceof ThisType.

src/Node/ClassStatementsGatherer.php
```diff
@@ -196,4 +211,31 @@ private function gatherNodes(Node $node, Scope $scope): void
...
+		) {
+			return;
+		}
+
+		$classProperties = $this->classReflection->getNativeReflection()->getProperties();
```
ondrejmirtes, line 229: These PHPStan failures need to be added to this file https://github.com/phpstan/phpstan-src/blob/1.9.x/build/enum-adapter-errors.neon

src/Node/ClassStatementsGatherer.php
```diff
@@ -18,10 +18,17 @@
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\ShouldNotHappenException;
 use function count;
+use function in_array;
 
 class ClassStatementsGatherer
 {
 
+	/** @var array<string> */
+	private static array $propEnumerationFuncCalls = [
```
ondrejmirtes: Could be a private class constant

ondrejmirtes (changes requested):

src/Node/ClassStatementsGatherer.php
vovochka404: Fixed

src/Node/ClassStatementsGatherer.php
vovochka404: Fixed

src/Node/ClassStatementsGatherer.php
vovochka404: Fixed

vovochka404:
Thanks for your review! I was not sure if `ClassStatementsGatherer` is a proper place to add this logic with this hardcoded constants. May be there should be an extension point?

And when i was experimenting with another approach to fix this issue, i found an interesting moment. If i'm running analysis on several files, where `get_object_var` was add through trait, then everything works only if this trait in this subset if analyzed. In other case, functions from traits are ignored (and in such case there is no informations that get_object_vars was called). I was using phpstan 1.1.2.

May be you have some ideas how such behavior can be workaraounded? :)

ondrejmirtes:
> then everything works only if this trait in this subset if analyzed

That's an invalid use-case, please see: https://phpstan.org/blog/why-you-should-always-analyse-whole-project

ondrejmirtes:
Thank you.

PR #1903, opened 2022-10-24
Fix reordering unspecified named arguments

rvanvelzen:
Fix phpstan/phpstan#8204


ondrejmirtes:
This was pretty broken :) Thank you very much.

PR #1904, opened 2022-10-24
support integer-ranges in pow()

staabm:
closes the last remaining todo of #5614

fixes https://github.com/phpstan/phpstan/issues/5614


staabm:
//cc @orklah

ondrejmirtes (changes requested):
This doesn't handle unions of integer ranges on both sides. Instead of `instanceof *Type`, it'd be better to handle this with a new method on Type. But of course that needs to be designed in order to be useful, ideally to replace other instances of `instancoef IntegerRangeType` and even maybe `TypeUtils::getConstantIntegers()`.

staabm:
> it'd be better to handle this with a new method on Type

if I get you right we are talking about methods such as the following, right?
```diff
diff --git a/src/Type/Type.php b/src/Type/Type.php
index f02971559..5a00d7f1c 100644
--- a/src/Type/Type.php
+++ b/src/Type/Type.php
@@ -112,6 +112,16 @@ interface Type

        public function shuffleArray(): Type;

+       public function subtract(Type $subtrahend): Type;
+
+       public function sum(Type $summand): Type;
+
+       public function multiply(Type $multiplicand): Type;
+
+       public function divide(Type $divisor): Type;
+
+       public function pow(Type $exponent): Type;
+
```

or do you have somthing different in mind?

ondrejmirtes:
Yeah, that would be great, although "pow" isn't a word unlike the others, should be "exponentiate" probably.

ondrejmirtes:
We'll be able to get rid of a lot of code in InitializerExprTypeResolver :)

staabm:
implemented `Type->exponentiate()` for now and will do the others in separate PRs

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1122,54 +1122,9 @@ public function getPowType(Expr $left, Expr $right, callable $getTypeCallback):
...
-				}
-			}
-
-			return TypeCombinator::union(...$resultTypes);
+		$exponentiatedTyped = $leftType->exponentiate($rightType);
+		if (!$exponentiatedTyped instanceof ErrorType) {
+			return $exponentiatedTyped;
 		}
 
 		return $this->resolveCommonMath(new BinaryOp\Pow($left, $right), $leftType, $rightType);
```
staabm: I kept this pass thru `resolveCommonMath` because otherwise we would loose support for operator-type specifiying extensions (even though atm, there is no test-coverage for using it with `pow`).

staabm: test added with https://github.com/phpstan/phpstan-src/pull/1933

ondrejmirtes: What resolveCommonMath does here? Feels like it shouldn't be necessary to call it at all.

src/Type/Accessory/AccessoryNumericStringType.php
```diff
@@ -216,4 +216,9 @@ public function tryRemove(Type $typeToRemove): ?Type
 		return null;
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new ErrorType();
```
ondrejmirtes: Wrong https://3v4l.org/AoROm

src/Type/Accessory/AccessoryNonFalsyStringType.php
```diff
@@ -204,4 +204,9 @@ public static function __set_state(array $properties): Type
 		return new self();
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new ErrorType();
```
ondrejmirtes: Wrong https://3v4l.org/AoROm

src/Type/Accessory/AccessoryNonEmptyStringType.php
```diff
@@ -213,4 +213,9 @@ public function tryRemove(Type $typeToRemove): ?Type
 		return null;
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new ErrorType();
```
ondrejmirtes: Wrong https://3v4l.org/AoROm

src/Type/Accessory/AccessoryLiteralStringType.php
```diff
@@ -204,4 +204,9 @@ public static function __set_state(array $properties): Type
 		return new self();
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new ErrorType();
```
ondrejmirtes: Wrong https://3v4l.org/AoROm

src/Type/Constant/ConstantIntegerType.php
```diff
@@ -88,6 +91,40 @@ public function generalize(GeneralizePrecision $precision): Type
 		return new IntegerType();
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		if ($exponent instanceof IntegerRangeType) {
```
ondrejmirtes: What about `ConstantIntegerType ** "unions of integer ranges"`? Where is it handled?

src/Type/MixedType.php
```diff
@@ -685,6 +685,14 @@ public function tryRemove(Type $typeToRemove): ?Type
 		return null;
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new BenevolentUnionType([
```
ondrejmirtes, line 806: What if all numeric types are subtracted?

src/Type/NullType.php
```diff
@@ -238,6 +238,11 @@ public function getGreaterOrEqualType(): Type
 		return new MixedType();
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new ErrorType();
```
ondrejmirtes: Wrong https://3v4l.org/Vvfcu

src/Type/ObjectType.php
```diff
@@ -1320,4 +1320,13 @@ public function tryRemove(Type $typeToRemove): ?Type
 		return null;
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		$object = new ObjectWithoutClassType();
+		if (!$exponent instanceof NeverType && !$object->isSuperTypeOf($this)->no() && !$object->isSuperTypeOf($exponent)->no()) {
```
ondrejmirtes, line 1411: Objects can be exponentiated? Huh?

src/Type/StaticType.php
```diff
@@ -562,6 +562,11 @@ public function tryRemove(Type $typeToRemove): ?Type
 		return null;
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new ErrorType();
```
ondrejmirtes: Should call `$this->getStaticObjectType()->exponentiate`

src/Type/StringType.php
```diff
@@ -180,6 +180,11 @@ public function tryRemove(Type $typeToRemove): ?Type
 		return null;
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		return new ErrorType();
```
ondrejmirtes: Wrong

ondrejmirtes (changes requested):

src/Reflection/InitializerExprTypeResolver.php
staabm: its still required, see https://github.com/phpstan/phpstan-src/pull/1904#discussion_r1007684752

we may drop or at least de-duplicate it, after we moved all the operations into the type-system

src/Type/ObjectType.php
staabm: yep, see https://github.com/phpstan/phpstan-src/blob/285b49f2072e18ffe9ff72f221151512b8a92fe6/tests/PHPStan/Analyser/data/pow.php#L17-L25

I moved this pre-exisiting logic from `InitializerExprTypeResolver`

src/Reflection/InitializerExprTypeResolver.php
ondrejmirtes: So I'd call a different method that only calls those extensions and nothing else.

src/Type/MixedType.php
staabm: pow() supports a lot of types as exponent, see e.g. https://3v4l.org/ZtZVc

we would need a big union, which might not be really useful?

per https://www.php.net/manual/en/function.pow.php it supports `mixed` args, while phpstan only accepts `int|float` per functionMap.

src/Type/FloatType.php
```diff
@@ -149,6 +150,33 @@ public function traverse(callable $cb): Type
...
+
+		if ($exponent instanceof NeverType) {
+			return new NeverType();
+		}
+		if (!$numberType->isSuperTypeOf($exponent)->yes()) {
+			return new ErrorType();
+		}
+
+		if ($exponent instanceof ConstantScalarType) {
+			if ($exponent->getValue() == 0) {
```
staabm: the cs-fixer will turn these into strict comparisons, which in turn breaks a lot of tests.
I think we should ignore these from the cs fixer

src/Reflection/InitializerExprTypeResolver.php
staabm: I tried and realized there is more shared logic. 

I think it will be simpler after we moved more logic into the type system

src/Type/Constant/ConstantIntegerType.php
```diff
@@ -88,6 +93,56 @@ public function generalize(GeneralizePrecision $precision): Type
...
+				$max = $this->getValue() ** $exponent->getMax();
+			}
+
+			if (!is_float($min) && !is_float($max)) {
+				return IntegerRangeType::fromInterval($min, $max);
+			}
+		} elseif ($exponent instanceof ConstantScalarType) {
+			$exponentValue = $exponent->getValue();
+			if (is_int($exponentValue)) {
+				$min = $this->getValue() ** $exponentValue;
```
staabm: Memo to me: double check... seems duplicated with the following if (constant-scalar) case

staabm:
thanks for the feedback.

after I realized php-src pow() actually supports a lot more types then `int|float`, I refactored this PR and added a lot of more tests.

I could not locale where the last remaining phpstan error is originating from. any hint would be great.
maybe it should just be put into the baseline, since using `null` as a base for exponentiation seems to be something we might want to error about, even if php-src pow() actually allows it?

staabm:
Got an idea

src/Reflection/InitializerExprTypeResolver.php
staabm: after the final refactoring, I was able to drop the `resolveCommonMath` path

ondrejmirtes:
There are some conflicts.

staabm:
resolved

src/Type/FloatType.php
```diff
@@ -201,6 +201,12 @@ public function traverse(callable $cb): Type
 		return $this;
 	}
 
+	public function exponentiate(Type $exponent): Type
+	{
+		$helper = new ExponentiateHelper();
+		return $helper->exponentiate($this, $exponent);
```
ondrejmirtes: It's weird to `new` this. Typically we inject these services through DI. But I understand that we're in a value object where DI is not accessible. In these cases I'd write this as a static method: `ExponentiateHelper::exponentiate`. Thanks.

ondrejmirtes (changes requested):
Otherwise 👍

src/Type/FloatType.php
staabm: good point, fixed.

ondrejmirtes:
Thank you.

PR #1907, opened 2022-10-24
Improve non-array specification for recently cleaned-up array functions

herndlm:
array_fill_keys: https://3v4l.org/8ik8V
array_flip: https://3v4l.org/jBASO
array_keys: https://3v4l.org/TYWrU
array_values: https://3v4l.org/89UDN

I was lazy with the pre PHP 8 checks and just wrapped them in conditionals like I did it with array_search. should the test files be split maybe to assert this for PHP < 8 as well? or not worth the hassle? hmm. we could maybe also move them all out to 2 test files only. e.g. non-array-array-functions.php and non-array-array-functions-php8.php or so

### DIFF
--- a/src/Type/Php/ArrayFillKeysFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayFillKeysFunctionReturnTypeExtension.php
@@ -4,13 +4,20 @@ namespace PHPStan\Type\Php;
 
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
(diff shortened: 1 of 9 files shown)


ondrejmirtes:
Awesome, thank you!

herndlm:
are you fine with the tests like that or should I add more / split them so that we have more executing on PHP < 8 as well?

ondrejmirtes:
The best would be (for each function):

* Have common file that's always tested on every PHP version for things that don't change
* Have a PHP 7 specific file
* Have a PHP 8 specific file

PR #1910, opened 2022-10-25
Update deprecated set-output to $GITHUB_OUTPUT

samsonasik:
Ref https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -60,7 +60,7 @@ jobs:
         id: rector-cache-key
         if: matrix.php-version != '8.1' && matrix.php-version != '8.2'
         run: |
-          echo "::set-output name=sha::$(php build/rector-cache-files-hash.php)"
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you :) Please do the same here: https://github.com/phpstan/phpstan/tree/1.9.x/.github/workflows

ondrejmirtes:
I'm sorry, I have to revert your changes, they break the workflow: https://github.com/phpstan/phpstan-src/actions/runs/3320826153/jobs/5487697232

samsonasik:
That's strange, seems working ok on rector-src https://github.com/rectorphp/rector-src/commit/a59fd4879215f2c831e863894fe3447144d5feaf , probably missing quote somewhere ?

rez1dent3:
@ondrejmirtes Previously, text escaping was required. Apparently it shot you off. In theory, in this case it is necessary to test separately.

https://github.com/actions/runner/issues/2225
Need to follow the issue^

PR #1912, opened 2022-10-25
Check nullable only on level 8

VincentLanglet:


ondrejmirtes:
I just merged the fix 😊

PR #1899, opened 2022-10-22
implement FunctionReflection/ExtendedMethodReflection returnsByReference()

staabm:
closes https://github.com/phpstan/phpstan/issues/6005

PR #1913, opened 2022-10-25
cover ExtendedMethodReflection by bc-promise

staabm:
as requested in https://github.com/phpstan/phpstan-src/pull/1899#discussion_r1002685161

requires https://github.com/ondrejmirtes/BackwardCompatibilityCheck/pull/23

### DIFF
--- a/src/Reflection/ExtendedMethodReflection.php
+++ b/src/Reflection/ExtendedMethodReflection.php
@@ -14,6 +14,8 @@ use PHPStan\Type\Type;
  * and its methods in their code.
  *
  * Methods on ExtendedMethodReflection are subject to change.
+ *
+ * @api
  */
 interface ExtendedMethodReflection extends MethodReflection
 {
--- a/src/Rules/Api/ApiInterfaceExtendsRule.php
+++ b/src/Rules/Api/ApiInterfaceExtendsRule.php
@@ -9,7 +9,6 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleError;
 use PHPStan\Rules\RuleErrorBuilder;
-use PHPStan\Type\Type;
 use function array_merge;
 use function count;
 use function in_array;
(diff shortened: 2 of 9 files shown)


src/Rules/Api/ApiInterfaceExtendsRule.php
```diff
@@ -66,10 +65,7 @@ private function checkName(Scope $scope, Node\Name $name): array
 			'https://github.com/phpstan/phpstan/discussions',
 		))->build();
 
-		if (in_array($extendedInterfaceReflection->getName(), [
-			Type::class,
-			ReflectionProvider::class,
-		], true)) {
+		if (in_array($extendedInterfaceReflection->getName(), BcUncoveredInterface::CLASSES, true)) {
```
staabm, line 68: I am re-using the same constant as in `ApiClassImplementsRule`.. thats how I understand the todo.

since we are now ignoring 5 classes here instead of only 2, I could add more tests.
but I wasn't sure whether I should DRY only `Type::class, ReflectionProvider::class` or all 5 class-strings..?

src/Reflection/ExtendedMethodReflection.php
```diff
@@ -14,6 +14,8 @@
  * and its methods in their code.
  *
  * Methods on ExtendedMethodReflection are subject to change.
+ *
+ * @api
```
staabm, line 18: above this interface is this big comment, mentioning that this interface is not meant for re-use. I guess it should be dropped now?

ondrejmirtes: I'll rewrite it after merge.

src/Rules/Api/BcUncoveredInterface.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace PHPStan\Rules\Api;
+
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ExtendedMethodReflection;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Type\Type;
+
+final class BcUncoveredInterface
```
ondrejmirtes, line 11: A food for thought - in the future we could add a new annotation like `@api-interface-forbid-implements`. That would allow us not to maintain these lists and also BackwardCompatibilityCheck with each new interface.

ondrejmirtes:
Thank you!

PR #1914, opened 2022-10-25
Fix closure return type based on by-ref use

rvanvelzen:
Fix phpstan/phpstan#6806


ondrejmirtes:
Thank you!

staabm:
Nice. This one unlocked a few issues. Will do regression tests for them (please stop me Richard if you prefer doing them yourself)

PR #1915, opened 2022-10-25
Fix closure return type for different same-code functions in same scope

rvanvelzen:
Fix phpstan/phpstan#8179


ondrejmirtes:
Thank you!

PR #1916, opened 2022-10-25
Add `Type::intersectKeyArray()`

herndlm:


src/Type/ArrayType.php
```diff
@@ -388,6 +388,15 @@ public function flipArray(): Type
 		return new self($this->getIterableValueType()->toArrayKey(), $this->getIterableKeyType());
 	}
 
+	public function intersectKeyArray(Type $arrays): Type
+	{
+		if ($arrays->getIterableKeyType()->isSuperTypeOf($this->getIterableKeyType())->no()) {
+			return ConstantArrayTypeBuilder::createEmpty()->getArray();
+		}
+
+		return new self($arrays->getIterableKeyType()->generalize(GeneralizePrecision::lessSpecific()), $this->getIterableValueType());
```
herndlm: The generalization here is not correct yet I think. Needs also more tests

tests/PHPStan/Analyser/data/array-intersect-key.php
```diff
@@ -0,0 +1,144 @@
+<?php
+
+namespace ArrayIntersectKey;
+
+use function array_intersect_key;
+use function PHPStan\Testing\assertType;
+
+class Foo
```
ondrejmirtes, line 8: Hi, file renames are hard to review, there isn't a good reason to do it :) So please keep the original file name. You can follow my advice from the other PR, and keep testing it on PHP 7 :)

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/array-intersect-key.php
herndlm: adapted. I just moved up the yield of the constant test file so that they are all together and don't feel lonely :)

ondrejmirtes:
Thank you!

PR #1917, opened 2022-10-25
Upgrade deprecated set-*

rez1dent3:
Second try #1910...

@ondrejmirtes Hello. I'm not sure that there will be a correct log (log escaping). Unfortunately, I can't check it in my fork. At least there is no more error: https://github.com/rez1dent3/phpstan-src/actions/runs/3322690041/jobs/5492131881

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -59,8 +59,7 @@ jobs:
       - name: "Rector downgrade cache key"
         id: rector-cache-key
         if: matrix.php-version != '8.1' && matrix.php-version != '8.2'
-        run: |
-          echo "::set-output name=sha::$(php build/rector-cache-files-hash.php)"
+        run: echo "sha=$(php build/rector-cache-files-hash.php)" >> $GITHUB_OUTPUT
 
       - name: "Rector downgrade cache"
         if: matrix.php-version != '8.1' && matrix.php-version != '8.2'
--- a/.github/workflows/phar.yml
+++ b/.github/workflows/phar.yml
@@ -53,8 +53,7 @@ jobs:
 
       - name: "Rector downgrade cache key"
         id: rector-cache-key
-        run: |
-          echo "::set-output name=sha::$(php build/rector-cache-files-hash.php)"
+        run: echo "sha=$(php build/rector-cache-files-hash.php)" >> $GITHUB_OUTPUT
 
       - name: "Rector downgrade cache"
         uses: actions/cache@v3
@@ -107,7 +106,7 @@ jobs:
 
       - name: "Save checksum"
         id: "checksum"
-        run: echo "::set-output name=md5::$(md5sum tmp/phpstan.phar | cut -d' ' -f1)"
+        run: echo "md5=$(md5sum tmp/phpstan.phar | cut -d' ' -f1)" >> $GITHUB_OUTPUT
 
       - uses: actions/upload-artifact@v3
         with:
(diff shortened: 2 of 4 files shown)


rez1dent3:
@ondrejmirtes ~While translated into draft. Tomorrow I will check another option, maybe it will work better.~ Done

https://github.com/rez1dent3/phpstan-src/actions/runs/3322894879/jobs/5492570187

rez1dent3:
This solved my problem, you shouldn't have any more problems either.

Example:
https://github.com/rez1dent3/ghacts/pull/1/files#diff-1eb4e5fd5611777d4e597ef299a1cb5ba8050c28a2dabbd4fbc56205d69e5ddaR26-R38

https://github.com/rez1dent3/ghacts/pull/1#issuecomment-1291092461

ondrejmirtes:
Thank you very much!

ondrejmirtes:
Hey @rez1dent3 it looks like you know your stuff when it comes to Bash. Could you please help me out a bit?

It's about this line: https://github.com/phpstan/phpstan-src/blob/34dabddb9001404f8c7667cf6a71629cbc8253bc/.github/workflows/phar.yml#L243

When the Git log contains a commit that contains message with `""`, like `Revert "abc"`, the command fails. How can I pass the Git log string correctly to `git commit -m ""`? Thanks.

The failure is:

```
  git config --global user.name "Ondrej Mirtes"
  git config --global user.email "ondrej@mirtes.cz"
  git add .
  git commit --gpg-sign -m "Updated PHPStan to commit e16c1df489c7002e9e4609ee4f3490ed70d13814" -m "https://github.com/phpstan/phpstan-src/commit/e16c1df489c7002e9e4609ee4f3490ed70d13814 Revert "Improve expression type resolving of superglobals"" --author "Ondrej Mirtes <ondrej@mirtes.cz>"
  lucky_commit e16c1df
  git push
  shell: /usr/bin/bash -e {0}
error: pathspec 'expression' did not match any file(s) known to git
error: pathspec 'type' did not match any file(s) known to git
error: pathspec 'resolving' did not match any file(s) known to git
error: pathspec 'of' did not match any file(s) known to git
error: pathspec 'superglobals' did not match any file(s) known to git
```

rez1dent3:
```bash
git commit --gpg-sign -m "Updated PHPStan to commit e16c1df489c7002e9e4609ee4f3490ed70d13814" -m "https://github.com/phpstan/phpstan-src/commit/e16c1df489c7002e9e4609ee4f3490ed70d13814 Revert \"Improve expression type resolving of superglobals\"" --author "Ondrej Mirtes <ondrej@mirtes.cz>"
```

Hello. The line escaping is missing. At least I think so.

rez1dent3:
I'll try to make a fix in my free time.

ondrejmirtes:
Surprisingly newlines work correctly if you look at some examples of commits here https://github.com/phpstan/phpstan/commit/7380ed007f26ffb1ac80bcaad288b431c96fe30e.

What doesn't work are double quotes. Bash is really mysterious to me here. If I had a variable then `git commit -m "$INPUT"` works flawlessly with any special characters in the variable. But I don't know how to write it when GitHub Actions simply replace the `${{ }}` thing with the text.

herndlm:
Isn't it that the unescaped double quotes inside the double quotes end the string / arg leading to the following things being interpreted as new args / commands. Or am I oversimplifying this?

ondrejmirtes:
It is but I don't know what to do about it.

herndlm:
Ah OK, the dynamic part with github actions replacament, got it. I also wonder if there is a better solution than trying to string replace it or use something like https://unix.stackexchange.com/questions/379181/escape-a-variable-for-use-as-content-of-another-script which would need the string to be assigned to a a variable I suppose. Looking forward to the solution :)

PR #1687, opened 2022-09-01
Deprecate TypeUtils::getArrays

herndlm:
Deprecates `TypeUtils::getArrays()` and `TypeUtils::getAnyArrays()` in favor of `Type::getArrays()`

I might be getting a bit greedy now, but this is the first step towards reducing usage of `ConstantArrayType::getAllArrays()`

PR #1920, opened 2022-10-26
Remove `Type::getConstantArrays()`

herndlm:
As preparation for the 1.9.0 release :) 
See https://github.com/phpstan/phpstan-src/pull/1687#issuecomment-1282840241 where I mentioned that I think this was not the correct solution

This also reverts the deprecation of `TypeUtils::getOldConstantArrays()` and `TypeUtils::getConstantArrays()`. The latter is unused in PHPStan and I don't like it, but I'm not sure if we can just deprecate it. Or if we do if we should point people to getOldConstantArrays() or something else..

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1905,7 +1905,7 @@ class NodeScopeResolver
 					}
 				};
 
-				$constantArrays = $arrayType->getConstantArrays();
+				$constantArrays = TypeUtils::getOldConstantArrays($arrayType);
 				if (count($constantArrays) > 0) {
 					$newArrayTypes = [];
 					$prepend = $functionReflection->getName() === 'array_unshift';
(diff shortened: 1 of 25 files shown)


ondrejmirtes:
What? I don't get it at all, this is a step in the other direction than what I'd like to be taking.

herndlm:
> What? I don't get it at all, this is a step in the other direction than what I'd like to be taking.

I thought we said to get rid of the method on Type again because it was too specific. And I'd continue getting rid of the the getConstantArrays() usage in general. See also your comment after my linked one. Or did I misunderstand it? 😊

ondrejmirtes:
I don't like to go the original method that has "old" in its name. I think it's fine if people start using `Type::getConstantArrays()`. And once we have a better replacements for that, we can deprecate it.

herndlm:
> I don't like to go the original method that has "old" in its name. I think it's fine if people start using `Type::getConstantArrays()`. And once we have a better replacements for that, we can deprecate it.

ah ok, makes sense to me as well. I also don't like the "old" method. I can just continue slowly getting rid of Type::get*Arrays() then and don't have to worry that people start using them and we'll eventually deprecate them at some point again :)

PR #1923, opened 2022-10-26
Split tests for recently cleaned-up array functions

herndlm:
As mentioned in https://github.com/phpstan/phpstan-src/pull/1907#issuecomment-1291649517


ondrejmirtes:
Nice, thank you!

PR #1924, opened 2022-10-26
Deprecate `TypeUtils::getDirectClassNames()`

herndlm:
Just a simple migration over to Type::, as requested in https://github.com/phpstan/phpstan-src/pull/1922#issuecomment-1291875264

Are you fine with that name? E.g. we also have `Type::getReferencedClasses()` which is somehow related, right?

I'd try to get rid of more `instanceof TypeWithClassName` in a follow-up if that's fine


herndlm:
Looks likes a deprecated method rule in the deprecation rules extension is causing a deprecation now that needs fixing then 🙃

herndlm:
The new failing tests here are from the deprecation extension which needs fixing afterwards. But it would be ready

src/Type/UnionTypeHelper.php
```diff
@@ -32,6 +34,24 @@ public static function getReferencedClasses(array $types): array
...
+	/**
+	 * @param Type[] $types
+	 * @return list<string>
+	 */
+	public static function getObjectClassNames(array $types): array
+	{
+		return array_values(
+			array_unique(
+				array_merge(
+					...array_map(
```
ondrejmirtes: I really don't like this notation, a good old foreach would be much better. I have no idea what's the result of array_merge with `...array_map` inside :)

As a bonus, with foreach, I can easily put some debugging statements between the calls.

ondrejmirtes (changes requested):
Yeah, the name is OK :) Additionally, we could get rid of `instanceof TypeWithClassName` in this PR. Thank you.

herndlm:
getting rid of all `instanceof TypeWithClassName` here can get ugly really fast. currently there are 40 of those and with many I'm not sure if we should really use the new method, introduce something else, or just leave it as it is.

e.g. some examples

calls where we explicitly want to know the class for whatever reasons
```php
$classType = TypeTraverser::map($classType, static function (Type $type, callable $traverse) use (&$uncertainty): Type {
		return $traverse($type);
	}
	if ($type instanceof TypeWithClassName) {
		$uncertainty = true;
		return $type;
	}
...
```
```php
if (!$returnType instanceof TypeWithClassName) {
	return new MixedType();
}
```

calls where we want to know explicitly the **one** className
```php
$thisType instanceof TypeWithClassName ? $thisType->getClassName() : null,
```
```php
$isThrowable = $originalCatchType instanceof TypeWithClassName && strtolower($originalCatchType->getClassName()) === 'throwable';
```
```php
 elseif (
	$argValueType instanceof GenericClassStringType
	&& $argValueType->getGenericType() instanceof TypeWithClassName
) {
	$scopeClass = $argValueType->getGenericType()->getClassName();
	$thisType = $argValueType->getGenericType();
}
```

I'm not entirely sure how to get rid of them yet, e.g. in some cases I'd need to check that `getObjectClassNames()` is not empty, in some I'd need to check that one entry is returned, in some I can maybe traverse them and adapt the surrounding logic (but many places seem to be using `GenericTypeVariableResolver::getType()` which expects a `TypeWithClassName`) and in some maybe I want to check that only one entry is returned and use that. any ideas? or should we do this in more PRs maybe? I want to avoiding messing this one up too much :)

ondrejmirtes:
It shouldn't be hard to come up with examples where those `instanceof TypeWithClassName` are wrong - for example having an intersection with another type (but the condition where we're asking we'd still want it to be true).

Also - UnionType::getObjectClassNames() doesn't match how it should behave. If we for example have Foo|null, the returned array should be empty. Look at getArrays() for an example how to do it.

ondrejmirtes:
You can now just call `pickTypes` in UnionType :) https://github.com/phpstan/phpstan-src/commit/480626ecb52d2e98cc28cee8a18dfb86112b7f8f

herndlm:
> You can now just call `pickTypes` in UnionType :) [480626e](https://github.com/phpstan/phpstan-src/commit/480626ecb52d2e98cc28cee8a18dfb86112b7f8f)

that's great! I just realised that you type-hinted `T of Type` but here T would be `string` :/
can I just remove it or should I make it a `T of Type|string` or do you have some other idea?

ondrejmirtes:
@herndlm Nothing might break if we remove the `of Type` part?

herndlm:
> @herndlm Nothing might break if we remove the `of Type` part?

no, but it starts to get weird naming-wise. `pickTypes()`, `$getTypes`, `$innerTypes` and `$types` are all kind of wrong then

ondrejmirtes:
Yeah, you can rename that to something else :) Also maybe `pickFromTypes` instead of `pickTypes`.

src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
```diff
@@ -25,16 +25,22 @@ public function determineType(
...
 		return TypeTraverser::map(
 			$classType,
-			static function (Type $type, callable $traverse) use ($objectOrClassTypeClassName, $allowString, $allowSameClass): Type {
+			static function (Type $type, callable $traverse) use ($objectOrClassTypeClassNames, $allowString, $allowSameClass): Type {
 				if ($type instanceof UnionType || $type instanceof IntersectionType) {
 					return $traverse($type);
 				}
 				if ($type instanceof ConstantStringType) {
-					if (!$allowSameClass && $type->getValue() === $objectOrClassTypeClassName) {
+					if (!$allowSameClass && $objectOrClassTypeClassNames === [$type->getValue()]) {
```
staabm, line 44: `array_unique` preserves keys. I think therefore this comparison might not work, even when equal named classes are compared because keys might be different?

https://3v4l.org/raB2N

herndlm: true, thx. actually this is one of the weird cases where it's only relevant if it's one value. I don't quite like my current solution and might come back to this again..

ondrejmirtes:
Nice, I'd merge this if it wasn't a draft 👍

herndlm:
Soon 😊

src/Type/BenevolentUnionType.php
```diff
@@ -43,11 +43,11 @@ protected function unionTypes(callable $getType): Type
 		return TypeUtils::toBenevolentUnion(TypeCombinator::union(...$resultTypes));
 	}
 
-	protected function pickTypes(callable $getTypes): array
+	protected function pickFromTypes(callable $getValues): array
 	{
 		$types = [];
 		foreach ($this->getTypes() as $type) {
-			$innerTypes = $getTypes($type);
+			$innerTypes = $getValues($type);
```
herndlm: Rename those vars too

src/Type/StringAlwaysAcceptingObjectWithToStringType.php
```diff
@@ -10,19 +10,25 @@ class StringAlwaysAcceptingObjectWithToStringType extends StringType
...
+		$thatClassNames = $type->getObjectClassNames();
+		if ($thatClassNames === []) {
+			return parent::accepts($type, $strictTypes);
 		}
 
-		return parent::accepts($type, $strictTypes);
+		return TrinaryLogic::createNo()->lazyOr(
+			$thatClassNames,
+			static function (string $thatClassName) {
+				$reflectionProvider = ReflectionProviderStaticAccessor::getInstance();
```
herndlm, line 21: Move this out of callable

src/Type/StringType.php
```diff
@@ -98,19 +98,25 @@ public function accepts(Type $type, bool $strictTypes): TrinaryLogic
...
+		$thatClassNames = $type->getObjectClassNames();
+		if ($thatClassNames === [] || $strictTypes) {
+			return TrinaryLogic::createNo();
 		}
 
-		return TrinaryLogic::createNo();
+		return TrinaryLogic::createNo()->lazyOr(
+			$thatClassNames,
+			static function (string $thatClassName) {
+				$reflectionProvider = ReflectionProviderStaticAccessor::getInstance();
```
herndlm: Move this out of callable

src/Type/UnionType.php
```diff
@@ -895,22 +902,34 @@ protected function unionTypes(callable $getType): Type
 	 * @template T of Type
 	 * @param callable(Type $type): list<T> $getTypes
 	 * @return list<T>
+	 *
+	 * @deprecated Use pickFromTypes() instead.
```
herndlm, line 916: weird, BC check was complaining when I removed the whole protected method initially. wonder why. UPDATE: I guess because the class can still be extended, makes sense then..

src/Type/StringType.php
```diff
@@ -98,19 +98,25 @@ public function accepts(Type $type, bool $strictTypes): TrinaryLogic
...
-			return TrinaryLogic::createFromBoolean(
-				$typeClass->hasNativeMethod('__toString'),
-			);
+		$thatClassNames = $type->getObjectClassNames();
+		if ($thatClassNames === [] || $strictTypes) {
+			return TrinaryLogic::createNo();
 		}
 
-		return TrinaryLogic::createNo();
+		return TrinaryLogic::createNo()->lazyOr(
```
staabm: sorry if my question is dump,.. I don't foresee the overall change of this PR .. 

I am wondering, whether this change here means, that the new logic accepts a union of types with a `__toString` method, which it did not accept before...?
(in other words, does this change add a feature which needs a new test?)

src/Type/ClosureType.php
```diff
@@ -97,6 +97,11 @@ public function getReferencedClasses(): array
 		return array_merge($classes, $this->returnType->getReferencedClasses());
 	}
 
+	public function getObjectClassNames(): array
```
staabm, line 100: another maybe dumb question:

~~before this PR the `ClosureType` was not considered a `TypeWithClassName`, additionally it was not handled in `TypeUtil::getDirectClassNames()` etc.~~

but we are now using `Type->getObjectClassNames()` in a few places where we do not handle/did not handle the `ClosureType` type before, but it is now returned.

does this mean we have new features/new edge cases in this area which require new tests?

**scratch this comment.. I did a wrong code-search**

src/Type/StringType.php
herndlm: No, you're right. Things like that were the main reason I wanted to not change too much here and try to get rid of instanceof in follow-ups.
I basically tried to use the new method everywhere instead and this is how I came up with such changes. I'm not sure of all yet, and yes, this one for example should be tested somehow I guess.
I'm still OK with only merging the first commit here. But I can also start looking into adding more tests.

herndlm: @ondrejmirtes what do you think - I just took a second look at this. some of my adaptions where I tried to handle cases where `$type->getObjectClassNames()` would return more than one entry, are adapted in the same fashion using TrinaryLogin::lazy*(). BUT, AFAIK so far only `UnionType` and `IntersectionType` would return more than one. And e.g. here further up above `CompoundType` is already handled, so this code is basically useless.

I could simplify it again to e.g.
- throw a `ShouldNotHappenException` in case it's more than one and handle it if it's one or zero entries
- always only handle the first entry which would be set or null

not sure if my adaptions like this one make sense. I can't test them right now for sure :/

herndlm: nevermind, cleaning it up already, I'll mark it as ready to review when I'm done. no need to check now

src/Type/ObjectType.php
```diff
@@ -248,16 +248,22 @@ public function accepts(Type $type, bool $strictTypes): TrinaryLogic
 			return TrinaryLogic::createMaybe();
 		}
 
-		if (!$type instanceof TypeWithClassName) {
+		$thatClassNames = $type->getObjectClassNames();
+		if (count($thatClassNames) > 1) {
+			throw new ShouldNotHappenException();
```
ondrejmirtes, line 253: Why it couldn't be more than one? In the world where the only CompoundType is NeverType, this should be supported and implemented. By calling `checkSubclassAcceptability` for each class name and combining the results correctly.

src/Type/ObjectType.php
```diff
@@ -309,20 +315,22 @@ public function isSuperTypeOf(Type $type): TrinaryLogic
 		}
 
 		$thisClassName = $this->className;
-		$thatClassName = $type->getClassName();
+		if (count($thatClassNames) > 1) {
+			throw new ShouldNotHappenException();
```
ondrejmirtes, line 319: dtto

src/Type/StringType.php
```diff
@@ -98,19 +100,24 @@ public function accepts(Type $type, bool $strictTypes): TrinaryLogic
...
 		}
 
-		if ($type instanceof TypeWithClassName && !$strictTypes) {
-			$reflectionProvider = ReflectionProviderStaticAccessor::getInstance();
-			if (!$reflectionProvider->hasClass($type->getClassName())) {
-				return TrinaryLogic::createNo();
-			}
+		$thatClassNames = $type->getObjectClassNames();
+		if (count($thatClassNames) > 1) {
+			throw new ShouldNotHappenException();
```
ondrejmirtes, line 105: dtto

src/Analyser/MutatingScope.php
```diff
@@ -2730,6 +2730,11 @@ public function enterClosureCall(Type $thisType, Type $nativeThisType): self
 		$nativeExpressionTypes = $this->nativeExpressionTypes;
 		$nativeExpressionTypes['$this'] = ExpressionTypeHolder::createYes(new Variable('this'), $nativeThisType);
 
+		$thisObjectClassNames = $thisType->getObjectClassNames();
+		if (count($thisObjectClassNames) > 1) {
+			throw new ShouldNotHappenException();
```
ondrejmirtes: Can be more than one. You can typically have intersection type of ThisType and another type, for example:

```php
if ($this instanceof FooInterface) {
    // voila
}
```

You can change MutatingScope internals to correctly represent this.

ondrejmirtes: Should be tested of course :)

src/PhpDoc/TypeNodeResolver.php
```diff
@@ -629,9 +628,14 @@ private function resolveGenericTypeNode(GenericTypeNode $typeNode, NameScope $na
 			return new ErrorType();
 		} elseif ($mainTypeName === 'value-of') {
 			if (count($genericTypes) === 1) { // value-of<ValueType>
-				if ($genericTypes[0] instanceof TypeWithClassName) {
-					if ($this->getReflectionProvider()->hasClass($genericTypes[0]->getClassName())) {
-						$classReflection = $this->getReflectionProvider()->getClass($genericTypes[0]->getClassName());
+				$genericTypeObjectClassNames = $genericTypes[0]->getObjectClassNames();
+				if (count($genericTypeObjectClassNames) > 1) {
+					throw new ShouldNotHappenException();
```
ondrejmirtes, line 633: I think this can happen by writing something crazy in `value-of`.

ondrejmirtes (changes requested):
Awesome, I'm really looking forward to this :)

src/Type/ObjectType.php
herndlm: I even had something like that already but removed it again 😅 the problem for me was that I couldn't come up with a test (because as you hinted already, CompoundType is already handled before) and I was afraid that this "dead" code might break at some point in the future in weird ways. That's why I thought it might make sense to just not support it for now and break in more expected ways

ondrejmirtes: That's fine - write something that works, try it with a unit test and maybe temporarily make UnionType not implement CompoundType to see if your logic works...

When I make most types no longer implement CompoundType, I don't want to hunt for ShouldNotHappenExceptions 😊

src/Analyser/MutatingScope.php
herndlm: I pushed a fixup trying to adapt it. I'm not fully sure about it though. I added a test that at least leads now to passing 2 classNames there, which would have thrown an exception before, but the assertion was the same before. In theory something should be smarter now around `canAccessClassMember()` but I couldn't figure out how to test this yet :/

staabm:
Would be great if we could land this one, so we can continue with the refactoring without conflicts

ondrejmirtes:
Yeah, sure, could be out of draft already 😊 This thing is slowing down my mail client, with 50 messages in the thread 😂

herndlm:
Bump 😅

I'll try to undraft it soon™

src/Type/ObjectType.php
herndlm: this is a good example, I tried to
> By calling checkSubclassAcceptability for each class name and combining the results correctly. 

added test cases and commented out the CompoundType handling. I either had
```php
return TrinaryLogic::createNo()->lazyOr(
	$thatClassNames,
	fn (string $thatClassName) => $this->checkSubclassAcceptability($thatClassName),
);
```

or 

```php
return TrinaryLogic::createYes()->lazyAnd(
	$thatClassNames,
	fn (string $thatClassName) => $this->checkSubclassAcceptability($thatClassName),
);
```

but both are wrong. I don't know anymore if it was a union or an intersection, so I have no clue how to deal with them :/

my added test cases for `accepts()` were
```php
			[
				new ObjectType(DateTimeInterface::class),
				new UnionType([
					new ObjectType(stdClass::class),
					new ObjectType(stdClass::class),
				]),
				TrinaryLogic::createNo(),
			],
			[
				new ObjectType(DateTimeInterface::class),
				new UnionType([
					new ObjectType(stdClass::class),
					new ObjectType(DateTime::class),
				]),
				TrinaryLogic::createMaybe(), // was no with lazyAnd(), was yes with lazyOr()
			],
			[
				new ObjectType(DateTimeInterface::class),
				new UnionType([
					new ObjectType(DateTime::class),
					new ObjectType(DateTime::class),
				]),
				TrinaryLogic::createYes(),
			],
			[
				new ObjectType(DateTimeInterface::class),
				new IntersectionType([
					new ObjectType(stdClass::class),
					new ObjectType(stdClass::class),
				]),
				TrinaryLogic::createNo(),
			],
			[
				new ObjectType(DateTimeInterface::class),
				new IntersectionType([
					new ObjectType(stdClass::class),
					new ObjectType(DateTime::class),
				]),
				TrinaryLogic::createYes(), // was no with lazyAnd()
			],
			[
				new ObjectType(DateTimeInterface::class),
				new IntersectionType([
					new ObjectType(DateTime::class),
					new ObjectType(DateTime::class),
				]),
				TrinaryLogic::createYes(),
			],
```

these are my main problems here.. any ideas?

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

staabm:
Great job @herndlm 🙏

ondrejmirtes:
An idea - `Type::getStringClassNames()` for these cases:

* ConstantStringType that's also a class string
* GenericClassStringType - a shortcut to `$type->getGenericType()->getObjectClassNames()`

There are some cases of `instanceof GenericClassStringType` that we'd be able to get rid of.

PR #1926, opened 2022-10-27
improve hasExpressionType api

rajyan:


rajyan:
Hmm, my guess was wrong... This change doesn't fix the composer error.

rajyan:
Reverted because doesn't fix the composer integration test error, but I think this PR is mergable now.

ondrejmirtes:
Thank you!

PR #1927, opened 2022-10-27
Improve unset on ConstantArray

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8225

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -672,10 +672,11 @@ class ConstantArrayType extends ArrayType implements ConstantType
 		$optionalKeys = $this->optionalKeys;
 		$isList = $this->isList;
 		foreach ($this->keyTypes as $i => $keyType) {
-			if ($offsetType->isSuperTypeOf($keyType)->yes()) {
-				$optionalKeys[] = $i;
-				$isList = false;
+			if (!$offsetType->isSuperTypeOf($keyType)->yes()) {
+				continue;
 			}
+			$optionalKeys[] = $i;
+			$isList = false;
 		}
 		$optionalKeys = array_values(array_unique($optionalKeys));
 
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -669,7 +669,7 @@ class ConstantArrayType extends ArrayType implements ConstantType
 			return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, false);
 		}
 
-		return new ArrayType($this->getKeyType(), $this->getItemType());
+		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, array_keys($this->keyTypes), false);
 	}
 
 	public function fillKeysArray(Type $valueType): Type
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -669,7 +669,17 @@ class ConstantArrayType extends ArrayType implements ConstantType
 			return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, false);
 		}
 
-		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, array_keys($this->keyTypes), false);
+		$optionalKeys = $this->optionalKeys;
+		$isList = $this->isList;
+		foreach ($this->keyTypes as $i => $keyType) {
+			if ($offsetType->isSuperTypeOf($keyType)->yes()) {
+				$optionalKeys[] = $i;
+				$isList = false;
+			}
+		}
+		$optionalKeys = array_values(array_unique($optionalKeys));
+
+		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, $isList);
 	}
 
 	public function fillKeysArray(Type $valueType): Type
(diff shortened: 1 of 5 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -669,7 +669,7 @@ public function unsetOffset(Type $offsetType): Type
 			return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, false);
 		}
 
-		return new ArrayType($this->getKeyType(), $this->getItemType());
+		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, array_keys($this->keyTypes), false);
```
rvanvelzen: This also unsets keys of types that `$offsetType` does not match. Something like this should make it even more useful:
```suggestion
		$optionalKeys = $this->optionalKeys;
		$isList = $this->isList;
		foreach ($this->keyTypes as $i => $keyType) {
			if ($offsetType->isSuperTypeOf($keyType)->yes()) {
				$optionalKeys[] = $i;
				$isList = false;
			}
		}

		$optionalKeys = array_values(array_unique($optionalKeys));
		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, $isList);
```

herndlm: great idea! (this PR too) @VincentLanglet make sure to also run/update the tests locally to avoid many force-pushes 😊

VincentLanglet: Just discovered `make build`. I'll try to think about it next time.

VincentLanglet: For the record I get the error
```
1) PHPStan\Command\ErrorFormatter\TableErrorFormatterTest::testEditorUrlWithRelativePath
Failed asserting that ' ------ ----------------\n
  [32mLine[39m   [32mFoo.php[39m\n
 ------ ----------------\n
  12     Test\n
         ✏️  rel/Foo.php\n
 ------ ----------------\n
\n
\n
[37;41m                                                                                [39;49m\n
[37;41m [ERROR] Found 1 error                                                          [39;49m\n
[37;41m                                                                                [39;49m\n
\n
' contains "editor://custom/path/rel/Foo.php".
```
on local.

rajyan:
This can be the edge case of array keys.
Numeric string keys are converted to integers.
https://3v4l.org/AitVn

VincentLanglet:
> This can be the edge case of array keys. Numeric string keys are converted to integers. https://3v4l.org/AitVn

I added a test with your edge case, and it works fine.

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -3100,13 +3100,21 @@ public function dataBinaryOperations(): array
 				'$coalesceArray',
 			],
 			[
-				'array<0|1|2, 1|2|3>',
+				'array{1, 2, 3}',
```
rajyan, line 3103: Sorry for lack of explanation. I thought that maybe this test should be kept, because of https://3v4l.org/AitVn

VincentLanglet: I don't know if that's the good direction to go.

This PR is consistent with the current behavior, see https://phpstan.org/r/cd2147fb-6b73-4bdf-abd6-d1a05b0b8334
If we're following what you're asking here, we should report that
```
\PHPStan\dumpType($array[$foo]);
```
is `int|bool|string` then.

IMHO,
When you provide a constant string, phpstan understand `unset($array['1'])` so it's ok.
When you provide a non-constant string, phpstan should report `unset($array[$string])` as useless instead.

It would be consistent with the fact that `$array[$string]` is an error when the array is `array<int, mixed>`.

rajyan: Thank you for the example! Agreed with your explanation.

VincentLanglet: In the original PR I was unsetting everything (for int and string).
This is the current behavior of psalm, see
- https://psalm.dev/r/1296eb7185
- https://psalm.dev/r/be28362a63

@rvanvelzen made this proposal https://github.com/phpstan/phpstan-src/pull/1927#discussion_r1006664191 to only set the key of the same type ; which was not a bad idea.

We can still decide that
- Unsetting a `int` key will only unset the int keys
- Unsetting a `string` key will unset all the possibly keys

That's not a big change.

WDYT @ondrejmirtes ?

ondrejmirtes:
I'm fine with this, let's try it out! Thank you.

src/Type/Constant/ConstantArrayType.php
herndlm: that looks like some terminal ANSI weirdness or so? how are you running the tests exactly? any special tools/terminal involved. I definitely don't get that error 🤔

rajyan: I always have the same error too running `make tests`

rvanvelzen: I get the same kind of error when running `make` in the PhpStorm built-in terminal. `symfony/console` [disables hyperlinks](https://github.com/symfony/symfony/blob/v5.4.13/src/Symfony/Component/Console/Formatter/OutputFormatterStyle.php#L98-L99) because it appears to not handle them correctly.

I just run `make` in a regular terminal instead, or I ignore the error.

VincentLanglet: I run it inside phpstorm too

PR #1928, opened 2022-10-27
Regression test

rvanvelzen:
Trying to figure out what's going on with https://github.com/phpstan/phpstan-src/commit/921fd9a09c455a39d185399915b19dd3ec8b3b5f#commitcomment-87929523 - it looks like `unescapeStrings` is not set properly


rvanvelzen:
The phpstan-doctrine failure seems unrelated - fails locally on 1.8.11 as well

ondrejmirtes:
Thank you!

PR #1930, opened 2022-10-27
Add config to check BenevolentUnion

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8223

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -76,6 +76,7 @@ parameters:
 	checkNullables: false
 	checkThisOnly: true
 	checkUnionTypes: false
+	checkBenevolentUnionTypes: false
 	checkListType: false
 	checkExplicitMixedMissingReturn: false
 	checkPhpDocMissingReturn: false
@@ -290,6 +291,7 @@ parametersSchema:
 	checkNullables: bool()
 	checkThisOnly: bool()
 	checkUnionTypes: bool()
+	checkBenevolentUnionTypes: bool()
 	checkListType: bool()
 	checkExplicitMixedMissingReturn: bool()
 	checkPhpDocMissingReturn: bool()
@@ -1067,6 +1069,7 @@ services:
 			checkExplicitMixed: %checkExplicitMixed%
 			checkImplicitMixed: %checkImplicitMixed%
 			checkListType: %checkListType%
+			checkBenevolentUnionTypes: %checkBenevolentUnionTypes%
 
 	-
 		class: PHPStan\Rules\UnusedFunctionParametersCheck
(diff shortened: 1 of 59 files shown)


conf/config.neon
```diff
@@ -76,6 +76,7 @@ parameters:
 	checkNullables: false
 	checkThisOnly: true
 	checkUnionTypes: false
+	checkBenevolentUnionTypes: false
```
staabm, line 79: should it be true on bleedingEdge?

VincentLanglet: CheckUnionTypes is not true on bleedingEdge, so checkBenevolentUnionTypes which is stricter shouldn't be true too.

ondrejmirtes: `checkBenevolentUnionTypes` will never be true by default, that's why it's "benevolent".

src/Rules/RuleLevelHelper.php
```diff
@@ -48,6 +49,16 @@ public function isThis(Expr $expression): bool
 	/** @api */
 	public function accepts(Type $acceptingType, Type $acceptedType, bool $strictTypes): bool
 	{
+		$checkForUnion = $this->checkUnionTypes;
+
+		if (
+			$this->checkBenevolentUnionTypes
+			&& $acceptedType instanceof BenevolentUnionType
```
ondrejmirtes: This is too naive - will not see nested BenevolentUnionType. Needs similar logic as `checkExplicitMixed` and use `TypeTraverser`.

ondrejmirtes (changes requested):

src/Rules/RuleLevelHelper.php
VincentLanglet: I have a test for (\DateTimeImmutable|false) and for array<(\DateTimeImmutable|false)>.

What do you have in mind with "nested BenevolentUnionType", I'll add new tests.

staabm: I guess things like `(float|int)|string`

conf/config.neon
VincentLanglet: But I'm kinda worried about someone using the config
```
checkUnionTypes: false
checkBenevolentUnionTypes: true
```
what should be the behavior ?
- checking both ?
- checking none ?
- only checking the benevolent ?
- throwing an error ?

ondrejmirtes:
I think it's fine as it is. Please also send a documentation update to https://phpstan.org/config-reference (https://github.com/phpstan/phpstan/edit/1.9.x/website/src/config-reference.md).

Thank you.

VincentLanglet:
> I think it's fine as it is. Please also send a documentation update to https://phpstan.org/config-reference (https://github.com/phpstan/phpstan/edit/1.9.x/website/src/config-reference.md).
> 
> Thank you.

Sure, I started https://github.com/phpstan/phpstan/pull/8234

PR #1931, opened 2022-10-28
Fix composer integration

rajyan:
solve composer integration test errors caused by
https://github.com/phpstan/phpstan-src/pull/1919

Finally found the cause by deep debugging!

When promoting native types, variable with certainty "maybe" was promoted which led to wrong specification in `$treatPhpDocTypesAsCertain: false`.
We can promote "maybe" native types again after
https://github.com/phpstan/phpstan-src/pull/1919#issuecomment-1292596647
> Change $nativeExpressionTypes into array<string, ExpressionTypeHolder>

and native type is handled correctly in all places.

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2193,7 +2193,7 @@ class MutatingScope implements Scope
 		$expressionTypes = $this->expressionTypes;
 		foreach ($this->nativeExpressionTypes as $exprString => $type) {
 			$has = $this->hasVariableType(substr($exprString, 1));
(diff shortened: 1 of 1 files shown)


rajyan:
sorry, it wasn't true
I was looking at the wrong result locally

rajyan:
but I can say 
> Change $nativeExpressionTypes into array<string, ExpressionTypeHolder>

should solve the problem.

There is a 

`$this->getOption('latest') (maybe) => true`
leading
`native $this->getOption('latest') => true`
which is wrong.

PR #1932, opened 2022-10-28
temporal solution to handle "maybe" certainty types assign in nativeE…

rajyan:
…xpressionTypes

hope this fixes the problem
see https://github.com/phpstan/phpstan-src/pull/1931


rajyan:
I know this isn't a "correct" solution for native types, but I think it should be a working solution while the refactoring is running.

ondrejmirtes:
Thank you 😊

PR #1933, opened 2022-10-28
added operator-specifiying test for the pow-operator

staabm:
coverage was missing for this single operators, as mentioned in https://github.com/phpstan/phpstan-src/pull/1904#discussion_r1007684752

### DIFF
--- a/tests/PHPStan/Type/TestDecimalOperatorTypeSpecifyingExtension.php
+++ b/tests/PHPStan/Type/TestDecimalOperatorTypeSpecifyingExtension.php
@@ -10,7 +10,7 @@ final class TestDecimalOperatorTypeSpecifyingExtension implements OperatorTypeSp
 
 	public function isOperatorSupported(string $operatorSigil, Type $leftSide, Type $rightSide): bool
 	{
-		return in_array($operatorSigil, ['-', '+', '*', '/'], true)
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Type/TestDecimalOperatorTypeSpecifyingExtension.php
```diff
@@ -10,7 +10,7 @@ final class TestDecimalOperatorTypeSpecifyingExtension implements OperatorTypeSp
 
 	public function isOperatorSupported(string $operatorSigil, Type $leftSide, Type $rightSide): bool
 	{
-		return in_array($operatorSigil, ['-', '+', '*', '/'], true)
+		return in_array($operatorSigil, ['-', '+', '*', '/', '^'], true)
```
ondrejmirtes: `^` is XOR, POW is `**`.

ondrejmirtes (changes requested):

tests/PHPStan/Type/TestDecimalOperatorTypeSpecifyingExtension.php
staabm: oh my, nice catch. it means, we have 2 uncovered cases :).

added POW for real now

ondrejmirtes:
Thank you.

PR #1934, opened 2022-10-28
Mutating scope refactor and regression test

rajyan:
Added some regression test for the logic of keeping var related expression types.
Fixed the last test case to work correctly.

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2880,6 +2880,7 @@ class MutatingScope implements Scope
 			$nativeTypes[$paramExprString] = $parameterType;
 		}
 
+		$nonRefVariableNames = [];
 		foreach ($closure->uses as $use) {
 			if (!is_string($use->var->name)) {
 				throw new ShouldNotHappenException();
@@ -2891,6 +2892,7 @@ class MutatingScope implements Scope
 				$nativeTypes[$paramExprString] = new MixedType();
 				continue;
 			}
+			$nonRefVariableNames[$variableName] = true;
 			if ($this->hasVariableType($variableName)->no()) {
 				$variableType = new ErrorType();
 			} else {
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/data/specified-types-closure-use.php
```diff
@@ -34,4 +34,39 @@ function () use ($call, &$a): void {
...
+	}
+
+	public function doBuzz(array $arr, string $key): void
+	{
+		if (isset($arr[$key])) {
+			assertType('array', $arr);
+			assertType("mixed~null", $arr[$key]);
+			function ($key) use ($arr): void {
+				assertType('array', $arr);
+				assertType("mixed", $arr[$key]);
```
rajyan, line 67: `$arr[$key]` was mixed~null before this PR

src/Analyser/MutatingScope.php
```diff
@@ -2891,34 +2892,41 @@ private function enterAnonymousFunctionWithoutReflection(
...
+			$expr = $this->exprStringToExpr((string) $exprString);
+			if ($expr === null) {
+				continue;
+			}
+			if ($expr instanceof Variable) {
+				continue;
+			}
+			$variables = (new NodeFinder())->findInstanceOf([$expr], Variable::class);
+			if ($variables === []) {
+				continue;
```
rajyan, line 2915: I think https://github.com/phpstan/phpstan-src/pull/1929 can be merged by adding `$experssionTypes[$exprString] = $typeHolder` before this line.

rajyan:
@rvanvelzen 
Cherry picked your tests from https://github.com/phpstan/phpstan-src/pull/1929 because the refactoring now includes your fix too.

ondrejmirtes:
I tried to break the logic in https://github.com/phpstan/phpstan-src/pull/1929 and https://github.com/phpstan/phpstan-src/pull/1934 and I did it. This example breaks in these PRs: https://phpstan.org/r/704ec47a-e5b0-4d5b-9bf6-037c4b563278

Relying on `Variable` being in there and invalidating those expressions is pretty smart, but there are expressions that don't contain a `Variable` and need to be invalidated too.

I guess that's what those failures in phpstan-webmozart-assert are about too.

rajyan:
Thank you for the example. I see, this logic isn't correct...
I'll have to investigate them further.

For the phpstan-webmozart-assert failure, I think rvanvelzen's explanation https://github.com/phpstan/phpstan-src/pull/1929#issuecomment-1294541893 is right. (so it should fail as same as arrow function fails)

rajyan:
@ondrejmirtes 
Do you think the logic of https://github.com/phpstan/phpstan-src/pull/1934/commits/93ef0b8a7f86c0ebee0aa14bb452fca7ea5947d1 is correct enough?

ondrejmirtes:
I don't think that any changes around the Variables are correct. There's only one bug reported, and that's about function_exists(). So we need to retain this expression, and everything else needs to stay the same.

ondrejmirtes:
Also class_exists maybe.

rajyan:
I added the variable related specification logic in the migration of variableTypes and moreSpecificTypes
https://github.com/rajyan/phpstan-src/commit/3e20db31885a79361d3cc0c81d4b2353e3542f33

to keep the logic before
https://github.com/rajyan/phpstan-src/blob/76be530e341d6452fbd4a910b8302438904caa55/src/Analyser/MutatingScope.php#L2910-L2922

but I think it's not correct enough
because of the https://github.com/phpstan/phpstan-src/pull/1934#discussion_r1008005278 case.

This pull request was first intended to fix that issue.

rajyan:
so, I think I can reset this pull request to https://github.com/phpstan/phpstan-src/pull/1934/commits/93ef0b8a7f86c0ebee0aa14bb452fca7ea5947d1
and leave the https://github.com/phpstan/phpstan/issues/8205 open

rajyan:
Made your example explicit that there are non variable expressions that should be invalidated.
https://3v4l.org/9bXe9

ondrejmirtes:
> so, I think I can reset this pull request to https://github.com/phpstan/phpstan-src/commit/93ef0b8a7f86c0ebee0aa14bb452fca7ea5947d1

Yeah, sure :)

rajyan:
@ondrejmirtes 
Reset and added your comment as a test case :smile:

ondrejmirtes:
Thank you!

PR #1935, opened 2022-10-29
refactor arrow function scope logic to make it consistent with anonym…

rajyan:
…ous functions

Refactor anonymous function scope to use invalidate logic of assignVariable instead of the duplication.
This makes the logic consistent with
https://github.com/phpstan/phpstan-src/pull/1934

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2983,13 +2983,10 @@ class MutatingScope implements Scope
 	 */
 	private function enterArrowFunctionWithoutReflection(Expr\ArrowFunction $arrowFunction, ?array $callableParameters): self
 	{
-		$expressionTypes = $this->expressionTypes;
-		$mixed = new MixedType();
-		$parameterExprStrings = [];
-		$parameterVariableExpressions = [];
+		$arrowFunctionScope = $this;
 		foreach ($arrowFunction->params as $i => $parameter) {
 			if ($parameter->type === null) {
-				$parameterType = $mixed;
+				$parameterType = new MixedType();
 			} else {
 				$isNullable = $this->isParameterValueNullable($parameter);
 				$parameterType = $this->getFunctionType($parameter->type, $isNullable, $parameter->variadic);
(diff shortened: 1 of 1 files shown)


rajyan:
My thinking is that, `conditionalExpressionTypes` should be only handled in two places.
Specified in `specifyExpressionType` and invalidated in `assignVariable`.
This can make the merge of `conditionalExpressionTypes` and `ExpressionTypes` much easier.

ondrejmirtes:
Does this change anything in type inference? Should there be a test?

src/Analyser/MutatingScope.php
```diff
@@ -3013,93 +3010,31 @@ private function enterArrowFunctionWithoutReflection(Expr\ArrowFunction $arrowFu
...
-					}
-
-					$newHolders[] = $holder;
-				}
-
-				if (count($newHolders) === 0) {
-					continue;
-				}
-
-				$conditionalExpressions[$conditionalExprString] = $newHolders;
```
rajyan, line 3075: This line was never called because of  the above foreach in line 3028.

src/Analyser/MutatingScope.php
```diff
@@ -3013,93 +3010,31 @@ private function enterArrowFunctionWithoutReflection(Expr\ArrowFunction $arrowFu
...
-			foreach ($holders as $holder) {
-				foreach ($parameterExprStrings as $exprString) {
-					foreach (array_keys($holder->getConditionExpressionTypes()) as $conditionalExprString2) {
-						if ($exprString === $conditionalExprString2) {
-							continue 3;
-						}
-					}
-				}
-
-				$newHolders[] = $holder;
```
rajyan, line 3045: It is always newHolders === holders because of the continue 3

src/Analyser/MutatingScope.php
```diff
@@ -3013,93 +3010,31 @@ private function enterArrowFunctionWithoutReflection(Expr\ArrowFunction $arrowFu
...
-			[],
-			$this->afterExtractCall,
-			$this->parentScope,
+			$arrowFunctionScope->afterExtractCall,
+			$arrowFunctionScope->parentScope,
 		);
-
-		foreach ($parameterVariableExpressions as $expr) {
-			$scope = $scope->invalidateExpression($expr, true);
-		}
```
rajyan, line 3100: This line is handled in `assignVariable`'s `invalidateExpression`

rajyan:
@ondrejmirtes 

> Does this change anything in type inference? Should there be a test?

From the investigations above, I believe using `assignVarible` is equal to the code before.

ondrejmirtes:
Alright, great, thank you :)

PR #1939, opened 2022-10-30
Try to solve issue 5005

VincentLanglet:
I'm trying to solve https://github.com/phpstan/phpstan/issues/5005
the most of the issue seems solve, the only remaining part can be resume to
https://phpstan.org/r/c5b1a21c-3365-4cf3-b9b6-dca415afdfff

Things are that `resolveCommonMath` has a special behavior for UnionType.
When we're doing something like `int|null / int`, there is
```
foreach ($leftType->getTypes() as $type) {
     if ($type instanceof IntegerRangeType || $type instanceof ConstantIntegerType) {
	$resultType = $this->integerRangeMath($type, $expr, $rightType);
     } else {
	$resultType = $type;
    }

    $unionParts[] = $resultType;
}
```
but even if the left part is an int, the computed type might be an int|float when div is used.

This misbehave can be reproduce with other operators:
https://phpstan.org/r/b65b3e08-8f55-4ac7-920d-46e52f1fae06

So, instead of `$resultType = $type;`, I believe that we should write
```
$resultType = $this->resolveCommonMath($expr, $type, $rightType);
```
The result of an operation on a union of type is the union of the result of the operation on the types.

I still have one legacy test failing and I'm not sure how to solve this:
```
$number / $int
```
should be float|int, but I currently get `(float|int)`.

This is because
- $number is an union type `int|float`
- I compute `int / int` it's a BenevolentUnion(float, int)
- I compute 'float / int' it's a Float
- Then I do an union Union(BenevolentUnion(float, int), Float), it's a BenevolentUnion(float, int).

I'm not sure how I should solve this...

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -1418,42 +1418,42 @@ class InitializerExprTypeResolver
 	 */
 	private function resolveCommonMath(Expr\BinaryOp $expr, Type $leftType, Type $rightType): Type
 	{
-		if (($leftType instanceof IntegerRangeType || $leftType instanceof ConstantIntegerType || $leftType instanceof UnionType) &&
-			($rightType instanceof IntegerRangeType || $rightType instanceof ConstantIntegerType || $rightType instanceof UnionType) &&
-			!$expr instanceof BinaryOp\Pow) {
-
-			if ($leftType instanceof ConstantIntegerType) {
-				return $this->integerRangeMath(
-					$leftType,
-					$expr,
-					$rightType,
-				);
-			} elseif ($leftType instanceof UnionType) {
-
-				$unionParts = [];
-
-				foreach ($leftType->getTypes() as $type) {
-					if ($type instanceof IntegerRangeType || $type instanceof ConstantIntegerType) {
-						$resultType = $this->integerRangeMath($type, $expr, $rightType);
-					} else {
-						$resultType = $this->resolveCommonMath($expr, $type, $rightType);
-					}
-
-					$unionParts[] = $resultType;
-				}
-
-				$union = TypeCombinator::union(...$unionParts);
-				if ($union instanceof ErrorType) {
-					return new ErrorType();
-				}
-
-				if ($leftType instanceof BenevolentUnionType) {
-					return TypeUtils::toBenevolentUnion($union)->toNumber();
-				}
-
-				return $union->toNumber();
+		if ($leftType instanceof UnionType) {
+			$unionParts = [];
+			foreach ($leftType->getTypes() as $type) {
+				$unionParts[] = $this->resolveCommonMath($expr, $type, $rightType);
 			}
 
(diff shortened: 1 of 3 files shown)


src/Rules/RuleLevelHelper.php
```diff
@@ -221,23 +221,25 @@ public function findTypeToCheck(
...
-
-					$newTypes[] = $innerType;
-				}
+		if (!$this->checkUnionTypes && $type instanceof ObjectWithoutClassType) {
+			return new FoundTypeResult(new ErrorType(), [], [], null);
+		}
 
-				if (count($newTypes) > 0) {
-					return new FoundTypeResult(TypeCombinator::union(...$newTypes), $directClassNames, [], null);
+		if (
```
VincentLanglet, line 228: While working on this issue, I got multiple error message
```
Binary operation "+" between (int|string) and int results in an error.
```
which should indeed be an error when checking benevolent union.

I discovered that in the PR https://github.com/phpstan/phpstan-src/pull/1930, I modified `accepts` but not `findTypeToCheck`.

This is fixed with this.

VincentLanglet:
I'll need some advices/guidelines to move on this.

I might be wrong but I feel like the refacto done is going in a good way, calling `resolveCommonMath` on an UnionType should be resolved as the union of the call `resolveCommonMath` on every types. But I have some issues:
- `$number / $int` is `(float|int)` as already explained.
- I get a lot of `Binary operation "+" between (int|string) and 1 results in an error`

I feel like the second part is kinda tricky because technically, the error should be reported with `$this->checkBenevolentUnionTypes = true`.
If I try to add support for `checkBenevolentUnionTypes` inside the `RuleLevelHelper::findTypeToCheck` method, with something like
```
if (
     !$this->checkUnionTypes && $type instanceof UnionType && !$type instanceof BenevolentUnionType
     || !$this->checkBenevolentUnionTypes && $type instanceof BenevolentUnionType
) {
```
It solve all my issues, but create new ones because of the ArrayUnpackingRule

This rule is currently reporting error when unpacking an `array<string, mixed>`, `array<int|string, mixed>` or `array<array-key, mixed>`. This seems to be the only rule which reports error for BenevolentUnion (here array-key).
So if I add the `checkBenevolentUnionTypes` support, I need to remove the error from `array-key` like I did in the PR https://github.com/phpstan/phpstan-src/pull/1940

staabm:
This code should be moved into the typesystem, like we do it with https://github.com/phpstan/phpstan-src/pull/1904

We shoud decide wether this fix should be the last fix using the old logic, or maybe should instead introduce `Type->divide(...): Type` which in turn would cleanup some of the `instanceof *Type` cases

VincentLanglet:
> This code should be moved into the typesystem, like we do it with #1904
> 
> We shoud decide wether this fix should be the last fix using the old logic, or maybe should instead introduce `Type->divide(...): Type` which in turn would cleanup some of the `instanceof *Type` cases

Then I close this.
It's not a big issue, and I still didn't have find the perfect solution.

Might be easier after the new type system.

PR #1940, opened 2022-10-30
Add more support for benevolentUnion config

VincentLanglet:
While working on https://github.com/phpstan/phpstan-src/pull/1939, I discovered that `findTypeToCheck` were also checking BenevolentUnion when the option `checkUnionType` were used.

This would report error like 
```
Binary operation "+" between (int|string) and 1 results in an error
```
only if `checkBenevolentUnion` is enabled and solve most of the error of the #1939 pr.

The only issue I have is that I had to modify the `ArrayUnpackingRule` since it was the only rule which was already checking benevolent union (array-key). I dunno if we should make a special behavior, and how.

The integration tests failure doesn't seem related to the PR.

### DIFF
--- a/src/Rules/RuleLevelHelper.php
+++ b/src/Rules/RuleLevelHelper.php
@@ -226,8 +226,14 @@ class RuleLevelHelper
 		}
 
 		if (
-			!$this->checkUnionTypes && $type instanceof UnionType && !$type instanceof BenevolentUnionType
-			|| !$this->checkBenevolentUnionTypes && $type instanceof BenevolentUnionType
+			(
+				!$this->checkUnionTypes
+				&& $type instanceof UnionType
+				&& !$type instanceof BenevolentUnionType
+			) || (
+				!$this->checkBenevolentUnionTypes
+				&& $type instanceof BenevolentUnionType
+			)
 		) {
 			$newTypes = [];
 
--- a/src/Rules/RuleLevelHelper.php
+++ b/src/Rules/RuleLevelHelper.php
@@ -221,23 +221,26 @@ class RuleLevelHelper
 			return new FoundTypeResult(new ErrorType(), [], $errors, null);
 		}
 
-		if (!$this->checkUnionTypes) {
-			if ($type instanceof ObjectWithoutClassType) {
-				return new FoundTypeResult(new ErrorType(), [], [], null);
+		if (!$this->checkUnionTypes && $type instanceof ObjectWithoutClassType) {
+			return new FoundTypeResult(new ErrorType(), [], [], null);
+		}
+
+		if (
+			!$this->checkUnionTypes && $type instanceof UnionType && !$type instanceof BenevolentUnionType
+			|| !$this->checkBenevolentUnionTypes && $type instanceof BenevolentUnionType
+		) {
+			$newTypes = [];
+
+			foreach ($type->getTypes() as $innerType) {
+				if (!$unionTypeCriteriaCallback($innerType)) {
+					continue;
+				}
+
+				$newTypes[] = $innerType;
 			}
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Arrays/data/array-unpacking.php
```diff
@@ -18,12 +18,18 @@ function stringKeyedArray(array $bar)
 	$baz = [...$bar];
 }
 
-/** @param array<array-key, string> $bar */
+/** @param array<int|string, string> $bar */
```
ondrejmirtes: Do not change existing tests, add new ones instead.

tests/PHPStan/Rules/Arrays/data/array-unpacking.php
```diff
@@ -18,12 +18,18 @@ function stringKeyedArray(array $bar)
...
 function unionKeyedArray(array $bar)
 {
 	$baz = [...$bar];
 }
 
+/** @param array<array-key, string> $bar */
+function benevolentUnionKeyedArray(array $bar)
+{
+	$baz = [...$bar];
```
ondrejmirtes, line 24: Add new tests at the end of the file, otherwise it's hard to review the changes.

ondrejmirtes (changes requested):
The errors seemed familiar, looks like you based your work on top of old commit, I'm trying to rebase your branch.

tests/PHPStan/Rules/Arrays/data/array-unpacking.php
VincentLanglet: The diff should be better now

src/Rules/RuleLevelHelper.php
```diff
@@ -221,23 +221,26 @@ public function findTypeToCheck(
...
-				}
+		if (!$this->checkUnionTypes && $type instanceof ObjectWithoutClassType) {
+			return new FoundTypeResult(new ErrorType(), [], [], null);
+		}
 
-				if (count($newTypes) > 0) {
-					return new FoundTypeResult(TypeCombinator::union(...$newTypes), $directClassNames, [], null);
+		if (
+			!$this->checkUnionTypes && $type instanceof UnionType && !$type instanceof BenevolentUnionType
+			|| !$this->checkBenevolentUnionTypes && $type instanceof BenevolentUnionType
```
ondrejmirtes: This needs parentheses for disambiguation, I have no idea what it does now.

ondrejmirtes (changes requested):

src/Rules/RuleLevelHelper.php
VincentLanglet: You mean
```
(
     !$this->checkUnionTypes
     && $type instanceof UnionType
     && !$type instanceof BenevolentUnionType
) || (
    !$this->checkBenevolentUnionTypes
    && $type instanceof BenevolentUnionType
)
```

Or is it more clear another way, like
```
if ($type instanceof UnionType) {
     if (
             (!$this->checkUnionTypes && !$type instanceof BenevolentUnionType)
              || (!$this->checkBenevolentUnionTypes && $type instanceof BenevolentUnionType)
         ) {
         ...
         }
}
```

What do you prefer ?

ondrejmirtes: This:

```php
(
	!$this->checkUnionTypes
	&& $type instanceof UnionType
	&& !$type instanceof BenevolentUnionType
) || (
	!$this->checkBenevolentUnionTypes
	&& $type instanceof BenevolentUnionType
)
```

VincentLanglet: Done

ondrejmirtes:
Thank you.

PR #636, opened 2021-08-22
verify constants via array_key_exists

voku:
issue: https://github.com/phpstan/phpstan/issues/4174

I looked into ```ArrayKeyExistsFunctionTypeSpecifyingExtension```, but it seems like I didn't fully understand the logic of the specifying classes. In this case we want to specify parameter 1 or/and 2, can we currently do that?

PR #1941, opened 2022-10-30
Improve ArrayKeyExists specifyingExtension

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/2970

### DIFF
--- a/src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
@@ -65,6 +65,8 @@ class ArrayKeyExistsFunctionTypeSpecifyingExtension implements FunctionTypeSpeci
 				$arrayKeyType = $arrayType->getIterableKeyType();
 				if ($keyType->isString()->yes()) {
 					$arrayKeyType = $arrayKeyType->toString();
+				} elseif ($keyType->isString()->maybe()) {
+					$arrayKeyType = TypeCombinator::union($arrayKeyType, $arrayKeyType->toString());
 				}
 
 				$specifiedTypes = $this->typeSpecifier->create(
--- a/src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
@@ -19,6 +19,7 @@ use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\FunctionTypeSpecifyingExtension;
 use PHPStan\Type\MixedType;
+use PHPStan\Type\StringType;
 use PHPStan\Type\TypeCombinator;
 use function count;
 use function in_array;
@@ -57,13 +58,37 @@ class ArrayKeyExistsFunctionTypeSpecifyingExtension implements FunctionTypeSpeci
 		$array = $node->getArgs()[1]->value;
 		$keyType = $scope->getType($key);
 		$arrayType = $scope->getType($array);
-		if (!$keyType instanceof ConstantIntegerType && !$keyType instanceof ConstantStringType && !$arrayType->isIterableAtLeastOnce()->no()) {
+
+		if (!$keyType instanceof ConstantIntegerType
+			&& !$keyType instanceof ConstantStringType
+			&& !$arrayType->isIterableAtLeastOnce()->no()) {
 			if ($context->truthy()) {
+				$arrayKeyType = $arrayType->getIterableKeyType();
+				if ((new StringType())->isSuperTypeOf($keyType)->yes()) {
+					$arrayKeyType = $arrayKeyType->toString();
+				}
+
+				$specifiedTypes = $this->typeSpecifier->create(
+					$key,
+					$arrayKeyType,
+					$context,
+					false,
+					$scope,
+				);
+
 				$arrayDimFetch = new ArrayDimFetch(
 					$array,
 					$key,
 				);
-				return $this->typeSpecifier->create($arrayDimFetch, $arrayType->getIterableValueType(), $context, false, $scope, new Identical($arrayDimFetch, new ConstFetch(new Name('__PHPSTAN_FAUX_CONSTANT'))));
+
+				return $specifiedTypes->unionWith($this->typeSpecifier->create(
+					$arrayDimFetch,
+					$arrayType->getIterableValueType(),
+					$context,
+					false,
+					$scope,
+					new Identical($arrayDimFetch, new ConstFetch(new Name('__PHPSTAN_FAUX_CONSTANT')))
+				));
 			}
 
 			return new SpecifiedTypes();
--- a/src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
@@ -19,7 +19,6 @@ use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\FunctionTypeSpecifyingExtension;
 use PHPStan\Type\MixedType;
-use PHPStan\Type\StringType;
 use PHPStan\Type\TypeCombinator;
 use function count;
 use function in_array;
@@ -64,7 +63,7 @@ class ArrayKeyExistsFunctionTypeSpecifyingExtension implements FunctionTypeSpeci
 			&& !$arrayType->isIterableAtLeastOnce()->no()) {
 			if ($context->truthy()) {
 				$arrayKeyType = $arrayType->getIterableKeyType();
-				if ((new StringType())->isSuperTypeOf($keyType)->yes()) {
+				if ($keyType->isString()->yes()) {
 					$arrayKeyType = $arrayKeyType->toString();
 				}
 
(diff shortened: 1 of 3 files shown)


src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
```diff
@@ -57,13 +58,37 @@ public function specifyTypes(
...
 		$arrayType = $scope->getType($array);
-		if (!$keyType instanceof ConstantIntegerType && !$keyType instanceof ConstantStringType && !$arrayType->isIterableAtLeastOnce()->no()) {
+
+		if (!$keyType instanceof ConstantIntegerType
+			&& !$keyType instanceof ConstantStringType
+			&& !$arrayType->isIterableAtLeastOnce()->no()) {
 			if ($context->truthy()) {
+				$arrayKeyType = $arrayType->getIterableKeyType();
+				if ((new StringType())->isSuperTypeOf($keyType)->yes()) {
+					$arrayKeyType = $arrayKeyType->toString();
```
herndlm, line 67: Jfyi we have `isString()` already too which I personally prefer. But it should come down to the same thing I guess. UPDATE: this was meant for the line above, sorry

src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
```diff
@@ -57,13 +58,37 @@ public function specifyTypes(
...
 				);
-				return $this->typeSpecifier->create($arrayDimFetch, $arrayType->getIterableValueType(), $context, false, $scope, new Identical($arrayDimFetch, new ConstFetch(new Name('__PHPSTAN_FAUX_CONSTANT'))));
+
+				return $specifiedTypes->unionWith($this->typeSpecifier->create(
+					$arrayDimFetch,
+					$arrayType->getIterableValueType(),
+					$context,
+					false,
+					$scope,
+					new Identical($arrayDimFetch, new ConstFetch(new Name('__PHPSTAN_FAUX_CONSTANT')))
```
herndlm:
```suggestion
					new Identical($arrayDimFetch, new ConstFetch(new Name('__PHPSTAN_FAUX_CONSTANT'))),
```
missing trailing comma as also reported by phpcs

VincentLanglet: Fixed

src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php
VincentLanglet: I changed

voku:
Maybe you can add some extra tests from this old pull request? 🤔 : https://github.com/phpstan/phpstan-src/pull/636

VincentLanglet:
> Maybe you can add some extra tests from this old pull request? 🤔 : #636

I take a look and there is only 6 check about `array_key_exists` and I feel like they already covered by the test:
```
if (array_key_exists($key, $b)) {
     assertType("'3'|'foo'", $key);
}
```

But I can add more tests if needed

VincentLanglet:
Failure seems unrelated

VincentLanglet:
There is a new error reported by phpstan in the prestashop codebase:
```
/**
     * Apply filters to address query builder.
     *
     * @param array $filters
     * @param QueryBuilder $qb
     */
    private function applyFilters(QueryBuilder $qb, array $filters)
    {
        $allowedFiltersMap = [
            'id_customer' => 'a.id_customer',
        ];

        foreach ($filters as $filterName => $value) {
            if (!array_key_exists($filterName, $allowedFiltersMap) || empty($value)) {
                continue;
            }

            if ('id_customer' === $filterName) {
                $qb->andWhere('a.`id_customer` = :' . $filterName);
                $qb->setParameter($filterName, $value);

                continue;
            }

            $qb->andWhere($allowedFiltersMap[$filterName] . ' LIKE :' . $filterName)
                ->setParameter($filterName, '%' . $value . '%');
        }
    }
```
The end of the foreach is indeed never reach, because of the array_key_exists check, the only `$filterName` value allowed is  `'id_customer'` which is handled by the following if.

There is also a new error in doctrine/dbal reporting that
```
if (array_key_exists($join['joinAlias'], $knownAliases)) {
    throw QueryException::nonUniqueAlias($join['joinAlias'], array_keys($knownAliases));
}
```
`$join['joinAlias`] might be an int, when `$knownAliases` is `array<string, mixed>`

This might be considered as annoying.

VincentLanglet:
> There is also a new error in doctrine/dbal reporting that
> 
> ```
> if (array_key_exists($join['joinAlias'], $knownAliases)) {
>     throw QueryException::nonUniqueAlias($join['joinAlias'], array_keys($knownAliases));
> }
> ```
> 
> `$join['joinAlias`] might be an int, when `$knownAliases` is `array<string, mixed>`
> 
> This might be considered as annoying.

I added more tests, `array_key_exists($string, $arrayOfKeyString)`, is correctly considered as a string.

The fact is in the doctrine code, we are inside a foreach, and we're modifying `$knowAliases` by setting new value with keys that are mixed, so the initial `array<string, mixed>` array become `array<int|string, mixed>`. Therefor the new error reported by PHPStan is correct too.

tests/PHPStan/Analyser/data/array-key-exists.php
```diff
@@ -12,20 +12,89 @@ class Foo
...
+		}
+
+		if (array_key_exists($key1, $c)) {
+			assertType('int', $key1);
+		}
+		if (array_key_exists($key2, $c)) {
+			assertType('string', $key2);
+		}
+		if (array_key_exists($key3, $c)) {
+			assertType('(int|string)', $key3);
```
VincentLanglet, line 88: This is the same behavior than https://phpstan.org/r/5b1ebeca-e8a5-4e01-a2cb-9abda793ce4f

VincentLanglet:
Just rebased the branch if you have time to review @ondrejmirtes

To me the test failure are correct, cf https://github.com/phpstan/phpstan-src/pull/1941#issuecomment-1304788241

tests/PHPStan/Analyser/data/array-key-exists.php
```diff
@@ -12,20 +12,89 @@ class Foo
...
 	 * @return void
 	 */
-	public function doFoo(array $a, string $key): void
+	public function doFoo(array $a, string $key, int $anotherKey): void
 	{
 		assertType('false', array_key_exists(2, $a));
 		assertType('bool', array_key_exists('foo', $a));
 		assertType('false', array_key_exists('2', $a));
 
-		$a = ['foo' => 2, 3 => 'bar'];
```
ondrejmirtes, line 21: Please do not change existing tests like this. This diff is hard to review.

Of course, if you improved behaviour and existing asserts changed, you need to edit them, but otherwise add new tests instead. I don't want to change meaning of variables here, please continue testing `$a` there as it was tested before.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/array-key-exists.php
VincentLanglet: Sure, I changed to keep the old name of the variable.
I also rebase the branch on 1.9.x.

ondrejmirtes (changes requested):
I'd like some more tests about tricky situations. Let's say we have an array shape like:

```
$a = [3 => 'foo', 4 => 'bar']
```

When we have string `$key` and we ask about `array_key_exists($key, $a)`, the `$key` should become `'3'|'4'`.

It should work generally, so `int|string` should become `3|4|'3'|'4'`.

VincentLanglet:
> I'd like some more tests about tricky situations. Let's say we have an array shape like:
> 
> ```
> $a = [3 => 'foo', 4 => 'bar']
> ```
> 
> When we have string `$key` and we ask about `array_key_exists($key, $a)`, the `$key` should become `'3'|'4'`.
> 
> It should work generally, so `int|string` should become `3|4|'3'|'4'`.

Test added with a little fix to get the expected behavior

VincentLanglet:
For the build failure:
https://github.com/phpstan/phpstan-src/actions/runs/3869539129/jobs/6595709562

```
/**
     * Apply filters to address query builder.
     *
     * @param array $filters
     * @param QueryBuilder $qb
     */
    private function applyFilters(QueryBuilder $qb, array $filters)
    {
        $allowedFiltersMap = [
            'id_customer' => 'a.id_customer',
        ];

        foreach ($filters as $filterName => $value) {
            if (!array_key_exists($filterName, $allowedFiltersMap) || empty($value)) {
                continue;
            }

            if ('id_customer' === $filterName) {
                $qb->andWhere('a.`id_customer` = :' . $filterName);
                $qb->setParameter($filterName, $value);

                continue;
            }

            $qb->andWhere($allowedFiltersMap[$filterName] . ' LIKE :' . $filterName)
                ->setParameter($filterName, '%' . $value . '%');
        }
    }
```
The end of the foreach is indeed never reach, because of the array_key_exists check, the only `$filterName` value allowed is  `'id_customer'` which is handled by the following if.

https://github.com/phpstan/phpstan-src/actions/runs/3869539129/jobs/6595709711
```
if (array_key_exists($join['joinAlias'], $knownAliases)) {
    throw QueryException::nonUniqueAlias($join['joinAlias'], array_keys($knownAliases));
}
```
The fact is in the doctrine code, we are inside a foreach, and we're modifying `$knowAliases` by setting new value with keys that are mixed, so the initial `array<string, mixed>` array become `array<int|string, mixed>`. Therefor the new error reported by PHPStan is correct too.

VincentLanglet:
@ondrejmirtes Do I need to target 1.10.x now ?

ondrejmirtes:
@VincentLanglet No https://github.com/phpstan/phpstan-src/pull/2043#issuecomment-1375607286

ondrejmirtes:
Thank you.

PR #1943, opened 2022-10-31
Promote native types

ondrejmirtes:


rajyan:
I checked
Rector, composer, phpstan-webmozart-assert test failures looks like caused from improvement of native types, so valid.

not sure enough with nextras/orm
https://github.com/phpstan/phpstan-src/actions/runs/3364624981/jobs/5579240300
but basically look correct.

The first failure of phpstan-phpunit is wrong.
https://github.com/phpstan/phpstan-src/actions/runs/3364624981/jobs/5579231181

rajyan:
Looks like we are almost there!

ondrejmirtes:
Some of the existing errors are because getNativeType() for MethodCall should only read native typehints. I'm already working on it locally 😊

And the todos in that issue will help with various little bugs...

I think we can release 1.9.0 at any time and leave this for a future version, it's already feature-packed and 1.9.x branch is stable enough now 😊

Another problem in this PR is performance, calling TypeSpecifier twice slows PHPStan down, I'd love to see some ideas for optimizations.

rajyan:
> I think we can release 1.9.0 at any time and leave this for a future version, it's already feature-packed and 1.9.x branch is stable enough now 😊

Looking forward for the release!!

> Another problem in this PR is performance, calling TypeSpecifier twice slows PHPStan down, I'd love to see some ideas for optimizations.

Yeah, performance was one problem I was worried about when thinking about native types.
I would look for optimizations too.

ondrejmirtes:
It's not perfect but it's fine :)

rajyan:
I'm glad to know and thank you for improving the Native types idea to a level where it can be merged!

PR #1944, opened 2022-10-31
handle ClassConstFetch in scope

rajyan:
fixes https://github.com/phpstan/phpstan/issues/8034

blocking
https://github.com/phpstan/phpstan/issues/8191

regression from
https://github.com/phpstan/phpstan-src/pull/1938

Steps to fix

- [x] fix DefinedConstantTypeSpecifyingExtension to detect ClassConstFetch and specify them correctly
- [x]  fix ClassConst assigned as ConstFetch in NodeScopeResolver https://github.com/rajyan/phpstan-src/blob/69d7e8185af250ce7315a2000b88a3d31f84cf3a/src/Analyser/NodeScopeResolver.php#L1439  assign them as ClassConstFetch
- [x] skip ClassConstantRule if hasExpressionType->yes
- [x] add expressionTypes handling in resolveType of ClassConstFetch

### DIFF
--- a/src/Type/Php/DefinedConstantTypeSpecifyingExtension.php
+++ b/src/Type/Php/DefinedConstantTypeSpecifyingExtension.php
@@ -15,6 +15,7 @@ use PHPStan\Type\FunctionTypeSpecifyingExtension;
 use PHPStan\Type\MixedType;
 use function count;
 use function explode;
+use function ltrim;
 
 class DefinedConstantTypeSpecifyingExtension implements FunctionTypeSpecifyingExtension, TypeSpecifierAwareExtension
 {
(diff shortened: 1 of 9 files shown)


rajyan:
It's obvious that 
https://github.com/phpstan/phpstan-webmozart-assert/blob/944b629c3cf3995aab6eb1ce28bfe0fe8bc20f96/tests/Type/WebMozartAssert/data/type.php#L250
failure comes from the last commit
https://github.com/phpstan/phpstan-src/pull/1944/commits/db2ecb7e6a5319662b067ef1d6f02129793df73e

tests/PHPStan/Rules/Classes/data/class-const-fetch-defined.php
```diff
@@ -0,0 +1,51 @@
...
+			Foo::TEST;
+			\Foo::TEST;
+			\ClassConstFetchDefined\Foo::TEST;
+		} else {
+			Foo::TEST;
+			\Foo::TEST;
+			\ClassConstFetchDefined\Foo::TEST;
+		}
+	}
+}
```
rajyan, line 51: verified the behavior by https://3v4l.org/ueJTH

src/Type/Php/DefinedConstantTypeSpecifyingExtension.php
```diff
@@ -35,7 +35,7 @@ public function isFunctionSupported(
 	{
 		return $functionReflection->getName() === 'defined'
 			&& count($node->getArgs()) >= 1
-			&& !$context->null();
+			&& $context->truthy();
```
rajyan, line 38: Not exist !== NeverType so I think this makes things better.

ondrejmirtes:
Perfect, thank you!

PR #1946, opened 2022-11-01
added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/7913

### DIFF
--- a/tests/PHPStan/Rules/Arrays/data/bug-7913.php
+++ b/tests/PHPStan/Rules/Arrays/data/bug-7913.php
@@ -1,0 +1,17 @@
+<?php declare(strict_types = 1);
+
+namespace Bug7913;
+
+const X = [];
+if (!empty(X)) {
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Arrays/data/bug-7913.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types = 1);
+
+namespace Bug7913;
+
+const X = [];
```
staabm, line 5: I tried wrapping it in a method/function, but using `const` like in the repro example, only works in top level code, it seems

ondrejmirtes:
I'd like a NodeScopeResolverTest with assertType before and after empty().

ondrejmirtes:
Rebase please :) (always make sure you're on the latest commit before starting the work).

ondrejmirtes:
Thank you!

PR #1947, opened 2022-11-01
Fix impossible type checks for array offset/property fetch

rvanvelzen:


staabm:
Do we also need some love for constants?

https://phpstan.org/r/47694dbc-f2ff-40eb-a2cc-30f037961a1c

rvanvelzen:
> Do we also need some love for constants?

While possible in theory, I couldn't even write a simple reproducer that actually makes sense, so I'm not really interested :)

rvanvelzen:
I believe the failures are all legitimate issues uncovered by this PR.

ondrejmirtes:
Thank you!

ondrejmirtes:
I decided to revert this one, it's pretty annoying because PHPStan is now more pedant about this https://phpstan.org/blog/remembering-and-forgetting-returned-values and it has some non-zero impact in real-world projects.

I want to release PHPStan 1.9.0 tomorrow and this is gonna need some more work.

Also it uncovered some related bugs, for example (in PHPUnit test):

```php
$data = $di->get(xxx::class)->getBundle($bundle);
self::assertTrue($data->isDefault);

$bundle->setDefaultProduct(null);
$entityManager->flush($bundle);
$data = $di->get(xxx::class)->getBundle($bundle);
self::assertFalse($data->isDefault); // PHPStan says that true can't be false
```

PR #1948, opened 2022-11-01
Add rule for casting objects to string

rvanvelzen:
Rule extracted from #1905 because it's useful without explicit `stringable-object` support as well


ondrejmirtes (changes requested):
I feel like we don't need a new rule here.

We need `ObjectWithoutClassType::toString()` to return ErrorType.

ObjectType already does this correctly.

Also, `MixedType::toString()` should be looked into for subtracted types.

ondrejmirtes:
Nice! I also love this error:

```

 ------ ---------------------------------------------------------------- 
  Line   lib/Doctrine/ORM/Query/Expr/Base.php                            
 ------ ---------------------------------------------------------------- 
  104    Parameter #2 $array of function implode expects array<string>,  
         array<int, object|string> given.                                
 ------ ---------------------------------------------------------------- 
```

ondrejmirtes:
Thank you!

PR #1949, opened 2022-11-01
support subtracted-mixed in toNumber()

staabm:


src/Type/MixedType.php
```diff
@@ -403,6 +403,18 @@ public function toBoolean(): BooleanType
...
+		if ($this->subtractedType !== null) {
+			$float = new FloatType();
+			$int = new IntegerType();
+
+			if ($float->isSuperTypeOf($this->subtractedType)->yes()) {
+				return $int;
+			}
+			if ($int->isSuperTypeOf($this->subtractedType)->yes()) {
+				return $float;
+			}
```
rvanvelzen: What if both `float` and `int` are subtracted?

herndlm: sorry, can someone explain to me how this works in general? if something is e.g. not a float it could be e.g. an object still and objects can't be incremented?

staabm: > What if both `float` and `int` are subtracted?

yeah, we could return a `NeverType` then.

> sorry, can someone explain to me how this works in general? if something is e.g. not a float it could be e.g. an object still and objects can't be incremented?

the incrementing test-case is kind of synthetic.. I "reverse engineered" it from looking at `toNumber()` uses in the codebase:
https://github.com/phpstan/phpstan-src/blob/213863e23a5f0736f6d46511f3315e3dc22d3cb5/src/Analyser/MutatingScope.php#L975-L977

decting invalid cases like `$object++` is the job of a rule, not the narrowing done here I think

tests/PHPStan/Analyser/data/mixed-to-number.php
```diff
@@ -0,0 +1,37 @@
...
+		}
+	}
+
+	public function sayFooBar($mixed): void
+	{
+		if (!is_float($mixed)) {
+			if (!is_int($mixed)) {
+				assertType('mixed~float|int', $mixed);
+				$mixed++;
+				assertType('*NEVER*', $mixed);
```
rvanvelzen: Could still be a string though? https://3v4l.org/D4BHP

staabm: yeah,.. I think the src/ changes actually makes sense .. but my test is to synthetic and your example shows that we cannot return a `Never` type for `toNumber`

ondrejmirtes (changes requested):
Please return `ErrorType` if the subtracted type makes sure it can never be a number. Thanks.

ondrejmirtes:
Oh actually, other types can be cast to number as well 🤦

src/Type/MixedType.php
ondrejmirtes: Even if float is subtracted, the result can still be a float... Because `MixedType` without `FloatType` can still be numeric-string for example... https://3v4l.org/StqSP

staabm:
thank you guys for all the examples and ideas.

I think we / I learnt from it, that we can't take subtractable into account in `MixedType->toNumber`, as there are a lot of situations which will lead to `int`,  `float` or `int|float` even when the summand types are not even numbers.

I put all the examples into a unit test and fixed a edge case with `mixed~array` by that.

the PR no longer does what its title suggested, but at least our time discussing it is not lost, but manifested in more testcoverage.

thanks again.

ondrejmirtes:
Yeah, at least it was still productive :) Thank you.

PR #1951, opened 2022-11-01
Merge expression types and native expression types

rajyan:
resolving todos in https://github.com/phpstan/phpstan/issues/8191

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -248,6 +248,13 @@ class MutatingScope implements Scope
 			null,
 			null,
 			$this->expressionTypes,
+			[],
+			null,
+			null,
+			true,
+			[],
+			[],
+			$this->nativeExpressionTypes,
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


rajyan:
After resolving the inconsistencies, I think it might be better to have `nativeType` and `nativeCertainty` in `ExpressionTypeHolder` and merge nativeExpression types. With it, we can merge some duplicate of  code for native types which can improve performance in some cases.

rajyan:
I realized that, the current conditionalExpressionTypes logic can make the count of native and expression types differ.

src/Analyser/MutatingScope.php
```diff
@@ -3861,8 +3963,10 @@ public function filterBySpecifiedTypes(SpecifiedTypes $specifiedTypes): self
 
 			if ($typeHolder->getCertainty()->no()) {
 				unset($scope->expressionTypes[$variableExprString]);
+				unset($scope->nativeExpressionTypes[$variableExprString]);
 			} else {
 				$scope->expressionTypes[$variableExprString] = $typeHolder;
+				$scope->nativeExpressionTypes[$variableExprString] = $typeHolder;
```
rajyan, line 3966: > I realized that, the current conditionalExpressionTypes logic can make the count of native and expression types differ.

this is the incorrect one. Added just to see the temporal validation passes.

rajyan:
On second thought I think https://github.com/phpstan/phpstan-src/pull/1951#issuecomment-1299371448 is possible. I would clean up this PR first.

rajyan:
If expressionType exists for some expression, nativeExpressionType should always exist too. Although the certainty can be different.

ondrejmirtes:
> After resolving the inconsistencies, I think it might be better to have nativeType and nativeCertainty in ExpressionTypeHolder

My problem is that currently it's not possible to have an entry in these arrays with certainty=no.

src/Analyser/MutatingScope.php
```diff
@@ -365,6 +372,40 @@ public function afterClearstatcacheCall(): self
...
 			}
 		}
+		$nativeExpressionTypes = $this->nativeExpressionTypes;
+		foreach (array_keys($nativeExpressionTypes) as $exprString) {
+			// list from https://www.php.net/manual/en/function.clearstatcache.php
+
+			// stat(), lstat(), file_exists(), is_writable(), is_readable(), is_executable(), is_file(), is_dir(), is_link(), filectime(), fileatime(), filemtime(), fileinode(), filegroup(), fileowner(), filesize(), filetype(), and fileperms().
+			foreach ([
+						 'stat',
+						 'lstat',
```
ondrejmirtes, line 382: This is exactly the kind of duplication I'd like to avoid :)

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
rajyan: Definitely. I just wanted to test that it is possible to make count of native and full type same, if we add them everywhere.

rajyan:
> My problem is that currently it's not possible to have an entry in these arrays with certainty=no.

Yeah, I think we currently don't have a situation that expressionTypes holds certainty "no", because we are unsetting the expression when invalidated.

src/Analyser/MutatingScope.php
```diff
@@ -248,6 +248,13 @@ public function enterDeclareStrictTypes(): self
 			null,
 			null,
 			$this->expressionTypes,
+			[],
+			null,
+			null,
+			true,
+			[],
+			[],
+			$this->nativeExpressionTypes,
```
rajyan, line 257: maybe this one is not correct too. There should be no situation that native type assigned before declare

rajyan:
Sorry, but I would close this one for now.

PR #1952, opened 2022-11-02
support subtracted-mixed in toArray()

staabm:


staabm:
ahh neh... doesn't make sense

PR #1954, opened 2022-11-03
Fix return type of Ds\Sequence#copy()

MidnightDesign:
I don't know why the return type isn't correctly inferred. [It seems fine to me.](https://github.com/phpstan/phpstan-src/blob/1.9.x/stubs/ext-ds.stub#L398) Any ideas?

[Playground link](https://phpstan.org/r/c7881107-0ded-4599-8225-78cdb089b0c6)

### DIFF
--- a/stubs/ext-ds.stub
+++ b/stubs/ext-ds.stub
@@ -18,7 +18,7 @@ use UnderflowException;
 interface Collection extends IteratorAggregate, Countable, JsonSerializable
 {
 	/**
-	 * @return Collection<TKey, TValue>
+	 * @return static
 	 */
 	public function copy();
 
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
The method probably isn't declared on `Sequence`. The correct solution might be to write `@return static` above `Collection::copy()`.

MidnightDesign:
I've tried removing `copy()` from `Sequence` and making the return type of `Collection` `static<TKey, TValue>`, but that didn't change anything. I think it can't work because `Collection` takes two type arguments, and `Sequence` takes just one. I have no idea how to fix this.

ondrejmirtes:
Try just `static`

MidnightDesign:
You're right, that works. Thank you for your help!

ondrejmirtes:
Thank you!

PR #1956, opened 2022-11-03
Fix mt_rand return type

enumag:
Closes https://github.com/phpstan/phpstan/issues/8272

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-8272.php
+++ b/tests/PHPStan/Analyser/data/bug-8272.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Bug8272;
+
+use function PHPStan\Testing\assertType;
+
+function test(): void {
+	assertType('int<1, 5>', mt_rand(1, 5));
+	assertType('int<0, max>', mt_rand());
+};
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
What did you do to the icon? 👓

enumag:
Huh? Nothing at all. I didn't touch it.

enumag:
It must have somehow been deleted by git automatically because I have .idea in global gitignore... I'll add it back...

ondrejmirtes:
You deleted the file.

enumag:
Fixed.

tests/PHPStan/Analyser/data/bug-8272.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+use function PHPStan\Testing\assertType;
+
+assertType('int<1, 5>', mt_rand(1, 5));
```
staabm: This test file is not included from any test-class?

Should be yielded from NodeScopeResolverTest
(And the line wrapped in a method or function)

ondrejmirtes: And in a namespace.

canvural: And maybe tested with no arguments.

ondrejmirtes:
I can't review it when the tests are failing.

enumag:
@ondrejmirtes Fixed. There is still one failed job but I don't think it's related to my changes: https://github.com/phpstan/phpstan-src/actions/runs/3393592592/jobs/5641134975

ondrejmirtes:
Thank you.

PR #1958, opened 2022-11-03
Fix bug 8280

rajyan:
fixes https://github.com/phpstan/phpstan/issues/8280

There are other situations that constant arrays are not correctly marked as list, I would fix them with a following PR


ondrejmirtes:
Thank you!

PR #1959, opened 2022-11-03
Fix array casts as list type

rajyan:
fixes array cast of types as list


rajyan:
For numeric, falsy, non-empty string, it's weird that the casted type is `array<0, non-empty-string>` ~~(which is wrong)~~. I think there is another underlying bug.

rajyan:
ok, it was not wrong, but intersection of two constant arrays narrowed to array type was the cause.

rajyan:
Very weird only failing on windows...

rajyan:
solved
the namespace was wrong

rajyan:
Noticed that failing tests of composer is caused from my previous PR https://github.com/phpstan/phpstan-src/commit/8b456c524c6dd8a782569c8c23e727189494ec32
I'll look into it.

rajyan:
I think the composer failure is caused because of this `composer update`
https://github.com/phpstan/phpstan/blob/c1ea106f9a101643d4e224c0917cef4eb0c1f60c/.github/workflows/integration-tests.yml#L127
The CI of composer is failing right now while migrating to PHPstan 1.9.0
https://github.com/composer/composer

I'm not sure why the composer integration test is running `composer update`.

tests/PHPStan/Rules/Functions/data/array-cast-list-types.php
```diff
@@ -0,0 +1,25 @@
...
+
+foo((array) true);
+foo((array) 'literal');
+foo((array) 1.0);
+foo((array) 1);
+foo((array) $resource);
+foo((array) (fn () => 'closure'));
+foo((array) $nonEmptyString);
+foo((array) $nonFalsyString);
+foo((array) $numericString);
```
staabm: The test code should be wrapped into a method or function (no top level code, when possible as it would get executed when running the testsuite)

ondrejmirtes: It's better to define parameter types as `@param` rather than `@var`.

rajyan:
fixed the tests and moved to CallMethodsRule
Playground result to show that the test is failing without this PR.
https://phpstan.org/r/e7dcb950-2cdc-4a07-9d4d-e89b58856b48

rajyan:
I was taking literal-string wrong.
Fixed the test and the correct playground result is this
https://phpstan.org/r/75e6f175-55e5-42aa-9cfc-64343c705af8

ondrejmirtes:
Thank you!

PR #1962, opened 2022-11-04
Update DateTimeZone::listAbbreviations signature

franmomu:
Apparently, based on https://www.php.net/manual/en/datetimezone.listabbreviations.php it cannot return `false`


ondrejmirtes (changes requested):
Would be nice to sync that with `timezone_abbreviations_list`

franmomu:
> Would be nice to sync that with `timezone_abbreviations_list`

yeah sure

ondrejmirtes:
Thank you.

PR #1963, opened 2022-11-04
`strtr` returns `non-empty-string`

staabm:


tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -9060,7 +9060,7 @@ public function dataGeneralizeScope(): array
 	{
 		return [
 			[
-				'array<non-empty-array<int|string, array{saveCount: int<0, max>, removeCount: int<0, max>, loadCount: int<0, max>, hitCount: int<0, max>}>>',
+				'array<string, non-empty-array<int|string, array{saveCount: int<0, max>, removeCount: int<0, max>, loadCount: int<0, max>, hitCount: int<0, max>}>>',
```
staabm, line 9063: tbh, I don't know why this test got more precise with this change :)

tests/PHPStan/Analyser/data/strtr.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Strtr;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @param non-empty-string $nonEmptyString
+ * @param non-falsy-string $nonFalseyString
+ */
+function doFoo(string $s, $nonEmptyString, $nonFalseyString) {
```
staabm, line 11: results before this PR:
https://phpstan.org/r/8833f01d-4202-4163-bf72-982f9729c3c9

src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
```diff
@@ -53,7 +55,11 @@ public function getTypeFromFunctionCall(
 	{
 		$type = $this->getPreliminarilyResolvedTypeFromFunctionCall($functionReflection, $functionCall, $scope);
 
-		$possibleTypes = ParametersAcceptorSelector::selectSingle($functionReflection->getVariants())->getReturnType();
+		$possibleTypes = ParametersAcceptorSelector::selectFromArgs(
```
staabm, line 58: since `strtr` has multiple signatures, I had to use `selectFromArgs()` here.

ondrejmirtes (changes requested):
Looking at the extension, could we replace it with stubs with conditional return types instead?

staabm:
> could we replace it with stubs with conditional return types instead?

I think we could, but it would be a rather complicated conditional, which needs to cover `non-empty-string`, `non-falsey-string` and `array` types, and even combine those logical across parameters `subject` and `replace`.

for `strtr` we would also need to incorporate the 2 function variants.. I don't feel this would lead to a readable stub-file.

src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
ondrejmirtes: Because preg_replace is impacted by your changes. By using selectFromArgs, conditional return type is resolved.

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
ondrejmirtes: Because preg_replace is impacted by your changes. By using selectFromArgs, conditional return type is resolved.

ondrejmirtes:
Thank you.

PR #1966, opened 2022-11-04
Error on unknown parameter name in parameter conditional type

rvanvelzen:
Fix phpstan/phpstan#8284

### DIFF
--- a/tests/PHPStan/Rules/PhpDoc/MethodConditionalReturnTypeRuleTest.php
+++ b/tests/PHPStan/Rules/PhpDoc/MethodConditionalReturnTypeRuleTest.php
@@ -70,4 +70,14 @@ class MethodConditionalReturnTypeRuleTest extends RuleTestCase
 		]);
 	}
 
+	public function testBug8284(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-8284.php'], [
+			[
+				'Conditional return type references unknown parameter $callable.',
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/PhpDoc/MethodConditionalReturnTypeRuleTest.php
```diff
@@ -70,4 +70,14 @@ public function testRule(): void
 		]);
 	}
 
+	public function testBug8284(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-8284.php'], [
+			[
+				'Conditional return type references unknown parameter $callable.',
```
mad-briller, line 77: there is no `@return` tag on the `replaceCallback()` method so is this error message a bit misleading?
it looks like it's really a conditional parameter type?

ondrejmirtes: The feature is actually "conditonal types", not "conditional return types" - it was implemented as a first thing after we released "conditional return types".

ondrejmirtes: But for the sake of backward compatibility, let's leave it like this.

ondrejmirtes:
Thank you.

PR #1967, opened 2022-11-04
AccessoryArrayListType::intersectWith() compatibility promise

enumag:
See https://github.com/phpstan/phpstan-symfony/pull/309

I didn't do similar PR before... is it correct to add the `@api` annotation to both the class and the method?

Also do I need to update some test?

### DIFF
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -22,6 +22,7 @@ use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
 
+/** @api */
 class AccessoryArrayListType implements CompoundType, AccessoryType
 {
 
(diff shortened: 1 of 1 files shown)


src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -332,6 +333,7 @@ public static function isListTypeEnabled(): bool
 		return self::$enabled;
 	}
 
+	/** @api */
```
ondrejmirtes, line 336: This doesn't need to be here.

ondrejmirtes (changes requested):

src/Type/Accessory/AccessoryArrayListType.php
ondrejmirtes: I solved it: https://github.com/phpstan/phpstan-src/commit/071320dab361ef35597a524cd472eff3e3531490

PR #1968, opened 2022-11-04
Fix array and iterable type action methods on `NeverType`

herndlm:
What does he mean with that cryptic title again?
Basically, methods that work on the type / modify it instead of only describing its characteristics should not magically turn `*NEVER*` into `mixed`. Things like `popArray()` or `shiftArray()` for example.

And to be on the safe side and avoid similar future failures here I removed usage of the two relevant traits completely.

Closes  https://github.com/phpstan/phpstan/issues/8277


ondrejmirtes:
Nice, thank you!

PR #1969, opened 2022-11-05
support 'HasMethodType' on StaticCall

staabm:
implements https://github.com/phpstan/phpstan/issues/8297
closes https://github.com/phpstan/phpstan/issues/1267

### DIFF
--- a/src/Type/Php/MethodExistsTypeSpecifyingExtension.php
+++ b/src/Type/Php/MethodExistsTypeSpecifyingExtension.php
@@ -47,16 +47,28 @@ class MethodExistsTypeSpecifyingExtension implements FunctionTypeSpecifyingExten
 		TypeSpecifierContext $context,
 	): SpecifiedTypes
 	{
-		$objectType = $scope->getType($node->getArgs()[0]->value);
-		if (!$objectType instanceof ObjectType) {
-			if ($objectType->isString()->yes()) {
-				return new SpecifiedTypes([], []);
-			}
-		}
-
 		$methodNameType = $scope->getType($node->getArgs()[1]->value);
 		if (!$methodNameType instanceof ConstantStringType) {
 			return new SpecifiedTypes([], []);
+		}
+
+		$objectType = $scope->getType($node->getArgs()[0]->value);
+		if (!$objectType instanceof ObjectType) {
+			if ($objectType instanceof ConstantStringType && $objectType->isClassString()) {
+				return $this->typeSpecifier->create(
+					$node->getArgs()[0]->value,
+					new IntersectionType([
+						$objectType,
+						new HasMethodType($methodNameType->getValue()),
+					]),
+					$context,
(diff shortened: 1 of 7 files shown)


tests/PHPStan/Rules/Methods/data/static-has-method.php
```diff
@@ -0,0 +1,18 @@
...
+
+class HelloWorld
+{
+	public function sayHello(): void
+	{
+		if (!method_exists(rex_var::class, 'varsIterator')) {
+			return;
+		}
+
+		$it = rex_var::varsIterator();
```
staabm: the `NodeScopeResolverTest` seems to work as expected with this PR.

what is missing is a proper fix in `StaticMethodCallCheck` so no error is emitted for this line.
was not yet able to figure out how this can work

src/Type/Php/MethodExistsTypeSpecifyingExtension.php
```diff
@@ -47,15 +46,26 @@ public function specifyTypes(
...
+		if (!$methodNameType instanceof ConstantStringType) {
+			return new SpecifiedTypes([], []);
+		}
+
 		$objectType = $scope->getType($node->getArgs()[0]->value);
-		if (!$objectType instanceof ObjectType) {
-			if ($objectType->isString()->yes()) {
-				return new SpecifiedTypes([], []);
+		if ($objectType->isString()->yes()) {
+			if ($objectType instanceof ConstantStringType && $objectType->isClassString()) {
```
staabm, line 56: Should we add isClassString on Type as a followup?

ondrejmirtes: Of course :)

ondrejmirtes:
Yeah, I guess this is usable. Thank you :)

ondrejmirtes (changes requested):
We should also test a scenario when a method doesn't exist - for example when we're not in method_exists at all, or when we call a different method than what was asked in method_exists.

ondrejmirtes:
Thank you!

src/Type/Php/MethodExistsTypeSpecifyingExtension.php
staabm: we already have `ConstantStringType->isClassString(): bool`, 

https://github.com/phpstan/phpstan-src/blob/47ba52bd68e13b4242cdace06ea4df7e20159f30/src/Type/Constant/ConstantStringType.php#L75-L84

therefore we need a different name for this new method... any suggestions?

ondrejmirtes: isClassStringType

ondrejmirtes: And make it nicer for 2.0

PR #1970, opened 2022-11-05
implement isClassStringType() on Type

staabm:


staabm:
in [11ca353](https://github.com/phpstan/phpstan-src/pull/1970/commits/11ca35386c6bed7f14a579af5c8d1c125920570b) I added a test from https://github.com/phpstan/phpstan/issues/1267#issuecomment-1107708169 which was not covered by https://github.com/phpstan/phpstan-src/pull/1969

phpstan-baseline.neon
```diff
@@ -392,6 +392,14 @@ parameters:
 			count: 1
 			path: src/Testing/PHPStanTestCase.php
 
+		-
```
staabm: I was able to drop all uses of `isClassString` and `instanceof ClassStringType` with [48f8635](https://github.com/phpstan/phpstan-src/pull/1970/commits/48f863561c9dcfa359a51ba8ee395970eab8a050) but these 2 occurences. These seemed harder to solve, therefore I put them into the baseline.

herndlm: I might be missing something, but it looks like the internals of `isClassString()` should be moved into `isClassStringType()` and wrapped in a trinary. At least you don't have to use the deprecated method in the new one then. You could even turn it around and let the deprecated one use the new one internally?

Not sure what the problem with the other usage was 🤔

staabm: after a couple of hours sleep I was able to resolve the remining stuff :)

src/Type/MixedType.php
```diff
@@ -676,6 +676,18 @@ public function isLiteralString(): TrinaryLogic
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isClassStringType(): TrinaryLogic
+	{
+		if ($this->subtractedType !== null) {
+			$stringType = new StringType();
+			if ($this->subtractedType->isSuperTypeOf($stringType)->yes()) {
```
ondrejmirtes: Alright, so `mixed~string` is definitely no. `mixed~class-string` should also be no.

ondrejmirtes (changes requested):

src/Type/ClassStringType.php
```diff
@@ -44,7 +44,7 @@ public function accepts(Type $type, bool $strictTypes): TrinaryLogic
 	public function isSuperTypeOf(Type $type): TrinaryLogic
 	{
 		if ($type instanceof ConstantStringType) {
-			return TrinaryLogic::createFromBoolean($type->isClassString());
+			return TrinaryLogic::createFromBoolean($type->isClassStringType()->yes());
```
herndlm: You can simplify this and return the original trinary here and above maybe?

staabm: I thought so too, but I think the alternative would be `$type->isClassStringType()->and(Trinary::createYes())` which is not that readable either. just returning doesn't work, because of `Maybe`.

src/Type/ClassStringType.php
```diff
@@ -27,7 +27,7 @@ public function accepts(Type $type, bool $strictTypes): TrinaryLogic
 		}
 
 		if ($type instanceof ConstantStringType) {
-			return TrinaryLogic::createFromBoolean($type->isClassString());
+			return TrinaryLogic::createFromBoolean($type->isClassStringType()->yes());
```
staabm: I made a 1:1 translation of pre-existing logic. didn't realize that returning the trinary from `isClassStringType` would be even more precise.

fixed

src/Type/ClassStringType.php
staabm: I was wrong, see https://github.com/phpstan/phpstan-src/pull/1970/files#r1014798873 ;)

src/Type/MixedType.php
staabm: fixed, and tests added.

src/Type/CallableType.php
```diff
@@ -347,6 +347,11 @@ public function isLiteralString(): TrinaryLogic
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isClassStringType(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
```
ondrejmirtes: maybe https://3v4l.org/6rfY6

src/Type/ClassStringType.php
ondrejmirtes: I don't get it, why degrade maybe to no?

You can do just `return $type->isClassStringType();` and it's gonna be even better.

src/Type/Constant/ConstantStringType.php
```diff
@@ -445,7 +446,7 @@ public function getGreaterOrEqualType(): Type
 
 	public function canAccessConstants(): TrinaryLogic
 	{
-		return TrinaryLogic::createFromBoolean($this->isClassString());
+		return TrinaryLogic::createFromBoolean($this->isClassStringType()->yes());
```
ondrejmirtes: return `$this->isClassStringType();`

ondrejmirtes:
BTW :) https://github.com/phpstan/phpstan-src/pull/1970/commits/b78ff975bc61ab47c75df4ccb5e8b16cff27dd60 Similarly it should be possible to simplify complicated GenericClassStringType::accepts() and isSuperTypeOf(), but that's gonna need more methods on Type.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #1971, opened 2022-11-06
implement isInteger() on Type

staabm:


staabm:
Memo to me: adjust per loose comparison rules

ondrejmirtes:
> Memo to me: adjust per loose comparison rules

No. Only an integer is an integer.

ondrejmirtes:
It's up to `Type::looseCompare()` to tell that `int == string` might be true by asking `$otherType->isInteger()->or($otherType->isString())` in `IntegerType::looseCompare()`.

ondrejmirtes:
Thank you.

herndlm:
👍 I wanted to do the same later today :)

staabm:
We also need it for bool and float :-)

staabm:
I am working on bool now

herndlm:
Go ahead, I won't be working on something today any more.

staabm:
working on isFloat now

PR #1973, opened 2022-11-06
implement isBoolean() on Type

staabm:


ondrejmirtes:
Would be great to do isTrue and isFalse at the same time to get rid of instanceof ConstantBooleanType. Thanks!

ondrejmirtes:
I love it, thank you!

PR #1978, opened 2022-11-07
implement isNull() on Type

staabm:

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -875,7 +875,7 @@ class MutatingScope implements Scope
 					$node->left instanceof Node\Expr\PropertyFetch
 					|| $node->left instanceof Node\Expr\StaticPropertyFetch
 				)
-				&& $rightType instanceof NullType
+				&& $rightType->isNull()->yes()
 				&& !$this->hasPropertyNativeType($node->left)
 			) {
 				return new BooleanType();
@@ -886,7 +886,7 @@ class MutatingScope implements Scope
 					$node->right instanceof Node\Expr\PropertyFetch
 					|| $node->right instanceof Node\Expr\StaticPropertyFetch
 				)
-				&& $leftType instanceof NullType
+				&& $leftType->isNull()->yes()
 				&& !$this->hasPropertyNativeType($node->right)
 			) {
 				return new BooleanType();
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2167,7 +2167,7 @@ class NodeScopeResolver
 							$thisType = null;
 							if (isset($expr->getArgs()[1])) {
 								$argType = $scope->getType($expr->getArgs()[1]->value);
-								if ($argType instanceof NullType) {
+								if ($argType->isNull()->yes()) {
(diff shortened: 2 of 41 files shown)


staabm:
> BTW I have a small thing to ask - please use first letter upper-case in your PR titles and commit messages.

starting with my next PR, I will take this into account.

VincentLanglet:
Small thought for ConstantInteger, ConstantString, ConstantBoolean, ...

Rather than methods `isConstantInteger` on so on, did you think/do you plan to create `isConstant` since you'll already have `isInteger`, `isBoolean`, `isString`, ... , :)

staabm:
> Rather than methods `isConstantInteger` on so on, did you think/do you plan to create `isConstant` since you'll already have `isInteger`, `isBoolean`, `isString`, ... , :)

these types can already be checked with the `ConstantScalarType` interface, so I think there is currently no need to have them on `Type` again

ondrejmirtes:
@VincentLanglet I think a method like `Type::getConstantStrings(): list<ConstantStringType>` makes the most sense.

But integers complicate it a bit. I don't know how `getConstantIntegers` should be implemented in IntegerRangeType, I don't want huge arrays. Maybe we'll have to treat integers in a different way.

VincentLanglet:
> > Rather than methods `isConstantInteger` on so on, did you think/do you plan to create `isConstant` since you'll already have `isInteger`, `isBoolean`, `isString`, ... , :)
> 
> these types can already be checked with the `ConstantScalarType` interface, so I think there is currently no need to have them on `Type` again

I thought the purpose of all those methods were to get rid of all the `instanceof` checks.
So how do you get rid of the `instanceof` check in (one of you current diff)
```
($returnType->isNull()->yes() || $returnType instanceof ConstantBooleanType)
```
?

staabm:
ondrej has a better idea of it, I am pretty sure.

> I thought the purpose of all those methods were to get rid of all the `instanceof` checks.

I think we are getting rid of `instanceof X` while X is a concrete type class. I am not sure we need to get rid of it when `X` is an interface.

ondrejmirtes:
@VincentLanglet There's already `Type::isTrue(): TrinaryLogic` and `Type::isFalse(): TrinaryLogic`.

@staabm We want to get rid of `instanceof Constant*Type` too because very often people just do `$type instanceof ConstantStringType` and it doesn't occur to them they should handle `'foo'|'bar'` as well.

ondrejmirtes:
Thank you!

herndlm:
good, then there's still work left for me 😅

ondrejmirtes:
@herndlm There's ton of work to be done, I'm looking forward to get rid of `instanceof TypeWithClassName` 😊 And there's a lot more.

herndlm:
I know that there's no need to be worried hehe. great that multiple people are now actively cleaning up!

ondrejmirtes:
Also SubtractableType, that's gonna be a lot of fun 😂

mad-briller:
> We want to get rid of instanceof Constant*Type too because very often people just do $type instanceof ConstantStringType and it doesn't occur to them they should handle 'foo'|'bar' as well.

i ran into this exact case last week in a custom rule :sweat_smile:

PR #1979, opened 2022-11-07
Implement getConstantStrings() on Type

staabm:
refs https://github.com/phpstan/phpstan/issues/8311

### DIFF
--- a/src/Type/TypeUtils.php
+++ b/src/Type/TypeUtils.php
@@ -93,6 +93,8 @@ class TypeUtils
 
 	/**
 	 * @return ConstantStringType[]
+	 *
+	 * @deprecated Use PHPStan\Type\Type->getConstantStrings() instead
 	 */
 	public static function getConstantStrings(Type $type): array
 	{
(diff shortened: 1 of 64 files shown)


ondrejmirtes:
Thank you!

src/Type/TypeUtils.php
```diff
@@ -93,6 +93,8 @@ public static function getConstantArrays(Type $type): array
 
 	/**
 	 * @return ConstantStringType[]
+	 *
+	 * @deprecated Use PHPStan\Type\Type::getConstantStrings() instead
```
schlndh, line 97: @staabm The two methods don't work exactly the same (i.e. the new implementation for `UnionType` doesn't have `$stopOnUnmatched` equivalent). I think it causes a regression here: https://phpstan.org/r/33df1d6b-50e1-4aed-ac3b-bb5f1a8ac66b (constant string concat with `int|null` - the result shouldn't be a constant string).

staabm: thx for the headsup. I will have a look.

I opened a issue for it: https://github.com/phpstan/phpstan/issues/8568

PR #1980, opened 2022-11-07
Making `json_encode()` always produce a `non-empty-string`, when successful

Slamdunk:
See https://3v4l.org/E2P9K and https://github.com/vimeo/psalm/pull/8681


ondrejmirtes:
Thank you!

PR #1981, opened 2022-11-07
`sodium_crypto_generichash*` always produce a `non-empty-string`

Slamdunk:
`sodium_crypto_generichash` functions always return states or hashes as `non-empty-string` by design, see

1. https://doc.libsodium.org/hashing/generic_hashing
2. https://github.com/jedisct1/libsodium/blob/1.0.18/src/libsodium/crypto_generichash/blake2b/ref/generichash_blake2b.c


resources/functionMap.php
```diff
@@ -10449,11 +10449,11 @@
 'sodium_crypto_box_seal_open' => ['string|false', 'message'=>'string', 'recipient_keypair'=>'string'],
 'sodium_crypto_box_secretkey' => ['string', 'keypair'=>'string'],
 'sodium_crypto_box_seed_keypair' => ['string', 'seed'=>'string'],
-'sodium_crypto_generichash' => ['string', 'msg'=>'string', 'key='=>'?string', 'length='=>'?int'],
-'sodium_crypto_generichash_final' => ['string', 'state'=>'string', 'length='=>'?int'],
-'sodium_crypto_generichash_init' => ['string', 'key='=>'?string', 'length='=>'?int'],
-'sodium_crypto_generichash_keygen' => ['string'],
-'sodium_crypto_generichash_update' => ['bool', 'state'=>'string', 'string'=>'string'],
+'sodium_crypto_generichash' => ['non-empty-string', 'msg'=>'string', 'key='=>'?string', 'length='=>'?int'],
+'sodium_crypto_generichash_final' => ['non-empty-string', 'state'=>'non-empty-string', 'length='=>'?int'],
```
ondrejmirtes, line 10453: Declaring parameter as `non-empty-string` gives me a little pause - what happens when the function receives an empty string? Unless it crashes, I wouldn't require it here.

ondrejmirtes (changes requested):

resources/functionMap.php
Slamdunk: That's indeed what happens when you pass an empty string variable:

```
PHP Fatal error:  Uncaught SodiumException: incorrect state length
```

`libsodium` is very reliable

ondrejmirtes:
Thank you.

PR #1983, opened 2022-11-07
Fix issue 8292

rajyan:
fixes https://github.com/phpstan/phpstan/issues/8292

All static expression types should be invalidated in anonymous/arrow functions, because the type is not reliable in delayed calls (skipped ClassConstFetch because I think it cannot change).
Also, this fixes the inconsistency between anonymous and arrow functions, which rvanvelzen mentioned in
https://github.com/phpstan/phpstan-src/pull/1929#issuecomment-1294553610
static expressions should be invalidate for arrows as well.

https://github.com/phpstan/phpstan-src/pull/1929#issuecomment-1295885526

I have a feeling that maybe we can get rid of `expressionTypeIsUnchangeable` if example like
https://github.com/phpstan/phpstan-src/pull/1934#issuecomment-1295766300
is not possible without using static expressions, but not sure enough.

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3058,7 +3058,7 @@ class MutatingScope implements Scope
 			$this->isDeclareStrictTypes(),
 			$arrowFunctionScope->getFunction(),
 			$arrowFunctionScope->getNamespace(),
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Thank you!

PR #1985, opened 2022-11-08
`sodium_crypto_sign*` always produce `non-empty-string`

Slamdunk:
By design `libsodium` never produce or allow consumption of empty keys or an empty signature.

That's true for basically all `sodium_*` functions, but I prefer to propose here PRs only for what I know and I have first hand experience on.


ondrejmirtes:
Thank you.

PR #1986, opened 2022-11-08
Extend Xdebug note

curry684:
If a problem is emitted proper DX says you should also suggest a solution.

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -96,7 +96,7 @@ class CommandHelper
 		if ($allowXdebug && !XdebugHandler::isXdebugActive()) {
 			$errorOutput->getStyle()->note('You are running with "--xdebug" enabled, but the Xdebug PHP extension is not active. The process will not halt at breakpoints.');
 		} elseif (!$allowXdebug && XdebugHandler::isXdebugActive()) {
-			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. This may slow down performance and the process will not halt at breakpoints.');
+			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. This may slow down performance and the process will not halt at breakpoints. Run with XDEBUG_MODE=off to disable Xdebug and remove this message.');
 		}
 
 		if (!$allowXdebug) {


staabm (approved):
I like.

drupol:
I like it too !

There's also this that might be an option : https://packagist.org/packages/composer/xdebug-handler

staabm:
> There's also this that might be an option

its already used in phpstan, see https://github.com/phpstan/phpstan-src/blob/582a9cb8b9b4fce2bd069bac26bf1d31dd52e7e2/composer.json#L12

src/Command/CommandHelper.php
```diff
@@ -96,7 +96,7 @@ public static function begin(
 		if ($allowXdebug && !XdebugHandler::isXdebugActive()) {
 			$errorOutput->getStyle()->note('You are running with "--xdebug" enabled, but the Xdebug PHP extension is not active. The process will not halt at breakpoints.');
 		} elseif (!$allowXdebug && XdebugHandler::isXdebugActive()) {
-			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. This may slow down performance and the process will not halt at breakpoints.');
+			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. This may slow down performance and the process will not halt at breakpoints. Run with XDEBUG_MODE=off to disable Xdebug and remove this message.');
```
ondrejmirtes, line 99: I'd argue that this message isn't still clear enough. What does "Run with XDEBUG_MODE=off" mean?

ondrejmirtes (changes requested):

src/Command/CommandHelper.php
curry684: I agree that it's not totally clear, but phpstan is a *developer* tool, and to *developers* this is at least a firm push in the right direction. A better explanation would take at least 4 lines, likely more, and at that point I think you're crossing the line it should be explained *thoroughly* at https://phpstan.org/user-guide/command-line-usage, and the warning itself should just be as it is now.

If the verbose explanation is in the docs people who need it can and will google "phpstan xdebug_mode=off" and find it right away.

curry684:
If https://github.com/phpstan/phpstan/pull/8343 is merged we could link to https://phpstan.org/user-guide/command-line-usage#--xdebug in the note.

(you may not want to add documentation permalinks to the sources, in that case the google argument still holds)

ondrejmirtes:
I'm not really keen on this. It's not the only solution you can do, and the environment variable was introduced only pretty recently with Xdebug 3... Sorry, I decided no to merge this.

PR #1987, opened 2022-11-08
Fix unset on nested array

rvanvelzen:
Fix phpstan/phpstan#8113


tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -1126,6 +1126,7 @@ public function dataFileAsserts(): iterable
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/strtr.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/static-has-method.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/mixed-to-number.php');
+		yield from $this->gatherAssertTypes(__DIR__ . '/../Rules/Variables/data/bug-8113.php');
```
ondrejmirtes, line 1129: You've added the test file here but there are no `assertType` calls inside?

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/NodeScopeResolverTest.php
rvanvelzen: I accidentally reverted the calls and didn't realize :sweat_smile:

ondrejmirtes:
Thank you!

PR #1988, opened 2022-11-08
Simplify array collections in `UnionTypeHelper`

herndlm:
as requested in https://github.com/phpstan/phpstan-src/pull/1924#discussion_r1016672366


ondrejmirtes:
Thank you!

PR #1989, opened 2022-11-09
Fix false positive of access to static private property of parent class

staabm:
closes https://github.com/phpstan/phpstan/issues/8333

### DIFF
--- a/src/Rules/Properties/AccessStaticPropertiesRule.php
+++ b/src/Rules/Properties/AccessStaticPropertiesRule.php
@@ -23,6 +23,7 @@ use PHPStan\Type\ThisType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeUtils;
+use PHPStan\Type\TypeWithClassName;
 use PHPStan\Type\VerbosityLevel;
 use function array_map;
 use function array_merge;
@@ -182,6 +183,18 @@ class AccessStaticPropertiesRule implements Rule
 		if (!$has->yes()) {
 			if ($scope->hasExpressionType($node)->yes()) {
 				return $messages;
+			}
+
+			if ($typeForDescribe instanceof TypeWithClassName && $typeForDescribe->getClassReflection() !== null) {
+				foreach ($typeForDescribe->getClassReflection()->getParents() as $parentClassReflection) {
+					if (!$parentClassReflection->hasProperty($name)) {
+						continue;
+					}
+
+					if ($scope->canAccessProperty($parentClassReflection->getProperty($name, $scope))) {
+						return [];
+					}
+				}
 			}
 
 			return array_merge($messages, [
--- a/src/Rules/Properties/AccessStaticPropertiesRule.php
+++ b/src/Rules/Properties/AccessStaticPropertiesRule.php
@@ -185,9 +185,9 @@ class AccessStaticPropertiesRule implements Rule
 				return $messages;
 			}
 
-			$classNames = $classType->getReferencedClasses();
+			$classNames = TypeUtils::getDirectClassNames($classType);
 			if (count($classNames) === 1) {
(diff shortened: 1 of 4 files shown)


staabm:
just realized that this rule also missed a separate error case, which the`AccessPropertiesRule` contained in

https://github.com/phpstan/phpstan-src/blob/80ba5a0a570834bc96b81ccc614a9cc84fe54c9a/src/Rules/Properties/AccessPropertiesRule.php#L124-L135

added the missing check and a test

src/Rules/Properties/AccessStaticPropertiesRule.php
```diff
@@ -184,6 +185,26 @@ private function processSingleProperty(Scope $scope, StaticPropertyFetch $node,
 				return $messages;
 			}
 
+			if ($typeForDescribe instanceof TypeWithClassName && $typeForDescribe->getClassReflection() !== null) {
```
ondrejmirtes: 1) typeForDescribe is just for error message description
2) `instanceof *Type` is always wrong

ondrejmirtes (changes requested):

src/Rules/Properties/AccessStaticPropertiesRule.php
staabm: refactored it to make it more like 
https://github.com/phpstan/phpstan-src/blob/582a9cb8b9b4fce2bd069bac26bf1d31dd52e7e2/src/Rules/Properties/AccessPropertiesRule.php#L119-L139

src/Rules/Properties/AccessStaticPropertiesRule.php
```diff
@@ -185,23 +185,27 @@ private function processSingleProperty(Scope $scope, StaticPropertyFetch $node,
 				return $messages;
 			}
 
-			if ($typeForDescribe instanceof TypeWithClassName && $typeForDescribe->getClassReflection() !== null) {
-				foreach ($typeForDescribe->getClassReflection()->getParents() as $parentClassReflection) {
-					if (!$parentClassReflection->hasProperty($name)) {
-						continue;
+			$classNames = $classType->getReferencedClasses();
```
ondrejmirtes: This is also wrong. `getReferencedClasses` will return nested classes too. Which means that if you get `Foo<Bar>`, getReferencedClasses will contain both Foo and Bar.

I think that in context of this PR, you need `TypeUtils::getDirectClassNames`. Feel free to fix the other PR too.

ondrejmirtes (changes requested):

src/Rules/Properties/AccessStaticPropertiesRule.php
staabm: at this point I just did what you said. I don't know what I am doing here right now.

> Feel free to fix the other PR too.

not sure which one you mean.

I will be in a meeting for the rest of the day. would be great in case you already know what needs to be done, if you could finish it.

thank you

ondrejmirtes: You said: refactored it to make it more like 
https://github.com/phpstan/phpstan-src/blob/582a9cb8b9b4fce2bd069bac26bf1d31dd52e7e2/src/Rules/Properties/AccessPropertiesRule.php#L119-L139

But that other rule is wrong too, not sure for how long.

staabm: ok. I pushed a fix. hopefully thats what you had in mind.

thank you

staabm: > But that other rule is wrong too, not sure for how long.

btw: I found the commit https://github.com/staabm/phpstan-src/commit/455fb7b890614c7f402fc1d87f7b2c26aa18424b

ondrejmirtes:
Thank you!

PR #1990, opened 2022-11-10
Passing a variable by reference to a function and method has side effects

ryium:
- fixes https://github.com/phpstan/phpstan/issues/8084

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-8084.php
+++ b/tests/PHPStan/Analyser/data/bug-8084.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Bug8084;
+
+use function array_shift;
+use function PHPStan\Testing\assertType;
+
+class Bug8084
+{
+	public function run(array $arr): void
+	{
+		/** @var array{a: 0, b?: 1} $arr */
+		assertType('0', array_shift($arr) ?? throw new \Exception());
+		assertType('1|null', array_shift($arr));
+	}
+}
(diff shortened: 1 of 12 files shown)


ondrejmirtes (changes requested):
Could you please test and fix other similar functions that might have the same issue? array_push, array_unshift, array_pop, array_splice

ondrejmirtes:
And maybe it'd be a better to idea to do this in `public function hasSideEffects(): TrinaryLogic` in various FunctionReflection and MethodReflection implementations - to return `TrinaryLogic::createYes()` if the function accepts parameters as references.

tests/PHPStan/Analyser/data/bug-8084.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Bug8084;
+
+use function array_shift;
+use function PHPStan\Testing\assertType;
+
+class Bug8084
+{
+	public function run(array $arr): void
+	{
+		/** @var array{a: 0, b?: 1} $arr */
```
ondrejmirtes: Hi, please put this type into `@param` above the method, here it's ambiguous what it means.

Also currently this example https://phpstan.org/r/de642079-e5b0-492a-9782-d78c324f70a7 shows `Expression on left side of ?? is not nullable.` so we'll need another test in NullCoalesceRuleTest to make sure this message goes away.

Thank you!

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/bug-8084.php
ryium: Hi, thank you for your review.
Added a test, but included a merge commit to follow the latest branch.  
Should I recreate?

ondrejmirtes (approved):

ondrejmirtes:
Brought it over the finish line :) Thank you.

ondrejmirtes:
Sorry, I had to revert this, it led to some false positives.

PR #1992, opened 2022-11-11
Fix non-working type assertions in `NodeScopeResolverTest` cases

herndlm:
Discovered via https://github.com/phpstan/phpstan-src/pull/1991


herndlm:
because I saw it here - most likely running the issue bot for `tests/`-only changes is overkill?

staabm:
Maybe we should have a rule which detects when TypeInferenceTest included files don‘t contain asserts (or doesnt import the function)

ondrejmirtes:
Thank you!

ondrejmirtes:
> Maybe we should have a rule

Done https://github.com/phpstan/phpstan-src/commit/d0207ed79e94e4b4daceb0f83abb1f481ee3250b

PR #1994, opened 2022-11-13
parse_str() might return arrays by-ref

staabm:
closes https://github.com/phpstan/phpstan/issues/8356

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -70,7 +70,7 @@ function bin2hex(string $string): string {}
 
 /**
  * @param array<mixed> $result
- * @param-out array<int|string, string|list<string>|array<mixed>> $result
+ * @param-out array<int|string, array|string> $result
  */
 function parse_str(string $string, array &$result): void {}
 
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -76,7 +76,7 @@ function parse_str(string $string, array &$result): void {}
 
 /**
  * @param array<mixed> $result
- * @param-out array<string, string|list<string>|array<mixed>> $result
+ * @param-out array<string, array|string> $result
  */
 function mb_parse_str(string $string, array &$result): bool {}
 
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -70,9 +70,15 @@ function bin2hex(string $string): string {}
 
 /**
  * @param array<mixed> $result
- * @param-out array<int|string, string|list<string>> $result
+ * @param-out array<int|string, string|list<string>|array<mixed>> $result
  */
 function parse_str(string $string, array &$result): void {}
+
+/**
+ * @param array<mixed> $result
+ * @param-out array<string, string|list<string>|array<mixed>> $result
+ */
+function mb_parse_str(string $string, array &$result): bool {}
 
 /** @param-out float $percent */
 function similar_text(string $string1, string $string2, float &$percent = null) : int {}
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -700,4 +700,15 @@ public function testBug6243(): void
 		$this->analyse([__DIR__ . '/data/bug-6243.php'], []);
 	}
 
+	public function testBug8356(): void
+	{
+		$this->bleedingEdge = true;
+		$this->analyse([__DIR__ . '/data/bug-8356.php'], [
+			[
+				"Offset 'x' might not exist on array|string.",
```
staabm, line 708: before this PR, the rule reported 
`Offset 'x' does not exist on array|string.`
.. now we get  ..
`Offset 'x' might not exist on array|string.`

when running this test without `$this->bleedingEdge = true;` the signature changes in the stub do not have any effect and we still get `Offset 'x' does not exist on array|string.`

staabm (commented):
note: the issue-bot did not realize a change in the origin issue snippet, even though the testsuite shows that the PR will change it

stubs/core.stub
```diff
@@ -70,10 +70,16 @@ function bin2hex(string $string): string {}
 
 /**
  * @param array<mixed> $result
- * @param-out array<int|string, string|list<string>> $result
+ * @param-out array<int|string, string|list<string>|array<mixed>> $result
  */
 function parse_str(string $string, array &$result): void {}
```
staabm, line 75: moved `parse_str` and `mb_parse_str` next to each other, and equalized the type signatures

stubs/core.stub
```diff
@@ -70,10 +70,16 @@ function bin2hex(string $string): string {}
 
 /**
  * @param array<mixed> $result
- * @param-out array<int|string, string|list<string>> $result
+ * @param-out array<int|string, string|list<string>|array<mixed>> $result
```
staabm: `int` keys are only possible in php7 when using `parse_str($str);` without a second arg.
[per php manual this is highly discouraged](https://www.php.net/manual/en/function.parse-str.php) and was removed with php8:

> Warning
> Using this function without the result parameter is highly DISCOURAGED and DEPRECATED as of PHP 7.2. As of PHP 8.0.0, the result parameter is mandatory.

`mb_parse_str` in contrast does not at all have this feature.

we could drop the `int` part from the array key to get a more precise type for the sugged use-cases, but would get errors for the discouraged use-case in return. wdyt?

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
ondrejmirtes: The bot only goes through triaged issues labeled as bug or feature-request.

stubs/core.stub
ondrejmirtes: `string|list<string>|array<mixed>` - this doesn't make sense to me, that type is going to be normalized: https://phpstan.org/r/7006b26a-88ec-4766-ab84-bc4c8b93160a

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #1999, opened 2022-11-14
Fix expr string cast

rajyan:
I was wondering why we need string casts for exprs at all, because array key string which causes an conversion to integer is a `LNumber` and should be skipped by
```php
$expr instanceof Node\Scalar || $expr instanceof Array_
```

The culprit was LNumber with UnaryMinus:smile:

### DIFF
--- a/src/Analyser/DirectInternalScopeFactory.php
+++ b/src/Analyser/DirectInternalScopeFactory.php
@@ -41,11 +41,11 @@ class DirectInternalScopeFactory implements InternalScopeFactory
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


rajyan:
Wait, found an edge case
https://3v4l.org/bMGnT

Maybe we should make type specifier don't specify scalar types.

rajyan:
On second thought, this was fine, because it wouldn't converted to integer if used as array keys.
https://3v4l.org/aj358

ondrejmirtes:
Thank you!

PR #2001, opened 2022-11-15
Fix polluted types when disabling polluteScopeWithLoopInitialAssignments

jlherren:
Fixes phpstan/phpstan#8358

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -113,6 +113,7 @@ use PHPStan\Type\VerbosityLevel;
 use PHPStan\Type\VoidType;
 use Throwable;
 use function abs;
+use function array_diff;
 use function array_key_exists;
 use function array_key_first;
 use function array_keys;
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
Hi, I've been sitting on this one and thinking about it for a while. The problem is that you want `polluteScopeWithAlwaysIterableForeach` for a for-loop. And there isn't a config option for that.

If you look in NodeScopeResolver how `polluteScopeWithAlwaysIterableForeach` is used for foreach, you should be able to implement a new option called `polluteScopeWithAlwaysIterableFor` that works for a for-loop the same way.

But a method like `unpolluteVariables` is not the correct approach. I can't pinpoint to why, I'm just seeing it from experience.

jlherren:
That makes sense. My main concern was the merging of the scopes before/after the loop, which causes types to just be wrong, when compared with php-src. I believe the only reason this is done is to reduce the certainty of variables, which in turn triggers errors about 'Maybe' variables being used. But it also messes with the types.

I might be opinionated on this, but maybe what I really want is this:

```
function foo(): void {
    $var = null;
    for ($i = 0; $i < 10; $i++) {
        $var = 1;
    }
    // <--
}
```

At the arrow, I want `$var` to have certainty `Yes` and type `'1'`, because the loop is always iterable. But I want `$i` to have certainty `Maybe`, for the sole purpose of triggering errors about using a loop variable after the loop.

So in the end, what I really want is maybe an entirely different feature... Something like "Access to variable $i, which was defined in a nested scope". PHPStan is currently abusing the certainty of variables to produce this kind of errors.

ondrejmirtes:
You want something weird and in-between 😊 Either the loop didn't run and the certainty is Maybe with a combined type, or the loop ran and the certainty is Yes with a type coming from inside the loop 😊

I don't see something in-between as really valuable and easily doable.

jlherren:
Yeah, I figured what I want is weird :) But frankly, it's also what I'd technically expect from `polluteScopeWithLoopInitialAssignments = false`, since it's precisely the loop initial assignment that I don't want to leak into the scope after the loop.

All in all, your suggestion is probably the best.

PR #2003, opened 2022-11-15
Add `TypeCombinator::removeFalsey()`

axlon:
This PR adds the ability for the type combinator to remove falsey types from a given type. I am in need of this functionality within an extension, but the static type factory is currently not part of the public API. Rather than changing that, this seemed like a better way to expose the functionality.

PHPStan also contains a few places where it removes falsey types, so I replaced those occurrences.

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -99,7 +99,6 @@ use PHPStan\Type\ObjectType;
 use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\ParserNodeTypeToPHPStanType;
 use PHPStan\Type\StaticType;
-use PHPStan\Type\StaticTypeFactory;
 use PHPStan\Type\StringType;
 use PHPStan\Type\ThisType;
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Hi, what's your extension and why do you have the need to remove all falsey types in it? :)

axlon:
> Hi, what's your extension and why do you have the need to remove all falsey types in it? :)

Our extension is currently still privated, because its in very early stages of development. But we're working on an extension that aims to make Laravel strictly typed, which could be used alongside or separate of Larastan.

Currently what I'm trying to tackle is Laravel's `Collection::filter()` which works the same as `array_filter()`; when the method is called without any parameters it will remove any falsey values from the collection.

```php
/** @var \Illuminate\Support\Collection<int, string|null> $foo */
$bar = $foo->filter(); // \Illuminate\Support\Collection<int, non-falsey-string>
```

My thinking was that since this is the way the array_filter extension does it, this is probably the best way to remove falsey types from a type.

ondrejmirtes:
Thank you!

PR #1950, opened 2022-11-01
Improve conditionalExpressionTypes

rajyan:
Steps left
- [x] permit all expression types as conditional expression keys
- [x] change the type guard logic by using the specified scope
- [ ] Improve `createConditionalExpressions`
- [x] fix conditional expression invalidations
- [x] use `ExpressionTypeHolders` for conditions
- [ ] change `addConditionalExpressions` first arg to expr

fixes https://github.com/phpstan/phpstan/issues/3677
fixes https://github.com/phpstan/phpstan/issues/5623
fixes https://github.com/phpstan/phpstan/issues/5401 (the playground in the comment is another issue, we should open a new issue)
fixes https://github.com/phpstan/phpstan/issues/7292
fixes https://github.com/phpstan/phpstan/issues/8212

PR #2004, opened 2022-11-16
Do not invalidate related expressions in ensureNonNullability

rajyan:
fixes https://github.com/phpstan/phpstan/issues/8361
As `specifyExpressionType`  and `assignExpression` is separated in 
https://github.com/phpstan/phpstan-src/pull/1950
I want to make it public to use them when appropriate.

In `ensureShallowNonNullability`, I think there are no reasons to invalidate related expressions

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3401,7 +3401,7 @@ class MutatingScope implements Scope
 		);
 	}
 
-	private function specifyExpressionType(Expr $expr, Type $type, Type $nativeType): self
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Perfect, thank you!

herndlm:
Btw @rajyan I'm just curious - how did you know where to fix that? Do you know MutatingScope already so well or do you also have to step-debug like I do often? Or something else?

rajyan:
@herndlm 
I do step debugging too. By debugging, I found that conditionalExpression of `$from` was invalidated unexpectedly after coalesce and could predict the cause from that.

PR #2006, opened 2022-11-17
Include filename when reporting that result cache wasn't saved due to non-ignorable exceptions

5lava:
Before:

```
Result cache was not saved because of non-ignorable exception: Syntax error, unexpected '}' on line 9
```

After:
```
Result cache was not saved because of non-ignorable exception: Syntax error, unexpected '}' on line 9 in /project/path/file.php
```


ondrejmirtes:
Tha parse error can be in a different file so this would be really confusing :)

PR #2007, opened 2022-11-17
Improve create conditional expressions to handle all expressions

rajyan:
This pull request covers
https://github.com/phpstan/phpstan/issues/8360

*  Solve the infinite loop problem
* Make createConditionalExpressions work for all expressions

improves result of https://github.com/phpstan/phpstan/issues/4173 but not completely fixed yet

### DIFF
--- a/tests/PHPStan/Rules/Variables/data/bug-4173.php
+++ b/tests/PHPStan/Rules/Variables/data/bug-4173.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+namespace Bug4173;
+
+class HelloWorld
+{
+	protected function _sub_render_condition($row)
+	{
+		if (count($row) === 3) {
+			[$field, $cond, $value] = $row;
+		} elseif (count($row) === 2) {
+			[$field, $cond] = $row;
+		} elseif (count($row) === 1) {
+			[$field] = $row;
+		} else {
+			throw new \InvalidArgumentException();
+		}
+
+		$field = trim($field);
+
+		if (count($row) === 1) {
+			return $field;
+		}
+
+		if (count($row) === 2) {
+			$value = $cond; // $cond always set, "count($row) === 1" never true here
+		}
+
+
+		if ($value === null) { // $value always set
+
+		}
+	}
+}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -114,6 +114,7 @@ use PHPStan\Type\VoidType;
 use Throwable;
 use function abs;
 use function array_key_exists;
+use function array_key_first;
 use function array_keys;
 use function array_map;
 use function array_merge;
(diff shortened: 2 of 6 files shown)


rajyan:
Issue bot seems not working well. Looks like a cache problem?
This change should notify issue-7705 and issue-4173 changes like in
https://github.com/phpstan/phpstan-src/pull/1950#issuecomment-1307318676

staabm:
> Issue bot seems not working well. Looks like a cache problem?
> This change should notify issue-7705 and issue-4173 changes like in

hmm I have the impression it works:

https://github.com/phpstan/phpstan-src/actions/runs/3488347704

grafik

rajyan:
Oh, thank you for noticing! Not sure what I was looking at...

rajyan:
@ondrejmirtes 
I think the result of issue bot is wrong when there is a infinite loop. issue-7705 was causing an infinite loop before https://github.com/phpstan/phpstan-src/pull/2007/commits/dd478c0b5b1999856fab94c1486235060eb4261f , but the issue bot was saying no error.

herndlm:
Loops are generalised and skipped after a couple of iterations AFAIK. Could that be related?

rajyan:
Not that "loop" I think.
This change causes something like
Resolve conditionalExpressionType of `$foo` => Need to `getType` of contional `$isFoo || $isBar` => Resolve type of `$isFoo || $isBar`  => `getType` of BooleanOr uses `filterByFalsyScope` => In `filterByFalsyScope` conditionalExpressionType of `$foo` is evaluate again...

herndlm:
Ah sorry, yeah, completely wrong loop I was thinking about.. xD

ondrejmirtes:
Thank you!

tests/PHPStan/Rules/Variables/data/bug-4173.php
```diff
@@ -0,0 +1,34 @@
...
+			throw new \InvalidArgumentException();
+		}
+
+		$field = trim($field);
+
+		if (count($row) === 1) {
+			return $field;
+		}
+
+		if (count($row) === 2) {
```
mvorisek, line 25: This PR seems to cache result of function result 1:1 and retrive the result from cache later. I observed this in phpstan v 1.9.2 (the latest rel).

Is that desired? For `count()` it is legit and thank you. However for expression like `TraitUtil::hasTrait($object, DebugTrait::class)` (a custom function) how does php know/guarantee the result will be the same, does it reason it based on the `TraitUtil::hasTrait` impl or just guess it when the passed args are the same?

rajyan: Yes this is intended

> how does php know/guarantee the result will be the same

When
* TraitUtil::hasTrait is not marked impure
* The exact same `TraitUtil::hasTrait($object, DebugTrait::class)` is called
* $object is not invalidated

> just guess it when the passed args are the same?

so it's something like that.

rajyan: By the way, remembering function results was working before this PR already.
https://phpstan.org/blog/remembering-and-forgetting-returned-values
this PR made them work for more complex conditions too.

mvorisek: Thank you, I did not know user functions are "pure"/remembered by default.

staabm: See https://phpstan.org/blog/remembering-and-forgetting-returned-values

PR #2010, opened 2022-11-18
Specify return type for `filter_input()`

herndlm:
Closes https://github.com/phpstan/phpstan/issues/6261

See https://www.php.net/manual/en/function.filter-input.php, the important difference to `filter_var` is the handling of the non-set variables in the superglobal arrays
> Return Values
> Value of the requested variable on success, false if the filter fails, or null if the var_name variable is not set. If the flag FILTER_NULL_ON_FAILURE is used, it returns false if the variable is not set and null if the filter fails.

### DIFF
--- a/src/Type/Php/FilterInputDynamicReturnTypeExtension.php
+++ b/src/Type/Php/FilterInputDynamicReturnTypeExtension.php
@@ -14,7 +14,6 @@ use PHPStan\Type\NullType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use function count;
-use function strtolower;
 
 class FilterInputDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
 {
--- a/src/Type/Php/FilterInputDynamicReturnTypeExtension.php
+++ b/src/Type/Php/FilterInputDynamicReturnTypeExtension.php
@@ -44,7 +44,7 @@ class FilterInputDynamicReturnTypeExtension implements DynamicFunctionReturnType
 		// See https://github.com/phpstan/phpstan-src/pull/2012 for details
 		$inputType = new ArrayType(new StringType(), new MixedType());
 
-		return $this->filterFunctionReturnTypeHelper->getArrayOffsetValueType(
+		return $this->filterFunctionReturnTypeHelper->getOffsetValueType(
 			$inputType,
 			$scope->getType($functionCall->getArgs()[1]->value),
 			isset($functionCall->getArgs()[2]) ? $scope->getType($functionCall->getArgs()[2]->value) : null,
--- a/src/Type/Php/FilterInputDynamicReturnTypeExtension.php
+++ b/src/Type/Php/FilterInputDynamicReturnTypeExtension.php
@@ -2,29 +2,21 @@
 
 namespace PHPStan\Type\Php;
 
-use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\FuncCall;
-use PhpParser\Node\Expr\Variable;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
-use PHPStan\Type\NullType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\NeverType;
+use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
-use PHPStan\Type\TypeCombinator;
 use function count;
+use function in_array;
 
 class FilterInputDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
 {
-
-	private const INPUT_TYPE_VARIABLE_NAME_MAP = [
-		'INPUT_GET' => '_GET',
-		'INPUT_POST' => '_POST',
-		'INPUT_COOKIE' => '_COOKIE',
-		'INPUT_SERVER' => '_SERVER',
-		'INPUT_ENV' => '_ENV',
-	];
 
 	public function __construct(private FilterFunctionReturnTypeHelper $filterFunctionReturnTypeHelper)
 	{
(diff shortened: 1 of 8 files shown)


herndlm:
Oh, I forgot to let it return false/null when the offset is not existing. Need to change a couple of things for that still..

herndlm:
do you think we can get that this one in the next release? 😊 it's 90% refactor and would help me get rid of some errors in a legacy app which helps me finish the 7.4 -> 8.x upgrade. at least that's what I was telling myself when I was last procrastinating when working on it :D

tests/PHPStan/Analyser/data/filter-input.php
```diff
@@ -0,0 +1,36 @@
...
+use function PHPStan\Testing\assertType;
+
+class FilterInput
+{
+
+	public function invalidTypesOrVarNamesAreIgnored($mixed): void
+	{
+		assertType('mixed', filter_input(INPUT_GET, $mixed, FILTER_VALIDATE_INT));
+		assertType('mixed', filter_input(INPUT_GET, 17, FILTER_VALIDATE_INT));
+		assertType('mixed', filter_input(-1, 'foo', FILTER_VALIDATE_INT));
```
herndlm, line 13: fyi things like that lead to
```
Fatal error: Uncaught ValueError: filter_input(): Argument #1 ($type) must be an INPUT_* constant in /in/47GPB:3
```
which is not caught by phpstan yet. not sure if needed, but we could create a rule for it I suppose

https://3v4l.org/D6eor
https://phpstan.org/r/235a242d-c6b3-4a8f-a746-a766b4a9c687

tests/PHPStan/Rules/Functions/data/bug-6261.php
```diff
@@ -0,0 +1,13 @@
...
+
+function needs_int(int $x) : void {}
+
+$x = filter_input(INPUT_POST, 'row_id', FILTER_VALIDATE_INT);
+
+if($x === false || $x === null) {
+	die("I expected a numeric string!\n");
+}
+
+needs_int($x);
```
staabm: should be moved into a method/function

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+		if ($inputVariable === null || !$varNameType->isString()->yes()) {
+			return null;
+		}
+
+		$inputVariableType = $scope->getType($inputVariable);
+		$inputHasOffsetValue = $inputVariableType->hasOffsetValueType($varNameType);
+		$flagsType = isset($functionCall->getArgs()[3]) ? $scope->getType($functionCall->getArgs()[3]->value) : null;
+		if ($inputHasOffsetValue->no()) {
+			return $this->determineNonExistentOffsetReturnType($flagsType);
+		}
```
herndlm: without superglobal narrowing I can't test this unfortunately :/

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+		'INPUT_ENV' => '_ENV',
+	];
+
+	public function __construct(private FilterFunctionReturnTypeHelper $filterFunctionReturnTypeHelper)
+	{
+	}
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return strtolower($functionReflection->getName()) === 'filter_input';
```
ondrejmirtes: No need for strtolower

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
+	{
+		if (count($functionCall->getArgs()) < 2) {
+			return null;
+		}
+
+		$inputVariable = $this->determineInputVariable($functionCall->getArgs()[0]);
```
ondrejmirtes: What's special about variables and this function? What do you do with the variable? What about other expression types?

ondrejmirtes (changes requested):

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
herndlm: filter_input() only works with superglobal variables like GET and such. But this is an internal thing. From the outside you don't pass an expression but a constant that defines which superglobal input is used. This is why I have this weird mapping in the class that creates the superglobal variable expression. I hope that makes sense.

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+use PhpParser\Node\Expr\Variable;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\NullType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use function count;
+use function strtolower;
```
herndlm:
```suggestion
```

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
herndlm:
```suggestion
		return $functionReflection->getName() === 'filter_input';
```

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
herndlm: Ah and of course my helper thingy expects an expression. So that's basically re-using filter_var() (there you have the var :)) internally for filter_input()

herndlm:
Fyi the return type in case the array offset does not exist and the force array flag is set is not handled properly. It's a bit tricky and I might need to refactor the helper maybe. Good that you didn't merge it yet 😅

tests/PHPStan/Analyser/data/filter-input.php
herndlm: will be handled by returning a `*NEVER*` at least

herndlm:
I'm happy with this now, maybe the new method in the helper is useful for https://github.com/phpstan/phpstan-src/pull/2257 too.

herndlm:
I'm happy with this now, maybe the new method in the helper is useful for https://github.com/phpstan/phpstan-src/pull/2257 too.

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,63 @@
...
+		if (count($functionCall->getArgs()) < 2) {
+			return null;
+		}
+
+		$typeArgExpr = $functionCall->getArgs()[0]->value;
+		$varNameType = $scope->getType($functionCall->getArgs()[1]->value);
+		$varNameTypeIsString = $varNameType->isString();
+		if (
+			$varNameTypeIsString->no()
+			|| !($typeArgExpr instanceof ConstFetch)
```
herndlm: Still doesn't feel like a good way to handle those constants though. Can this be done better? E.g. what if people pass constant ints or a general integer. Like https://3v4l.org/uYgUd

herndlm: decided to simplify some sanity checks. the preciser type is only determined when a constant is passed, but if not it doesn't assume a fatal error anymore, but simply falls back to the default return type. there are too many weird PHP-version-dependent edge cases most likely nobody is interested it. e.g. https://3v4l.org/OH6Nt#veol

it can be still made stricter / preciser any time in the future

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
herndlm: this has become more or less irrelevant I'd say, at least I don't have this mapping anymore since all superglobals are currently assumed to have the same type anyways. but that simplified things

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,55 @@
...
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
+	{
+		if (count($functionCall->getArgs()) < 2) {
+			return null;
+		}
+
+		$typeExpr = $functionCall->getArgs()[0]->value;
+		if (
+			!($typeExpr instanceof ConstFetch)
+			|| !in_array((string) $typeExpr->name, ['INPUT_GET', 'INPUT_POST', 'INPUT_COOKIE', 'INPUT_SERVER', 'INPUT_ENV'], true)
```
ondrejmirtes: I don't like that this condition only works for literal `INPUT_GET` ConstFetch in the code, but doesn't work for `@param INPUT_GET|INPUT_POST $abc` from PHPDocs.

ondrejmirtes (changes requested):

src/Type/Php/FilterInputDynamicReturnTypeExtension.php
herndlm: so rebasing does summon you :D

good point, I also dislike that. I'm not good with constants, do you have any better ideas you can think of? Do I need to resolve them somehow I suppose?

ondrejmirtes: I thought this is already a solved problem, FilterFunctionReturnTypeHelper does this with `hasFlag`, `$this->getConstant()` etc.

herndlm: yeah that would work I guess, I thought there's a better approach. I'll switch it to use those helpers then :)

herndlm: adapted. I could not re-use those methods but found another approach. not sure yet if there's a better one though

ondrejmirtes:
Thank you!

PR #2011, opened 2022-11-19
Infinite loop regression test

rajyan:
small follow up fixes for https://github.com/phpstan/phpstan-src/pull/2007

I miss used `unsetExpression` while fixing https://github.com/phpstan/phpstan-src/pull/2007/files#diff-8ebf997b275a1b4e9be07f399f4b71c00617f43cfb16c3f9e4d088aabd522fa2R226 test.

Confirmed that the test causes an infinite loop without invalidateExpression.


ondrejmirtes:
Thank you!

PR #1799, opened 2022-10-06
Add @phpstan-self-out support

rvanvelzen:
This is a "missing" feature from the assert family which allows changing the `$this` type after a method call, which is particularly useful for setters.

See https://psalm.dev/docs/annotating_code/adding_assertions/#asserting-return-values-of-methods for the docs for psalm.

Syntax implementation in https://github.com/phpstan/phpdoc-parser/pull/149

PR #2012, opened 2022-11-19
Improve expression resolving of superglobals

herndlm:
Handles superglobals as any other expressions, which fixes narrowing them.

~~Closes https://github.com/phpstan/phpstan/issues/5805 (which I was totally not expecting, most likely unlocked also by recent changes from @rajyan)~~ nope, not anymore, this was fixed in https://github.com/phpstan/phpstan-src/pull/2030, I do understand that this change then is only improving things if people use superglobal expressions to narrow types, which might not be the case often in modern apps. Therefore I also get it if this is not wanted in the scope :)

The superglobals creation is most likely also a good integration point for https://github.com/phpstan/phpstan/issues/8037

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -190,7 +190,6 @@ class MutatingScope implements Scope
 		private ?Scope $parentScope = null,
 		private bool $nativeTypesPromoted = false,
 		private bool $explicitMixedInUnknownGenericNew = false,
-		private bool $explicitMixedForGlobalVariables = false,
 	)
 	{
 		if ($namespace === '') {
@@ -2150,7 +2149,6 @@ class MutatingScope implements Scope
 			$this->parentScope,
 			false,
 			$this->explicitMixedInUnknownGenericNew,
-			$this->explicitMixedForGlobalVariables,
 		);
 	}
 
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -468,10 +468,6 @@ class MutatingScope implements Scope
 	/** @api */
 	public function hasVariableType(string $variableName): TrinaryLogic
 	{
-		if ($this->isGlobalVariable($variableName)) {
-			return TrinaryLogic::createYes();
-		}
-
 		$varExprString = '$' . $variableName;
 		if (!isset($this->expressionTypes[$varExprString])) {
 			if ($this->canAnyVariableExist()) {
@@ -508,10 +504,6 @@ class MutatingScope implements Scope
 
 		$varExprString = '$' . $variableName;
 		if (!array_key_exists($varExprString, $this->expressionTypes)) {
-			if ($this->isGlobalVariable($variableName)) {
-				return new ArrayType(new StringType(), new MixedType($this->explicitMixedForGlobalVariables));
-			}
-
 			return new MixedType();
 		}
 
@@ -537,21 +529,6 @@ class MutatingScope implements Scope
 		}
 
 		return $variables;
-	}
-
-	private function isGlobalVariable(string $variableName): bool
-	{
-		return in_array($variableName, [
-			'GLOBALS',
-			'_SERVER',
-			'_GET',
-			'_POST',
-			'_FILES',
-			'_COOKIE',
-			'_SESSION',
-			'_REQUEST',
-			'_ENV',
-		], true);
 	}
 
 	/** @api */
@@ -2470,12 +2447,8 @@ class MutatingScope implements Scope
 			$this->isDeclareStrictTypes(),
 			null,
 			$this->getNamespace(),
-			array_merge($this->getConstantTypes(), [
-				'$this' => $thisHolder,
-			]),
-			array_merge($this->getNativeConstantTypes(), [
-				'$this' => $thisHolder,
-			]),
+			array_merge($this->getSuperglobalExpressionTypes(), $this->getConstantTypes(), ['$this' => $thisHolder]),
+			array_merge($this->getSuperglobalExpressionTypes(), $this->getNativeConstantTypes(), ['$this' => $thisHolder]),
 			[],
 			null,
 			null,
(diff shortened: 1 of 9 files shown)


herndlm:
wait, the prestashop errors do look like bugs..

herndlm:
Tried to implement the "bringing the superglobals expressions already into the scope from the beginning" mentioned from the previous comment, mostly to show how this could look. not sure if I like it much, it kind of makes sense and feels correct and even https://www.php.net/manual/en/language.variables.superglobals.php states "Superglobals — Built-in variables that are always available in all scopes",  but it apparently broke Rector in some weird way I don't understand yet. I had this in the past already and mentioned it in https://github.com/phpstan/phpstan-src/pull/1799#issuecomment-1269696300

src/Analyser/MutatingScope.php
```diff
@@ -467,10 +466,6 @@ public function afterOpenSslCall(string $openSslFunctionName): self
 	/** @api */
 	public function hasVariableType(string $variableName): TrinaryLogic
 	{
-		if ($this->isGlobalVariable($variableName)) {
-			return TrinaryLogic::createYes();
```
rajyan, line 472: I guess that the same thing can be done only by bring these conditions(here and getVariableType) inside !isset, before canAnyVariableExist without touching ScopeFacotrys

herndlm: what do you mean? my problem here was that scope merging does not work properly because the expression was missing. `hasVariableType()` always answered correctly I think

rajyan: Sorry, I read your comments in pull request afterwards. Never mind with this

herndlm:
Still not a 100% happy with how the superglobals are created and set, but at least it seems to be working fine now.

herndlm:
It's better, but still not optimal. If an approach like this is acceptable, I think the superglobals should be created outside the scope in the ScopeFactory and only passed around via constructor arg. But most likely Rector transforming will break again if I touch the scope constructor signature :/

herndlm:
Took me a while, but I finally figured out how I think it should be. The last commit though (a simple cleanup commit that removes an now unused argument from `MutatingScope`) touches `LazyInternalScopeFactory` and that is breaking the Rector downgrade apparently. Without that commit it was working fine.

src/Analyser/ScopeFactory.php
```diff
@@ -2,17 +2,51 @@
...
+		private InternalScopeFactory $internalScopeFactory,
+		private bool $explicitMixedForGlobalVariables,
+	)
 	{
 	}
 
 	public function create(ScopeContext $context): MutatingScope
 	{
-		return $this->internalScopeFactory->create($context);
+		$superglobalType = new ArrayType(new StringType(), new MixedType($this->explicitMixedForGlobalVariables));
```
staabm, line 23: as you are working on super global types: psalm recently made them more precise. maybe thats something we could also think about

https://github.com/vimeo/psalm/blob/7869cb548a5d88399f971ce9f897fe83abbd4213/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php#L594-L813

maybe as a followup

herndlm: Check the description 😊 I found your issue by luck

staabm: Ha awesome 😎

src/Analyser/MutatingScope.php
```diff
@@ -5008,4 +4987,19 @@ private function getNativeConstantTypes(): array
 		return $constantTypes;
 	}
 
+	/** @return array<string, ExpressionTypeHolder> */
+	private function getSuperglobalExpressionTypes(): array
+	{
+		$superglobalExpressionTypes = [];
+		$exprStrings = ['$GLOBALS', '$_SERVER', '$_GET', '$_POST', '$_FILES', '$_COOKIE', '$_SESSION', '$_REQUEST', '$_ENV'];
+		foreach ($this->expressionTypes as $exprString => $typeHolder) {
```
rajyan, line 5002: I think we need to handle super globals for native expressions separately.

herndlm: thx, that makes sense I guess, I'm currently trying to implement that, let's see where this goes..

btw., what's the reason that so many scope creations don't pass over native expression? E.g. `enterDeclareStrictTypes()` or `enterArrowFunction()` only retains `$this->expressionTypes`

herndlm:
very weird, rector downgrading worked now fine, I wonder if I really did fix this with moving phpdoc around or in general with the changes from https://github.com/phpstan/phpstan-src/pull/2022

I guess this is ready then again :)

herndlm:
rector error again WHY? I only rebased on 1.9.x and removed a commit that added a test which is not needed anymore... :/
update: looks like I could "fix" it by removing phpdoc from the scope factory classes that is already on the interface. very weird

ondrejmirtes:
Thank you!

herndlm:
On, tbh I didn't expect that this one would be merged without changes 😊 I think I duplicated the list of superglobals somewhere. Are you fine with that or should I put it somewhere central? Maybe I didn't because it would have needed a public static method or so and felt weird 🤔

ondrejmirtes:
When you open a pull request, there's always a risk I'm gonna merge it :) Feel free to improve your code in follow-up PRs.

herndlm:
Saw that it was in MutatingScope, but I believe that it's fine to not deduplicate it. So, nevermind and sorry for the noise

ondrejmirtes:
Sadly I have to revert this because of performance reasons. Blackfire comparison on top of 1.10.x-dev: https://blackfire.io/profiles/compare/21309b29-d2de-4f60-b6fe-1873b47e31b4/graph

herndlm:
oh that's unfortunate. just for my understanding  - how did you find out that it was this commit that caused it in the performance graph? or was it just a lucky guess?

ondrejmirtes:
Yesterday I saw a million calls to ArrayType::equals() which looked weird to me. Today I looked at recent releases and started git bisect, and the results made my suspicion obvious 😊

staabm:
> Blackfire comparison on top of 1.10.x-dev

is this a blackfire run over `php bin/phpstan -vvv --debug` - or some other script ?
I want to reproduce and fiddle a bit with this PRs impl

ondrejmirtes:
Yes

PR #2014, opened 2022-11-20
Fix `pathinfo($s, PATHINFO_ALL)` return type

staabm:
some tests fail with the current release: https://phpstan.org/r/adc64e75-4570-4aeb-acfd-bdcbf88ff872

`PATHINFO_ALL` is php8 only, see https://3v4l.org/4319K

### DIFF
--- a/tests/PHPStan/Analyser/data/pathinfo.php
+++ b/tests/PHPStan/Analyser/data/pathinfo.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace pathinfo;
+namespace pathinfoInference;
 
 use function PHPStan\Testing\assertType;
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/data/pathinfo.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace pathinfo;
```
ondrejmirtes: Don't add two files with the same namespace please :)

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #2017, opened 2022-11-21
Do not set native type to `mixed` on annotation assignment

herndlm:
Not sure, but this looked weird to me. Or did I misunderstand it?

My shuffle tests started failing partly, but that makes sense to me because they were influencing each other and not cleanly separated. E.g. this should fix https://phpstan.org/r/20b7a14a-21e8-4e2e-aca7-c1119fc8528c

dang, this doesn't seem to fix any open issues, I was hoping it would, should have created one for my snippet 😅

CC @rajyan

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -3863,7 +3863,12 @@ class NodeScopeResolver
 					$certainty = TrinaryLogic::createYes();
 				}
 
-				$scope = $scope->assignVariable($name, $varTag->getType(), new MixedType(), $certainty);
+				$scope = $scope->assignVariable(
+					$name,
+					$varTag->getType(),
+					$scope->getNativeType(new Variable($name)),
+					$certainty,
+				);
 			}
 		}
 
--- a/tests/PHPStan/Analyser/data/native-types.php
+++ b/tests/PHPStan/Analyser/data/native-types.php
@@ -202,3 +202,15 @@ function fooFunction(
 	assertType('string', $nonNullableString);
 	assertNativeType('string', $nonNullableString);
 }
+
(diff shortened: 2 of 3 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -3863,7 +3863,12 @@ private function processStmtVarAnnotation(MutatingScope $scope, Node\Stmt $stmt,
 					$certainty = TrinaryLogic::createYes();
 				}
 
-				$scope = $scope->assignVariable($name, $varTag->getType(), new MixedType(), $certainty);
+				$scope = $scope->assignVariable(
+					$name,
+					$varTag->getType(),
+					$scope->getNativeType(new Variable($name)),
```
herndlm, line 3869: since this internally falls back to `getType()` which uses `getVariableType()` it could be potentially (not sure yet if that fallback would be ever triggered here) wrong, right? is there any better way to get the native (variable/expression) type then?
oh wait, and the phpdoc states
> Gets type of an expression with no regards to phpDocs. Works for function/method parameters only.

is that still true? :/

rajyan:
Looks good to me!

I'm now thinking maybe making native type for assign variable required was not the right direction.
The purpose was to make native type exist in scope as full types too, so the variable certainty is handled correctly.

> dang, this doesn't seem to fix any open issues

I think changing the type by `@var` after the first assign is a rather rare case? and `treatPhpDocTypesAsCertain` is not used that much currently.

src/Analyser/NodeScopeResolver.php
rajyan: I think there are no ways now.
We have to finish https://github.com/phpstan/phpstan-src/pull/1943 to make `getNativeType` correct.

ondrejmirtes:
Thank you!

PR #2019, opened 2022-11-21
feat: disable tips via config

DrWursterich:
Hi, I'd like to be able to configure wether tips are displayed or not, which is what this pr adds.

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -109,6 +109,7 @@ parameters:
 	treatPhpDocTypesAsCertain: true
 	rememberPossiblyImpureFunctionValues: true
 	tipsOfTheDay: true
+	tips: true
 	reportMagicMethods: false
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
What's your use-case for disabling tips? I worry that someone could turn this off but then they're gonna post a support question that would already be answered by the tip, so this would be a net-negative for both parties.

DrWursterich:
That is a fair point.

Mainly it's the unicode symbol `💡` I would like to prevent beeing output. `--ansi` does not hide them.

ondrejmirtes:
Yeah, I'd accept a change that's more smart about this.

DrWursterich:
Okay, I understand. I will probably not have the time to implement this.

Thanks for the quick answers and have a nice day!

PR #1267, opened 2022-04-30
Add DateTimeModifyReturnTypeExtension

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/4927

I think it also closes https://github.com/phpstan/phpstan/issues/1934. There is a lot more method listed:
- diff
- format
- getOffset
- getTimezone
- setTime

But I never succeed to get `false` with those method, except when I passed a parameter with a wrong type.
```
$dateTime->diff('foo') // false in php 7.4, exception in 8.0
```
This is reported by phpstan as an error, so I don't think a returnType extension might be needed to say it's return false in php 7.4.

PR #2020, opened 2022-11-21
ImagickPixel::getColor() normalized param accept int instead of bool

blankse:
Should be fix https://phpstan.org/r/991c445c-3376-4892-947f-a1e5cd591d2d

See PHP doc:
https://www.php.net/manual/en/imagickpixel.getcolor.php

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -5166,7 +5166,7 @@ return [
 'ImagickPixel::clear' => ['bool'],
 'ImagickPixel::clone' => ['void'],
 'ImagickPixel::destroy' => ['bool'],
-'ImagickPixel::getColor' => ['array', 'normalized='=>'int'],
+'ImagickPixel::getColor' => ['array', 'normalized='=>'0|1|2'],
 'ImagickPixel::getColorAsString' => ['string'],
 'ImagickPixel::getColorCount' => ['int'],
 'ImagickPixel::getColorQuantum' => ['mixed'],
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -5166,7 +5166,7 @@ return [
 'ImagickPixel::clear' => ['bool'],
 'ImagickPixel::clone' => ['void'],
 'ImagickPixel::destroy' => ['bool'],
-'ImagickPixel::getColor' => ['array', 'normalized='=>'0|1|2'],
+'ImagickPixel::getColor' => ['array{r: int|float, g: int|float, b: int|float, a: int|float}', 'normalized='=>'0|1|2'],
 'ImagickPixel::getColorAsString' => ['string'],
 'ImagickPixel::getColorCount' => ['int'],
 'ImagickPixel::getColorQuantum' => ['mixed'],
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -5166,7 +5166,7 @@ return [
 'ImagickPixel::clear' => ['bool'],
 'ImagickPixel::clone' => ['void'],
 'ImagickPixel::destroy' => ['bool'],
-'ImagickPixel::getColor' => ['array', 'normalized='=>'bool'],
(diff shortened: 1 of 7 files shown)


resources/functionMap.php
```diff
@@ -5166,7 +5166,7 @@
 'ImagickPixel::clear' => ['bool'],
 'ImagickPixel::clone' => ['void'],
 'ImagickPixel::destroy' => ['bool'],
-'ImagickPixel::getColor' => ['array', 'normalized='=>'bool'],
+'ImagickPixel::getColor' => ['array', 'normalized='=>'int'],
```
blankse: Allow only 0, 1 or 2 integers
```suggestion
'ImagickPixel::getColor' => ['array', 'normalized='=>'0|1|2'],
```

resources/functionMap.php
```diff
@@ -5166,7 +5166,7 @@
 'ImagickPixel::clear' => ['bool'],
 'ImagickPixel::clone' => ['void'],
 'ImagickPixel::destroy' => ['bool'],
-'ImagickPixel::getColor' => ['array', 'normalized='=>'bool'],
+'ImagickPixel::getColor' => ['array', 'normalized='=>'0|1|2'],
```
mad-briller: you could also specify the return array shape:

```suggestion
'ImagickPixel::getColor' => ['array{r: int|float, g: int|float, b: int|float, a: int|float}', 'normalized='=>'0|1|2'],
```

this might be better off implemented as a return-type extension though, due to the nature of the 2nd arg and the way it affects the return types.

blankse: Can you give me an example of such a return type extension? Then I can try to implement it.

VincentLanglet: You can find a lot here https://github.com/phpstan/phpstan-src/tree/1.9.x/src/Type/Php

You'll need a DynamicMethodReturnTypeExtension, for instance https://github.com/phpstan/phpstan-src/blob/1.9.x/src/Type/Php/DateTimeModifyReturnTypeExtension.php
(You can see the whole PR here: https://github.com/phpstan/phpstan-src/pull/1267)

rajyan: and maybe we can write a conditional return type stub instead
https://phpstan.org/r/a8720857-2c4d-4bc9-a3bf-c8a46b51b02f

ex.
https://github.com/phpstan/phpstan-src/blob/2d4904895921ea3b37b5076c66690d4e067f9ed6/stubs/core.stub#L38-L42

blankse:
@VincentLanglet @rajyan I added the stub file. I think this should be enough right?

mad-briller:
apologies for my lack of reply @blankse github didn't notify me of your comment the first time

> @VincentLanglet @rajyan I added the stub file. I think this should be enough right?

it's worth adding some NodeScopeResolverTest cases to prove that it's working as intended:

1. create a file in `tests/PHPStan/Analyser/data` that contains examples of usage, for each value of the 2nd param
2. using the `PHPStan\Testing\assertType` method write assertions in the file; [for example](https://github.com/phpstan/phpstan-src/blob/1.9.x/tests/PHPStan/Analyser/data/bug-8373.php)
3. add a new yield statement to the [NodeScopeResolverTest](https://github.com/phpstan/phpstan-src/blob/1.9.x/tests/PHPStan/Analyser/NodeScopeResolverTest.php#L1135) that yields the file you created

blankse:
@mad-briller I added a test.

ondrejmirtes:
You should also verify that https://phpstan.org/r/991c445c-3376-4892-947f-a1e5cd591d2d is actually fixed with CallMethodsRuleTest.

Thank you.

blankse:
@ondrejmirtes I add a CallMethodsRuleTest for the params

ondrejmirtes:
Thank you!

PR #2021, opened 2022-11-21
Consistently pass `$nativeExpressionTypes` in MutatingScope

herndlm:
because why not? no really, I have no clue and am asking myself exactly that.

I went through all ScopeFactory::create() calls and this makes the native expression types passing consistent with the expression types passing AFAIK.

This passes native types properly for
- enterDeclareStrictTypes
- enterClosureBind / restoreOriginalScopeAfterClosureBind
- enterClosureCall
- enterArrowFunction

CC @rajyan

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -247,6 +247,7 @@ class MutatingScope implements Scope
 			null,
 			null,
 			$this->expressionTypes,
+			$this->nativeExpressionTypes,
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
For these changes, there should definitely be some tests that fail without these changes.

And one idea - we should move nativeExpressionTypes to be right after expressionTypes in MutatingScope constructor + in InternalScopeFactory.

herndlm:
> For these changes, there should definitely be some tests that fail without these changes.

Right, I thought the same, but was struggling with tests. I just managed to come up with something and can hopefully push it soon.

> And one idea - we should move nativeExpressionTypes to be right after expressionTypes in MutatingScope constructor + in InternalScopeFactory.

OK, then I'll quickly create a PR for that, that simplifies my other PRs including this one 👍

herndlm:
Oh, and unfortunately I need https://github.com/phpstan/phpstan-src/pull/2017 first to make the regression tests here work

herndlm:
while adding some tests I realized that currently $thisType is passed into nativeExpressionTypes, which isn't correct. Do we want to retain and pass a $nativeThisType? it would definitely work, but duplicate some code basically.

UPDATE: I think I dealt with it, also added tests for everything and ensured that they fail without the commit. they should turn magically green after merging #2017 and rebasing

ondrejmirtes:
This one should be rebased now I guess :)

herndlm:
yes and no. it needs https://github.com/phpstan/phpstan-src/pull/2017 unfortunately :) but really, no worries!

also, I need to fix conflicts now

ondrejmirtes:
Thank you!

PR #2022, opened 2022-11-21
Group `$expressionTypes` and  `$nativeExpressionTypes` in Scope creation methods

herndlm:
As requested in https://github.com/phpstan/phpstan-src/pull/2021#issuecomment-1322289671

I expect the Rector downgrade to fail again, let's see 🤞. local `make` was green. UPDATE: mysteriously it worked, weird. maybe also moving the phpdoc (which was done by PHPStorm) did something, well, not important as long as it works I guess..


herndlm:
CC @rajyan again. I hope this does not cause too many conflicts with whatever you're working on :/

herndlm:
btw do we also want to adapt the $expressionTypes variable names or is it not worth it? e.g. in some places they are still called variableTypeHolders and such AFAIK

ondrejmirtes:
I found a PR from you I can merge without testing locally, so that your withdrawal symptoms go away too :D Thank you very much!

PR #2023, opened 2022-11-21
Introduce class_implements dynamic return type

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/4335


src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,65 @@
...
+		return TypeCombinator::remove(
+			ParametersAcceptorSelector::selectSingle($functionReflection->getVariants())->getReturnType(),
+			new ConstantBooleanType(false),
+		);
+	}
+
+	private function canFunctionReturnFalse(Type $type, bool $autoload): bool
+	{
+		return !$type instanceof ObjectType
+			&& !$type instanceof ObjectWithoutClassType
```
VincentLanglet: Didn't find something like `isObject()` method.

VincentLanglet:
The failing build seems legit.
This assert https://github.com/doctrine/persistence/blob/3.1.x/src/Persistence/Mapping/RuntimeReflectionService.php#L44 won't be necessary thanks to the dynamic return type extension.

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
ondrejmirtes: `(new ObjectWithoutClassType())->isSuperTypeOf($type)`

ondrejmirtes (changes requested):

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
VincentLanglet: Welcome back @ondrejmirtes.
I hope you got good rest.

Thanks for the review, I made the change.

VincentLanglet:
I just checked the failing build for the integration builds and this is expected.

`class_uses()` or `class_implements()` never return false where they are used so a `!== false` check is now useless.

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
+	{
+		if (count($functionCall->getArgs()) < 1) {
+			return null;
+		}
+
+		$objectOrClassType = $scope->getType($functionCall->getArgs()[0]->value);
+		$autoload = !isset($functionCall->getArgs()[1])
```
ondrejmirtes, line 38: `$autoload` should be a TrinaryLogic. So you should do `(new ConstantBooleanType(true))->isSuperTypeOf($secondArg)`. Because you might have `true` (yes), `bool` or `mixed` (maybe), or `false` (no).

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,62 @@
...
+			|| $scope->getType($functionCall->getArgs()[1]->value)->equals(new ConstantBooleanType(true));
+
+		if ($objectOrClassType instanceof UnionType) {
```
ondrejmirtes: No reason to do `instanceof *Type` here. A single `$this->canFunctionReturnFalse($objectOrClassType, $autoload)` call should be sufficient.

ondrejmirtes (changes requested):
Also there are situations the function will always return `false`. Which aren't covered here.

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
VincentLanglet: Why ?
If I'm not sure that `true` is passed, I cannot remove the `false` return type.
So `bool`, `mixed` or `false` (ie, maybe and no) are working the same way.

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
VincentLanglet: If I just write
```
if ($this->canFunctionReturnFalse($objectOrClassType, $autoload)) {
     return null;
}
```
my tests are failing

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
ondrejmirtes: Please test these scenarios.

VincentLanglet:
> Also there are situations the function will always return false. Which aren't covered here.

Do you have examples ? How can I know that the class is not loaded ?

ondrejmirtes:
@VincentLanglet A constant string that isn't a class.

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
VincentLanglet: Tests added

src/Type/Php/ClassImplementsFunctionReturnTypeExtension.php
VincentLanglet: Nevermind, I found a way to do it.

VincentLanglet:
> @VincentLanglet A constant string that isn't a class.

Done :)

ondrejmirtes (changes requested):
I think the array values should be class-string in all cases.

VincentLanglet:
> I think the array values should be class-string in all cases.

I used `TypeCombinator::remove` to rely on the original definition:
https://github.com/phpstan/phpstan-src/blob/1.10.x/resources/functionMap.php#L951-L953

I suspect that `class-uses` return `array<string>` because it's a list of trait.
But maybe `FooTrait::class` is considered as a class-string too ; I didn't check.

ondrejmirtes:
For this purpose it is. You can use more specific trait-string and interface-string which currently mean the same thing but in the future they might be different.

VincentLanglet:
> For this purpose it is. You can use more specific trait-string and interface-string which currently mean the same thing but in the future they might be different.

Thanks for the precision ; I did the changes

ondrejmirtes (changes requested):
Regression test please for 4335, based on the issue bot output.

Can you also please go through these failures and think about whether they're valid or not? Thanks.

1) https://github.com/phpstan/phpstan-src/actions/runs/3891490930/jobs/6641853062
2) https://github.com/phpstan/phpstan-src/actions/runs/3891490930/jobs/6641856141

VincentLanglet:
> Regression test please for 4335, based on the issue bot output.

Added

> Can you also please go through these failures and think about whether they're valid or not? Thanks.
> 
> 1. [phpstan/phpstan-src/actions/runs/3891490930/jobs/6641853062](https://github.com/phpstan/phpstan-src/actions/runs/3891490930/jobs/6641853062)

This is expected since the false check is not needed for PHPStan.
https://github.com/nextras/orm/blob/main/src/Entity/Reflection/MetadataParser.php#L138-L139
https://github.com/nextras/orm/blob/main/src/Entity/Reflection/MetadataParser.php#L153-L154
https://github.com/nextras/orm/blob/main/src/Entity/Reflection/MetadataParser.php#L422-L423

> 2. [phpstan/phpstan-src/actions/runs/3891490930/jobs/6641856141](https://github.com/phpstan/phpstan-src/actions/runs/3891490930/jobs/6641856141)

This is expected too
https://github.com/doctrine/persistence/blob/3.1.x/src/Persistence/Mapping/RuntimeReflectionService.php#L38-L44
Since the class exists, the class_parents won't return false.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

ondrejmirtes:
Thank you.

PR #2027, opened 2022-11-23
Show combined memory usage in parallel mode

janedbal:
Closes https://github.com/phpstan/phpstan/issues/4683

Looks like this:
```
$ php bin/phpstan clear-result-cache -q && php bin/phpstan -v
Note: Using configuration file /usr/src/myapp/phpstan.neon.
 1486/1486 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100% 56 secs/56 secs

 [OK] No errors                                                                 

Took about 1.35 GB memory
```

### DIFF
--- a/src/Command/ErrorsConsoleStyle.php
+++ b/src/Command/ErrorsConsoleStyle.php
@@ -143,13 +143,22 @@ class ErrorsConsoleStyle extends SymfonyStyle
 
 	private function getProgressBarFormat(): ?string
 	{
-		$formatName = match ($this->getVerbosity()) {
-			OutputInterface::VERBOSITY_NORMAL => ProgressBar::FORMAT_NORMAL,
-			OutputInterface::VERBOSITY_VERBOSE => ProgressBar::FORMAT_VERBOSE,
-			OutputInterface::VERBOSITY_VERY_VERBOSE,
-			OutputInterface::VERBOSITY_DEBUG => ProgressBar::FORMAT_VERY_VERBOSE, // FORMAT_DEBUG shows invalid memory, avoid that
-			default => null,
-		};
+		switch ($this->getVerbosity()) {
+			case OutputInterface::VERBOSITY_NORMAL:
+				$formatName = ProgressBar::FORMAT_NORMAL;
+				break;
+			case OutputInterface::VERBOSITY_VERBOSE:
+				$formatName = ProgressBar::FORMAT_VERBOSE;
+				break;
+			case OutputInterface::VERBOSITY_VERY_VERBOSE:
+			case OutputInterface::VERBOSITY_DEBUG:
+				$formatName = ProgressBar::FORMAT_VERY_VERBOSE;
+				break;
+			default:
+				$formatName = null;
+				break;
+		}
+
 		if ($formatName === null) {
 			return null;
 		}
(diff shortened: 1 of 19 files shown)


src/Command/ErrorsConsoleStyle.php
```diff
@@ -118,6 +118,7 @@ private function wrap(array $rows, int $terminalWidth, int $maxHeaderWidth): arr
 	public function createProgressBar(int $max = 0): ProgressBar
 	{
 		$this->progressBar = parent::createProgressBar($max);
+		$this->progressBar->setFormat(' %current%/%max% [%bar%] %percent:3s%% %elapsed:6s%/%estimated:-6s%');
```
janedbal: Shows as 
```
 1486/1486 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100% 56 secs/56 secs
```

I copied [original debug verbosity format](https://symfony.com/doc/current/components/console/helpers/progressbar.html#custom-formats) and removed memory info to avoid confusion. Now everybody will see it this way no matter the verbosity level is selected. I think it is acceptable.

src/Parallel/ParallelAnalyser.php
```diff
@@ -221,6 +227,7 @@ public function analyse(
 			$internalErrorsCount === 0 ? $dependencies : null,
 			$exportedNodes,
 			$reachedInternalErrorsCountLimit,
+			(int) array_sum($peakMemoryUsages), // not 100% correct as the peak usages of workers might not have met
```
janedbal, line 237: I dont think there is a better option. Still imo much better than before. It seems working well when comparing with system memory used.

src/Parallel/ParallelAnalyser.php
```diff
@@ -198,7 +200,12 @@ public function analyse(
 
 				$job = array_pop($jobs);
 				$process->request(['action' => 'analyse', 'files' => $job]);
-			}, $handleError, function ($exitCode, string $output) use (&$internalErrors, &$internalErrorsCount, $processIdentifier): void {
+			}, $handleError, function ($exitCode, string $output) use (&$someChildEnded, &$peakMemoryUsages, &$internalErrors, &$internalErrorsCount, $processIdentifier): void {
+				if ($someChildEnded === false) {
+					$peakMemoryUsages['main'] = memory_get_usage(true);
```
janedbal, line 205: This is the point of time where phpstan consumes most memory (I believe in >99% cases), so I add main thread memory at that specific time.

src/Command/ErrorsConsoleStyle.php
ondrejmirtes: Doesn't this mean that the progress bar will be more verbose even without `-vvv`?

janedbal: It does, I dont see it as issue, but I can check how to avoid that if you want.

janedbal: Fixed, it should behave as it was, just not showing the memory in debug mode.

src/Parallel/ParallelAnalyser.php
janedbal: Tested against linux `ps` and it gives very good results, almost equal.

ondrejmirtes:
Thank you!

ondrejmirtes:
FYI these numbers were becoming irrelevant and not as precise when forking workers instead of spawning. Trying to improve it here: https://github.com/phpstan/phpstan-src/pull/6297

PR #2029, opened 2022-11-25
GithubAction: added job to verify tests fail without src/ changes

staabm:
Refs https://github.com/laminas/laminas-continuous-integration-action/issues/117

Refs https://github.com/phpstan/phpstan/discussions/8228#discussioncomment-3991522


ondrejmirtes:
Some suggestions:

* Run it only if both src/ and tests/ were changed
* Run it only if the usual tests passed
* Somehow you need to evaluate that one of the items from the matrix failed, and consider that a success build. Otherwise the build should fail. Possibly you'll need to save the tests result in a machine-readable format, upload the artifact, and do an "evaluate" step with some custom logic written in PHP. I do a similar thing already for the issue bot.

staabm:
I think it works as expected regarding
- Run it only if the usual tests passed
- evaluate that one of the items from the matrix failed

missing atm is 
- a check to run all these checks only if both src/ and tests/ were changed
- a prettty action summary

(and of course the temp changes need to be reverted, as soon as we get everything running as expected)

staabm:
I think we are feature complete.

- summary with accumulated failing test cases https://github.com/phpstan/phpstan-src/actions/runs/3553045875#summary-9719268731

- a run which skipped checks because it only contains tests/ but not src/ changes https://github.com/phpstan/phpstan-src/actions/runs/3553104875/jobs/5968433529 on commit [8b688fc](https://github.com/phpstan/phpstan-src/pull/2029/commits/8b688fc17587b246046140dd7128234524f62651)

- a run which verified tests because it contained src/ and tests/ changes https://github.com/phpstan/phpstan-src/actions/runs/3553108621 on commit [7d73752](https://github.com/phpstan/phpstan-src/pull/2029/commits/7d7375252b4822a961afddbad0a40c2247e37529)

the job dependency graph

grafik


if you are fine with the result, I will re-enable all lines I previously commented to speedup the feedback loop.

staabm:
If we agreed on this one, I can cleanup all the test code and remove the temporary changes

staabm:
squashed and rebased

staabm:
@ondrejmirtes is this PR useful? should I rebase it for 1.10.x?

PR #2033, opened 2022-11-28
Example test of huge slowdown in 1.9.x

janedbal:
Hi, I'm meeting huge slowdown (feels like infinite loop, but probably is some exponential time degradation) in 1.9.x, here is a separated (minimal) example.

- takes **minutes** on 1.9.x
- takes **4 secs** on 1.9.2 (which still feels a lot)

Possibly related to https://github.com/phpstan/phpstan-src/pull/2030


rajyan:
Thank you for the simple test case :+1:
> feels like infinite loop, but probably is some exponential time degradation

Yes, it is causing an exponential time degradation.

Could confirm https://github.com/phpstan/phpstan-src/pull/2030 solves this.

PR #2038, opened 2022-12-01
fix returntype for DateIntervall::createFromDateString

verfriemelt-dot-org:
this should fix https://github.com/phpstan/phpstan/issues/8442 
seems to be a simple fix, not sure if i done everything correct.

### DIFF
--- a/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
+++ b/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
@@ -28,15 +28,26 @@ class DateIntervalDynamicReturnTypeExtension implements DynamicStaticMethodRetur
 
 	public function getTypeFromStaticMethodCall(MethodReflection $methodReflection, StaticCall $methodCall, Scope $scope): ?Type
 	{
-		$dateTimeString = TypeUtils::getConstantStrings($scope->getType($methodCall->getArgs()[0]->value));
-
-		if (!($dateTimeString instanceof ConstantStringType)) {
-			return null;
+		$strings = $scope->getType($methodCall->getArgs()[0]->value)->getConstantStrings();
+		
+		if ($strings === []) {
+		   return null;
+		}
+		
+		if (count($strings) === 1) {
+           if (DateInterval::createFromDateString($dateTimeString->getValue()) === false) {
+    		 return new ConstantBooleanType(false);
+           }
+           return new ObjectType(DateInterval::class);
+		}
+		
+		foreach($strings as $string) {
+		  if (DateInterval::createFromDateString($dateTimeString->getValue()) === false) {
+		    return null;
+		  }
 		}
 
-		$isValid = DateInterval::createFromDateString($dateTimeString->getValue()) !== false;
-
-		return $isValid ? new ObjectType(DateInterval::class) : new ConstantBooleanType(false);
+		return new ObjectType(DateInterval::class);
 	}
 
 }
--- a/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
+++ b/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
@@ -31,7 +31,13 @@ implements DynamicStaticMethodReturnTypeExtension
 
 	public function getTypeFromStaticMethodCall(MethodReflection $methodReflection, StaticCall $methodCall, Scope $scope): ?Type
 	{
-		$strings = $scope->getType($methodCall->getArgs()[0]->value)->getConstantStrings();
+		$arguments = $methodCall->getArgs();
+
+		if (!isset($arguments[0])) {
+			return null;
+		}
+
+		$strings = $scope->getType($arguments[0]->value)->getConstantStrings();
 
 		$possibleReturnTypes = array_map(
 			static fn (ConstantStringType $s): string => gettype(@DateInterval::createFromDateString($s->getValue())),
--- a/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
+++ b/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
@@ -12,6 +12,7 @@ use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use function array_map;
+use function count;
 use function gettype;
 use function in_array;
 
@@ -43,6 +44,11 @@ implements DynamicStaticMethodReturnTypeExtension
 			static fn (ConstantStringType $s): string => gettype(@DateInterval::createFromDateString($s->getValue())),
 			$strings,
 		);
+
+		// the error case, when wrong types are passed
+		if (count($possibleReturnTypes) === 0) {
+			return null;
+		}
 
 		if (in_array('boolean', $possibleReturnTypes, true) && in_array('object', $possibleReturnTypes, true)) {
 			return null;
--- a/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
+++ b/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use DateInterval;
+use PhpParser\Node\Expr\StaticCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
+use PHPStan\Type\ObjectType;
+
+class DateIntervalDynamicReturnTypeExtension implements DynamicStaticMethodReturnTypeExtension
+{
+	public function getClass(): string
+	{
+		return DateInterval::class;
+	}
+
+	public function isStaticMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === 'createFromDateString';
+	}
+
+	public function getTypeFromStaticMethodCall( MethodReflection $methodReflection, StaticCall $methodCall, Scope $scope ): ?\PHPStan\Type\Type
+	{
+		$defaultReturnType = ParametersAcceptorSelector::selectFromArgs(
+			$scope,
+			$methodCall->getArgs(),
+			$methodReflection->getVariants(),
+		)->getReturnType();
+
+		$dateTimeString = $scope->getType($methodCall->getArgs()[0]->value);
+
+		if (!($dateTimeString instanceof ConstantStringType)) {
+			return $defaultReturnType;
+		}
+
+		$isValid = DateInterval::createFromDateString($dateTimeString->getValue()) !== false;
+
+		return $isValid ? new ObjectType(DateInterval::class) : new ConstantBooleanType(false);
+	}
+}
--- a/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
+++ b/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
@@ -7,13 +7,10 @@ use PhpParser\Node\Expr\StaticCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Type\Constant\ConstantBooleanType;
-use PHPStan\Type\Constant\ConstantStringType;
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Analyser/data/bug-6842.php
```diff
@@ -16,6 +16,7 @@ public function getScheduledEvents(
 		\DateTimeInterface $startDate,
 		\DateTimeInterface $endDate
 	): \Iterator {
+		/** @var \DateInterval $interval */
```
staabm: a DynamicReturnTypeExtension should be implemented, to make sure this case here does not require this inline `@var` to realize it cannot be `false`

similar to https://github.com/phpstan/phpstan-src/blob/1.9.x/src/Type/Php/DateTimeDynamicReturnTypeExtension.php which is handling the constant string argument

staabm (changes requested):
the snippet from https://github.com/phpstan/phpstan/issues/8442 should be added to this PR as a `NodeScopeResolverTest` including proper `assertType` calls

verfriemelt-dot-org:
i will look into this! 👍

tests/PHPStan/Analyser/data/bug-8442.php
```diff
@@ -0,0 +1,9 @@
+<?php declare(strict_types = 1);
+
+namespace Bug8442;
+
+use function PHPStan\Testing\assertType;
+use DateInterval;
+
+assertType('false', DateInterval::createFromDateString('foo'));
+assertType('DateInterval', DateInterval::createFromDateString('1 Day'));
```
staabm: please wrap this code into a function or method.

that way we can make sure that this code is not evaluated while phpstan is scanning the codebase

verfriemelt-dot-org: done

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+		$defaultReturnType = ParametersAcceptorSelector::selectFromArgs(
+			$scope,
+			$methodCall->getArgs(),
+			$methodReflection->getVariants(),
+		)->getReturnType();
+
+		$dateTimeString = $scope->getType($methodCall->getArgs()[0]->value);
+
+		if (!($dateTimeString instanceof ConstantStringType)) {
+			return $defaultReturnType;
```
herndlm: fyi you can also just return `null` here to simplify it even more and get rid of the `$defaultReturnType` block. PHPStan uses the default then on calling-site AFAIK

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+	{
+		$defaultReturnType = ParametersAcceptorSelector::selectFromArgs(
+			$scope,
+			$methodCall->getArgs(),
+			$methodReflection->getVariants(),
+		)->getReturnType();
+
+		$dateTimeString = $scope->getType($methodCall->getArgs()[0]->value);
+
+		if (!($dateTimeString instanceof ConstantStringType)) {
```
staabm: You should use `TypeUtils::getConstantStrings` instead

staabm: oohh I guess your extension is inspired by existing ones, which also miss to handle unions constant string etc. like in https://github.com/phpstan/phpstan-src/blob/1.9.x/src/Type/Php/DateTimeCreateDynamicReturnTypeExtension.php

if you adjust it for `TypeUtils::getConstantStrings()` you will also support cases like

```
if (rand(0,1)) {
  $interval = 'P1Y';
} else {
  $interval = 'P2Y';
}

$di = DateInterval::createFromDateString($interval);
```

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
verfriemelt-dot-org: nice, i'll do that

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
verfriemelt-dot-org: > oohh I guess your extension is inspired by existing ones

yes, excatly :) first time contribution, needed some inspiration :)

if i check the argument type like this:

```
$dateTimeString = TypeUtils::getConstantStrings($scope->getType($methodCall->getArgs()[0]->value));
```

i will get an array of string types? do i need to check if the `ConstantStringType::class` is contained within that array? i am a bit lost here?

staabm: see e.g. https://github.com/phpstan/phpstan-src/blob/253316f50dcafb60a6e1e603d16de9078d27ad44/src/Type/Php/StrSplitFunctionReturnTypeExtension.php#L91

you just loop over the array and ask `getValue` on it

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,42 @@
...
+	{
+		$dateTimeString = TypeUtils::getConstantStrings($scope->getType($methodCall->getArgs()[0]->value));
+
+		if (!($dateTimeString instanceof ConstantStringType)) {
+			return null;
+		}
+
+		$isValid = DateInterval::createFromDateString($dateTimeString->getValue()) !== false;
+
+		return $isValid ? new ObjectType(DateInterval::class) : new ConstantBooleanType(false);
```
staabm: since https://github.com/phpstan/phpstan-src/pull/1979 got merged this can now be simplified to something like this (and requires a few more tests)

`TypeUtils::getConstantStrings` is deprecated now.

```suggestion
		$strings = $scope->getType($methodCall->getArgs()[0]->value)->getConstantStrings();
		
		if ($strings === []) {
		   return null;
		}
		
		if (count($strings) === 1) {
           if (DateInterval::createFromDateString($dateTimeString->getValue()) === false) {
    		 return new ConstantBooleanType(false);
           }
           return new ObjectType(DateInterval::class);
		}
		
		foreach($strings as $string) {
		  if (DateInterval::createFromDateString($dateTimeString->getValue()) === false) {
		    return null;
		  }
		}

		return new ObjectType(DateInterval::class);
```

verfriemelt-dot-org: intresting, so if i understand this correctly, `getConstantStrings()` will return me _all possible constant strings_ with which the method could be called given the current context.
so for the example you gave earlier:

```php
if (rand(0,1)) {
  $interval = 'P1Y';
} else {
  $interval = 'P2Y';
}
```

i have two entries once `P1Y` and once `P2Y`. and were are testing both?

i think now it clicked, i was a bit confused by that.

tests/PHPStan/Analyser/data/bug-8442.php
```diff
@@ -0,0 +1,28 @@
...
+	if (rand(0,1)) {
+		$interval = '1 day';
+	} else {
+		$interval = '2 day';
+	}
+
+	assertType('DateInterval', DateInterval::createFromDateString($interval));
+
+	if (rand(0,1)) {
+		$interval = 'foo';
```
verfriemelt-dot-org, line 22: those are emitting warnings during testruns, how is that usually dealt with? just `@` the method in the extension?

staabm: Yes

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
staabm: Correct

staabm:
Would be great if you could adjust other similar existing extensions which also miss the `getConstantStrings()` handling, after this one was merged

tests/PHPStan/Analyser/data/bug-8442.php
staabm: do the tests emit the warnings or the extension code?

the tests itself should not need the `@`

verfriemelt-dot-org: from the extension code:

```
Warning: DateInterval::createFromDateString(): Unknown or bad format (foo) at position 0 (f): The timezone could not be found in the database in phpstan-src/src/Type/Php/DateIntervalDynamicReturnTypeExtension.php on line 38
```

staabm: yep, just add the `@` there

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	public function isStaticMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === 'createFromDateString';
+	}
+
+	public function getTypeFromStaticMethodCall(MethodReflection $methodReflection, StaticCall $methodCall, Scope $scope): ?Type
+	{
+		$strings = $scope->getType($methodCall->getArgs()[0 ]->value)->getConstantStrings();
```
ondrejmirtes: This code will crash if someone has `DateInterval::createFromDateString()` without an argument in their code.

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,54 @@
...
+
+		if ( $strings === []) {
+			return null;
+		}
+
+		if ( count($strings) === 1) {
```
ondrejmirtes: A single string shouldn't need any special case in this code, you should always iterate through all the strings and make the logic work generally.

ondrejmirtes (changes requested):
Also - you have some random spaces in there that I don't like but CS isn't picking up them as errors. Please get rid of them.

Like: `if ( count` or `$strings[0 ]`.

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
verfriemelt-dot-org: yeah, i agree. i stole this implementation from another file :) but you are absolutely right.

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
verfriemelt-dot-org: true, fixed that. but it leaves me wondering, what to return when this case is found?

```php
		$arguments = $methodCall->getArgs();

		if (!isset($arguments[0])) {
			return null;
		}
```
i opted for the default type but running code like this with warnings supressed does not execute at all. so what would be the correct thing to do here?

verfriemelt-dot-org:
> Also - you have some random spaces in there that I don't like but CS isn't picking up them as errors. Please get rid of them.
> 
> Like: `if ( count` or `$strings[0 ]`.

intresting, `make cs-fix` will not clean this up :thinking: i fixed them by hand :)

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
verfriemelt-dot-org: i opted for this:

```
	assertType('DateInterval|false',DateInterval::createFromDateString());
	assertType('DateInterval|false',DateInterval::createFromDateString(new stdClass()));
```

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,64 @@
...
+		$arguments = $methodCall->getArgs();
+
+		if (!isset($arguments[0])) {
+			return null;
+		}
+
+		$strings = $scope->getType($arguments[0]->value)->getConstantStrings();
+
+		$possibleReturnTypes = array_map(
+			static fn (ConstantStringType $s): string => gettype(@DateInterval::createFromDateString($s->getValue())),
```
ondrejmirtes: You don't need to use `gettype` if the only two possibilities are DateInterval or false. Write less clever code please :)

ondrejmirtes (changes requested):

src/Type/Php/DateIntervalDynamicReturnTypeExtension.php
verfriemelt-dot-org: maybe i am missing something, for the `false` i see your point.
but for the `DateInverval` case? :thinking: 

`in_array(DateInterval::class, $returnTypes)` does not work here? maybe i am too blind to see :D

verfriemelt-dot-org: ah, you mean this?

```
DateInterval::createFromDateString($s->getValue()) !== false ? 'DateInterval' : 'boolean'
```

ondrejmirtes:
Thank you!

PR #2039, opened 2022-12-01
Improve error messages for logical/boolean AND/OR (bleeding edge)

Wirone:
As discussed [here](https://github.com/phpstan/phpstan/discussions/8385) I would like to improve error messages for logical/boolean AND/OR checks.

- include kind (`logical` vs `boolean`), taken from processed node
- uppercased sigil (`AND` instead of `and`, `OR` instead of `or`) for better readability

### DIFF
--- a/src/Rules/Comparison/BooleanAndConstantConditionRule.php
+++ b/src/Rules/Comparison/BooleanAndConstantConditionRule.php
@@ -3,6 +3,8 @@
 namespace PHPStan\Rules\Comparison;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr\BinaryOp\BooleanAnd;
+use PhpParser\Node\Expr\BinaryOp\LogicalAnd;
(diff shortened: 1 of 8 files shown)


Wirone:
My bad, did not run integration tests before creating PR 🙄 Anyway, hope you'll like it!

ondrejmirtes:
Hi, thank you, but I don't want to change the messages again 😀

Wirone:
@ondrejmirtes but they were changed in bleeding edge so it's last moment to improve them before release 🙁 Current ones are not user friendly when it comes to logical ones: `Left side of and is always true` - the `and` does not look like operator, it can be misleading. As I wrote in [the discussion](https://github.com/phpstan/phpstan/discussions/8385), we have large legacy area where `and`/`or` are used and new release will require baseline change, if we need to do it just consider better error message.. I worked on it because @VincentLanglet, who as far as I see correctly worked on these errors before, agreed with me 😕

PR #2041, opened 2022-12-03
Fix sprintf() argument count edge cases

staabm:
refs https://github.com/phpstan/phpstan/issues/2342

### DIFF
--- a/src/Rules/Functions/PrintfParametersRule.php
+++ b/src/Rules/Functions/PrintfParametersRule.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Rules\Functions;
 
+use ArgumentCountError;
 use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Expr\FuncCall;
@@ -14,6 +15,7 @@ use function count;
 use function in_array;
 use function max;
 use function sprintf;
+use function sscanf;
 use function strlen;
 use function strtolower;
 use const PREG_SET_ORDER;
@@ -108,6 +110,16 @@ class PrintfParametersRule implements Rule
 
 	private function getPlaceholdersCount(string $functionName, string $format): int
 	{
+		try {
+			@sprintf($format);
+		} catch (ArgumentCountError $e) {
+			$message = $e->getMessage();
+			sscanf($message, '%d arguments are required, %d given', $required, $given);
+
+			return $required - 1;
+		}
+
+		// fallback for php < 8
 		$specifiers = in_array($functionName, ['sprintf', 'printf'], true) ? '[bcdeEfFgGosuxX%s]' : '(?:[cdDeEfinosuxX%s]|\[[^\]]+\])';
 		$addSpecifier = '';
 		if ($this->phpVersion->supportsHhPrintfSpecifier()) {
--- a/src/Rules/Functions/PrintfParametersRule.php
+++ b/src/Rules/Functions/PrintfParametersRule.php
@@ -13,6 +13,7 @@ use PHPStan\Rules\RuleErrorBuilder;
 use function array_filter;
 use function count;
 use function in_array;
+use function is_int;
 use function max;
 use function sprintf;
 use function sscanf;
@@ -113,10 +114,11 @@ class PrintfParametersRule implements Rule
 		try {
 			@sprintf($format);
 		} catch (ArgumentCountError $e) {
-			$message = $e->getMessage();
-			sscanf($message, '%i arguments are required, %i given', $required, $given);
+			sscanf($e->getMessage(), '%i arguments are required, %i given', $required, $given);
 
-			return $required - 1;
+			if (is_int($required)) {
+				return $required - 1;
+			}
 		}
 
 		// fallback for php < 8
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Functions/data/printf.php
```diff
@@ -9,7 +9,7 @@
 sprintf('%2$s %1$s %% %1$s %%%', 'one'); // one parameter missing
 sprintf('%2$s %%'); // two parameters required
 sprintf('%2$s %1$s %1$s %s %s %s %s'); // four parameters required
-sprintf('%2$s %1$s %% %s %s %s %s %%% %%%%', 'one', 'two', 'three', 'four'); // ok
```
staabm, line 12: this test-case was wrong, see https://3v4l.org/2dPI6

the changes in this PR result in properly detecting the error within this test, but only on php8+.

for php 7 the "wrong" test assumption of this test is not changed.
I figured its not worth complicating the already complex regex based parsing (which - as of this PR - is only required for php7 and can be dropped when we raise the min php version to 8).

tests/PHPStan/Rules/Functions/data/printf.php
```diff
@@ -52,3 +52,5 @@
 sprintf($variousPlaceholderCount, 'bar');
 sprintf($variousPlaceholderCount, 'bar', 'baz');
 sprintf($variousPlaceholderCount, 'bar', 'baz', 'lorem');
+
+sprintf('UPDATE {{%loans}} SET field=1');
```
staabm: from https://github.com/phpstan/phpstan/issues/2342#issuecomment-516841318

src/Rules/Functions/PrintfParametersRule.php
```diff
@@ -109,6 +112,17 @@ public function processNode(Node $node, Scope $scope): array
 
 	private function getPlaceholdersCount(string $functionName, string $format): int
 	{
+		try {
```
staabm, line 115: instead of re-implemeting the complex php-src argument parsing, I just delegate to the runtime and interpret the error message.

this works as of php8+ when php-src started throwing `ArgumentCountError`

src/Rules/Functions/PrintfParametersRule.php
```diff
@@ -109,6 +112,17 @@ public function processNode(Node $node, Scope $scope): array
...
+			@sprintf($format);
+		} catch (ArgumentCountError $e) {
+			sscanf($e->getMessage(), '%i arguments are required, %i given', $required, $given);
+
+			if (is_int($required)) {
+				return $required - 1;
+			}
+		}
+
+		// fallback for php < 8
```
staabm, line 129: the whole regex parsing can be deleted when phpstan min-version is raised to php8+

staabm:
The error is legit https://github.com/PrestaShop/PrestaShop/blob/ca9d81aae0bb17f3e767bb5d835348ed144a3ab5/src/Core/Shop/LogoUploader.php#L112

src/Rules/Functions/PrintfParametersRule.php
VincentLanglet: Shouldnt an early return be added for php>8 users then if the previous checks are enough then ?

ondrejmirtes (changes requested):
Please also analyse `printf-php8.php` on PHP 7 in a separate test case method, probably with different errors asserted.

ondrejmirtes:
Does this close https://github.com/phpstan/phpstan/issues/2342? If yes, please change the PR description.

staabm:
> Does this close [phpstan/phpstan#2342](https://github.com/phpstan/phpstan/issues/2342)? If yes, please change the PR description.

it is only fixed for php8+. for php7 the issue still exists. if you are fine with that, I can update the description to a `closes`.

ondrejmirtes:
If PHP 7 still crashes then it's not really fixed...

staabm:
Ok but PR is stll acceptable after I add more tests?

ondrejmirtes:
It signals to me that this is wrong approach. `getPlaceholdersCount` should simply report the correct number, without catching exceptions. So the regexes should be fixed.

staabm:
> It signals to me that this is wrong approach. `getPlaceholdersCount` should simply report the correct number, without catching exceptions. So the regexes should be fixed.

I agree, fixing the regexes would be the best move regarding php version support for this fix.
I tried and couldn't come up with a good regex which works and doesn't regress existing functionality.

I just committed a small fix in [cf6ded3](https://github.com/phpstan/phpstan-src/pull/2041/commits/cf6ded3ab37d1b57084a5b7fad3f7da698d9c042) so you get an idea, that the whole regex thing is only doing php7 stuff after this PR. the newly added 5 lines of code cover all existing logic on php8 (and maybe even cases we did not yet implement - or may be implemented in php-src in the future - because it re-uses the php-src native format checking).

from my point of view this is a big step forward, which allows us to remove a lot of code when phpstan min php-version will be raised to php8+. its also a great way to get rid of clumsy regex stuff which is prone to errors in the future.

I was not able to come up with a regex fix for this problem. I will not invest time in doing more research for such a edge case bug. I am fine with closing the PR so, in case you don't like this php version inconsistency.

ondrejmirtes:
I feel like we should continue with the regex path into the future. I like it more than parsing exception error message which can change at any time.

staabm:
fwiw, this PR also detected https://github.com/phpstan/phpstan/issues/8547

PR #2043, opened 2022-12-03
Fix resolving of __DIR__, __NAMESPACE__ in traits

staabm:
closes https://github.com/phpstan/phpstan/issues/3019

the test fails without the fix like
```diff
There were 2 failures:

1) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-3019.php:11" ('type', '/Users/staabm/workspace/phpst...19.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\Constant\ConstantStringType Object (...), 11)
Expected type string, got type 'Bug3019\\Bar' in /Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-3019.php on line 11.
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'string'
+''Bug3019\\Bar''

/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:96
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:1150

2) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-3019.php:12" ('type', '/Users/staabm/workspace/phpst...19.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\Constant\ConstantStringType Object (...), 12)
Expected type string, got type 'Bug3019' in /Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-3019.php on line 12.
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'string'
+''Bug3019''

3) PHPStan\Rules\Comparison\UnreachableTernaryElseBranchRuleTest::testBug3019
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'
+'09: Else branch is unreachable because ternary operator condition is always true.
 '
```


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -331,6 +331,10 @@ public function getType(Expr $expr, InitializerExprContext $context): Type
 		}
 
 		if ($expr instanceof MagicConst\Class_) {
+			if ($context->getTraitName() !== null) {
+				return new StringType();
```
ondrejmirtes: ClassStringType combined with AccessoryLiteralStringType

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -339,6 +343,10 @@ public function getType(Expr $expr, InitializerExprContext $context): Type
 		}
 
 		if ($expr instanceof MagicConst\Namespace_) {
+			if ($context->getTraitName() !== null) {
+				return new StringType();
```
ondrejmirtes: combined with AccessoryLiteralStringType

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/bug-3019.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Bug3019;
+
+use function PHPStan\Testing\assertType;
+
+trait FooTrait
+{
+	public function doFoo(): void
+	{
+		assertType('string', __CLASS__);
```
staabm, line 11: I am wondering a bit that we don't get a class-string here

staabm: I am wondering that we don't get a class-string here

tests/PHPStan/Rules/Comparison/data/bug-3019.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Bug3019;
```
ondrejmirtes: Duplicate namespaces/duplicate classes

ondrejmirtes: Just use a single file used in both RuleTestCase and NodeScopeResolverTest.

tests/PHPStan/Analyser/data/bug-3019.php
ondrejmirtes: My guess it's not checked because a duplicate trait with the same name exists.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Comparison/data/bug-3019.php
staabm: ahh good catch.

when I have some time I will work on the rule to detect duplicate namespaces :)

staabm:
should this be re-targeted for 1.10.x or will we still have 1.9.x releases before 1.10 is out?

ondrejmirtes:
> or will we still have 1.9.x releases before 1.10 is out

Yeah, sure, that's the point of these branches :)

tests/PHPStan/Analyser/data/bug-3019.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Bug3019;
```
staabm, line 3: I verified this code still fails without the patch for both Test-cases

ondrejmirtes:
Nice, thank you!

PR #2044, opened 2022-12-03
Fix comparison with get_class() in traits always evaluate to true

staabm:
closes https://github.com/phpstan/phpstan/issues/3633

see https://3v4l.org/LalXd

test fails without the src change:

```diff
1) PHPStan\Rules\Comparison\StrictComparisonOfDifferentTypesRuleTest::testBug3633
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'23: Strict comparison using === between class-string<Bug3633\HelloWorld> and 'Bug3633\\OtherClass' will always evaluate to false.
+'10: Strict comparison using === between class-string<Bug3633\HelloWorld> and 'Bug3633\\OtherClass' will always evaluate to false.
+23: Strict comparison using === between class-string<Bug3633\HelloWorld> and 'Bug3633\\OtherClass' will always evaluate to false.
+07: Strict comparison using === between class-string<Bug3633\OtherClass> and 'Bug3633\\HelloWorld' will always evaluate to false.
 35: Strict comparison using === between class-string<Bug3633\OtherClass> and 'Bug3633\\HelloWorld' will always evaluate to false.
+07: Strict comparison using === between class-string<Bug3633\FinalClass> and 'Bug3633\\HelloWorld' will always evaluate to false.
+10: Strict comparison using === between class-string<Bug3633\FinalClass> and 'Bug3633\\OtherClass' will always evaluate to false.
 50: Strict comparison using === between class-string<Bug3633\FinalClass> and 'Bug3633\\HelloWorld' will always evaluate to false.
 53: Strict comparison using === between class-string<Bug3633\FinalClass> and 'Bug3633\\OtherClass' will always evaluate to false.
 59: Strict comparison using === between 'Bug3633\\FinalClass' and 'Bug3633\\FinalClass' will always evaluate to true.
 '
```

### DIFF
--- a/src/Type/Php/GetClassDynamicReturnTypeExtension.php
+++ b/src/Type/Php/GetClassDynamicReturnTypeExtension.php
@@ -17,7 +17,7 @@ use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\StaticType;
-use PHPStan\Type\StringType;
+use PHPStan\Type\ThisType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\TypeWithClassName;
@@ -37,6 +37,10 @@ class GetClassDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExt
 		$args = $functionCall->getArgs();
 
 		if (count($args) === 0) {
+			if ($scope->isInTrait()) {
+				return new ClassStringType();
+			}
+
 			if ($scope->isInClass()) {
 				return new ConstantStringType($scope->getClassReflection()->getName(), true);
 			}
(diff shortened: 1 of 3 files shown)


src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -42,6 +44,10 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 			return new ConstantBooleanType(false);
 		}
 
+		if ($scope->isInTrait()) {
+			return new StringType();
```
ondrejmirtes: We can at least say that it's going to be ClassStringType.

ondrejmirtes (changes requested):

ondrejmirtes (changes requested):
Oh, I just realized this PR is more wrong than right.

1) We need to cover the same scenario for `get_class()` without arguments.
2) We can still return correct specific result for an argument that isn't `$this`.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #2045, opened 2022-12-03
Fixed '$this instanceof X will always be false' in traits

staabm:
closes https://github.com/phpstan/phpstan/issues/3632

test fails without the fix
```diff
1) PHPStan\Rules\Classes\ImpossibleInstanceOfRuleTest::testBug3632
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'
+'10: Instanceof between *NEVER* and Bug3632\OtherClass will always evaluate to false.
 '
```

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2424,18 +2424,18 @@ class MutatingScope implements Scope
 	public function enterClass(ClassReflection $classReflection): self
 	{
 		$thisHolder = ExpressionTypeHolder::createYes(new Variable('this'), new ThisType($classReflection));
+		$constantTypes = $this->getConstantTypes();
+		$constantTypes['$this'] = $thisHolder;
+		$nativeConstantTypes = $this->getNativeConstantTypes();
+		$nativeConstantTypes['$this'] = $thisHolder;
 
 		return $this->scopeFactory->create(
 			$this->context->enterClass($classReflection),
 			$this->isDeclareStrictTypes(),
 			null,
 			$this->getNamespace(),
-			array_merge($this->getConstantTypes(), [
-				'$this' => $thisHolder,
-			]),
-			array_merge($this->getNativeConstantTypes(), [
-				'$this' => $thisHolder,
-			]),
+			$constantTypes,
+			$nativeConstantTypes,
 			[],
 			null,
 			null,
@@ -2462,19 +2462,20 @@ class MutatingScope implements Scope
 		if ($classReflection === null) {
 			throw new ShouldNotHappenException();
 		}
+
 		$thisHolder = ExpressionTypeHolder::createYes(new Variable('this'), new ThisType($classReflection, null, $traitReflection));
+		$expressionTypes = $this->expressionTypes;
+		$expressionTypes['$this'] = $thisHolder;
+		$nativeExpressionTypes = $this->nativeExpressionTypes;
+		$nativeExpressionTypes['$this'] = $thisHolder;
 
 		return $this->scopeFactory->create(
 			$traitContext,
 			$this->isDeclareStrictTypes(),
 			$this->getFunction(),
 			$namespace,
-			array_merge($this->expressionTypes, [
-				'$this' => $thisHolder,
-			]),
-			array_merge($this->nativeExpressionTypes, [
-				'$this' => $thisHolder,
-			]),
+			$expressionTypes,
+			$nativeExpressionTypes,
 			[],
 			$this->inClosureBindScopeClass,
 			$this->anonymousFunctionReflection,
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2456,13 +2456,25 @@ class MutatingScope implements Scope
 		if (count($traitNameParts) > 1) {
 			$namespace = implode('\\', array_slice($traitNameParts, 0, -1));
 		}
+
+		$traitContext = $this->context->enterTrait($traitReflection);
+		$classReflection = $traitContext->getClassReflection();
+		if ($classReflection === null) {
+			throw new ShouldNotHappenException();
+		}
+		$thisHolder = ExpressionTypeHolder::createYes(new Variable('this'), new ThisType($classReflection, null, $traitReflection));
+
 		return $this->scopeFactory->create(
-			$this->context->enterTrait($traitReflection),
+			$traitContext,
 			$this->isDeclareStrictTypes(),
 			$this->getFunction(),
 			$namespace,
-			$this->expressionTypes,
-			$this->nativeExpressionTypes,
+			array_merge($this->expressionTypes, [
+				'$this' => $thisHolder,
+			]),
+			array_merge($this->nativeExpressionTypes, [
+				'$this' => $thisHolder,
+			]),
 			[],
 			$this->inClosureBindScopeClass,
 			$this->anonymousFunctionReflection,
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -1131,10 +1131,4 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		return $errors;
 	}
 
-	public function testTraitInstanceOf(): void
-	{
-		$errors = $this->runAnalyse(__DIR__ . '/data/trait-instance-of.php');
-		$this->assertNoErrors($errors);
-	}
-
 }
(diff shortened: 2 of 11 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -909,12 +909,17 @@ private function resolveType(Expr $node): Type
...
 				$expressionType = $this->getNativeType($node->expr);
 			}
-			if (
-				$this->isInTrait()
-				&& TypeUtils::findThisType($expressionType) !== null
-			) {
-				return new BooleanType();
+
+			if ($this->isInTrait()) {
+				if ($node->expr instanceof Expr\Variable
```
staabm: when inside a trait, it seems `$node->expr` is resolved into a `NeverType`.

actually I thought it should return a `ThisType` and it should be catched by `TypeUtils::findThisType` below, but it doesn't work like that.

staabm: looked into it again and I am still not know more then the comment above.

maybe it even intentionally resolves a 'this' variable to NeverType within a trait, as technically the trait only has a 'this' when used in a class context?

I find it interessting that `TypeUtils::findThisType`  is just used in 2 cases over the whole codebase and both within a `isInTrait` case.

staabm: ohh I got an idea. I guess the cause is that we are trying to remove `ObjectType<T>` from `ThisType<T>`.
https://phpstan.org/r/7a4872ff-c9e4-4d40-bced-226efd0288f1

staabm:
I have approached the problem from a few different angles. All worked more or less. 

At the end I get the feeling maybe this should be fixed at the core by e.g. treating `$this` just like a simple `ObjectType` when analysing in trait context

src/Rules/Classes/ImpossibleInstanceOfRule.php
```diff
@@ -59,6 +60,14 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		if ($scope->isInTrait()
+			&& $node->expr instanceof Node\Expr\Variable
+			&& is_string($node->expr->name)
+			&& $node->expr->name === 'this'
```
staabm: I tried different ways on how to detect this properly. since `$this` in traits is class-context dependent the inference turns the type into a `NeverType`. 

checking for the `NeverType` here would be too agressive. therefore I figured using a explicit `$this` check would be the most readable way to handle the case in question.

----

since `$this` is turned into a `NeverType` the `TypeUtils::findThisType()` also doesn't work.

when looking at how we use `TypeUtils::findThisType()` all over phpstan-src, we could also add a new `TypeUtils`-method:

```php
public static function isThisType(Node $expr, Scope $scope): bool
{
	if ($expr instanceof Node\Expr\Variable
		&& is_string($expr->name)
		&& $expr->name === 'this'
	) {
		return true;
	}

	$type = $scope->getType($expr);
	if ($type instanceof ThisType) {
		return true;
	}

		foreach ($type->getTypes() as $innerType) {
			if ($innerType instanceof ThisType) {
				return true;
			}
		}
	}

	return false;
}
```

and use it all over the codebase instead of `TypeUtils::findThisType()`

tests/PHPStan/Rules/Classes/data/trait-instance-of.php
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types = 1);
+
+namespace TraitInstanceOf;
+
+trait Foo {
+	public function test(HelloWorld $hw): string {
+		if ($hw instanceof HelloWorld) {
```
staabm: since I had different implementations in between which also swalled errors for lokal vars within traits, I added a separate test-case for it

staabm:
regarding the build: I can see the same errors in other PRs so I think these are unrelated

ondrejmirtes (changes requested):
The underyling issues we need to fix is: https://phpstan.org/r/d8b183e1-7099-478e-af58-fdd5e8ab7d21

So it's not sufficient to try to if-else our way out of this problem in a single rule, we need to fix the typesystem here.

So something needs to happen to handling of `Instanceof_` in TypeSpecifier.

staabm:
> So something needs to happen to handling of `Instanceof_` in TypeSpecifier.

that was a great hint, thank you. I feel dumb for searching the solution in a lot of other phpstan components, but not the TypeSpecifier :)

ondrejmirtes:
I decided this still doesn't work and we need a different approach. ThisType in a trait context needs to behave a bit differently than in a class. It needs to allow intersection with other classes, almost like an interface.

staabm:
So the bottom line is, that we need a new type?

ondrejmirtes:
Some boolean arg in ThisType would suffice. But maybe TraitThisType would work too.

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1131,4 +1131,10 @@ private function runAnalyse(string $file, ?array $allAnalysedFiles = null): arra
 		return $errors;
 	}
 
+	public function testTraitInstanceOf(): void
```
staabm: figured a new test to make sure the newly added type-narrowing on $this in traits allows to call methods of narrowed types could be useful

tests/PHPStan/Analyser/data/trait-instance-of.php
```diff
@@ -0,0 +1,41 @@
...
+			return 'hello world';
+		}
+		assertType('$this(TraitInstanceOf\HelloWorld~$this(TraitInstanceOf\HelloWorld))', $this);
+		if ($this instanceof OtherClass) {
+			$this->doOther();
+
+			assertType('$this(TraitInstanceOf\HelloWorld~$this(TraitInstanceOf\HelloWorld))&$this(TraitInstanceOf\OtherClass~$this(TraitInstanceOf\HelloWorld))', $this);
+			return 'other class';
+		}
+		assertType('$this(TraitInstanceOf\HelloWorld~$this(TraitInstanceOf\HelloWorld))', $this);
```
staabm: I am not sure about this type here. maybe it could/should contain also that `OtherClass` is subtracted.
If so, I don't now what is missing to make this work

ondrejmirtes (changes requested):
I see some problems with this PR.

1) TypeSpecifier shouldn't be touched. Instead, MutatingScope::enterTrait should handle ThisType.
2) Any changes to isSuperTypeOf logic needs to be heavily covered by tests in TypeCombinatorTest - in dataUnion and dataIntersection.
3) There are some cases the method should return `no`: a) for a final class that does not use the trait. b) For an ObjectType with a trait name c) For a type that's already subtracted in ThisType. This influences how unions and intersections are gonna end up looking like.

tests/PHPStan/Rules/Classes/ImpossibleInstanceOfRuleTest.php
```diff
@@ -368,4 +368,24 @@ public function testBug5333(): void
...
+	}
+
+	public function testTraitInstanceOf(): void
+	{
+		$this->checkAlwaysTrueInstanceOf = true;
+		$this->treatPhpDocTypesAsCertain = false;
+
+		$this->analyse([__DIR__ . '/../../Analyser/data/trait-instance-of.php'], [
+			[
+				'Instanceof between $this(TraitInstanceOf\ATrait1Class) and trait TraitInstanceOf\Trait2 will always evaluate to false.',
```
staabm: for some reason this error is not reported. I get a error though, when running phpstan on this file in isolation.

otherwise I think this looks fine. I will add new TypeCombinator Tests now

src/Analyser/MutatingScope.php
```diff
@@ -2456,13 +2456,25 @@ public function enterTrait(ClassReflection $traitReflection): self
...
 		return $this->scopeFactory->create(
-			$this->context->enterTrait($traitReflection),
+			$traitContext,
 			$this->isDeclareStrictTypes(),
 			$this->getFunction(),
 			$namespace,
-			$this->expressionTypes,
-			$this->nativeExpressionTypes,
+			array_merge($this->expressionTypes, [
+				'$this' => $thisHolder,
```
ondrejmirtes: No need for array_merge. Simply do this before:

```php
$expressionTypes = $this->expressionTypes;
$expressionType['$this'] = $thisHolder;
```

And the same thing for nativeExpressionTypes.

tests/PHPStan/Rules/Classes/ImpossibleInstanceOfRuleTest.php
ondrejmirtes: That's ExistingClassInInstanceOfRule rule, not ImpossibleInstanceOfRule rule.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: I see. I was inspired by another place where it was done this way. fixed both in 749fbab

ondrejmirtes:
Thank you.

staabm:
thanks for the directions. I couldn't have finished this one without your feedback <3.

PR #2046, opened 2022-12-03
Fix class-const resolving will always evaluate to true/false in traits

staabm:
closes https://github.com/phpstan/phpstan/issues/4570

test fails without the src changes like
```diff
1) PHPStan\Rules\Comparison\ImpossibleCheckTypeFunctionCallRuleTest::testBug4570
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'
+'08: Call to function array_key_exists() with 'valueToFetch' and array{someValue: 'abc', valueToFetch: '123'} will always evaluate to true.
+08: Call to function array_key_exists() with 'valueToFetch' and array{someValue: 'abc', someOtherValue: '123'} will always evaluate to false.
 '
```

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -1621,11 +1621,13 @@ class MutatingScope implements Scope
 			if ($this->hasExpressionType($node)->yes()) {
 				return $this->expressionTypes[$exprString]->getType();
 			}
+
 			return $this->initializerExprTypeResolver->getClassConstFetchTypeByReflection(
 				$node->class,
 				$node->name->name,
 				$this->isInClass() ? $this->getClassReflection() : null,
 				fn (Expr $expr): Type => $this->getType($expr),
+				InitializerExprContext::fromScope($this),
 			);
 		}
 
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -2,9 +2,6 @@
 
 namespace PHPStan\Analyser;
 
-use Bug4288\MyClass;
-use Bug4713\Service;
-use ExtendingKnownClassWithCheck\Foo;
 use PHPStan\File\FileHelper;
 use PHPStan\Reflection\InitializerExprContext;
 use PHPStan\Reflection\InitializerExprTypeResolver;
@@ -13,6 +10,7 @@ use PHPStan\Reflection\SignatureMap\SignatureMapProvider;
 use PHPStan\Testing\PHPStanTestCase;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\MixedType;
 use function extension_loaded;
 use function restore_error_handler;
 use function sprintf;
@@ -84,7 +82,7 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertNoErrors($errors);
 
 		$reflectionProvider = $this->createReflectionProvider();
-		$this->assertTrue($reflectionProvider->hasClass(Foo::class));
+		$this->assertTrue($reflectionProvider->hasClass(\ExtendingKnownClassWithCheck\Foo::class)); // phpcs:ignore
 	}
 
 	public function testInfiniteRecursionWithCallable(): void
@@ -354,11 +352,11 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertSame('Method Bug4713\Service::createInstance() should return Bug4713\Service but returns object.', $errors[0]->getMessage());
 
 		$reflectionProvider = $this->createReflectionProvider();
-		$class = $reflectionProvider->getClass(Service::class);
+		$class = $reflectionProvider->getClass(\Bug4713\Service::class); // phpcs:ignore
 		$parameter = ParametersAcceptorSelector::selectSingle($class->getNativeMethod('createInstance')->getVariants())->getParameters()[0];
 		$defaultValue = $parameter->getDefaultValue();
 		$this->assertInstanceOf(ConstantStringType::class, $defaultValue);
-		$this->assertSame(Service::class, $defaultValue->getValue());
+		$this->assertSame(\Bug4713\Service::class, $defaultValue->getValue()); // phpcs:ignore
 	}
 
 	public function testBug4288(): void
@@ -367,7 +365,28 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertNoErrors($errors);
 
 		$reflectionProvider = $this->createReflectionProvider();
-		$class = $reflectionProvider->getClass(MyClass::class);
+		$class = $reflectionProvider->getClass(\Bug4288\MyClass::class); // phpcs:ignore
+		$parameter = ParametersAcceptorSelector::selectSingle($class->getNativeMethod('paginate')->getVariants())->getParameters()[0];
+		$defaultValue = $parameter->getDefaultValue();
+		$this->assertInstanceOf(MixedType::class, $defaultValue);
+
+		$nativeProperty = $class->getNativeReflection()->getProperty('test');
+		$initializerExprTypeResolver = self::getContainer()->getByType(InitializerExprTypeResolver::class);
+		$defaultValueType = $initializerExprTypeResolver->getType(
+			$nativeProperty->getDefaultValueExpression(),
+			InitializerExprContext::fromClassReflection($class->getNativeProperty('test')->getDeclaringClass()),
+		);
+		$this->assertInstanceOf(ConstantIntegerType::class, $defaultValueType);
+		$this->assertSame(10, $defaultValueType->getValue());
+	}
+
+	public function testBug4288b(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-4288b.php');
+		$this->assertNoErrors($errors);
+
+		$reflectionProvider = $this->createReflectionProvider();
+		$class = $reflectionProvider->getClass(\Bug4288b\MyClass::class); // phpcs:ignore
 		$parameter = ParametersAcceptorSelector::selectSingle($class->getNativeMethod('paginate')->getVariants())->getParameters()[0];
 		$defaultValue = $parameter->getDefaultValue();
 		$this->assertInstanceOf(ConstantIntegerType::class, $defaultValue);
--- a/tests/PHPStan/Rules/Comparison/data/bug-4570.php
+++ b/tests/PHPStan/Rules/Comparison/data/bug-4570.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+namespace Bug4570;
+
+trait MyLogic
+{
+	public function fetchValue() : string {
+		if (array_key_exists('valueToFetch', self::MY_CONST_ARRAY)) {
+			return self::MY_CONST_ARRAY['valueToFetch'];
+		}
+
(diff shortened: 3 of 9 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -1621,12 +1621,18 @@ private function resolveType(string $exprString, Expr $node): Type
...
 				$node->name->name,
 				$this->isInClass() ? $this->getClassReflection() : null,
 				fn (Expr $expr): Type => $this->getType($expr),
 			);
+
+			if ($this->isInTrait() && !$constantType->isClassStringType()->yes()) {
+				return $constantType->generalize(GeneralizePrecision::lessSpecific());
+			}
+
+			return $constantType;
```
ondrejmirtes: No idea what this code is going to do, can you verify that with NodeScopeResolverTest?

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: the idea is, when analyzing in a trait we generalize constant values to their base type, e.g. `'hallo'` -> `string`.

that way we don't run into the always true/false errors

tests/PHPStan/Rules/Comparison/data/bug-4570.php
```diff
@@ -0,0 +1,52 @@
+<?php declare(strict_types = 1);
+
+namespace Bug4570;
+
+use function PHPStan\Testing\assertType;
+
+trait MyLogic
+{
+	public function fetchValue() : string {
+		assertType('string', self::MY_CONST);
```
ondrejmirtes: How can this assert worj when the FirstConsumer class has `MY_CONST = 1;`?

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Comparison/data/bug-4570.php
staabm: asserting types when using a trait from 2 classes is kind of hard to express. I therefore moved the type assertions into a separate case which only uses a single class

staabm:
just re-verified the tests added with this PR

src/Analyser/MutatingScope.php
```diff
@@ -1621,12 +1621,18 @@ private function resolveType(string $exprString, Expr $node): Type
...
 			}
-			return $this->initializerExprTypeResolver->getClassConstFetchTypeByReflection(
+			$constantType = $this->initializerExprTypeResolver->getClassConstFetchTypeByReflection(
 				$node->class,
 				$node->name->name,
 				$this->isInClass() ? $this->getClassReflection() : null,
 				fn (Expr $expr): Type => $this->getType($expr),
 			);
+
+			if ($this->isInTrait() && !$constantType->isLiteralString()->yes()) {
```
staabm: literals are not generalized, to make sure class-strings still work like expected.

see existing test at

https://github.com/phpstan/phpstan-src/blob/f4cfe25b25e6ee26d7227f20dc754e11d6de2ee4/tests/PHPStan/Analyser/data/type-aliases.php#L170

tests/PHPStan/Rules/Comparison/data/bug-4570.php
ondrejmirtes: Bu that kind of defeats the purpose of this fix. When one class has `const MY_CONST = 1;` and another one `const MY_CONST = 'foo';`, PHPStan is still going to complain.

src/Analyser/MutatingScope.php
ondrejmirtes: This is wrong assumption and a wrong fix. Sure, `Test::FOO` (including `::class`) should not change, but that has nothing to do with literal strings.

Instead, you need to look at the class the constant is fetched on. If it's the current class and we're in a trait, we essentially need to make it `mixed`... But if it's fetched from an external place, we can keep it as it is.

Of course, `self::class` doesn't need to be mixed, but should be kept as `class-string&literal-string`.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
```diff
@@ -1618,21 +1618,23 @@ private function resolveType(string $exprString, Expr $node): Type
 
 			return new ErrorType();
 		} elseif ($node instanceof Node\Expr\ClassConstFetch && $node->name instanceof Node\Identifier) {
+			if ($this->isInTrait()
+				&& $node->class instanceof Name
+				&& $node->class->toString() === 'self'
```
ondrejmirtes: That's not the only way how to reference the current class ;) You need to resolve the name and then compare it against the current class name.

Also test how self::class looks like.

Also - the correct place to do this is probably InitializerExprTypeResolver.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: > That's not the only way how to reference the current class ;) You need to resolve the name and then compare it against the current class name.

generalizing `static::class` to `'class-string&literal-string'` makes some existing tests fail unexpectedly

```
1) PHPStan\Rules\Comparison\ImpossibleCheckTypeFunctionCallRuleTest::testImpossibleCheckTypeFunctionCall
--- Expected
+++ Actual
@@ @@
 634: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'method' will always evaluate to true.
 637: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'someAnother' will always evaluate to true.
 640: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'unknown' will always evaluate to false.
-643: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'method' will always evaluate to true.
-646: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'someAnother' will always evaluate to true.
-649: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'unknown' will always evaluate to false.
+643: Call to function method_exists() with class-string&literal-string and 'method' will always evaluate to false.
+646: Call to function method_exists() with class-string&literal-string and 'someAnother' will always evaluate to false.
+649: Call to function method_exists() with class-string&literal-string and 'unknown' will always evaluate to false.
 678: Call to function is_string() with string will always evaluate to true.
     💡 Because the type is coming from a PHPDoc, you can turn off this check by setting <fg=cyan>treatPhpDocTypesAsCertain: false</> in your <fg=cyan>%configurationFile%</>.
 693: Call to function assert() with true will always evaluate to true.

C:\dvl\Workspace\phpstan-src-staabm\src\Testing\RuleTestCase.php:153
C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Rules\Comparison\ImpossibleCheckTypeFunctionCallRuleTest.php:48

2) PHPStan\Rules\Comparison\ImpossibleCheckTypeFunctionCallRuleTest::testImpossibleCheckTypeFunctionCallWithoutAlwaysTrue
--- Expected
+++ Actual
@@ @@
 598: Call to function method_exists() with 'UndefinedClass' and 'test' will always evaluate to false.
 631: Call to function method_exists() with $this(CheckTypeFunctionCall\MethodExistsWithTrait) and 'unknown' will always evaluate to false.
 640: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'unknown' will always evaluate to false.
-649: Call to function method_exists() with 'CheckTypeFunctionCall\\MethodExistsWithTrait' and 'unknown' will always evaluate to false.
+643: Call to function method_exists() with class-string&literal-string and 'method' will always evaluate to false.
+646: Call to function method_exists() with class-string&literal-string and 'someAnother' will always evaluate to false.
+649: Call to function method_exists() with class-string&literal-string and 'unknown' will always evaluate to false.
 694: Call to function assert() with false will always evaluate to false.
 694: Call to function is_numeric() with 'blabla' will always evaluate to false.
 '

C:\dvl\Workspace\phpstan-src-staabm\src\Testing\RuleTestCase.php:153
C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Rules\Comparison\ImpossibleCheckTypeFunctionCallRuleTest.php:260

3) PHPStan\Analyser\AnalyserIntegrationTest::testBug4288
Failed asserting that PHPStan\Type\MixedType Object (...) is an instance of class "PHPStan\Type\Constant\ConstantIntegerType".

C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Analyser\AnalyserIntegrationTest.php:373
```

ondrejmirtes: Looks like a bug, you need to check MethodExistsTypeSpecifyingExtension and also there's something about `method_exists` in ImpossibleCheckTypeHelper.

staabm: thank you.

with https://github.com/phpstan/phpstan-src/pull/2046/commits/4d6944791918ecc50ec66564ca7e43fe930c362b I have adjusted `MethodExistsTypeSpecifyingExtension` so it no longer makes the impossible type checker emit errors for `static::class` when used within traits. My reasoning is, since we generalized the class-string type when used within traits, we can no longer judge about always true/false method errors for this expression.

the remaining failing test in https://github.com/phpstan/phpstan-src/blob/e3bf63090b9da7d3e8ef8496eb8c5e7f1fa47da4/tests/PHPStan/Analyser/data/bug-4288.php#L21 is also related to the generalization. since we generalize the constant to `mixed`, we no longer can hold the expectations of this testcase.

I guess this is not accaptable.. but lets see what you think :)

staabm: > Looks like a bug, you need to check MethodExistsTypeSpecifyingExtension and also there's something about `method_exists` in ImpossibleCheckTypeHelper.

extracted into a separate issue: https://github.com/phpstan/phpstan/issues/8752

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -370,8 +371,7 @@ public function testBug4288(): void
 		$class = $reflectionProvider->getClass(MyClass::class);
 		$parameter = ParametersAcceptorSelector::selectSingle($class->getNativeMethod('paginate')->getVariants())->getParameters()[0];
 		$defaultValue = $parameter->getDefaultValue();
-		$this->assertInstanceOf(ConstantIntegerType::class, $defaultValue);
-		$this->assertSame(10, $defaultValue->getValue());
+		$this->assertInstanceOf(MixedType::class, $defaultValue);
```
staabm, line 371: this default value is now generalized, since it is referenced from within a trait via class-constant

staabm:
with 10ec187 I added a new test-variant for bug 4288, which shows that we can still use the concrete constant value as a parameter default value, when referencing the class-const by name, instead of with `self::` which is generalized and less precise with this PR

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -2,9 +2,6 @@
 
 namespace PHPStan\Analyser;
 
-use Bug4288\MyClass;
-use Bug4713\Service;
-use ExtendingKnownClassWithCheck\Foo;
```
staabm, line 7: a few tests relied on importing classes, which can easily backfire when adding new tests, with same named classes.

therefore I adjusted `AnalyzerIntegrationTest` to use FQCN everywhere - inspired by `TypeCombinatorTest` where we do simliar stuff.

staabm:
replaced by https://github.com/phpstan/phpstan-src/pull/2335 for 1.10.x

PR #2047, opened 2022-12-04
Integrate FidryCpuCoreCounter

theofidry:
See https://github.com/theofidry/cpu-core-counter/issues/11 for the motivations.

The code should be sensibly the same, the notable differences are:

- fixed the deprecated usage of `hw.ncpu`
- /proc/cpuinfo is check _last_
- nproc is checked first (before was not checked at all, it's considered to be more accurate and less convoluted than cpuinfo though)
- not sure about the `return 2`, but this was not too clear [here](https://github.com/phpstan/phpstan-src/pull/514#pullrequestreview-660926363) neither. I could otherwise change the fallback value to return `1` if `proc_open` doesn't exist and 2 otherwise

I would also wait on 1.x before merging this if you're happy with it. Technically the current version is fine, I'm just looking for a round of feedback before publishing 1.0.

### DIFF
--- a/src/Process/CpuCoreCounter.php
+++ b/src/Process/CpuCoreCounter.php
@@ -2,16 +2,8 @@
 
 namespace PHPStan\Process;
 
-use function count;
-use function fgets;
-use function file_get_contents;
-use function function_exists;
-use function is_file;
-use function is_resource;
-use function pclose;
-use function popen;
-use function preg_match_all;
-use const DIRECTORY_SEPARATOR;
+use Fidry\CpuCoreCounter\CpuCoreCounter as FidryCpuCoreCounter;
+use Fidry\CpuCoreCounter\NumberOfCpuCoreNotFound;
 
 class CpuCoreCounter
 {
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
One idea I have - it should differentiate between physical and virtual cores, offer different getters for that. That'd be really valuable. I want to use the virtual cores, no matter the operating system.

I understand that on systems with hyperthreading, `virtual cores = 2 * physical cores`. The CPU wouldn't be fully utilized if I used only the physical cores.

But of course it might be more complicated, I'm not a hardware expert.

ondrejmirtes:
And if PHPStan should be able to use this library, right now it needs to support PHP 7.2+.

theofidry:
@ondrejmirtes should be good; there is now a distinction possible between physical and logical cores. The default picked is logical (as PHP source does and as was done previously here).

If you fancy checking out what finder finds what, to find a bug or play around with the order or something, I can recommend [those commands](https://github.com/theofidry/cpu-core-counter#checks-what-finders-find-what-on-your-system)

ondrejmirtes:
I really like your work, the library looks nice!

But it looks like there's a related failure: https://github.com/phpstan/phpstan-src/actions/runs/3647730351/jobs/6160310175

src/Process/CpuCoreCounter.php
```diff
@@ -24,42 +16,11 @@ public function getNumberOfCpuCores(): int
 			return $this->count;
 		}
 
-		if (!function_exists('proc_open')) {
-			return $this->count = 1;
+		try {
+			return (new FidryCpuCoreCounter())->getCount();
```
ondrejmirtes: also the caching into `$this->count` should probably be kept here :)

ondrejmirtes (changes requested):

src/Process/CpuCoreCounter.php
theofidry: Missed that 🤦

theofidry:
Indeed and the failure is related I think. It's curious though as this is a piece of code coming from Infection and I thought I tested the case where the commands would fail... will look into it, but will probably have to rely on the CI here for that, at least partially

theofidry:
Nevermind it's windows related, I'll be able to check that on a machine at home soon :)

theofidry:
@ondrejmirtes green

ondrejmirtes:
Not really:

* https://github.com/phpstan/phpstan-src/actions/runs/3648030637/jobs/6161060710
* https://github.com/phpstan/phpstan-src/actions/runs/3648030605/jobs/6160966865 (still the same problem with Windows)

theofidry:
@ondrejmirtes FINALLY it is green

ondrejmirtes (approved):

ondrejmirtes:
Nice :) Please definitely be ready for a round of feedback once this is released, maybe we broke some scenario that isn't in the test suite, but that's fine, it's an "unknown unknown", so if we learn about it, the better :)

ondrejmirtes:
Thank you!

PR #2049, opened 2022-12-05
phpcs: use parallel mode

janedbal:
Ideally, available vcpus should be used...

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -7,6 +7,7 @@
 	<arg name="tab-width" value="4"/>
 	<arg name="cache" value="tmp/cache/phpcs"/>
 	<arg name="ignore" value="compiler/tests/*/data,tests/*/data,tests/*/traits,tests/notAutoloaded,tests/*/cache,src/Reflection/SignatureMap/functionMap.php,tests/e2e/magic-setter,tests/e2e/anon-class"/>
+	<arg name="parallel" value="120"/>
 	<arg value="sp"/>
 	<file>bin</file>
 	<file>src</file>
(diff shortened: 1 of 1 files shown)


phpcs.xml
```diff
@@ -7,6 +7,7 @@
 	<arg name="tab-width" value="4"/>
 	<arg name="cache" value="tmp/cache/phpcs"/>
 	<arg name="ignore" value="compiler/tests/*/data,tests/*/data,tests/*/traits,tests/notAutoloaded,tests/*/cache,src/Reflection/SignatureMap/functionMap.php,tests/e2e/magic-setter,tests/e2e/anon-class"/>
+	<arg name="parallel" value="120"/>
```
janedbal: This might be a little too high number :)

johnbillion:
What improvement does this give? And does GitHub Actions provide more than one core to run this in parallel during CI?

staabm:
github actions run with 2 cores on the default instance. whether its a measurable difference need to be proved though.

staabm:
@theofidry maybe phpcs should have some cpu core autodetection ;)

janedbal:
> What improvement does this give?

Minor CI & local speedup.

> maybe phpcs should have some cpu core autodetection ;)

Yeah, ideally. Current solution is really naive.

---- 

I dont really care about this PR much, but it feels wrong not to utilize faster solution :)

alies-dev:
@johnbillion 
 
>What improvement does this give? And does GitHub Actions provide more than one core to run this in parallel during CI?

GH provides 2, 4, 8, 16, 64 core runners, also supports custom runners: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

I see this PR very beneficial - I can remove some custom core that I wrote to utility all CPUs for different tools

PR #2050, opened 2022-12-05
Option usePathConstantsAsConstantString

MartinMystikJonas:


ondrejmirtes (changes requested):
The `new ConstantStringType($file)` part is repeated here. I'd like the code to change so that the last line call looks like this:

```php
return $this->usePathConstantsAsConstantString ? $stringType : $stringType->generalize(GeneralizePrecision::moreSpecific());
```

ondrejmirtes (approved):

ondrejmirtes:
Please remind me - why would need this again? It's causing problems with https://github.com/phpstan/phpstan/issues/12203.

thg2k:
@ondrejmirtes I think there is something explained here: https://github.com/phpstan/phpstan/discussions/8454#discussioncomment-4297664

MartinMystikJonas:
It was needed when we evaluated paths to latte tempaltes with expressions containing __DIR__. Type was resolved as generic string not COnstantString with evaluated value. So I implemented this to be able to turn on treating results of expressions with __DIR__ as constantString so it can be evaluated later. It seems like problem described in https://github.com/phpstan/phpstan/issues/12203 is basically same thing we encountered. So this is not causing that problem but is solution to similar problem.

ondrejmirtes:
I'd say this option shouldn't be needed. We should always strive for the type to be as precise as possible, and only degrade it (generalize it) in contexts where the precise type is counter-productive.

MartinMystikJonas:
Agreed.

PR #1492, opened 2022-07-01
Add support for @template-contravariant

autaut03:
Closes https://github.com/phpstan/phpstan/issues/3960

PR #2051, opened 2022-12-05
Support contravariant template types

jiripudil:
Closes phpstan/phpstan#3960.

Basically, I'm picking up where #1492 left off, with a couple more tests.

Once this lands, I'd like to follow up with a fix to the bug mentioned in the original PR (https://github.com/phpstan/phpstan-src/pull/1492#discussion_r912294565). The logs are no longer available for the relevant builds, but I understand it caused some turmoil, and should therefore be discussed separately.


ondrejmirtes (changes requested):
I'd like some "accepts" tests, meaning "does this parameter accept this argument"? For example in CallMethodsRuleTest, to see that the objects behave correctly :)

jiripudil:
I'm assuming you have something like this in mind? Otherwise, apart from a detail irrelevant to this case, `GenericObjectType::accepts()` shares the implementation with `isSuperTypeOf()` which is already covered by tests. But can't hurt to have an integration test on top of that :)

ondrejmirtes:
Yes, thank you :)

oprypkhantc:
Thank you for picking up on it @jiripudil :)

PR #2053, opened 2022-12-05
Fix node PHPDoc type hints

herndlm:
Those 3 methods are also called with properties that are determined more dynamically like e.g.
```php
foreach ($node->getSubNodeNames() as $subNodeName) {
    $subNode = $node->{$subNodeName};
    $this->processNodes($subNode, $nodeCallback, $endNodeCallback);
}
```
and `$subNode` can be `null` here for e.g. `$node->stmts`.

noticed this when trying to enforce native union type hints for params return types via phpcs. btw what is the reason this is not done yet? BC for public methods I suppose?

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4064,7 +4064,7 @@ class NodeScopeResolver
 	}
 
 	/**
-	 * @param Node[]|Node|scalar $node
+	 * @param Node[]|Node|scalar|null $node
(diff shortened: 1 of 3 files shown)


herndlm:
I do wonder though why PHPStan did not report this. Looks like the type is `mixed`, implicit mixed I suppose?

ondrejmirtes:
Thank you! :) Yeah, PHPStan is on level 8, and you can't expect any useful typechecking with `$node->{$subNodeName}` :)

PR #2056, opened 2022-12-06
Fix issue 8467 ShowCommand regression

rajyan:
In https://github.com/phpstan/phpstan-src/pull/1950

I replaced this specification
https://github.com/phpstan/phpstan-src/commit/46a8f32f5474b6f6127fb17a1638e7e11897be52#diff-dc817f2bab8672057a95375a542e68599164f6ab2380e4cccd1b50583e295d85R3514-R3544

with a conditional expression
https://github.com/phpstan/phpstan-src/blob/106cf1bc332242e94fce4bea25cea4cdc856e562/src/Analyser/NodeScopeResolver.php#L4021-L4024

which caused a regression for `non-empty-array` because of this line.
https://github.com/phpstan/phpstan-src/blob/024e98ecd612b3ca8a25cc1e612ed30d46d94202/src/Analyser/NodeScopeResolver.php#L3997

Thanks to `isConstantArray` and `getConstantArrays` we can handle this case correctly too.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -3992,16 +3992,18 @@ class NodeScopeResolver
 			}
 		}
 
+		$constantArrays = $iterateeType->getConstantArrays();
 		if (
 			$stmt->getDocComment() === null
-			&& $iterateeType instanceof ConstantArrayType
+			&& $iterateeType->isConstantArray()->yes()
+			&& count($constantArrays) === 1
 			&& $stmt->valueVar instanceof Variable && is_string($stmt->valueVar->name)
 			&& $stmt->keyVar instanceof Variable && is_string($stmt->keyVar->name)
 		) {
 			$valueConditionalHolders = [];
 			$arrayDimFetchConditionalHolders = [];
-			foreach ($iterateeType->getKeyTypes() as $i => $keyType) {
-				$valueType = $iterateeType->getValueTypes()[$i];
+			foreach ($constantArrays[0]->getKeyTypes() as $i => $keyType) {
+				$valueType = $constantArrays[0]->getValueTypes()[$i];
 				$holder = new ConditionalExpressionHolder([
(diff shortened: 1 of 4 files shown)


herndlm:
Nice to see that those Type functions are useful :)

ondrejmirtes (changes requested):
I think that needs kinds of problems benefit more from NodeScopeResolverTest and not rule-specific tests. But feel free to have both 😊

And those Type methods are 😍

rajyan:
> I think that needs kinds of problems benefit more from NodeScopeResolverTest and not rule-specific tests. But feel free to have both blush

Indeed. Added some assertTypes:+1:

ondrejmirtes:
Thank you very much!

src/Analyser/NodeScopeResolver.php
```diff
@@ -3992,16 +3992,18 @@ static function (): void {
...
+			&& count($constantArrays) === 1
 			&& $stmt->valueVar instanceof Variable && is_string($stmt->valueVar->name)
 			&& $stmt->keyVar instanceof Variable && is_string($stmt->keyVar->name)
 		) {
 			$valueConditionalHolders = [];
 			$arrayDimFetchConditionalHolders = [];
-			foreach ($iterateeType->getKeyTypes() as $i => $keyType) {
-				$valueType = $iterateeType->getValueTypes()[$i];
+			foreach ($constantArrays[0]->getKeyTypes() as $i => $keyType) {
+				$valueType = $constantArrays[0]->getValueTypes()[$i];
```
staabm, line 4006: do you think it might make sense to move `$constantArrays[0]->getValueTypes()` before the loop?

rajyan: https://github.com/phpstan/phpstan-src/blob/1.9.x/src/Type/Constant/ConstantArrayType.php#L248
It’s just a getter, I don’t think it benefits that much for performance.

PR #2058, opened 2022-12-06
Do not merge `BenevolentUnionType` with `UnionType`

herndlm:
This was inspired by https://github.com/phpstan/phpstan/issues/8458#issuecomment-1336402081

Closes https://github.com/phpstan/phpstan/issues/7049
Closes https://github.com/phpstan/phpstan/issues/7279
Closes https://github.com/phpstan/phpstan/issues/7423
Closes https://github.com/phpstan/phpstan/issues/8268

After this https://phpstan.org/r/498cd1d0-4568-4cbf-8b0b-11ed722552e8 becomes green.

#### Regarding the performance impact (https://github.com/phpstan/phpstan-src/pull/2058#issuecomment-1340500218)
before:
```
Benchmark 1: make phpstan
  Time (mean ± σ):     56.978 s ± 10.578 s    [User: 338.941 s, System: 26.704 s]
  Range (min … max):   49.723 s … 82.109 s    10 runs
```
after:
```
Benchmark 1: make phpstan
  Time (mean ± σ):     51.790 s ±  5.459 s    [User: 330.922 s, System: 25.508 s]
  Range (min … max):   48.548 s … 66.973 s    10 runs
```
not sure if that is enough, the system is a bit trashy I guess. but looks like things don't really change perf.-wise. I'm pretty sure things also don't get faster, the variance is just too big on my system, I think we can go with 49s before and after.

#### Regarding the 3 failures
- PocketMine: `Cannot access offset (float|int<1, max>) on mixed.` https://github.com/pmmp/PocketMine-MP/blob/e0b07ff3087b652407439a29c941f3b66ca92c86/src/pocketmine/CrashDump.php#L314 / https://github.com/pmmp/PocketMine-MP/blob/e0b07ff3087b652407439a29c941f3b66ca92c86/src/pocketmine/CrashDump.php#L111
- Prestashop: modification of a `ArrayObject<*NEVER*, *NEVER*>` https://github.com/PrestaShop/PrestaShop/blob/ca9d81aae0bb17f3e767bb5d835348ed144a3ab5/src/Adapter/Presenter/AbstractLazyArray.php#L98
- PHPUnit: modification of a `ArrayObject<*NEVER*, *NEVER*>`, e.g. https://github.com/sebastianbergmann/phpunit/blob/93d4bf4c37aec6384bb9e5d390d9049a463a7256/tests/unit/Framework/AssertTest.php#L153

all of those are expected IMO, but I can create PRs for Prestashop and PHPUnit to add PHPDoc types to the collection if wanted

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -145,25 +145,12 @@ final class TypeCombinator
 			return new NeverType();
 		}
 
-		$benevolentTypes = [];
-		$benevolentUnionObject = null;
 		// transform A | (B | C) to A | B | C
 		for ($i = 0; $i < $typesCount; $i++) {
-			if ($types[$i] instanceof BenevolentUnionType) {
-				if ($types[$i] instanceof TemplateBenevolentUnionType && $benevolentUnionObject === null) {
-					$benevolentUnionObject = $types[$i];
-				}
-				$benevolentTypesCount = 0;
-				$typesInner = $types[$i]->getTypes();
-				foreach ($typesInner as $benevolentInnerType) {
-					$benevolentTypesCount++;
-					$benevolentTypes[$benevolentInnerType->describe(VerbosityLevel::value())] = $benevolentInnerType;
-				}
-				array_splice($types, $i, 1, $typesInner);
-				$typesCount += $benevolentTypesCount - 1;
+			if (!($types[$i] instanceof UnionType)) {
 				continue;
 			}
-			if (!($types[$i] instanceof UnionType)) {
+			if ($types[$i] instanceof BenevolentUnionType) {
 				continue;
 			}
 			if ($types[$i] instanceof TemplateType) {
@@ -345,25 +332,6 @@ final class TypeCombinator
 		}
 		if ($typesCount === 1) {
 			return $types[0];
-		}
-
-		if ($benevolentTypes !== []) {
-			$tempTypes = $types;
-			foreach ($tempTypes as $i => $type) {
-				if (!isset($benevolentTypes[$type->describe(VerbosityLevel::value())])) {
-					break;
-				}
-
-				unset($tempTypes[$i]);
-			}
-
-			if ($tempTypes === []) {
-				if ($benevolentUnionObject instanceof TemplateBenevolentUnionType) {
-					return $benevolentUnionObject->withTypes($types);
-				}
-
-				return new BenevolentUnionType($types);
-			}
 		}
 
 		return new UnionType($types, true);
--- a/src/Type/UnionType.php
+++ b/src/Type/UnionType.php
@@ -70,6 +70,9 @@ class UnionType implements CompoundType
 			if (!($type instanceof UnionType)) {
 				continue;
 			}
+			if ($type instanceof BenevolentUnionType) {
+				continue;
+			}
 			if ($type instanceof TemplateType) {
 				continue;
 			}
--- a/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
@@ -4942,11 +4942,11 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 				'array_search(new stdClass, $generalStringKeys, true)',
 			],
 			[
-				'int|string|false',
+				'(int|string)|false',
 				'array_search($mixed, $array, true)',
 			],
 			[
-				'int|string|false',
+				'(int|string)|false',
 				'array_search($mixed, $array, false)',
 			],
 			[
@@ -5006,15 +5006,15 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 				'array_search(\'id\', doFoo() ? $thisDoesNotExistAndIsMixedInUnion : false, true)',
 			],
 			[
-				'int|string|false',
+				'(int|string)|false',
 				'array_search(1, $generalIntegers, true)',
 			],
 			[
-				'int|string|false',
+				'(int|string)|false',
 				'array_search(1, $generalIntegers, false)',
 			],
 			[
-				'int|string|false',
+				'(int|string)|false',
 				'array_search(1, $generalIntegers)',
 			],
 			[
@@ -8820,11 +8820,11 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 				'json_decode($mixed, false, 512, $integer | JSON_THROW_ON_ERROR | JSON_NUMERIC_CHECK)',
 			],
 			[
-				'int|string|null',
+				'(int|string)|null',
 				'array_key_first($mixedArray)',
 			],
 			[
-				'int|string|null',
+				'(int|string)|null',
 				'array_key_last($mixedArray)',
 			],
 			[
@@ -8904,7 +8904,7 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 				'$hrtime3',
 			],
 			[
-				'array{int, int}|float|int',
+				'(float|int)|array{int, int}',
 				'$hrtime4',
 			],
 		];
--- a/tests/PHPStan/Analyser/nsrt/array-search-php7.php
+++ b/tests/PHPStan/Analyser/nsrt/array-search-php7.php
@@ -12,9 +12,9 @@ class Foo
 	public function mixedAndSubtractedArray($mixed, string $string): void
 	{
 		if (is_array($mixed)) {
-			assertType('int|string|false', array_search('foo', $mixed, true));
-			assertType('int|string|false', array_search('foo', $mixed));
-			assertType('int|string|false', array_search($string, $mixed, true));
+			assertType('(int|string)|false', array_search('foo', $mixed, true));
+			assertType('(int|string)|false', array_search('foo', $mixed));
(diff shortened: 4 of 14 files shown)


herndlm:
OK close, better than I thought. I guess I deleted too much and am loosing the benevolent template object..

staabm:
You should measure whether these changes have a perf impact

VincentLanglet:
Since `A|(B|C)` would be now a possible type, does it require a test to check that a benevolent union
`(A|(B|C)|D)` is simplified to `(A|B|C|D)` ?

herndlm:
Good question, I'm not sure if it should be tbh. If we explicitly want a union, shouldn't it be non-benevolent? 🤔 I guess what I'm saying is, are we expecting this to be simplified if we call `TypeCombinator::union()`?
UPDATE: ah sorry, got your point now. Those should be simplified I suppose, yeah

tests/PHPStan/Analyser/data/bug-7279.php
```diff
@@ -0,0 +1,87 @@
...
+ * @var array{} $emptyMap
+ * @var array<int, string> $unknownList
+ * @var array{id?: int, name?: string} $unknownMap
+ * @var non-empty-array<int, Timeline<Percentage>> $nonEmptyList
+ * @var array{work: Timeline<Percentage>} $nonEmptyMap
+ */
+
+// Everything works great for find()
+
+assertType('null', find([], $callback));
```
staabm: Should be wrapped into a method/function

tests/PHPStan/Analyser/data/bug-7279.php
```diff
@@ -0,0 +1,87 @@
...
+
+	assertType('null', findKey([], $callback));
+	assertType('null', findKey($emptyList, $callback));
+	assertType('null', findKey($emptyMap, $callback));
+
+	assertType('int|null', findKey($unknownList, $callback));
+	assertType('string|null', findKey($unknownMap, $callback));
+
+	assertType('int|null', findKey($nonEmptyList, $callback));
+	assertType('string|null', findKey($nonEmptyMap, $callback));
```
herndlm: both L83 and L86 are different to the snippet where `'id'|'name'|null` was asserted/expected. but this is either not supported or something different than the main problem that was fixed here IMO

ondrejmirtes:
I'm struggling a bit with this one. What is your thinking behind this PR? I see that it fixes a bunch of issues, but to me it looks like it's just a matter of luck, not an intention (please take no offense :)).

What's a typical situation that this PR improves/fixes?

Also, I think we need some more tests around this. Let's say for example we have `(int|string)` and we want to add some types into it by calling TypeCombinator::union. What's the expected result?

* When we add a `ConstantStringType` like `'1'`.
* When we add a `StringType`.
* When we add ConstantBooleanType like `true`.

herndlm:
The main intention behind this is that I believe, and I think the linked issues confirm it, that union with a benevolent union should not suddenly get rid of the benevolence. E.g.  https://phpstan.org/r/498cd1d0-4568-4cbf-8b0b-11ed722552e8

Calling `TypeCombinator::union()` will always add a real union and not get rid of or extend the benevolence. More tests make sense for sure :) some already existed though I believe, and I just adapted their result. Your mentioned cases are good ones that worry me that the edges are still rough here hehe

But it's basically what you suggested at https://github.com/phpstan/phpstan/issues/8268#issuecomment-1301834601 (which I just found later via issue bot).

herndlm:
Added the test cases for union with `(int|string)` you requested, we get the following

> When we add a `ConstantStringType` like `'1'`.

`(int|string)`


> When we add a `StringType`.

`(int|string)`


>  When we add ConstantBooleanType like `true`.

`(int|string)|true`

that looks OK to me 😊

ondrejmirtes:
I'd argue that `TypeCombinator::union` of `(int|string)` and `string` should result in non-benevolent union type.

Another idea for a test - if we have `(1|2)` and `string`, it should result in `string` (which it probably will).

Also we can test what happens with `(1|2|3)` and `int<2, 3>`.

Also `(InvalidArgumentException|LengthException|stdClass)` and `LogicException`.

herndlm:
Ok, added your mentioned cases again, the result is a bit different, I got
- `(1|2)|string` UPDATE: but `('1'|'2')|string` does result in `string`, not sure if I misunderstood that or not, anyway, also added it as a test
- `(1|2|3)`
- `(InvalidArgumentException|LengthException|stdClass)|LogicException`

so most likely there's still work needed / it can be improved? I pushed them in this state, but need to think about this another time. Such benevolent cases are still hard for me to think through properly

VincentLanglet:
> What's a typical situation that this PR improves/fixes?

https://phpstan.org/r/9be20c04-429f-4dc0-b245-9d23628ae2db

Both `$key` and `$bool` are accepted by the method, but the union of both type is not accepted.
The current behavior doesn't feel natural.

> I'd argue that TypeCombinator::union of (int|string) and string should result in non-benevolent union type.

Technically, it should be `(int)|string` which is not really a possible type currently.

Because if it's transformed into `int|string`
```
		$key = current(array_keys($array));
		$this->acceptString($key); // valid
		$this->acceptString($string); // valid
		
		$keyOrString = rand(0,1) ? $key : $string;
		$this->acceptString($keyOrString); // won't be but should still be valid
```

drupol:
Hello @herndlm,

I don't think this is closing [#742](https://github.com/phpstan/phpstan/issues/7423), the output is different for sure, but it's not fixed actually.

herndlm:
@drupol are you referring to the bot message you got in your issue (which is unrelated to this PR) or https://github.com/phpstan/phpstan-src/actions/runs/3910108845? :)

drupol:
> @drupol are you referring to the bot message you got in your issue (which is unrelated to this PR) or https://github.com/phpstan/phpstan-src/actions/runs/3910108845? :)

Sorry, here's the proper link: https://github.com/phpstan/phpstan/issues/7423 I also fixed it in my previous message.

herndlm:
> Sorry, here's the proper link: [phpstan/phpstan#7423](https://github.com/phpstan/phpstan/issues/7423) I also fixed it in my previous message.

yeah and why do you mean this PR does not fix it? the bot message there is unrelated. but in https://github.com/phpstan/phpstan-src/actions/runs/3910108845 I can see no errors for PHP 8+ because the type should be correct now? and that is also checked via regression test here.

drupol:
Oh if it's unrelated, please ignore my message then.

Sorry for the noise.

herndlm:
Did we come to a conclusion about this one? :) I rebased it on 1.10.x in the meantime. Thought the behavior change might be better there. In case it gets merged. So far it still makes sense to me

herndlm:
Going to rebase this one tomorrow morning. Is there anything else I can do here? This still makes sense to me :)

ondrejmirtes:
@herndlm I have some issues with it. For example I've seen a type `(X)|Y` mentioned somewhere and I don't think that one-item union type makes sense... So I'm not sure how you're handling it. Last time I looked at this PR it felt a bit undertested.

For example if you compare `int|stdClass|string` and `(int|string)|stdClass` - when is the first one going to be accepted and the second one won't?

herndlm:
Naming this PR "Do not merge" was a stupid idea, that way it will never get merged... 😂

jokes aside, I try to get this updated and tests added soon. would be great for 2.x I suppose.

ondrejmirtes:
I'm sorry, I don't see this being merged :/

herndlm:
> I'm sorry, I don't see this being merged :/

that's ok, no worries. is there anything I can do to change that or is it mostly based on experience/gut feeling (don't meant that in a bad way)? To me this still makes sense and kind of is the only way IMO to deal with that the linked issues show, also https://github.com/phpstan/phpstan-src/pull/2058#issuecomment-1376493368 and the example there.

but maybe there are edge cases I'm just not seeing yet and already have in mind or are worried about :)

ondrejmirtes:
I don't want to screw around https://phpstan.org/developing-extensions/type-system#type-normalization, and keeping benevolent union type separate from non-benevolent union type does that.

herndlm:
> I don't want to screw around [phpstan.org/developing-extensions/type-system#type-normalization](https://phpstan.org/developing-extensions/type-system#type-normalization), and keeping benevolent union type separate from non-benevolent union type does that.

sorry, don't want to be too annoying here, but right now they are merged and not merging them keeps them more separate, no? just want to fully understand what might cause issues here.

ondrejmirtes:
I don't want to keep them separate.

herndlm:
ok, thx! I'll close this for now, not sure atm, but maybe there's a better way of dealing with the linked issues 😊

herndlm:
what about, yeah I know, I can't leave it alone 😂, the other way around? merging more aggressively: `(int|string)|float` => `(float|int|string)`. my current reasoning: if benevolence is involved type checks are less strict and maybe this should be kept like that if additional types are added as in the (outdated) original example of https://phpstan.org/r/498cd1d0-4568-4cbf-8b0b-11ed722552e8

if stricter type checks are needed `checkBenevolentUnionTypes` could be enabled. obviously, I don't know yet what exact impacts this will have, less errors I suppose, so nobody is annoyed. there is of course also https://github.com/phpstan/phpstan/issues/8458#issuecomment-1336402081 which sounds like it would miss more potential runtime errors then I suppose :(

UPDATE: the more I think about it, the less I like it. Being less strict sounds bad :/ can we deprecate BenevolentUnionType? 😂

PR #2059, opened 2022-12-07
Fix issue 8467 intersect resolved conditional types

rajyan:
I thought about this once in https://github.com/phpstan/phpstan-src/pull/2030/commits/42bddf0262d38dccb34aa9670335956ed6047a15 but wasn't sure when this is needed.

The composer integration test failure was a very good example.
The conditional type knows `$initialClone=true => $origCwd=string|null`, but `isset($originCwd)` has already specified `$originCwd=>'string'` in the `if ($initialClone && isset($origCwd))` truthy scope. Conditional type can be resolved in the scope too, so it overwrites the `$originCwd`'s type as `$origCwd=string|null`.

Before my changes (in 1.9.2) resolving conditional types for specified types were skipped, but I think it is better to intersect, because the conditional type can be narrower than the specified type.

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-8467b.php
+++ b/tests/PHPStan/Analyser/data/bug-8467b.php
@@ -1,0 +1,56 @@
+<?php declare(strict_types = 1);
+
+namespace Bug8467b;
+
+use function PHPStan\Testing\assertType;
+
+class Test {
+	public function foo (?string $cwd, bool $initialClone = false): void {
+		if ($initialClone) {
+			$origCwd = $cwd;
+			$cwd = null;
+		}
+
+		if ($initialClone && isset($origCwd)) {
+			assertType('string', $origCwd);
+			assertType('null', $cwd);
+		}
+	}
+
+	/**
+	 * @param mixed[]|null $mirrors
+	 *
+	 * @return list<non-empty-string>
+	 *
+	 * @phpstan-param list<array{url: non-empty-string, preferred: bool}>|null $mirrors
+	 */
+	protected function getUrls(?string $url, ?array $mirrors, ?string $ref, ?string $type, string $urlType): array
+	{
+		if (!$url) {
+			return [];
+		}
+
+		if ($urlType === 'dist' && false !== strpos($url, '%')) {
+			assertType('string|null', $type);
+			$url = 'test';
+		}
+		assertType('non-falsy-string', $url);
+
+		$urls = [$url];
+		if ($mirrors) {
+			foreach ($mirrors as $mirror) {
+				if ($urlType === 'dist') {
+					assertType('string|null', $type);
+				} elseif ($urlType === 'source' && $type === 'git') {
+					assertType("'git'", $type);
+				} elseif ($urlType === 'source' && $type === 'hg') {
+					assertType("'hg'", $type);
+				} else {
+					continue;
+				}
+			}
+		}
+
+		return $urls;
+	}
+}
--- a/tests/PHPStan/Analyser/data/bug-8467b.php
+++ b/tests/PHPStan/Analyser/data/bug-8467b.php
@@ -13,7 +13,7 @@ class Test {
 
 		if ($initialClone && isset($origCwd)) {
 			assertType('string', $origCwd);
-			assertType('null', $cwd);
+			assertType('string|null', $cwd); // could be null
 		}
 	}
 
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -119,7 +119,6 @@ use function array_keys;
 use function array_map;
 use function array_merge;
 use function array_pop;
-use function array_reverse;
 use function array_slice;
 use function count;
 use function explode;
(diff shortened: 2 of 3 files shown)


tests/PHPStan/Analyser/data/bug-8467b.php
```diff
@@ -0,0 +1,56 @@
...
+class Test {
+	public function foo (?string $cwd, bool $initialClone = false): void {
+		if ($initialClone) {
+			$origCwd = $cwd;
+			$cwd = null;
+		}
+
+		if ($initialClone && isset($origCwd)) {
+			assertType('string', $origCwd);
+			assertType('null', $cwd);
```
rajyan: I believe this one is not covered in `createConditionalExpressions` yet and not working before 1.9.2 too.
I think it might be better to solve it in a different PR

rajyan:
The first error of larastan test failure looks like an improvement, but not sure with the second one.

rajyan:
https://github.com/phpstan/phpstan-src/actions/runs/3635043616/jobs/6133787739

Tested in the playground and I believe both larastan test failure are improvements.
https://laravelplayground.com/#/snippets/45e88341-c412-4c26-b3eb-5956890561aa

First one
https://github.com/nunomaduro/larastan/blob/325d2f377f17d99529ea76833479e6ab6d6ff64c/tests/Type/data/collection-stubs.php#L33
we know that the items is  `<int, App\User>`

second one
https://github.com/nunomaduro/larastan/blob/325d2f377f17d99529ea76833479e6ab6d6ff64c/tests/Type/data/collection-generic-static-methods.php#L171
the result is always an empty items because the callback is invalid, that means TKey cannot be resolved, and should be array-key (int|string).

Not sure enough with the second test. Maybe @canvural knows something?

rajyan:
The test is failing in https://github.com/phpstan/phpstan-src/pull/2058 too, it may have nothing to do with this change anyway.

ondrejmirtes:
Thank you 😊

canvural:
> [phpstan/phpstan-src/actions/runs/3635043616/jobs/6133787739](https://github.com/phpstan/phpstan-src/actions/runs/3635043616/jobs/6133787739)
> 
> Tested in the playground and I believe both larastan test failure are improvements. [laravelplayground.com/#/snippets/45e88341-c412-4c26-b3eb-5956890561aa](https://laravelplayground.com/#/snippets/45e88341-c412-4c26-b3eb-5956890561aa)
> 
> First one [nunomaduro/larastan@`325d2f3`/tests/Type/data/collection-stubs.php#L33](https://github.com/nunomaduro/larastan/blob/325d2f377f17d99529ea76833479e6ab6d6ff64c/tests/Type/data/collection-stubs.php#L33) we know that the items is `<int, App\User>`
> 
> second one [nunomaduro/larastan@`325d2f3`/tests/Type/data/collection-generic-static-methods.php#L171](https://github.com/nunomaduro/larastan/blob/325d2f377f17d99529ea76833479e6ab6d6ff64c/tests/Type/data/collection-generic-static-methods.php#L171) the result is always an empty items because the callback is invalid, that means TKey cannot be resolved, and should be array-key (int|string).
> 
> Not sure enough with the second test. Maybe @canvural knows something?



> [phpstan/phpstan-src/actions/runs/3635043616/jobs/6133787739](https://github.com/phpstan/phpstan-src/actions/runs/3635043616/jobs/6133787739)
> 
> Tested in the playground and I believe both larastan test failure are improvements. [laravelplayground.com/#/snippets/45e88341-c412-4c26-b3eb-5956890561aa](https://laravelplayground.com/#/snippets/45e88341-c412-4c26-b3eb-5956890561aa)
> 
> First one [nunomaduro/larastan@`325d2f3`/tests/Type/data/collection-stubs.php#L33](https://github.com/nunomaduro/larastan/blob/325d2f377f17d99529ea76833479e6ab6d6ff64c/tests/Type/data/collection-stubs.php#L33) we know that the items is `<int, App\User>`
> 
> second one [nunomaduro/larastan@`325d2f3`/tests/Type/data/collection-generic-static-methods.php#L171](https://github.com/nunomaduro/larastan/blob/325d2f377f17d99529ea76833479e6ab6d6ff64c/tests/Type/data/collection-generic-static-methods.php#L171) the result is always an empty items because the callback is invalid, that means TKey cannot be resolved, and should be array-key (int|string).
> 
> Not sure enough with the second test. Maybe @canvural knows something?

I think this was caused because of a change in Laravel, not by this change. Because Larastan's own CI is also failing. So I'll fix this there 👍🏽

ondrejmirtes:
Yeah, I'll update the commit on our side once you fix it, thank you :)

rajyan:
Thank you!

canvural:
@ondrejmirtes https://github.com/nunomaduro/larastan/commit/1a11021385b8d79dfb5f5c5a24ba1e82521a7fac Larastan's CI is green now.

ondrejmirtes:
PHPStan is green too 🎉 https://github.com/phpstan/phpstan/actions/runs/3649224831/jobs/6163628972

PR #2060, opened 2022-12-07
Notify in issuebot github summary results if there are changes

rajyan:


rajyan:
Works, but github doesn't notify mentions in step result...

staabm:
Maybe it would make sense to update the PR description or add a new comment with a link to the issue bot github summary page via the running github action job?

Or maybe even just post (and update) the results in a bots' comment instead of the summary...?

I guess/assume the idea woth thos initial PR was, to make it easier to access the isdue bot results?

ondrejmirtes:
I have this thought-out, I just need to get from under the pile of issues and pull requests of three weeks when I was not able to work 😊

rajyan:
> I guess/assume the idea woth thos initial PR was, to make it easier to access the isdue bot results?

Yes!

> Or maybe even just post (and update) the results in a bots' comment instead of the summary...?

I would prefer this one:+1:
I like it better to delete the previous comment and add a new comment for each push so the issue bot change is notified without messing the PR.

rajyan:
sorry miss clicked "Reopen and comment"

rajyan:
> I have this thought-out, I just need to get from under the pile of issues and pull requests of three weeks when I was not able to work blush

I'm looking forward to it:smile:

PR #2062, opened 2022-12-07
Add `deprecationRulesInstalled` configuration

villfa:
Follow up on https://github.com/phpstan/phpstan-phpunit/pull/150#issuecomment-1341164151:
This new configuration indicates whether [phpstan-deprecation-rules](https://github.com/phpstan/phpstan-deprecation-rules) is installed.


ondrejmirtes:
Thank you!

PR #2065, opened 2022-12-08
ArrayFilterRule should care bout treatPhpDocTypesAsCertain

VincentLanglet:
Hi @ondrejmirtes, 

When using `treatPhpDocTypesAsCertain: false`, a call `array_filter` shouldn't be reported as useless if the type is coming from phpdoc.

So 
```
if ($this->treatPhpDocTypesAsCertain) {
     $arrayType = $scope->getType($args[0]->value);
} else {
     $arrayType = $scope->getNativeType($args[0]->value);
}
```
should be used instead of `$arrayType = $scope->getType($args[0]->value);`

But I don't know if there is something to do in order to not forgot this for others rules,
like changing `getType` to return the nativeType if treatPhpDocTypesAsCertain is false or exposing another method. 
Maybe you would have an idea @staabm, @herndlm or @rajyan ?

### DIFF
--- a/conf/config.level5.neon
+++ b/conf/config.level5.neon
@@ -23,3 +23,5 @@ services:
 
 	-
 		class: PHPStan\Rules\Functions\ArrayFilterRule
+		arguments:
+			treatPhpDocTypesAsCertain: %treatPhpDocTypesAsCertain%
--- a/src/Rules/Functions/ArrayFilterRule.php
+++ b/src/Rules/Functions/ArrayFilterRule.php
@@ -20,7 +20,10 @@ use function strtolower;
 class ArrayFilterRule implements Rule
 {
 
-	public function __construct(private ReflectionProvider $reflectionProvider)
+	public function __construct(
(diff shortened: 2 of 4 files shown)


staabm:
as I understood it, `treatPhpDocTypesAsCertain` is missing in a lot of places and the current in-progress refactoring will help us to concentrate the `if (treatPhpDocTypesAsCertain)`-logic into the `Scope` class, so we don't need to sprinkle it all over the rules.

ondrejmirtes:
@staabm That's a wrong understanding of the current work in progress. Rules will specifically need to ask about getNativeType as opposed to getType with the `treat` parameter set to false.

ondrejmirtes (changes requested):
This diff doesn't make it obvious which errors are no longer reported. Please create a new file specific to your issue and test it with both true and false. Thanks. Otherwise 👍

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2067, opened 2022-12-09
Report dead types even in multi-exception catch

janedbal:
Closes https://github.com/phpstan/phpstan/issues/7312


src/Analyser/NodeScopeResolver.php
```diff
@@ -1301,6 +1322,9 @@ private function processStmtNode(
 						$catchScope = $catchScope->mergeWith($matchingThrowPoint->getScope());
 					}
 				}
+				if ($catchScope === null) {
+					$catchScope = $branchScope;
```
janedbal, line 1287: This means there is no matching throw point, maybe we should rather continue?

phpstan-baseline.neon
```diff
@@ -208,7 +208,7 @@ parameters:
 
 		-
 			message: "#^Dead catch \\- PHPStan\\\\BetterReflection\\\\Identifier\\\\Exception\\\\InvalidIdentifierName is never thrown in the try block\\.$#"
-			count: 2
+			count: 3
```
janedbal, line 254: Not sure why it was ignored in this file so I added it here, but it looks reported correctly.

src/Rules/Exceptions/CatchWithUnthrownExceptionRule.php
```diff
@@ -34,7 +34,7 @@ public function processNode(Node $node, Scope $scope): array
 
 		return [
 			RuleErrorBuilder::message(
-				sprintf('Dead catch - %s is never thrown in the try block.', $node->getCaughtType()->describe(VerbosityLevel::typeOnly())),
```
janedbal: I kept the original getter to maintain BC.

conf/config.neon
```diff
@@ -61,6 +61,7 @@ parameters:
 		closureDefaultParameterTypeRule: false
 		newRuleLevelHelper: false
 		instanceofType: false
+		detectDeadTypeInMultiCatch: true
```
ondrejmirtes, line 64: Should be false here

src/Node/CatchWithUnthrownExceptionNode.php
```diff
@@ -10,7 +10,7 @@
 class CatchWithUnthrownExceptionNode extends NodeAbstract implements VirtualNode
 {
 
-	public function __construct(private Catch_ $originalNode, private Type $caughtType, private Type $originalCaughtType)
+	public function __construct(private Catch_ $originalNode, private Type $unthrownType, private Type $originalCaughtType)
```
ondrejmirtes, line 13: We don't need to rename the parameter

tests/PHPStan/Rules/Exceptions/CatchWithUnthrownExceptionRuleTest.php
```diff
@@ -12,11 +13,21 @@
...
 	protected function getRule(): Rule
 	{
 		return new CatchWithUnthrownExceptionRule();
 	}
 
+	public static function getAdditionalConfigFiles(): array
+	{
+		return array_merge(
+			parent::getAdditionalConfigFiles(),
+			!self::$detectDeadMultiCatch ? [__DIR__ . '/disable-detect-multi-catch.neon'] : [],
```
ondrejmirtes, line 27: Should be two separate test cases instead of switching the property. I haven't tested if this works, and the convention is to have two separate test cases... You can't be sure when `getAdditionalConfigFiles` will be called first and thus it can lead to unexpected behaviour.

ondrejmirtes (changes requested):

janedbal:
Solved by https://github.com/phpstan/phpstan-src/pull/2399

PR #2068, opened 2022-12-09
Optimize path related utils on windows

staabm:
the changes mainly prevent excessive use of PCRE on windows. this leads to a small performance improvement of ~3%. 

more suprising is blackfire reporting a 56% improvement on memory usage (830 MB instead of 1.9 GB).

grafik

tested on windows 11
```
php -v
PHP 8.1.2 (cli) (built: Jan 19 2022 10:13:52) (NTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.1.2, Copyright (c) Zend Technologies
    with blackfire v1.76.0~win-x64-non_zts81, https://blackfire.io, by Blackfire
```

### DIFF
--- a/src/File/FileHelper.php
+++ b/src/File/FileHelper.php
@@ -59,7 +59,7 @@ class FileHelper
 		if ($originalPath !== '') {
 			if ($originalPath[0] === '/') {
 				$isLocalPath = true;
-			} elseif (strlen($originalPath) >= 3 && in_array($originalPath[0], range('A', 'Z'), true) && $originalPath[1] === ':' && $originalPath[2] === '\\') {
+			} elseif (strlen($originalPath) >= 3 && $originalPath[1] === ':' && $originalPath[2] === '\\') { // e.g. C:\\
 				$isLocalPath = true;
 			}
 		}
--- a/src/File/FileHelper.php
+++ b/src/File/FileHelper.php
@@ -10,6 +10,7 @@ use function ltrim;
 use function rtrim;
 use function str_replace;
 use function str_starts_with;
+use function strlen;
 use function strpos;
 use function substr;
 use function trim;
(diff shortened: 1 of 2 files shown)


src/Type/FileTypeMapper.php
```diff
@@ -88,6 +84,10 @@ public function getResolvedPhpDoc(
 			return ResolvedPhpDocBlock::createEmpty();
 		}
 
+		if ($fileName !== null) {
```
staabm, line 87: moved this normalization below the early return, to prevent unnecessary work

src/File/FileHelper.php
```diff
@@ -52,7 +53,14 @@ public function absolutizePath(string $path): string
 	/** @api */
 	public function normalizePath(string $originalPath, string $directorySeparator = DIRECTORY_SEPARATOR): string
 	{
-		$isLocalPath = $originalPath !== '' && $originalPath[0] === '/';
+		$isLocalPath = false;
+		if ($originalPath !== '') {
+			if ($originalPath[0] === '/') {
+				$isLocalPath = true;
+			} elseif (strlen($originalPath) >= 3 && $originalPath[0] === 'C' && $originalPath[1] === ':' && $originalPath[2] === '\\') {
+				$isLocalPath = true;
```
staabm, line 61: on windows we now take this fast path and prevent PCRE usage for regular paths starting with `C:\\`

ondrejmirtes: What about other drive letters?

ondrejmirtes (changes requested):

src/File/FileHelper.php
staabm: are you "ok" with allowing any uppercase character?

ondrejmirtes: Yeah, but I feel like there has to be a better way other than comparing letters as if they were numbers :)

staabm: wdyt about using `ord()`? :)

to be honest, I don't like both ways. using regex would work, but we wanne be fast, right? :)

staabm: ok, another idea.  now using `range()`

staabm:
@ondrejmirtes any reasons this was not merged yet? the memory improvement is even more valuable then the time improvement IMO (less likely the process needs to go to swap)

ondrejmirtes:
I think you're overthinking it - what about not checking the letter at all and just checking that 2nd and 3rd character is `:\` ?

staabm:
I am fine with that, too. here we go.

ondrejmirtes:
Thank you!

PR #2069, opened 2022-12-09
Remove unnecessary file IO

staabm:
remove file IO when not necessary. the IO subsystem can be super slow or at least performance can vary a lot dependent on when/where you read data from

grafik

### DIFF
--- a/src/Cache/FileCacheStorage.php
+++ b/src/Cache/FileCacheStorage.php
@@ -52,10 +52,6 @@ class FileCacheStorage implements CacheStorage
 		[,, $filePath] = $this->getFilePaths($key);
 
 		return (static function () use ($variableKey, $filePath) {
-			if (!is_file($filePath)) {
-				return null;
-			}
-
 			$cacheItem = @include $filePath;
 			if (!$cacheItem instanceof CacheItem) {
 				return null;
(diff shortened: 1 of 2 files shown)


src/Cache/FileCacheStorage.php
```diff
@@ -52,10 +52,6 @@ public function load(string $key, string $variableKey)
 		[,, $filePath] = $this->getFilePaths($key);
 
 		return (static function () use ($variableKey, $filePath) {
-			if (!is_file($filePath)) {
-				return null;
-			}
-
 			$cacheItem = @include $filePath;
```
staabm, line 55: since we already swallow possible errors and validate the result we read just after that, we don't need to check beforehand whether `$filePath` actually is a file.

src/Reflection/Php/PhpMethodReflection.php
```diff
@@ -223,8 +222,8 @@ private function isVariadic(): bool
 			$filename = $this->declaringTrait->getFileName();
 		}
 
-		if (!$isNativelyVariadic && $filename !== null && is_file($filename)) {
```
staabm, line 226: `filemtime` will return false when $filename is not a file. the code already handles this case.

lets drop this unnecessary file IO

ondrejmirtes:
What's the command you're profiling this with?

staabm:
`blackfire run php bin/phpstan -vvv --debug`

it is compared against phpstan 1.9.x @ da8413c

ondrejmirtes:
Thank you!

PR #2070, opened 2022-12-09
When an array of replacements is passed to `strtr()` it should be an associative array of strings

johnbillion:
This specifies the type of the replacements array to prevent accidental mistakes with the format of the array.

Refs:

* https://phpstan.org/r/43f08118-c4d2-46aa-b05e-3a5c1c9bc4e3
* https://3v4l.org/XXcWD

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -11804,7 +11804,7 @@ return [
 'strtotime' => ['int|false', 'time'=>'string', 'now='=>'int'],
 'strtoupper' => ['string', 'str'=>'string'],
 'strtr' => ['string', 'str'=>'string', 'from'=>'string', 'to'=>'string'],
-'strtr\'1' => ['string', 'str'=>'string', 'replace_pairs'=>'array'],
+'strtr\'1' => ['string', 'str'=>'string', 'replace_pairs'=>'array<string,string>'],
 'strval' => ['string', 'var'=>'mixed'],
 'substr' => ['__benevolent<string|false>', 'string'=>'string', 'start'=>'int', 'length='=>'int'],
 'substr_compare' => ['int<-1, 1>|false', 'main_str'=>'string', 'str'=>'string', 'offset'=>'int', 'length='=>'int', 'case_sensitivity='=>'bool'],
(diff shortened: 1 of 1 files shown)


VincentLanglet:
This won't be valid anymore
https://3v4l.org/UOOL0

But I think we can assume it's not a real use case, and people should use
```
[ '0' => 'e', '1' => 'u' ] 
```
instead

resources/functionMap.php
```diff
@@ -11804,7 +11804,7 @@
 'strtotime' => ['int|false', 'time'=>'string', 'now='=>'int'],
 'strtoupper' => ['string', 'str'=>'string'],
 'strtr' => ['string', 'str'=>'string', 'from'=>'string', 'to'=>'string'],
-'strtr\'1' => ['string', 'str'=>'string', 'replace_pairs'=>'array'],
+'strtr\'1' => ['string', 'str'=>'string', 'replace_pairs'=>'array<string,string>'],
```
VincentLanglet:
```suggestion
'strtr\'1' => ['string', 'str'=>'string', 'replace_pairs'=>'array<non-empty-string,string>'],
```

According to
> If replace_pairs contains a key which is an empty string (""), the element is ignored; as of PHP 8.0.0 E_WARNING is raised in this case.

https://www.php.net/manual/en/function.strtr.php

VincentLanglet (changes requested):

johnbillion:
That's a pretty wild use case. Is that something you've seen before?

VincentLanglet:
> That's a pretty wild use case. Is that something you've seen before?

No, I just preferred to mention it.

I checked https://www.php.net/manual/en/function.strtr.php
and it's said
> The replace_pairs parameter may be used instead of to and from, in which case it's an array in the form array('from' => 'to', ...).

So `[ 'e', 'u' ] ` is not supposed to be supported by PHP.

VincentLanglet:
Thinking again about this PR, this will create false-positive @johnbillion 

Since array like `['1' => 'u' ]` are inferred as an array with integer keys (see https://phpstan.org/r/ee405b11-251f-4e03-83a6-4db613e4e55f), it will be reported with the change you propose.

So it will be impossible to replace a number by something else with strtr without a PHPStan error.

Unless a test prove I'm wrong I think this should be closed

PR #2071, opened 2022-12-09
Reduce calls to Scope->getType(). Cheap checks first.

staabm:
```
 php -v
PHP 8.1.13 (cli) (built: Nov 24 2022 15:58:42) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.13, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.13, Copyright (c), by Zend Technologies
    with blackfire v1.81.0~mac-x64-non_zts81, https://blackfire.io, by Blackfire
```

`blackfire run php bin/phpstan -vvv --debug`

grafik

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -208,9 +208,9 @@ class FunctionCallParametersCheck
 		}
 
 		if (
-			$scope->getType($funcCall)->isVoid()->yes()
+			!$funcCall instanceof Node\Expr\New_
 			&& !$scope->isInFirstLevelStatement()
-			&& !$funcCall instanceof Node\Expr\New_
+			&& $scope->getType($funcCall)->isVoid()->yes()
 		) {
 			$errors[] = RuleErrorBuilder::message($messages[7])->line($funcCall->getLine())->build();
 		}


ondrejmirtes:
Oh yeah 😊 Thank you 😊 If you find more of these that'd be great!

voku:
I think the Php Inspections (EA Extended) plugin for PhpStorm had a inspection for non optional conditions. Maybe it can find some more stuff? 🤔

staabm:
@voku I don't have this plugin. if you find something useful, even if its just a bottleneck without a idea to optimize it, feel free to report. someone will help you :)

staabm:
> If you find more of these that'd be great!

@ondrejmirtes I thought about whether we should have a few performance oriented rules.

somethink like `ExpensiveCallsInBooleanAndRule` which you would give a list of method references, known to be expensive, and the rule would report if these method calls happen in a `&&`  condition, but could be re-ordered to be called later...  (like we saw in this PR)

another rule could be `ExpensivePureCallInLoop` which would report when a pure-call of a expensive method (e.g. `TypeCombinator::union`) happens within a loop, but does not depend on variables which dependend on the loop-scope.
(and the returned type is not relevant for the loop-scope).

wdty?

ondrejmirtes:
These rules are really hard to do. Maybe once we have the contro flow graph, it will become easier.

PR #2073, opened 2022-12-10
Don't re-calculate node-key over and over again

staabm:
we call this method milions of times. lets do it upfront and re-use the value.

node-key calculation is fast already, but it requires a lot of method call across a lot of classes.
doing less calls saves this method call overhead.

grafik

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3478,9 +3478,11 @@ class MutatingScope implements Scope
 		$expressionTypes = $this->expressionTypes;
 		$nativeExpressionTypes = $this->nativeExpressionTypes;
 		$invalidated = false;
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Yes! 🎉 

BTW if you'd find a way to include Blackfire in CI so it'd give us useful report each time, that'd be great! Otherwise the performance will get worse over time again.

staabm:
> include Blackfire in CI so it'd give us useful report each time

blackfire supports performance assertions. but this requires a paid plan. 

maybe there is room for a sponsoring by blackfire so we can achieve that

ondrejmirtes:
Blackfire sponsors PHPStan, maybe they could chip in with a plan too 😀 Currently I'm using Slevomat's paid account.

ondrejmirtes:
Thank you!

PR #2077, opened 2022-12-11
Improve performance again by dumbing down nested arrays

ondrejmirtes:


src/Rules/RuleLevelHelper.php
```diff
@@ -115,13 +123,12 @@ public function accepts(Type $acceptingType, Type $acceptedType, bool $strictTyp
 
 		$accepts = $acceptingType->accepts($acceptedType, $strictTypes);
 		if (!$accepts->yes() && $acceptingType instanceof UnionType) {
+			$results = [];
 			foreach ($acceptingType->getTypes() as $innerType) {
-				if (self::accepts($innerType, $acceptedType, $strictTypes)) {
-					return true;
-				}
+				$results[] = self::innerAccepts($innerType, $acceptedType, $strictTypes);
```
staabm: it might make sense to use `lazyOr` so it may early exit

staabm: (or you just return from within the loop on YES)

PR #2079, opened 2022-12-11
Remove unnecessary regex matching

staabm:
~~.. testing this small change while my laptop is busy running blackfire profiles~~

seems the change is not useful, since I missed single vs. double quotes.
leaving open for now, as it serves as a example for https://github.com/phpstan/phpstan/discussions/8482


ondrejmirtes:
Closing since this isn't going to be merged. Regarding https://github.com/phpstan/phpstan/discussions/8482, feel free to update the formatter, along with some tests :)

PR #2081, opened 2022-12-11
Faster ConstantStringType->toArrayKey()

staabm:
11% speedup of https://github.com/phpstan/phpstan/issues/8504

> This file [mmarton/phpstan-issue8146@master/src/DataFixtures/LocationFixtures.php](https://github.com/mmarton/phpstan-issue8146/blob/master/src/DataFixtures/LocationFixtures.php?rgh-link-date=2022-12-11T08%3A29%3A10Z) from https://github.com/phpstan/phpstan/issues/8146 really slows PHPStan down, because it contains a lot of literal arrays.

-----

grafik


ondrejmirtes:
Thank you!

PR #2082, opened 2022-12-11
Get rid of `TypeTraverser` in `ArrayChunkFunctionReturnTypeExtension`

herndlm:
just a tiny cleanup/refactor, I often misused `TypeTraverser` in the recent past. if this makes sense, I'll create a couple more PRs like this. this extension also looks a bit complicated since the last changes, I'll look into simplifying it as well maybe


ondrejmirtes:
> I'll create a couple more PRs like this

Yeah, sure :)

ondrejmirtes:
Thank you!

PR #2083, opened 2022-12-11
Faster and simpler TrinaryLogic->lazyExtremeIdentity()

staabm:
8% speedup for snippet in https://github.com/phpstan/phpstan/issues/8504#issuecomment-1345491965

-----

grafik

### DIFF
--- a/src/TrinaryLogic.php
+++ b/src/TrinaryLogic.php
@@ -158,24 +158,25 @@ class TrinaryLogic
 		callable $callback,
 	): self
 	{
+		if ($objects === []) {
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Wow, this is extremely smart, it looks like the function almost doesn't do anything 😀

ondrejmirtes:
Thank you!

staabm:
> it looks like the function almost doesn't do anything 😀

hehe, I was also pleasantly suprised. as always: you need to stare at code for several hours until you realize the solution can be that simple :)

PR #1290, opened 2022-05-05
`OptimizedDirectorySourceLocator` based only on `PhpParser`

kukulich:

PR #2086, opened 2022-12-11
Faster ConstantArrayType->isSuperTypeOf()

staabm:
23% speedup of https://github.com/phpstan/phpstan/issues/8504

> This file [mmarton/phpstan-issue8146@master/src/DataFixtures/LocationFixtures.php](https://github.com/mmarton/phpstan-issue8146/blob/master/src/DataFixtures/LocationFixtures.php?rgh-link-date=2022-12-11T08%3A29%3A10Z) from https://github.com/phpstan/phpstan/issues/8146 really slows PHPStan down, because it contains a lot of literal arrays.

----

grafik

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -330,7 +330,12 @@ class ConstantArrayType extends ArrayType implements ConstantType
 				} elseif ($hasOffset->maybe() && !$this->isOptionalKey($i)) {
 					$results[] = TrinaryLogic::createMaybe();
 				}
-				$results[] = $this->valueTypes[$i]->isSuperTypeOf($type->getOffsetValueType($keyType));
+
+				$isValueSuperType = $this->valueTypes[$i]->isSuperTypeOf($type->getOffsetValueType($keyType));
+				if ($isValueSuperType->no()) {
+					return TrinaryLogic::createNo();
+				}
+				$results[] = $isValueSuperType;
 			}
 
 			return TrinaryLogic::createYes()->and(...$results);


src/Type/Constant/ConstantArrayType.php
```diff
@@ -330,7 +330,12 @@ public function isSuperTypeOf(Type $type): TrinaryLogic
 				} elseif ($hasOffset->maybe() && !$this->isOptionalKey($i)) {
 					$results[] = TrinaryLogic::createMaybe();
 				}
-				$results[] = $this->valueTypes[$i]->isSuperTypeOf($type->getOffsetValueType($keyType));
+
+				$isValueSuperType = $this->valueTypes[$i]->isSuperTypeOf($type->getOffsetValueType($keyType));
+				if ($isValueSuperType->no()) {
+					return TrinaryLogic::createNo();
```
herndlm, line 336:
```suggestion
					return $isValueSuperType;
```
should be possible, no? or any other reason why you didn't do that?

staabm: `createNo` is cheap and I like seeing on the first sight without applying any logics what the return type is.

but its the same thing, you are right.

ondrejmirtes:
Thank you!

staabm:
@ondrejmirtes we could apply a similar early return NO on NO here:
https://github.com/phpstan/phpstan-src/blob/e4a5e98d2e13358086ca1f6f489f340dd82839bd/src/Type/Generic/GenericObjectType.php#L165

I have no concrete perf relevant case to measure it. wdyt?

ondrejmirtes:
Only if it shows up in Blackfire :) This is gonna run the same number of times as there are `@template` above a class... so not that many :)

staabm:
ok.

profiling https://github.com/mmarton/phpstan-issue8146/blob/master/src/DataFixtures/LocationFixtures.php I can see the top consumer is now `PhpFileCleaner`

grafik

this makes me think of https://github.com/phpstan/phpstan-src/pull/1290 which deletes said `PhpFileCleaner` class

ondrejmirtes:
Nope, that's completely unrelated to the arrays, it just means there's a lot of dependencies in vendor and otherwise the project (analysed files) is relatively small.

#1290 is going to be slower than the current solution with PhpFileCleaner.

PR #2088, opened 2022-12-11
Loops analysed in linear time instead of exponential time

ondrejmirtes:
Closes https://github.com/phpstan/phpstan/issues/8503


staabm:
Really nice 🎉

PR #2094, opened 2022-12-12
Remove unnecessary FileIO from FileReader

staabm:
immediately try to read the file instead of doing 2 separate IO calls


staabm:
the build error is not caused by the PR

ondrejmirtes:
Thank you!

PR #2095, opened 2022-12-12
Implement identical/equal comparisons on EnumCaseObjectType

staabm:
closes https://github.com/phpstan/phpstan/issues/8485

for comparison see the handling of regular constants/class-constants: https://phpstan.org/r/a6a9d005-a9d5-4f7d-86ad-00a3a3422cec

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -1304,10 +1304,7 @@ class InitializerExprTypeResolver
 		}
 
 		if ($leftType instanceof EnumCaseObjectType && $rightType instanceof EnumCaseObjectType) {
-			return new ConstantBooleanType(
-				$leftType->getClassName() === $rightType->getClassName()
-				&& $leftType->getEnumCaseName() === $rightType->getEnumCaseName(),
-			);
+			return new ConstantBooleanType($leftType->equals($rightType));
 		}
 
 		$isSuperset = $leftType->isSuperTypeOf($rightType);
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -1303,6 +1303,13 @@ class InitializerExprTypeResolver
 			return new ConstantBooleanType($leftType->getValue() === $rightType->getValue());
 		}
 
+		if ($leftType instanceof EnumCaseObjectType && $rightType instanceof EnumCaseObjectType) {
+			return new ConstantBooleanType(
+				$leftType->getClassName() === $rightType->getClassName()
+				&& $leftType->getEnumCaseName() === $rightType->getEnumCaseName(),
+			);
+		}
+
 		$isSuperset = $leftType->isSuperTypeOf($rightType);
 		if ($isSuperset->no()) {
 			return new ConstantBooleanType(false);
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/Comparison/MatchExpressionRuleTest.php
```diff
@@ -153,7 +153,11 @@ public function testEnums(): void
 				56,
 			],
 			[
-				'Match arm comparison between *NEVER* and MatchEnums\Foo is always false.',
+				'Match arm comparison between MatchEnums\Foo::THREE and MatchEnums\Foo::THREE is always true.',
```
staabm, line 156: this new error is a bit misleading. not sure we want to keep it like that.

tests/PHPStan/Rules/Comparison/MatchExpressionRuleTest.php
```diff
@@ -153,7 +153,11 @@ public function testEnums(): void
...
+				76,
+			],
+			[
+				'Match arm is unreachable because previous comparison is always true.',
```
staabm, line 160: in contrast, I like this error way more then the one we had before this PR

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1303,6 +1303,13 @@ public function resolveIdenticalType(Type $leftType, Type $rightType): BooleanTy
 			return new ConstantBooleanType($leftType->getValue() === $rightType->getValue());
 		}
 
+		if ($leftType instanceof EnumCaseObjectType && $rightType instanceof EnumCaseObjectType) {
+			return new ConstantBooleanType(
```
ondrejmirtes: Use EnumCaseObjectType::equals() instead.

tests/PHPStan/Rules/Comparison/MatchExpressionRuleTest.php
ondrejmirtes: It's fine :)

tests/PHPStan/Rules/Comparison/MatchExpressionRuleTest.php
ondrejmirtes: 👍

ondrejmirtes (changes requested):

src/Reflection/InitializerExprTypeResolver.php
staabm: good catch - fixed

ondrejmirtes:
Thank you!

ondrejmirtes:
Sorry, I had to revert this, it's pretty annoying in practice. We first need to do a different feature in order for this one to be less annoying.

staabm:
no problem. in case there is something I can do, please open a issue about the thing we need, so I can investigate.

ondrejmirtes:
The problem is that this now reports "always true" for the last `elseif` in an if-elseif-else blocks. In real-world projects this now reports dozens of new errors.

Also arguably last match arm shouldn't report this. Because people might prefer to write `Foo::Bar => doFoo()` instead of `default => doFoo()` to be notified when there's a new enum case to handle.

So, I have an idea: We need a new custom node visitor to mark the condition expression in last match arm and last elseif with an attribute. And ConstantConditionRuleHelper can use this node attribute to decide that "always true" shouldn't be reported there.

There's already an issue for this: https://github.com/phpstan/phpstan/issues/8042

And after this is implemented, we can afford to report all "always true" conditions for everyone in bleedingEdge/2.0, not just phpstan-strict-rules users. Because the most annoying ones will not be reported anymore.

PR #2096, opened 2022-12-12
Declare more PDOStatement method types

staabm:
These types are part of phpstan-dba for a couple of releases, and therefore can be considered kind of battle tested

### DIFF
--- a/stubs/PDOStatement.stub
+++ b/stubs/PDOStatement.stub
@@ -7,5 +7,16 @@
  */
 class PDOStatement implements Traversable, IteratorAggregate
 {
+    /**
+     * @template T of object
+     * @param class-string<T> $class
+     * @param array<mixed> $ctorArgs
+     * @return false|T
+     */
+    public function fetchObject($class = \stdClass::class, array $ctorArgs = array()) {}
 
+    /**
+     * @return array{name: string, table?: string, native_type?: string, len: int, flags: array<int, string>, precision: int<0, max>, pdo_type: PDO::PARAM_* }|false
+     */
+    public function getColumnMeta(int $column) {}
 }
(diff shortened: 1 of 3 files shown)


stubs/PDOStatement.stub
```diff
@@ -7,5 +7,16 @@
  */
 class PDOStatement implements Traversable, IteratorAggregate
 {
+    /**
+     * @template T
```
ondrejmirtes: should be `T of object`

stubs/PDOStatement.stub
```diff
@@ -7,5 +7,16 @@
...
+     * @param class-string<T> $class
+     * @param array<mixed> $ctorArgs
+     * @return false|T
+     */
+    public function fetchObject($class = \stdclass::class, array $ctorArgs = array()) {}
```
ondrejmirtes: It's `stdClass`

ondrejmirtes (changes requested):

stubs/PDOStatement.stub
staabm: thanks fixed

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

stubs/PDOStatement.stub
```diff
@@ -7,5 +7,16 @@
  */
 class PDOStatement implements Traversable, IteratorAggregate
```
staabm, line 8: would you agree adding a class-level generic for PDOStatement?

```
/**
 * @template rowType
 *
 * @implements Traversable<int|string, rowType>
 * @implements IteratorAggregate<int|string, rowType>
 *
 * @link https://php.net/manual/en/class.pdostatement.php
 */
 ```

I need the class to be generic for phpstan-dba to enrich this type with real-time db schema information.

ondrejmirtes: I have no idea how this class is used in practice in real-world code so I can't decide this :/

PR #2098, opened 2022-12-13
Prevent overly greedy $scope->getType() calls in ImpossibleInstanceOfRule

staabm:
refs https://github.com/phpstan/phpstan/issues/8146

### DIFF
--- a/src/Rules/Classes/ExistingClassInInstanceOfRule.php
+++ b/src/Rules/Classes/ExistingClassInInstanceOfRule.php
@@ -77,9 +77,10 @@ class ExistingClassInInstanceOfRule implements Rule
 		}
 
 		$classReflection = $this->reflectionProvider->getClass($name);
-		$expressionType = $scope->getType($node->expr);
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
FYI this could have been a single PR ;)

It doesn't make much sense to open a bunch of related PRs at the same time:

1) You're risking conflicts.
2) It'd be tedious to apply learnings from the first PR in all the other PRs.

Instead, open a single PR and wait for it to be merged. That way both of these problems are solved.

And these are pretty simple PRs only with a few lines changed...

ondrejmirtes:
Thank you.

staabm:
> And these are pretty simple PRs only with a few lines changed...

I thought it would be easier to review in smaller PRs.. but I am fine with a big one.

fwiw, 1.9.3 vs. 1.9.x-dev has ~10.000 less getType() calls after these PRs :)

`bin/phpstan --debug -vvv`

grafik

PR #2099, opened 2022-12-13
Prevent overly greedy $scope->getType() calls in Comparison*Rules

staabm:
refs https://github.com/phpstan/phpstan/issues/8146


ondrejmirtes:
Thank you.

PR #2102, opened 2022-12-13
Remove redundant getType() calls in MutatingScope

staabm:


src/Analyser/MutatingScope.php
```diff
@@ -3352,10 +3352,7 @@ public function unsetExpression(Expr $expr): self
 			$exprVarType = $scope->getType($expr->var);
 			$dimType = $scope->getType($expr->dim);
 			$unsetType = $exprVarType->unsetOffset($dimType);
-			$exprVarNativeType = $scope->getType($expr->var);
```
staabm, line 3355: I was wondering since these lines are the same as a few lines above. maybe it should instead be `getNativeType` instead of deleting the lines?

I tried both variants, but since all these variants did not regress any tests I went with the faster one.

//cc @rajyan

rajyan: This was intended to use native type, not sure why this  happened… Thanks for pointing it out.
I’ll fix them with regression test.

staabm: thx

PR #2105, opened 2022-12-13
Implement identical/equal comparisons on EnumCaseObjectType

staabm:
implements the idea of https://github.com/phpstan/phpstan-src/pull/2095#issuecomment-1347945023

closes https://github.com/phpstan/phpstan/issues/8042
closes https://github.com/phpstan/phpstan/issues/8485
closes https://github.com/phpstan/phpstan/issues/7721
closes https://github.com/phpstan/phpstan/issues/4242
closes https://github.com/phpstan/phpstan/issues/8240

re-implements https://github.com/phpstan/phpstan-src/pull/2095

### DIFF
--- a/tests/PHPStan/Rules/Comparison/data/strict-comparison.php
+++ b/tests/PHPStan/Rules/Comparison/data/strict-comparison.php
@@ -982,3 +982,23 @@ function () {
 	INF !== INF;
 	NAN !== NAN;
 };
+
+class ArrayWithLongStrings
+{
+
+	public function doFoo()
+	{
+		$array = ['foofoofoofoofoofoofoo','foofoofoofoofoofoofob'];
+
+		foreach ($array as $value) {
+			if ('foofoofoofoofoofoofoo' === $value) {
+				echo 'nope';
+			} elseif ('foofoofoofoofoofoofob' === $value) {
+				echo 'nop nope';
+			} elseif (rand(0, 1) === 0) {
+				echo 'nope';
+			}
+		}
+	}
+
+}
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -489,6 +489,11 @@ services:
 			- phpstan.parser.richParserNodeVisitor
 
 	-
+		class: PHPStan\Parser\LastConditionVisitor
+		tags:
+			- phpstan.parser.richParserNodeVisitor
+
+	-
 		class: PhpParser\NodeVisitor\NodeConnectingVisitor
 
 	-
--- a/src/Parser/LastConditionVisitor.php
+++ b/src/Parser/LastConditionVisitor.php
@@ -1,0 +1,33 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Parser;
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+use function count;
+
+class LastConditionVisitor extends NodeVisitorAbstract
+{
+
+	public const ATTRIBUTE_NAME = 'isLastCondition';
+
+	public function enterNode(Node $node): ?Node
+	{
+		if ($node instanceof Node\Stmt\If_ && $node->elseifs !== []) {
+			$lastElseIf = count($node->elseifs) - 1;
+
+			$node->elseifs[$lastElseIf]->cond->setAttribute(self::ATTRIBUTE_NAME, true);
+		}
+
+		if ($node instanceof Node\Expr\Match_ && $node->arms !== []) {
+			$lastArm = count($node->arms) - 1;
+
+			if ($node->arms[$lastArm]->conds !== null && $node->arms[$lastArm]->conds !== []) {
+				$node->arms[$lastArm]->conds[0]->setAttribute(self::ATTRIBUTE_NAME, true);
+			}
+		}
+
+		return null;
+	}
+
+}
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -1333,6 +1333,23 @@ class InitializerExprTypeResolver
 			return new ConstantBooleanType($leftType->equals($rightType));
 		}
 
+		if ($leftType instanceof TypeWithClassName && $rightType instanceof EnumCaseObjectType) {
+			$classReflection = $leftType->getClassReflection();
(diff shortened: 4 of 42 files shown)


staabm:
I decided against implementing the check in `ConstantConditionRuleHelper` since the 2 rules in question which trigger the annoying errors, do not yet depend on this helper

ondrejmirtes:
Sure, these rules needed update too, but I really need this in the helper. The idea is to stop reporting each "always true" in the last elseif and last match arm :) Also make sure to test that these "always true" are still being reported when in not-last elseif and match arm etc. :) Thank you.

staabm:
what about

https://github.com/phpstan/phpstan-src/blob/926ec9dabb8a293dee02b822f20c7bd859fc728b/tests/PHPStan/Rules/Comparison/data/strict-comparison.php#L647-L660

do we expect a error on line 654, or should this also be prevented? before this PR we get an error there.

staabm:
> The idea is to stop reporting each "always true" in the last elseif and last match arm :)

ok reading this again, I think it should not be reported. we want to stop reporting of _each_ "always true"

tests/PHPStan/Rules/Comparison/data/strict-comparison.php
```diff
@@ -982,3 +982,23 @@ function () {
...
+	public function doFoo()
+	{
+		$array = ['foofoofoofoofoofoofoo','foofoofoofoofoofoofob'];
+
+		foreach ($array as $value) {
+			if ('foofoofoofoofoofoofoo' === $value) {
+				echo 'nope';
+			} elseif ('foofoofoofoofoofoofob' === $value) {
+				echo 'nop nope';
+			} elseif (rand(0, 1) === 0) {
```
staabm, line 998: this is a new variant of the existing test in the same file on line 654, with a additional else-if so we can see the error are reported when its not the last else-if

staabm:
since this PR already fixes 4 open issues and AFAIK will fix the anoying errors the [1st implementation had](https://github.com/phpstan/phpstan-src/pull/2095) I think we should consider merging it.

Taking the new visitor into account in ConstantConditionRuleHelper and all related rules can be done in a separate PR.. wdyt?

ondrejmirtes:
Something is broken here, the pipeline didn't run. Also I don't see the expected changes in `src/Rules/Comparison/ConstantConditionRuleHelper.php`, only an unused use.

staabm:
> Also I don't see the expected changes in `src/Rules/Comparison/ConstantConditionRuleHelper.php`

I can add it back, but I figured its not required for the current PR feature-set/fixes and we could leave it out for a next PR.

see https://github.com/phpstan/phpstan-src/pull/2105#issuecomment-1349498295

ondrejmirtes:
To me to have all the "always true" changes together is more cohesive than only part of the changes for "always true" + enums fix.

To summarize, it'd be best to have two PRs:

1) All changes for "always true"
2) Enums comparison

But I'm fine to have everything in this single PR. Thanks.

staabm:
lets see:
- `MatchExpressionRule` already prevents always-true in the last arm
- I have skimmed thru the issue tracker for open "always-true" errors, which are not yet tested here -> couldn't find one
- I have looked at `Unreachable*Rule`s, but cannot think of a case which requires a new test/is missing

I am a bit lost, as I don't have an idea on how to move forwared here/what exactly needs to be done

ondrejmirtes:
There's also:

* ConstantLooseComparisonRule
* Rules that use `checkAlwaysTrueCheckTypeFunctionCall`: ImpossibleCheckTypeStaticMethodCallRule, ImpossibleCheckTypeMethodCallRule, ImpossibleCheckTypeFunctionCallRule (which means that examples like these should not be reported https://phpstan.org/r/c7affa79-7bcf-4eb5-baff-5f5c4c2587b0)
* And finally the rules that use `ConstantConditionRuleHelper::getBooleanType()`. I don't need to come up with new examples - if you look at tests for these rules, and see something reported as "always true" and you put it into a last `elseif`, it shouldn't be reported anymore.

staabm:
ok, lets see how you like the latest iteration.

I have worked thru all the mentioned rules. what a ride ;).

I think there is no need to adjust `IfConstantConditionRule`, `TernaryOperatorConstantConditionRule`, `WhileLoopAlwaysFalseConditionRule`, `WhileLoopAlwaysTrueConditionRule`.
all the other ones were extended with new test-cases and fixes

staabm:
thinking about `TernaryOperatorConstantConditionRule` again..

this one might be doable:
```php
class ConditionalAlwaysTrue
{
	public function doFoo(int $i, \stdClass $std)
	{
		if ($i > 0) {
		} elseif ($std ? 'foo' : 'bar') { // always-true should not be reported because last condition
		}

		if ($i > 0) {
		} elseif ($std ? 'foo' : 'bar') { // always-true should be reported, because another condition below
		} elseif (rand(0,1)) {
		}
	}
}
```

ondrejmirtes:
Please also add a test for https://github.com/phpstan/phpstan/issues/8240, thanks.

staabm:
https://github.com/phpstan/phpstan/issues/8240 is interessting. its not fixed by the PR as is.

the problem is similar but different. ~~the root cause is that, $foo in this snippet should be resolved in a union of all cases, but is resolved into a object type and therefore the match-arm cases are not substracted it seems.~~ 

https://phpstan.org/r/7f7b8b7b-f780-41f0-a1fb-02276482859b

ondrejmirtes:
Thank you! I have more improvements in mind, I started them already: https://github.com/phpstan/phpstan-src/commit/565fb0f6da9cdc58e8686598015561a848693972

staabm:
cool. if you want me to work on more, just open a new issue and give me an idea what we are targeting :)

ondrejmirtes:
To match the current feature set, we need something like `reportAlwaysTrueInLastCondition` that's going to be false by default but it's going to be turned on in strict rules :) I'm not gonna open an issue, I'm gonna do it myself if no one beats me to it :)

ondrejmirtes:
Also there should be a tip about this config option when such error is reported.

ondrejmirtes:
And also: There is some complicated logic in UnreachableIfBranchesRule which we should rethink for bleedingEdge, because when "always true" is always reported, we might not need another rule to say the same thing...

PR #2106, opened 2022-12-13
Resolve parameter types only when checkArgumentTypes=true

staabm:


src/Rules/FunctionCallParametersCheck.php
```diff
@@ -249,43 +249,42 @@ public function check(
...
-						'#%d %s',
-						$i + 1,
-						$parameterDescription,
-					) : $parameterDescription,
-					$parameterType->describe($verbosityLevel),
-					$argumentValueType->describe($verbosityLevel),
-				))->line($argumentLine)->build();
-			}
+			if ($this->checkArgumentTypes) {
+				$parameterType = TypeUtils::resolveLateResolvableTypes($parameter->getType());
```
staabm, line 253: defer this resolving until we realy need the value

src/Rules/FunctionCallParametersCheck.php
```diff
@@ -249,43 +249,42 @@ public function check(
...
+							$parameterDescription,
+						) : $parameterDescription,
+						$parameterType->describe($verbosityLevel),
+						$argumentValueType->describe($verbosityLevel),
+					))->line($argumentLine)->build();
+				}
+
+				if ($this->checkUnresolvableParameterTypes
+					&& $originalParameter !== null
+					&& isset($messages[13])
```
staabm, line 274: moved the cheap `isset($messages[13])` condition before the 2 resolving calls

ondrejmirtes:
Thank you!

PR #2107, opened 2022-12-13
Fix native type on unset

rajyan:
Two fixes

1. Previously added native type test forgot to use assertNativeType and wasn't working correctly
2. getType was used for native type as mentioned in https://github.com/phpstan/phpstan-src/pull/2102

I'm sorry that I didn't add enough tests when adding these native type changes...

proof of wrong native type on unset
https://phpstan.org/r/85d3f157-6989-4e60-90f3-9aa7a90546ee

### DIFF
--- a/tests/PHPStan/Analyser/data/native-expressions.php
+++ b/tests/PHPStan/Analyser/data/native-expressions.php
@@ -3,6 +3,7 @@
 namespace NativeExpressions;
 
 use function PHPStan\Testing\assertType;
+use function PHPStan\Testing\assertNativeType;
 
 function doFoo(): string
 {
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you, don't worry about it 😊

ondrejmirtes:
Regression test please 😊 https://github.com/phpstan/phpstan/issues/7805#issuecomment-1350400074

tests/PHPStan/Analyser/data/native-expressions.php
```diff
@@ -3,6 +3,7 @@
 namespace NativeExpressions;
 
 use function PHPStan\Testing\assertType;
+use function PHPStan\Testing\assertNativeType;
```
staabm, line 6: @ondrejmirtes wdyt about adding a rule/check which verifies that tests included by `TypeInferenceTestCase` contain proper full-qualified method calls or imports when `assertType` or `assertNativeType` is contained?

ondrejmirtes: @staabm We can make `TypeInferenceTestCase` crash on the supported functions with correct names but without (or with a wrong) namespace :)

PR #2109, opened 2022-12-14
Refactor FilterVarDynamicReturnTypeExtension to pass around Types instead of Args and the Scope

herndlm:
This is preparation for a fix and re-usage of the internals of this thing for `filter_input`


ondrejmirtes:
Thank you!

PR #2110, opened 2022-12-14
NeverType: accept nothing

jiripudil:
I was surprised to find out that `NeverType` accepts everything. Shouldn't it be the opposite? It's the bottom type, it's empty. No value can be of this type. Why accept _every_ value then?

Consider this PR me poking around and waiting to see what builds break.

### DIFF
--- a/src/Type/NeverType.php
+++ b/src/Type/NeverType.php
@@ -63,6 +63,10 @@ class NeverType implements CompoundType
 
 	public function accepts(Type $type, bool $strictTypes): TrinaryLogic
 	{
+		if ($this->isExplicit()) {
+			return $this->isSuperTypeOf($type);
+		}
+
 		return TrinaryLogic::createYes();
 	}
 
--- a/tests/PHPStan/Rules/Functions/CallCallablesRuleTest.php
+++ b/tests/PHPStan/Rules/Functions/CallCallablesRuleTest.php
@@ -262,4 +262,14 @@ class CallCallablesRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/static-call-in-functions.php'], []);
 	}
 
+	public function testBug6485(): void
+	{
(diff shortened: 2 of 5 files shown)


ondrejmirtes:
An idea - only explicit never should have this behaviour. If you have implicit never, it means that the function signature is wrong, and you already have an error at the declaration - no need for a new error when calling it :)

staabm:
interessting :)

you should have a look at the impact this change has on open issues: https://github.com/phpstan/phpstan-src/actions/runs/3693563281#summary-10085912763 (Issue-Bot summary)

jiripudil:
> An idea - only explicit never should have this behaviour. If you have implicit never, it means that the function signature is wrong, and you already have an error at the declaration - no need for a new error when calling it :)

Oh, thanks a lot for this comment! This probably explains why it has been implemented to accept everything. I'll update this so that only the behaviour of explicit never is changed 👍

ondrejmirtes (changes requested):
Thank you, it makes total sense and issue bot agrees :)

Please:

1) Add a regression test for the one implemented feature request that asks for this :)
2) Please add a test that shows that explicit `never` can be passed into explicit `never` :)

ondrejmirtes:
Thank you!

ondrejmirtes:
I'm really gonna go crazy here :D Because of this change, this now reports problem like this:

```

  Line   test.php                                                
  73     Parameter #1 $element of method                                        
         Doctrine\Common\Collections\ArrayCollection<*NEVER*,*NEVER*>::add()    
         expects *NEVER*,                                                       
         object given.                                                    
```

So if you create a new ArrayCollection and immediately try to add something to it, it's gonna report an error.

Maybe we'll have to postpone this until generics are smarter 🤔

herndlm:
That's https://github.com/phpstan/phpstan/discussions/8511 / https://github.com/phpstan/phpstan/discussions/6731 again, right? Did you not see the failing test Ondřej? :P

PR #2112, opened 2022-12-14
Keep NeverType isExplicit-flag in TypeCombinator::intersect()

staabm:
inspired by https://github.com/phpstan/phpstan-src/pull/2110 where we addd logic based on NeverType explicitness, I figured it would be helpful if the TypeCombinator would keep the exlicitness, instead of create a new non-explicit NeverType every-time it combines NeverTypes'.

Before this PR there was no way to differentiate between NeverType explicit and implicit - I therefore added the same "hack" we use to differentiate explicit and implicit MixedType in tests.

---

this could be followed up by changes in e.g. https://github.com/phpstan/phpstan-src/blob/b76e7b1d316126b10c0f2f9058818f75b205b41f/src/Reflection/InitializerExprTypeResolver.php#L500-L502 where we also loose the explicitness (several occurences in the `InitializerExprTypeResolver` with this pattern)

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -946,7 +946,7 @@ class TypeCombinator
 						$valueType = $types[$j]->getValueType();
 						$newValueType = self::intersect($types[$i]->getOffsetValueType($offsetType), $valueType);
 						if ($newValueType instanceof NeverType) {
-							return new NeverType();
+							return $newValueType;
 						}
 						$types[$i] = $types[$i]->setOffsetValueType($offsetType, $newValueType);
 						array_splice($types, $j--, 1);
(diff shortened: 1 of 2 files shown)


src/Type/TypeCombinator.php
```diff
@@ -946,7 +946,7 @@ public static function intersect(Type ...$types): Type
 						$valueType = $types[$j]->getValueType();
 						$newValueType = self::intersect($types[$i]->getOffsetValueType($offsetType), $valueType);
 						if ($newValueType instanceof NeverType) {
-							return new NeverType();
```
staabm, line 949: these lines dropped the explicitness flag before this PR

tests/PHPStan/Type/TypeCombinatorTest.php
```diff
@@ -3507,6 +3507,22 @@ public function dataIntersect(): iterable
...
 				'mixed',
 			],
+			[
+				[
+					new NeverType(true),
+					new IntegerType(),
+				],
+				NeverType::class,
+				'*NEVER*=explicit',
+			],
```
staabm, line 3517: here we have a test which now returns a explicit NeverType. all other tests where adjusted to compensate the additional `implicit` information in the description. no changes regarding the typtes itself.

ondrejmirtes:
Thank you!

PR #2113, opened 2022-12-15
Remove unnecessary md5 hashing in TypeCombinator

staabm:
I don't think we need this hashing here, and since the TypeCombinator is on a very hot path every unncessary computation should be removed

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -31,7 +31,6 @@ use function array_values;
 use function count;
 use function get_class;
 use function is_int;
-use function md5;
 use function sprintf;
 use function usort;
 
@@ -149,7 +148,7 @@ class TypeCombinator
 		for ($i = 0; $i < $typesCount; $i++) {
 			if ($types[$i] instanceof ConstantScalarType) {
 				$type = $types[$i];
-				$scalarTypes[get_class($type)][md5($type->describe(VerbosityLevel::cache()))] = $type;
+				$scalarTypes[get_class($type)][$type->describe(VerbosityLevel::cache())] = $type;
 				unset($types[$i]);
 				continue;
 			}


ondrejmirtes:
Oh boy, it is necessary. You don't want ConstantStringType '1', ConstantIntegerType 1, and ConstantFloatType 1 collapsed into a single entry.

staabm:
running 
```
$int = ((new ConstantIntegerType(1))->describe(VerbosityLevel::cache()));
$string = ((new ConstantStringType('1'))->describe(VerbosityLevel::cache()));
$float = ((new ConstantFloatType(1))->describe(VerbosityLevel::cache()));

var_dump($int, $string, $float);

$arr = [];
$arr[$int] = 'int';
$arr[$string] = 'string';
$arr[$float] = 'float';

var_dump($arr);
```

gives me 

```
string(1) "1"
string(3) "'1'"
string(3) "1.0"
array(3) {
  [1]=>
  string(3) "int"
  ["'1'"]=>
  string(6) "string"
  ["1.0"]=>
  string(5) "float"
}
```

so I think it seems these are different. I am wondering how a md5-hash of a string will create different array-indizes?
I think since `describe()` already returns a string, the indizes cannot collapse into a single entry.

staabm:
also tried a few more examples:
```
$int = ((new ConstantIntegerType(0))->describe(VerbosityLevel::cache()));
$falseString = ((new ConstantStringType('false'))->describe(VerbosityLevel::cache()));
$string = ((new ConstantStringType('0'))->describe(VerbosityLevel::cache()));
$float = ((new ConstantFloatType(0))->describe(VerbosityLevel::cache()));
$false = ((new ConstantBooleanType(false))->describe(VerbosityLevel::cache()));

var_dump($int, $falseString, $string, $float, $false);

$arr = [];
$arr[$int] = 'int';
$arr[$falseString] = 'false-string';
$arr[$string] = 'string';
$arr[$float] = 'float';
$arr[$false] = 'false';

var_dump($arr);
```

gives

```
string(1) "0"
string(7) "'false'"
string(3) "'0'"
string(3) "0.0"
string(5) "false"
array(5) {
  [0]=>
  string(3) "int"
  ["'false'"]=>
  string(12) "false-string"
  ["'0'"]=>
  string(6) "string"
  ["0.0"]=>
  string(5) "float"
  ["false"]=>
  string(5) "false"
}
````

ondrejmirtes:
This is a huge mess. I'm really surprised at the inconsistent PHP behaviour and I'm not sure I want to exploit it. So an integer in a string gets cast to an integer (https://3v4l.org/mfBIZ), but float in a string isn't...

I'm talking about the difference what happens between ConstantIntegerType(0) (ends up as an integer key in the array) vs. ConstantFloatType(0.0) (ends up as a string key in the array).

Also, md5 never showed up in Blackfire profiles for me, so I don't think so that it's significant.

Thank you anyway :)

PR #2114, opened 2022-12-15
Keep NeverType isExplicit-flag in InitializerExprTypeResolver

staabm:
similar to https://github.com/phpstan/phpstan-src/pull/2112

before this PR we lost the explicit-flag on NeverType, when expressions were used in arithmentic operations.
with the changes the explicit flag is preserved.

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -5,14 +5,20 @@ namespace PHPStan\Analyser;
 use Bug4288\MyClass;
 use Bug4713\Service;
 use ExtendingKnownClassWithCheck\Foo;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Scalar\LNumber;
+use PhpParser\Node\Scalar\String_;
 use PHPStan\File\FileHelper;
 use PHPStan\Reflection\InitializerExprContext;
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1075,6 +1080,24 @@ public function testBug8503(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testExplicitNever(): void
+	{
+		$initializerExprTypeResolver = self::getContainer()->getByType(InitializerExprTypeResolver::class);
+
+		$result = $initializerExprTypeResolver->getPlusType(
```
staabm, line 1094: I did only a single test as a representative for one operation.

should I test every single one?

ondrejmirtes: That's fine.

ondrejmirtes: But you should test all the branches obviously, ideally with a dataProvider. So when the left type is explicit, when the right type is explicit, and when none of them are.

staabm: good idea. done.

ondrejmirtes:
Thank you.

PR #2116, opened 2022-12-15
Implement OversizedArrayVisitor to improve huge constant array performance

staabm:
implements the idea described in https://github.com/phpstan/phpstan/issues/8215#issuecomment-1353232909

closes https://github.com/phpstan/phpstan/issues/8215

with this fix, [the snippet ](https://gist.github.com/dpeukert/eadf816dc5cde760c1aa295fc065815e)is analyzable in 2 seconds, while it takes several minutes before.


staabm:
it needs a few fixes for existing tests, but what do you think about the approach in general?

ondrejmirtes:
Can you explain your reasoning behind this code? I don't entirely understand it.

staabm:
the idea is, that phpstan is slow when analyzing huge arrays, like [gist](https://gist.github.com/dpeukert/eadf816dc5cde760c1aa295fc065815e) because it tries to get the most precise result.

we already have an OversizedArrayType, but it works in a way that it needs to build up the constant array using ConstantTypes and then collapsing it into a OversizedArrayType, after realizing its too big. in huge cases even the steps to build up the Constant types and collapse them into OversizedArrayType already takes minutes.

this PR tries to add a very lightweight, but imprecise type inference when very big constant arrays are analyzed.
we determine a rough shape based on only the simplest types, to make sure we don't later run into the very slow Constant types -> OversizedArrayType case.

see the blackfire profile of said gist, where we are super slow because of the sheer number of constant-integer types invovled

grafik

ondrejmirtes:
1) Does this also make https://github.com/phpstan/phpstan-src/blob/aa7b1eabd197ef6b7fb2c8eebbb7876aebc0ba6a/tests/PHPStan/Analyser/data/bug-8146.php fast?
2) Does this make https://github.com/phpstan/phpstan-src/blob/98c66f30dea8bbb075e327cb5c415427000446e8/tests/PHPStan/Analyser/data/bug-8146a.php fast?
3) Does this get rid of false positives when an array type is simplified but the array is returned from a method that typehints a precise array shape?

I need at least 1) and 3) to be true. I understand that 2) might be a different problem.

staabm:
> 1. Does this also make [`aa7b1ea`/tests/PHPStan/Analyser/data/bug-8146.php](https://github.com/phpstan/phpstan-src/blob/aa7b1eabd197ef6b7fb2c8eebbb7876aebc0ba6a/tests/PHPStan/Analyser/data/bug-8146.php) fast?

runs locally in 13 seconds. on the latest 1.9.x-dev it runs in 1min 18 secs.

> 2\. Does this make [`98c66f3`/tests/PHPStan/Analyser/data/bug-8146a.php](https://github.com/phpstan/phpstan-src/blob/98c66f30dea8bbb075e327cb5c415427000446e8/tests/PHPStan/Analyser/data/bug-8146a.php) fast?

I don't think this counts as a constant array in a sense this Visitor could improve. analyzing this source requires a already built up scope, or at least knowledge about the returned types of methods.

> 3\. Does this get rid of false positives when an array type is simplified but the array is returned from a method that typehints a precise array shape?

I will work on this one now

staabm:
with this fix, we achieve the following runtimes:

```
➜  phpstan-src git:(oversized-array) ✗ time php bin/phpstan analyze tests/PHPStan/Analyser/data/bug-8146b.php --debug -vvv
Result cache not used because of debug mode.
    0 [▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░] < 1 sec 24.0 MiB

Result cache was not saved because only files were passed as analysed paths.

                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        

Used memory: 74 MB
php bin/phpstan analyze tests/PHPStan/Analyser/data/bug-8146b.php --debug -vv  0.60s user 0.13s system 99% cpu 0.727 total
➜  phpstan-src git:(oversized-array) ✗ time php bin/phpstan analyze tests/PHPStan/Analyser/data/bug-8215.php --debug -vvv 
Result cache not used because of debug mode.
    0 [▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░] < 1 sec 24.0 MiB

Result cache was not saved because only files were passed as analysed paths.

                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        

Used memory: 74 MB
php bin/phpstan analyze tests/PHPStan/Analyser/data/bug-8215.php --debug -vvv  0.60s user 0.13s system 99% cpu 0.729 total
```

before this PR both tests ran several minutes

src/Analyser/TypeSpecifier.php
```diff
@@ -1080,31 +1076,7 @@ private function specifyTypesForConstantStringBinaryExpression(
 			&& strtolower($exprNode->name->toString()) === 'gettype'
 			&& isset($exprNode->getArgs()[0])
 		) {
-			$type = null;
-			if ($constantType->getValue() === 'string') {
```
staabm: extracted into new helper for re-use in new visitor

src/Rules/FunctionReturnTypeCheck.php
```diff
@@ -93,7 +94,7 @@ public function checkReturnType(
 			];
 		}
 
-		if (!$this->ruleLevelHelper->accepts($returnType, $returnValueType, $scope->isDeclareStrictTypes())) {
+		if (!$returnValue->hasAttribute(OversizedConstantArrayVisitor::IS_OVERSIZED) && !$this->ruleLevelHelper->accepts($returnType, $returnValueType, $scope->isDeclareStrictTypes())) {
```
ondrejmirtes: This isn't a systemic solution. We can't just completely ignore those arrays. Also, there are many other places that check whether Type accepts a Type.

staabm: Ok I think I need to do this check within the Type classes. 

Do you think I can reuse OversizedArrayType or do I need a completely new one?

ondrejmirtes: I've tried to do this (the ConstantArrayType::accepts() change) but it didn't work for me: https://github.com/phpstan/phpstan-src/pull/2077/files#diff-2e28cc3b285eae2b224df51cda5d6d532675e945e2995e2cd76293bca56799d6L303-R309

So maybe you can investigate this path and see what can be done there. The idea is that if an array is dumbed down, it should still be reported if the type is always wrong, like trying to put `array<int>` into `array{foo: string}`.

staabm: hmm maybe we can `generalize` the accepting type when the accepted type is oversized?

staabm: ahh nice, inspired by #2077 I guess I found a working solution

staabm:
> Backward Compatibility

fails because I have removed a `@deprecated` annotation - I guess.
I think its a false positive

staabm:
https://github.com/phpstan/phpstan/issues/8215 sees a ~25x improvement

```
➜  phpstan-src git:(oversized-array) ✗ php bin/phpstan analyze tests/PHPStan/Analyser/data/bug-8146b.php --debug -vvv
Result cache not used because of debug mode.
    0 [▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░] < 1 sec 24.0 MiB

Result cache was not saved because only files were passed as analysed paths.
 [OK] No errors

Used memory: 74 MB
➜  phpstan-src git:(oversized-array) ✗ php bin/phpstan analyze test.php --debug -vvv
Result cache not used because of debug mode.
/Users/staabm/workspace/phpstan-src/test.php
--- consumed 91.98 MB, total 142 MB, took 23.03 s
Result cache was not saved because only files were passed as analysed paths.
 [OK] No errors

Used memory: 142 MB
```


https://github.com/phpstan/phpstan/issues/8146#issuecomment-1351329301 sees a 2x improvement
```
➜  phpstan-src git:(oversized-array) ✗ php bin/phpstan analyze test.php --debug -vvv
Result cache not used because of debug mode.
/Users/staabm/workspace/phpstan-src/test.php
--- consumed 109.98 MB, total 160 MB, took 21.31 s
Result cache was not saved because only files were passed as analysed paths.

➜  phpstan-src git:(1.9.x) ✗ php bin/phpstan analyze test.php --debug -vvv
Result cache not used because of debug mode.
/Users/staabm/workspace/phpstan-src/test.php
--- consumed 136 MB, total 160 MB, took 42.41 s
```

ondrejmirtes:
I don't get what's exactly happening in the visitor, it's a lot of code.

And I don't think it has to be a visitor, to me it looks like it's essentially caching for something that could be in InitializerExprTypeResolver where Array_ type is resolved.

staabm:
ohh thats a good point. let me see whether I can get rid of the Visitor.

staabm:
great suggestion, it got even simpler now.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -440,6 +440,16 @@ public function getConcatType(Expr $left, Expr $right, callable $getTypeCallback
 	 */
 	public function getArrayType(Expr\Array_ $expr, callable $getTypeCallback): Type
 	{
+		// degrade oversized constant arrays, supports only basic types. less precise but slim and fast.
+		if (count($expr->items) > ConstantArrayTypeBuilder::ARRAY_COUNT_LIMIT) {
```
ondrejmirtes, line 455: 1) You need to cache the result in the `$expr` attributes
2) You don't need to reimplement `Expr -> Type` again, you need to take advantage of the fact you're inside InitializerExprTypeResolver already and you have `$getTypeCallback` at your disposal.
3) No need to write a new class - OversizedConstantArrayResolver, it can be just a private method here.
4) You're not gonna need to refactor and create GetType helper either.

ondrejmirtes (changes requested):

src/Reflection/InitializerExprTypeResolver.php
staabm: thank you. Fixed 2,3,4.

why do we need the caching within the attributes?

do you mean we need to cache the result here and use it if existing instead of building it again? or do we need the cached result somewhere else? I did not see evidence that a caching might be useful in a profile.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -497,6 +514,90 @@ public function getArrayType(Expr\Array_ $expr, callable $getTypeCallback): Type
...
+			}
+
+			$isEmptyArray = false;
+
+			$key = $item->key;
+			if ($key !== null) {
+				$isList = false;
+
+				$itemKeyType = $getTypeCallback($key);
+				if (!$itemKeyType instanceof ConstantScalarType || $key instanceof ClassConstFetch) {
```
staabm: We skip const class fetch, as I was not able to make ArrayType accept a OversizedArrayType with reduced precision when using Constant enumerations like `self::SOME_*`.
See failing test when removing this check

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -497,6 +514,90 @@ public function getArrayType(Expr\Array_ $expr, callable $getTypeCallback): Type
...
+			}
+
+			$value = $item->value;
+			if ($value instanceof Array_) {
+				$valueTypes[] = $this->builtDegradedConstantArray($value, $getTypeCallback);
+				continue;
+			}
+
+			$itemValueType = $getTypeCallback($value);
+			if (!$itemValueType instanceof ConstantScalarType || $value instanceof ClassConstFetch) {
```
staabm: I think we could also add support for arrays where const arrays call methods/functions as long as the return type is defined.

example:
```php
…

public function getData(): array
 	{
 		return [
 			'Bács-Kiskun' => [
 				'Ágasegyháza' => [
 					'constituencies' => ['Bács-Kiskun 4.'],
 					'coordinates' => ['lat' => 46.8386043, 'lng' => 19.4502899],
 				],
 			         $this->returnsString() => [
 					'constituencies' => ['Bács-Kiskun 3.'],
 					'coordinates' => ['lat' => 46.6898175, 'lng' => 19.205086],
 				],
 				'Apostag' => [
 					'constituencies' => ['Bács-Kiskun 3.'],
 					'coordinates' => ['lat' => 46.8812652, 'lng' => $this->returnsFloat()],
 				],
 ```

I did not yet work on it, as I think we are still discussing whether the change itself is acceptable.

**Comment by ondrejmirtes** [comment]
I just pushed a simplification of the buildDegradedConstantArray method as I imagine it should work.

A few problems with it though - some tests fail. I think that we'll have to modify not only accepts, but `isSuperTypeOf` as well.

Also - it's not always correct. When `$key === null` needs to be handled, and also `$isList` isn't always truthful.

**Comment by ondrejmirtes** [comment]
Hi, I'm taking over this PR and I'll try to finish it. Thanks.

**Comment by staabm** [comment]
I tried several things which didn't work. Thanks for taking it

File: tests/PHPStan/Rules/Methods/ReturnTypeRuleTest.php [hunk 6]
```diff
@@ -809,4 +809,15 @@ public function testBug8223(): void
 		]);
 	}
 
+	public function testBug8146b(): void
```
staabm: nit: I misstyped the testname

```suggestion
	public function testBug8146c(): void
```

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

staabm:
Thank you 🙏

ddebin:
Do you think we can disable this new behavior? Now, my huge const is oversimplified and the array shape is not guessed as precisely as before.

ondrejmirtes:
@ddebin Please open a new issue and make sure to include a code sample.

ddebin:
@ondrejmirtes https://github.com/phpstan/phpstan/issues/8636

PR #2117, opened 2022-12-15
Add support for strict `$filter_value` of `array_keys()`

herndlm:
Closes https://github.com/phpstan/phpstan/issues/3300

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -1259,7 +1259,6 @@ class ConstantArrayType extends ArrayType implements ConstantType
 		}
 
 		$array = $builder->getArray();
-
 		if (!$array instanceof ConstantArrayType) {
 			throw new ShouldNotHappenException();
 		}
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -1239,26 +1239,32 @@ class ConstantArrayType extends ArrayType implements ConstantType
 	/**
 	 * @return self
 	 */
-	public function getKeysArray(): Type
+	public function getKeysArray(?Type $filterValueType = null): Type
 	{
-		$keyTypes = [];
-		$valueTypes = [];
-		$optionalKeys = [];
-		$autoIndex = 0;
+		$builder = ConstantArrayTypeBuilder::createEmpty();
 
 		foreach ($this->keyTypes as $i => $keyType) {
-			$keyTypes[] = new ConstantIntegerType($i);
-			$valueTypes[] = $keyType;
-			$autoIndex++;
+			$isOptionalKey = $this->isOptionalKey($i);
+			if ($filterValueType !== null) {
+				if ($filterValueType->isSuperTypeOf($this->valueTypes[$i])->no()) {
+					continue;
+				}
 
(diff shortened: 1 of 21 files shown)


src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
```diff
@@ -36,7 +37,16 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 			return $this->phpVersion->arrayFunctionsReturnNullWithNonArray() ? new NullType() : new NeverType();
 		}
 
-		return $arrayType->getKeysArray();
+		$searchValueType = isset($functionCall->getArgs()[1]) ? $scope->getType($functionCall->getArgs()[1]->value) : null;
+		$strictType = isset($functionCall->getArgs()[2]) ? $scope->getType($functionCall->getArgs()[2]->value) : null;
+		$strict = $strictType instanceof ConstantBooleanType && $strictType->getValue();
```
ondrejmirtes: instaneceof *Type is rarely correct, and in this case it isn't :) You also need to account for mixed or bool being passed in here (so we cannot be sure if it's strict or not), so as always, isSuperTypeOf() wins :)

src/Type/Constant/ConstantArrayType.php
```diff
@@ -1212,26 +1212,32 @@ public function generalizeToArray(): Type
...
-			$autoIndex++;
+			$isOptionalKey = $this->isOptionalKey($i);
+			if ($filterValueType !== null) {
+				if ($filterValueType->isSuperTypeOf($this->valueTypes[$i])->no()) {
+					continue;
+				}
 
-			if (!$this->isOptionalKey($i)) {
-				continue;
+				if (!$filterValueType->equals($this->valueTypes[$i])) {
```
ondrejmirtes, line 1253: This logic should be able to use isSuperTypeOf yes/no/maybe.

ondrejmirtes (changes requested):

src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
herndlm: adapted, I think it should be more correct now? I still exit early the same way later though

src/Type/Constant/ConstantArrayType.php
herndlm: this is an edge case I think, it does already above to skip values for `no`, but here (with `maybe|yes` remaining) I want to mark `maybe|yes` as optional EXCEPT they are actually the same. example: filtering with `bool` makes the key with value `true` optional (because it's a supertype), but filtering with `true` should not make it optional (it's still a supertype though). does that make sense? I had a couple of such things already in the past, where I'm interested in basically `$foo->isSuperTypeOf($bar)->yes() && !$foo->equals($bar)`

herndlm:
What do you think about this one? :)

herndlm:
it might make sense to wait if https://github.com/phpstan/phpstan-src/pull/2216 becomes available soon, then this can support loose comparison too relatively easily I hope

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/pull/3066

PR #2078, opened 2022-12-11
Typecombinator union optimization

rajyan:

PR #2118, opened 2022-12-16
Use isArray, isConstantArray instead of instanceof in TypeCombinartor::union

rajyan:
Refactored union of ArrayTypes by extracting `processArrayAccessoryTypes` as a private function.
Also, changed to use `isArray` `isConstantArray` instead of `instanceof ArrayType`.

I think we can apply the bucket idea of https://github.com/phpstan/phpstan/issues/8499 if we filter ArrayTypes, ConstantIntegerTypes|IntegerRangeType and union them before creating buckets.

### DIFF
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -239,7 +239,7 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 
 	public function isArray(): TrinaryLogic
 	{
-		return TrinaryLogic::createYes();
+		return TrinaryLogic::createMaybe();
 	}
 
 	public function isConstantArray(): TrinaryLogic
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -239,7 +239,7 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 
 	public function isArray(): TrinaryLogic
 	{
-		return TrinaryLogic::createMaybe();
+		return TrinaryLogic::createYes();
 	}
 
 	public function isConstantArray(): TrinaryLogic
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -20,9 +20,7 @@ use PHPStan\Type\Generic\TemplateBenevolentUnionType;
 use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\Generic\TemplateTypeFactory;
 use PHPStan\Type\Generic\TemplateUnionType;
-use function array_intersect_key;
 use function array_key_exists;
-use function array_keys;
 use function array_map;
 use function array_merge;
 use function array_slice;
@@ -331,7 +329,7 @@ class TypeCombinator
 					self::union($a->getIterableKeyType(), $b->getIterableKeyType()),
 					self::union($a->getIterableValueType(), $b->getIterableValueType()),
 				),
-				null
+				null,
 			];
 		}
 
@@ -505,7 +503,8 @@ class TypeCombinator
 	 * @param Type[] $arrayTypes
 	 * @return Type[]
 	 */
-	private static function processArrayAccessoryTypes(array $arrayTypes) {
+	private static function processArrayAccessoryTypes(array $arrayTypes): array
+	{
 		$accessoryTypes = [];
 		foreach ($arrayTypes as $i => $arrayType) {
 			if ($arrayType instanceof IntersectionType) {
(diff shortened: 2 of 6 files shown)


src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -239,7 +239,7 @@ public function getLastIterableValueType(): Type
 
 	public function isArray(): TrinaryLogic
 	{
-		return TrinaryLogic::createYes();
+		return TrinaryLogic::createMaybe();
```
ondrejmirtes: I haven't checked the rest of this PR but this is certainly wrong. If something is a list then we know it's an array...

ondrejmirtes: I understand you're changing this so that some other algorithm results in an outcome we want, but the typesystem layer always has to tell the truth, otherwise something else breaks...

rajyan: > I haven't checked the rest of this PR but this is certainly wrong. If something is a list then we know it's an array...

I thought that AccessoryType themselves are not isArray, they have to be always intersected with an array, but looking at the failing test, this assumption seems false as you are saying...

I can solve this PR without this change, I just wanted to check what fails with this :) (at least the tests didn't fail, maybe we need some regression tests)

ondrejmirtes: I'm likely to accept whatever you come up with, but make sure to do some benchmarks too :) It's really easy to think you've sped things up, but in reality you might have slowed them down :) Thanks.

rajyan:
I tried to just refactor and keep the original behavior, but there seems to be a slight difference somewhere...

rajyan:
```
╭─ ~/Contribution/phpstan-src  refactor-union *2 !1 ···································· 1 ✘  4s  22:01:50 
╰─ hyperfine 'bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php'
Benchmark 1: bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php
  Time (mean ± σ):      9.168 s ±  0.106 s    [User: 8.924 s, System: 0.195 s]
  Range (min … max):    9.056 s …  9.407 s    10 runs
 
╭─ ~/Contribution/phpstan-src  refactor-union *2 ·············································· ✔  22:06:01 
╰─ git switch 1.9.x
Switched to branch '1.9.x'
Your branch is up to date with 'origin/1.9.x'.
╭─ ~/Contribution/phpstan-src  1.9.x *2 ······················································· ✔  22:06:04 
╰─ hyperfine 'bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php'
Benchmark 1: bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php
  Time (mean ± σ):      9.754 s ±  0.070 s    [User: 9.481 s, System: 0.214 s]
  Range (min … max):    9.696 s …  9.893 s    10 runs
```

I believe this refactor should have a slight performance improvement as I expected, because number of union calls has been decreased.

rajyan:
```
╭─ ~/Contribution/phpstan-src  refactor-union *2 ··· INT ✘  30s  09:53:14 
╰─ hyperfine 'bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php'
Benchmark 1: bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php
  Time (mean ± σ):      9.121 s ±  0.069 s    [User: 8.884 s, System: 0.190 s]
  Range (min … max):    9.055 s …  9.259 s    10 runs
 
╭─ ~/Contribution/phpstan-src  refactor-union *2 ···· ✔  1m 31s  09:54:46 
╰─ git switch 1.9.x
Switched to branch '1.9.x'
Your branch is up to date with 'origin/1.9.x'.
╭─ ~/Contribution/phpstan-src  1.9.x *2 ······················ ✔  09:54:53 
╰─ hyperfine 'bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php'
Benchmark 1: bin/phpstan analyse src/Analyser/NodeScopeResolver.php src/Analyser/MutatingScope.php
  Time (mean ± σ):      9.160 s ±  0.061 s    [User: 8.921 s, System: 0.185 s]
  Range (min … max):    9.067 s …  9.271 s    10 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (9.271 s). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
```

This is the final performance comparison.
This PR is preparation for https://github.com/phpstan/phpstan-src/pull/2078 looks like there isn't a significant performance improvement.

src/Type/TypeCombinator.php
```diff
@@ -165,106 +162,26 @@ public static function union(Type ...$types): Type
...
-		$arrayAccessoryTypesToProcess = [];
-		foreach ($commonArrayAccessoryTypesKeys as $commonKey) {
-			$typesToUnion = [];
-			foreach ($arrayAccessoryTypes as $array) {
-				foreach ($array[$commonKey] as $arrayAccessoryType) {
-					$typesToUnion[] = $arrayAccessoryType;
-				}
-			}
-			$arrayAccessoryTypesToProcess[] = self::union(...$typesToUnion);
-		}
```
rajyan, line 239: this logic is replaced by `count($arrayType) === count($accessoryType[$key])` in
https://github.com/phpstan/phpstan-src/pull/2118/files#diff-fef252fe8071200a31b3d282d1442135d62297098c9b20fd731f3888398ee973R549-R562

also, there is no need to union accessoryTypes except `HasOffsetValueType`, because the types are all the same.

ondrejmirtes:
Thank you!

PR #2120, opened 2022-12-16
Refactor options handling in `FilterVarDynamicReturnTypeExtension`

herndlm:
It looks like I need a `hasOptions(): TrinaryLogic` for a bug fix which is how I ended up with this refactor. I could also get rid of a couple of `instanceof ConstantArrayType` and one `instanceof ErrorType` in the process. Most likely more similar expressions could be refactored, but this extension is a beast and I wanted to keep the changes small on purpose.


ondrejmirtes:
Thank you!

herndlm:
thx, hopefully I can finally do the actual bug fix now 😅

PR #2122, opened 2022-12-18
NodeScopeResolver analysed files not as global state and optional

MartinMystikJonas:
@ondrejmirtes We discussed this on our friday call. I prepared PR that would make NodeScopeResolver usable for our use case of collecting latte templates context.

1) `analysedFiles` are not global state anymore but passed as optional parameter to `resolveNodes` and `resolveStmtNodes`
2) It is optional parameter - when omitted `NodeScopeResolver` will not limit what traits it resolves

I split `resolveStmtNodes` to public `resolveStmtNodes` which can set `analysedFiles` and private `iterateStmtNodes` used internally in `NodeScopeResolver` to avoid reseting `analysedFiles` or passing it arround all private methods of `NodeScopeResolver`

### DIFF
--- a/src/Analyser/Analyser.php
+++ b/src/Analyser/Analyser.php
@@ -19,7 +19,6 @@ class Analyser
 		private FileAnalyser $fileAnalyser,
 		private RuleRegistry $ruleRegistry,
 		private CollectorRegistry $collectorRegistry,
-		private NodeScopeResolver $nodeScopeResolver,
 		private int $internalErrorsCountLimit,
 	)
 	{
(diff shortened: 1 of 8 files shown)


MartinMystikJonas:
Now I see this global state is used by rector. Would you preffer to just make it optional and kept as global state? It would be enough for our use case. I just want to remove global state when I was already making changes to it.

Or make `setAnalysedFiles` public and mark it as deprecated?

MartinMystikJonas:
I created simpler version that only makes setting analysedPaths optionsl for NodeScopeREsolver (used to limit analysis of trait outside of analysedPaths). See #2123

VincentLanglet:
Hi, sorry for the delay. 
The 1.9.x branch is not maintain, if you're still interested in this feature please open it on the PR on the 2.1.x branch instead and this time it should be review quicker.

PR #2123, opened 2022-12-18
NodeScopeResolver setting analysedFiles to limit analysis of traits outside analysedPaths is optional

MartinMystikJonas:
Alternative version of #2122 

This simpler version without aby BC break would be enough for our use case.


VincentLanglet:
Hi, sorry for the delay. 
The 1.9.x branch is not maintain, if you're still interested in this feature please open it on the PR on the 2.1.x branch instead and this time it should be review quicker.

PR #2127, opened 2022-12-19
Fix typo, the related issue is phpstan/phpstan#7270

rajyan:
noticed while investigating https://github.com/phpstan/phpstan/issues/8555 :)


ondrejmirtes:
Thank you :)

PR #2134, opened 2022-12-21
Fix constant-string handling in union-types

staabm:
closes https://github.com/phpstan/phpstan/issues/8568
closes https://github.com/phpstan/phpstan/issues/8562


src/Type/UnionType.php
```diff
@@ -117,7 +136,7 @@ public function getConstantArrays(): array
 
 	public function getConstantStrings(): array
 	{
-		return UnionTypeHelper::getConstantStrings($this->getTypes());
+		return UnionTypeHelper::getConstantStrings($this->getTypesOfClass(ConstantStringType::class));
```
staabm: @herndlm do you think we have a similar problem for other methods in UnionType calling into UnionTypeHelper?

src/Type/UnionTypeHelper.php
```diff
@@ -70,6 +70,10 @@ public static function getConstantStrings(array $types): array
 			$strings[] = $type->getConstantStrings();
 		}
 
+        if ($strings === []) {
```
staabm: php <= 7.3 doesn't like array_merge with an variadic empty array

```
7) PHPStan\Rules\Properties\TypesAssignedToPropertiesRuleTest::testTypesAssignedToStaticPropertiesExpressionNames
ArgumentCountError: array_merge() expects at least 1 parameter, 0 given
```

src/Type/UnionType.php
herndlm: pretty much looks like that I guess :/ there's a difference at least. there might be another though that worried me already a bit in the past: the old `TypeUtils::map()` was often called with `$inspectIntersections` set to `false`, which would do exactly that, not inspect intersections, but I think inspecting them made things better / stricter, so maybe that one's fine

herndlm: tbh I also don't like the "stop on unmatched" logic though, it would also be fixable by e.g. ensuring that `isConstantArray()->yes()` on calling site, right? hmm. sorry, still just thinking out loud here

staabm: I think its only relevant for the cases where the 4th arg to `map` is `true`.

e.g. for these array-methods this is not the case

https://github.com/phpstan/phpstan-src/blob/4a72d1513c86df6aab55672bd676c1fcb8c03f9b/src/Type/TypeUtils.php#L120-L136

ondrejmirtes (changes requested):
1) Please show and fix the problems with the other methods as well. It's also about getArrays and getConstantArrays. Feel free to write unit tests.
2) Do no longer call UnionTypeHelper at all. Obviously the implementation has to be different for UnionType and IntersectionType, this just complicates and obfuscates things.
3) Please also verify these methods work correctly in IntersectionType. As I said - do not call UnionTypeHelper in UnionType/IntersectionType for getArrays, getConstantArrays, getConstantStrings at all.

src/Type/IntersectionType.php
```diff
@@ -98,22 +99,42 @@ public function inferTemplateTypesOn(Type $templateType): TemplateTypeMap
 	 */
 	public function getReferencedClasses(): array
 	{
-		return UnionTypeHelper::getReferencedClasses($this->types);
+		$classes = [];
+		foreach ($this->types as $type) {
+			$classes[] = $type->getReferencedClasses();
```
ondrejmirtes: We could build the array directly and avoid an expensive `array_merge` at the end.

src/Type/IntersectionType.php
```diff
@@ -98,22 +99,42 @@ public function inferTemplateTypesOn(Type $templateType): TemplateTypeMap
...
 
 	public function getArrays(): array
 	{
-		return UnionTypeHelper::getArrays($this->getTypes());
+		$arrays = [];
+		foreach ($this->types as $type) {
+			$arrays[] = $type->getArrays();
+		}
+
+		return array_merge(...$arrays);
```
ondrejmirtes: dtto

src/Type/IntersectionType.php
```diff
@@ -98,22 +99,42 @@ public function inferTemplateTypesOn(Type $templateType): TemplateTypeMap
...
 
 	public function getConstantArrays(): array
 	{
-		return UnionTypeHelper::getConstantArrays($this->getTypes());
+		$constantArrays = [];
+		foreach ($this->types as $type) {
+			$constantArrays[] = $type->getConstantArrays();
+		}
+
+		return array_merge(...$constantArrays);
```
ondrejmirtes: dtto

src/Type/IntersectionType.php
```diff
@@ -98,22 +99,42 @@ public function inferTemplateTypesOn(Type $templateType): TemplateTypeMap
...
 
 	public function getConstantStrings(): array
 	{
-		return UnionTypeHelper::getConstantStrings($this->getTypes());
+		$strings = [];
+		foreach ($this->types as $type) {
+			$strings[] = $type->getConstantStrings();
+		}
+
+		return array_merge(...$strings);
```
ondrejmirtes: dtto

src/Type/UnionType.php
```diff
@@ -102,22 +123,54 @@ private function getSortedTypes(): array
...
+
+		return array_merge(...$arrays);
 	}
 
 	public function getConstantArrays(): array
 	{
-		return UnionTypeHelper::getConstantArrays($this->getTypes());
+		$constantArrays = [];
+		foreach ($this->getTypesOfClass(ConstantArrayType::class) as $type) {
+			$constantArrays[] = $type->getConstantArrays();
```
ondrejmirtes: This logic is wrong. For example if I have `(array{1?: int, 2?: string}&non-empty-array)|(array{Foo, Bar}`, these two arrays should still be returned.

ondrejmirtes: But if I have `array{1?: int, 2?: string}|int`, this method should return an empty array.

Basically - the function should return something if all the types in the union return a non-empty array from `getConstantArrays`. If one of them returns an empty array, an empty array should be returned. The `getTypesOfClass` method is backwards - it returns doing `instanceof *Type` which is something that no longer works.

src/Type/UnionType.php
```diff
@@ -102,22 +123,54 @@ private function getSortedTypes(): array
...
+		}
+
+		return array_merge(...$constantArrays);
 	}
 
 	public function getConstantStrings(): array
 	{
-		return UnionTypeHelper::getConstantStrings($this->getTypes());
+		$strings = [];
+		foreach ($this->getTypesOfClass(ConstantStringType::class) as $type) {
```
ondrejmirtes: Same logic as with getConstantArrays.

You could test this with:

* `''|numeric-string` - should return empty array
* Create a fake denormalized type and test that `getConstantStrings` returns an array of 'foo' and 'bar':

```php
new UnionType([
    new IntersectionType([new ConstantStringType('foo'), new AccessoryLiteralStringType()]),
    new IntersectionType([new ConstantStringType('bar'), new AccessoryLiteralStringType()]),
])
```

(In the future there could be a valid intersection like this - for example if we add nominal types and people are able to define new types like `email-string`, we could have `'ondrej@mirtes.cz'&email-string`.
```

**Review by ondrejmirtes — CHANGES_REQUESTED** [review]

File: src/Type/UnionType.php (hunk 8 above)
Comment (staabm) [reply src/Type/UnionType.php]: thx for the detailled feedback. I couldn't have come up with such tests on my own.

File: src/Type/UnionType.php [hunk 9]
```diff
@@ -102,22 +102,62 @@ private function getSortedTypes(): array
...
+		return $classes;
 	}
 
 	public function getArrays(): array
 	{
-		return UnionTypeHelper::getArrays($this->getTypes());
+		$arrays = [];
+		foreach ($this->types as $type) {
+			foreach ($type->getArrays() as $array) {
+				$arrays[] = $array;
```
ondrejmirtes, line 125: I think getArrays should behave the same way as getConstantArrays here.

ondrejmirtes (changes requested):
Otherwise 👍

src/Type/UnionType.php
staabm: implemented it like https://github.com/phpstan/phpstan-src/blob/5a0d7368bf3540a58626c28708c45c933fe2dda2/src/Type/TypeUtils.php#L20-L64

not sure what todo with the generalization test error

ondrejmirtes:
Pushed some fixes. There's one remaining failing test (`PHPStan\Analyser\AnalyserIntegrationTest::testBug7903`). It seemed familiar - I get the same error when I try to do this change:

```
diff --git a/src/Type/IntersectionType.php b/src/Type/IntersectionType.php
index 913e5a455..e3d745274 100644
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -551,7 +551,12 @@ class IntersectionType implements CompoundType

 	public function getOffsetValueType(Type $offsetType): Type
 	{
-		return $this->intersectTypes(static fn (Type $type): Type => $type->getOffsetValueType($offsetType));
+		$result = $this->intersectTypes(static fn (Type $type): Type => $type->getOffsetValueType($offsetType));
+		if ($this->isOversizedArray()->yes()) {
+			return TypeUtils::toBenevolentUnion($result);
+		}
+
+		return $result;
 	}

```

So it seems like the test case is very brittle. Can you please investigate what types are involved on top of 1.9.x (so that there are no errors) and what types are involved in your PR, so that we can decide what should be done there?

Thanks.

src/Analyser/MutatingScope.php
```diff
@@ -4200,15 +4200,15 @@ private function generalizeVariableTypeHolders(
 
 			$variableTypeHolders[$variableExprString] = new ExpressionTypeHolder(
 				$variableTypeHolder->getExpr(),
-				self::generalizeType($variableTypeHolder->getType(), $otherVariableTypeHolders[$variableExprString]->getType()),
+				self::generalizeType($variableTypeHolder->getType(), $otherVariableTypeHolders[$variableExprString]->getType(), 0),
```
staabm, line 4203: I printed the types after generalization at this line in this PR vs. 1.9.x-dev at 9128321b0

most types are identical. the only different are:

```diff
- string(71) "non-empty-array<literal-string&non-falsy-string, mixed>&oversized-array"
+ string(206) "non-empty-array<literal-string&non-falsy-string, non-empty-array<int|string, 0|(non-empty-array<int|string, 0|(non-empty-array<int|string, *ERROR*>&oversized-array)|float>&oversized-array)>>&oversized-array"
```

so after this PR we no longer get MixedType and therefore we get the errors

src/Analyser/MutatingScope.php
```diff
@@ -4514,17 +4513,14 @@ private static function generalizeType(Type $a, Type $b): Type
...
+		while (count($arrays) > 0) {
 			$temp = $type->getIterableValueType();
-			$arrays = $temp->getArrays();
-			if (count($arrays) === 1) {
-				$type = $arrays[0];
-			} else {
-				$type = $temp;
-			}
+			$type = $temp;
+			$arrays = TypeUtils::getAnyArrays($type);
```
staabm, line 4523: I think our problem is that `getAnyArrays` does not return arrays for the [types in question](https://github.com/phpstan/phpstan-src/pull/2134/files#r1059638126) because we are working with a union of intersection types

grafik

and `TypeUtils::getAnyArrays` does not recursivly check the intersections within unions

ondrejmirtes:
Thank you.

herndlm:
yeah, thx for fixing this! :)

PR #2135, opened 2022-12-21
Fix `MixedType::hasOffsetValueType()` for subtracted types

herndlm:
just stumbled over this by luck. I don't expect it to have much of an influence because most of the code using `hasOffsetValueType()` does a `isOffsetAccessible()` before already. which kind of made me thinking - maybe we should look into how much of a performance impact it has that e.g. `hasOffsetValueType()` is not cached for big constant arrays, but maybe it doesn't matter any more now..

### DIFF
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -514,6 +514,10 @@ class MixedType implements CompoundType, SubtractableType
 
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
+		if ($this->isOffsetAccessible()->no()) {
+			return TrinaryLogic::createNo();
+		}
+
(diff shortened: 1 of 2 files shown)


src/Type/MixedType.php
```diff
@@ -514,6 +514,10 @@ public function isOffsetAccessible(): TrinaryLogic
 
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
+		if ($this->isOffsetAccessible()->no()) {
```
staabm, line 517: this makes we wonder whether the API is meant in a way that the caller is expected to check with `isOffsetAccessible`  before calling `hasOffsetValueType`

ondrejmirtes:
Thank you!

PR #2136, opened 2022-12-21
Result cache should not be invalidated by changes to editorUrl, editorUrlTitle and errorFormat parameters

bendavies:
- Result cache should not be invalidated by changes to editorUrl and editorUrlTitle parameters


ondrejmirtes (changes requested):
Also `errorFormat` can be unset.

ondrejmirtes:
Thank you.

PR #2138, opened 2022-12-21
Fix: Return of parent::__set_state() method is only object

blankse:
Resolves https://github.com/phpstan/phpstan/issues/8573


ondrejmirtes:
Thank you!

PR #2139, opened 2022-12-21
Faster Scope->shouldInvalidateExpression()

staabm:
when analyzing [tcpdf.php](https://github.com/tecnickcom/TCPDF/blob/2fb1c01bc37487d1f94fe1297f8d8ad1b5c290bb/tcpdf.php) most of the time is spend in FindFirst

`blackfire run --ignore-exit-status php ../phpstan-src/bin/phpstan --debug -vvv analyze tcpdf.php`

grafik

speedup:

grafik

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3500,6 +3500,11 @@ class MutatingScope implements Scope
 			return false;
 		}
 
+		// Variables will not contain traversable expressions. skip the NodeFinder overhead
+		if ($expr instanceof Variable && is_string($expr->name)) {
+			return $exprStringToInvalidate === $this->getNodeKey($expr);
+		}
+
 		$nodeFinder = new NodeFinder();
 		$expressionToInvalidateClass = get_class($exprToInvalidate);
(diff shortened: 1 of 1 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -3500,6 +3500,11 @@ private function shouldInvalidateExpression(string $exprStringToInvalidate, Expr
 			return false;
 		}
 
+		// Variables will not contain traversable expression. skip the NodeFinder overhead
+		if ($expr instanceof Variable) {
```
ondrejmirtes: Only if the variable name is string

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: good catch. fixed and re-measured. same characteristics.. seems most variables don't reference other variables :)

ondrejmirtes:
Thank you!

williamdes:
Thank you @staabm!

PR #2140, opened 2022-12-22
Fix a few MongoDB definitions in functionMap.php

UCIS:
I think a few function definitions related to the MongoDB extension are incorrect or outdated. This results in incorrect errors as shown here: https://phpstan.org/r/0bdf022f-164a-4daa-bd51-6f127a5c3e44

I have updated two of the obvious errors, but there might be more outdated definitions. Are these definitions required at all? It appears I can get the same information using runtime reflection.


ondrejmirtes:
Thank you!

ondrejmirtes:
> Are these definitions required at all?

Yes. Reflection is often wrong or incomplete. PHPStan uses multiple sources of data, see https://phpstan.org/user-guide/troubleshooting-types#internal-stubs for more details.

PR #2142, opened 2022-12-24
Make http_response_header a non-empty-list<non-falsy-string>

fluffycondor:
This list always contains at least one element: http status code. So I think it's safe to assume it's a non-empty-list of non-falsy-strings.
https://www.php.net/manual/en/reserved.variables.httpresponseheader.php

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -123,6 +123,8 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\Properties\ReadWritePropertiesExtensionProvider;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
+use PHPStan\Type\Accessory\AccessoryArrayListType;
+use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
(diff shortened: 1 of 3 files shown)


fluffycondor:
Looks like all the failed tests are unrelated:
>  fatal: unable to access 'https://github.com/phpstan/phpstan/': The requested URL returned error: 429

src/Analyser/NodeScopeResolver.php
```diff
@@ -1994,7 +1996,31 @@ static function (?Type $offsetType, Type $valueType, bool $optional) use (&$arra
...
+					TypeCombinator::intersect(
+						new ArrayType(
+							new IntegerType(),
+							TypeCombinator::intersect(
+								new StringType(),
+								new AccessoryNonFalsyStringType(),
+							)
+						),
+						new AccessoryArrayListType(),
+						new NonEmptyArrayType()
```
ondrejmirtes, line 2021: Please try to follow the boy scout rule - always leave the firecamp in a better state than you entered it :) So save the type into a variable and reuse it twice.

Also coding standard check failed.

ondrejmirtes (changes requested):

PR #2146, opened 2022-12-27
Fix `filter_var()` narrowing with unknown options

herndlm:
Closes https://github.com/phpstan/phpstan/issues/8516

started the filter-var.php node scope resolver test file because I'd like to move some other filter-var tests over there

also: the fix is the first commit, the second one is still related, but the rest are unrelated fixes and improvements in additional commits.. thought I'll save you from 10 more PRs here :)

### DIFF
--- a/tests/PHPStan/Analyser/data/filter-var-returns-non-empty-string.php
+++ b/tests/PHPStan/Analyser/data/filter-var-returns-non-empty-string.php
@@ -16,7 +16,7 @@ class Foo
 		assertType('non-empty-string', $str);
 
 		$return = filter_var($str, FILTER_DEFAULT);
-		assertType('non-empty-string|false', $return);
+		assertType('non-empty-string', $return);
 
 		$return = filter_var($str, FILTER_DEFAULT, FILTER_FLAG_STRIP_LOW);
 		assertType('string|false', $return);
@@ -98,7 +98,7 @@ class Foo
 
 		$str2 = '';
 		$return = filter_var($str2, FILTER_DEFAULT);
-		assertType('string|false', $return);
+		assertType("''", $return);
 
 		$return = filter_var($str2, FILTER_VALIDATE_URL);
 		assertType('string|false', $return);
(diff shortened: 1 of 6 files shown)


herndlm:
ah ok, the failing GenericsIntegrationTest / LevelsTest case is unrelated, exists also in other PRs

tests/PHPStan/Analyser/data/filter-var-returns-non-empty-string.php
```diff
@@ -82,7 +82,7 @@ public function run(string $str, int $int, int $positive_int, int $negative_int)
 		assertType('9', $return);
 
 		$return = filter_var(1.0, FILTER_VALIDATE_INT, ['options' => ['min_range' => 1, 'max_range' => 9]]);
```
staabm, line 84: I guess you could now also test/assert `filter_var(111.0, FILTER_VALIDATE_INT, ['options' => ['min_range' => 1, 'max_range' => 9]])` that it returns `false`?

herndlm: good idea, it does luckily 😅 I do not like this extension..

tests/PHPStan/Analyser/data/filter-var.php
```diff
@@ -0,0 +1,41 @@
...
+		assertType('array<false>', filter_var(false, FILTER_VALIDATE_BOOLEAN, FILTER_FORCE_ARRAY | FILTER_NULL_ON_FAILURE));
+
+		assertType('17', filter_var(17, FILTER_VALIDATE_INT));
+
+		assertType('17.0', filter_var(17, FILTER_VALIDATE_FLOAT));
+
+		assertType("'17'", filter_var(17));
+		assertType("'17'", filter_var(17.0));
+		assertType("'1'", filter_var(true));
+		assertType("''", filter_var(false));
```
staabm, line 85: Does this also mean?

```suggestion
		assertType("'1'", filter_var(true));
		assertType("''", filter_var(false));
     assertType("''|'1'", filter_var($bool));

```

herndlm: Yes, it does, this is correctly implemented in BoolType already. Will add more tests tomorrow, thx

herndlm:
sorry for the pressure, but what do you think about this one? it's big I know.. afterwards I should be finally able to re-use the logic and add support for `filter_input()` in the other PR. and I'm looking forward very much to that one, to help with many errors in some legacy code base.. 😊

ondrejmirtes:
This is so complicated 😂 Thank you.

PR #2147, opened 2022-12-27
Add namespace to bug-8573.php / fix `GenericsIntegrationTest`

herndlm:
makes the build green again.

the missing namespace lead to `A` and `B` leaking into other tests and messing up templates in `GenericsIntegrationTest`  with the `functions` case.

the commit history on 1.9.x also shows that this was introduced with https://github.com/phpstan/phpstan-src/commit/630430fe810339da2a933903d2f3639ea3d0cbd5


staabm:
@ondrejmirtes would be great if this fix could be merged. it fixes a build error in all open pull requests

ondrejmirtes:
Thank you!

PR #2148, opened 2022-12-27
Error in TypeInferenceTestCase when missing namespace

staabm:
inspired by https://github.com/phpstan/phpstan-src/pull/2147 added a error condition when test files contain named classes or functions but do not declare a namespace

catched the following error on first run (and even more on subsequent runs):

```
There was 1 error:

1) Error
PHPUnit\Framework\AssertionFailedError: File C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Analyser/data/preg_split.php does not contain a namespace declaration
C:\dvl\Workspace\phpstan-src-staabm\src\Testing\TypeInferenceTestCase.php:195
C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Analyser\NodeScopeResolverTest.php:142

2) Error
PHPUnit\Framework\AssertionFailedError: File C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Analyser/data/bug-1014.php does not contain a namespace declaration
C:\dvl\Workspace\phpstan-src-staabm\src\Testing\TypeInferenceTestCase.php:195
C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Analyser\NodeScopeResolverTest.php:154

```

### DIFF
--- a/src/Testing/TypeInferenceTestCase.php
+++ b/src/Testing/TypeInferenceTestCase.php
@@ -120,7 +120,7 @@ abstract class TypeInferenceTestCase extends PHPStanTestCase
 			if ($node instanceof Node\Stmt\Namespace_) {
 				$isNamespaced = true;
 			}
-			if ($node instanceof Node\Stmt\Class_ && $node->name !== null || $node instanceof Node\Stmt\Function_) {
+			if ($node instanceof Node\Stmt\Class_ && $node->name !== null || $node instanceof Node\Stmt\Function_ || $node instanceof Node\Stmt\Trait_) {
 				$requiresNamespace = true;
 			}
 
--- a/src/Testing/TypeInferenceTestCase.php
+++ b/src/Testing/TypeInferenceTestCase.php
@@ -112,8 +112,18 @@ abstract class TypeInferenceTestCase extends PHPStanTestCase
 	 */
 	public function gatherAssertTypes(string $file): array
 	{
+		$isNamespaced = false;
+		$requiresNamespace = false;
+
 		$asserts = [];
-		$this->processFile($file, function (Node $node, Scope $scope) use (&$asserts, $file): void {
+		$this->processFile($file, function (Node $node, Scope $scope) use (&$asserts, $file, &$isNamespaced, &$requiresNamespace): void {
+			if ($node instanceof Node\Stmt\Namespace_) {
+				$isNamespaced = true;
+			}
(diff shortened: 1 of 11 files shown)


staabm:
~~if this PR is accepted, we should also tackle the same errors in the phpstan extension repos, e.g.~~ - anonymous class detection was fixed via [b333809](https://github.com/phpstan/phpstan-src/pull/2148/commits/b333809fc495e88e88717f8f8dbc97312980df68)

```
1) Warning
The data provider specified for PHPStan\Type\Symfony\ExtensionTest::testFileAsserts is invalid.
File /home/runner/work/phpstan-src/phpstan-src/extension/tests/Type/Symfony/data/kernel_interface.php does not contain a namespace declaration

home/runner/work/phpstan-src/phpstan-src/extension/vendor/phpunit/phpunit/phpunit:60
```

rajyan:
The idea looks great to me! Is it possible to detect duplicate namespaces?

staabm:
I think we could implement a phpstan collector to check for duplicate namespaces across all test files. 


What I don't like with this PR is that it does not cover files e.g. included from Rule*Tests.
But maybe its still good enough for now as it discovered a lot of bugs already

herndlm:
something like https://github.com/slevomat/coding-standard/blob/master/doc/namespaces.md#slevomatcodingstandardnamespacesrequireonenamespaceinfile configured to run only in all `test/*/data/*` dirs or so could be a less aggressive alternative?

staabm:
> so could be a less aggressive alternative?

good idea.

what I like about this PR is, that it only enforces a namespace when you have named symbols contained.
therefore I was able to pinpoint which files are missing a namespace.

some tests might not have a namespace and adding one might imply changing a lot of test expectations (e.g. reported line numbers)

ondrejmirtes:
I don't think this is the right approach. We need something more general because:

1) There are also RuleTestCase child classes with the same requirement.
2) There's also LegacyNodeScopeResolverTest.

What would make more sense to me is a new workflow with a separate PHPStan config that would run PHPStan only on `tests/*/data/*` (which is currently excluded) and only with a custom ruleset with a few rules:

* Every file needs to have a namespace
* There cannot be duplicate functions and classes

Since that implementation is not going to have anything common with this PR, I'm closing it.

ondrejmirtes:
(If you want to send fixes for the discovered problems, sure, send a separate PR with no other changes :))

PR #2149, opened 2022-12-27
Add `Type::isScalar()`

herndlm:
Returns `yes` for bool, float, int or string only. I found myself in need of something like that in the `filter_var()` extension. The alternative would be to combine `isBoolean()`, `isFloat()`, `isInteger()` and `isString()` of course, but that feels somehow wrong too.

### DIFF
--- a/src/Type/CallableType.php
+++ b/src/Type/CallableType.php
@@ -392,6 +392,11 @@ class CallableType implements CompoundType, ParametersAcceptor
 		return TrinaryLogic::createNo();
 	}
 
+	public function isScalar(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
+	}
+
 	public function isCommonCallable(): bool
 	{
 		return $this->isCommonCallable;
--- a/src/Type/CallableType.php
+++ b/src/Type/CallableType.php
@@ -394,7 +394,7 @@ class CallableType implements CompoundType, ParametersAcceptor
 
 	public function isScalar(): TrinaryLogic
 	{
-		return TrinaryLogic::createNo();
+		return TrinaryLogic::createMaybe();
 	}
 
 	public function isCommonCallable(): bool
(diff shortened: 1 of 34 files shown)


src/Type/UnionType.php
```diff
@@ -526,6 +526,11 @@ public function isVoid(): TrinaryLogic
 		return $this->unionResults(static fn (Type $type): TrinaryLogic => $type->isVoid());
 	}
 
+	public function isScalar(): TrinaryLogic
```
staabm, line 529: a few `UnionTypeTest`s would be great

src/Type/CallableType.php
```diff
@@ -392,6 +392,11 @@ public function isVoid(): TrinaryLogic
 		return TrinaryLogic::createNo();
 	}
 
+	public function isScalar(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
```
ondrejmirtes: maybe - strings might be callables

src/Type/JustNullableTypeTrait.php
```diff
@@ -117,4 +117,9 @@ public function isVoid(): TrinaryLogic
 		return TrinaryLogic::createNo();
 	}
 
+	public function isScalar(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
```
ondrejmirtes: I'd remove this from the trait. Where it's going to be missing, I'd just add it there in a specific class.

ondrejmirtes (changes requested):

src/Type/CallableType.php
herndlm: oh good catch, thx

src/Type/JustNullableTypeTrait.php
herndlm: adapted, makes sense. that trait is a bit weird actually, yeah. thought the same when implementing it, but went with "consistency" with other is* methods instead. reminds me also that the traits still need cleaning-up and a rule that keep them clean :/

ondrejmirtes:
Thank you!

PR #2152, opened 2022-12-28
Improve constant string support in DateTime return type extensions

herndlm:
the other DateTime* extensions seem to be using `Type::getConstantStrings()` already

There are many more extensions that need this and I'll open more PRs after this one most likely :)


ondrejmirtes:
Thank you!

PR #2153, opened 2022-12-29
Declare FileReadTrapStreamWrapper::$context property

staabm:
should fix warnings like

```
 Deprecated: Creation of dynamic property                                   
     PHPStan\Reflection\BetterReflection\SourceLocator\FileReadTrapStreamWrapper::$context is deprecated in                                              
     /home/runner/work/redaxo/redaxo/redaxo/src/core/lib/autoload.php on line       256                                                                        
```

this property is defined in the [streamwrapper prototype ](https://www.php.net/manual/en/class.streamwrapper.php#streamwrapper.props.context) and when missing creates a runtime warning

----

the fix is related to https://github.com/phpstan/phpstan/issues/8610 but will not fix said issue

### DIFF
--- a/src/Reflection/BetterReflection/SourceLocator/FileReadTrapStreamWrapper.php
+++ b/src/Reflection/BetterReflection/SourceLocator/FileReadTrapStreamWrapper.php
@@ -27,6 +27,7 @@ use const STREAM_URL_STAT_QUIET;
  */
 final class FileReadTrapStreamWrapper
 {
+
 	/** @var resource|null */
 	public $context;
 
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
This needs a reproduction first. Here's an example how to reproduce open_basedir issues in an integration test: https://github.com/phpstan/phpstan/pull/7518

staabm:
I tried a lot of different things but the warning did not reproduce.

maybe this fix is obsolete with https://github.com/phpstan/phpstan-src/pull/2154.. lets see

ondrejmirtes:
Well if you have a project where you see it, it should be easy to reduce the project to something small that still shows it.

ondrejmirtes:
I expect https://github.com/phpstan/phpstan-src/pull/2154 to fix the problem, because there's just one reproducer -> one PR to fix it.

PR #2155, opened 2022-12-30
Fix hasSideEffects for AnnotationMethodReflection

VincentLanglet:
Related to https://github.com/phpstan/phpstan/discussions/8586#discussioncomment-4558154

### DIFF
--- a/src/Reflection/Annotations/AnnotationMethodReflection.php
+++ b/src/Reflection/Annotations/AnnotationMethodReflection.php
@@ -11,6 +11,7 @@ use PHPStan\Reflection\ParametersAcceptor;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\Type;
+use function strtolower;
 
 class AnnotationMethodReflection implements ExtendedMethodReflection
 {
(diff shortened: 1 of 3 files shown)


src/Reflection/Annotations/AnnotationMethodReflection.php
```diff
@@ -109,6 +110,13 @@ public function getThrowType(): ?Type
 
 	public function hasSideEffects(): TrinaryLogic
 	{
+		if (
+			strtolower($this->getName()) !== '__construct'
```
ondrejmirtes: I think you can delete the thing about __construct.

src/Reflection/Annotations/AnnotationMethodReflection.php
```diff
@@ -109,6 +110,13 @@ public function getThrowType(): ?Type
 
 	public function hasSideEffects(): TrinaryLogic
 	{
+		if (
```
VincentLanglet: I took the logic from https://github.com/phpstan/phpstan-src/blob/97efd4d36b89fe8a5da39bc71abf5a25b44d4d26/src/Reflection/Php/PhpMethodReflection.php#L410-L415

ondrejmirtes (changes requested):

src/Reflection/Annotations/AnnotationMethodReflection.php
VincentLanglet: Removed

ondrejmirtes:
Thank you.

PR #2156, opened 2022-12-31
Added missing namespaces in tests

staabm:
taken from https://github.com/phpstan/phpstan-src/pull/2148


tests/PHPStan/Analyser/data/bug-1014.php
```diff
@@ -1,5 +1,7 @@
 <?php declare(strict_types=1);
 
+namespace bug1014;
```
VincentLanglet:
```suggestion
namespace Bug1014;
```
?

PR #2160, opened 2023-01-03
Add non regression test for Issue 8620

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8620


ondrejmirtes:
Please also add NodeScopeResolverTest that asserts the type after `$key` assignment. Thank you.

tests/PHPStan/Analyser/data/bug-8620.php
```diff
@@ -0,0 +1,33 @@
+<?php declare(strict_types = 1);
+
+namespace Bug8620;
+
+use function PHPStan\Testing\assertType;
+
+class HelloWorld
```
ondrejmirtes: To have duplicate classes in the same namespace is not acceptable. Use a single file at tests/PHPStan/Rules/DeadCode/data/bug-8620.php and just add assertType calls into it.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/DeadCode/data/bug-8620.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Bug8620;
+
+use function PHPStan\Testing\assertType;
+
+class HelloWorld
+{
+	public function nullCoalesceAndConcatenation (?int $a = null): int
+	{
+		$key = ($a ?? "x") . "-";
+		assertType('non-falsy-string', $key);
```
ondrejmirtes, line 12: You still need to reference this file in NodeScopeResolverTest in order for these asserts to fail.

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #2161, opened 2023-01-03
min() and max() do not return false on php8

staabm:
refs https://github.com/phpstan/phpstan/issues/7239#issuecomment-1368407228

bottom line: in php8 `min()` and `max()` do no longer return false
https://www.php.net/manual/en/function.max.php

### DIFF
--- a/tests/PHPStan/Analyser/NodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/NodeScopeResolverTest.php
@@ -133,7 +133,13 @@ class NodeScopeResolverTest extends TypeInferenceTestCase
 
 		yield from $this->gatherAssertTypes(__DIR__ . '/../Reflection/data/staticReturnType.php');
 
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/minmax-arrays.php');
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/minmax.php');
+		if (PHP_VERSION_ID < 80000) {
+			yield from $this->gatherAssertTypes(__DIR__ . '/data/minmax-arrays.php');
+		}
+		if (PHP_VERSION_ID >= 80000) {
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -133,7 +133,13 @@ public function dataFileAsserts(): iterable
 
 		yield from $this->gatherAssertTypes(__DIR__ . '/../Reflection/data/staticReturnType.php');
 
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/minmax-arrays.php');
```
staabm, line 136: turned the existing assertions into a php<8 case (`minmax-arrays.php`)

copied the existing assertion files and adjusted them for php8+ (`minmax-php8.php`) - meaning turn `false` into `*ERROR*`

extracted the assertions which are php version independent into `minmax.php`

staabm:
the build error is legit: we no longer trigger the false-positive

```
Error: Ignored error pattern #^Parameter \#2 \$length of function str_pad expects int, int\<0, max\>\|false given\.$# in path /home/runner/work/phpstan-src/phpstan-src/e2e/integration/repo/src/TextUI/DefaultResultPrinter.php was not matched in reported errors.
  Line   src/TextUI/DefaultResultPrinter.php                                    
         Ignored error pattern #^Parameter \#2 \$length of function str_pad     
         expects int, int<0, max>\|false given\.$# in path                      
         /home/runner/work/phpstan-src/phpstan-src/e2e/integration/repo/src/Te  
         xtUI/DefaultResultPrinter.php was not matched in reported errors.      
 ```

in https://github.com/sebastianbergmann/phpunit/blob/93d4bf4c37aec6384bb9e5d390d9049a463a7256/src/TextUI/DefaultResultPrinter.php#L541-L547

ondrejmirtes:
Thank you.

PR #2174, opened 2023-01-09
Remove or test iterable check in `MissingTypehintCheck`

herndlm:
This is not making any tests fail and I'd like to know why. It was apparently introduced in 41db2c908b1f57561e779d9cd81ea0022d78b28c

### DIFF
--- a/src/Rules/MissingTypehintCheck.php
+++ b/src/Rules/MissingTypehintCheck.php
@@ -6,7 +6,6 @@ use Closure;
 use Generator;
 use Iterator;
 use IteratorAggregate;
-use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Accessory\AccessoryType;
 use PHPStan\Type\CallableType;
@@ -20,7 +19,6 @@ use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeTraverser;
-use PHPStan\Type\TypeWithClassName;
 use Traversable;
 use function array_keys;
 use function array_merge;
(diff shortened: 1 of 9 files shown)


herndlm:
I was hoping it would at least make a CI test fail.
WHY CAN'T IT FAIL
:D

herndlm:
no idea what to do with this. it should be preferably tested. or removed of course

ondrejmirtes:
Running the old commit where this was added (https://github.com/phpstan/phpstan-src/commit/41db2c908b1f57561e779d9cd81ea0022d78b28c) was quite a challenge.

The relevant test that fails when the code is removed is:

```
1) PHPStan\Rules\Methods\MissingMethodParameterTypehintRuleTest::testRule
--- Expected
+++ Actual
@@ @@
 58: Method MissingMethodParameterTypehint\Foo::unionTypeWithUnknownArrayValueTypehint() has parameter $a with no value type specified in iterable type array.
 91: Method MissingMethodParameterTypehint\Bar::acceptsGenericInterface() has parameter $i with generic interface MissingMethodParameterTypehint\GenericInterface but does not specify its types: T, U
 101: Method MissingMethodParameterTypehint\Bar::acceptsGenericClass() has parameter $c with generic class MissingMethodParameterTypehint\GenericClass but does not specify its types: A, B
+111: Method MissingMethodParameterTypehint\CollectionIterableAndGeneric::acceptsCollection() has parameter $collection with no value type specified in iterable type DoctrineIntersectionTypeIsSupertypeOf\Collection.
 111: Method MissingMethodParameterTypehint\CollectionIterableAndGeneric::acceptsCollection() has parameter $collection with generic interface DoctrineIntersectionTypeIsSupertypeOf\Collection but does not specify its types: TKey, T
 119: Method MissingMethodParameterTypehint\CollectionIterableAndGeneric::acceptsCollection2() has parameter $collection with generic interface DoctrineIntersectionTypeIsSupertypeOf\Collection but does not specify its types: TKey, T
```

ondrejmirtes:
The test still failed even on 1.4.0. But no longer does after this modification on 1.5.0.

And unsurprisingly, it no longer fails after this commit: https://github.com/phpstan/phpstan-src/commit/0808a89469a0d28bebd0b8d7b14af1639e441215

So the mystery is answered :)

ondrejmirtes:
Thank you!

herndlm:
Ah nice, thx for debugging/testing. I was a bit lazy here

PR #2178, opened 2023-01-12
Allow default values for allow-private-mutation properties

zonuexe:
closes https://github.com/phpstan/phpstan/issues/8701

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -534,6 +534,7 @@ class NodeScopeResolver
 						false,
 						$scope->isInTrait(),
 						$scope->getClassReflection()->isReadOnly(),
+						false,
 					), $methodScope);
 				}
 			}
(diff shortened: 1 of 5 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -534,6 +534,7 @@ private function processStmtNode(
 						false,
 						$scope->isInTrait(),
 						$scope->getClassReflection()->isReadOnly(),
+						false
```
zonuexe: It seems that there is no processing to get the PHPDoc of the parameter of the constructor promotion, so I ignored it with `false`.

ondrejmirtes: There is: https://phpstan.org/r/f22ccc98-94c2-4b7e-9d0f-152bdc575803

But beware - it doesn't mean there's a default value because the code isn't desugared like that.

That's why one is allowed and the other isn't: https://phpstan.org/r/c9757b83-4b83-4ee5-b517-e92e89b44126

ondrejmirtes:
Thank you.

PR #2179, opened 2023-01-13
Call TypeSpecifier for native types too

ondrejmirtes:


src/Analyser/MutatingScope.php
```diff
@@ -3551,6 +3551,11 @@ public function filterByTruthyValue(Expr $expr): Scope
 
 		$specifiedTypes = $this->typeSpecifier->specifyTypesInCondition($this, $expr, TypeSpecifierContext::createTruthy());
 		$scope = $this->filterBySpecifiedTypes($specifiedTypes);
+
+		$nativeScope = $this->promoteNativeTypes();
+		$nativeSpecifiedTypes = $this->typeSpecifier->specifyTypesInCondition($nativeScope, $expr, TypeSpecifierContext::createTruthy());
+		$scope->nativeExpressionTypes = $nativeScope->filterBySpecifiedTypes($nativeSpecifiedTypes)->expressionTypes;
+
```
rajyan: Currently I don't have a good idea to solve the performance problem, except making native type related options/rules as opt-in.
We simply have to call the methods twice...

tests/PHPStan/Analyser/data/bug-7805.php
```diff
@@ -21,12 +21,10 @@ function foo(array $params)
 		assertNativeType("array<mixed~'help', mixed>", $params);
 		$params = $params === [] ? ['list'] : $params;
 		assertType("array{'list'}", $params);
-		assertNativeType("array{'list'}", $params);
-		// assertNativeType("array<mixed~'help', mixed>", $params); // not working yet
+		assertNativeType("non-empty-array<mixed~'help', mixed>", $params);
 		array_unshift($params, 'help');
 		assertType("array{'help', 'list'}", $params);
```
staabm, line 26: shouldn't the `array_unshift($params, 'help');` also remove the `help` from the array shape (like it does for the native type)?

PR #2181, opened 2023-01-14
Reduce unnecessary work

staabm:


src/Analyser/ExpressionTypeHolder.php
```diff
@@ -35,15 +35,15 @@ public function equals(self $other): bool
 
 	public function and(self $other): self
```
staabm, line 36: this method is on the hot path, prevent unncessary function/method call overhead (we do already similar stuff in other methods of this class)

src/Analyser/MutatingScope.php
```diff
@@ -3143,10 +3143,12 @@ public function enterCatchType(Type $catchType, ?string $variableName): self
 			return $this;
 		}
 
+		$intersectedCatchType = TypeCombinator::intersect($catchType, new ObjectType(Throwable::class));
```
staabm, line 3146: prevent building the same type twice

ondrejmirtes:
Please share Blackfire profile comparison.

staabm:
unable to find a measurable substantial repro

PR #2186, opened 2023-01-17
Support type narrowing with casted parameters

staabm:
closes https://github.com/phpstan/phpstan/issues/8714

https://3v4l.org/LPe72

### DIFF
--- a/src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type\Php;
 
+use PhpParser\Node\Expr\Cast;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Analyser\SpecifiedTypes;
@@ -38,7 +39,8 @@ class CtypeDigitFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyin
 			throw new ShouldNotHappenException();
 		}
 
-		if ($context->true() && $scope->getType($node->getArgs()[0]->value)->isNumericString()->yes()) {
+		$exprArg = $node->getArgs()[0]->value;
+		if ($context->true() && $scope->getType($exprArg)->isNumericString()->yes()) {
 			return new SpecifiedTypes();
 		}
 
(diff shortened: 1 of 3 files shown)


src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
```diff
@@ -54,7 +56,18 @@ public function specifyTypes(FunctionReflection $functionReflection, FuncCall $n
 			]);
 		}
 
-		return $this->typeSpecifier->create($node->getArgs()[0]->value, TypeCombinator::union(...$types), $context, false, $scope);
+		$unionType = TypeCombinator::union(...$types);
+		$specifiedTypes = $this->typeSpecifier->create($exprArg, $unionType, $context, false, $scope);
+
+		if ($context->true() && $exprArg instanceof Cast) {
```
staabm, line 62: my initial attempt was to do this directly within the `TypeSpecifier` ... but I then got the feeling that the type narrowing of a cast-expression is not a concept which can be applied to all specifying extensions.

ondrejmirtes: Yeah, well, `is_string((string) $int)` is always true, but it doesn't make `$int` a string :D

src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
```diff
@@ -54,7 +56,18 @@ public function specifyTypes(FunctionReflection $functionReflection, FuncCall $n
...
+			$castOriginType = TypeCombinator::intersect($scope->getType($exprArg->expr), $unionType);
+
+			$specifiedTypes = $specifiedTypes->unionWith(
+				$this->typeSpecifier->create($exprArg->expr, $castOriginType, $context, false, $scope)
```
ondrejmirtes: Are you sure you can't just use `$unionType` instead of `$castOriginType` here in the 2nd arg?

ondrejmirtes (changes requested):
Also please test the negative context.

src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
staabm: good point. fixed.

ondrejmirtes:
Thank you.

ondrejmirtes:
Thank you.

PR #2189, opened 2023-01-19
Fix types after ctype_digit() with casted parameter

staabm:
closes https://github.com/phpstan/phpstan/issues/8736

### DIFF
--- a/src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
@@ -12,11 +12,13 @@ use PHPStan\Analyser\TypeSpecifierContext;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Accessory\AccessoryNumericStringType;
+use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\FunctionTypeSpecifyingExtension;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\UnionType;
 use function strtolower;
 
 class CtypeDigitFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyingExtension, TypeSpecifierAwareExtension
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
Can you please explain the changes and your findings since the last PR?

tests/PHPStan/Analyser/data/callsite-cast-narrowing.php
```diff
@@ -6,17 +6,21 @@
...
-			assertType('int<48, 57>|int<256, max>|numeric-string', $mixed);
+			assertType('int<0, max>|numeric-string|true', $mixed);
 		} else {
 			assertType('mixed', $mixed);
 		}
 		assertType('mixed', $mixed);
 
 		if (ctype_digit((int) $mixed)) {
-			assertType('int<48, 57>|int<256, max>|numeric-string', $mixed);
+			assertType('mixed', $mixed); // could be *NEVER*
```
staabm, line 23: as can be seen in https://3v4l.org/k3Qce casting the parameter for `ctype-digit` to int beforehand does not yield a usefull result at all. 

therefore I adjusted the extension to only narrow on `(string)` call.

I think its not worth to add logic into the extension which turns all "bad uses" of casted parameters into a NeverType.

src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php
```diff
@@ -59,9 +61,17 @@ public function specifyTypes(FunctionReflection $functionReflection, FuncCall $n
 		$unionType = TypeCombinator::union(...$types);
 		$specifiedTypes = $this->typeSpecifier->create($exprArg, $unionType, $context, false, $scope);
 
-		if ($context->true() && $exprArg instanceof Cast) {
+		if ($context->true() && $exprArg instanceof Cast\String_) {
```
staabm: with this change, I just turn the narrowing for non-string casts back to what it was before the recent PR https://github.com/phpstan/phpstan-src/pull/2186

tests/PHPStan/Analyser/data/callsite-cast-narrowing.php
ondrejmirtes: https://github.com/phpstan/phpstan/issues/8736 is about `(string)` cast, not `(int)` cast, so please explain what changes for `(string)` that makes 1.9.13 not work.

staabm: in 1.9.13 we assumed a parameter type before casting would be the same as the one the `ctype_digit` function narrows.

this is not entirely correct. you can actually pass in e.g. a `1` which casted to string is a numeric-string, but is not contained within the 

https://github.com/phpstan/phpstan-src/blob/a6a3441c829b81bb3630651dbb4664f41ff82649/src/Type/Php/CtypeDigitFunctionTypeSpecifyingExtension.php#L47-L57

type.

what we actually need is a type which represents all values which after casted with `(string)` will lead in a `numeric-string`. thats also the reason why e.g. `bool` is narrowed to `true`

ondrejmirtes:
Thank you.

PR #2190, opened 2023-01-19
Extract `FilterFunctionReturnTypeHelper`

herndlm:
First step to support the same (crazy, but apparently working) logic via `filter_input()` too.


ondrejmirtes:
How do you expect this helper to be called for `filter_input`? What's the difference between `filter_var` and `filter_input`? Thank you.

herndlm:
the core logic is the same, `filter_var()` expects a single mixed var as input, while `filter_input()` expects an input type (GET/POST/..) and a var name (which would be the array key on the input super global). Additionally filter_input handles the case when the var name offset is not existing at all on the superglobal input array.

https://www.php.net/manual/en/function.filter-var.php vs https://www.php.net/manual/en/function.filter-input.php

See https://github.com/phpstan/phpstan-src/pull/2010/files#diff-8d424599365db1271856abb9d8444194e495d7eebcd0894d0e05dc04d871a6ae for planned usage

ondrejmirtes:
Thank you 😊

PR #2191, opened 2023-01-19
Added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/6776
closes https://github.com/phpstan/phpstan/issues/7075


ondrejmirtes:
There's also 7075 please.

ondrejmirtes (changes requested):
Please make sure you're seeing these tests fail first. My guess is that the second one is missing `$this->treatPhpDocTypesAsCertain = false;`.

staabm:
I tested it on 1.9.x and it failed there. wasn't the best idea for this particular test though it seems ;)

ondrejmirtes:
Thank you.

PR #2193, opened 2023-01-20
Fixed method_exists() on class-string&literal-string

staabm:
closes https://github.com/phpstan/phpstan/issues/8752

test fails locally without the patch

```
There was 1 failure:

1) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-8752.php:13" ('type', '/Users/staabm/workspace/phpst...52.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\NeverType Object (...), 13)
Expected type class-string&hasMethod(abc)&literal-string, got type *NEVER* in /Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-8752.php on line 13.
--- Expected
+++ Actual
@@ @@
-'class-string&hasMethod(abc)&literal-string'
+'*NEVER*'

/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:96
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:1193
```

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -17,12 +17,16 @@ use PHPStan\TrinaryLogic;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\Generic\GenericClassStringType;
+use PHPStan\Type\IntersectionType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\TypeUtils;
 use PHPStan\Type\TypeWithClassName;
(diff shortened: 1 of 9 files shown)


ondrejmirtes:
You're not testing the important things:

1) That the message "Call to function method_exists()..." is no longer present.
2) That you can actually call `$s::abc()`.
3) That you can't call `$s->abc()` because that doesn't work.

Also - it'd be nice to correctly implement this for GenericClassStringType, because in that case we know the class name, and we know whether `method_exists()` returns always true (the method exists), always false (the class is final and doesn't have the method), or bool (typical case - method doesn't exist but class isn't final).

staabm:
Pushed the 3 main points. Will work on the GenericClassString in the next days. 

You may merge this in the meantime if you like

staabm:
> Also - it'd be nice to correctly implement this for GenericClassStringType, because in that case we know the class name, and we know whether `method_exists()` returns always true (the method exists), always false (the class is final and doesn't have the method), or bool (typical case - method doesn't exist but class isn't final).

implemented

ondrejmirtes:
Thank you.

PR #2198, opened 2023-01-26
Simplify default type handling in `FilterFunctionReturnTypeHelper`

herndlm:
This thing is still a complexity beast. My plan is to simplify it a bit, this is the first idea that came to mind. Without simplifications fixing things just makes it even more messy..

Basically this avoids unnecessary passing around, setting and unsetting of the default type.

It also fixes the following which was `false` before because it exited before knowing the correct default type.
```php
assertType("'invalid'", filter_var(new stdClass(), FILTER_DEFAULT, ['options' => ['default' => 'invalid']]));
```
https://3v4l.org/OCZ3J


ondrejmirtes:
Thank you!

PR #2199, opened 2023-01-26
Fix big constant arrays not recognized as oversized

staabm:
closes https://github.com/phpstan/phpstan/issues/8775


tests/PHPStan/Analyser/data/bug-8775.php
```diff
@@ -0,0 +1,279 @@
...
+ */
+class Test
+{
+	/**
+	 * @var Fixtures
+	 */
+	public static $fixtures;
+
+	function doFoo():void {
+		assertType('(int|string)', Test::$fixtures['257']);
```
bendavies, line 276: this appears to be the type before this fix? so the fix isn't actually tested here?
https://phpstan.org/r/2e0b1bf7-c51c-42ff-b6f4-eba9e384506c

bendavies: so is there no way for this to be marked as just an `int` then? 256 is a hard limit for type inference?

staabm: Before the fix the type was `int|string`. Now it is `(int|string)` which means its benevolent (more forgiving). 

It should prevent the followup errors. I cannot say for sure as the repro given in the origin issue does not show these followup errors. 

For arrays bigger then 256 elements we cannot be that precise anymore for performance reasons

bendavies: thanks, i didn't know what benevolent union was!

ondrejmirtes:
Changed it a bit. Thank you.

PR #2200, opened 2023-01-27
PropertyFetch used in by-ref arg is a PropertyWrite

staabm:
closes https://github.com/phpstan/phpstan/issues/8781

### DIFF
--- a/src/Node/ClassStatementsGatherer.php
+++ b/src/Node/ClassStatementsGatherer.php
@@ -15,6 +15,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Node\Constant\ClassConstantFetch;
 use PHPStan\Node\Property\PropertyRead;
 use PHPStan\Node\Property\PropertyWrite;
+use PHPStan\Parser\CallLikePropertyFetchArgVisitor;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ReflectionProvider;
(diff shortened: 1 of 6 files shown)


src/Node/ClassStatementsGatherer.php
```diff
@@ -158,6 +162,36 @@ private function gatherNodes(Node $node, Scope $scope): void
 			$this->tryToApplyPropertyReads($node, $scope);
 			return;
 		}
+
+		if ($node instanceof Expr\FuncCall && $node->name instanceof Node\Name && $this->reflectionProvider->hasFunction($node->name, $scope)) {
+			$function = $this->reflectionProvider->getFunction($node->name, $scope);
```
staabm: I am pretty sure, you want us doing this for all cases of `CallLike`. I implemented it for now only for `FuncCall` to get the first round of feedback.

ondrejmirtes (changes requested):
No, this is a wrong approach. Please note that a wrong property type assigned should be reported by TypesAssignedToPropertiesRule but currently isn't. Which means that the code should trigger PropertyAssignNode. If we do that, there's no need to modify ClassStatementsGatherer.

Which means this has to be fixed somewhere around https://github.com/phpstan/phpstan-src/blob/769a141d0c96514603884f70ad07035c2aae26c6/src/Analyser/NodeScopeResolver.php#L3389-L3394. Ideally by calling `processAssignVar` which takes care of the same thing in other places.

Additionally, PHPStan should notice the change of the expression type after such call, not just for variables.

So, in points:

1) Failing test for TypesAssignedToPropertiesRule
2) Failing test in NodeScopeResolverTest for property type change after a call.
3) Fix by calling processAssignVar.

VincentLanglet:
I think this can be close @staabm
the issue was solved by https://github.com/phpstan/phpstan/issues/8781#issuecomment-2340820205

PR #2202, opened 2023-01-29
{"WIP"=>"Add to support List shapes"}

zonuexe:
refs https://github.com/phpstan/phpstan/issues/8789, https://github.com/phpstan/phpdoc-parser/pull/175

no tests yet

### DIFF
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -823,7 +823,12 @@ class TypeNodeResolver
 			$builder->setOffsetValueType($offsetType, $this->resolve($itemNode->valueType, $nameScope), $itemNode->optional);
 		}
 
-		return $builder->getArray();
+		$arrayType = $builder->getArray();
+		if ($typeNode->kind === ArrayShapeNode::KIND_LIST) {
+			$arrayType = AccessoryArrayListType::intersectWith($arrayType);
+		}
+
+		return $arrayType;
 	}
 
 	private function resolveConstTypeNode(ConstTypeNode $typeNode, NameScope $nameScope): Type
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -348,22 +348,6 @@ class IntersectionType implements CompoundType
 				continue;
 			}
 
-			if (
-				substr($typeDescription, 0, strlen('array{')) === 'array{'
-					&& in_array('array', $skipTypeNames, true)
-			) {
-				$isList = false;
-				foreach ($typesToDescribe as $j => $typeToDescribe) {
-					if ($typeToDescribe instanceof AccessoryArrayListType) {
-						$isList = true;
-						unset($typesToDescribe[$j]);
-						break;
-					}
-				}
-				$describedTypes[$i] = 'list{' . substr($typeDescription, strlen('array{'));
-				continue;
-			}
-
 			if (in_array($typeDescription, $skipTypeNames, true)) {
 				continue;
 			}
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -348,6 +348,22 @@ class IntersectionType implements CompoundType
 				continue;
 			}
 
+			if (
+				substr($typeDescription, 0, strlen('array{')) === 'array{'
+					&& in_array('array', $skipTypeNames, true)
+			) {
+				$isList = false;
+				foreach ($typesToDescribe as $j => $typeToDescribe) {
+					if ($typeToDescribe instanceof AccessoryArrayListType) {
+						$isList = true;
+						unset($typesToDescribe[$j]);
+						break;
+					}
+				}
+				$describedTypes[$i] = 'list{' . substr($typeDescription, strlen('array{'));
+				continue;
+			}
+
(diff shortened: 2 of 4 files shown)


zonuexe:
continue work in a few days

tests/PHPStan/Analyser/data/list-shapes.php
```diff
@@ -0,0 +1,25 @@
...
+	 * @param list{} $l1
+	 * @param list{'a'} $l2
+	 * @param list{0: 'a'} $l3
+	 * @param list{0: 'a', 1: 'b'} $l4
+	 * @param list{0: 'a', 1?: 'b'} $l5
+	 * @param list{'a', 'b', ...} $l6
+	 */
+	public function bar($l1, $l2, $l3, $l4, $l5, $l6): void
+	{
+		assertType('list{}', $l1);
```
zonuexe: Test failed, fixed by https://github.com/phpstan/phpdoc-parser/pull/177

ondrejmirtes (changes requested):
I'd choose a different implementation path.

1) Reset everything in src/, leave just the tests.
2) In TypeNodeResolver, intersect `$builder->getArray()` with `AccessoryArrayListType` if kind is list.
3) In `IntersectionType::describeItself` there's a section that makes sure that `array<...>&list` is described as `list<...>`. Please write similar logic there so that `array{...}&list` is described as `list{...}`.

Thank you.

src/PhpDoc/TypeNodeResolver.php
```diff
@@ -823,7 +823,13 @@ private function resolveArrayShapeNode(ArrayShapeNode $typeNode, NameScope $name
 			$builder->setOffsetValueType($offsetType, $this->resolve($itemNode->valueType, $nameScope), $itemNode->optional);
 		}
 
-		return $builder->getArray();
+		$arrayType = $builder->getArray();
+		if ($typeNode->kind === ArrayShapeNode::KIND_LIST) {
+			$arrayType = AccessoryArrayListType::intersectWith($arrayType);
+			var_dump($arrayType);
```
zonuexe: `array{0: 'a', 1: 'b'}&list` is normalized to `array{0: 'a', 1: 'b'}`.

`AccessoryArrayListType::intersectWith($arrayType)` always returns the operand `$arrayType: ConstantArrayType` instead of `IntersectionType` because the following equality always holds.

```
(new AccessoryArrayListType)->isSuperTypeOf($arrayType)->yes() === true
```

zonuexe: @ondrejmirtes `AccessoryArrayListType::intersectWith($arrayType)` never returns `IntersectionType` so changing `IntersectionType::describeItself()` doesn't seem to accomplish what you want, any ideas?

ondrejmirtes: `array{0: 'a', 1: 'b'}` is always a list so that's why the `&list` part is omitted and that's why it's not an intersection type. So we probably don't need to worry about this at all. The point of `list{...}` should be to reject types like `list{foo: 1}`.

zonuexe: ok.

src/Type/IntersectionType.php
```diff
@@ -348,6 +348,22 @@ private function describeItself(VerbosityLevel $level, bool $skipAccessoryTypes)
...
+				foreach ($typesToDescribe as $j => $typeToDescribe) {
+					if ($typeToDescribe instanceof AccessoryArrayListType) {
+						$isList = true;
+						unset($typesToDescribe[$j]);
+						break;
+					}
+				}
+				$describedTypes[$i] = 'list{' . substr($typeDescription, strlen('array{'));
+				continue;
+			}
```
zonuexe: > 3. In `IntersectionType::describeItself` there's a section that makes sure that `array<...>&list` is described as `list<...>`. Please write similar logic there so that `array{...}&list` is described as `list{...}`.

> `array{0: 'a', 1: 'b'}` is always a list so that's why the `&list` part is omitted and that's why it's not an intersection type. So we probably don't need to worry about this at all. The point of `list{...}` should be to reject types like `list{foo: 1}`.


As you say `array{0: 'a', 1: 'b'}&list` is always `ConstantArrayType` not `IntersectionType`. Changing `IntersectionType::describeItself()` for print `list{}` seems to have no effect.

> Reset everything in src/, leave just the tests.

Maybe I need to reapply the reverted fa7bcee09da8565bfe9f5a557c2894922f74faf1 and change  `ConstantArrayType`.

ondrejmirtes: Do not overthink this :) It's fine to describe `list{1}` in PHPDoc as `array{1}` with dumpType.

ondrejmirtes: I think you can delete this change without anything failing.

ondrejmirtes (changes requested):
Please add a test so that we can see that `list{foo: 1}` reports:

> `PHPDoc tag @param for parameter $a contains unresolvable type.`

in IncompatiblePhpDocTypeRuleTest

ondrejmirtes:
Rebasing to see if the Rector failures are fixed.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #2205, opened 2023-01-30
Remove `$otherTypes` in `FilterFunctionReturnTypeHelper`

herndlm:
Since the last PR `$otherTypes` was only dealing with applying the range options, so I thought it might be a better idea to reflect this somehow and remove the special if conditions in the main method then.

I dislike the special handling for `ConstantScalarType` still, but did not touch it on purpose so far because it started to break existing tests.


ondrejmirtes:
I think it should be possible to do this on 1.9.x.

herndlm:
yeah, I just selected the wrong branch here / forgot to change to 1.9.x :)

ondrejmirtes:
Thank you!

PR #2206, opened 2023-01-30
Fix `filter_var()` with float input and FILTER_VALIDATE_INT

herndlm:
Closes https://github.com/phpstan/phpstan/issues/8797


ondrejmirtes:
Thank you.

PR #2208, opened 2023-01-30
Fix sqliteCreateFunction signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8802

cf https://www.php.net/manual/en/pdo.sqlitecreatefunction.php
According to https://www.php.net/manual/en/doc.changelog.php it was added in 7.1.4.


ondrejmirtes:
Thank you.

PR #2209, opened 2023-01-30
Add support for maths with constant numeric string

VincentLanglet:
When doing math, constantString can be casted to ConstantInt by Phpstan to compute the result.
Closes https://github.com/phpstan/phpstan/issues/8803

In the same way boolean can be casted to `0|1`, I added a commit 
Closes https://github.com/phpstan/phpstan/issues/8827

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -65,6 +65,8 @@ use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\TypeUtils;
 use PHPStan\Type\TypeWithClassName;
 use PHPStan\Type\UnionType;
+use PHPStan\Type\VerbosityLevel;
+
 use function array_keys;
 use function array_merge;
 use function assert;
@@ -1479,7 +1481,7 @@ class InitializerExprTypeResolver
 				return $union->toNumber();
 			}
 
-			return $this->integerRangeMath($leftType, $expr, $rightType);
+			return $this->integerRangeMath($leftNumberType, $expr, $rightNumberType);
 		}
 
 		$specifiedTypes = $this->callOperatorTypeSpecifyingExtensions($expr, $leftType, $rightType);
@@ -1532,7 +1534,6 @@ class InitializerExprTypeResolver
 	/**
 	 * @param ConstantIntegerType|IntegerRangeType $range
 	 * @param BinaryOp\Div|BinaryOp\Minus|BinaryOp\Mul|BinaryOp\Plus $node
-	 * @param IntegerRangeType|ConstantIntegerType|UnionType $operand
 	 */
 	private function integerRangeMath(Type $range, BinaryOp $node, Type $operand): Type
 	{
@@ -1549,8 +1550,9 @@ class InitializerExprTypeResolver
 			$unionParts = [];
 
 			foreach ($operand->getTypes() as $type) {
-				if ($type instanceof IntegerRangeType || $type instanceof ConstantIntegerType) {
-					$unionParts[] = $this->integerRangeMath($range, $node, $type);
+				$numberType = $type->toNumber();
+				if ($numberType instanceof IntegerRangeType || $numberType instanceof ConstantIntegerType) {
+					$unionParts[] = $this->integerRangeMath($range, $node, $numberType);
 				} else {
 					$unionParts[] = $type->toNumber();
 				}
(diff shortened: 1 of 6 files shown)


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1453,23 +1453,23 @@ private function callOperatorTypeSpecifyingExtensions(Expr\BinaryOp $expr, Type
...
 				);
 			} elseif ($leftType instanceof UnionType) {
-
 				$unionParts = [];
 
 				foreach ($leftType->getTypes() as $type) {
-					if ($type instanceof IntegerRangeType || $type instanceof ConstantIntegerType) {
-						$unionParts[] = $this->integerRangeMath($type, $expr, $rightType);
+					$numberType = $type->toNumber();
+					if ($numberType instanceof IntegerRangeType || $numberType instanceof ConstantIntegerType) {
```
staabm, line 1472: Would this work with `isInteger()->yes()` instead of `instanceof`?

tests/PHPStan/Analyser/data/bug-8803.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Bug8803;
+
+use function PHPStan\Testing\assertType;
+
+class HelloWorld
```
staabm, line 7: I think we should add a test for constant numeric-strings which are not integer like, e.g. `"1e3"`

src/Reflection/InitializerExprTypeResolver.php
VincentLanglet: We don't want `instanceof Integer` but only IntegerRange and ConstantInteger.

tests/PHPStan/Analyser/data/bug-8803.php
VincentLanglet: Test added

VincentLanglet:
Friendly ping @ondrejmirtes, I'm not sure if it's just a time-issue or if this PR got away from your notifications

ondrejmirtes:
On one real world project, it shows 101 new errors.

One example: 

```php
	/**
	 * @return array<int, array<string, int>>
	 */
	public function getIdsByStatus(BankTransactionStatus $status): array
	{
		return $this->entityManager->createQueryBuilder()
			->select('bt.id')
			->from(BankTransaction::class, 'bt')
			->andWhere('bt.status = :status')->setParameter('status', $status)
			->addOrderBy('bt.id', 'ASC')
			->getQuery()
			->getArrayResult();
	}
```

PHPStan reports:

```
 ------ -------------------------------------------------------------------------------------------
  Line   app/services/Bank/BankTransactionRepository.php
 ------ -------------------------------------------------------------------------------------------
  473    Method Slevomat\Bank\BankTransactionRepository::getIdsByStatus() should return array<int,
         array<string, int>> but returns array<int, array<string, int|null>>.
 ------ -------------------------------------------------------------------------------------------
```

The primary key really can't be null. The definition:

```
	#[ORM\Id]
	#[ORM\GeneratedValue]
	#[ORM\Column(type: 'big_integer', options: ['unsigned' => true])]
	#[Get]
	private ?int $id = null;
```

Another similar example:

```php
			$id = $this->entityManager->createQueryBuilder()
				->select('c.id')
				->from(Campaign::class, 'c')
				->andWhere('c.hash = ?', $hash)
				->setMaxResults(1)
				->getQuery()
				->getSingleScalarResult();
```

PHPStan thinks the `$id` can be `int|null` but in fact can only be `int`.

VincentLanglet:
> On one real world project, it shows 101 new errors.

I think you meant to post this on https://github.com/phpstan/phpstan-doctrine/pull/412
I'll take a look.

ondrejmirtes:
Oh yeah, sorry :)

ondrejmirtes (changes requested):
Hi, type inference tests are great but the linked bug reports are about false positives in specific rules. Please also add regression tests for those rules. Thank you.

VincentLanglet:
> Hi, type inference tests are great but the linked bug reports are about false positives in specific rules. Please also add regression tests for those rules. Thank you.

I added non-regression tests for those rules :)

VincentLanglet:
I rebased the PR, and the three failing build seems unrelated @ondrejmirtes :)

VincentLanglet:
Friendly ping @ondrejmirtes, I rebased the PR. This should be ready to be reviewed :)

ondrejmirtes:
Thank you!

PR #2211, opened 2023-01-31
Regression tests for issue 393

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/393


ondrejmirtes (changes requested):
Please do not write code that would exexute if those classes were autoloaded. At least put it inside anonymous functions.

ondrejmirtes:
Thank you.

PR #2214, opened 2023-02-02
NodeScopeResolver::earlyTerminatingMethodNames remove useless default

janedbal:
I just tried running https://github.com/shipmonk-rnd/phpstan-rules#uselessprivatepropertydefaultvalue, nothing else was found.


ondrejmirtes:
Thank you!

PR #2216, opened 2023-02-02
Implement Type->looseCompare(type)

staabm:
closes https://github.com/phpstan/phpstan/issues/8224


staabm:
for now I have implemented the `looseCompare` methods for all types.
now I need to have a look at the failling tests and afterwards type narrowing

ondrejmirtes:
Type narrowing is a separate concern, I wouldn't do that here.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1354,7 +1354,7 @@ public function resolveEqualType(Type $leftType, Type $rightType): BooleanType
 			return $this->resolveConstantArrayTypeComparison($leftType, $rightType, fn ($leftValueType, $rightValueType): BooleanType => $this->resolveEqualType($leftValueType, $rightValueType));
 		}
 
-		return new BooleanType();
+		return $leftType->looseCompare($rightType);
```
ondrejmirtes: This whole method needs to be just `$leftType->looseCompare($rightType)`.

src/Type/Constant/ConstantStringType.php
```diff
@@ -278,6 +278,16 @@ public function toArrayKey(): Type
...
 	}
 
+	public function isInteger(): TrinaryLogic
+	{
+		return TrinaryLogic::createMaybe();
+	}
+
+	public function isFloat(): TrinaryLogic
+	{
+		return TrinaryLogic::createMaybe();
```
ondrejmirtes: These additions are wrong. ConstantStringType can never be an integer, nor a float.

ondrejmirtes (changes requested):
I don't see the table of `==` comparison described anywhere in this PR https://www.php.net/manual/en/types.comparisons.php.

All types need to take into account being compared to any other type. For example this implementation in NullType is too naive:

```php
return $type->isNull()->toBooleanType();
```

Because all of the following examples are true:

```php
null == []
null == false
null == 0
null == ""
```

ondrejmirtes:
Also an idea for a preliminary or a future PR - `NanType`. See the last comment here: https://www.php.net/manual/en/function.is-nan.php

staabm:
I have implemented a few basic types and added tests for them in `loose-compare.php`. please have a look whether this is done how you imagined it.

btw: this PR will get really big if we implement all combinations for all types here.
do you think we could split this, e.g. doing basic and simple object type in one PR, and all more specialized types in separate PRs?

thank you

src/Type/Traits/ConstantScalarTypeTrait.php
```diff
@@ -46,6 +49,29 @@ public function isSuperTypeOf(Type $type): TrinaryLogic
...
 
+	public function looseCompare(Type $type): BooleanType
+	{
+		if (!$this instanceof ConstantScalarType) {
+			throw new ShouldNotHappenException();
+		}
+
+		if ($type instanceof ConstantScalarType) {
+			// @phpstan-ignore-next-line
+			return new ConstantBooleanType($this->getValue() == $type->getValue()); // phpcs:ignore
```
ondrejmirtes: I'd prefer to implement logic instead of letting PHP runtime do the operation. Because when it's different across PHP versions, it means that the result on the playground wouldn't be representative.

Why? Playground runs on PHP 8.1, but shows results PHP 7.2-PHP8.2. So if `==` between scalars results in different boolean on different version, the playground wouldn't show it.

And there are definitely some differences. A friend recently gave a talk about some gotchas, I asked him for the slides.

src/Type/IntegerType.php
```diff
@@ -128,6 +129,22 @@ public function isScalar(): TrinaryLogic
 		return TrinaryLogic::createYes();
 	}
 
+	public function looseCompare(Type $type): BooleanType
+	{
+		foreach ($type->getConstantStrings() as $stringType) {
+			if ($stringType->isNumericString()->yes()) {
+				continue;
+			}
+			return new ConstantBooleanType(false);
```
ondrejmirtes: I think this is wrong. So you get a union like `'a'|'123'`, it doesn't have to be false.

ondrejmirtes (changes requested):
I like StringType::looseCompare but otherwise it's still gonna be a lot of work 😅 And we're not even diving into type narrowing now. I think it's obvious why I was avoiding this topic until now 😂

src/Type/Traits/ConstantScalarTypeTrait.php
herndlm: I think I triggered one oft those gotchas last week resulting in a very serious bug https://3v4l.org/MEjDC

staabm: nice catch. lets do this first as a [decoupled PR](https://github.com/phpstan/phpstan-src/pull/2220)

src/Reflection/InitializerExprTypeResolver.php
staabm: In case we want to move all the logic we have here into `$leftType->looseCompare($rightType)` we would need to pass the PhpVersion as an additional param, right?
(Loose comparison differs between php versions)

ondrejmirtes:
FYI you might find a lot of wisdom and test cases here in these slides from @janedbal (not everything is about `==`): 
[Is-PHP-still-fucked-up_.pdf](https://github.com/phpstan/phpstan-src/files/10663056/Is-PHP-still-fucked-up_.pdf)


We might also need to do NanType and InfType because of these...

staabm:
I added tons of tests and implemented `looseCompare` on a lot of types.

I happy for everyone giving feedback, since there are a lot of test-cases added and its pretty likely I did not assert everything right :)

src/Type/Accessory/AccessoryNonEmptyStringType.php
```diff
@@ -262,6 +264,12 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		$looseFalse = new NullType();
+
+		if ($looseFalse->isSuperTypeOf($type)->yes()) {
```
ondrejmirtes, line 283: We can be sure it's going to be always false with:

* `''`
* `false`

src/Type/ArrayType.php
```diff
@@ -348,6 +348,26 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		$looseTrue = new ConstantBooleanType(true);
+
+		if ($looseTrue->isSuperTypeOf($type)->yes()) {
+			return new ConstantBooleanType(true);
```
ondrejmirtes: I don't think so. Array might be empty, and `[] == true` is false.

src/Type/ArrayType.php
```diff
@@ -348,6 +348,26 @@ public function isScalar(): TrinaryLogic
...
+		}
+
+		$looseFalse = new UnionType([
+			new ObjectWithoutClassType(),
+			new ConstantBooleanType(false),
```
ondrejmirtes: Array can be equal to `false` if it's empty.

src/Type/BooleanType.php
```diff
@@ -121,6 +121,10 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		if ($type->isObject()->yes()) {
+			return new ConstantBooleanType(true);
```
ondrejmirtes: `false` is equal with any object? I don't think so.

src/Type/Accessory/AccessoryNonEmptyStringType.php
ondrejmirtes: I see that this test passes but I have no idea how:

```
assertType('false', $nonEmptyString == $emptyStr);
```

ondrejmirtes (changes requested):

ondrejmirtes (changes requested):
`InitializerExprTypeResolver::resolveEqualType` should be just `return $leftType->looseCompare($rightType, $this->phpVersion);`, nothing more.

src/Type/Accessory/AccessoryNonEmptyStringType.php
staabm: > I see that this test passes but I have no idea how

it works because it is resolved via the "identical" type in https://github.com/phpstan/phpstan-src/blob/d01b91999311fbcdc091e105ac46df536dcc5648/src/Reflection/InitializerExprTypeResolver.php#L1330-L1337

it will start failing as expected, after https://github.com/phpstan/phpstan-src/pull/2228 is merged :)

src/Type/ArrayType.php
```diff
@@ -348,6 +348,19 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		$looseFalse = new UnionType([
+			new ObjectWithoutClassType(),
+			new StringType(),
+			new NullType(),
+			new IntegerType(),
+			new FloatType(),
+			new NeverType(),
```
staabm, line 366: should all type turn a possible `*NEVER*` operand into a `false`, or just leave it a undecided boolean?

tests/PHPStan/Analyser/data/loose-comparisons.php
```diff
@@ -518,5 +600,881 @@ public function sayEmptyStr(
...
+		assertType('bool', $callable == 'myFunction');
+		assertType('true', $callable == $true);
+		assertType('false', $callable == $false);
+		assertType('true', $callable == $one);
+		assertType('false', $callable == $zero);
+		assertType('false', $callable == 10);
+		assertType('false', $callable == $minusOne);
+		assertType('bool', $callable == $oneStr); // could be false, because invalid function name
+		assertType('bool', $callable == $zeroStr); // could be false, because invalid function name
+		assertType('bool', $callable == $minusOneStr); // could be false, because invalid function name
```
staabm, line 1207: within the type system we cannot use reflection to check whether given strings are valid callables.
we could do some kind of regular expression matching, whether a constant string contains only of chars allowed within a function name, but I don't think its worth it

staabm:
ondrej, do you agree that we can implement only the most common types (all scalars, arrays, basic object types) with this PR, while leaving more specialized ones for a future PR?

src/Type/Accessory/AccessoryNumericStringType.php
```diff
@@ -275,6 +276,10 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		if ($type->isString()->yes() && $type->isNumericString()->no()) {
+			return new ConstantBooleanType(false);
```
ondrejmirtes, line 287: This one depends on the PhpVersion, right? Because of https://wiki.php.net/rfc/string_to_number_comparison

src/Type/Accessory/NonEmptyArrayType.php
```diff
@@ -354,6 +355,18 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		if ($type->isTrue()->yes()) {
+			return new ConstantBooleanType(true);
+		}
+
+		if ($type->isFalse()->yes() || $type->isNull()->yes()) {
```
ondrejmirtes: This kind of condition fails for `false|null`, because both left and right part are maybe. You need to use (new UnionType(false, null))->isSuperTypeOf($type)->yes()`.

You also need to include the below `if` in that union type too - an empty ConstantArrayType.

src/Type/IntegerType.php
```diff
@@ -131,6 +133,20 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		if (count($type->getConstantStrings()) > 0) {
+			foreach ($type->getConstantStrings() as $stringType) {
+				if ($stringType->isNumericString()->yes()) {
```
ondrejmirtes: Depends on PhpVersion, right?

ondrejmirtes (changes requested):
I just looked at several implementations, not all of them. Also please limit your pushes, the email conversation is unbearably big in my email client, there's no reason you needed 74 force-pushes. Push only when the build is green for you locally and you want to show me your code. Thank you.

src/Type/Accessory/AccessoryNonEmptyStringType.php
staabm: `'0' == false` is `true`

src/Type/Accessory/AccessoryNumericStringType.php
staabm: I don't think so, since we compare string with string here.

The rfc describes phpversion specific differences in comparing non-numeric string and numbers

src/Type/IntegerType.php
staabm: right, fixed

src/Type/Accessory/NonEmptyArrayType.php
staabm: fixed

PR #2217, opened 2023-02-02
fix bcmath types.

Warxcell:
Reference in psalm: https://psalm.dev/r/3786d05a6c


ondrejmirtes:
I'm hesitating to merge this for same reason as here: https://github.com/phpstan/phpstan-src/pull/2163#issuecomment-1371065918

With this update, users need to update their PHPDocs with `numeric-string`...

orklah:
If it can help with deciding, bcadd(and others) is documented as taking numeric-strings in Psalm for almost 3 years now: https://github.com/vimeo/psalm/pull/3189

As far as I know, no one complained about that. Though I admit it may be biased because users likely to come and create an issue are users that generally will want more strict checks,

Warxcell:
> I'm hesitating to merge this for same reason as here: [#2163 (comment)](https://github.com/phpstan/phpstan-src/pull/2163#issuecomment-1371065918)
> 
> With this update, users need to update their PHPDocs with `numeric-string`...

but that's a good thing, right? I mean: if they accept user string - the method that returns that string will be just `string` - so phpstan will force them to take that string through `is_numeric` check, otherwise they will get an exception in runtime - which is basically what static analysers are trying to prevent.

image

ondrejmirtes:
Commited as https://github.com/phpstan/phpstan-src/commit/3fb919aa72428f285ee76f45f9d495011b1cf7f7

PR #2220, opened 2023-02-05
Implement php version dependent loose const comparison

staabm:
refs https://github.com/phpstan/phpstan-src/pull/2216#discussion_r1096637226

see https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.string-number-comparision

https://3v4l.org/pWAjV
https://3v4l.org/0VWtq

https://phpstan.org/r/b2748c38-b33e-4db7-a60f-4f5edd991270

### DIFF
--- a/tests/PHPStan/Analyser/looseConstComparisonPhp7.neon
+++ b/tests/PHPStan/Analyser/looseConstComparisonPhp7.neon
@@ -1,0 +1,2 @@
+parameters:
+	phpVersion: 70400 # PHP 7.4
--- a/src/Php/PhpVersion.php
+++ b/src/Php/PhpVersion.php
@@ -216,4 +216,10 @@ class PhpVersion
 		return $this->versionId < 80000;
 	}
 
+	// see https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.string-number-comparision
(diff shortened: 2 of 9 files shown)


tests/PHPStan/Analyser/looseConstComparisonPhp7.neon
```diff
@@ -0,0 +1,2 @@
+parameters:
+	phpVersion: 70400 # PHP 7.4
```
staabm, line 2: when running the test-cases in isolation, outside of the phpstan repo they behave as expected. 

I somehow have the feeling that using `phpVersion` within the test-suite is kind of a unsupported feature in the current test infrastructure (I think we see some kind of cache problem)

ondrejmirtes: You need to change `new PhpVersion(PHP_VERSION_ID)` in PHPStanTestCase

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/looseConstComparisonPhp7.neon
staabm: thank you.

ondrejmirtes:
Thank you.

PR #1116, opened 2022-03-25
Update various DB functions for more specific row counts

Seldaek:
Found all these while working on https://github.com/doctrine/dbal/pull/5334

PR #2223, opened 2023-02-06
Use "numeric-string" type for some MySQLi functions

phansys:
>If the number of rows is greater than PHP_INT_MAX, the number will be returned as a string.

* https://www.php.net/manual/en/mysqli-result.num-rows.php
* https://www.php.net/manual/en/mysqli-stmt.num-rows.php

BTW, is there a way to declare the type for properties like [`mysqli_stmt::$affected_rows`](https://www.php.net/manual/en/mysqli-stmt.affected-rows.php)?

Related to #1116.


resources/functionMap.php
```diff
@@ -7272,7 +7272,7 @@
 'mysqli_multi_query' => ['bool', 'link'=>'mysqli', 'query'=>'string'],
 'mysqli_next_result' => ['bool', 'link'=>'mysqli'],
 'mysqli_num_fields' => ['int', 'link'=>'mysqli_result'],
-'mysqli_num_rows' => ['int', 'link'=>'mysqli_result'],
+'mysqli_num_rows' => ['int<-1,max>|numeric-string', 'link'=>'mysqli_result'],
```
ondrejmirtes: When it's -1?

ondrejmirtes (changes requested):

ondrejmirtes:
> BTW, is there a way to declare the type for properties

Try to add a stub file, but please add a test for it too (NodeScopeResolverTest), it might be buggy.

resources/functionMap.php
phansys: Good catch. I just copied the type from `mysqli_affected_rows`. It's fixed.

ondrejmirtes:
Thank you.

PR #2224, opened 2023-02-06
Use "numeric-string" type for some `mysqli_stmt` properties

phansys:
Follow up of #2223.

#### To-Do:
- [x] Add test.


ondrejmirtes:
Thank you.

PR #2225, opened 2023-02-06
Preparational refactoring to ease implementation of `Type->looseCompare(type, phpVersion)`

staabm:
lets see whether we agree on this intermediate step

### DIFF
--- a/src/Type/Type.php
+++ b/src/Type/Type.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type;
 
+use PHPStan\Php\PhpVersion;
 use PHPStan\Reflection\ClassMemberAccessAnswerer;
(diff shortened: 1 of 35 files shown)


src/Type/Type.php
```diff
@@ -195,6 +196,8 @@ public function isVoid(): TrinaryLogic;
 
 	public function isScalar(): TrinaryLogic;
 
+	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType;
```
staabm, line 199: in case we agree on this signature, I would move more logic from `InitializerExprTypeResolver->resolveEqualType` into the type-system

ondrejmirtes: Yes please, it makes sense :)

ondrejmirtes:
Thank you!

PR #2228, opened 2023-02-07
Refactor `InitializerExprTypeResolver->resolveEqualType()`

staabm:
move the logic into the type-classes, as requested in https://github.com/phpstan/phpstan-src/pull/2216#pullrequestreview-1286571372

### DIFF
--- a/src/Type/ComparisonHelper.php
+++ b/src/Type/ComparisonHelper.php
@@ -9,10 +9,10 @@ use PHPStan\Type\Constant\ConstantBooleanType;
 use function array_keys;
 use function count;
 
-final class LooseComparisonHelper
+final class ComparisonHelper
 {
 
-	public static function compareConstantScalars(ConstantScalarType $leftType, ConstantScalarType $rightType, PhpVersion $phpVersion): BooleanType
+	public static function looseCompareConstantScalars(ConstantScalarType $leftType, ConstantScalarType $rightType, PhpVersion $phpVersion): BooleanType
 	{
 		if ($phpVersion->castsNumbersToStringsOnLooseComparison()) {
 			$isNumber = new UnionType([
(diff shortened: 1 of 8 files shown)


src/Type/ComparisonHelper.php
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		// @phpstan-ignore-next-line
+		return new ConstantBooleanType($leftType->getValue() == $rightType->getValue()); // phpcs:ignore
+	}
+
+	/**
+	 * @param callable(Type, Type): BooleanType $valueComparisonCallback
+	 */
+	public static function compareConstantArrayType(ConstantArrayType $leftType, ConstantArrayType $rightType, callable $valueComparisonCallback): BooleanType
```
ondrejmirtes: You've missed my point. This is supposed to be in `ConstantArrayType::looseCompare()`.

ondrejmirtes (changes requested):

src/Type/ComparisonHelper.php
staabm: So should I duplicate the logic?

Its also called from resolveIdentical.

staabm: Or is the goal to also reduce resolveIdentical to just

isSuperTypeOf()->no() ?

ondrejmirtes: > Or is the goal to also reduce resolveIdentical to just

Yeah, probably, in the future. I'm fine if momentarily the code is a bit duplicated.

staabm: just duplicated the method and removed the callable arg to improve readability

staabm:
should be good to go

staabm:
rebased

PR #2229, opened 2023-02-08
Skip `testEditorUrlWithRelativePath()` test on PHPStorm console

staabm:
Skip the only test which does not run successfully in the PHPStorm console.

see https://github.com/phpstan/phpstan/discussions/8854
I am not the only one with this problem, see https://github.com/phpstan/phpstan-src/pull/1927#discussion_r1009129184

### DIFF
--- a/tests/PHPStan/Command/ErrorFormatter/TableErrorFormatterTest.php
+++ b/tests/PHPStan/Command/ErrorFormatter/TableErrorFormatterTest.php
@@ -217,6 +217,10 @@ class TableErrorFormatterTest extends ErrorFormatterTestCase
 
 	public function testEditorUrlWithRelativePath(): void
 	{
+		if ('JetBrains-JediTerm' === getenv('TERMINAL_EMULATOR')) {
+			$this->markTestSkipped('PHPStorm console does not support links in console.');
+		}
+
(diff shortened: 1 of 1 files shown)


tests/PHPStan/Command/ErrorFormatter/TableErrorFormatterTest.php
```diff
@@ -217,6 +217,10 @@ public function testEditorUrlWithTrait(): void
 
 	public function testEditorUrlWithRelativePath(): void
 	{
+		if ('JetBrains-JediTerm' === getenv('TERMINAL_EMULATOR')) {
+			$this->markTestSkipped('PHPStorm console does not support links in console.');
```
ondrejmirtes: 1) I don't like yoda conditions
2) It's `PhpStorm`.

ondrejmirtes (changes requested):

tests/PHPStan/Command/ErrorFormatter/TableErrorFormatterTest.php
staabm: thanks, fixed

staabm:
I verified this change makes the phpunit job run successfully on my machine.

ondrejmirtes:
Thank you.

PR #2232, opened 2023-02-12
Draft: hide class-level template types from static members

jiripudil:
This is a solution I was talking about [previously](https://github.com/phpstan/phpstan-src/pull/2075#issuecomment-1410638235).

The failing test discovered exactly the kind of class-string magic I was afraid people might be doing. I don't know how many actual real-life use cases there are, but it puts doubt in my mind about this.

Nevertheless, I still firmly believe this is a must for static properties, which, sadly, I have no idea how to cover in this way.

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -31,3 +31,4 @@ parameters:
 		closureDefaultParameterTypeRule: true
 		newRuleLevelHelper: true
 		instanceofType: true
+		disallowClassLevelTemplatesInStatic: true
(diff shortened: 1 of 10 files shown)


ondrejmirtes:
Added a few commits so that this is ready for static methods if we decide to go with it :) But yeah, the CarFactoryProcessor example is somewhat reasonable even for static method.

About static properties - yeah, that might not be trivial at all :/ Because when retrieving PHPDocs, FileTypeMapper is handed these parameters:

```php
	public function getResolvedPhpDoc(
		?string $fileName,
		?string $className,
		?string $traitName,
		?string $functionName,
		string $docComment,
	): ResolvedPhpDocBlock
```

We don't know when we're getting a PHPDoc for a property, they're not stored separately :/

PR #2233, opened 2023-02-13
iterator_*() accepts array since PHP 8.2

takaram:
`iterator_to_array()` and `iterator_count()` accept an array since PHP 8.2.
https://wiki.php.net/rfc/iterator_xyz_accept_array


ondrejmirtes:
Duplicate of https://github.com/phpstan/phpstan-src/pull/1831, see the comments there.

PR #2236, opened 2023-02-14
Add non regression test for bug 5288

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/5288
(But the close need to be manually since default branch is 1.10.x)


staabm:
> But the close need to be manually since default branch is 1.10.x

It will be automatically closed, when merged from 1.9.x into 1.10.x

ondrejmirtes:
> It will be automatically closed, when merged from 1.9.x into 1.10.x

It won't

ondrejmirtes:
There's no reason this test has to be added in 1.9.x. You can submit it for 1.10.x, solving two issues at once:

1) There will be no conflicts when merging
2) The issue will get closed automatically

VincentLanglet:
> There's no reason this test has to be added in 1.9.x. You can submit it for 1.10.x, solving two issues at once:
> 
> 1. There will be no conflicts when merging
> 2. The issue will get closed automatically

So bugfix+tests on 1.9.x
but non-regression-test alones on 1.10.x.

Sure, no problem. I changed the base branch

ondrejmirtes (changes requested):
Test against the wrong rule

VincentLanglet:
> Test against the wrong rule

Indeed... Might be better now.

ondrejmirtes:
Thank you.

PR #2238, opened 2023-02-15
Remove unneded code in `ConstantArrayType::hasOffsetValueType()`

herndlm:
was added in 28c5ff61, but everything seems to be still fine without it. not sure yet if this changes something performance-wise, I hope CI tells me

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -548,9 +548,6 @@ class ConstantArrayType extends ArrayType implements ConstantType
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
 		$offsetType = $offsetType->toArrayKey();
-		if ($offsetType instanceof UnionType) {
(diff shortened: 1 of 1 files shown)


herndlm:
tried to benchmark it, but my system is not good.. I think there's basically no performance difference

before
```sh
hyperfine 'make phpstan'
Benchmark 1: make phpstan
  Time (mean ± σ):     50.324 s ±  4.197 s    [User: 328.346 s, System: 22.814 s]
  Range (min … max):   47.511 s … 59.977 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```
I re-ran this 3 times, but always had some outliers and gave up..

after
```sh
$ hyperfine 'make phpstan'
Benchmark 1: make phpstan
  Time (mean ± σ):     48.054 s ±  0.398 s    [User: 326.638 s, System: 22.087 s]
  Range (min … max):   47.398 s … 48.890 s    10 runs
```

ondrejmirtes:
Yeah, that's a pretty old commit :) Thank you.

PR #2239, opened 2023-02-15
Update param and return type for `db2_autocommit()`

phansys:
Prior to version ibm_db2 2.1, the `db2_autocommit()` function was expecting a boolean as argument 2.

See:
* https://www.php.net/manual/en/function.db2-autocommit.php
* https://bugs.php.net/bug.php?id=77591
* php/pecl-database-ibm_db2@e9fb4b173046ab4fa1bb69125adf3de8d3818b97
* php/doc-en#2300

Additionally, the return type of this function can be `DB2_AUTOCOMMIT_OFF|DB2_AUTOCOMMIT_ON|bool` depending on the received arguments:
>When db2_autocommit() receives only the connection parameter, it returns the current state of AUTOCOMMIT for the requested connection as an integer value. A value of DB2_AUTOCOMMIT_OFF indicates that AUTOCOMMIT is off, while a value of DB2_AUTOCOMMIT_ON indicates that AUTOCOMMIT is on.
>
>When db2_autocommit() receives both the connection parameter and autocommit parameter, it attempts to set the AUTOCOMMIT state of the requested connection to the corresponding state. Returns true on success or false on failure.

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -1644,7 +1644,7 @@ return [
 'DateTimeZone::getTransitions' => ['list<array{ts: int, time: string, offset: int, isdst: bool, abbr: string}>', 'timestamp_begin='=>'int', 'timestamp_end='=>'int'],
 'DateTimeZone::listAbbreviations' => ['array<string, list<array{dst: bool, offset: int, timezone_id: string|null}>>'],
 'DateTimeZone::listIdentifiers' => ['list<string>', 'what='=>'int', 'country='=>'string'],
-'db2_autocommit' => ['mixed', 'connection'=>'resource', 'value='=>'int'],
+'db2_autocommit' => ['DB2_AUTOCOMMIT_OFF|DB2_AUTOCOMMIT_ON|bool', 'connection'=>'resource', 'value='=>'DB2_AUTOCOMMIT_OFF|DB2_AUTOCOMMIT_ON'],
 'db2_bind_param' => ['bool', 'stmt'=>'resource', 'parameter_number'=>'int', 'variable_name'=>'string', 'parameter_type='=>'int', 'data_type='=>'int', 'precision='=>'int', 'scale='=>'int'],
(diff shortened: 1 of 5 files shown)


resources/functionMap.php
```diff
@@ -1644,7 +1644,7 @@
 'DateTimeZone::getTransitions' => ['list<array{ts: int, time: string, offset: int, isdst: bool, abbr: string}>', 'timestamp_begin='=>'int', 'timestamp_end='=>'int'],
 'DateTimeZone::listAbbreviations' => ['array<string, list<array{dst: bool, offset: int, timezone_id: string|null}>>'],
 'DateTimeZone::listIdentifiers' => ['list<string>', 'what='=>'int', 'country='=>'string'],
-'db2_autocommit' => ['mixed', 'connection'=>'resource', 'value='=>'int'],
+'db2_autocommit' => ['DB2_AUTOCOMMIT_OFF|DB2_AUTOCOMMIT_ON|bool', 'connection'=>'resource', 'value='=>'DB2_AUTOCOMMIT_OFF|DB2_AUTOCOMMIT_ON'],
```
ondrejmirtes, line 1647: What does the function do if it receives something that isn't `\DB2_AUTOCOMMIT_OFF|\DB2_AUTOCOMMIT_ON`? Like `1337` for example?

ondrejmirtes (changes requested):

resources/functionMap.php
phansys: A warning is emitted and the function returns `false`:
```
Warning: db2_autocommit(): value must be one of DB2_AUTOCOMMIT_ON or DB2_AUTOCOMMIT_OFF
```
See https://github.com/php/pecl-database-ibm_db2/blob/4eebfe146a873e25af26fbca30f480747cd237af/ibm_db2.c#L3522-L3525.

ondrejmirtes:
Thank you.

PR #2240, opened 2023-02-16
Fix comparison of ThisType and GenericObjectType with equal class-names

staabm:
closes https://github.com/phpstan/phpstan/issues/8879

### DIFF
--- a/src/Type/Generic/GenericObjectType.php
+++ b/src/Type/Generic/GenericObjectType.php
@@ -15,6 +15,7 @@ use PHPStan\Type\CompoundType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\ObjectType;
+use PHPStan\Type\ThisType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeWithClassName;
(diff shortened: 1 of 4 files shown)


src/Type/Generic/GenericObjectType.php
```diff
@@ -121,6 +122,11 @@ private function isSuperTypeOfInternal(Type $type, bool $acceptsContext): Trinar
 			return $nakedSuperTypeOf;
 		}
 
+		if ($type instanceof ThisType && $type->getClassName() === $this->getClassName()) {
+			$type = $type->getStaticObjectType();
+			$nakedSuperTypeOf = TrinaryLogic::createYes();
```
staabm, line 127: Initially I tried making `parent::isSuperTypeOf($type);` return `TrinaryLogic::createYes()`, but thats no possible because that would make the `TypeCombinator` combine types like `$this(TraitInstanceOf\FinalTrait2Class)&TraitInstanceOf\FinalTrait2Class` into `$this(TraitInstanceOf\FinalTrait2Class)` which makes other tests fail.

therefore I figured special casing here seems to be the way to go.

staabm:
issue was resolved by reverting the initial implementation.

PR #2241, opened 2023-02-17
Give class template context to conditional checker

KmeCnin:
This PR fixes bug reported here: https://github.com/phpstan/phpstan/issues/8609

### DIFF
--- a/src/Rules/PhpDoc/MethodConditionalReturnTypeRule.php
+++ b/src/Rules/PhpDoc/MethodConditionalReturnTypeRule.php
@@ -6,6 +6,8 @@ use PhpParser\Node;
 use PHPStan\Analyser\Scope;
 use PHPStan\Node\InClassMethodNode;
 use PHPStan\Rules\Rule;
+use PHPStan\Type\Generic\TemplateTypeMap;
+use function array_merge;
 use function count;
 
 /**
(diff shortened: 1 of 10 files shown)


tests/PHPStan/Generics/data/bug8609.php
```diff
@@ -0,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+namespace Generics\Bug8609;
```
staabm: it seems you are adding a new test-case here, but you did not analyze this file from any of the *Test classes.
you need to make sure that this file is analyzed from the correct *Test class - the one which is testing the rule which emits the error you are trying to fix

src/Rules/PhpDoc/MethodConditionalReturnTypeRule.php
```diff
@@ -31,7 +31,12 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
-		return $this->helper->check($variants[0]);
+		$variant = $variants[0];
+		$templateTypeMap = $node->getClassReflection()->getTemplateTypeMap()->union(
```
ondrejmirtes: `union` isn't the best operation for this. I suggest `array_merge` instead, like here: https://github.com/phpstan/phpstan-src/blob/b250a8bf799b6d22b979260ecfa7539f16449a0f/src/Analyser/NameScope.php#L164-L167

tests/PHPStan/Generics/data/bug8609.php
ondrejmirtes: 1) File name needs to be in `bug-8609.php` format.
2) File needs to be analysed by tests/PHPStan/Rules/PhpDoc/MethodConditionalReturnTypeRuleTest.php.

ondrejmirtes (changes requested):
Additionally, we need to test that this actually works. A new file tested in NodeScopeResolverTest would be nice.

ondrejmirtes:
Finished it myself :) Thank you!

ondrejmirtes (approved):

PR #2244, opened 2023-02-19
Solve issue 8917 with array_column

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8917


herndlm:
Phew, the array column extension also looks multiple times more complicated than I thought it would 😅

VincentLanglet:
> Phew, the array column extension also looks multiple times more complicated than I thought it would 😅

And the extension could be improve for a better optional keys support like
https://phpstan.org/r/ec725950-6493-4079-979f-9b753f93985f

But that's another issue.

ondrejmirtes:
Thank you!

PR #2248, opened 2023-02-21
Fix Bug 8932 with match and treatPhpDocTypesAsCertain

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8932


ondrejmirtes:
Thank you.

PR #2250, opened 2023-02-22
Fix bug 8937 with match and treatPhpDocTypesAsCertain

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8937

I dunno if it's the best fix, but seems like it's solving both https://github.com/phpstan/phpstan/issues/8932 and https://github.com/phpstan/phpstan/issues/8937.

### DIFF
--- a/src/Rules/Comparison/MatchExpressionRule.php
+++ b/src/Rules/Comparison/MatchExpressionRule.php
@@ -44,12 +44,13 @@ class MatchExpressionRule implements Rule
 	{
 		$matchCondition = $node->getCondition();
 		$matchConditionType = $scope->getType($matchCondition);
-		$nextArmIsDead = false;
+		$nextArmIsDeadForType = false;
+		$nextArmIsDeadForNativeType = false;
 		$errors = [];
 		$armsCount = count($node->getArms());
 		$hasDefault = false;
 		foreach ($node->getArms() as $i => $arm) {
-			if ($nextArmIsDead) {
+			if ($nextArmIsDeadForNativeType || $nextArmIsDeadForType && $this->treatPhpDocTypesAsCertain) {
 				if (!$this->disableUnreachable) {
 					$errors[] = RuleErrorBuilder::message('Match arm is unreachable because previous comparison is always true.')->line($arm->getLine())->build();
 				}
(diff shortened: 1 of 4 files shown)


src/Rules/Comparison/MatchExpressionRule.php
```diff
@@ -65,13 +66,23 @@ public function processNode(Node $node, Scope $scope): array
 					$matchCondition,
 					$armCondition->getCondition(),
 				);
-				$armConditionResult = ($this->treatPhpDocTypesAsCertain ? $armConditionScope->getType($armConditionExpr) : $armConditionScope->getNativeType($armConditionExpr));
```
VincentLanglet, line 68: I split this in `nextArmIsDeadForType` and `nextArmIsDeadForNativeType` because
- `Match arm is unreachable` should rely on `$nextArmIsDeadForNativeType || $nextArmIsDeadForType && $this->treatPhpDocTypesAsCertain` in order to not be reported in the issue 8932.
-  `Match expression does not handle remaining` should rely only on `nextArmIsDeadForType` in order to not be reported in the issue 8937.

This way extra checks are allowed when using `treatPhpDocTypesAsCertain: false`, but phpstan are not complaining if they are missing.

src/Rules/Comparison/MatchExpressionRule.php
```diff
@@ -44,12 +44,13 @@ public function processNode(Node $node, Scope $scope): array
...
 		$matchConditionType = $scope->getType($matchCondition);
-		$nextArmIsDead = false;
+		$nextArmIsDeadForType = false;
+		$nextArmIsDeadForNativeType = false;
 		$errors = [];
 		$armsCount = count($node->getArms());
 		$hasDefault = false;
 		foreach ($node->getArms() as $i => $arm) {
-			if ($nextArmIsDead) {
+			if ($nextArmIsDeadForNativeType || $nextArmIsDeadForType && $this->treatPhpDocTypesAsCertain) {
```
ondrejmirtes: Please use parentheses to disambiguate the condition. And please rebase on top of 1.10.x so we can see whether there are new failures. Thanks.

ondrejmirtes (changes requested):

src/Rules/Comparison/MatchExpressionRule.php
VincentLanglet: It's rebased with parentheses added

ondrejmirtes:
Thank you.

PR #2251, opened 2023-02-22
Detect not existing classes in @method annotations

alfredbez:
Monday I asked in the #static-analysis Slack channel of the Symfony Devs workspace if PHPStan is able to detect not-existing classes in the return type of `@method` annotations. I provided [this example](https://phpstan.org/r/9c76cf12-bd31-478d-95db-d3b97f71787a), but looks like this is not handled right now, so here's the Pull Request for that.

### DIFF
--- a/src/Rules/PhpDoc/MethodAnnotationReturnTypeClassDoesNotExistRule.php
+++ b/src/Rules/PhpDoc/MethodAnnotationReturnTypeClassDoesNotExistRule.php
@@ -1,0 +1,65 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\PhpDoc;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\InClassNode;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\ClassCaseSensitivityCheck;
+use PHPStan\Rules\ClassNameNodePair;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function array_merge;
+use function sprintf;
+
+/**
+ * @implements Rule<InClassNode>
+ */
+class MethodAnnotationReturnTypeClassDoesNotExistRule implements Rule
+{
+
+	public function __construct(
+		private ReflectionProvider $reflectionProvider,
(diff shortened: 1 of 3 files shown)


src/Rules/PhpDoc/MethodAnnotationReturnTypeClassDoesNotExistRule.php
```diff
@@ -0,0 +1,50 @@
...
+		$errors = [];
+		foreach ($classReflection->getMethodTags() as $methodName => $methodTag) {
+			foreach ($methodTag->getReturnType()->getObjectClassNames() as $className) {
+				if (class_exists($className)) {
+					continue;
+				}
+
+				$errors[] = RuleErrorBuilder::message(
+					sprintf(
+						'PHPDoc tag @method %s() has invalid return type "%s", class does not exist',
```
ondrejmirtes: There are many existing rules you can draw inspiration from. They have one thing in common - they all call ClassCaseSensitivityCheck.

The one that's probably the closest is MixinRule. The ExistingClassesInPropertiesRule is also pretty close. That's gonna give you some ideas how to implement this.

Also, some more points:

1) Your rule isn't registered for any level.
2) You need to check return type, and parameter types too.
3) There should be similar rules for `@property` and also type aliases.

The last point can be done in a separate PR.

ondrejmirtes (changes requested):

src/Rules/PhpDoc/MethodAnnotationReturnTypeClassDoesNotExistRule.php
alfredbez: 1. What would be a good level for that check?
2. You mean `@return` and `@param`, right?

I pushed some changes after looking at the examples. Sorry if something is off, I'm new to the codebase

ondrejmirtes: You're checking only `$methodTag->getReturnType()`, you need to check `$methodTag->getParameters()` too.

Some things are still missing in your rule, like UnresolvableTypeHelper, and checking that the referenced class isn't a trait. As I said, please inspire yourself by the existing rules more :)

As for the level: level 2 would be great. But the rule needs to be registered only for bleedingEdge. Inspire yourself by the existing `conditionalTags` sections :) Thanks.

PR #2253, opened 2023-02-24
[RFC] Allow resources to document its type

phansys:
Do you think something like this could be possible? https://phpstan.org/r/f11c30dc-3a4d-4f95-abea-c9627460110f

### DIFF
--- a/src/Type/ResourceType.php
+++ b/src/Type/ResourceType.php
@@ -38,8 +38,25 @@ class ResourceType implements Type
 	{
 	}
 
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Resources are on their way out, being replaced by classes like CurlHandle. So I'm not interested in improving them. Thanks.

phansys:
Thank you for the feedback.
Just for the records, I found this issue including the status of the migration to objects: php/php-tasks#6.

PR #2254, opened 2023-02-24
Remove 3 leading spaces in PHAR

ruudk:
Fixes https://github.com/phpstan/phpstan/issues/8952

I think the problem comes from the nowdoc related to Box/Rector.

### DIFF
--- a/src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
+++ b/src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
@@ -59,20 +59,12 @@ class BaselinePhpErrorFormatter
 			ksort($fileErrorsCounts, SORT_STRING);
 
 			foreach ($fileErrorsCounts as $message => $count) {
-				$template = <<<'PHP'
-$ignoreErrors[] = [
-	'message' => %s,
-	'count' => %d,
-	'path' => __DIR__ . %s,
-];
-PHP;
 				$php .= sprintf(
-					$template,
+					"\$ignoreErrors[] = [\n\t'message' => %s,\n\t'count' => %d,\n\t'path' => __DIR__ . %s,\n];\n",
 					var_export(Helpers::escape('#^' . preg_quote($message, '#') . '$#'), true),
 					var_export($count, true),
 					var_export(Helpers::escape($file), true),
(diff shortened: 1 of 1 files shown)


ruudk:
> As a CI artifact in a PR here you get the compiled PHAR so you could easily debug the end result this way.

@ondrejmirtes I must be blind but I don't see the artifact.

staabm:
in the "compile phar" action

grafik

 at the bottom

https://github.com/phpstan/phpstan-src/actions/runs/4260429654

grafik

sorry for the huge screenshots 😅

ruudk:
@staabm Thanks!

ruudk:
Oke, the generated PHAR now looks like this:
```php
            ksort($fileErrorsCounts, SORT_STRING);
            foreach ($fileErrorsCounts as $message => $count) {
                $php .= '$ignoreErrors[] = [';
                $php .= "\n";
                $php .= "\t'message' => " . var_export(Helpers::escape('#^' . preg_quote($message, '#') . '$#'), \true) . ',';
                $php .= "\n";
                $php .= "\t'count' => " . $count . ',';
                $php .= "\n";
                $php .= "\t'path' => __DIR__ . " . var_export(Helpers::escape($file), \true) . ',';
                $php .= "\n";
                $php .= '];';
                $php .= "\n";
            }
```

I think this looks good.

I also noticed that the generated baseline did not have a line ending, so I added that.
Screenshot 2023-02-24 at 09 40 40@2x

This is ready.

ondrejmirtes:
CS failed: https://github.com/phpstan/phpstan-src/actions/runs/4260736633/jobs/7414242127

mvorisek:
When PHP-CS-Fixer is present in the project, what about invoking it and keeping the phpstan PHP file CS always fixed?

ruudk:
Feels like a waste of resources especially on large baselines with 15k lines. PHP CS Fixer is not the fastest.... 

I will fix the CS in this PR so that we can merge the actual fix.

ruudk:
@ondrejmirtes CS is fixed. And I verified that the new PHAR generates the baseline correctly.

ondrejmirtes:
Thank you!

PR #2255, opened 2023-02-24
Add support for a not-deprecated annotation

stof:
This annotation allows disabling the inheritance of method deprecations when a child class does not want to deprecate its method even though it corresponds to a deprecated parent method.

Solves https://github.com/phpstan/phpstan/discussions/7422


ondrejmirtes:
Thank you!

VincentLanglet:
If I understand correctly the code 
```
/**
 * @deprecated
 * @not-deprecated
 */
```
Will still consider the method as deprecated, but might be confusing.

Should phpstan report that those annotations are incompatible ?

stof:
@VincentLanglet this might make sense as a rule indeed.

PR #2256, opened 2023-02-25
Support json format in dump-parameters cli command

staabm:
I want to inspect the phpstan options/configuration in a machine readable format.

I want to run other tools on the paths configured within phpstan


ondrejmirtes:
Neon is already machine-readable format. You can read it with an open-source package nette/neon.

PR #2257, opened 2023-02-25
Add `FilterVarArrayDynamicReturnTypeExtension`

zonuexe:
Add a return type extension for [`filter_var_array()`](https://www.php.net/filter_var_array) and [`filter_input_array()`](https://www.php.net/filter_input_array).

This function is defined with the following signature:

```php
filter_var_array(array $array, array|int $options = FILTER_DEFAULT, bool $add_empty = true): array|false|null
```

This function supports three types of filters:

```php
// bit flag
$values = filter_var_array($_POST, FILTER_VALIDATE_INT);

// simple array
$values = filter_var_array($_POST, [
    'key' => FILTER_VALIDATE_INT,
]);

// complex array
$values = filter_var_array($_POST,[
    'key' => [
        'filter' => FILTER_VALIDATE_INT,
        'options' => ['min_range' => 1, 'max_range' => 10],
    ],
]);
```

The array keys returned from `filter_var_array()` change depending on how the filter is passed, so the implementation is complex.

| filter passed by | key is determined by |
|------------------|----------------------|
| bit flag         | `$options`           |
| simple array     | `$array`             |
| complex array    | `$array`             |

This PR was implemented thanks to @herndlm's `FilterFunctionReturnTypeHelper` https://github.com/phpstan/phpstan-src/pull/2190 and https://github.com/phpstan/phpstan-src/pull/2010.

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -1467,6 +1467,11 @@ services:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
 
 	-
+		class: PHPStan\Type\Php\FilterVarArrayDynamicReturnTypeExtension
+		tags:
+			- phpstan.broker.dynamicFunctionReturnTypeExtension
+
+	-
 		class: PHPStan\Type\Php\GetCalledClassDynamicReturnTypeExtension
 		tags:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
--- a/src/Type/Php/FilterFunctionReturnTypeHelper.php
+++ b/src/Type/Php/FilterFunctionReturnTypeHelper.php
@@ -11,6 +11,7 @@ use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\BooleanType;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantFloatType;
 use PHPStan\Type\Constant\ConstantIntegerType;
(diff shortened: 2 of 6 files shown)


herndlm:
I was hoping it might be possible to reuse that thing, but didn't plan it through :) feel also free to change it if needed. Reminds me that I also still need to finish filter_input..

zonuexe:
Since the processing for `filter_input_array()` is almost common to `filter_var_array()`, I started moving the code into `FilterFunctionReturnTypeHelper`.

herndlm:
I started the filter_input support via https://github.com/phpstan/phpstan-src/pull/2010 but this is not the latest state and I wasn't happy yet with the local refactor. Will try to push it soon.
Or do you also have already something (in mind)?
Moving the part that deals with arrays into the helper makes sense to me 👍

zonuexe:
@herndlm ~~No changes planned except for the two methods I added to the `FilterFunctionReturnTypeHelper`.~~

zonuexe:
I intended to implement `FilterInputArrayDynamicReturnTypeExtension` as a pared-down version of `FilterVarArrayDynamicReturnTypeExtension`, but it's probably simpler to support both functions in one extension. So the method move to `FilterFunctionReturnTypeHelper` has been reverted.

herndlm:
a missing interesting test case for `filter_input_array` might be one with the `FILTER_NULL_ON_FAILURE` flag set.
> An array value will be false if the filter fails, or null if the variable is not set. Or if the flag FILTER_NULL_ON_FAILURE is used, it returns false if the variable is not set and null if the filter fails.

I handled this incorrectly in https://github.com/phpstan/phpstan-src/pull/2010 initially but added a new method in the helper now that should deal with this and maybe you can make use of it here too.

herndlm:
hi @zonuexe, the filter_input stuff has been merged, maybe it is useful here in some way

zonuexe:
@herndlm Thanks for the reminder!

zonuexe:
@ondrejmirtes Could you please review this PR?

ondrejmirtes:
Thank you!

PR #2258, opened 2023-02-26
Add non regression test for 4969

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/4969


ondrejmirtes:
Thank you.

PR #2259, opened 2023-02-26
Add non regression test for 5695

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/5695


ondrejmirtes:
Thank you.

PR #2260, opened 2023-02-26
Fix password_hash return type

VincentLanglet:
Cf https://3v4l.org/DMv87

Closes https://github.com/phpstan/phpstan/issues/5978

### DIFF
--- a/tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
+++ b/tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
@@ -882,4 +882,14 @@ class StrictComparisonOfDifferentTypesRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/bug-8776-2.php'], []);
 	}
 
+	public function testBug5978(): void
+	{
+		if (PHP_VERSION_ID >= 80000) {
+			$this->markTestSkipped('Test does not run on PHP >= 8.0.');
+		}
+
+		$this->checkAlwaysTrueStrictComparison = true;
+		$this->analyse([__DIR__ . '/data/bug-5978.php'], []);
+	}
+
 }
--- a/tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
+++ b/tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
@@ -885,11 +885,22 @@ class StrictComparisonOfDifferentTypesRuleTest extends RuleTestCase
 	public function testBug5978(): void
 	{
 		if (PHP_VERSION_ID >= 80000) {
-			$this->markTestSkipped('Test does not run on PHP >= 8.0.');
+			$expectedErrors = [
+				[
+					'Strict comparison using === between string and false will always evaluate to false.',
+					7,
+				],
+				[
+					'Strict comparison using === between string and null will always evaluate to false.',
+					7,
+				],
+			];
+		} else {
+			$expectedErrors = [];
 		}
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
This isn't great. If the inputs are correct, users shouldn't be forced to handle `null`.

VincentLanglet:
After some try, password_hash returns null
- if the second parameter is not `PASSWORD_DEFAULT| PASSWORD_BCRYPT| PASSWORD_ARGON2I| PASSWORD_ARGON2ID`
- if the second parameter is `PASSWORD_DEFAULT| PASSWORD_BCRYPT` and the salt option is not a 22-length string 
- if the second parameter is `PASSWORD_DEFAULT|PASSWORD_BCRYPT` and the cost option is not int<4, 31>
- if the second parameter is `PASSWORD_ARGON2I| PASSWORD_ARGON2ID` and `ext-libargon2` is not installed
- if the second parameter is `PASSWORD_ARGON2I| PASSWORD_ARGON2ID` and the `memory_cost/time_cost/threads` is not a correct int but I don't know all the allowed values.

This won't be an easy dynamicReturnTypeExtension...

> This isn't great. If the inputs are correct, users shouldn't be forced to handle `null`.

I understand, but currently PHP 7.4 users are already forced to handle `false`
https://phpstan.org/r/5464625f-1925-4c4d-a8c5-7aea19ace58f

When I looked at
https://stackoverflow.com/questions/39729941/php-password-hash-returns-false
https://bugs.php.net/bug.php?id=77218
we find that
> So basically, for PHP ≤ 7.3, the function returned FALSE for a
failing implementation, and NULL for invalid parameters in
combination with a warning (the latter likely according to the
general convention to return NULL on ZPP failures).

So even without dynamicReturnTypeExtension it would be more useful (and correct ?) for `functionMap_php74delta.php` to have
```
- 'password_hash' => ['string|false', 'password'=>'string', 'algo'=>'string|null', 'options='=>'array'],
+ 'password_hash' => ['string|null', 'password'=>'string', 'algo'=>'string|null', 'options='=>'array'],
```
while I can let 
```
'password_hash' => ['string|false', 'password'=>'string', 'algo'=>'string|null', 'options='=>'array'],
```
for the original `functionMap_php`.

Would you be ok with this first step ?

ondrejmirtes:
I'd just do benevolent union of string|false|null and call it a day.

VincentLanglet:
> I'd just do benevolent union of string|false|null and call it a day.

Nice idea. Done :)

ondrejmirtes (changes requested):
You're closing an issue with this PR - needs a regression test.

tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
```diff
@@ -882,4 +882,14 @@ public function testBug8776Part2(): void
 		$this->analyse([__DIR__ . '/data/bug-8776-2.php'], []);
 	}
 
+	public function testBug5978(): void
+	{
+		if (PHP_VERSION_ID >= 80000) {
+			$this->markTestSkipped('Test does not run on PHP >= 8.0.');
```
ondrejmirtes: Instead, you could assert different errors on PHP 7.x vs. 8.x.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
VincentLanglet: Indeed, that's better

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2261, opened 2023-02-26
fix detecting @var classes imported

Odachannnn:
closes https://github.com/phpstan/phpstan/issues/8950

This bug report is about the situation when `define` is written before the `use` statements, but I found that it occurs when other functions, example `sleep()`, `unpack()`, etc.


Odachannnn:
another bugs were occurred, so close this pull request for a moment

PR #2263, opened 2023-02-26
Do not treat definition of functions and class-likes as unreachable

takaram:
PHPStan treats the statement right after always-terminating statement, like `return` or `exit`, as unreachable statement.
With this PR, definition of classes, interfaces, traits and functions on top level will not marked as unreachable.
Nop nodes will be ignored as well.

I haven't implemented [PHP's early binding](https://www.npopov.com/2021/10/20/Early-binding-in-PHP.html) mechanism since it is so complicated.
This may cause some false negative on code [like this](https://3v4l.org/dnK0M), but I think the advantage of decreasing false positive is larger.

Closes phpstan/phpstan#4002
Closes phpstan/phpstan#8966
Closes phpstan/phpstan#8319 (although the early binding mechanism is not completely compatible with [that of PHP](https://www.npopov.com/2021/10/20/Early-binding-in-PHP.html))

### DIFF
--- a/tests/PHPStan/Rules/DeadCode/UnreachableStatementRuleTest.php
+++ b/tests/PHPStan/Rules/DeadCode/UnreachableStatementRuleTest.php
@@ -149,6 +149,34 @@ class UnreachableStatementRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/bug-8620.php'], []);
 	}
 
+	public function testBug4002(): void
+	{
+		$this->treatPhpDocTypesAsCertain = true;
+		$this->analyse([__DIR__ . '/data/bug-4002.php'], []);
+		$this->analyse([__DIR__ . '/data/bug-4002-2.php'], []);
+		$this->analyse([__DIR__ . '/data/bug-4002-3.php'], [
+			[
+				'Unreachable statement - code above always terminates.',
+				13,
+			],
+		]);
+		$this->analyse([__DIR__ . '/data/bug-4002-4.php'], [
+			[
+				'Unreachable statement - code above always terminates.',
+				9,
+			],
+		]);
+		$this->analyse([__DIR__ . '/data/bug-4002_class.php'], []);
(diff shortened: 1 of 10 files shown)


antonsacred:
What is the blockers to merge it?

herndlm:
Oh interesting, this looks super close to what I did in https://github.com/phpstan/phpstan-src/pull/2279, which explains the conflicts, which makes this not mergable. I wasn't even aware of this PR when I did those changes. It also closed one of the linked issues here.

Most likely it makes sense that somebody checks what's still different / missing and adds those missing changes or adapts my changes to also fix the other issue.

I would gladly do that, but can't because I'm traveling.

takaram:
I've just rebased the branch and merged @herndlm's changes and mine.

tests/PHPStan/Rules/DeadCode/UnreachableStatementRuleTest.php
```diff
@@ -149,6 +149,34 @@ public function testBug8620(): void
...
+		]);
+		$this->analyse([__DIR__ . '/data/bug-4002-4.php'], [
+			[
+				'Unreachable statement - code above always terminates.',
+				9,
+			],
+		]);
+		$this->analyse([__DIR__ . '/data/bug-4002_class.php'], []);
+		$this->analyse([__DIR__ . '/data/bug-4002_interface.php'], []);
+		$this->analyse([__DIR__ . '/data/bug-4002_trait.php'], []);
```
ondrejmirtes, line 201: For each `test*` method, only a single `analyse()` call should be made. So these calls should each have their own test method.

ondrejmirtes (changes requested):
And also please add a note in the PR description that this closes #8319, thanks.

tests/PHPStan/Rules/DeadCode/UnreachableStatementRuleTest.php
takaram: I've split the method.

ondrejmirtes:
Thank you!

PR #2266, opened 2023-02-27
Replace newlines in ignored errors with `\n` and keep every `$ignoreErrors[]` on a single line too

ruudk:
This way, large baselines will have a single ignore error per line. 

This helps with quickly identifying the number of ignored errors (not to be confused with the total count).

### DIFF
--- a/src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
+++ b/src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
@@ -9,7 +9,10 @@ use PHPStan\File\RelativePathHelper;
 use function ksort;
 use function preg_quote;
 use function sprintf;
+use function str_replace;
+use function substr;
 use function var_export;
+use const PHP_EOL;
 use const SORT_STRING;
 
 class BaselinePhpErrorFormatter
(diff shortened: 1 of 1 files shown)


ruudk:
It seems that the `1.10.x` branch is erroring, and therefore the PHAR generation fails to complete. Will have to wait until that's resolved.

ondrejmirtes:
In this case I think whatever’s failing is the cause of this PR. Because the change you made changes the behaviour.

src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
```diff
@@ -77,4 +80,12 @@ public function formatErrors(
 		return 1;
 	}
 
+	/**
+	 * Exports the input to a double-quoted string making sure newlines are converted to `\n`.
+	 */
+	public static function exportDoubleQuoted(string $input): string
+	{
+		return '"' . substr(str_replace([PHP_EOL, '$', '"'], ['\n', '\$', '\"'], var_export($input, true)), 1, -1) . '"';
```
ondrejmirtes, line 88: Oh god, I have no idea what's going on here, and I'm afraid to merge it. Also, it'd be much nicer to escape the thing that's going into `preg_quote` (the most-inner one) instead of the thing that's already PCRE-escaped, and also nette/di-escaped.

ondrejmirtes (changes requested):

src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
ruudk: I tried that, but that doesn't work. 

It's important that `$`  and `"` are escaped to be put inside a double quoted string.

If I replace before putting in `preg_quote` it escapes the escape characters.
Example: https://3v4l.org/O4JGm#v8.1.16

I think this is hard to solve. Closed this PR in favor of https://github.com/phpstan/phpstan-src/pull/2269

I will try to fix the places that produce a multiline error (like https://github.com/phpstan/phpstan-deprecation-rules)

PR #2268, opened 2023-02-28
Add dynamic return type extension for `get_object_vars`

BackEndTea:


src/Reflection/ClassReflection.php
```diff
@@ -528,6 +529,17 @@ public function evictPrivateSymbols(): void
 		$this->getPhpExtension()->evictPrivateSymbols($this->getCacheKey());
 	}
 
+	/**
+	 * @return array<string>
+	 */
+	public function getPropertyNames(): array
```
BackEndTea, line 535: I'm not sure if this is the correct way of doing this, i'm open to suggestions.

tests/PHPStan/Analyser/data/get-object-vars.php
```diff
@@ -0,0 +1,45 @@
...
+{
+
+	public function getExtendedVars(): void
+	{
+		assertType('array{a: int, b: string, c: string}', get_object_vars($this));
+	}
+
+}
+
+function check(Base $base, Base|Other $class, string $input): void
```
BackEndTea, line 37: Need to add testcases and functionality for Enums

tests/PHPStan/Analyser/data/get-object-vars-81.php
```diff
@@ -0,0 +1,37 @@
...
+}
+
+enum NumBacked: int
+{
+	case One = 1;
+	case Two = 2;
+	case Three = 3;
+}
+
+function check(Suit $suit, SuitBacked $backed, NumBacked $num): void
```
BackEndTea, line 30: I personally want aware how get_object vars worked on an enum: https://3v4l.org/6hV36

BackEndTea:
So, the issue now is that there may be 'extra' keys in the array, if the class is not final. (https://3v4l.org/hM8MU)

Is there any way to indicate that the array may have more keys?

jrmajor:
> Is there any way to indicate that the array may have more keys?

@BackEndTea Currently there isn't, see [Psalm 5 announcment](https://psalm.dev/articles/psalm-5) and phpstan/phpstan#8438.

VincentLanglet:
> So, the issue now is that there may be 'extra' keys in the array, if the class is not final. ([3v4l.org/hM8MU](https://3v4l.org/hM8MU))
> 
> Is there any way to indicate that the array may have more keys?

Currently you have to return `null` as soon as you encounter a non final class.

But more important, I think your implementation is wrong (and the right one might be really complex) because uninitialized properties are not in the array https://3v4l.org/XDYC7

PR #2269, opened 2023-03-01
Keep ignored errors on a single line

ruudk:
When the message contains a newline, it still produces multiple lines. But most errors are single line so that's nice.

### DIFF
--- a/src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
+++ b/src/Command/ErrorFormatter/BaselinePhpErrorFormatter.php
@@ -60,7 +60,7 @@ class BaselinePhpErrorFormatter
 
 			foreach ($fileErrorsCounts as $message => $count) {
 				$php .= sprintf(
-					"\$ignoreErrors[] = [\n\t'message' => %s,\n\t'count' => %d,\n\t'path' => __DIR__ . %s,\n];\n",
+					"\$ignoreErrors[] = ['message' => %s, 'count' => %d, 'path' => __DIR__ . %s];\n",
 					var_export(Helpers::escape('#^' . preg_quote($message, '#') . '$#'), true),
 					var_export($count, true),
 					var_export(Helpers::escape($file), true),


ondrejmirtes:
Hi, I don't really know if I want to change anything here. Please consult it with me before you decide to sink a bunch of time into it :)

gnutix:
@ondrejmirtes I think the changes are good, as I mentioned in [here](https://github.com/phpstan/phpstan/issues/8870#issuecomment-1448449280). Any reason why you would not like these changes to be made ? Can this be discussed ?

ondrejmirtes:
I think that code on more lines is more readable for very long lines.

ondrejmirtes:
Anyway, why would you like these changes?

gnutix:
Sure, but no one "reads" the baseline IMHO. Sometimes you have to manipulate it by hand (in which case I use CTRL+F search to find the line, then copy/paste it, and thus the "tab to space" change is welcome), and most of the time you simply review changes that happened it in via `git diff` (because you regenerated the baseline after changing some code in your application), in which case having one error per line is much more compact than multiple lines, especially when a bunch of inspections got removed.
 
And even for long lines, if something got modified, `git diff` will most of the time be able to highlight the part of the line that got changed. And it wraps lines in the terminal.

ondrejmirtes:
In order to manipulate it, you need to read it -> it has to be readable.

I believe that multiple lines make diff more readable too.

I don't want to discuss tabs vs. spaces. I prefer tabs :)

gnutix:
Then I guess there's nothing more to discuss, and these personal preferences will be dealt with using code styling tools like ECS. :shrug:

gnutix:
... except ECS crashes, even with 4G of memory and run *only* on the baseline (28'490 lines in our app, ~ 7000 inspections) :

`php -d memory_limit=4G vendor/bin/ecs check -c easy-coding-standard.php --fix`

> Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 7864352 bytes) in /app/vendor/symplify/easy-coding-standard/vendor/symplify/package-builder/src/Console/Formatter/ColorConsoleDiffFormatter.php on line 70

Well, too bad! Editorconfig it is : 

```
# .editorconfig
[phpstan-baseline.php]
indent_style = tab
tab_width = 4
```

VincentLanglet:
> Sure, but no one "reads" the baseline IMHO.

I personally read it, and even sometimes comment it.
So it's better multi-line for this

gnutix:
Comment it ? So you never regenerate it, or else you'd lose your comments, right ?

PR #2271, opened 2023-03-04
Check filter_input* type param type

herndlm:
Specifying invalid ints there fails with a fatal error since PHP 8, see https://3v4l.org/2QvBP and https://3v4l.org/EP3T8
It would also be wrong on pre PHP 8 but there's no fatal error there at least.

This is the easiest I could come up with to check this, but it feels a bit weird. Any better ideas?

See also https://www.php.net/manual/en/function.filter-input.php and https://www.php.net/manual/en/function.filter-input-array.php
The 2 failing tests seem unrelated.

Some special handling based on PHP version could also be additionally built into e.g. https://github.com/phpstan/phpstan-src/pull/2010 (which can be merged I believe ;))

### DIFF
--- a/resources/functionMap_php80delta_bleedingEdge.php
+++ b/resources/functionMap_php80delta_bleedingEdge.php
@@ -1,0 +1,11 @@
+<?php // phpcs:ignoreFile
+
+return [
+	'new' => [
+		'filter_input' => ['mixed', 'type'=>'INPUT_GET|INPUT_POST|INPUT_COOKIE|INPUT_SERVER|INPUT_ENV', 'variable_name'=>'string', 'filter='=>'int', 'options='=>'array|int'],
+		'filter_input_array' => ['array|false|null', 'type'=>'INPUT_GET|INPUT_POST|INPUT_COOKIE|INPUT_SERVER|INPUT_ENV', 'definition='=>'int|array', 'add_empty='=>'bool'],
+	],
+	'old' => [
+
+	]
+];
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Although it makes the code safer, it can be quite annoying to require users to write a more specific PHPDoc type in their `@param`.

We need to figure out a way how to do these changes only in bleedingEdge. Then we'd also be albe to merge this https://github.com/phpstan/phpstan-src/pull/2163 and this https://github.com/phpstan/phpstan-src/pull/2217.

staabm:
> We need to figure out a way how to do these changes only in bleedingEdge

what do you think about enabling these errors/use more precise signatures only when bleedingEdge and strict-rules are enabled?

herndlm:
makes sense, thx for the inputs! I thought here it might be okay, because nobody ever should call these functions differently. I think it's a bit different compared to the linked PRs, but I get the general problem and point :)

resources/functionMap_php80delta.php
```diff
@@ -41,6 +41,8 @@
 		'date_timezone_set' => ['DateTime', 'object'=>'DateTime', 'timezone'=>'DateTimeZone'],
 		'explode' => ['list<string>', 'separator'=>'non-empty-string', 'str'=>'string', 'limit='=>'int'],
 		'fdiv' => ['float', 'dividend'=>'float', 'divisor'=>'float'],
+		'filter_input' => ['mixed', 'type'=>'INPUT_GET|INPUT_POST|INPUT_COOKIE|INPUT_SERVER|INPUT_ENV', 'variable_name'=>'string', 'filter='=>'int', 'options='=>'array|int'],
```
ondrejmirtes: Please do stricter parameter types in `functionMap_bleedingEdge.php`. If this is only PHP 8+ specific, please create `functionMap_php80delta_bleedingEdge.php`. Thank you!

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2276, opened 2023-03-07
Fix false positive with `array_combine()` on php8

staabm:
closes https://github.com/phpstan/phpstan/issues/9011

test fails without fix:
```
There was 1 failure:

1) PHPStan\Rules\Methods\ReturnTypeRuleTest::testBug9011
--- Expected
+++ Actual
@@ @@
-'
+'16: Method Bug9011\HelloWorld::getX() should return array<string> but returns false.
 '

C:\dvl\Workspace\phpstan-src-staabm\src\Testing\RuleTestCase.php:154
C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Rules\Methods\ReturnTypeRuleTest.php:851
```

----

the [2nd example ](https://github.com/phpstan/phpstan/issues/9011#issuecomment-1458414871) posted in the issue works as expected in my eyes.


ondrejmirtes:
Thank you.

PR #2277, opened 2023-03-07
Prevent false positive on `method_exists()` and `function_exists()`

staabm:
Pre-existing tests have different expectations regarding the "is always true/false" results, then what was discussed in the linked issue.
I wonder when/how we should/can differentiate whether we expect this "is always true/false" error or not.

I just did the trivial version, so we can see how this violates existing expectations and start discussion/feedback from here.

closes https://github.com/phpstan/phpstan/issues/8980

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -11,6 +11,7 @@ use PhpParser\Node\Expr\StaticCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Analyser\TypeSpecifier;
 use PHPStan\Analyser\TypeSpecifierContext;
+use PHPStan\Reflection\Native\NativeFunctionReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\TrinaryLogic;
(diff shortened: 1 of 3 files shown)


staabm:
Maybe there should be a config switch to enable/disable this rule for function/method_exists?

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -79,6 +80,18 @@ public function findSpecifiedType(
...
+				if ($functionName === 'function_exists') {
+					$functionNameType = $scope->getType($node->getArgs()[0]->value);
+
+					foreach ($functionNameType->getConstantStrings() as $constantString) {
+						$functionReflection = $this->reflectionProvider->getFunction(new Node\Name($constantString->getValue()), $scope);
+						if ($functionReflection instanceof NativeFunctionReflection) {
+							// native functions existance depends php configuration,
+							// therefore don't judge them regarding will always exist or not
+							return null;
+						}
```
staabm, line 98: let me propose to not judge about "always exist yes/no", when `function_exists()` is called on a php-src native/php-extension function - these functions might not exist at runtime and code trying to detect such optional dependency should not get flagged IMO.

wdyt?

---

going one step further I think it might also make sense to prevent the "always exists yes/no" error for `method_exists()`, when the class involved is located in the composer vendor/ folder.

the assumption would be that the current project might support different versions of a external dependency and therefore needs to feauture detect the existance of methods.

wdyt?

westonruter:
Should this also account for `property_exists()`?

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
staabm: @ondrejmirtes do you have a opinion on the above comment?

PR #2287, opened 2023-03-13
Configure composer platform constraint everywhere

herndlm:
Adds a php require of ^8.1 and a config.platform.php of `8.1.99` everywhere as it exists in the root composer.json already. This basically tells renovate to use PHP 8.1 instead of the latest reachable one (e.g. 8.2) which could otherwise result in invalid update PRs like https://github.com/phpstan/phpstan-src/pull/2285

### DIFF
--- a/apigen/composer.json
+++ b/apigen/composer.json
@@ -9,5 +9,10 @@
 		"psr-4": {
 			"PHPStan\\ApiGen\\": "src"
 		}
+	},
+	"config": {
+		"platform": {
+			"php": "8.1.99"
+		}
 	}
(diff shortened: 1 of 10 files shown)


ondrejmirtes:
Hi, I disagree with this. phpstan-src uses `config.platform.php` as a hack so that Composer doesn't refuse to install dependencies on PHP 7.x. I don't want to have this key in all composer.json files :)

herndlm:
Ok, understood. There are other ways to configure renovate properly if it bothers you with weird PRs :)

PR #2288, opened 2023-03-13
Ensure consistent file traversal order

dravnic:
File traversal order depends on how files are read from the filesystem by symfony/finder. As well it depends on how paths are defined in configuration file.

We noticed unsorted traversal, as well reverse order in different systems with same code and same phpstan version.

This caused issues if there were different php files with same class names but different definitions. It gave different list of undefined methods depending on which php file was read first which caused issues of keeping our phpstan-baseline consistent.

### DIFF
--- a/tests/PHPStan/Rules/Methods/OverridingMethodRuleTest.php
+++ b/tests/PHPStan/Rules/Methods/OverridingMethodRuleTest.php
@@ -84,7 +84,7 @@ class OverridingMethodRuleTest extends RuleTestCase
 				115,
 			],
 			[
-				'Parameter #1 $size (int) of method OverridingFinalMethod\FixedArray::setSize() is not ' . $contravariantMessage . ' with parameter #1 $size (mixed) of method SplFixedArray::setSize().',
+				'Parameter #1 $size (int) of method OverridingFinalMethod\FixedArray::setSize() is not ' . $contravariantMessage . ' with parameter #1 $size (mixed) of method SplFixedArray<mixed>::setSize().',
 				125,
 			],
 			[


dravnic:
Here is how to reproduce this issue. This was tested on macOS, arm cpu.

I used same phpstan version. One from homebrew the other from docker. It seems that docker container reads file system differently than host from a local file system.

two php files:

a.php:

```php

<?php

class Test {
    function methodA() {
    }
}

$test = new Test();
$test->methodA();

```

b.php:

```php

<?php

class Test {
    function methodB() {
    }
}

$test = new Test();
$test->methodB();

```

Here is the result.

phpstan from docker:

```bash
% docker run -v $PWD:/app -v /tmp:/tmp --rm ghcr.io/phpstan/phpstan --version
PHPStan - PHP Static Analysis Tool 1.10.6
% docker run -v $PWD:/app -v /tmp:/tmp --rm ghcr.io/phpstan/phpstan --debug  
Note: Using configuration file /app/phpstan.neon.
/app/a.php
/app/b.php
 ------ ---------------------------------------------- 
  Line   a.php                                         
 ------ ---------------------------------------------- 
  9      Call to an undefined method Test::methodA().  
 ------ ---------------------------------------------- 

 [ERROR] Found 1 error                                                                                                                    
                                                                                                                        
```

phpstan from homebrew:

```bash
% phpstan --version
PHPStan - PHP Static Analysis Tool 1.10.6
% phpstan --debug                                                  
Note: Using configuration file /Users/dino/projects/test/phpstan.neon.
/Users/dino/projects/test/b.php
/Users/dino/projects/test/a.php

 ------ ---------------------------------------------- 
  Line   b.php                                         
 ------ ---------------------------------------------- 
  9      Call to an undefined method Test::methodB().  
 ------ ---------------------------------------------- 
                                                                                                                        
 [ERROR] Found 1 error                                                                                                  
                                                                                                                        
```

ondrejmirtes:
First and foremost, the root issue here is that you have duplicate classes. Your problem would go away if you get rid of them.

But I understand the appeal of consistent and deterministic baseline. However, I’m not sure your change will fix this issue. This needs to be reproduced first, preferably in https://github.com/phpstan/phpstan/blob/1.10.x/.github/workflows/other-tests.yml

You can run the same project with different OSes, to highlight the issue.

Please reproduce the problem first, then we can continue here.

ondrejmirtes:
First and foremost, the root issue here is that you have duplicate classes. Your problem would go away if you get rid of them.

But I understand the appeal of consistent and deterministic baseline. However, I’m not sure your change will fix this issue. This needs to be reproduced first, preferably in https://github.com/phpstan/phpstan/blob/1.10.x/.github/workflows/other-tests.yml

You can run the same project with different OSes, to highlight the issue.

Please reproduce the problem first, then we can continue here.

dravnic:
Thanks @ondrejmirtes 

The root issue has been fixed, but, yes, it would be great to have a consistent baseline. 

You are right, the traversal order doesn't seem to have impact on this. Maybe it's due to the autoloader. I'll try to explore this further and provide a reproducible scenario.

PR #2290, opened 2023-03-14
Allow autodiscovery of PHP based configuration files

yguedidi:
# Context

- we're PHP developers
- I saw in a tweet that PHPStan already supports PHP format for configuration
- PHPStan supports PHP based baseline file since [1.10.2](https://github.com/phpstan/phpstan/releases/tag/1.10.2)
- [NEON support](https://plugins.jetbrains.com/plugin/7060-neon-support) is broken on recent PHPStorm versions and the plugin seems to not be likely to be updated
- PHPStan config file and its baseline were the only files using the NEON format in my project

# Issue

Without this PR, the PHP format configuration file is already supported but must be passed explicitly in the command line.

# Workarounds

- Use VSCode for NEON format support
- Use a make recipe to not have to type the configuration argument

---

Still, I think it could be good if PHPStan autodiscover PHP based configuration 🙂

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -159,7 +159,7 @@ class CommandHelper
 			})($autoloadFile);
 		}
 		if ($projectConfigFile === null) {
-			foreach (['phpstan.neon', 'phpstan.neon.dist', 'phpstan.dist.neon'] as $discoverableConfigName) {
+			foreach (['phpstan.neon', 'phpstan.neon.dist', 'phpstan.dist.neon', 'phpstan.php', 'phpstan.php.dist', 'phpstan.dist.php', 'phpstan.neon.php', 'phpstan.neon.php.dist', 'phpstan.dist.neon.php'] as $discoverableConfigName) {
 				$discoverableConfigFile = $currentWorkingDirectory . DIRECTORY_SEPARATOR . $discoverableConfigName;
 				if (is_file($discoverableConfigFile)) {
 					$projectConfigFile = $discoverableConfigFile;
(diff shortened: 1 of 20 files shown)


src/Command/CommandHelper.php
```diff
@@ -159,7 +159,7 @@ public static function begin(
 			})($autoloadFile);
 		}
 		if ($projectConfigFile === null) {
-			foreach (['phpstan.neon', 'phpstan.neon.dist', 'phpstan.dist.neon'] as $discoverableConfigName) {
+			foreach (['phpstan.neon', 'phpstan.neon.dist', 'phpstan.dist.neon', 'phpstan.php', 'phpstan.php.dist', 'phpstan.dist.php', 'phpstan.neon.php', 'phpstan.neon.php.dist', 'phpstan.dist.neon.php'] as $discoverableConfigName) {
```
yguedidi, line 162: `*.dist` files are configured to be NEON files, so it doesn't work with `phpstan.php.dist` and `phpstan.neon.php.dist` (see failing tests).
It will require a bigger refactor to support them, so I suggest to not autodiscover them, would you be OK?

ondrejmirtes:
Hi, I disagree with this. There's no advantage to a PHP configuration file over NEON. Sure, in the future, I'd like to offer a better configuration format, but it will look differently than the current `.php` file.

Because it's undocumented and because it's not "nicer" than what NEON offers, I'm not going to pursue this direction. Thanks for understanding.

Editing NEON in PhpStorm works totally fine, I do it without any plugins (since the format is very simple).

I added baseline generation in `.php` format only for performance reasons. I don't want to encourage writing your configuration in `.php` because there might be some quirks and because it's not documented and not officially supported.

yguedidi:
OK, no problem, thanks for taking the time to explain

PR #2291, opened 2023-03-14
Allow baseline generation to detect a PHP format baseline file

yguedidi:
If there is alreadu a PHP baseline file created using the default naming, this will shorten the command line allowing to use just `--generate-baseline` instead of `--generate-baseline phpstan-baseline.php` to get a PHP baseline.

Question: I don't know how this could be tested the best way, I'd need some help here

### DIFF
--- a/src/Command/AnalyseCommand.php
+++ b/src/Command/AnalyseCommand.php
@@ -122,7 +122,7 @@ class AnalyseCommand extends Command
 		if ($generateBaselineFile === false) {
 			$generateBaselineFile = null;
 		} elseif ($generateBaselineFile === null) {
-			$generateBaselineFile = 'phpstan-baseline.neon';
+			$generateBaselineFile = is_file('phpstan-baseline.php') ? 'phpstan-baseline.php' : 'phpstan-baseline.neon';
 		}
 
 		$allowEmptyBaseline = (bool) $input->getOption('allow-empty-baseline');


src/Command/AnalyseCommand.php
```diff
@@ -122,7 +122,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		if ($generateBaselineFile === false) {
 			$generateBaselineFile = null;
 		} elseif ($generateBaselineFile === null) {
-			$generateBaselineFile = 'phpstan-baseline.neon';
+			$generateBaselineFile = file_exists('phpstan-baseline.php') ? 'phpstan-baseline.php' : 'phpstan-baseline.neon';
```
staabm: micro optimazation, which requires less syscalls
```suggestion
			$generateBaselineFile = is_file('phpstan-baseline.php') ? 'phpstan-baseline.php' : 'phpstan-baseline.neon';
```

yguedidi: thank you, didn't know, updated

ondrejmirtes:
I disagree. `--generate-baseline` generates `phpstan-baseline.neon`, this would create a weird inconsistency.

If you don't like to type that many letters every time, you can use Composer scripts, or Makefile. Thanks for understanding.

yguedidi:
No problem, thanks!

PR #2292, opened 2023-03-18
Don't report unknown class in inline-@var after class_exists()

staabm:
closes https://github.com/phpstan/phpstan/issues/9055


ondrejmirtes:
Thank you.

PR #2293, opened 2023-03-19
DX: Added info about use of `setUp()` and `tearDown()` in `TypeInferenceTestCase`

staabm:
technically these methods are not deprecated, but I think its useful to have a hint to the developer of phpstan-extensions that using `setUp()` and `tearDown()` within a TypeInferenceTestCase because the test-cases/file-asserts are analyzed statically in the background


ondrejmirtes:
Errr, what? Can you show a piece of code where it does not work?

PR #2294, opened 2023-03-20
Asymmetric @property types

jtojnar:
Fixes: https://github.com/phpstan/phpstan/issues/9062

### DIFF
--- a/src/Reflection/Annotations/AnnotationPropertyReflection.php
+++ b/src/Reflection/Annotations/AnnotationPropertyReflection.php
@@ -15,6 +15,8 @@ class AnnotationPropertyReflection implements PropertyReflection
 		private Type $type,
 		private bool $readable = true,
 		private bool $writable = true,
+		private ?Type $readableType = null,
+		private ?Type $writableType = null,
 	)
 	{
 	}
@@ -41,12 +43,12 @@ class AnnotationPropertyReflection implements PropertyReflection
 
 	public function getReadableType(): Type
 	{
-		return $this->type;
+		return $this->readableType ?? $this->type;
 	}
 
 	public function getWritableType(): Type
 	{
-		return $this->type;
+		return $this->writableType ?? $this->type;
 	}
 
 	public function canChangeTypeAfterAssignment(): bool
--- a/src/Reflection/Annotations/AnnotationPropertyReflection.php
+++ b/src/Reflection/Annotations/AnnotationPropertyReflection.php
@@ -6,15 +6,13 @@ use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\PropertyReflection;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Type;
+use PHPStan\Type\VoidType;
 
 class AnnotationPropertyReflection implements PropertyReflection
 {
 
 	public function __construct(
 		private ClassReflection $declaringClass,
-		private Type $type,
-		private bool $readable = true,
-		private bool $writable = true,
 		private ?Type $readableType = null,
 		private ?Type $writableType = null,
 	)
@@ -43,12 +41,12 @@ class AnnotationPropertyReflection implements PropertyReflection
 
 	public function getReadableType(): Type
 	{
-		return $this->readableType ?? $this->type;
+		return $this->readableType ?? new VoidType();
 	}
 
 	public function getWritableType(): Type
 	{
-		return $this->writableType ?? $this->type;
+		return $this->writableType ?? new VoidType();
 	}
 
 	public function canChangeTypeAfterAssignment(): bool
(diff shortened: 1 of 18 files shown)


src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
```diff
@@ -37,16 +37,25 @@ private function findClassReflectionWithProperty(
...
+
 			return new AnnotationPropertyReflection(
 				$declaringClass,
-				TemplateTypeHelper::resolveTemplateTypes(
-					$propertyTags[$propertyName]->getType(),
-					$classReflection->getActiveTemplateTypeMap(),
-				),
-				$propertyTags[$propertyName]->isReadable(),
-				$propertyTags[$propertyName]->isWritable(),
+				$readableType ?? $writableType,
```
jtojnar: The most problematic part is that we need to provide `type` if we do not want to break  `AnnotationPropertyReflection` API.

src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
```diff
@@ -37,16 +37,25 @@ private function findClassReflectionWithProperty(
...
+				isset($propertyReadTags[$propertyName]) ? $propertyReadTags[$propertyName]->isReadable() : false,
+				isset($propertyWriteTags[$propertyName]) ? $propertyWriteTags[$propertyName]->isWritable() : false,
```
jtojnar: These could be made redundant if we could make `AnnotationPropertyReflection::get{Readable,Writable}Type()` return `?Type` instead of `Type`.

src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
```diff
@@ -37,16 +37,25 @@ private function findClassReflectionWithProperty(
...
+				$propertyReadTags[$propertyName]->getType(),
+				$classReflection->getActiveTemplateTypeMap(),
+			);
+			$writableType = !isset($propertyWriteTags[$propertyName]) ? null : TemplateTypeHelper::resolveTemplateTypes(
+				$propertyWriteTags[$propertyName]->getType(),
+				$classReflection->getActiveTemplateTypeMap(),
+			);
+
 			return new AnnotationPropertyReflection(
 				$declaringClass,
```
jtojnar, line 53: The limitation that any `@property{,-read,write}` will shadow all magic properties from parent. But that is probably not needed for most use cases.

src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
jtojnar: Actually, `AnnotationPropertyReflection` is not part of the API so I was able to remove the redundant properties.

tests/PHPStan/Type/FileTypeMapperTest.php
```diff
@@ -32,10 +41,22 @@ public function testGetResolvedPhpDoc(): void
  */');
 		$this->assertCount(0, $resolvedA->getVarTags());
 		$this->assertCount(0, $resolvedA->getParamTags());
-		$this->assertCount(2, $resolvedA->getPropertyTags());
+		$this->assertCount(7, $resolvedA->getPropertyTags());
```
jtojnar, line 44: Should I just remove the deprecated methods from tests?

jtojnar: Oh, right. I can just add it to the baseline.

src/Reflection/Annotations/AnnotationPropertyReflection.php
```diff
@@ -41,12 +41,12 @@ public function isPublic(): bool
...
 	public function getReadableType(): Type
 	{
-		return $this->type;
+		return $this->readableType ?? new VoidType();
 	}
 
 	public function getWritableType(): Type
 	{
-		return $this->type;
+		return $this->writableType ?? new VoidType();
```
jtojnar, line 49: Looks like nextras/orm test tries to write to `@property-read` https://github.com/nextras/orm/blob/58764efa7035bd163988eaa367115fece1afbf7e/tests/inc/model/tag/Tag.php#L13. Opened https://github.com/nextras/orm-phpstan/pull/28.

jtojnar:
The rest of the integration failure tests look unrelated.

hrach:
Isn't this fixing phpstan/phpstan#4845 as well? (Probably not :/)

jtojnar:
@hrach With this PR, it would be expressible as:

```php
/**
 * @property DateTimeImmutable|null $createdAt
 * @property-write DateTimeImmutable|'now'|null $createdAt
 */
class Entity {}

$e = new Entity();
$e->createdAt = 'now';
$e->createdAt->format();
```

But you would still need `getTypeAfterAssignment()` to know 'now' will be converted to non-`null` value.

ondrejmirtes:
Hello, thank you for your work! I took the liberty to take your tests and implement it in an easier way: https://github.com/phpstan/phpstan-src/pull/2327

We just needed to add different types to PropertyTag, there was no need to deprecate `getPropertyTags()` and change ResolvedPhpDocBlock etc.

PR #2295, opened 2023-03-20
[mysqli] Add stubs for `mysqli` and `mysqli_result` classes

phansys:


ondrejmirtes:
What are the types without the stubs added? Also, isn't this platform-dependent? Isn't the precision going to be annoying for someone who knows they're using a specific DB engine?

phansys:
>What are the types without the stubs added?

I'm not sure about this question. Are you asking about other missing properties that are defined for these classes?

>Also, isn't this platform-dependent?

AFAIK, these types are the same for any platform. At least, the docs are not exposing a distinction regarding this.

>Isn't the precision going to be annoying for someone who knows they're using a specific DB engine?

IIUC, these types are specific for the MySQLi extension, regardless the MySQL engine version.

ondrejmirtes:
What error are you trying to fix? How the tests fail if you remove changes in stubs?

phansys:
[This is the error](https://phpstan.org/r/78c9f18e-4670-4230-abdd-eba9c79feb2f) I'm trying to fix.

ondrejmirtes:
Please add a regression test for that too.

phansys:
>Please add a regression test for that too.

Done. Please, let me know if there is something else to address.

ondrejmirtes (changes requested):
1) The namespaces should be more unique. Something like `namespace MySqlAffectedRowsType`.
2) Your playground example is about ReturnTypeRule, which means the test should be added in ReturnTypeRuleTest (for methods).

phansys:
I've changed the namespaces and added the test to [`returnTypes.php`](https://github.com/phpstan/phpstan-src/pull/2295/files#diff-d77186c96b8806c22f5ee0fd82eb084bf1b8e8bdb4ae6600d734d844cd483ef1).
I hope these changes are going in the right direction, but please let me know if there is something missing.

Thank you in advance.

ondrejmirtes:
Thank you!

PR #2297, opened 2023-03-21
feat: improve `Map::get()` @throws inference

simPod:
Opening based on https://github.com/phpstan/phpstan/issues/9066, hope it's correct place where to put the extension

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -1414,6 +1414,11 @@ services:
 			- phpstan.broker.dynamicMethodReturnTypeExtension
 
 	-
+		class: PHPStan\Type\Php\DsMapDynamicMethodThrowTypeExtension
+		tags:
+			- phpstan.dynamicMethodThrowTypeExtension
+
+	-
 		class: PHPStan\Type\Php\DioStatDynamicFunctionReturnTypeExtension
 		tags:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
Please provide a test that fails without the extension (and also test the code that is supposed to throw the exception).

tests/PHPStan/Analyser/data/throw-points/ext-ds.php
```diff
@@ -0,0 +1,27 @@
...
+	}
+};
+
+function () {
+	$map = new \Ds\Map();
+	try {
+		$map->get('1', null);
+		$foo = 1;
+	} finally {
+		assertVariableCertainty(TrinaryLogic::createYes(), $foo);
```
simPod: Any idea why this failed? I've debugged the extension and it returns VoidType in this case so phpstan should know nothing is thrown 🤔

ondrejmirtes: No idea. You can try to show the effect of the extension in a rule test (like a dead catch...).

tests/PHPStan/Rules/Exceptions/data/ext-ds.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace PHPStan\Rules\Exceptions\data;
```
ondrejmirtes: Bad namespace and filename, please use `namespace Bug9066` + `bug-9066.php`.

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2298, opened 2023-03-21
Clarify --xdebug warning message

AJenbo:
I have seen several developers getting confused by this message thinking that --xdebug will result in better performance.

Since it can be read as:
> "--xdebug" is not used. This may slow down performance

(missing the meaning of the period and surrounding text)

Having xdebug and phpstan installed is common for developers :)

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -96,7 +96,7 @@ class CommandHelper
 		if ($allowXdebug && !XdebugHandler::isXdebugActive()) {
 			$errorOutput->getStyle()->note('You are running with "--xdebug" enabled, but the Xdebug PHP extension is not active. The process will not halt at breakpoints.');
 		} elseif (!$allowXdebug && XdebugHandler::isXdebugActive()) {
-			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. This may slow down performance and the process will not halt at breakpoints.');
+			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. Using "--xdebug" may slow down performance, but will allow you to debug PHPStan using breakpoints.');
 		}
 
 		if (!$allowXdebug) {


src/Command/CommandHelper.php
```diff
@@ -96,7 +96,7 @@ public static function begin(
 		if ($allowXdebug && !XdebugHandler::isXdebugActive()) {
 			$errorOutput->getStyle()->note('You are running with "--xdebug" enabled, but the Xdebug PHP extension is not active. The process will not halt at breakpoints.');
 		} elseif (!$allowXdebug && XdebugHandler::isXdebugActive()) {
-			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. This may slow down performance and the process will not halt at breakpoints.');
+			$errorOutput->getStyle()->note('The Xdebug PHP extension is active, but "--xdebug" is not used. Using "--xdebug" may slow down performance, but will allow you to debug PHPStan using breakpoints.');
```
staabm, line 99: the point of this message is, that php processes may slow down because xdebug beeing loaded.

phpstan doesn't get slowser because of `--xdebug`

AJenbo: Doesn't xdebug handler prevent that by disabling the extension?

My understanding of --xdebug is that it prevents xdebug handler from doing so.

The current phrasing makes it sound like --xdebug will improve performance which I don't think is the case either.

staabm: When xdebug is loaded things get a lot slower - e.g. when running `make tests`.

If you don't run with `--xdebug` you get a slow process without any benefits - hence the warning.

AJenbo: The whole point of `XdebugHandler` is to restart the process without Xdebug exactly to avoid this slow down, even if it is installed. By testing for `XdebugHandler::isXdebugActive()` before `$xdebug->check();` this message is being shown when it's not relevant.

Here are some numbers from a fairly large project.
php phpstan: 117 sec
php+xdebug phpstan: 110 sec
php+xdebug phpstan --xdebug: 264 sec

Maybe you had `PHPSTAN_ALLOW_XDEBUG=1` set in your enviroment when testing this so that Xdebug was left active? This would cause Xdebug to still be loaded and result in the behavior you describe (https://github.com/composer/xdebug-handler#basic-usage).

Edit: see #2299 for a fix for that issue.

AJenbo: I see the confusing nature of the message has been pointed out by others as well: https://github.com/phpstan/phpstan-src/pull/1878#issuecomment-1336982466, [issue/WI-69996](https://youtrack.jetbrains.com/issue/WI-69996/Quality-tools-phpstan-psalm-warning-if-interpreter-with-xdebug-extension-is-used?_gl=1*10swwrx*_ga*MTk1MjY1NzQzNi4xNjc5NDU0Nzg2*_ga_9J976DJZ68*MTY3OTQ1NDc4NS4xLjAuMTY3OTQ1NDc4OS42MC4wLjA.&_ga=2.184729193.17751043.1679454786-1952657436.1679454786)

How about this phrasing instead?

> The Xdebug PHP extension is active, but "--xdebug" is not used. Consider disabling Xdebug when not debugging PHPStan, to avoid performance issues.

I took out `and the process will not halt at breakpoints` to keep the message relatively short as users are unlikely to have set any breakpoints if they where unaware that Xdebug is loaded.

If neither of these versions seem suitable, could you please help me rephrase it? It sounds like the message might be related to how `make tests` works internally, something I'm not familiar with. 
The warning is causing confusion for end-users in its current form. I have noticed that some people are setting up their CI to run slower due to misunderstanding the message and thinking that using --xdebug will actually improve performance when Xdebug is active. Your help with resolving this issue would be greatly appreciated. Thank you

ondrejmirtes:
Please rebase since the code moved.

ondrejmirtes:
I did one more thing regarding this: https://github.com/phpstan/phpstan-src/commit/1d4ede19f6f08d6a07216cfe8536dc7dec2f4001

ondrejmirtes:
I closed this PR because I don't think we need to improve this anymore.

AJenbo:
Agreed, the new message shown since https://github.com/phpstan/phpstan-src/commit/1d4ede19f6f08d6a07216cfe8536dc7dec2f4001 makes it much more clear which options is the right one for ones use case.

PR #2299, opened 2023-03-22
Attempt unloading Xdebug before showing warnings

AJenbo:
The Xdebug warning was being shown even when XdebugHandler would avoid the slow down by unloading the extension. This avoids showing the warning for users who have not disabled XdebugHandler via env or other means.

This also has the benefit of making the second message less misleading as it is now only sown when users have a setup that is currently causing a needless slowdown of performance.


ondrejmirtes:
Makes sense, thank you!

PR #2301, opened 2023-03-22
Support of unofficial specifiers in `sprintf`/`printf`

kukulich:
Fixes https://github.com/phpstan/phpstan/issues/9072


ondrejmirtes:
Thank you very much!

PR #2302, opened 2023-03-24
Fix BackedEnum::tryFrom not being nullable

schlndh:
I encountered this case: https://phpstan.org/r/c1013cf7-d2ea-4d7a-91a9-9839456e5740 where the result of `tryFrom` is incorrectly being reported as not nullable when converting a superset enum to a subset enum. Here is my attempt to fix it.

In addition, the original extension ignores the possibility of `strict_types` not being enabled. I kept this behavior, as I'm not sure what the policy is with regards to `strict_types` (i.e. it seems that `isDeclareStrictTypes` is not called from any other type extension). But I at least added a few tests for `strict_types=1`.


ondrejmirtes:
Hi, I'm sorry, but I don't understand your change in BackedEnumFromMethodDynamicReturnTypeExtension. But I fixed it myself while keeping you as the author (plus I used your tests): https://github.com/phpstan/phpstan-src/commit/bb31aabc9d86bfe2d048669dbdf1bdca05163cab?w=1

PR #2303, opened 2023-03-26
Specify `never` for `array_combine` with different number of elements

herndlm:
Was found by @gnutix in https://github.com/phpstan/phpstan/issues/9011#issuecomment-1482916959
CC @staabm

### DIFF
--- a/src/Type/Php/ArrayCombineFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayCombineFunctionReturnTypeExtension.php
@@ -56,7 +56,7 @@ class ArrayCombineFunctionReturnTypeExtension implements DynamicFunctionReturnTy
 
 			if (count($keyTypes) !== count($valueTypes)) {
 				if ($this->phpVersion->throwsTypeErrorForInternalFunctions()) {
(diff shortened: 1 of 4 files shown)


staabm:
Thanks for working on it. 

I think we need another test which shows a actual useful rule error.

The never type alone might not create a useful error

herndlm:
not sure if this really needs another rule test? It feels not much related to me, but what were you thinking of, a rule that reports that this always throws an exception or smth like that? is this even existing already? Everything else I can think of is maybe not really related to `array_combine`, not sure 🤔

staabm:
my thinking is that a in-detail change like this is not really useful, when we don't know how it fits in the bigger picture (so which errors are generated because of this change and whether these are useful for a particular use-case)

herndlm:
I think this one's a tiny change that fixes a bug and is good to go, or does anybody have any better ideas about more tests? @ondrejmirtes
It's essentially also doing the same return-type-wise as we did in some other extensions and I just did recently via filter_input*

ondrejmirtes:
Thank you.

PR #1647, opened 2022-08-24
Upgrade vaimo/composer-patches

herndlm:
If this works it would fix more PHP 8.1 deprecation warnings as well

PR #2304, opened 2023-03-26
Upgrade vaimo/composer-patches

herndlm:
was tried already in https://github.com/phpstan/phpstan-src/pull/1647 and https://github.com/phpstan/phpstan-src/pull/2129 but I forgot what the problem was..
I hope CI tells me and this time I'll properly note it down

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -45,6 +45,10 @@ jobs:
       - name: "Validate Composer"
         run: "composer validate"
 
+      - name: "Downgrade dependencies"
+        if: matrix.php-version == '7.2'
+        run: "composer require --dev vaimo/composer-patches:^4.22 --no-install --update-with-dependencies --ignore-platform-reqs"
+
       - name: "Install dependencies"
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
There's also a different library (check what Rector is using), maybe we could switch.

herndlm:
ah ok, I guess the problem is that ` vaimo/composer-patches` dropped the dependency of `drupol/phposinfo` which was replaced with `loophp/phposinfo` and the latest version there is not compatible with PHP 7.2.
using an older version of `loophp/phposinfo` that is (e.g. 1.6.5) still uses the drupol namespace which breaks ` vaimo/composer-patches`.

so effectively ` vaimo/composer-patches` needs `loophp/phposinfo` in `^1.7` which is not compatible with PHP 7.2.

herndlm:
hmm close, it would be possible to make this work I guess, but it looks like it involves more github actions / composer install adaptions / hacks. I'll check out the other lib too.

ondrejmirtes:
`cweagans/composer-patches` was merged

PR #2307, opened 2023-03-27
Switch to `cweagans/composer-patches`

herndlm:
Alternative to https://github.com/phpstan/phpstan-src/pull/2304

upsides:
- supports PHP 5.3+ in v1 which receives still bug and security fixes

downsides:
- v2 (going to be released soon) will need PHP 8+
- does not support version-conditional patches, e.g. the paratest patch for PHP 7.2 will always fail and be skipped

docs are at https://github.com/cweagans/composer-patches/blob/1.x/README.md


herndlm:
I still don't like the patch failures because of non-matching PHP version, but apart from that it is as good as it gets I suppose.

ondrejmirtes:
If there's something in `patches` that isn't supported (like `@version ^4.0`?), please remove those lines.

herndlm:
done, just to make sure that I communicate this properly. Because of the missing PHP version info support in patches, it results in the following with every composer install (except on PHP 7.2 of course :)):
```
Gathering patches for dependencies. This might take a minute.
  - Installing brianium/paratest (v6.6.2): Extracting archive
  - Applying patches for brianium/paratest
    patches/paratest.patch (0)
   Could not apply patch! Skipping. The error was: Cannot apply patch patches/paratest.patch
```
apart from that all seems to be fine though.

ondrejmirtes:
Thank you!

PR #2309, opened 2023-03-29
Fix incorrect type for NumberFormatter::parseCurrency

robertvansteen:
Just like `parse`, `format` and `formatCurrency` the `parseCurrency` method can also return `false` in the case of failure.

https://www.php.net/manual/en/numberformatter.parsecurrency.php


ondrejmirtes:
Thank you!

PR #2312, opened 2023-03-31
Fix `Imagick::profileImage()` type

zonuexe:
[`Imagick::profileImage()`](https://www.php.net/manual/en/imagick.profileimage.php) says <q>If the profile is NULL, it is removed from the image otherwise added.</q>

The original stub is below:
https://github.com/Imagick/imagick/blob/28f27044e435a2b203e32675e942eb8de620ee58/Imagick.stub.php#L1050

PHP manual seems to be wrong as well: https://github.com/php/doc-en/pull/2396


ondrejmirtes:
Thank you!

PR #2311, opened 2023-03-31
do not check template type variance in constructor

jiripudil:
Hi, I'm gonna pick the less controversial changes from #2075 and submit them in individual PRs. Starting with this one :)

PR #2313, opened 2023-04-01
fix position variance of static method parameters

jiripudil:
As a follow-up on #2311, I think the `$isStatic` condition was originally meant as an exception for constructors, and was widened to static methods by mistake, because it produces false negatives (https://phpstan.org/r/1bb14b3f-5dc5-4417-9e27-8c0e4c886987).

Now that variance in constructors is ignored entirely, we can just remove the conditional and treat parameters in static methods as contravariant.

I’m not sure about the BC impact of this change though. It’s technically a bug fix, but it might disrupt some existing builds. I’m curious to see if any integration tests fail.

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -33,3 +33,4 @@ parameters:
 		instanceofType: true
 		paramOutVariance: true
 		allInvalidPhpDocs: true
+		strictStaticMethodTemplateTypeVariance: true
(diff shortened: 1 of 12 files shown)


ondrejmirtes:
There aren't many heavy-generics projects in the integration test suite so that's why we might not see any failures here. But I feel like this should be behind bleedingEdge config because it detects NEW situations in the code.

ondrejmirtes:
Thank you!

PR #2317, opened 2023-04-03
Fix NeverType accepting non-never types

JanTvrdik:
Closes phpstan/phpstan#9133


ondrejmirtes:
This has already been implemented with NonAcceptingNeverType.

PR #2318, opened 2023-04-03
Improve `filter_*` array handling, support `FILTER_REQUIRE_ARRAY`

herndlm:
Refs: https://github.com/phpstan/phpstan/discussions/9134

fyi @GMQS

See https://www.php.net/manual/en/filter.filters.flags.php and https://3v4l.org/p8jA9 and 

Support for more specialised arrays like constant arrays or lists as inputs could be improved still, but this is already complicated enough and I'm not sure if it's worth it. Could be added later still I suppose.


herndlm:
I assume you're busy with other things, but since it's in the same scope - support for `filter_input()` is still open via https://github.com/phpstan/phpstan-src/pull/2010 and should be ready :)

ondrejmirtes:
Thank you!

PR #2322, opened 2023-04-04
Error when on closed-resources is passed to io functions

staabm:


ondrejmirtes:
Sorry, I don't think I'm interested in this. Resources in PHP are on their way out, and adding a new Type impl to complicate the typesystem isn't something I want to do.

staabm:
even if php-src got rid of resources and we are back to regular objects, we still have to cope with the "closed-state" of e.g. curl objects etc.

do you think it would make sense to somehow represent this closed-state in the type system, so we can pinpoint errors like

```php
<?php
$ch = curl_init();
curl_setopt($ch, CURLOPT_URL, "http://www.example.com/");
curl_setopt($ch, CURLOPT_HEADER, 0);
curl_exec($ch);
curl_close($ch);

// error because of using the curl resource/object after closing it
curl_setopt($ch, CURLOPT_HEADER, 0);
curl_exec($ch);
```

which is pretty similar to

```php
function doFoo() {
	$fp = tmpfile();
	if ($fp === false) {
		die('could not open temp file');
	}
	fread($fp, 1);
	feof($fp);
	fwrite($fp, 'should work');
	fclose($fp);

	// working on a already closed resource should error
	fread($fp, 1);
	feof($fp);
	fwrite($fp, 'err - already closed');
	fclose($fp);
}
```

ondrejmirtes:
What happens in runtime if you do such a mistake?

staabm:
we get warnings

grafik

grafik

ondrejmirtes:
I worried this would be less detectable. This is pretty easy to figure out and fix in the source code, I don't think we need to support such legacy use-cases here.

In case you want to pursue this, we'd need some kind of `ClosedType` AccessoryType to intersect with existing ResourceType and object types.

ondrejmirtes:
Also, I'd like to remind you you already have 15 open pull requests here, please finish them or close them first ;)

staabm:
most PRs in the pipeline atm are waiting for your feedback I fear.

I will stop submitting new ones

ondrejmirtes:
I'm sorry, I sounded too harsh. Of course submitting a 19th PR is totally fine if it's going to be mergable and I'm gonna like it. The number of open PRs doesn't matter, the throughput (bandwidth) does 😊

staabm:
no worries. I can totally see that a lot of people try to contribute and you need to prioritize your time.

PR #2323, opened 2023-04-04
Fix Memcache/MemcachePool get method signature

dravnic:
https://www.php.net/manual/en/memcache.get.php

### DIFF
--- a/issue-bot/playground.neon
+++ b/issue-bot/playground.neon
@@ -1,3 +1,4 @@
 rules:
 	- PHPStan\Rules\Playground\FunctionNeverRule
 	- PHPStan\Rules\Playground\MethodNeverRule
+	- PHPStan\Rules\Playground\NoPhpCodeRule
--- a/src/Rules/Playground/NoPhpCodeRule.php
+++ b/src/Rules/Playground/NoPhpCodeRule.php
@@ -1,0 +1,41 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Playground;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\FileNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function count;
+
+/**
+ * @implements Rule<FileNode>
+ */
+class NoPhpCodeRule implements Rule
+{
+
+	public function getNodeType(): string
+	{
+		return FileNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (count($node->getNodes()) !== 1) {
+			return [];
+		}
+
+		$html = $node->getNodes()[0];
+		if (!$html instanceof Node\Stmt\InlineHTML) {
+			return [];
+		}
+
+		return [
+			RuleErrorBuilder::message('The example does not contain any PHP code. Did you forget the opening <?php tag?')
+				->identifier('phpstanPlayground.noPhp')
+				->build(),
+		];
+	}
+
+}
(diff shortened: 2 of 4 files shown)


resources/functionMap.php
```diff
@@ -6414,7 +6414,7 @@
 'Memcache::decrement' => ['int', 'key'=>'string', 'value='=>'int'],
 'Memcache::delete' => ['bool', 'key'=>'string', 'timeout='=>'int'],
 'Memcache::flush' => ['bool'],
-'Memcache::get' => ['string|array|false', 'key'=>'string', '&flags='=>'array', '&keys='=>'array'],
+'Memcache::get' => ['string|array|false', 'key'=>'string|array', '&flags='=>'int|array'],
```
staabm: Could we get more precise, e.g. string[] ?

dravnic: @staabm Good idea. Just added commit with string[] for key and int[] for flags

ondrejmirtes (changes requested):
Can you please add these as separate variants? Because they are separate variants at https://www.php.net/manual/en/memcache.get.php

What I mean by that - look at the top of the file and see the function `abs()`:

```
'abs' => ['0|positive-int', 'number'=>'int'],
'abs\'1' => ['float', 'number'=>'float'],
'abs\'2' => ['float|0|positive-int', 'number'=>'string'],
```

You can also test that in NodeScopeResolverTest - passing string should return string, passing array should return array.

dravnic:
@ondrejmirtes Sure, I didn't know it was possible to define variants. makes more sense.

dravnic:
Using variants in the latest commit. 

Also fixed return value. Memcache::get deserializes so return value should be mixed i.e. mixed[]|false

ondrejmirtes:
You can also test that in NodeScopeResolverTest - passing string should return string, passing array should return array.

dravnic:
It seems that PHP documentation isn't correct regarding `Memcache::get` return type. It doesn't return `string|array`, but `mixed|mixed[]` type depending on value. Below is a simple script that checks this. 

How to test this in NodeScopeResolverTest? Is there a stub/mock of Memcache class that I could use? I didn't find any.

```php
<?php
declare(strict_types=1);

$memcache = new Memcache();
$memcache->connect('localhost');

$memcache->set('key-number', 123);
$memcache->set('key-string', "abc");
$memcache->set('key-array', array(1,2,3));
$memcache->set('key-object', (object) array(1,2,3));

echo "\n---- Memcache::get (single) ----\n";
echo "key-number type -> " . gettype($memcache->get('key-number')) . "\n";
echo "key-string type -> " . gettype($memcache->get('key-string')) . "\n";
echo "key-array type -> " . gettype($memcache->get('key-array')) . "\n";
echo "key-object type -> " . gettype($memcache->get('key-object')) . "\n";

echo "\n---- Memcache::get (multi)  ----\n";
$multi = $memcache->get(array('key-number', 'key-string', 'key-array', 'key-object'));
foreach ($multi as $key => $value) {
    echo "$key type -> " . gettype($value) . "\n";
}

```

and here is the output of running it (php 8.1):

```
---- Memcache::get (single) ----
key-number type -> integer
key-string type -> string
key-array type -> array
key-object type -> object

---- Memcache::get (multi)  ----
key-number type -> integer
key-string type -> string
key-array type -> array
key-object type -> object

```

ondrejmirtes:
Here's the documentation about testing: https://phpstan.org/developing-extensions/testing#type-inference

You can also check out all the files linked in NodeScopeResolverTest to get the idea how it works.

dravnic:
@ondrejmirtes just added tests to the NodeScopeResolverTest. hope it's ok now.

ondrejmirtes:
@dravnic 1) Please base it on top of 1.10.x.
2) Rebase it so that it contains only your own commits and no merge commits.

PR #2324, opened 2023-04-04
Add types for `IntlPartsIterator`

zonuexe:
[`IntlPartsIterator`](https://www.php.net/manual/en/class.intlpartsiterator.php) just seems to inherit from [IntlIterator::current(): mixed](https://www.php.net/manual/en/intliterator.current.php), but it actually only returns non-empty-string. https://3v4l.org/CS5Tp

HHVM has types with generics as below:
https://github.com/facebook/hhvm/blob/77e37f7129e72d3434eea3ca9c8aad33e550350b/hphp/hack/hhi/stdlib/builtins_intl.hhi#L1550-L1552


resources/functionMap.php
```diff
@@ -5357,7 +5357,7 @@
 'IntlBreakIterator::getErrorCode' => ['int'],
 'IntlBreakIterator::getErrorMessage' => ['string'],
 'IntlBreakIterator::getLocale' => ['string', 'locale_type'=>'string'],
-'IntlBreakIterator::getPartsIterator' => ['IntlPartsIterator', 'key_type='=>'int'],
+'IntlBreakIterator::getPartsIterator' => ['IntlPartsIterator', 'key_type='=>'IntlPartsIterator::KEY_*'],
```
zonuexe, line 5360: https://www.php.net/manual/en/intlbreakiterator.getpartsiterator.php

ondrejmirtes:
Thank you.

PR #2327, opened 2023-04-05
Asymmetric @property types - thinner version

ondrejmirtes:
Closes https://github.com/phpstan/phpstan/issues/9062

### DIFF
--- a/src/PhpDoc/Tag/PropertyTag.php
+++ b/src/PhpDoc/Tag/PropertyTag.php
@@ -10,25 +10,44 @@ class PropertyTag
 
 	public function __construct(
 		private Type $type,
-		private bool $readable,
-		private bool $writable,
+		private ?Type $readableType,
+		private ?Type $writableType,
 	)
 	{
 	}
 
+	/**
+	 * @deprecated Use getReadableType() / getWritableType()
+	 */
 	public function getType(): Type
 	{
 		return $this->type;
 	}
 
-	public function isReadable(): bool
+	public function getReadableType(): ?Type
 	{
-		return $this->readable;
+		return $this->readableType;
 	}
 
+	public function getWritableType(): ?Type
+	{
+		return $this->writableType;
+	}
+
+	/**
+	 * @phpstan-assert-if-true !null $this->getReadableType()
+	 */
+	public function isReadable(): bool
+	{
+		return $this->readableType !== null;
+	}
+
+	/**
+	 * @phpstan-assert-if-true !null $this->getWritableType()
+	 */
 	public function isWritable(): bool
 	{
-		return $this->writable;
+		return $this->writableType !== null;
 	}
 
 }
--- a/src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
+++ b/src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
@@ -5,6 +5,7 @@ namespace PHPStan\Reflection\Annotations;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\PropertiesClassReflectionExtension;
 use PHPStan\Reflection\PropertyReflection;
+use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Generic\TemplateTypeHelper;
 
 class AnnotationsPropertiesClassReflectionExtension implements PropertiesClassReflectionExtension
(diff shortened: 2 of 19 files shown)


src/PhpDoc/Tag/PropertyTag.php
```diff
@@ -10,25 +10,44 @@ class PropertyTag
 
 	public function __construct(
 		private Type $type,
-		private bool $readable,
-		private bool $writable,
+		private ?Type $readableType,
+		private ?Type $writableType,
```
jtojnar, line 14: One downside compared to #2294 is that sharing `PropertyTag` makes it impossible to e.g. inherit `@property` definition and then override `@property-write` in the subclass. (Though I did not bother implementing it yet.)

tests/PHPStan/Analyser/data/asymmetric-properties.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+#[AllowDynamicProperties]
+class Foo
+{
+
+	public function doFoo(): void
+	{
+		assertType('int', $this->asymmetricPropertyRw);
+		assertType('int', $this->asymmetricPropertyXw);
+		assertType('int|string', $this->asymmetricPropertyRx);
```
jtojnar, line 26: This does not look right. The `@property` should be shadowed by `@property-read`.

ondrejmirtes: To me it doesn't make sense to allow both `@property` and `@property-read`. My expectation is that `@property` overwrites everything.

jtojnar: I would expect the opposite – the more specific `@property-read` and `@property-write` should overwrite the less specific `@property` (at least within the context of a single class). In fact, that was the behaviour before this PR.

ondrejmirtes: Oh, please send an updating PR that restores this behaviour, and adds some tests for it. Thank you!

jtojnar: I do not care that much either way – converting my existing `@property` tags to `@property-read` when adding `@property-write` tags is only mildly annoying. Though, if we are counter-intuitively ignoring the more specific tags, we might want to add a check that warns that there are conflicting tags.

ondrejmirtes: Because the tags are indexed by name, the check cannot really easily be written.

src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
```diff
@@ -39,10 +40,20 @@ private function findClassReflectionWithProperty(
...
+			$defaultType = $propertyTag->getReadableType() ?? $propertyTag->getWritableType();
+			if ($defaultType === null) {
+				throw new ShouldNotHappenException();
+			}
+
 			return new AnnotationPropertyReflection(
 				$declaringClass,
 				TemplateTypeHelper::resolveTemplateTypes(
-					$propertyTags[$propertyName]->getType(),
+					$propertyTag->getReadableType() ?? $defaultType,
```
jtojnar, line 52: Another point of difference from #2294, this will continue to return an incorrect readable type when the property is not readable. It would be more elegant if we used `VoidType` as the fallback type – then instead of checking if the property is readable/writeable and checking if a value can inhabit the type, we could simply check the value, since `void` is a bottom type that cannot be inhabited by any value.

tests/PHPStan/Analyser/data/asymmetric-properties.php
ondrejmirtes: But backward compatibility is important to me so please research it and send a PR.

src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
ondrejmirtes: That's NeverType, not VoidType. Yeah, I didn't understand that change. Feel free to update that scenario to NeverType, thank you.

jtojnar: Opened #2328.

tests/PHPStan/Analyser/data/asymmetric-properties.php
jtojnar: Opened #2329.

PR #2328, opened 2023-04-05
Use bottom type for read/write-only properties

jtojnar:
Follow-up to https://github.com/phpstan/phpstan-src/pull/2327

It is not meaningful to return the writable type from getReadableType() type for non-readable properties.
Let’s return a bottom type to make it clear that no value can be read from the property.
And similarly, for non-writable properties, getReadableType() will return bottom type
to emphasize that no value exists that can be written to the property.

Probably best rebase-merge so that the second commit can be easily reverted if it turns out it causes too many issues.


ondrejmirtes (changes requested):
Please uncomment the test in `tests/PHPStan/Analyser/data/classPhpDocs-phpstanPropertyPrefix.php`

jtojnar:
Done.

ondrejmirtes:
Thank you!

PR #2329, opened 2023-04-05
AnnotationsPropertiesClassReflectionExtension: More specific property prevail

jtojnar:
Restore the behaviour from before asymmetric properties were introduced, when `@property-read` and `@property-write` would shadow `@property`. This is slightly more intuitive since those tags are more specific.

Another follow-up to https://github.com/phpstan/phpstan-src/pull/2327


ondrejmirtes:
Thank you!

PR #2333, opened 2023-04-07
Improve return type precision of `filter_input` with invalid first args

herndlm:
This is essentially https://3v4l.org/vLIZK / makes the return type more precise when the first arg (type constant) is invalid.

In order to do that I had to refactor this thing again and move more into the helper. I added caching for `supportedFilterInputTypes` at the same time as the helper was also doing it for other similar props.

The idea is also to keep the return type correct in combination with https://github.com/phpstan/phpstan-src/pull/2271


ondrejmirtes:
Thank you!

PR #2336, opened 2023-04-10
Prevent false-positive in get_parent_class() on interfaces

staabm:
closes https://github.com/phpstan/phpstan/issues/4302


ondrejmirtes:
Thank you!

PR #2337, opened 2023-04-10
generics: fix position variance of readonly-by-phpdoc properties

jiripudil:
closes phpstan/phpstan#9153

### DIFF
--- a/tests/PHPStan/Rules/Generics/data/bug-9153.php
+++ b/tests/PHPStan/Rules/Generics/data/bug-9153.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace ClxProductNet\Recombee\Model;
+
+/**
+ * @template-covariant Value of string|list<string>
+ *
+ * @immutable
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/Generics/data/bug-9153.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace ClxProductNet\Recombee\Model;
+
+/**
+ * @template-covariant Value of string|list<string>
+ *
+ * @immutable
```
staabm, line 8: ~~I think this needs another test-case with a `@readonly`-phpdoc property on a non-immutable class~~

seems this is working as expected already: https://phpstan.org/r/c1a29fdf-09c7-4500-a7c8-87a64de4ef9f

ondrejmirtes:
Thank you!

staabm:
thank you @jiripudil

PR #2348, opened 2023-04-18
Add extension to add additional constructors through code

raalderink:
This provides the basis to be able to fix this issue;
https://github.com/phpstan/phpstan-symfony/issues/346

### DIFF
--- a/src/Reflection/AdditionalConstructorsExtension.php
+++ b/src/Reflection/AdditionalConstructorsExtension.php
@@ -1,0 +1,29 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Reflection;
+
+/**
+ * This is the extension interface to implement if you want to dynamically
+ * mark methods as constructor. As opposed to simply list them in the configuration file.
+ *
+ * To register it in the configuration file use the `phpstan.additionalConstructorsExtension` service tag:
+ *
+ * ```
+ * services:
+ * 	-
+ *		class: App\PHPStan\MyExtension
+ *		tags:
+ *			- phpstan.additionalConstructorsExtension
+ * ```
+ *
+ * @api
+ */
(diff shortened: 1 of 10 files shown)


src/Reflection/AdditionalConstructorsExtension.php
```diff
@@ -0,0 +1,13 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Reflection;
+
+interface AdditionalConstructorsExtension
```
ondrejmirtes, line 21: Would be nice to add a PHPDoc description similar to https://github.com/phpstan/phpstan-src/blob/979175ae6fa9ad2c3a2a5f228ac0449cded7900d/src/PhpDoc/StubFilesExtension.php#L5-L20 (also `@api` is needed)

tests/PHPStan/Rules/Properties/UninitializedPropertyRuleTest.php
```diff
@@ -15,8 +18,29 @@ class UninitializedPropertyRuleTest extends RuleTestCase
 
 	protected function getRule(): Rule
 	{
+		$containerMock = $this->createMock(Container::class);
+		$containerMock->method('getServicesByTag')
```
ondrejmirtes: Instead of mocking, please create a separate test case similar to this one, but also create a new `.neon` file next to the test, and configure the extension there.

And then reference the new `.neon` file in `public static function getAdditionalConfigFiles()` in the test case 👍

ondrejmirtes (changes requested):
And listen to the bot :) I'd release this quickly and you could continue your work in phpstan-symfony. Thanks!

ondrejmirtes:
Perfect, thank you!

PR #2349, opened 2023-04-18
Make `NonEmptyArrayType::toArray()` return `$this`

zonuexe:
close https://github.com/phpstan/phpstan/issues/9208

### DIFF
--- a/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
@@ -7,7 +7,6 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
-use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\Type;
--- a/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
@@ -5,8 +5,7 @@ namespace PHPStan\Type\Php;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\Type\Accessory\AccessoryArrayListType;
-use PHPStan\Type\Accessory\NonEmptyArrayType;
+use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\Type;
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Analyser/data/array-slice.php
```diff
@@ -12,7 +12,7 @@ class Foo
 	 */
 	public function nonEmpty(array $a): void
 	{
-		assertType('array', array_slice($a, 1));
+		assertType('non-empty-array', array_slice($a, 1));
```
zonuexe: `array_slice()` can return empty arrays. https://3v4l.org/8VGrv

src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
```diff
@@ -50,7 +51,7 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 		}
 
 		if ($valueType->isIterableAtLeastOnce()->yes()) {
-			return $valueType->toArray();
+			return TypeCombinator::union($valueType->toArray(), new NonEmptyArrayType());
```
ondrejmirtes: intersect, not union

tests/PHPStan/Analyser/data/bug-9208.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+/**
+ * @param int|non-empty-list<int> $id_or_ids
+ * @return non-empty-list<int>
+ */
+function f(int|array $id_or_ids): array
+{
+	if (is_array($id_or_ids)) {
+		\PHPStan\Testing\assertType('non-empty-list<int>', (array)$id_or_ids);
```
ondrejmirtes: The file needs to have namespace `Bug9208`.

ondrejmirtes: The file isn't referenced in NodeScopeResolverTest

ondrejmirtes: Please use `PHPStan\Testing\assertType` in a `use function` statement.

ondrejmirtes (changes requested):

src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
zonuexe: I'm trying to make a `non-empty-array&array<T>` into an `array<T>`. Isn't it a union?

tests/PHPStan/Analyser/data/bug-9208.php
zonuexe: fixed

src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
ondrejmirtes: So you mean that you want an array that’s non-empty to no longer be non-empty? I’m afraid you need to recreate it from scratch by doing new ArrayType.

tests/PHPStan/Analyser/data/array-slice.php
```diff
@@ -9,10 +9,14 @@ class Foo
...
 	/**
 	 * @param non-empty-array $a
+	 * @param non-empty-list $b
+	 * @param non-empty-array<int>|non-empty-list<string> $c
 	 */
-	public function nonEmpty(array $a): void
+	public function nonEmpty(array $a, array $b, array $c): void
 	{
 		assertType('array', array_slice($a, 1));
+		assertType('list', array_slice($b, 1));
```
staabm: seems this assertion fails atm in CI 
```suggestion
		assertType('list<mixed>', array_slice($b, 1));
```

zonuexe:
@ondrejmirtes Could you please review this PR?

herndlm:
Ah interesting, this is the same problem I found independently via https://github.com/phpstan/phpstan-src/pull/2451#issuecomment-1587862340 but with a different solution that not just only removes that one accessory

ondrejmirtes:
There's a much easier solution to make a non-empty array possibly empty again :) https://github.com/phpstan/phpstan-src/pull/2349/commits/9e84671c475f2e1fded227965852ce9a32788484

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

zonuexe:
> There's a much easier solution to make a non-empty array possibly empty again :) https://github.com/phpstan/phpstan-src/commit/9e84671c475f2e1fded227965852ce9a32788484

[I agree.](https://github.com/phpstan/phpstan-src/pull/2349#discussion_r1170464977) 😄

ondrejmirtes:
Oh wow, so you actually planted it in my brain two months ago :D Well I'm a bit embarassed now :D

PR #2350, opened 2023-04-19
Fix `get_class()` on HasMethodType

staabm:
closes https://github.com/phpstan/phpstan/issues/4890

### DIFF
--- a/src/Type/Php/GetClassDynamicReturnTypeExtension.php
+++ b/src/Type/Php/GetClassDynamicReturnTypeExtension.php
@@ -9,14 +9,17 @@ use PHPStan\Type\ClassStringType;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\Enum\EnumCaseObjectType;
 use PHPStan\Type\Generic\GenericClassStringType;
+use PHPStan\Type\IntersectionType;
 use PHPStan\Type\ObjectShapeType;
+use PHPStan\Type\ObjectType;
 use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\TypeUtils;
-use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
 use function count;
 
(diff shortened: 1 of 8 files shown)


src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -85,6 +86,8 @@ static function (Type $type, callable $traverse): Type {
 					return new GenericClassStringType($type);
 				} elseif ($type instanceof ObjectWithoutClassType) {
 					return new ClassStringType();
+				} elseif ($type instanceof HasMethodType) {
+					return new ClassStringType();
```
ondrejmirtes: I think it'd make more sense to get rid of this whole monstrosity (because it'd still fail for HasPropertyType for example - feel free to add a test for it) and also other types like ObjectShapeType.

Instead the whole `TypeTraverser::map` should be replacable with some methods called on `Type`.

* `Type::isObject()->no()` - always returns false
* `Type::isObject()->yes()` - never returns false
* `Type::isObject()->maybe()` - might return false

As for the actual type, you could do something like this:

```php
new GenericClassStringType(TypeCombinator::intersect($argType, new ObjectWithoutClassType()));
```

It will change the result in some cases but it will still be correct.

ondrejmirtes (changes requested):

src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -53,43 +48,22 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-				$objectClassNames = $type->getObjectClassNames();
-				if ($type instanceof TemplateType && $objectClassNames === []) {
-					if ($type instanceof ObjectWithoutClassType) {
-						return new GenericClassStringType($type);
-					}
+		$isObject = $argType->isObject();
+		if ($isObject->yes()) {
+			if ($argType instanceof ObjectShapeType) {
+				return new ClassStringType();
+			}
```
staabm: I am sure you won't like it, but I think it seems to be necessary.

without this case we get things like
`class-string<object{foo: Bug4890\HelloWorld, bar: int, baz?: string}>`

when intersecting `ObjectShapeType` and `ObjectWithoutClassType`.

src/Type/Php/GetClassDynamicReturnTypeExtension.php
staabm: I think your idea works pretty good. one exception is

```
 PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Analyser/data/bug-7167.php:11" ('type', 'C:\dvl\Workspace\phpstan-src-...67.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\Generic\GenericClassStringType Object (...), 11)
Expected type class-string<Bug7167\Foo>, got type class-string<Bug7167\Foo::Value> in C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Analyser/data/bug-7167.php on line 11.
--- Expected
+++ Actual
@@ @@
-'class-string<Bug7167\Foo>'
+'class-string<Bug7167\Foo::Value>'
```

which I am not yet sure how to handle

tests/PHPStan/Analyser/data/bug-4890.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Bug4890;
+
+use function PHPStan\Testing\assertType;
+
+interface Proxy {}
+
+class HelloWorld
```
staabm, line 9: see https://phpstan.org/r/b6b95d5a-a3fb-4c29-be47-4e7832f00ddc regarding how this test fails before this PR

staabm:
think I figured it out - I had to re-introduce the TypeTraverser for the EnumCaseObjectType

https://github.com/phpstan/phpstan-src/pull/2350#discussion_r1171591087

src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -64,27 +64,26 @@ static function (Type $type, callable $traverse): Type {
 					return new GenericClassStringType(new ObjectType($type->getClassName()));
 				}
 
-				$objectClassNames = $type->getObjectClassNames();
-				if ($type instanceof TemplateType && $objectClassNames === []) {
-					if ($type instanceof ObjectWithoutClassType) {
-						return new GenericClassStringType($type);
+				$isObject = $type->isObject();
+				if ($isObject->yes() || $isObject->maybe()) {
```
ondrejmirtes, line 68: For which type here it's going to be a `maybe`?

src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -64,27 +64,26 @@ static function (Type $type, callable $traverse): Type {
...
+					if ($type instanceof ObjectShapeType) {
+						return new ClassStringType();
+					}
+
+					$objectType = TypeCombinator::intersect($type, new ObjectWithoutClassType());
+					if ($objectType instanceof StaticType) {
+						$objectType = $objectType->getStaticObjectType();
```
ondrejmirtes, line 75: Why are you checking this after an intersection?

ondrejmirtes (changes requested):

PR #2351, opened 2023-04-19
Fix intval() and floatval() call with a non scalar value

dravnic:
Here is a code snippet that shows a misleading error when `intval()` is used with a non-scalar.

```php
<?php

class Foo
{
    private int $x = 0;
    
    function __construct() {
        $this->x += intval(new stdClass());
    }
}
```

```
8      Property Foo::$x (int) does not accept (float|int).  
```

It's reproducible with `checkBenevolentUnionTypes: true`

I've traced the issue.

`intval()` with non-scalar argument resolves to `ErrorType` which is inherited from `MixedType`.  

So `ErrorType::toNumber()` returns `UnionType of int and float`. When above math expression is resolved it has `int` on the left side and `float|int` on the right side which produces error when `checkBenevolentUnionTypes` is on.

The root cause is fixed in this PR. It adds non-scalar value handling by making sure that return type for intval/floatval in all cases is `int` i.e. `float`. 

PHP is throwing a warning on use of non-scalar with intval/floatval and not an error like with strval, so even though non-scalar arguments shouldn't be used we should handle them.


ondrejmirtes:
Besides this it would also be nice to change the type of intval and floatval parameters to `scalar` in resources/functionMap.php. Because with non-scalars it produces a warning: https://3v4l.org/Bdim5

dravnic:
Makes sense. Just added it.

The above code now produces a proper error

```
Parameter #1 $value of function intval expects bool|float|int|string, stdClass given.
```

ondrejmirtes:
Thank you!

PR #2354, opened 2023-04-20
More precise strval family function signatures

dravnic:
This PR makes consistent error reporting on strval family non-scalars params with PHP's reporting.

Here is output how PHP handles these values: https://3v4l.org/O2KY5

Unfortunately PHP isn't consistent there, but it would be better that we comply with that than report false positives.


ondrejmirtes:
Thank you!

PR #2358, opened 2023-04-22
Fix ConstantFloatType value dump precision

mvorisek:
When dumping a float value, we need:
- ensure `.` is always present in finite values (/wo `.` the number type is integer)
- always dump with full precision (`(string) float_expr` cast is affected by `precision` php.ini)
- never dump more digits than needed

### DIFF
--- a/src/Type/Constant/ConstantFloatType.php
+++ b/src/Type/Constant/ConstantFloatType.php
@@ -15,9 +15,9 @@ use PHPStan\Type\Traits\ConstantScalarTypeTrait;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
 use function abs;
+use function ini_get;
+use function ini_set;
 use function is_finite;
-use function rtrim;
-use function sprintf;
 use function strpos;
 use const PHP_FLOAT_EPSILON;
 
@@ -45,18 +45,27 @@ class ConstantFloatType extends FloatType implements ConstantScalarType
 		return $type instanceof self && abs($this->value - $type->value) < PHP_FLOAT_EPSILON;
 	}
 
+	private function castFloatToString(float $value): string
+	{
+		$precisionBackup = ini_get('precision');
+		ini_set('precision', '-1');
+		try {
+			$valueStr = (string) $value;
+			if (is_finite($value) && strpos($valueStr, '.') === false) {
+				$valueStr .= '.0';
+			}
+
+			return $valueStr;
+		} finally {
+			ini_set('precision', $precisionBackup);
+		}
+	}
+
 	public function describe(VerbosityLevel $level): string
 	{
 		return $level->handle(
 			static fn (): string => 'float',
-			function (): string {
-				$formatted = (string) $this->value;
-				if (is_finite($this->value) && strpos($formatted, '.') === false) {
-					$formatted .= '.0';
-				}
-
-				return $formatted;
-			},
+			fn (): string => $this->castFloatToString($this->value),
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


src/Type/Constant/ConstantFloatType.php
```diff
@@ -45,18 +45,24 @@ public function equals(Type $type): bool
 		return $type instanceof self && abs($this->value - $type->value) < PHP_FLOAT_EPSILON;
 	}
 
+	private function castFloatToString(float $value): string
+	{
+		$precisionBackup = ini_get('precision');
```
mvorisek, line 50: needs https://github.com/phpstan/phpstan/issues/3063 - how to solve it in this PR?

tests/PHPStan/Type/TypeToPhpDocNodeTest.php
```diff
@@ -290,12 +302,22 @@ public function dataToPhpDocNodeWithoutCheckingEquals(): iterable
...
+		];
+
+		yield [
+			new ConstantFloatType(8.202343767574732),
+			'8.202343767574732',
+		];
+
+		yield [
+			new ConstantFloatType(1e80),
+			'1.0E+80',
```
mvorisek, line 353: phpdoc parser needs to be fixed to parse numbers in scientific format, at least the ones with `.` where the type is unambiguous - `float`

src/Type/Constant/ConstantFloatType.php
```diff
@@ -45,18 +45,24 @@ public function equals(Type $type): bool
...
 			static fn (): string => 'float',
-			function (): string {
-				$formatted = (string) $this->value;
-				if (is_finite($this->value) && strpos($formatted, '.') === false) {
-					$formatted .= '.0';
-				}
-
-				return $formatted;
-			},
+			fn (): string => $this->castFloatToString($this->value),
```
staabm, line 68: IIRC this is a perf critical path. 

I think we need a benchmark to show whether this edge-case fixing leads to a perf regression in the common case

staabm: Scratch that. I mixed it up with https://github.com/phpstan/phpstan-src/pull/2080

mvorisek: #2080 - please see https://github.com/phpstan/phpstan/issues/9228, the comparison should be even simplifid to `===` only, as any other float value is different

staabm:
fwiw, I just stumbled over a pretty similar logic in sebastian/exporter:

https://github.com/sebastianbergmann/exporter/blob/997c7034005f045466fcf66d001821153d537c4e/src/Exporter.php#L224-L240

src/Type/Constant/ConstantFloatType.php
staabm: fixed in https://github.com/phpstan/phpstan-src/pull/2362 & https://github.com/phpstan/phpstan-src/pull/2365

mvorisek: Thank you!

tests/PHPStan/Type/TypeToPhpDocNodeTest.php
mvorisek: The problem is not scientific format not supported, but only the `+` before the exponent is not supported.

https://phpstan.org/r/81b725d6-b525-4c04-98d1-6d16b64c08ce
https://3v4l.org/SamhX - this is how php outputs exponents by default

mvorisek:
PR is done.

mvorisek:
PR is done.

ondrejmirtes:
Thank you.

PR #2362, opened 2023-04-24
Implement IniGetReturnTypeExtension

staabm:
closes https://github.com/phpstan/phpstan/issues/3063

required for https://github.com/phpstan/phpstan-src/pull/2358#discussion_r1174357092

### DIFF
--- a/src/Type/Php/IniGetReturnTypeExtension.php
+++ b/src/Type/Php/IniGetReturnTypeExtension.php
@@ -47,7 +47,7 @@ class IniGetReturnTypeExtension implements DynamicFunctionReturnTypeExtension
 		$results = [];
 		foreach ($argType->getConstantStrings() as $constantString) {
 			if (!array_key_exists($constantString->getValue(), $types)) {
-				continue;
+				return null;
 			}
 			$results[] = $types[$constantString->getValue()];
 		}
--- a/src/Type/Php/IniGetReturnTypeExtension.php
+++ b/src/Type/Php/IniGetReturnTypeExtension.php
@@ -1,0 +1,62 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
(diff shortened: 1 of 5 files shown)


src/Type/Php/IniGetReturnTypeExtension.php
```diff
@@ -0,0 +1,62 @@
...
+			'memory_limit' => new StringType(),
+			'max_execution_time' => $numericString,
+			'max_input_time' => $numericString,
+		];
+
+		$argType = $scope->getType($args[0]->value);
+		$results = [];
+		foreach ($argType->getConstantStrings() as $constantString) {
+			if (!array_key_exists($constantString->getValue(), $types)) {
+				continue;
```
ondrejmirtes: So what's the default return type? If we encounter `'unknown'|'max_execution_time'`, the return type from this extension will be wrong.

ondrejmirtes (changes requested):

src/Type/Php/IniGetReturnTypeExtension.php
staabm: good catch

ondrejmirtes:
Thank you.

ondrejmirtes:
Although I don't understand how it's gonna help https://github.com/phpstan/phpstan-src/pull/2358#discussion_r1174357092 because for that key it's still gonna be `string|false` AFAIK.

staabm:
I think `precision` can be added to the type list to make it return a `numeric-string`:

```
$ php -r 'var_dump(ini_get("precision"));'
string(2) "14"

$  php -n -r 'var_dump(ini_get("precision"));'
string(2) "14"
```

but thats something I would expect to be part of https://github.com/phpstan/phpstan-src/pull/2358

see also https://www.php.net/manual/en/ini.core.php#ini.precision

PR #2363, opened 2023-04-24
Fix Imagick::scaleImage parameter name

davidkmenta:
According to https://github.com/php/doc-en/pull/2444


ondrejmirtes:
What kind of bug are you experiencing that is fixed with this change?

davidkmenta:
When using named attributes, the phpstan reports errors according to the badly documented method signature. This code:

```php
$image = new \Imagick();
$image->readImageFile($fileStream);
$image->scaleImage(columns: 1500, rows: 1500, bestfit: true);
```

produces these errors:

>Missing parameter $cols (int) in call to method Imagick::scaleImage().    
>Unknown parameter $columns in call to method Imagick::scaleImage().

but it should not as the named attributes are correct and the code works.

ondrejmirtes:
Please verify that your change actually helps by writing a new test in CallMethodsRuleTest.

Your change might help, but it might also need a different fix in https://github.com/JetBrains/phpstorm-stubs/blob/077d371b23d38d7f6f039994bfb03732d7965fbc/imagick/imagick.php#L1962.

davidkmenta:
Fixed by #2580.

PR #2365, opened 2023-04-25
Add type for ini_get() 'default_socket_timeout' and 'precision'

staabm:
`default_socket_timeout` is pretty commonly used and `precision` seems to be something we need for https://github.com/phpstan/phpstan-src/pull/2358


ondrejmirtes:
Thank you.

PR #2369, opened 2023-05-01
`CURLOPT_HTTPHEADER` should not accept a dictionary

staabm:
closes https://github.com/phpstan/phpstan/issues/9263


ondrejmirtes:
Thank you.

PR #2372, opened 2023-05-04
Enforce magic __serialize() and __unserialize() return types

staabm:
types as defined in php-src

https://www.php.net/manual/en/language.oop5.magic.php#language.oop5.magic.serialize
https://3v4l.org/bLjLG

my motivation for this change is, as I am upgrading a php7 app for php8 compat per https://wiki.php.net/rfc/phase_out_serializable and [I experienced php segfaults](https://github.com/php/php-src/issues/11187) when I use the wrong signature or return the wrong data

returning the wrong result is a fatal error on php8

### DIFF
--- a/src/Reflection/Php/PhpMethodFromParserNodeReflection.php
+++ b/src/Reflection/Php/PhpMethodFromParserNodeReflection.php
@@ -14,6 +14,7 @@ use PHPStan\Type\ArrayType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\IntegerType;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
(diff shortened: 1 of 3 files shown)


src/Reflection/Php/PhpMethodFromParserNodeReflection.php
```diff
@@ -78,6 +79,13 @@ public function __construct(
 			$realReturnType = TypeCombinator::intersect(new ObjectWithoutClassType(), $realReturnType);
 		}
 
+		if ($name === '__unserialize') {
+			$realReturnType = new VoidType();
```
ondrejmirtes, line 83: This change would make sense if there's a type missing in the signature, like: https://3v4l.org/Wji2j

So please change the current test and remove the types from it.

But we need another rule and test - that checks the correctness of the signatures. In the example with `__unserialize(string $data): string`, both parameter and return type are wrong, but these aren't picked up by any rule.

While we're at it we should verify the corectness of signatures of all magic methods. Needs a new rule probably.

ondrejmirtes (changes requested):

src/Reflection/Php/PhpMethodFromParserNodeReflection.php
staabm: thank you. 

> So please change the current test and remove the types from it.

I have fixed the test as per your feedback 

> But we need another rule and test - that checks the correctness of the signatures. In the example with __unserialize(string $data): string, both parameter and return type are wrong, but these aren't picked up by any rule.
> While we're at it we should verify the corectness of signatures of all magic methods. Needs a new rule probably.

I will prepare a separate PR for that

ondrejmirtes:
Thank you.

PR #2373, opened 2023-05-06
Fix SplDoublyLinkedList::offsetGet return type

takaram:
Closes phpstan/phpstan#9274

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-9274.php
+++ b/tests/PHPStan/Analyser/data/bug-9274.php
@@ -16,14 +16,20 @@ class A extends SplDoublyLinkedList {}
 /** @extends SplQueue<Point> */
 class B extends SplQueue {}
 
-$dll = new A();
-$p1 = $dll[0];
+function testSplDoublyLinkedList(): void
+{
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/data/bug-9274.php
```diff
@@ -0,0 +1,29 @@
...
+{
+	public int $x;
+}
+
+/** @extends SplDoublyLinkedList<Point> */
+class A extends SplDoublyLinkedList {}
+/** @extends SplQueue<Point> */
+class B extends SplQueue {}
+
+$dll = new A();
```
ondrejmirtes: Please do not mix definitions with side effects. Put this code into another class or a function. Thanks.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/bug-9274.php
takaram: I changed the code as you suggested. Thank you!

ondrejmirtes:
Thank you.

PR #2374, opened 2023-05-08
Fix 9280 - func_num_args() presence must not allow unlimited args

mvorisek:


ondrejmirtes:
Thank you, but I've already said in the issue: https://github.com/phpstan/phpstan/issues/9280

> This has been in PHPStan for a long time and there might be users relying on opposite behaviour than you want. I'm not willing to change it.
> You can adjust your code not to use func_num_args instead.

mvorisek:
I would really like to get this merged as `func_num_args` has legit uses on non-variadic methods. The tests also shows no big BC-break will be introduced, actually there is no single BC-break. If the variadic args are retrived using `func_get_arg` or `func_get_args`, which is typical for variadic functions with undeclared params, everything is analysed as before.

mvorisek:
@ondrejmirtes the E2E tests show no single issue, I have done a short brainstorming and with this PR I do not see any, even minor downside, using `func_num_args` without retriving the actual args (via `func_get_arg`/`func_get_args`) does not make much sense. Ondřej, what usecase do you have in your head?

The problem with the current (before this PR) behaviour is a lot of problems can be supressed and this is probably also the reason why noone reported it yet.

mvorisek:
x

PR #2375, opened 2023-05-08
Support different error levels in GithubErrorFormatter

staabm:
allows to configure the message severity when formatting phpstan-errors in the GithubErrorFormatter.

Our use case:
we want to run 2 phpstan jobs in parallel. 
- one with our currently supported php version
- another job where we use our future php version (the one we are aiming for to support within the next few months)

we don't want to allow new errors in the currently supported php version. additionally we want github action to report only warnings for the future php version, as not all members of the team are working on that topic.

as soon as we have fixed all future php version problems we will raise the error level in the pipeline for the future php version

-> long story short: would be cool we could configure the message severity of errors reported by the phpstan github formatter

effectivly this will make the errors on the github.com ui yellow instead of red - example (not using phpstan):

grafik

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -225,6 +225,7 @@ parameters:
 	editorUrl: null
 	editorUrlTitle: null
 	errorFormat: null
+	githubErrorLevel: 'error'
 	pro:
 		dnsServers:
 			- '1.1.1.1'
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Hello, this is a very limited use-case that I'm not willing to support and maintain in PHPStan.

You can already do this on your side, in multiple ways:

1) Pipe the PHPStan output into your own script that will make the output "yellow".
2) Write your own error formatter that will do this. You can even decorate the original formatter and make your code DRY.

staabm:
I see. Instead I have adjusted `cs2pr` so I can downgrade errors to warnings in https://github.com/staabm/annotate-pull-request-from-checkstyle/pull/112

that way I can utilize the checkstyle error format and leave the rest to cs2pr

PR #2397, opened 2023-05-09
Disallow extra parameters for callables

enumag:
Attempting to fix https://github.com/phpstan/phpstan/issues/9248

Should this be bleeding-edge only error for now?

### DIFF
--- a/src/Type/CallableTypeHelper.php
+++ b/src/Type/CallableTypeHelper.php
@@ -83,6 +83,27 @@ class CallableTypeHelper
 			}
 		}
 
+		foreach ($ourParameters as $i => $ourParameter) {
+			$parameterDescription = $ourParameter->getName() === '' ? sprintf('#%d', $i + 1) : sprintf('#%d $%s', $i + 1, $ourParameter->getName());
+			if (!isset($theirParameters[$i])) {
+				if ($ourParameter->isOptional()) {
+					continue;
+				}
+
+				$accepts = new AcceptsResult(TrinaryLogic::createNo(), [
+					sprintf(
+						'Accepting callable has a required parameter %s but the passed callable does not have that parameter. This can trigger a fatal error for internal functions and methods which do not accept extra arguments.',
+						$parameterDescription,
+					),
+				]);
+				if ($result === null) {
+					$result = $accepts;
+				} else {
+					$result = $result->and($accepts);
+				}
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
As you can see from the CI results this isn't the right fix. This is totally OK to do for situations where it doesn't cause fatal errors. A callable might be called with a parameter but the passed callable might not be interested in that parameter.

Instead, we need to somehow remember when a callable is created from an internal function affected by this behaviour. This includes `strlen(...)` (first-class callable - ClosureType) and also `'strlen'` - ConstantStringType. And then use this information in CallableTypeHelper.

We might need to require `ParametersAcceptorWithPhpDocs` instead of `ParametersAcceptor` in CallableTypeHelper. Think of `ParametersAcceptorWithPhpDocs` as `ExtendedParametersAcceptor`, similar to ExtendedMethodReflection. And add a relevant method on ParametersAcceptorWithPhpDocs so that we can ask about this.

enumag:
Yeah, I expected something like that. That's why I marked it as a draft.

CallableTypeHelper is used in [CallableType](https://github.com/phpstan/phpstan-src/blob/7e51e34a372b2ec59c57d50a00ac2eca4b7dca31/src/Type/CallableType.php#L135) and [ClosureType](https://github.com/phpstan/phpstan-src/blob/7e51e34a372b2ec59c57d50a00ac2eca4b7dca31/src/Type/ClosureType.php#L146).

I believe I only need `ParametersAcceptorWithPhpDocs` for the `$theirs` argument which is the passed callback function. But how can I obtain `ParametersAcceptorWithPhpDocs` implementation instead of `ParametersAcceptor` in those two places?

EDIT: It seems like both `ClosureType` and `CallableType` would need to implement `ParametersAcceptorWithPhpDocs` which won't be easy considering all the places where they can be created.

ondrejmirtes:
Your EDIT is right.

> which won't be easy considering all the places where they can be created

Exactly, these are the places that need to pass in the information about the function being internal or not.

enumag:
> Exactly, these are the places that need to pass in the information about the function being internal or not.

Right now I'm more concerned about how to correctly handle the extra data which ParametersAcceptorWithPhpDocs already has since correctly handling that is a pre-requisite:
- more restrictive `getParameters()`
- `getPhpDocReturnType()`
- `getNativeReturnType()`

PR #2399, opened 2023-05-11
Report dead types even in multi-exception catch II

JanTvrdik:
This is rewrite of https://github.com/phpstan/phpstan-src/pull/2067. The tests are the same, but the implementation is rewritten.

The first commit is pure refactoring to make it more obvious what is actually happening without altering the behavior. The second commit contains the new behavior.

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -38,3 +38,4 @@ parameters:
 		genericPrototypeMessage: true
 		stricterFunctionMap: true
 		invalidPhpDocTagLine: true
+		detectDeadTypeInMultiCatch: true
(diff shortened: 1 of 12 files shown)


ondrejmirtes:
Yeah, let's try this, thank you :)

ondrejmirtes:
FYI I did a small fix (https://github.com/phpstan/phpstan/issues/9406) - look at https://phpstan.org/r/9daa180b-9405-4e98-9b33-075e5464cbc4 - if `app2()` isn't marked as dead catch, `app()` can't be either.

See the fix: https://github.com/phpstan/phpstan-src/commit/c587dd1d870f7a67623e1a3bb7f2811fda688bab

PR #2400, opened 2023-05-13
Lazy initialization of AnalyserResult::$errors

takaram:
`AnalyserResult::$errors` array is sorted in the class' constructor, but it doesn't always need to be sorted.
This PR makes the property be initialized when `getErrors` is called.

In `AnalyseApplication::analyse` we may get two `AnalyserResult` instances. One of them is just an intermediate result, so its error array doesn't have to be sorted.
Sorting an array with `usort` may take a few seconds if the array has thousands of elements.

### DIFF
--- a/src/Analyser/AnalyserResult.php
+++ b/src/Analyser/AnalyserResult.php
@@ -9,18 +9,18 @@ use function usort;
 class AnalyserResult
 {
 
-	/** @var list<Error> */
-	private array $unorderedErrors;
+	/** @var list<Error>|null */
+	private ?array $errors = null;
 
 	/**
-	 * @param list<Error> $errors
+	 * @param list<Error> $unorderedErrors
 	 * @param list<CollectedData> $collectedData
 	 * @param list<string> $internalErrors
 	 * @param array<string, array<string>>|null $dependencies
 	 * @param array<string, array<RootExportedNode>> $exportedNodes
 	 */
 	public function __construct(
-		private array $errors,
+		private array $unorderedErrors,
 		private array $internalErrors,
 		private array $collectedData,
 		private ?array $dependencies,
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Can you please share a use case where AnalyserResult is created but errors aren't read? Thank you.

mad-briller:
@ondrejmirtes i think this isn't in relation to them being created but not read, but more to avoid the cost of sorting twice

on line 80 of AnalyseApplication there is this:
```php
$intermediateAnalyserResult = $this->runAnalyser(
```

which produces an `AnalyserResult` and as such has sorted the errors once,
but then later on on line 94 there is this:

```php
$intermediateAnalyserResult = new AnalyserResult(
	array_merge($intermediateAnalyserResult->getErrors(), $stubErrors),
```

which will re-sort the errors, including any additional errors produced in stubs 

this change makes it so that the errors are only sorted once

ondrejmirtes:
@mad-briller Are you sure? There's `$intermediateAnalyserResult->getErrors()` call in the code you posted.

mad-briller:
yeah, the `getErrors` call in the code i posted returns already-sorted errors, which will then be sorted again in the `AnalyserResult` [constructor](https://github.com/phpstan/phpstan-src/pull/2400/files#diff-e2aa8d931dedcc156cc1444bb2c89a3e2d2df401203931a0420ef19f27d2f650L34-L45)

ondrejmirtes:
But after this PR they will get sorted twice anyway, right?

mad-briller:
the pr actually changes the call to `getError` in the second instance to a `getUnorderedErrors`, the code samples i used were from before the change had happened to explain what it was targeting 
```php
			$intermediateAnalyserResult = new AnalyserResult(
					array_merge($intermediateAnalyserResult->getUnorderedErrors(), $stubErrors),
					```

ondrejmirtes:
I get it now 😊

ondrejmirtes:
Thank you.

takaram:
@mad-briller 
Thank you so much for answering the question! I didn't have enough time to answer😅

PR #2401, opened 2023-05-14
Fix MissingReadOnlyPropertyAssignRule reporting an error on NeverType

takaram:
Closes phpstan/phpstan#8412


ondrejmirtes:
I fixed it a different way and in multiple places: https://github.com/phpstan/phpstan-src/commit/d09401df4514a16eb38334cf4187093a4e66eefb

I reused your regression test: https://github.com/phpstan/phpstan-src/commit/ffc0495d89763b65d93256928cb95bc4b07f6138

Thank you.

PR #2404, opened 2023-05-15
Do less file IO in CachedParser

staabm:
we already read the file, don' tread it again in the wrapped parser

grafik


ondrejmirtes:
You'd break some stuff with this change, for example PathRoutingParser.

staabm:
would you be open to adjust the ParserInterface so we don't need to re-read files over and over?

ondrejmirtes:
No, I doubt there's actually any real performance benefit. Files are cached on the OS level and maybe even PHP level.

PR #2406, opened 2023-05-16
Filtering default stub files on Windows

maryo:
`ComposerHelper::getVendorDirFromComposerConfig($this->currentWorkingDirectory, $composerConfig)`
returns path separated by `/` (`$root . '/' . trim($vendorDirectory, '/')`) and  `$composerConfig['config']['vendor-dir']` could contain both separators.

On Windows, `$this->getStubFiles()` returns paths separated by `\`. So the vendor stub files are never filtered out on Windows.

`getVendorDirFromComposerConfig` could already return it normalized but I was not brave enough to change it.

### DIFF
--- a/src/PhpDoc/DefaultStubFilesProvider.php
+++ b/src/PhpDoc/DefaultStubFilesProvider.php
@@ -7,6 +7,7 @@ use PHPStan\Internal\ComposerHelper;
 use function array_filter;
 use function array_values;
 use function strpos;
+use function strtr;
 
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
This change needs a test (that fails when the changes in `src/` are removed).

maryo:
I see. Some tests added.
`testGetProjectStubFilesWhenPathContainsWindowsSeparator` fails without the changes

ondrejmirtes:
Thank you!

PR #2408, opened 2023-05-17
Fix ConstantArrayTypeBuilder's isList flag

jlherren:
This fixes the two `inspect()` calls not being reported in this example: https://phpstan.org/r/2d082d39-0875-4947-a864-0ce2a5ffeae1

Since lists aren't described as `list{}` (yet?), creating a proper test was not so trivial, I hope it's fine like this.

### DIFF
--- a/src/Type/Constant/ConstantArrayTypeBuilder.php
+++ b/src/Type/Constant/ConstantArrayTypeBuilder.php
@@ -18,6 +18,7 @@ use function count;
 use function in_array;
 use function is_float;
 use function max;
+use function min;
 use function range;
 
 /** @api */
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/ArrayIsListTest.php
```diff
@@ -0,0 +1,25 @@
...
+	public function testNotList(): void
+	{
+		$this->processFile(
+			__DIR__ . '/data/array-is-list.php',
+			function (Node $node, Scope $scope): void {
+				if (!$node instanceof Node\Stmt\Return_ || $node->expr === null) {
+					return;
+				}
+				$type = $scope->getType($node->expr);
+				$this->assertFalse($type->isList()->yes());
```
ondrejmirtes: This test is weird, no test in phpstan-src looks like this :) Please write a test that tests type acceptance, so for example CallMethodsRuleTest, or ReturnTypeRuleTest. Thank you!

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/ArrayIsListTest.php
jlherren: Apologies, I struggled with writing a classical test for this, but I can't remember why. I now converted it to a test in ReturnTypeRuleTest.

ondrejmirtes:
Thank you!

PR #2410, opened 2023-05-19
Fix loose comparison against enum

schlndh:
Here is an attempt to make PHPStan understand loose comparisons with enums better.

My motivation for this is that I'm converting a bunch of "enum classes" (based on `consistence/consistence`) into native enums. And I encountered a case like this, which was not reported by PHPStan, even though it doesn't work after the conversion: `in_array($enum, ['A', 'B'])`. Previously it worked because the enum class had `__toString()` implemented.

Here are the results for a bunch of loose comparisons: https://3v4l.org/Huc7t

Note that the PHP documentation says the following:

> While identity comparison (=== and !==) can be applied to arbitrary values, the other comparison operators should only be applied to comparable values. The result of comparing incomparable values is undefined, and should not be relied upon. 

So it is possible, that what I'm trying to achieve either doesn't make sense, or may be too version dependent. However, as far as I can tell, it does not clearly specify which types are comparable.

### DIFF
--- a/src/Type/Php/InArrayFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/InArrayFunctionTypeSpecifyingExtension.php
@@ -80,8 +80,8 @@ class InArrayFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyingEx
 			|| (
 				$context->false()
 				&& (
-					count(TypeUtils::getConstantScalars($needleType)) > 0
-					|| count(TypeUtils::getEnumCaseObjects($arrayValueType)) > 0
+					count(TypeUtils::getConstantScalars($needleType)) === 1
+					|| count(TypeUtils::getEnumCaseObjects($needleType)) === 1
 				)
 			)
 		) {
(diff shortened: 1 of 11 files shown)


ondrejmirtes (changes requested):
`InArrayFunctionTypeSpecifyingExtension` is a type-specifying extension, meaning the changes inside should also have their own tests - what the type of the argument is after true/false `in_array()` check.

src/Type/Php/InArrayFunctionTypeSpecifyingExtension.php
```diff
@@ -75,8 +80,8 @@ public function specifyTypes(FunctionReflection $functionReflection, FuncCall $n
 			|| (
 				$context->false()
 				&& (
-					count(TypeUtils::getConstantScalars($needleType)) > 0
-					|| count(TypeUtils::getEnumCaseObjects($arrayValueType)) > 0
+					count(TypeUtils::getConstantScalars($needleType)) === 1
+					|| count(TypeUtils::getEnumCaseObjects($needleType)) === 1
```
schlndh, line 84: This fixes https://phpstan.org/r/9c8038c6-783f-4eab-beab-e9ee525cff8c

src/Type/Php/InArrayFunctionTypeSpecifyingExtension.php
```diff
@@ -75,8 +80,8 @@ public function specifyTypes(FunctionReflection $functionReflection, FuncCall $n
 			|| (
```
schlndh, line 80: I haven't been able to figure out whether `$context->true()` can actually do anything. I tried commenting it out and it didn't break any tests. As far as I was able to tell, the specified types are mostly intersected with the original type, in which case widening the specified type doesn't do anything.

ondrejmirtes:
Thank you!

PR #2412, opened 2023-05-20
Run Rules-Tests in 2 parallel CI jobs

staabm:
The Rule Levels tests are pretty slow. its the slowest part of the CI pipline atm.

try running them in parallel to shorten the feedback loop.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -135,6 +135,11 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
 
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


staabm:
not sure its worth merging.

ondrejmirtes:
Hi, yeah, these are already pretty fast.

But if you want to make a difference, the tests in phpstan-doctrine needs rewriting. Ideally none of them should use LevelsTestCase. It's really slow. They were written before I made TypeInferenceTestCase reusable.

So most of them should be remade for TypeInferenceTestCase. And some of them (the ones who test EntityRepositoryClassReflectionExtension) should be tested through RuleTestCase for CallMethodsRule.

Feel free to send improvements there, I'd really appreciate it. You can do it one-by-one to make sure we're on the same page. Thank you!

PR #2415, opened 2023-05-21
fix enum type subtracting

schlndh:
It seems that enum type subtracting doesn't work correctly in some cases. See https://phpstan.org/r/ed7bac55-878e-4ecd-b1b6-6fbba5959741 PHPStan shouldn't complain about the second nested if.

I played around with it in a debugger and this is what I came up with. I checked the usages of `changeSubtractedType` and I wasn't able to figure out why considering `$this->subtractedType` would be necessary/correct. But my knowledge of PHPStan internals is very limited, so it's quite possible that I missed something.

In either case, I'm hoping that at least the test will be useful.


ondrejmirtes:
Perfect, thank you!

PR #2417, opened 2023-05-22
Smarter extract() with constant arrays

MartinMystikJonas:
As discussed in https://github.com/phpstan/phpstan/issues/9331

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2108,7 +2108,40 @@ class NodeScopeResolver
 			}
 
 			if (isset($functionReflection) && $functionReflection->getName() === 'extract') {
-				$scope = $scope->afterExtractCall();
+				$extractedArg = $expr->getArgs()[0]->value;
+				$extractedType = $scope->getType($extractedArg);
+				if (count($extractedType->getConstantArrays()) > 0) {
+					$properties = [];
+					$optionalProperties = [];
+					$refCount = [];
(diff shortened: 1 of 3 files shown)


MartinMystikJonas:
```php
if (!$keyType instanceof ConstantStringType) {
  // integers as variable names not allowed
  continue;
}
```
This code raises `Doing instanceof PHPStan\Type\Constant\ConstantStringType is error-prone and deprecated. Use Type::getConstantStrings() instead.  `

But it is used this way in other places. Should I rewrite it somehow or ignore it using baseline?

ondrejmirtes:
> But it is used this way in other places.

I don't see a single `instanceof ConstantStringType` in NodeScopeResolver. Please use what the tip tells you - either `Type::isString()` or `count($type->getConstantStrings()) === 0`. I'd prefer the first one.

ondrejmirtes:
please rebase your pull request and change its target to 1.10.x.

ondrejmirtes:
`git rebase --onto` is your friend ;) So `git rebase --onto origin/1.10.x HEAD^` is going to fix this for you. Alternatively, you can `git reset --hard origin/1.10.x` and then `git cherry-pick` that one commit of yours. The result is gonna be the same.

ondrejmirtes:
Really nice, thank you!

herndlm:
Thx, this most likely gets rid of a couple of baseline entries in one of our legacy apps 😊

PR #2418, opened 2023-05-27
Union of enum cases accepts the enum class type with negated cases

takaram:
This PR changes the result of removing an enum case from its class.

```php
enum TestEnum
{
    case A;
    case B;
    case C;
}

function test(TestEnum $enum): void
{
    if ($enum !== TestEnum::A) {
        \PHPStan\dumpType($enum);
    }
}
```
 - Before: `TestEnum~TestEnum::A`
 - After: `TestEnum::B|TestEnum::C`

Closes phpstan/phpstan#8846

### DIFF
--- a/tests/PHPStan/Type/UnionTypeTest.php
+++ b/tests/PHPStan/Type/UnionTypeTest.php
@@ -6,9 +6,6 @@ use DateTime;
 use DateTimeImmutable;
 use Exception;
 use Iterator;
-use PHPStan\Fixture\AnotherTestEnum;
-use PHPStan\Fixture\ManyCasesTestEnum;
-use PHPStan\Fixture\TestEnum;
 use PHPStan\Reflection\Native\NativeParameterReflection;
 use PHPStan\Reflection\PassedByReference;
 use PHPStan\Testing\PHPStanTestCase;
@@ -36,6 +33,7 @@ use function array_merge;
 use function array_reverse;
 use function get_class;
 use function sprintf;
+use const PHP_VERSION_ID;
 
 class UnionTypeTest extends PHPStanTestCase
 {
@@ -933,9 +931,9 @@ class UnionTypeTest extends PHPStanTestCase
 		$this->assertSame($expectedTypeOnlyDescription, $type->describe(VerbosityLevel::typeOnly()));
 	}
 
-	public function dataAccepts(): array
+	public function dataAccepts(): iterable
 	{
-		return [
+		yield from [
 			[
 				new UnionType([new CallableType(), new NullType()]),
 				new ClosureType([], new StringType(), false),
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Instead of changing what `TypeCombinator::remove()` does in this case, I'd rather be if you fixed so that `Foo::THREE|Foo::TWO` accepts `Foo~Foo::ONE`. You should be able to debug that by looking into what `UnionType::accepts()` does with `ObjectType` in this case. Some code that does `$type->getEnumCases()` and compares them (because both sides will report the same enum cases) should be put somewhere in there.

takaram:
@ondrejmirtes 
Thank you for your suggestion. I'll revise this PR.

tests/PHPStan/Type/UnionTypeTest.php
```diff
@@ -1186,6 +1188,72 @@ public function dataAccepts(): array
 			],
 
 		];
+
+		if (PHP_VERSION_ID < 80100) {
+			return;
```
ondrejmirtes: It's a bit tricky to mark the rest of the dataProvider as tested only on PHP 8.1. Please reverse the condition and put the enum `yield`s into that condition:

```php
if (PHP_VERSION_ID >= 80100) {
    yield...
}
```

If PHPCS complains, move one existing yield below the condition.

tests/PHPStan/Type/UnionTypeTest.php
```diff
@@ -1186,6 +1188,72 @@ public function dataAccepts(): array
...
+		];
+
+		yield [
+			new UnionType([
+				new EnumCaseObjectType('PHPStan\Fixture\TestEnum', 'ONE'),
+				new NullType(),
+			]),
+			new ObjectType(
+				'PHPStan\Fixture\TestEnum',
+				new EnumCaseObjectType('PHPStan\Fixture\TestEnum', 'TWO'),
```
ondrejmirtes: One more test please:

```php

		yield [
			new UnionType([
				new EnumCaseObjectType('PHPStan\Fixture\TestEnum', 'ONE'),
				new NullType(),
			]),
			new UnionType([
				new ObjectType(
					'PHPStan\Fixture\TestEnum',
					new EnumCaseObjectType('PHPStan\Fixture\TestEnum', 'TWO'),
				),
				new NullType(),
			]),
			TrinaryLogic::createYes(),
		);
```

ondrejmirtes (changes requested):

tests/PHPStan/Type/UnionTypeTest.php
takaram: I reversed the condition and moved some of the existing yields.

tests/PHPStan/Type/UnionTypeTest.php
takaram: I added this test case.

takaram:
@ondrejmirtes 
Thank you for your comments.
Could you check it again?

ondrejmirtes:
Alright, one more thing: you're trying to fix https://github.com/phpstan/phpstan/issues/8846. Which means we need a regression test in the test class for the rule that reports the false positive. Which means you should write a test in Functions\ReturnTypeRuleTest. And make sure it fails without your changes in `src/` :)

ondrejmirtes:
Perfect, thank you!

PR #2419, opened 2023-05-27
base64_decode returns false only in strict mode

kamil-tekiela:


ondrejmirtes:
Please verify in NodeScopeResolverTest this does what you think it does.

kamil-tekiela:
> Please verify in NodeScopeResolverTest this does what you think it does.

I have added a test and it seems to do what I expect. Unless, I didn't write the test correctly.

ondrejmirtes:
Thank you!

PR #2420, opened 2023-05-28
Static methods are no longer callable

staabm:
refs https://github.com/phpstan/phpstan/issues/5782

as suggested in https://github.com/phpstan/phpstan/issues/5782#issuecomment-1559321986


src/Type/Constant/ConstantArrayType.php
```diff
@@ -444,15 +445,32 @@ public function equals(Type $type): bool
...
+
+			$type = $typeAndMethodName->getType();
+			$methodName = $typeAndMethodName->getMethod();
+
+			if (!$type->hasMethod($methodName)->yes()) {
+				continue;
+			}
+
+			$method = $type->getMethod($methodName, new OutOfClassScope());
+			if ($method->isStatic()) {
```
staabm: actually this should depend on php-version.
within the type-classes we don't have access to `PhpVersion` though

staabm:
I think it works as expected now - at least what php8 is concerned.
(The PR title is wrong, as non-static methods are no longer callable)

we don't know the php version in the type-system, so I can't differentiate between php7/8. is it acceptable as is?

ondrejmirtes (changes requested):
There are examples how to approach DI-registered values or services in places where DI isn't available:

* https://github.com/phpstan/phpstan-src/blob/1.11.x/src/DependencyInjection/BleedingEdgeToggle.php
* https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Reflection/ReflectionProviderStaticAccessor.php

For current methods we could add `PhpVersionAccessor` so that we can ask about the PHP version.

For future methods we should require `PhpVersion` as a parameter.

tests/PHPStan/Rules/Methods/data/call-methods.php
```diff
@@ -656,7 +656,7 @@ public function test(callable $str)
 	{
 		$this->test('date');
 		$this->test('nonexistentFunction');
-		$this->test('Test\CheckIsCallable::test');
+		// $this->test('Test\CheckIsCallable::test'); differs between php7/8; tested separately
```
staabm, line 659: moved this line into a separate new test, so I don't need to duplicate the whole `call-methods.php` assertions and can keep the diff small

staabm:
the errors in the Carbon build job seem legit. [related source](https://github.com/briannesbitt/Carbon/blob/94e1178370787843424b276b334a03da6f0abb3d/src/Carbon/CarbonPeriod.php#L181-L200)

ondrejmirtes:
Thank you.

PR #2426, opened 2023-05-30
Add a way to create a RuleLevelHelper in tests

VincentLanglet:
Hi,

The `RuleLevelHelper::construct()` method is not covered by Backward compatibility.
If I write my own PHPStan custom rule, I can inject the RuleLevelHelper in the constructor,
but then, in tests, I'll need to write something like this:
https://github.com/phpstan/phpstan-src/blob/1.11.x/tests/PHPStan/Rules/Cast/EchoRuleTest.php#L16-L21

What could be great, would be to provide a `getRuleLevelHelper` method which allow to avoid writing `new RuleLevelHelper()` in my code base. It's just a default RuleLevelHelper with every config check configurable.
If later a new param `checkSomething` is added, it's just needed to add it with a default value and it won't change anything for the users of this method.

WDYT about it @ondrejmirtes ?

### DIFF
--- a/src/Testing/RuleTestCase.php
+++ b/src/Testing/RuleTestCase.php
@@ -25,6 +25,7 @@ use PHPStan\Rules\Properties\DirectReadWritePropertiesExtensionProvider;
 use PHPStan\Rules\Properties\ReadWritePropertiesExtension;
 use PHPStan\Rules\Properties\ReadWritePropertiesExtensionProvider;
 use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\Type\FileTypeMapper;
 use function array_map;
 use function count;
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Use `self::getContainer()->getByType(RuleLevelHelper::class)`.

VincentLanglet:
> self::getContainer()->getByType(RuleLevelHelper::class)

But then, how do I configure the `checkBenevolentUnionTypes`, `checkUnionTypes`, ... configuration ?

ondrejmirtes:
The ideal way to test different levels is through LevelsTestCase: https://apiref.phpstan.org/1.11.x/PHPStan.Testing.LevelsTestCase.html

You can see it in use for example here: https://github.com/phpstan/phpstan-strict-rules/blob/1.5.x/tests/Levels/LevelsIntegrationTest.php

The way to use it is to write dataTopics, have `.php` files with the same names in `getDataPath()`, and just run the test. It will generate or change the result JSONs automatically according to the latest results.

ondrejmirtes:
To test things that aren't part of any level, like `checkBenevolentUnionTypes`, is to configure it with `getAdditionalConfigFiles()` method in your RuleTestCase, as it's described in the documentation: https://phpstan.org/developing-extensions/testing#custom-rules

VincentLanglet:
Thanks

PR #2430, opened 2023-06-01
Minor refinements to MutatingScope empty() and typehint helper.

mad-briller:
a simplification to MutatingScope `empty()` handling

and a micro-optimization in `TypehintHelper` to prevent multiple calls to `strtolower` and prevent checking mutually exclusive cases when a match is already found


mad-briller:
not sure about the action failures, all seem related to the `processNode()` method which i have not changed here

ondrejmirtes:
Thank you!

PR #2432, opened 2023-06-04
Fix more in array issues

schlndh:
Here is a fix for https://phpstan.org/r/85c7b5e7-4993-4285-93b4-45e77aeeaeda

It should also fix https://github.com/phpstan/phpstan/issues/6959

### DIFF
--- a/tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
+++ b/tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
@@ -256,6 +256,11 @@ class ImpossibleCheckTypeFunctionCallRuleTest extends RuleTestCase
 					889,
 					'Remove remaining cases below this one and this error will disappear too.',
 				],
+				[
+					'Call to function in_array() with arguments 1, array<string> and true will always evaluate to false.',
+					927,
+					'Because the type is coming from a PHPDoc, you can turn off this check by setting <fg=cyan>treatPhpDocTypesAsCertain: false</> in your <fg=cyan>%configurationFile%</>.',
+				],
 			],
 		);
 	}
@@ -361,6 +366,11 @@ class ImpossibleCheckTypeFunctionCallRuleTest extends RuleTestCase
 					'Call to function is_numeric() with \'blabla\' will always evaluate to false.',
 					694,
 				],
+				[
+					'Call to function in_array() with arguments 1, array<string> and true will always evaluate to false.',
+					927,
+					'Because the type is coming from a PHPDoc, you can turn off this check by setting <fg=cyan>treatPhpDocTypesAsCertain: false</> in your <fg=cyan>%configurationFile%</>.',
+				],
 			],
 		);
 	}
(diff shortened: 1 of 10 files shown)


tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
```diff
@@ -910,7 +920,105 @@ public function testLooseComparisonAgainstEnums(): void
...
+			[
+				'Call to function in_array() with LooseComparisonAgainstEnums\FooUnitEnum::A and non-empty-array<LooseComparisonAgainstEnums\FooUnitEnum::A> will always evaluate to true.',
+				162,
+				$tipText,
+			],
+			[
+				'Call to function in_array() with arguments LooseComparisonAgainstEnums\FooUnitEnum::A, non-empty-array<LooseComparisonAgainstEnums\FooUnitEnum::A> and false will always evaluate to true.',
+				165,
+				'BUG',
+				//$tipText,
```
schlndh, line 982: There seem to be some issues with `treatPhpdocTypesAsCertain`, native types, ... I looked into it in a debugger, but I wasn't able to figure it out completely. Here is a smaller example of the issue: https://phpstan.org/r/ae2c5d39-7ecb-4dcc-81b6-15168ed320c7

tests/PHPStan/Rules/Comparison/data/loose-comparison-against-enums.php
```diff
@@ -106,4 +106,72 @@ public function nullableEnum(?FooUnitEnum $u, string $s): void
...
+	}
+
+    /**
+     * @param non-empty-array<FooUnitEnum::A> $nonEmptyA
+     * @return void
+     */
+    public function nonEmptyArray(array $nonEmptyA): void
+    {
+        if (in_array(FooUnitEnum::B, $nonEmptyA)) {
+        }
```
schlndh, line 160: This slightly weird order of the ifs is not accidental. I had to move this if first, otherwise phpstan would complain about it being always false with `treatPhpDocTypesAsCertain = false`.

tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
ondrejmirtes: Yeah, this a problem because TypeSpecifier is run only once on `expressionTypes`, not `nativeExpressionTypes`.

This should fix the problem (you can verify that) https://github.com/phpstan/phpstan-src/pull/2179 but comes with performance cost.

ondrejmirtes (changes requested):
Please add a regression test for #6959. Thanks.

tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
schlndh: Thanks for the info. I tried it (merged it into my branch and fixed the conflicts), but it didn't help in this specific case. I.e. this test still works without change and the snippet I linked still has incorrect native type.

It did generate the tip in some other tests, so I guess it's probably incomplete.

ondrejmirtes:
Thank you.

PR #2434, opened 2023-06-05
feat: improve Map::remove() @throws inference

baci266:
What applies to Ds\Map::get() should also apply to Ds\Map::remove().

https://github.com/phpstan/phpstan/issues/9066

### DIFF
--- a/tests/PHPStan/Rules/Exceptions/data/bug-9066.php
+++ b/tests/PHPStan/Rules/Exceptions/data/bug-9066.php
@@ -4,10 +4,18 @@ namespace Bug9066;
 
 class Foo
 {
-	public function mayThrow()
+	public function getMayThrow()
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Exceptions/data/bug-9066.php
```diff
@@ -8,6 +8,7 @@ public function mayThrow()
 	{	$map = new \Ds\Map();
 		try {
 			$map->get('1');
+			$map->remove('1');
```
ondrejmirtes, line 18: Please add a separate method for testing `remove` where it's supposed to throw an exception.

ondrejmirtes (changes requested):
And also base your PR on 1.10.x, thanks :)

ondrejmirtes:
Thank you!

PR #2435, opened 2023-06-05
Added failling test for bug 9391

staabm:
closes https://github.com/phpstan/phpstan/issues/9391


staabm:
let me have a look whether we have the same problem for `__serialize` and `__sleep`

ondrejmirtes:
I don't think so, union isn't possible in the first place.

staabm:
> I don't think so, union isn't possible in the first place.

you are right. I added a test which shows that we don't need a fix

ondrejmirtes:
Thank you.

PR #2436, opened 2023-06-05
Dectect reading class constants in PHPDoc

rvanvelzen:
Fix phpstan/phpstan#6415, fix phpstan/phpstan#7184, fix phpstan/phpstan#9385

### DIFF
--- a/src/PhpDoc/PhpDocPrivateClassConstFetchFinder.php
+++ b/src/PhpDoc/PhpDocPrivateClassConstFetchFinder.php
@@ -1,0 +1,44 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDoc;
+
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\Constant\ClassConstantFetch;
+use PHPStan\PhpDocParser\Ast\NodeTraverser;
+use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\FileTypeMapper;
+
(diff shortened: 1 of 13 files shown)


src/PhpDoc/PhpDocPrivateClassConstFetchFinder.php
```diff
@@ -0,0 +1,41 @@
...
+			throw new ShouldNotHappenException();
+		}
+
+		if (Strings::match($phpDoc, '(\bself\s*::)i') === null) {
+			return [];
+		}
+
+		$visitor = new PhpDocPrivateClassConstFetchFinderVisitor($scope->getClassReflection(), $scope);
+
+		$phpDocNode = $this->phpDocStringResolver->resolve($phpDoc);
```
ondrejmirtes: I really like this, it's almost perfect. What gives me a little pause is that all PHPDocs are now parsed multiple times which would lead to performance problems. The correct way to access them is through FileTypeMapper. Also you shouldn't have to do any regexes, but correctly resolve the class name with NameScope.

Of course the issue is that the thing returned by FileTypeMapper is ResolvedPhpDocBlock. Which is about accessing the useful PHPDoc things like ParamTag and many others.

One way to solve this would be to add `getClassConstFetches` on ResolvedPhpDocBlock. But the class is a bit tricky to work with - because it can be created multiple ways (create, createEmpty, merge), we need to be careful every time a new property is added.

ondrejmirtes (changes requested):

PR #2437, opened 2023-06-06
implement gettype() return type extension

staabm:
refs https://github.com/phpstan/phpstan/issues/8614#issuecomment-1367891117

### DIFF
--- a/src/Type/Php/GettypeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/GettypeFunctionReturnTypeExtension.php
@@ -1,0 +1,90 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\IntersectionType;
+use PHPStan\Type\ResourceType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\TypeTraverser;
+use PHPStan\Type\UnionType;
+use function count;
+
+class GettypeFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+{
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return $functionReflection->getName() === 'gettype';
+	}
+
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
+	{
+		if (count($functionCall->getArgs()) < 1) {
+			return null;
+		}
+
+		$valueType = $scope->getType($functionCall->getArgs()[0]->value);
+
+		return TypeTraverser::map($valueType, static function (Type $valueType, callable $traverse): Type {
+			if ($valueType instanceof UnionType || $valueType instanceof IntersectionType) {
+				return $traverse($valueType);
+			}
+
+			if ($valueType->isString()->yes()) {
+				return new ConstantStringType('string');
+			}
+			if ($valueType->isArray()->yes()) {
+				return new ConstantStringType('array');
+			}
+
+			if ($valueType->isBoolean()->yes()) {
+				return new ConstantStringType('boolean');
+			}
+
+			$resource = new ResourceType();
+			if ($resource->isSuperTypeOf($valueType)->yes()) {
+				return new UnionType([
+					new ConstantStringType('resource'),
+					new ConstantStringType('resource (closed)'),
+				]);
+			}
+
+			if ($valueType->isInteger()->yes()) {
+				return new ConstantStringType('integer');
+			}
+
+			if ($valueType->isFloat()->yes()) {
+				// for historical reasons "double" is returned in case of a float, and not simply "float"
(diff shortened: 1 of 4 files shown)


src/Type/Php/GettypeFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,85 @@
...
+
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
+	{
+		if (count($functionCall->getArgs()) < 1) {
+			return null;
+		}
+
+		$valueType = $scope->getType($functionCall->getArgs()[0]->value);
+
+		$type = TypeTraverser::map($valueType, static function (Type $valueType, callable $traverse): Type {
```
staabm: alternatively we might implement `Type->getType()` or `Type->getTypeString()` instead

src/Type/Php/GettypeFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,85 @@
...
+
+			if ($valueType->isNull()->yes()) {
+				return new ConstantStringType('NULL');
+			}
+
+			if ($valueType->isObject()->yes()) {
+				return new ConstantStringType('object');
+			}
+
+			return $valueType;
```
ondrejmirtes: I disagree with this last line. It should probably do `new StringType()` instead.

ondrejmirtes (changes requested):

src/Type/Php/GettypeFunctionReturnTypeExtension.php
staabm: good catch, fixed.

src/Analyser/TypeSpecifier.php
```diff
@@ -1419,15 +1419,19 @@ private function processBooleanConditionalTypes(Scope $scope, SpecifiedTypes $le
 	private function findTypeExpressionsFromBinaryOperation(Scope $scope, Node\Expr\BinaryOp $binaryOperation): ?array
 	{
 		$leftType = $scope->getType($binaryOperation->left);
-		$rightType = $scope->getType($binaryOperation->right);
 		if (
 			$leftType instanceof ConstantScalarType
+			&& !$binaryOperation->right instanceof Node\Scalar\String_
```
ondrejmirtes: Eh, what's so special about string? Why did you add this change?

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
staabm: its required to make https://github.com/staabm/phpstan-src/blob/1.10.x/tests/PHPStan/Analyser/data/bug-6901.php pass

otherwise - because of switch beeing re-written to comparisons and the new return type extension - `$leftType` and `$rightType`  will resolve to constant strings and the method - as it was before the PR - would not specify the `FuncCall` expression but the `String_`

ondrejmirtes: That feels like a wrong placement for such fix.

`gettype` is handled in TypeSpecifier around line 1072. Wouldn't it help to first check for `gettype` condition and only after that to perform the thing on line 1424?

staabm: I think we will have the same problem for other functions being used in a switch statement/identical comparison and have a return type extension with constant scalars.

we cannot detect it in line 1072, as the problem is not within the `FuncCall` itself, but the comparison.

see the bug in action in https://phpstan.org/r/d00e29f6-3163-414b-b83d-f591fd1eaf7d
(I have added a separate test-case for it in [e60ff34](https://github.com/phpstan/phpstan-src/pull/2437/commits/e60ff34b745a99fac5d6c7b18a664e3d8473e8ec))

staabm: If you prefer we could separate that bugfix in a new PR

ondrejmirtes: Yeah https://phpstan.org/r/d00e29f6-3163-414b-b83d-f591fd1eaf7d shouldn't be happening, but I'm not sure if it's the same bug...

staabm: ok, turned it into a separate bug: https://github.com/phpstan/phpstan/issues/9404

staabm: I had another look at it.

you were right that https://github.com/phpstan/phpstan/issues/9404 does not related to the problem here.

I think the fix is right though and the description in https://github.com/phpstan/phpstan-src/pull/2437/files#r1220239214 is accurate

staabm: to simplify the test in question:


```php
<?php

/**
 * @param int $y
 */
function foo($y): void
{
	switch (gettype($y)) {
		case "integer":
			assertType('int', $y);
			break;
		default:
			assertType('*NEVER*', $y);
	}
}
```

runs into 
```
------ ------------------------------------ 
  Line   tst.php                             
 ------ ------------------------------------ 
  15     Expected type *NEVER*, actual: int  
 ------ ------------------------------------ 
```

in this PR, when we drop the `... instanceof Node\Scalar\String_` changes

ondrejmirtes:
As I said, I didn't like the changes in TypeSpecifier. I made some changes on 1.10.x, essentially doing this https://github.com/phpstan/phpstan-src/pull/2437#discussion_r1221026739

And now the tests in your PR pass. I'm gonna merge it if the pipeline is going to be green. Thanks.

ondrejmirtes (approved):

staabm:
Thanks

PR #2438, opened 2023-06-06
More precise `getimagesize*` return type

staabm:
see https://www.php.net/manual/en/function.getimagesize.php

analog https://github.com/vimeo/psalm/pull/9875


ondrejmirtes (changes requested):
Changing return types does not need to happen in bleedingEdge, only the parameter types.

ondrejmirtes:
Thank you.

PR #2439, opened 2023-06-07
Fix param-type specification interferes with root-expression types

staabm:
closes https://github.com/phpstan/phpstan/issues/9404

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1099,7 +1099,9 @@ class TypeSpecifier
 			}
 
 			if ($type !== null) {
-				return $this->create($exprNode->getArgs()[0]->value, $type, $context, false, $scope, $rootExpr);
+				$callType = $this->create($exprNode, $constantType, $context, false, $scope, $rootExpr);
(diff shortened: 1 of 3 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -180,15 +180,13 @@ public function specifyTypesInCondition(
 				return $this->create($exprNode, new ObjectWithoutClassType(), $context, false, $scope, $rootExpr);
 			}
 		} elseif ($expr instanceof Node\Expr\BinaryOp\Identical) {
+			$binaryExpressionSpecifiedType = null;
 			$expressions = $this->findTypeExpressionsFromBinaryOperation($scope, $expr);
 			if ($expressions !== null) {
 				$exprNode = $expressions[0];
 				$constantType = $expressions[1];
 
-				$specifiedType = $this->specifyTypesForConstantBinaryExpression($exprNode, $constantType, $context, $scope, $rootExpr);
```
staabm: this case might specify types besides `$exprNode`, therefore remember this types and later on return them in union with other known things within the current scope instead of early returning

staabm:
the PR title is more generic then the actual fixed turned out in the end

ondrejmirtes:
Thank you.

PR #2441, opened 2023-06-08
Fix int-mask-of to work well with operations on the int mask

staabm:
closes https://github.com/phpstan/phpstan/issues/9384


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -659,10 +659,15 @@ public function getBitwiseOrType(Expr $left, Expr $right, callable $getTypeCallb
...
 		}
 
-		if (TypeCombinator::union($leftType->toNumber(), $rightType->toNumber()) instanceof ErrorType) {
+		$unionType = TypeCombinator::union($leftType->toNumber(), $rightType->toNumber());
+		if ($unionType instanceof ErrorType) {
 			return new ErrorType();
 		}
 
+		if ($unionType->isInteger()->yes()) {
+			return $unionType;
```
ondrejmirtes, line 668: 1) Other bitwise operations need to be fixed as well
2) Is this really correct? What if I do `int<100, 150> | int<500, 1000>`?

ondrejmirtes (changes requested):

src/Reflection/InitializerExprTypeResolver.php
jtojnar: 1. It would be a nice to have or at least.
2. For a simpler example, `1 ∈ int<1,1>` and `2 ∈ int<2,2>` but `1|2 = 3 ∉ (int<1,1> | int<2,2>)`

PR #2442, opened 2023-06-08
Fix #9413 - allow Redis::connect() to have 7 parameters

thomas-0816:
Fixes https://github.com/phpstan/phpstan/issues/9413

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -9262,7 +9262,7 @@ return [
 'Redis::client' => ['mixed', 'command'=>'string', 'arg='=>'string'],
 'Redis::close' => ['bool'],
 'Redis::config' => ['string', 'operation'=>'string', 'key'=>'string', 'value='=>'string'],
-'Redis::connect' => ['bool', 'host'=>'string', 'port='=>'int', 'timeout='=>'float', 'persistent_id='=>'?string', 'retry_interval='=>'?int', 'read_timeout='=>'float', 'context'=>'?array{auth?:array,stream?:array}'],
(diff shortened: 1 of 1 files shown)


resources/functionMap.php
```diff
@@ -9262,7 +9262,7 @@
 'Redis::client' => ['mixed', 'command'=>'string', 'arg='=>'string'],
 'Redis::close' => ['bool'],
 'Redis::config' => ['string', 'operation'=>'string', 'key'=>'string', 'value='=>'string'],
-'Redis::connect' => ['bool', 'host'=>'string', 'port='=>'int', 'timeout='=>'float', 'persistent_id='=>'?string', 'retry_interval='=>'?int', 'read_timeout='=>'float'],
+'Redis::connect' => ['bool', 'host'=>'string', 'port='=>'int', 'timeout='=>'float', 'persistent_id='=>'?string', 'retry_interval='=>'?int', 'read_timeout='=>'float', 'context'=>'?array{auth?:array,stream?:array}'],
```
ondrejmirtes: Should be optional - `'context='`

ondrejmirtes: And what kind of arrays are those? Could be described precisely

thomas-0816: seems there are multiple structures possible for auth:
https://github.com/phpredis/phpredis/blob/develop/library.c#L4358
[NULL, 'password']
[false, 'password']
['password']

thomas-0816: for "stream" I guess we can assume array<string, mixed> ?
https://github.com/phpredis/phpredis/blob/develop/library.c#L3257

ondrejmirtes: Yeah sure :)

thomas-0816: I've added a more precise type description for the context elements.

ondrejmirtes:
Thank you.

PR #2445, opened 2023-06-09
Detect invalid key in multi dimensional array fetch

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/6315


src/Analyser/NodeScopeResolver.php
```diff
@@ -3625,7 +3626,16 @@ private function processAssignVar(
...
 			$offsetNativeTypes = [];
-			foreach (array_reverse($dimExprStack) as $dimExpr) {
+			$dimFetchStack = array_reverse($dimFetchStack);
+			$lastDimKey = array_key_last($dimFetchStack);
+			foreach ($dimFetchStack as $key => $dimFetch) {
+				$dimExpr = $dimFetch->dim;
+
+				// Callback was already called for last dim at the beginning of the method.
+				if ($key !== $lastDimKey) {
+					$nodeCallback($dimFetch, $enterExpressionAssign ? $scope->enterExpressionAssign($dimFetch) : $scope);
```
schlndh, line 3636: I'm not sure whether the order of `$nodeCallback` calls is correct. First it's called with the outermost dim-fetch, and then the inner dim-fetches starting from the left. It would seem more natural to call the outermost dim-fetch last, but I wasn't able to make it work that way.

ondrejmirtes (changes requested):
Your PR also fixes https://github.com/phpstan/phpstan/issues/6605, please add a regression test for that. Thank you.

tests/PHPStan/Rules/Arrays/data/nonexistent-offset-multi-dim.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+class X {}
```
ondrejmirtes: Typically these tests are named after the issue - so `bug-6605.php` and put in same namespace - `Bug6605` to ensure classes are unique.

Also, we shouldn't mix definitions and side effects, so ideally the code below `class X {}` should be as one of the methods in `X` instead.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Arrays/data/nonexistent-offset-multi-dim.php
schlndh: Ok. What about the tests for bug 6315? Is it fine that I put them into the already existing file, or should I move them to a separate test as well?

ondrejmirtes: Ideally a separate file :)

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

ondrejmirtes:
Thank you!

PR #2446, opened 2023-06-09
Fix key type of superglobal variables

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/9306

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -507,7 +507,7 @@ class MutatingScope implements Scope
 		}
 
 		if ($this->isGlobalVariable($variableName)) {
-			return new ArrayType(new StringType(), new MixedType($this->explicitMixedForGlobalVariables));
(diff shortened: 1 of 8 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -507,7 +507,7 @@ public function getVariableType(string $variableName): Type
 		}
 
 		if ($this->isGlobalVariable($variableName)) {
-			return new ArrayType(new StringType(), new MixedType($this->explicitMixedForGlobalVariables));
+			return new ArrayType(TypeCombinator::union(new IntegerType(), new StringType()), new MixedType($this->explicitMixedForGlobalVariables));
```
ondrejmirtes: I'd like more if this was benevolent union type. It's annoying to require users to handle `int` if it can never happen inside their app. You can do `new BenevolentUnionType([new IntegerType(), new StringType()])`.

More about this: https://phpstan.org/config-reference#checkbenevolentuniontypes

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #2447, opened 2023-06-09
Add regression test for issue 7561.

mad-briller:
refs: https://github.com/phpstan/phpstan/issues/7561


ondrejmirtes:
Would be great to find what actually fixed it and if the test fails with the fix reverted :) Thank you.

PR #2448, opened 2023-06-09
Detect unused protected symbols in final classes.

mad-briller:
closes https://github.com/phpstan/phpstan/issues/9422

### DIFF
--- a/src/Rules/DeadCode/UnusedPrivateConstantRule.php
+++ b/src/Rules/DeadCode/UnusedPrivateConstantRule.php
@@ -39,7 +39,7 @@ class UnusedPrivateConstantRule implements Rule
 
 		$constants = [];
 		foreach ($node->getConstants() as $constant) {
-			if (!$constant->isPrivate()) {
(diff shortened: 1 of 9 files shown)


mad-briller:
every time

mad-briller:
the ci just raised an interesting issue, a protected method might be called by the parent in a final class, so it's very difficult to determine if it's unused... 
this might not be accurate enough to merge

staabm:
fwiw, there is also a separate extension package which even supports detecting unused public methods:
https://github.com/TomasVotruba/unused-public

you could use it to take inspiration from, or maybe ondrej could decide whether people who want the feature beeing implented here should use the extension instead and we keep phpstan-src simple

.. just thinking loud right now..

ondrejmirtes:
Of course it's not that easy 😊 I plan to tackle detecting dead code in a comprehensive way using Collectors, but it's not on my immediate roadmap.

I don't think we can and should do just this feature request. I have a todolist of about 40 rules we could do as part of a greater "dead code" effort.

Thank you.

ondrejmirtes:
Of course if you want to hear more about it and maybe help me implement it, I can describe what I have in mind 😊

PR #2449, opened 2023-06-10
Add non-zero int type

PrinsFrank:
For integers and floats that are used as divisors, a non-zero integer/float has to be used. This was easy to implement with integer ranges, but not easy for float ranges. How could we add that?

### DIFF
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -196,6 +196,12 @@ class TypeNodeResolver
 			case 'non-negative-int':
 				return IntegerRangeType::fromInterval(0, null);
 
+			case 'non-zero-int':
+				return new UnionType([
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Hi, how common is to require all integers except zero? You can already express them today with `int<min, -1>|int<1, max>`.

As for floats - we don't have a FloatRangeType, it'd have to be added, and update all arithmetic operations in order to correctly calculate it.

PrinsFrank:
I've used that exact annotation in three different projects now, so I thought It'd be time to add an alias. If you don't think it's needed I can still use the full written form, I just thought it would be nice to have this alias.

Would you see value in adding float ranges? I would be happy to contribute something more and check if I can implement this in a new PR.

ondrejmirtes:
Hi, yes, I see value in it, you can definitely try it out :) Thank you.

PR #2450, opened 2023-06-11
Catch implicit type conversions

ryium:
Catch deprecated implicit type conversions that occur in PHP 8.1 and later.
The following BinaryOp that may cause deprecation.

This PR fixes https://github.com/phpstan/phpstan/issues/8288

※O = safe, D = deprecated, X = not safe

## Pattern A
- `%` Mod
- `<<` ShiftLeft
- `>>` ShiftRight

|L \ R|int|float|string|numeric-string|Stringable|array|
|:---------------|:-:|:-:|:-:|:-:|:-:|:-:|
|int                    |O|D|X|D|X|X|
|float                 |D|D|X|D|X|X|
|string               |X|X|X|X|X|X|
|numeric-string  |D|D|X|D|X|X|
|Stringable         |X|X|X|X|X|X|
|array                |X|X|X|X|X|X|

## Pattern B
- `&` BitwiseAnd
- `|` BitwiseOr
- `^` BitwiseXor

|L \ R|int|float|string|numeric-string|Stringable|array|
|:---------------|:-:|:-:|:-:|:-:|:-:|:-:|
|int                    |O|D|X|D|X|X|
|float                 |D|D|X|D|X|X|
|string               |X|X|O|O|X|X|
|numeric-string  |D|D|O|O|X|X|
|Stringable         |X|X|X|X|X|X|
|array                |X|X|X|X|X|X|

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -82,6 +82,7 @@ use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\ConstantTypeHelper;
 use PHPStan\Type\DynamicReturnTypeExtensionRegistry;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\FloatType;
 use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\Generic\GenericObjectType;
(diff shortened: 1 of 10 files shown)


ondrejmirtes (changes requested):
Hello, thank you for this PR, but it's too big and very ambitious. Also I disagree with some parts.

I'm not sure where to start with the review.

Can you please send a separate smaller PR that changes/fixes only a single operator, so we can learn from that, get on the same page, review that, merge that, and then do the same work for other operators? Thank you.

ryium:
Hi, I will first focus on the Mod operator and open the PR again. Thanks.

PR #2451, opened 2023-06-11
Retain list type via `array_push()` and `array_unshift()`

herndlm:
Closes https://github.com/phpstan/phpstan/issues/8449

it has been a while :) I know I have still 2 open PRs, but I didn't feel like finishing them yet. soon™

I was wondering why the array here is even re-created like that. The reason is that this is after the isIterableAtLeastOnce / optional check and the re-creation is just there to _remove_ a NonEmptyArrayType. So if there would be a simpler/better way of removing it (maybe calling `->popArray()` but I'm not sure, that feels like a hack), the list would not be removed and this new code would not be needed 🤔

another more explicit way might be to intersect the union of `->getArrays()` with a `TypeUtils::getAccessoryTypes()` where NonEmptyArrayType is filtered away. also feels ugly. I'm not a 100% happy with any of the solutions so far.. like e.g.
```php
$arrayType = TypeCombinator::intersect(
	TypeCombinator::union(...$arrayType->getArrays()),
	...array_filter(
		TypeUtils::getAccessoryTypes($arrayType),
		static fn (AccessoryType $type) => !$type instanceof NonEmptyArrayType,
	),
);
```

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2928,7 +2928,7 @@ class NodeScopeResolver
 					return;
 				}
 
-				$arrayType = new ArrayType($arrayType->getIterableKeyType(), $arrayType->getIterableValueType());
+				$arrayType = TypeCombinator::union($arrayType, new ConstantArrayType([], []));
 			},
 		);
 
--- a/tests/PHPStan/Analyser/data/array-push.php
+++ b/tests/PHPStan/Analyser/data/array-push.php
@@ -12,8 +12,9 @@ use function PHPStan\Testing\assertType;
  * @param int[] $b
  * @param non-empty-array<int> $c
  * @param array<int|string> $d
+ * @param list<string> $e
  */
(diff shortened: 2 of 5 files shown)


herndlm:
Ha, I just found that the array_slice return type extension is using `toArray()` for the same thing essentially! (removing non-empty-array). But IMO `toArray()` should NOT remove it, that's also a bug I believe. I'd like to fix that one, but we should agree on a way to remove non-empty-array first most likely. maybe even move that to the good old TypeUtils even then hmm

update: this all feels dirty, maybe new type methods for push, unshift and slice or so are needed after all, but that feels like overhead...

let me know what you prefer or if more info is needed please

herndlm:
rebased with the trick the boss came up with in https://github.com/phpstan/phpstan-src/commit/9e84671c475f2e1fded227965852ce9a32788484 to make the array non-non-empty again (why did I not think of this..)

now I'm happy with this :)

ondrejmirtes:
Thank you! :)

PR #2456, opened 2023-06-13
Add false to the db2_num_rows() return type

morozov:
See the [PHP documentation](https://www.php.net/manual/en/function.db2-num-rows.php).

Fixes https://github.com/phpstan/phpstan/issues/9429.


morozov:
@ondrejmirtes please let me know if any of the build failures are relevant. The ones about Doctrine DBAL are expected.

ondrejmirtes:
Thank you!

PR #2457, opened 2023-06-16
fix array_splice with non-array replacement

schlndh:
Fixes https://phpstan.org/r/f1506ef5-1320-4bc4-942c-c98c897bd024

[PHP documentation](https://www.php.net/manual/en/function.array-splice.php) says

> Note: If replacement is not an array, it will be typecast to one (i.e. (array) $replacement). This may result in unexpected behavior when using an object or null replacement.


ondrejmirtes:
Thank you!

PR #2458, opened 2023-06-16
Fix typos

szepeviktor:
Tool used: https://github.com/crate-ci/typos


src/Php/PhpVersion.php
```diff
@@ -51,7 +51,7 @@ public function deprecatesRequiredParameterAfterOptional(): bool
 		return $this->versionId >= 80000;
 	}
 
-	public function supportsLessOverridenParametersWithVariadic(): bool
+	public function supportsLessOverriddenParametersWithVariadic(): bool
```
mad-briller, line 54: this is a public api that extension authors can use so this would be a bc-break

szepeviktor: So I should revert it?

ondrejmirtes:
I'm interested only in some of the changes, not all of them. I definitely not want any changes in `tests/`.

Please send a PR for one file (or related change) at a time. Wait for a PR to be reviewed and merged, before sending the next one. Thanks for understanding.

PR #2459, opened 2023-06-16
Remove redundant conditions from RemoveUnusedCodeByPhpVersionIdVisitor.

mad-briller:
an `If_` cannot be a classlike or functionlike so checking for these specifically is redundant when we check it's not an `If_`


ondrejmirtes:
Thank you!

PR #2461, opened 2023-06-16
Fix classname typo in tests

szepeviktor:
From #2458

### DIFF
--- a/tests/PHPStan/Type/Generic/GenericObjectTypeTest.php
+++ b/tests/PHPStan/Type/Generic/GenericObjectTypeTest.php
@@ -63,7 +63,7 @@ class GenericObjectTypeTest extends PHPStanTestCase
 				TrinaryLogic::createYes(),
 			],
 			'implementation with @extends with different type args' => [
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Type/Generic/GenericObjectTypeTest.php
```diff
@@ -63,7 +63,7 @@ public function dataIsSuperTypeOf(): array
 				TrinaryLogic::createYes(),
 			],
 			'implementation with @extends with different type args' => [
-				new GenericObjectType(B\I::class, [new ObjectType('DateTimeInteface')]),
+				new GenericObjectType(B\I::class, [new ObjectType('DateTimeInterface')]),
```
staabm, line 66: Would `DateTimeInterface::class` work?

szepeviktor: I think yes.
But I only do what I'm told.

ondrejmirtes:
Thank you.

szepeviktor:
You're welcome.

Advancing with the next one 🏃🏻‍♂️

PR #2462, opened 2023-06-17
Fix typo: removeable

szepeviktor:
From #2458


ondrejmirtes:
Sorry, I don't want this change. Both spellings are valid.

PR #2464, opened 2023-06-17
Treat ConstantArrayType as covariant in its keys and values

jiripudil:
Closes phpstan/phpstan#9161, as argued there, this should be safe because arrays are copy-on-write


ondrejmirtes:
Thank you!

PR #2465, opened 2023-06-17
Treat ArrayType as covariant in its keys and values

jiripudil:
Following up on #2464, I believe the implementation in ArrayType can be simplified accordingly


ondrejmirtes:
Is there a situation where the behaviour is different? Would be nice to have a test that fails without this change :)

jiripudil:
Actually, there is. Added a test. It looks like edge cases, but I guess it _could_ happen that this change introduces new errors for someone 😬

ondrejmirtes:
I just checked the last PR that touched this: https://github.com/phpstan/phpstan-src/pull/30/files

And looks like all the tests that were added back then still pass :)

Thank you!

PR #2466, opened 2023-06-17
Catch implicit conversions in mod operator

ryium:
Catch deprecated implicit type conversions that occur in PHP 8.1 and later.
The following BinaryOp that may cause deprecation.

This PR is part of https://github.com/phpstan/phpstan/issues/8288 and change regarding the mod operator of https://github.com/phpstan/phpstan-src/pull/2450

※O = safe, D = deprecated, X = not safe

## `%` Mod

|L \ R|int|float|string|numeric-string|Stringable|array|
|:---------------|:-:|:-:|:-:|:-:|:-:|:-:|
|int                    |O|D|X|D|X|X|
|float                 |D|D|X|D|X|X|
|string               |X|X|X|X|X|X|
|numeric-string  |D|D|X|D|X|X|
|Stringable         |X|X|X|X|X|X|
|array                |X|X|X|X|X|X|

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -874,6 +874,14 @@ class InitializerExprTypeResolver
 			}
 		}
 
+		$leftNumberType = $leftType->toNumber();
+		$rightNumberType = $rightType->toNumber();
+		if ($leftNumberType instanceof ErrorType || $rightNumberType instanceof ErrorType) {
+			return new ErrorType();
+		} elseif ($leftNumberType->isInteger()->or($leftNumberType->isFloat())->no() && $rightNumberType->isInteger()->or($rightNumberType->isFloat())->no()) {
+			return new ErrorType();
+		}
+
 		$positiveInt = IntegerRangeType::fromInterval(0, null);
 		if ($rightType->isInteger()->yes()) {
 			$rangeMin = null;
--- a/src/Rules/Operators/InvalidBinaryOperationRule.php
+++ b/src/Rules/Operators/InvalidBinaryOperationRule.php
@@ -6,6 +6,7 @@ use PhpParser\Node;
 use PHPStan\Analyser\MutatingScope;
 use PHPStan\Analyser\Scope;
 use PHPStan\Node\Printer\ExprPrinter;
+use PHPStan\Php\PhpVersion;
 use PHPStan\Rules\Rule;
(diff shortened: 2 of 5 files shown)


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -872,7 +877,18 @@ public function getModType(Expr $left, Expr $right, callable $getTypeCallback):
 			}
 		}
 
+		$float = new FloatType();
+		if ($leftType instanceof UnionType && $leftType->isSuperTypeOf($float)->yes()) {
+			return new ErrorType();
```
ondrejmirtes, line 880: How the type system responds shouldn't change because it's not actually true. The `%` operator still returns a number but only on some PHP versions it throws a deprecation notice.

src/Rules/Operators/InvalidBinaryOperationRule.php
```diff
@@ -105,6 +106,25 @@ public function processNode(Node $node, Scope $scope): array
 				return [];
 			}
 
+			if ($node instanceof Node\Expr\BinaryOp\Mod) {
+				$leftNumberType = $leftType->toNumber();
+				$rightNumberType = $rightType->toNumber();
+				if (!$leftNumberType instanceof ErrorType && !$rightNumberType instanceof ErrorType) {
```
ondrejmirtes: So just change the rule please. The condition should be written differently.

1) Please verify that all the `X` from table here https://github.com/phpstan/phpstan-src/pull/2466 are reported with `'Binary operation "%s" between %s and %s results in an error.'`.
2) Please write tests that all the `D` are reported with `'Deprecated in PHP 8.1: Implicit conversion from %s to %s loses precision.'`.
3) Please note that types can for example be `int|float` or `mixed`. You should be able to take advantage of RuleLevelHelper so that something that's always wrong like `int % float` is always reported, something that might be wrong like `int % float|int` is reported on level 7 and `int % mixed` is reported on level 9.

Please see how `RuleLevelHelper::findTypeToCheck` is used in many existing rules like InvalidCastRule. Basically there's a callback that checks the type and transforms the type based on the rule level. Then there's `$type instanceof ErrorType` with early return and then there's the callback called again on the transformed type.

ondrejmirtes (changes requested):

ondrejmirtes (changes requested):
I like this much more! 👍 

But the rule should report this only on 8.1+. Please add a new method in PhpVersion and ask about it in the rule.

PR #2474, opened 2023-06-20
Add return type extension for `constant()`

herndlm:
Had that idea when https://github.com/phpstan/phpstan/issues/9295 was created. 

See https://www.php.net/manual/en/function.constant.php and https://3v4l.org/66no7

UPDATE: prior to PHP 8 it would trigger a warning for inexistent constants and return `null`. But not sure if it's worth the effort to implement this. right now it would just return `*ERROR*` for such cases. UPDATE2: this feels related to also support non-existing constants as expressions, so I'll refrain from adding any code for it for now. E.g. https://3v4l.org/FTHPv / https://phpstan.org/r/c649b7bb-8031-4210-9c4b-c469eba63d3b

### DIFF
--- a/tests/PHPStan/Analyser/data/constant.php
+++ b/tests/PHPStan/Analyser/data/constant.php
@@ -1,0 +1,40 @@
+<?php // lint >= 8.1
+
+namespace Constant;
+
+use function PHPStan\Testing\assertType;
+
+define('FOO', 'foo');
+const BAR = 'bar';
+
+class Baz
+{
+	const BAZ = 'baz';
+}
+
+enum Suit
+{
+    case Hearts;
+}
+
+function doFoo(string $constantName): void
+{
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/data/constant.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    case Hearts;
+}
+
+function doFoo(string $constantName): void
+{
+	assertType('mixed', constant($constantName));
+}
+
+assertType("'foo'", FOO);
```
staabm, line 25: Can the asserts be moved into a  function or method?

herndlm: sure, but not sure if they really hurt here :)

ondrejmirtes:
Thank you!

PR #2476, opened 2023-06-21
More precise file() flags args in bleedingEdge

staabm:
https://www.php.net/manual/en/function.file.php

refs https://github.com/php/php-src/pull/11483

### DIFF
--- a/resources/functionMap_bleedingEdge.php
+++ b/resources/functionMap_bleedingEdge.php
@@ -107,6 +107,7 @@ return [
 		'ImagickKernel::scale' => ['void', 'scale'=>'float', 'normalizeFlag'=>'Imagick::NORMALIZE_KERNEL_*'],
 		'max' => ['', '...arg1'=>'non-empty-array'],
 		'min' => ['', '...arg1'=>'non-empty-array'],
+		'file' => ['list<string>|false', 'filename'=>'string', 'flags='=>'FILE_USE_INCLUDE_PATH|FILE_IGNORE_NEW_LINES|FILE_SKIP_EMPTY_LINES', 'context='=>'resource'],
 	],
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you.

resources/functionMap_bleedingEdge.php
```diff
@@ -107,6 +107,7 @@
 		'ImagickKernel::scale' => ['void', 'scale'=>'float', 'normalizeFlag'=>'Imagick::NORMALIZE_KERNEL_*'],
 		'max' => ['', '...arg1'=>'non-empty-array'],
 		'min' => ['', '...arg1'=>'non-empty-array'],
+		'file' => ['list<string>|false', 'filename'=>'string', 'flags='=>'0|FILE_USE_INCLUDE_PATH|FILE_IGNORE_NEW_LINES|FILE_SKIP_EMPTY_LINES', 'context='=>'resource'],
```
iluuu1994, line 110: This seems to be missing the `FILE_NO_DEFAULT_CONTEXT` flag (which is apparently undocumented).

staabm: thanks will have a look

staabm: added via https://github.com/phpstan/phpstan-src/pull/2482

I missed it, because its missing from https://www.php.net/manual/en/function.file.php - therefore just also [sumbitted a php.net docs PR](https://github.com/php/doc-en/pull/2540)

ondrejmirtes:
@staabm I don't think this should be reported: https://phpstan.org/r/7cdbe87f-3cd6-4ba7-a1d0-7e2be337d4ed Can you please look into it? Thank you.

PR #2477, opened 2023-06-22
Bleeding edge - More precise flock() operation flags

staabm:


tests/PHPStan/Rules/Functions/data/flock.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace flockFlags;
+
+class flock {
+	/** @param resource $fp */
+	public function ok($fp):void {
```
staabm, line 7: code examples from https://www.php.net/manual/en/function.flock

resources/functionMap_bleedingEdge.php
```diff
@@ -108,6 +108,7 @@
 		'max' => ['', '...arg1'=>'non-empty-array'],
 		'min' => ['', '...arg1'=>'non-empty-array'],
 		'file' => ['list<string>|false', 'filename'=>'string', 'flags='=>'0|FILE_USE_INCLUDE_PATH|FILE_IGNORE_NEW_LINES|FILE_SKIP_EMPTY_LINES', 'context='=>'resource'],
+		'flock' => ['bool', 'fp'=>'resource', 'operation'=>'int<1, 7>', '&w_wouldblock='=>'int'],
```
ondrejmirtes: Should probably be integer mask instead https://phpstan.org/writing-php-code/phpdoc-types#integer-masks

ondrejmirtes (changes requested):

resources/functionMap_bleedingEdge.php
```diff
@@ -108,6 +108,7 @@
 		'max' => ['', '...arg1'=>'non-empty-array'],
 		'min' => ['', '...arg1'=>'non-empty-array'],
 		'file' => ['list<string>|false', 'filename'=>'string', 'flags='=>'0|FILE_USE_INCLUDE_PATH|FILE_IGNORE_NEW_LINES|FILE_SKIP_EMPTY_LINES', 'context='=>'resource'],
+		'flock' => ['bool', 'fp'=>'resource', 'operation'=>'int-mask<1, 2, 4>', '&w_wouldblock='=>'int'],
```
staabm: semantically we can be more precise if we want. as not all combinations of the constants are allowed.
`0` is not allowed.
combinations of e.g. `LOCK_EX` and `LOCK_SH` are not valid

not sure its worth it though

ondrejmirtes: Can you use the constants instead of numbers?

staabm: it seems we do not yet support constants in `int-mask`:

https://phpstan.org/r/1a8e768d-0fd3-4390-8041-4f10ff639c0b

staabm: ohh my bad: it works

https://phpstan.org/r/e6a94130-e9da-4de0-b1ee-4eab8d73bc80

ondrejmirtes:
Thank you.

PR #2478, opened 2023-06-22
Break in ClassReflection when not allowing dynamic property extensions.

mad-briller:
break is more appropriate here as the condition will never become false after the first time it is true

additionally `allowsDynamicPropertyExtensions()` is a non-trivial method that is not memoized, so lets avoid calling it more than we have to given it's value will not change between iterations


ondrejmirtes:
Thank you!

PR #756, opened 2021-11-05
implemented `call_user_func` & `call_user_func_array` dynamic return type extensions

staabm:
closes https://github.com/phpstan/phpstan/issues/5942

PR #2479, opened 2023-06-22
Implement call_user_func() DynamicFunctionReturnTypeExtension

staabm:
submitting the non-controversial parts of https://github.com/phpstan/phpstan-src/pull/756 as a separate PR

refs https://github.com/phpstan/phpstan/issues/5942

closes https://github.com/phpstan/phpstan/issues/7057


src/Type/Php/CallUserFuncDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,32 @@
...
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return $functionReflection->getName() === 'call_user_func';
+	}
+
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
+	{
+		$args = $functionCall->getArgs();
+
+		$callbackType = $scope->getType($args[0]->value);
```
ondrejmirtes: This will crash for `call_user_func();`

src/Type/Php/CallUserFuncDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,32 @@
...
+		if ($callbackType->isCallable()->yes()) {
+			return $scope->getType(new FuncCall($args[0]->value, array_slice($args, 1)));
```
ondrejmirtes: I'd say this is a bit naive about named arguments. See https://wiki.php.net/rfc/named_params - section "call_user_func() and friends".

 Most likely we'll have to modify ArgumentsNormalizer as well.

ondrejmirtes (changes requested):

src/Analyser/ArgumentsNormalizer.php
```diff
@@ -23,12 +23,60 @@ final class ArgumentsNormalizer
...
+		FuncCall $callUserFuncCall,
+		Scope $scope,
+	): ?FuncCall
+	{
+		$args = $callUserFuncCall->getArgs();
+		if (count($args) < 1) {
+			return null;
+		}
+
+		$calledOnType = $scope->getType($args[0]->value);
```
ondrejmirtes: Still wrong. See: https://3v4l.org/F67Rj

staabm: added a test, which shows it works If I got your point.

at this point of execution named args are already re-ordered to positional ones

ondrejmirtes: Your logic isn't correct because I don't see the handling of `call_user_func` that would be necessary for PHPStan to understand it.

I added a rule to prove my point. You can see the rule doesn't see the signature correctly and so can't the dynamic return type extension.

The problem is that MutatingScope (before it calls your extension) already calls ArgumentsNormalizer which mangles the arguments so they're wrong. Because it doesn't know what to do with them.

staabm: Thank you. I have an idea now what you are talking about.

Do you expect this PR to still contain this new rule when finished, or do you think I should/can adjust the core in a way, that the rewritten FuncCall will be finally checked by the pre-existing `CallToFunctionParametersRule` ?

ondrejmirtes: I've got an idea:

1) It should not be an extension, just a piece of code in MutatingScope before extensions are called. Before the normalizer is used.
2) That way we can find the first/callback arg and remove it from args. And create a new FuncCall with the rest.
2) We can call $this->getType on the FuncCall and return that.
4) The new rule should just do the same thing as in 2) and pass the new FuncCall to FunctionCallParametersCheck.

tests/PHPStan/Rules/Functions/data/call-user-func.php
```diff
@@ -0,0 +1,25 @@
...
+
+		};
+		call_user_func($f);
+		call_user_func($f, 1);
+		call_user_func($f, 'foo');
+		call_user_func($f, i: 'foo');
+		call_user_func(i: 'foo', callback: $f);
+		call_user_func($f, i: 1);
+		call_user_func(i: 1, callback: $f);
+		call_user_func($f, j: 1);
```
staabm, line 22: this line cannot be re-ordered by `ArgumentsNormalizer::reorderCallUserFuncArguments` because the param names cannot be matched with the parameter acceptor.

therefore the "plain" FuncCall is validated instead and thats the reason why the error for this line is reported by `CallToFunctionParametersRule` instead of `CallUserFuncRule`

staabm:
one jobs CI errors we see is

```
Error: Parameter #1 $persisted of callable passed to call_user_func() expects array<E of Nextras\Orm\Entity\IEntity>, array<Nextras\Orm\Entity\IEntity> given.
Error: Parameter #2 $removed of callable passed to call_user_func() expects array<E of Nextras\Orm\Entity\IEntity>, array<Nextras\Orm\Entity\IEntity> given.
 ------ ---------------------------------------------------------------------- 
  Line   src/Repository/Repository.php                                         
 ------ ---------------------------------------------------------------------- 
  621    Parameter #1 $persisted of callable passed to call_user_func()        
         expects array<E of Nextras\Orm\Entity\IEntity>,                       
         array<Nextras\Orm\Entity\IEntity> given.                              
         💡 Type Nextras\Orm\Entity\IEntity is not always the same as E. It    
            breaks the contract for some argument types, typically subtypes.   
  621    Parameter #2 $removed of callable passed to call_user_func() expects  
         array<E of Nextras\Orm\Entity\IEntity>,                               
         array<Nextras\Orm\Entity\IEntity> given.                              
         💡 Type Nextras\Orm\Entity\IEntity is not always the same as E. It    
            breaks the contract for some argument types, typically subtypes.   
 ------ ---------------------------------------------------------------------- 
 ```
 
 I cannot judge whether this is something expected or not.

src/Analyser/ArgumentsNormalizer.php
staabm: thanks. 

I had to re-calibrate my expectations for this PR, as it turned from a return-type improving into a more overall goal :).
it seems to work as you described now.

ondrejmirtes:
Pushed some changes, feel free to check them out. Now please contribute the same thing for `call_user_func_array` :) Thank you.

staabm:
Thank you

PR #2480, opened 2023-06-22
Revert calling allowsDynamicPropertiesExtensions ahead of time.

mad-briller:
reverts part of the change i made before :see_no_evil: sorry for the bad commit, i should've measured more before making the pr

upon further investigation i realized that this actually causes the method to be called way more often than it needs to be, as the majority of cases of property access are managed by the first property reflection extension and as such this wouldn't be called...

keeping the `continue` -> `break` change as that is still useful however


ondrejmirtes:
Thank you!

PR #2481, opened 2023-06-23
Type projections, part 2: enforce elemental rules

jiripudil:
This PR should be straightforward. It prohibits type projections in incompatible places (e.g. `@extends`) and checks that the call-site variance is not in conflict with the declaration-site variance.


src/Rules/Generics/GenericAncestorsCheck.php
```diff
@@ -106,6 +110,18 @@ public function check(
...
+			foreach ($ancestorType->getVariances() as $index => $typeVariance) {
+				if ($typeVariance->invariant()) {
+					continue;
+				}
+
+				$messages[] = RuleErrorBuilder::message(sprintf(
+					$typeProjectionIsNotAllowedMessage,
+					TypeProjectionHelper::describe($ancestorType->getTypes()[$index], $typeVariance, VerbosityLevel::typeOnly()),
+					$ancestorType->describe(VerbosityLevel::typeOnly()),
+				))->identifier('generics.typeProjectionNotAllowed')->build();
```
jiripudil: Now that you want this in `1.10.x`, is adding an identifier here ok, or should I remove it and add it afterwards only to `1.11.x`?

src/Rules/Generics/GenericObjectTypeCheck.php
```diff
@@ -84,8 +89,35 @@ public function check(
...
 
 				$templateType = $templateTypes[$i];
-				$boundType = TemplateTypeHelper::resolveToBounds($templateType);
 				$genericTypeType = $genericTypeTypes[$i];
+
+				$genericTypeVariance = $genericTypeVariances[$i] ?? TemplateTypeVariance::createInvariant();
+				if ($templateType instanceof TemplateType && !$genericTypeVariance->invariant()) {
+					if ($genericTypeVariance->equals($templateType->getVariance())) {
+						$messages[] = RuleErrorBuilder::message(sprintf(
+							$typeProjectionIsRedundantMessage,
```
jiripudil, line 98: This is not technically a problem. I feel like it should be just a warning, but afaik PHPStan doesn't know that concept, so I've included it here along with a tip on how to fix it

ondrejmirtes:
Before we continue, I'd also like you to think about documentation. Right now there are these pages about generics:

* https://phpstan.org/blog/generics-in-php-using-phpdocs (doesn't mention declare-site variance at all, and it's already long, maybe we don't need to touch it at all)
* https://phpstan.org/blog/generics-by-examples - practical examples, because some people haven't been able to understand generics at all after reading the first article. Here some examples about variance should be added
* https://phpstan.org/blog/whats-up-with-template-covariant - this one is important, it's linked in an error tip when a type isn't accepted because of variance. It should definitely be updated with call-site variance

And also maybe we should write a completely new article explaining this whole issue and new posibilities with call-site variance.

The source code for all of this is in https://github.com/phpstan/phpstan/tree/1.11.x/website/src so feel free to send PRs there too :) Thank you.

I'll review your existing PRs as soon as possible but it might be a bit challenging, I'm travelling for the next two weeks but hopefully I'll find some time :)

jiripudil:
> Before we continue, I'd also like you to think about documentation

Yep, I'm thinking about that, and was mainly thinking along the lines of "maybe we should write a completely new article explaining this whole issue and new posibilities with call-site variance" :)

> I'll review your existing PRs as soon as possible but it might be a bit challenging, I'm travelling for the next two weeks but hopefully I'll find some time :)

Absolutely no worries, I'll actually be off (and possibly offline) too for some time within the next two weeks

src/Rules/Generics/GenericAncestorsCheck.php
ondrejmirtes: This is okay, I might change the identifier though :)

tests/PHPStan/Rules/Classes/MixinRuleTest.php
```diff
@@ -84,6 +84,15 @@ public function testRule(): void
 				'PHPDoc tag @mixin contains non-object type int.',
 				92,
 			],
+			[
+				'Type projection contravariant MixinRule\Foo in generic type MixinRule\Adipiscing<contravariant MixinRule\Foo> in PHPDoc tag @mixin is conflicting with variance of template type T of class MixinRule\Adipiscing.',
```
ondrejmirtes: Two things:

1) Should we call this "type projection"? What exactly is the projection? Is `MixinRule\Adipiscing<Foo>` (the common syntax we use today) also some kind of projection? I like "call-site variance" much better as opposed to "declaration variance".
2) The second part of the message "conflicting with variance of template type T" could specify the exact variance, so it could say:

* `"conflicting with covariance of template type T"`
* `"conflicting with invariance of template type T"`
* `"conflicting with contravariance of template type T"`

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Classes/MixinRuleTest.php
jiripudil: Good point, I've reworded the messages. Different sources use the term "type projection" rather vaguely, so it's for the best to use "call-site variance", hopefully it should clear any confusion about what the error refers to.

ondrejmirtes:
Thank you! I believe we should do 3 more things:

1) NonAcceptingNeverType PR :)
2) Rebase https://github.com/phpstan/phpstan-src/pull/2485
3) Adjustments to documentation + new article :)

Thank you very much.

PR #2484, opened 2023-06-23
Use case insensitive matching to detect usages of private methods

stof:
Closes https://github.com/phpstan/phpstan/issues/8346


ondrejmirtes:
Thank you!

PR #2488, opened 2023-06-26
Bleeding Edge - More precise json_decode parameter types

staabm:


tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
```diff
@@ -1388,4 +1388,22 @@ public function testFlockParams(): void
 		]);
 	}
 
+	public function testJsonEncodeDecodeParams(): void
+	{
+		$this->analyse([__DIR__ . '/data/json_encode_decode.php'], [
```
staabm: I named this test encode/decode because initially I had both cases covered here.

turns out though, that because of json_encode() accepts so many JSON flags, we should do this in a separate followup PR, because DX suffers from the huge int-mask generated from it.

lets discuss it in the followup then

ondrejmirtes: Right now it's just "decode" so name it decode please.

resources/functionMap_php80delta_bleedingEdge.php
```diff
@@ -5,6 +5,7 @@
 		'error_log' => ['bool', 'message'=>'string', 'message_type='=>'0|1|3|4', 'destination='=>'string', 'extra_headers='=>'string'],
 		'filter_input' => ['mixed', 'type'=>'INPUT_GET|INPUT_POST|INPUT_COOKIE|INPUT_SERVER|INPUT_ENV', 'variable_name'=>'string', 'filter='=>'int', 'options='=>'array|int'],
 		'filter_input_array' => ['array|false|null', 'type'=>'INPUT_GET|INPUT_POST|INPUT_COOKIE|INPUT_SERVER|INPUT_ENV', 'definition='=>'int|array', 'add_empty='=>'bool'],
+		'json_decode' => ['mixed', 'json'=>'string', 'assoc='=>'bool|null', 'depth='=>'positive-int', 'options='=>'int-mask<JSON_BIGINT_AS_STRING|JSON_INVALID_UTF8_IGNORE|JSON_INVALID_UTF8_SUBSTITUTE|JSON_OBJECT_AS_ARRAY|JSON_THROW_ON_ERROR>'],
```
staabm: the build fails on PHP 7.2 when running with PHP_VERSION_ID = 8.0, because `JSON_THROW_ON_ERROR` is not defined on 7.2.. the constant was added with 7.3

staabm: hmm the actual question should be, why the defined constant in phpstorm stubs is not picked up.

https://github.com/JetBrains/phpstorm-stubs/blob/e2e898a29929d2f520fe95bdb2109d8fa895ba4a/json/json.php#L438-L445

PR #2489, opened 2023-06-26
More precise ignore_user_abort() return type

staabm:
as [we recently stumbled over this functions ](https://github.com/sabre-io/http/pull/207/files#r1241885349) return type I figured it would be helpful to cover this more precisely in PHPStan.

The setter expects a bool, but the getter returns a int.

https://3v4l.org/jmCa8
https://www.php.net/manual/en/function.ignore-user-abort.php


ondrejmirtes:
Thank you!

PR #2491, opened 2023-06-27
Implement JsonValidateTypeSpecifyingExtension

staabm:
PHP8.3 Support for https://wiki.php.net/rfc/json_validate

this type-specifying extension narrows the `$json` arg to a non-empty-string

### DIFF
--- a/resources/functionMap_php83delta.php
+++ b/resources/functionMap_php83delta.php
@@ -1,0 +1,29 @@
+<?php // phpcs:ignoreFile
+
+/**
+ * Copied over from https://github.com/phan/phan/blob/8866d6b98be94b37996390da226e8c4befea29aa/src/Phan/Language/Internal/FunctionSignatureMap_php80_delta.php
+ * Copyright (c) 2015 Rasmus Lerdorf
+ * Copyright (c) 2015 Andrew Morrison
+ */
+
+/**
+ * This contains the information needed to convert the function signatures for php 8.0 to php 7.4 (and vice versa)
+ *
+ * This has two sections.
+ * The 'new' section contains function/method names from FunctionSignatureMap (And alternates, if applicable) that do not exist in php7.4 or have different signatures in php 8.0.
+ *   If they were just updated, the function/method will be present in the 'added' signatures.
+ * The 'old' signatures contains the signatures that are different in php 7.4.
+ *   Functions are expected to be removed only in major releases of php.
+ *
+ * @see FunctionSignatureMap.php
+ *
+ * @phan-file-suppress PhanPluginMixedKeyNoKey (read by Phan when analyzing this file)
+ */
+return [
+	'new' => [
+		'json_validate' => ['bool', 'json'=>'string', 'depth='=>'positive-int', 'flags='=>'int-mask<JSON_INVALID_UTF8_IGNORE>'],
+	],
+	'old' => [
+
+	]
+];
--- a/src/Type/Php/JsonValidateTypeSpecifyingExtension.php
+++ b/src/Type/Php/JsonValidateTypeSpecifyingExtension.php
@@ -9,7 +9,7 @@ use PHPStan\Analyser\TypeSpecifier;
 use PHPStan\Analyser\TypeSpecifierAwareExtension;
 use PHPStan\Analyser\TypeSpecifierContext;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
+use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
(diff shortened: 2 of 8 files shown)


src/Type/Php/JsonValidateTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,42 @@
...
+use PHPStan\Analyser\SpecifiedTypes;
+use PHPStan\Analyser\TypeSpecifier;
+use PHPStan\Analyser\TypeSpecifierAwareExtension;
+use PHPStan\Analyser\TypeSpecifierContext;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
+use PHPStan\Type\FunctionTypeSpecifyingExtension;
+use function count;
+
+class JsonValidateTypeSpecifyingExtension implements FunctionTypeSpecifyingExtension, TypeSpecifierAwareExtension
```
staabm, line 16: I tried using a stub signature instead, but this did not work for me locally:

```
/**
 * @phpstan-assert-if-true non-falsy-string $json
 * @return bool
 */
function json_validate(string $json, int $depth = 512, int $flags = 0): bool {}
```

as I can see this extension also only work in CI PHP 8.3.. I will retry using a stub signature on a different composer which has PHP 8.3 installed

staabm: ok, did a few more tests. It does not work with pure stub signatures, because these would also affect the ELSE case in a wrong way:

```
1) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/json_validate.php:13" ('type', '/Users/staabm/workspace/phpst...te.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\MixedType Object (...), 13)
Expected type mixed, got type mixed~non-falsy-string in /Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/json_validate.php on line 13.
--- Expected
+++ Actual
@@ @@
-'mixed'
+'mixed~non-falsy-string'
```

resources/functionMap_php83delta.php
```diff
@@ -0,0 +1,29 @@
...
+ * The 'old' signatures contains the signatures that are different in php 7.4.
+ *   Functions are expected to be removed only in major releases of php.
+ *
+ * @see FunctionSignatureMap.php
+ *
+ * @phan-file-suppress PhanPluginMixedKeyNoKey (read by Phan when analyzing this file)
+ */
+return [
+	'new' => [
+		'json_validate' => ['bool', 'json'=>'string', 'depth='=>'positive-int', 'flags='=>'int-mask<JSON_INVALID_UTF8_IGNORE>'],
```
staabm, line 24: only a single flag is supported, see https://github.com/php/php-src/pull/9399#issuecomment-1609741719

src/Type/Php/JsonValidateTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,42 @@
...
+
+	public function specifyTypes(FunctionReflection $functionReflection, FuncCall $node, Scope $scope, TypeSpecifierContext $context): SpecifiedTypes
+	{
+		$args = $node->getArgs();
+
+		if (count($args) < 1) {
+			return new SpecifiedTypes([], []);
+		}
+
+		return $this->typeSpecifier->create($node->getArgs()[0]->value, new AccessoryNonFalsyStringType(), $context, false, $scope);
```
mad-briller: is non falsy string type correct here?

https://3v4l.org/5IGU2/rfc#vgit.master

`'0'` is a valid json string (according to php anyway) but is also falsey

staabm: You are right. I did a 3v4l before with this exact case in mind. Seems I fucked it up. 

Need to go with non-empty-string then :-)

ondrejmirtes:
Hi, I usually start preparing next PHP version support after feature freeze but I decided to make an exception here.

Here's what I did:

* https://github.com/phpstan/php-8-stubs/commit/56ccadcf48b657bb7519d018dfd1927f468062ef (which caused https://github.com/phpstan/php-8-stubs/commit/2802b16850ea45c846c8ec0e82a3cdf6f542da32 to be generated)
* A stub: https://github.com/phpstan/phpstan-src/commit/828b269c0dccda404e33a5cca1ac4be1998e18ba (you don't need an extension). I took your tests there, thank you :)

staabm:
thank you

PR #2492, opened 2023-06-27
Fix int-mask in file()

staabm:
refs https://github.com/phpstan/phpstan-src/pull/2476#issuecomment-1609969911

https://phpstan.org/r/7cdbe87f-3cd6-4ba7-a1d0-7e2be337d4ed


ondrejmirtes:
Thank you.

PR #2494, opened 2023-06-28
feat: treat void as null for null coalescing operator

kesselb:
Hi :wave: 

WordPress often uses type definitions like `array|void` (e.g. https://github.com/WordPress/wordpress-develop/blob/68be569925ad2a88c97e1308b8a67a57bf17a14f/src/wp-includes/post.php#L2627).

`$custom_keys = get_post_custom_keys( $post_id ) ?? [];`

Shows: Expression on left side of ?? is not nullable.

### DIFF
--- a/src/Rules/Variables/NullCoalesceRule.php
+++ b/src/Rules/Variables/NullCoalesceRule.php
@@ -8,6 +8,7 @@ use PHPStan\Rules\IssetCheck;
 use PHPStan\Rules\Rule;
 use PHPStan\Type\NullType;
 use PHPStan\Type\Type;
+use PHPStan\Type\VoidType;
 
 /**
  * @implements Rule<Node\Expr>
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Using `void` in a union is completely invalid and when used in a native type even crashes PHP: https://3v4l.org/7iULB

Instead of `array|void` instead `array|null` should be used.

That said, I'd accept a PR that changes `array|void` into `array|null` as a result of FuncCall/StaticCall/MethodCall. So the change needs to happen in `MutatingScope::resolveType`. It's not enough to change a single rule.

kesselb:
Thanks for your prompt review :+1: 

I will take a look.

herndlm:
Looks like you are aware of it already @kesselb , but just for the sake of completeness and issue linking - we discussed this also in https://github.com/szepeviktor/phpstan-wordpress/issues/176, and yeah, WordPress types are wrong there IMO

Oh and https://github.com/phpstan/phpstan/issues/9185 too

mvorisek:
> That said, I'd accept a PR that changes `array|void` into `array|null` as a result of FuncCall/StaticCall/MethodCall. So the change needs to happen in `MutatingScope::resolveType`. It's not enough to change a single rule.

Yes, this is the right way, testcase from https://github.com/phpstan/phpstan/issues/9388 should be added.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/pull/2778

PR #2496, opened 2023-06-29
Added regression test

staabm:
phpstanbot told us this issue is fixed: https://github.com/phpstan/phpstan/issues/4608#issuecomment-1235674970
just locally confirmed that this no longer errors with 

> Parameter #1 $callback of function call_user_func expects callable(): mixed, array(class@anonymous/tmp.php:3, 'abc'|'not_abc') given.

refs https://github.com/phpstan/phpstan/issues/4608


staabm:
hmm ok.. I think the expected situation is to get 3 errors, instead of none

PR #2498, opened 2023-06-30
Fix ClassConstFetch Identical specification with Yoda conditions

herndlm:
Closes https://github.com/phpstan/phpstan/issues/9542

Is this how you would have done it? Might be the simplest way I guess. It's just an adapted version to handle the left/right mixed variant of the block directly before it.

UPDATE: I had the urge to somehow simplify/cleanup things, like the way left and right types are fetched from the scope, assigned to variables and used, in Identical in a follow-up but I failed. Better not touch it too much 😅

### DIFF
--- a/tests/PHPStan/Rules/Methods/data/bug-9542.php
+++ b/tests/PHPStan/Rules/Methods/data/bug-9542.php
@@ -33,7 +33,7 @@ class HelloWorld
 	public function testClass(TranslatableInterface $translatable): void
 	{
 		if ($translatable::class !== TranslatableMessage::class) {
-			assertType('Bug9542\TranslatableInterface~Bug9542\TranslatableMessage', $translatable);
+			assertType('Bug9542\TranslatableInterface', $translatable);
 			return;
 		}
 
@@ -44,7 +44,7 @@ class HelloWorld
 	public function testClassReverse(TranslatableInterface $translatable): void
 	{
 		if (TranslatableMessage::class !== $translatable::class) {
-			assertType('Bug9542\TranslatableInterface~Bug9542\TranslatableMessage', $translatable);
+			assertType('Bug9542\TranslatableInterface', $translatable);
 			return;
 		}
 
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
Hi, I'd be happier with NodeScopeResolverTest. We can leave the rule test but it'd be best to check the type inference (truthy/falsy context too). Thanks!

herndlm:
sure. added it.
UPDATE: missed the falsy context assertions. addem them too now..

tests/PHPStan/Rules/Methods/data/bug-9542.php
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		assertType('Bug9542\TranslatableMessage', $translatable);
+		$translatable->getMessage();
+	}
+
+	public function testClass(TranslatableInterface $translatable): void
+	{
+		if ($translatable::class !== TranslatableMessage::class) {
+			assertType('Bug9542\TranslatableInterface~Bug9542\TranslatableMessage', $translatable);
```
ondrejmirtes: There's a bad bug here. So you've eliminated that `$translatable` isn't `TranslatableMessage`. But it can still be a subclass of `TranslatableMessage`.

Most likely we should only specify the type for `$context->true()`.

Same thing is done for `get_class`: https://github.com/phpstan/phpstan-src/blob/2df1a5904a218038187e180861f757bb938a37b4/src/Analyser/TypeSpecifier.php#L194-L210

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Methods/data/bug-9542.php
staabm: could we be "smarter" if `TranslatableMessage` is a reference to a final class ~~(or a interface)~~?

ondrejmirtes: We could, but first we need to get rid of the bug, then we can improve it :)

herndlm: adapted. better? I didn't even notice this.. :)

herndlm: oh sorry, I was lazy and didn't run the full tests :/

herndlm: thx for the fix!

PR #2500, opened 2023-06-30
Add get_declared_classes() to runtime reflection function check.

mad-briller:


ondrejmirtes:
I feel like in this case we should give more advice about what to do instead. This deserves an article about the problem.

mad-briller:
unfortunately i don't think i'm qualified enough to write an article about the subject, as i'm not sure why people may be motivated to use this function and what to say / recommend in it's place

do you want me to leave this pr open incase you write up an article ?

ondrejmirtes:
> do you want me to leave this pr open

Yes, thank you 😊

PR #2501, opened 2023-07-01
Implement `call_user_func_array()` DynamicFunctionReturnTypeExtension and rule

staabm:
refs https://github.com/phpstan/phpstan-src/pull/2479#issuecomment-1613755928

closes https://github.com/phpstan/phpstan/issues/5942

### DIFF
--- a/conf/config.level5.neon
+++ b/conf/config.level5.neon
@@ -10,6 +10,8 @@ conditionalTags:
 		phpstan.rules.rule: %featureToggles.arrayFilter%
 	PHPStan\Rules\Functions\CallUserFuncRule:
 		phpstan.rules.rule: %featureToggles.callUserFunc%
+	PHPStan\Rules\Functions\CallUserFuncArrayRule:
+			phpstan.rules.rule: %featureToggles.callUserFunc%
 
 rules:
 	- PHPStan\Rules\DateTimeInstantiationRule
@@ -30,3 +32,6 @@ services:
 
 	-
 		class: PHPStan\Rules\Functions\CallUserFuncRule
+
+	-
+		class: PHPStan\Rules\Functions\CallUserFuncArrayRule
--- a/src/Analyser/ArgumentsNormalizer.php
+++ b/src/Analyser/ArgumentsNormalizer.php
@@ -76,6 +76,64 @@ final class ArgumentsNormalizer
 		)];
 	}
 
+	/**
+	 * @return array{ParametersAcceptor, FuncCall}|null
+	 */
+	public static function reorderCallUserFuncArrayArguments(
+		FuncCall $callUserFuncCall,
+		Scope $scope,
+	): ?array
+	{
+		$args = $callUserFuncCall->getArgs();
+		if (count($args) < 1 || count($args) > 2) {
+			return null;
+		}
+
+		$callbackArg = null;
+		$passThruArgs = [];
+		foreach ($args as $i => $arg) {
+			if ($callbackArg === null) {
+				if ($arg->name === null && $i === 0) {
+					$callbackArg = $arg;
+					continue;
+				}
+				if ($arg->name !== null && $arg->name->toString() === 'callback') {
+					$callbackArg = $arg;
+					continue;
+				}
+			}
+
+			$passThruArgs[] = new Arg(
+				$arg->value,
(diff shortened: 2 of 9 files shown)


conf/config.level5.neon
```diff
@@ -10,6 +10,8 @@ conditionalTags:
 		phpstan.rules.rule: %featureToggles.arrayFilter%
 	PHPStan\Rules\Functions\CallUserFuncRule:
 		phpstan.rules.rule: %featureToggles.callUserFunc%
+	PHPStan\Rules\Functions\CallUserFuncArrayRule:
+			phpstan.rules.rule: %featureToggles.callUserFunc%
```
staabm, line 14: should we have a separate flag?

staabm:
`call_user_func_array()` can return `false` on error - I need to verify when this can happen

src/Rules/Functions/CallUserFuncArrayRule.php
```diff
@@ -0,0 +1,81 @@
...
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\FunctionCallParametersCheck;
+use PHPStan\Rules\Rule;
+use function count;
+use function ucfirst;
+
+/**
+ * @implements Rule<FuncCall>
+ */
+class CallUserFuncArrayRule implements Rule
```
staabm, line 18: should this rule here also cover https://github.com/phpstan/phpstan/issues/5934 - or do we need a separate rule for that?

(and within this PR?)

conf/config.level5.neon
ondrejmirtes: No, this is fine

src/Analyser/ArgumentsNormalizer.php
```diff
@@ -76,6 +76,64 @@ public static function reorderCallUserFuncArguments(
...
+				if ($arg->name !== null && $arg->name->toString() === 'callback') {
+					$callbackArg = $arg;
+					continue;
+				}
+			}
+
+			$passThruArgs[] = new Arg(
+				$arg->value,
+				$arg->byRef,
+				true,
```
ondrejmirtes, line 109: Does this work solely thanks to `unpack` set to `true`? I don't really like this, we should process this ourselves. One reason is that if `$arg->unpack` in the original `call_user_func_array` is already true, this will lead to wrong results.

The easiest example is when the arg is just a single constant array, right. But we can also have multiple constant arrays in union, and also general arrays.

src/Analyser/ArgumentsNormalizer.php
```diff
@@ -76,6 +76,64 @@ public static function reorderCallUserFuncArguments(
...
+	/**
+	 * @return array{ParametersAcceptor, FuncCall}|null
+	 */
+	public static function reorderCallUserFuncArrayArguments(
+		FuncCall $callUserFuncCall,
+		Scope $scope,
+	): ?array
+	{
+		$args = $callUserFuncCall->getArgs();
+		if (count($args) < 1 || count($args) > 2) {
```
ondrejmirtes, line 88: Is call with three arguments really invalid? I'm not sure about it.

ondrejmirtes (changes requested):

src/Analyser/ArgumentsNormalizer.php
staabm: so to get it right, in case of

```php
	/**
	 * @param array{i: 5}|array{i: 15} $arr
	 */
	public function doFoo(array $arr): void
	{
		$f = function (int $i): void {

		};
		call_user_func_array(callback: $f, args: $arr);
	}
```

the argument normalizer needs to return multiple `FuncCall`s?

src/Analyser/ArgumentsNormalizer.php
staabm: at least its a fatal error https://3v4l.org/RBJfK in php8+

src/Analyser/ArgumentsNormalizer.php
ondrejmirtes: No. We just need to evaluate as if the callback was called like this:

```php
$f($i);
```

Where `$i` is `5|15`.

ondrejmirtes (changes requested):

PR #2502, opened 2023-07-01
Handle invokable exception thrown

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/6233

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1961,7 +1961,7 @@ class NodeScopeResolver
 				$throwPoints = $nameResult->getThrowPoints();
 				if (
 					$nameType->isObject()->yes()
-					&& $nameType->accepts(new ObjectType(Closure::class), true)->no()
+					&& (new ObjectType(Closure::class))->isSuperTypeOf($nameType)->no()
 				) {
 					$invokeResult = $this->processExprNode(
 						new MethodCall($expr->name, '__invoke', $expr->getArgs(), $expr->getAttributes()),
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1958,6 +1958,7 @@ class NodeScopeResolver
 				}
 
 				$nameResult = $this->processExprNode($expr->name, $scope, $nodeCallback, $context->enterDeep());
+				$throwPoints = $nameResult->getThrowPoints();
 				if ($nameType->isObject()->yes()) {
 					$invokeResult = $this->processExprNode(
 						new MethodCall($expr->name, '__invoke', $expr->getArgs(), $expr->getAttributes()),
@@ -1966,9 +1967,7 @@ class NodeScopeResolver
 						},
 						$context->enterDeep(),
 					);
-					$throwPoints = $invokeResult->getThrowPoints();
-				} else {
-					$throwPoints = $nameResult->getThrowPoints();
+					$throwPoints = array_merge($throwPoints, $invokeResult->getThrowPoints());
 				}
 				$scope = $nameResult->getScope();
 			} elseif ($this->reflectionProvider->hasFunction($expr->name, $scope)) {
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1961,6 +1961,7 @@ class NodeScopeResolver
 				$throwPoints = $nameResult->getThrowPoints();
 				if (
 					$nameType->isObject()->yes()
+					&& $nameType->isCallable()->yes()
 					&& (new ObjectType(Closure::class))->isSuperTypeOf($nameType)->no()
 				) {
 					$invokeResult = $this->processExprNode(
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1949,17 +1949,6 @@ class NodeScopeResolver
 			$throwPoints = [];
 			if ($expr->name instanceof Expr) {
 				$nameType = $scope->getType($expr->name);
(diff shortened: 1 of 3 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -1909,6 +1910,18 @@ function (MutatingScope $scope) use ($expr, $nodeCallback, $context): Expression
 			$throwPoints = [];
 			if ($expr->name instanceof Expr) {
 				$nameType = $scope->getType($expr->name);
+				if (
```
staabm: This looks like it would not support a union of object types. 

Could you test that?

VincentLanglet: The `TypeWithClassName` check wasn't excluding ClosureType, so I went back this way.
```
$nameType->isObject()->yes()
					&& !$nameType instanceof ClosureType
```

Is there a way to do this check without an `instanceof *Type` ? I know PHPStan codebase got rid of them.

staabm: Would Type-> getObjectClassReflections() work?

https://github.com/phpstan/phpstan-src/blob/3de7bbc6d1caad749873fb6b05560920fb046c36/src/Type/Type.php#L41

VincentLanglet: I found a way

VincentLanglet:
Friendly ping @ondrejmirtes 
I tried to follow your advice from https://github.com/phpstan/phpstan/issues/6233#issuecomment-1610198896

src/Analyser/NodeScopeResolver.php
```diff
@@ -1909,6 +1909,17 @@ function (MutatingScope $scope) use ($expr, $nodeCallback, $context): Expression
...
+					$nameType->isObject()->yes()
+					&& $nameType->accepts(new ObjectType(Closure::class), true)->no()
+				) {
+					return $this->processExprNode(
+						new MethodCall($expr->name, '__invoke', $expr->getArgs(), $expr->getAttributes()),
```
ondrejmirtes: Hi, I don't love this patch. It's gonna change how `$foo()` is being processed and it's gonna add another error message. Besides CallCallablesRule it's also gonna invoke CallMethodsRule and that's not great.

Also I'm not sure `$nameType->accepts(new ObjectType(Closure::class), true)->no()` is trying to do and it doesn't seem like there's a test for this scenario.

I'd accept if we called `processExprNode` with a "noop" `$nodeCallback` and just taken `$throwPoints` from that.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
```diff
@@ -1956,8 +1956,20 @@ function (MutatingScope $scope) use ($expr, $nodeCallback, $context): Expression
...
-				$throwPoints = $nameResult->getThrowPoints();
+				if ($nameType->isObject()->yes()) {
+					$invokeResult = $this->processExprNode(
+						new MethodCall($expr->name, '__invoke', $expr->getArgs(), $expr->getAttributes()),
+						$scope,
+						static function (): void {
+						},
+						$context->enterDeep(),
+					);
+					$throwPoints = $invokeResult->getThrowPoints();
```
ondrejmirtes: We don't want to ignore `$nameResult` throw points even in this branch.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
VincentLanglet: I used array_merge instead

src/Analyser/NodeScopeResolver.php
VincentLanglet: Tests are failing (https://github.com/phpstan/phpstan-src/actions/runs/6352469876/job/17255408049?pr=2502) without the extra check https://github.com/phpstan/phpstan-src/pull/2502/commits/838b9916f9350dd0b842c8dbe85716c38f047dab

This is because they are no method `__invoke` called on Closure

src/Analyser/NodeScopeResolver.php
```diff
@@ -1956,8 +1956,22 @@ function (MutatingScope $scope) use ($expr, $nodeCallback, $context): Expression
 						$nameType->getCallableParametersAcceptors($scope),
 					);
 				}
+
 				$nameResult = $this->processExprNode($expr->name, $scope, $nodeCallback, $context->enterDeep());
 				$throwPoints = $nameResult->getThrowPoints();
+				if (
+					$nameType->isObject()->yes()
+					&& $nameType->accepts(new ObjectType(Closure::class), true)->no()
```
ondrejmirtes: `accepts` isn't the right way to check for a type.

You need `(new ObjectType(Closure::class))->isSuperTypeOf($nameType)`.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
VincentLanglet: I updated the code with isSuperTypeOf

I always have difficulties to understand which method I should use.
Would it be possible to add comment with example above the method in the interface ?
Or at least explaining me the difference ?

Thanks.

staabm: I agree that its not that easy to find the correct method from `Type`.

most of the time we need `isSuperTypeOf` though ;-)

ondrejmirtes:
Added one more commit that makes sense here: https://github.com/phpstan/phpstan-src/pull/2502/commits/e38560a8ddf234ebaae68630c5e9fdcc18e441aa

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #2503, opened 2023-07-01
Prune non-wildcard excludePaths directories early

mvorisek:
uses backported symfony/symfony/pull/50877 with symfony/symfony/pull/50884 optimization

fix phpstan/phpstan#9554

in projects with massive amount of excluded files, this PR improves phpstan runtime by several times - in our case, the phpstan file scan speedup is 280 times (from 48 seconds to 0.17 seconds)

fully tested using VFS

### DIFF
--- a/src/File/FileFinder.php
+++ b/src/File/FileFinder.php
@@ -37,9 +37,12 @@ class FileFinder
 			} elseif (!file_exists($path)) {
 				throw new PathNotFoundException($path);
 			} else {
-				$finder = new Finder();
-				$finder->followLinks();
-				foreach ($finder->files()->name('*.{' . implode(',', $this->fileExtensions) . '}')->in($path) as $fileInfo) {
+				$finder = (new Finder())
+					->in($path)
+					->followLinks()
+					->files()
+					->name('*.{' . implode(',', $this->fileExtensions) . '}');
+				foreach ($finder as $fileInfo) {
 					$files[] = $this->fileHelper->normalizePath($fileInfo->getPathname());
 					$onlyFiles = false;
 				}
--- a/src/File/FileFinder.php
+++ b/src/File/FileFinder.php
@@ -9,6 +9,9 @@ use function array_values;
 use function file_exists;
 use function implode;
 use function is_file;
+use function str_starts_with;
+use function strlen;
+use function substr;
 
 class FileFinder
 {
@@ -25,6 +28,44 @@ class FileFinder
 	}
 
 	/**
+	 * Make relative directory prune pattern for Symfony Finder.
+	 *
+	 * https://github.com/symfony/symfony/blob/v6.2.12/src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php#L51
+	 * https://github.com/symfony/symfony/blob/v6.2.12/src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php#L70
+	 */
+	private function tryToMakeFinderExcludePattern(string $excludePath, string $inPath): ?string
+	{
+		$excludePath = $this->fileHelper->normalizePath($excludePath, '/');
+		$inPath = $this->fileHelper->normalizePath($inPath, '/');
+
+		if ($excludePath === $inPath || str_starts_with($inPath . '/', $excludePath)) {
+			return '.+';
+		} if (str_starts_with($excludePath, $inPath . '/')) {
+			return substr($excludePath, strlen($inPath) + 1);
+		}
+
+		return null;
+	}
+
+	/**
+	 * @return list<string>
+	 */
+	private function makeFinderExcludePatterns(string $inPath): array
+	{
+		$res = [];
+		foreach ($this->fileExcluder->getExcludedLiteralPaths() as $excludePath) {
+			$excludePattern = $this->tryToMakeFinderExcludePattern($excludePath, $inPath);
+			if ($excludePattern === null) {
+				continue;
+			}
+
+			$res[] = $excludePattern;
+		}
+
+		return $res;
+	}
+
+	/**
 	 * @param string[] $paths
 	 */
 	public function findFiles(array $paths): FileFinderResult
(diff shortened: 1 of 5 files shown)


src/File/FileFinder.php
```diff
@@ -23,6 +26,34 @@ public function __construct(
 	{
 	}
 
+	private function tryToMakeRelativeExcludePath(string $excludePath, string $inPath): ?string
+	{
+		$excludePath = $this->fileHelper->normalizePath($excludePath, '/');
+		$inPath = $this->fileHelper->normalizePath($inPath, '/');
+
+		return str_starts_with($excludePath, $inPath . '/')
+			? substr($excludePath, strlen($inPath) + 1)
```
mvorisek, line 35: how to fix the PHP 7.4- stan? https://github.com/phpstan/phpstan-src/actions/runs/5433198786/jobs/9880741952?pr=2503#step:12:23

src/File/FileFinder.php
```diff
@@ -23,6 +26,34 @@ public function __construct(
 	{
 	}
 
+	private function tryToMakeRelativeExcludePath(string $excludePath, string $inPath): ?string
```
mvorisek, line 29: for https://github.com/symfony/symfony/blob/v6.2.12/src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php#L62

ondrejmirtes (changes requested):
These changes need a test to show why they're done.

It can be a test calling PHPStan from the outside, there's plenty of those in phpstan/phpstan.

src/File/FileFinder.php
mvorisek: @ondrejmirtes may I have your advice here?

mvorisek:
> These changes need a test to show why they're done.
> 
> It can be a test calling PHPStan from the outside, there's plenty of those in phpstan/phpstan.

Can you please link here some test where phpstan is started from php? (as VFS from PHP needs to be created before phpstan is called)

mvorisek:
> These changes need a test to show why they're done.

done, tested using VFS, so minimal count of IO syscalls is asserted :)

mvorisek:
@ondrejmirtes this PR solves a major performance issue and in our case, it cuts the phpstan runtime by more than 20x, would you please mind reviwing this PR soon? Thank you.

ondrejmirtes:
Would be great if you were able to backport https://github.com/symfony/symfony/pull/50877 into Symfony 5.4, then we could just use their code without all the complex VFS tests.

mvorisek:
From Symfony perspective it is a feature and 5.x branch does no longer accept features. Is copying the pruning filtering here ok or when https://github.com/symfony/symfony/releases/tag/v6.4.0-BETA2 (or Symfony 6.4 in general) can be required in phpstan?

ondrejmirtes:
Not anytime soon, PHPStan still supports PHP 7.2+.

mvorisek:
This PR backports the needed class incl. tests. Can we merge it?

In our case, the file discovery speedup is 280 times.

Once Symfony 6.4+ could be used, we can remove the backported class.

ondrejmirtes:
> speedup is 280 times.

What it means in absolute numbers? 1ms vs. 280ms I'm not interested in but if the numbers are higher then for sure.

mvorisek:
see PR description - in our case, from 48 seconds to 0.17 seconds, the actual numbers depend on the files ignored

ondrejmirtes:
Sure we could merge this, but at the same time you could change your PHPStan config so that  you don't have "massive amount of excluded files". You could even change your project structure so that all your own is under `src/` or `app/` as you typically do with PSR-4.

mvorisek:
Let's merge this PR, it is highly supported by the community.

ondrejmirtes:
No, sorry, I can't merge it in the current state. It's too risky, I'm afrait of too much different behaviour of the newly introduced `FileFinderSymfonyBackport` class. It's 170 lines of code that looks for files completely differently than the currently used Symfony Finder class (which is 800+ different lines of code).

Compare that to your Symfony PR - https://github.com/symfony/symfony/pull/50877/files - it's just a 40 lines of actual production code changes in `src/Symfony/Component/Finder/Finder.php` and `src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php`.

If you'd write a `Symfony\Component\Finder\Finder` subclass that would only override the methods that it needs to override in order to achieve early pruning, it'd be much easier for me to merge this.

mvorisek:
Except https://github.com/phpstan/phpstan-src/pull/2503/commits/fa5ded5fd4d09fb572ad55ecd8762cfb527ead33 which can be reverted/dropped, it cannot be simplifed more. It is trully a backport of the newer Symfony. If you want this change, please either merge it as is or backport from scratch. If you do not want the change, close this PR. I will drop a tear but it is your repo.

ondrejmirtes:
This is not a backport. A backport would if you copied or extended this class https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Finder/Finder.php and made it work with PHP 7.2+ and in context of PHPStan codebase as it is.

I can't risk merging this. When I get a hunch some PR would cause problems, I'm usually right. I can't afford breaking users' analysis and spending the next week fixing bugs in code I don't understand.

To be able to merge this PR I'd have to see that the finder finds all the same files and finds them in the same order across various projects with complex configurations. But I don't see that here, sorry.

mvorisek:
> This is not a backport. A backport would if you copied or extended this class https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Finder/Finder.php and made it work with PHP 7.2+ and in context of PHPStan codebase as it is.

This is exactly what I did. I copied methods that are needed to backport the new feature, I extended the base class, as not all methods were needed to be copied. Also, with VFS test, I have prooven that the behaviour is the same as the one in Symfony, same test I contributed there.

ondrejmirtes:
> I extended the base class

The finder class introduced in this PR does not extend any class: `class FileFinderSymfonyBackport implements IteratorAggregate`

mvorisek:
The reason is https://github.com/symfony/symfony/blob/v7.0.1/src/Symfony/Component/Finder/Finder.php#L745 is private and implementing all `Finder` functionality will result in more LOC copied and untested. Thus I copied only the relevant/needed methods. Help with backporting https://github.com/symfony/symfony/pull/50877 welcomed.

PR #2506, opened 2023-07-02
Fix path with scheme normalization

mvorisek:
scheme grammar: https://www.rfc-editor.org/rfc/rfc3986#page-17

### DIFF
--- a/src/File/FileHelper.php
+++ b/src/File/FileHelper.php
@@ -7,11 +7,12 @@ use function array_pop;
 use function explode;
 use function implode;
 use function ltrim;
+use function preg_match;
 use function rtrim;
 use function str_replace;
-use function str_starts_with;
 use function strlen;
 use function strpos;
+use function strtolower;
 use function substr;
 use function trim;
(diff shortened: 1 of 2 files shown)


mvorisek:
PR is done, CI failures are unrelated.

src/File/FileHelper.php
```diff
@@ -43,7 +44,7 @@ public function absolutizePath(string $path): string
 				return $path;
 			}
 		}
-		if (str_starts_with($path, 'phar://')) {
+		if (str_contains($path, '://')) {
```
ondrejmirtes: AFAIK this change isn't covered by tests

ondrejmirtes (changes requested):

src/File/FileHelper.php
mvorisek: test added

mvorisek: done

mvorisek:
Can this smallfix be merged?

ondrejmirtes:
Thank you.

PR #2508, opened 2023-07-04
Allow null in mysqli*real_connect

staabm:
see https://www.php.net/mysqli_real_connect

fixes https://phpstan.org/r/468e65d5-36ab-4e4e-8f9d-dee8ce260373


ondrejmirtes:
Thank you.

PR #2509, opened 2023-07-04
More precise session_status() return type

staabm:
see https://www.php.net/session_status

locally tested.. not sure this trivial changes need a dedicated test?


ondrejmirtes:
Thank you!

PR #2511, opened 2023-07-05
Draft: Add AccessVariablePropertiesRule

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/9475

### DIFF
--- a/tests/PHPStan/Rules/Comparison/data/bug-9475.php
+++ b/tests/PHPStan/Rules/Comparison/data/bug-9475.php
@@ -8,6 +8,8 @@ class Foo
 {
 	/** @var string */
 	private $foo = 'Foo';
+	/** @var self */
+	protected $bar;
 
 	/** @param mixed[] $array */
 	function mustReport(Bug9475 $foo, array $array): void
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Comparison/data/bug-9475.php
```diff
@@ -0,0 +1,36 @@
...
+
+class Bar implements Stringable
+{
+	private string $foo = 'Foo';
+
+	function mustReport(Bug9475 $foo): void
+	{
+		echo $this->foo;
+		echo $this->$foo;
+		echo $this->$this;
```
zonuexe, line 39: `$this` is a Stringable so this line is legal, but heuristically it looks like a typo.
It is also possible to alert another message for this code.

ondrejmirtes:
Hi, thank you for this, but the approach isn't exactly right. To name a few things:

1) There's nothing special about `Variable` node, the rule needs to check all expressions. For example these things are possible too: https://3v4l.org/ULseu
2) There's no need to have different rules for reading/writing properties. We can just check PropertyFetch.
3) The logic in your rule misses some cases. We need to check that the expr type can be used as a string. The `!$type instanceof ErrorType` does not achieve that.

Please base your work on top of this branch (this rule just needs some tests to be written for it): https://github.com/phpstan/phpstan-src/tree/property-expr-name

And after that PR is merged, we can cover more expressions as listed here: https://github.com/phpstan/phpstan/issues/9475#issuecomment-1602082553

Thank you!

PR #2513, opened 2023-07-07
Bleeding edge - More precise ftp_* mode parameter type

staabm:
everything else unchanged

inspired by https://github.com/staabm/phpstorm-stubs/blob/master/ftp/ftp.php


ondrejmirtes:
Thank you.

PR #2514, opened 2023-07-07
Simplify access to class reflections in InClassNode rules.

mad-briller:
`InClassNodes` are guaranteed to be inside classes and can access the class reflection safely through the node instead of unsafely through the scope


ondrejmirtes:
Yes! 😊

ondrejmirtes:
Thank you.

mad-briller:
deleting code is my favorite

ondrejmirtes:
There are more opportunities like that, for example MethodReturnStatementsNode. It could have ClassReflection and MethodReflection inside but currently doesn't.

PR #2515, opened 2023-07-07
Set class and method reflection in the MethodReturnStatementNode.

mad-briller:
Is this what you meant by https://github.com/phpstan/phpstan-src/pull/2514#issuecomment-1625179766 ?

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -593,12 +593,22 @@ class NodeScopeResolver
 
 					$gatheredReturnStatements[] = new ReturnStatement($scope, $node);
 				}, StatementContext::createTopLevel());
+
+				$classReflection = $scope->getClassReflection();
+
+				$methodReflection = $methodScope->getFunction();
+				if (!$methodReflection instanceof MethodReflection) {
+					throw new ShouldNotHappenException();
+				}
+
 				$nodeCallback(new MethodReturnStatementsNode(
 					$stmt,
 					$gatheredReturnStatements,
 					$gatheredYieldStatements,
 					$statementResult,
 					$executionEnds,
+					$classReflection,
+					$methodReflection,
 				), $methodScope);
 			}
 		} elseif ($stmt instanceof Echo_) {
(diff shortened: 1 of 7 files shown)


mad-briller:
not sure if github is broken atm but i can't seem to change the target to `1.10.x` cos the dropdown isn't opening

src/Analyser/NodeScopeResolver.php
```diff
@@ -593,12 +593,25 @@ private function processStmtNode(
...
 					$gatheredReturnStatements[] = new ReturnStatement($scope, $node);
 				}, StatementContext::createTopLevel());
+
+				$classReflection = $methodScope->getClassReflection();
+				if ($classReflection === null) {
+					throw new ShouldNotHappenException();
+				}
+
+				$methodReflection = $methodScope->getFunction();
+				if (!$methodReflection instanceof MethodReflection) {
```
mad-briller, line 600: not sure if this is the best way to get the reflections to pass into the node

src/Analyser/NodeScopeResolver.php
```diff
@@ -593,12 +593,25 @@ private function processStmtNode(
 
 					$gatheredReturnStatements[] = new ReturnStatement($scope, $node);
 				}, StatementContext::createTopLevel());
+
+				$classReflection = $methodScope->getClassReflection();
```
ondrejmirtes: You could do this on `$scope` and then you don't need to check for `null` - because of `if (!$scope->isInClass()) {` on line 526.

ondrejmirtes (changes requested):
MethodReturnStatementsNode is read in NodeScopeResolver around line 4417, we could access ClassReflection on `$node` instead of `$scope` there too :)

src/Analyser/NodeScopeResolver.php
mad-briller: is the `isInClass` thing a custom type-specifier in phpstan?

ondrejmirtes: Not anymore, it just has `@phpstan-assert` above it.

mad-briller: ah i didn't know they could do that, nice

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2516, opened 2023-07-07
Fix array_keys() and array_values() on constant arrays

jlherren:


jlherren:
Fixes these wrong results: https://phpstan.org/r/a8999958-c27f-46c0-8110-e6b4f73dc896

ondrejmirtes:
Thank you!

PR #2517, opened 2023-07-08
Bleeding edge - more precise extract() signature

staabm:
see https://www.php.net/manual/en/function.extract

defines flag types


ondrejmirtes:
Conflict here.

staabm:
rebased

ondrejmirtes:
Thank you.

PR #2519, opened 2023-07-08
Bleeding edge - more precise stream_socket_client() signature

staabm:
https://www.php.net/manual/en/function.stream-socket-client

define connect flags


ondrejmirtes:
Thank you.

PR #2520, opened 2023-07-08
Simplify access to class reflections in ClassPropertiesNode rules.

mad-briller:


src/Node/ClassPropertiesNode.php
```diff
@@ -98,10 +103,7 @@ public function getUninitializedProperties(
 		if (!$this->getClass() instanceof Class_) {
 			return [[], [], []];
 		}
-		if (!$scope->isInClass()) {
-			throw new ShouldNotHappenException();
-		}
-		$classReflection = $scope->getClassReflection();
+		$classReflection = $this->getClassReflection();
```
mad-briller, line 106: the `$scope` param on this method is no longer used as we can access the class reflection on the node now, however removing it would be a BC break and this class is covered by `@api` so i just left it there

ondrejmirtes:
I could see a scenario where this would be a BC break (you don't know which Scope the user was passing into the method) but in practice it's very unlikely. Thank you!

mad-briller:
that's a great point i didn't consider to be honest yeah... it would be a weird way to use the node + scope together though

i'll try to get to the pr feedbacks after work later, thanks!

PR #2522, opened 2023-07-08
Simplify access to function reflection in FunctionReturnStatementsNode rules.

mad-briller:
you were right @ondrejmirtes, there were several opportunities :D


ondrejmirtes:
Thank you!

PR #2526, opened 2023-07-10
More precise return type for trigger_error

paulbalandan:
refs:
- https://3v4l.org/22DUP
- https://www.php.net/manual/en/function.trigger-error.php


ondrejmirtes:
Thank you.

PR #2527, opened 2023-07-11
Fix signature of `DOMNode::lookupNamespaceURI`

paulbalandan:
Fixes https://github.com/phpstan/phpstan/issues/9504

Ref:
- https://github.com/php/php-src/blob/c40dcf93d0b95d9a1026476b202f5aa965fb3fdd/ext/dom/node.c#L1622-L1650
- https://github.com/php/php-src/blob/b3553159a7b15d1ddd485aba9f83485f08e2e800/ext/dom/node.c#L1515-L1543
- https://www.w3.org/TR/DOM-Level-3-Core/core.html#Node3-lookupNamespaceURI

I think phpstorm-stubs should also be updated.

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -1988,7 +1988,7 @@ return [
 'DOMNode::isDefaultNamespace' => ['bool', 'namespaceuri'=>'string'],
 'DOMNode::isSameNode' => ['bool', 'node'=>'DOMNode'],
 'DOMNode::isSupported' => ['bool', 'feature'=>'string', 'version'=>'string'],
-'DOMNode::lookupNamespaceURI' => ['string', 'prefix'=>'string'],
+'DOMNode::lookupNamespaceURI' => ['?string', 'prefix'=>'?string'],
 'DOMNode::lookupPrefix' => ['string', 'namespaceuri'=>'string'],
 'DOMNode::normalize' => ['void'],
 'DOMNode::removeChild' => ['DOMNode', 'oldnode'=>'DOMNode'],


resources/functionMap.php
```diff
@@ -1988,7 +1988,7 @@
 'DOMNode::isDefaultNamespace' => ['bool', 'namespaceuri'=>'string'],
 'DOMNode::isSameNode' => ['bool', 'node'=>'DOMNode'],
 'DOMNode::isSupported' => ['bool', 'feature'=>'string', 'version'=>'string'],
-'DOMNode::lookupNamespaceURI' => ['string', 'prefix'=>'string'],
+'DOMNode::lookupNamespaceURI' => ['?string', 'prefix'=>'?string'],
```
staabm, line 1991: are you sure the prefix parameter is nullable?

per php.net docs it is not nullable
https://www.php.net/manual/en/domnode.lookupnamespaceuri.php

I don‘t know the details of the C php sources.. in case you are sure, please consider open a PR for the php.net docs

——

the return type seems fine

paulbalandan: Comment there says so: https://www.php.net/manual/en/domnode.lookupnamespaceuri.php#75556
Also in php-src master: https://github.com/php/php-src/blob/b3553159a7b15d1ddd485aba9f83485f08e2e800/ext/dom/php_dom.stub.php#L378-L379

Also, the referenced C sources don't change except for the `return;` (7.4.33) changing to `RETURN_THROWS();` (master).

ondrejmirtes:
Thank you.

resources/functionMap.php
mvorisek: @paulbalandan please update the https://github.com/php/doc-en too

PR #2529, opened 2023-07-13
Add base64_decode overloads to functionMap_php80delta

dktapps:
fixes phpstan/phpstan#9428

Longer term a better solution should probably be made for this, but I think this should work as duct tape for now.


ondrejmirtes:
Thank you!

PR #2531, opened 2023-07-14
report trait access to unititialized property in the trait instead of class

schlndh:
There is a bug in how access to uninitialized properties is reported. If the access is in a trait it was reported as being in the class, but the line number was from the trait, so it didn't make sense. This is an attempt to fix it so that the trait access gets reported similarly to other trait issues (i.e. `trait (in context of class)`).


ondrejmirtes:
Nice, thank you!

PR #2537, opened 2023-07-19
Avoid doing work before its necessary in mutating scope.

mad-briller:
Few instances of asking for information before it's really necessary

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -1119,11 +1119,11 @@ class MutatingScope implements Scope
 					return new ObjectType(Closure::class);
 				}
 
-				$classType = $this->resolveTypeByName($node->class);
 				if (!$node->name instanceof Node\Identifier) {
 					return new ObjectType(Closure::class);
 				}
 
+				$classType = $this->resolveTypeByName($node->class);
 				$methodName = $node->name->toString();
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

mad-briller:
@ondrejmirtes do you think it would be feasible to be able to detect this kind of thing with a custom phpstan rule?

i've written custom rules for single nodes but this would need to check more than one node so i don't know if it's possible, or if it is how i would approach it 

as a simple minimum, the rule could be checking for something like "if a variable is assigned and not used before an exit point"

ondrejmirtes:
Please see this: https://github.com/phpstan/phpstan/issues/8192

But read my last comment too - I was able to solve a bug without CFG even when I thought I needed CFG for that 😊 Maybe it will give you some ideas.

Also bear in mind that the called thing that we move lower needs to be pure/immutable otherwise moving the call might have unintended side effects.

mad-briller:
okay that gives me something to read and think about, thanks!

great point about purity, luckily phpstan already knows about the concept

mad-briller:
making changes to `NodeScopeResolver` isn't possible in an extension, does this mean you would be open to having this kind of rule in `phpstan-src` itself?

ondrejmirtes:
I don't understand the question. What kind of changes to the NodeScopeResolver?

PR #2538, opened 2023-07-21
Replaced text string "does not do anything" to "does nothing."

travisfont:
Should say "does nothing" to be simplified, more transparent, and proper English.


ondrejmirtes:
I believe that "does not do anything" is proper English, and also we cannot do this change for backward compatibility reason (existing `ignoreErrors` entries).

PR #2539, opened 2023-07-22
Array is non-empty when in_array is not strict

staabm:
implements true-context type narrowing to non-empty-array for non-strict use of `in_array`.
with this PR we only narrow when `$needle` is of the same type as the `$haystack`'s iterable type.

when non-strict, the loose-compare falsey context would be pretty complicated, therefore this PR just implements the true-context part (which already fixes 2 reported issues).

there is still room for improvement in true and false-context but this is a good first step which should already cover a lot of real-world code.

closes https://github.com/phpstan/phpstan/issues/9662
closes https://github.com/phpstan/phpstan/issues/10228


staabm:
@ondrejmirtes I think getting this merged would help a lot of non-strict codebases, because `in_array` is usally used a lot

ondrejmirtes:
Thank you!

ghost:
Hi, this PR does not close https://github.com/phpstan/phpstan/issues/10228. Here I have the updated playground after your change https://phpstan.org/r/026707d1-2de9-4bbb-8f1c-d47312de414c.

PR #2543, opened 2023-07-24
Support of printf with dynamic width

kukulich:
Fixes https://github.com/phpstan/phpstan/issues/9674

### DIFF
--- a/src/Rules/Functions/PrintfParametersRule.php
+++ b/src/Rules/Functions/PrintfParametersRule.php
@@ -116,7 +116,7 @@ class PrintfParametersRule implements Rule
 
 		$specifiers = sprintf($specifiers, $addSpecifier);
 
-		$pattern = '~(?<before>%*)%(?:(?<position>\d+)\$)?[-+]?(?:[ 0]|(?:\'[^%]))?-?\d*(?:\.\d*)?' . $specifiers . '~';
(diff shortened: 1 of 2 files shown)


src/Rules/Functions/PrintfParametersRule.php
```diff
@@ -116,7 +116,7 @@ private function getPlaceholdersCount(string $functionName, string $format): int
 
 		$specifiers = sprintf($specifiers, $addSpecifier);
 
-		$pattern = '~(?<before>%*)%(?:(?<position>\d+)\$)?[-+]?(?:[ 0]|(?:\'[^%]))?-?\d*(?:\.\d*)?' . $specifiers . '~';
+		$pattern = '~(?<before>%*)%(?:(?<position>\d+)\$)?[-+]?(?:(?:[ 0]|(?:\'[^%]))(?<width>\*)?)?-?\d*(?:\.\d*)?' . $specifiers . '~';
```
fmarchalemisys, line 119: Looking at this regex, I see the precision doesn't allow for a `*` either.

Sorry for not testing that case before opening the issue :confounded: 

From the [doc](https://www.php.net/manual/en/function.printf):

> Note: If the period is specified without an explicit value for precision, 0 is assumed. If * is used, the precision is supplied as an additional integer value preceding the one formatted by the specifier.

kukulich: @fmarchalemisys Please open new issue.

PR #2546, opened 2023-07-25
Fix `DOMDocument::loadXML` return type

fluffycondor:
https://www.php.net/manual/en/domdocument.loadxml.php#refsect1-domdocument.loadxml-returnvalues


VincentLanglet:
This was already resolved by https://github.com/phpstan/phpstan-src/pull/2975 (which closed https://github.com/phpstan/phpstan/issues/10057) and could be closed then @ondrejmirtes

ondrejmirtes:
Thank you.

PR #2548, opened 2023-07-25
Fix types for conditional parameters after checks

rvanvelzen:
Fix phpstan/phpstan#7915


ondrejmirtes:
Perfect, thank you!

mvorisek:
@rvanvelzen https://phpstan.org/r/aefdde67-72f5-4562-b374-9f5afb64391b can you please look into nested ternary? It seems it does work with one ternary/conditional type only.

PR #2550, opened 2023-07-25
fix: mark `Ds\Queue::pop()` as impure

simPod:
Resolves [#9682](https://github.com/phpstan/phpstan/issues/9682)

### DIFF
--- a/stubs/ext-ds.stub
+++ b/stubs/ext-ds.stub
@@ -457,6 +457,7 @@ interface Sequence extends Collection, ArrayAccess
 	/**
 	 * @return TValue
 	 * @throws \UnderflowException
+	 * @phpstan-impure
 	 */
 	public function pop();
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Can you please check other similar methods that might have the same problem? The key is that the method returns a value AND changes the underlying collection.

Such an example: `Stack::pop()`

If it does not return a value (it's `void`) then such a method is already understood as impure by PHPStan.

simPod:
I tried to mark all eligible I know of.

ondrejmirtes:
I don't think that first/last and peek are impure.

simPod:
ok i missed "changes" in `The key is that the method returns a value AND changes the underlying collection.`

ondrejmirtes:
Thank you.

PR #2551, opened 2023-07-25
Fix ftp related function signatures

thg2k:
Fixes the following issues:
 - Fixed optional 'mode' for file transfer functions (optional since php 7.3.0)
 - Added missing false value for 'ftp_nb_put'
 - ~Consistent naming of the first parameter 'stream' (it should actually be 'ftp' since 8.x, but then it should be changed in the delta file)~ (removed per maintainer request)


ondrejmirtes:
> Consistent naming of the first parameter 'stream'

Please remove this change. There's no need to change the name.

thg2k:
I removed the change of the parameter name. Although i don't understand why it should stay inconsistent across the same family of functions. Also rebased to current 1.10.x

ondrejmirtes:
Thank you!

PR #2552, opened 2023-07-27
Rework evictPrivateSymbols in PhpClassReflectionExtension to avoid unnecessary string comparisons.

mad-briller:
this avoids the workload of this method increasing as the amount of cached reflections increases


staabm:
the change itself makes sense. did this method show up in any profile?

mad-briller:
it did not, i was just looking around the codebase at how things work and noticed we were accessing a hash table as a list

PR #2554, opened 2023-07-29
Prevent unnecessary work in MutatingScope

staabm:
while looking into https://github.com/phpstan/phpstan/issues/9690 this stood out. but it does not yet resolve the perf problem


ondrejmirtes:
Thank you.

PR #2555, opened 2023-07-29
connection_* functions are impure

staabm:
~~requires https://github.com/JetBrains/phpstorm-stubs/pull/1560~~


bin/generate-function-metadata.php
```diff
@@ -60,6 +61,21 @@ public function enterNode(Node $node)
 			return null;
 		}
 
+		private function isHasSideEffectsAnnotated(Attribute $attr)
```
staabm: according to a conversation with a contributor on JetBrains/phpstorm-stubs, the `Pure` attribute has [different semantics based on the 1st parameter](https://github.com/JetBrains/phpstorm-stubs/pull/1560#issuecomment-1684166592)

staabm: the regenerated function map changed a lot.. I am not 100% sure all these changes make sense - at least no tests fail

staabm: hmm I think I over-think this thing and I should instead just have added a new entry to the `functionMetadata_original`.

ondrejmirtes:
Thank you.

PR #2563, opened 2023-08-06
Implement ClassReflection->getParentClassName()

staabm:
A class can declare a parent which we cannot locate via autoloading mechanics.
For some static analysis cases its still useful to know, whether a parent class is defined - even if we can't reflect on it for some reason.

I was not able to find a API which allows me to dermine whether a class has a parent declared but only BetterReflection cannot find its definition from whether there is no parent class declared at all.

Before this PR we had to use reflection hacks to get the private property out of the BetterReflection class, see https://github.com/rectorphp/rector/issues/8093#issuecomment-1657057947

requires https://github.com/ondrejmirtes/BetterReflection/pull/31

----

TODO: requires a similar thing for interfaces and traits

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -189,7 +189,11 @@ class ClassReflection
 	/** @return class-string|null */
 	public function getParentClassName(): string|null
 	{
-		return $this->getParentClassName();
+		if ($this->reflection instanceof ReflectionEnum) {
+			return null;
+		}
+
+		return $this->reflection->getParentClassName();
 	}
 
 	public function getParentClass(): ?ClassReflection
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
I'm sorry, but you're calling a method that doesn't exist even on BetterReflection 6.12.0. You haven't added the method on Adapter classes.

staabm:
> You haven't added the method on Adapter classes.

see https://github.com/Roave/BetterReflection/pull/1359#discussion_r1279105632 - the BetterReflection maintainers don't want the method on the Adapter classes and argumented that PHPStan has its own fork to add this methods.

thats why I created https://github.com/ondrejmirtes/BetterReflection/pull/31

ondrejmirtes:
FYI I still don't think we need this. From now on you can just call the Adapter methods on `ClassReflection::getNativeReflection()`.

PR #2566, opened 2023-08-08
Added AST based type retrieval methods on ClassReflection

staabm:


ondrejmirtes:
I've already told you in a comment this isn't necessary :) You can simply call these methods on `ClassReflection::getNativeReflection()` in your code. I'd rather not confuse users with these methods.

staabm:
I see. thank you.

PR #2574, opened 2023-08-17
ConditionalTagsExtension: add missing tags

janedbal:


ondrejmirtes (changes requested):
I don't think this list is complete yet.

Please search phpstan-src for `* services:`, you should get all the extensions that support registering via service tag in configurafion files. And add them to this list :) Thank you.

janedbal:
Added, hopefully I found all of them. Checked also `_TAG =` and `'phpstan.` usages.

ondrejmirtes:
Thank you!

PR #2576, opened 2023-08-18
Ability to disable implicit throws in tests

janedbal:


janedbal:
Currently you are unable to test things like https://github.com/phpstan/phpstan/issues/9779

ondrejmirtes:
Thank you!

PR #2577, opened 2023-08-18
Fix several issues with nullsafe type specification

schlndh:
The goal of this PR is to fix https://phpstan.org/r/c17b503c-a789-4d90-87a7-31f3436f1b68 .

These issues were also fixed:

- https://github.com/phpstan/phpstan/issues/9105
- https://github.com/phpstan/phpstan/issues/5172
- https://github.com/phpstan/phpstan/issues/8517
- https://github.com/phpstan/phpstan/issues/7980
- https://github.com/phpstan/phpstan/issues/8664
- https://github.com/phpstan/phpstan/issues/9293

The problem is that `TypeCombinator::containsNull` returns false for `MixedType` (there's even a test for that and some usages definitely rely on this behavior). That seems quite strange to me and it's possible that there are other issues due to this confusing method. But I replaced only those usages for which I could come up with a test.

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1395,7 +1395,7 @@ class TypeSpecifier
 	{
 		if ($scope !== null) {
 			if ($context->true()) {
-				$containsNull = TypeCombinator::containsNull($type) && TypeCombinator::containsNull($scope->getType($expr));
+				$containsNull = ! $type->isNull()->no() && ! $scope->getType($expr)->isNull()->no();
 			} elseif ($context->false()) {
 				$containsNull = !TypeCombinator::containsNull($type) && TypeCombinator::containsNull($scope->getType($expr));
 			}
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1395,9 +1395,9 @@ class TypeSpecifier
 	{
 		if ($scope !== null) {
 			if ($context->true()) {
-				$containsNull = ! $type->isNull()->no() && ! $scope->getType($expr)->isNull()->no();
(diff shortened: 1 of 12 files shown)


ondrejmirtes:
Yes, you got it right, there's a difference between `TypeCombinator::containsNull()` and `Type::isNull()`. Before there was `Type::isNull()` we used `(new NullType())->isSuperTypeOf($type)`. This way might still be used in some places and should be roughly equivalent to `Type::isNull()`.

These two methods are different on purpose, sometimes you want one and sometimes you want the other one. It's possible that the usage is wrong in some places as you already figured out :)

ondrejmirtes (changes requested):
Please move the tests out of `tests/PHPStan/Analyser/AnalyserIntegrationTest.php`. This test class is reserved for tests that crash PHPStan or cause performance problems.

Where to correctly put these tests? It depends on whether the issue creator complained about a specific error message (then the test belongs in the relevant rule test class), or if they complained about something that can be tested with `assertType` - then the test belongs to NodeScopeResolverTest.

It's also possible to reference the same file from a rule test class and from NodeScopeResolverTest if you'd want.

Thank you.

ondrejmirtes:
Of course this fix is hugely valuable and also satisfying given the number of issues issue-bot found it fixes :) Thank you very much.

schlndh:
@ondrejmirtes Ok, thanks for the feedback. I moved the tests out of the AnalyserIntegrationTest. I kept most of them both in a rule test and in the NodeScopeResolverTest to make sure that both the `assertType`s are valid and that the rule is not triggered.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

herndlm:
6 issues fixed with a two-line-change, nice! :)

PR #2578, opened 2023-08-18
Fix array_slice() dynamic return type

dravnic:
When `array_slice()` `offset` or `limit` parameters aren't constants we cannot dynamically override return type, so for those cases return type should stay `array`. 

This PR fixes this and as well it includes minor refactor and tests for array_slice offset/limit parameters as variables.

### DIFF
--- a/src/Testing/ErrorFormatterTestCase.php
+++ b/src/Testing/ErrorFormatterTestCase.php
@@ -78,6 +78,7 @@ abstract class ErrorFormatterTestCase extends PHPStanTestCase
 			throw new ShouldNotHappenException();
 		}
 
+		/** @var list<Error> */
 		$fileErrors = array_slice([
 			new Error('Foo', self::DIRECTORY_PATH . '/folder with unicode 😃/file name with "spaces" and unicode 😃.php', 4),
 			new Error('Foo', self::DIRECTORY_PATH . '/foo.php', 1),
@@ -86,6 +87,7 @@ abstract class ErrorFormatterTestCase extends PHPStanTestCase
 			new Error("Bar\nBar2", self::DIRECTORY_PATH . '/foo.php', null),
 		], 0, $numFileErrors);
 
+		/** @var list<string> */
 		$genericErrors = array_slice([
 			'first generic error',
 			'second generic error',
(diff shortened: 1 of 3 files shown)


src/Testing/ErrorFormatterTestCase.php
```diff
@@ -78,6 +78,7 @@ protected function getAnalysisResult(int $numFileErrors, int $numGenericErrors):
 			throw new ShouldNotHappenException();
 		}
 
+		/** @var list<Error> */
```
staabm, line 81: why are these phpdoc changes required? is it related to your extension changes?

dravnic: > why are these phpdoc changes required? is it related to your extension changes?

yes, it's related to the changes. the `array_slice` is invoked with `$numFileErrors` as the 3rd (limit) param, so with the new changes it will return `array` as type not a dynamic type as before (in this case `list<Error>`) so I had to add phpdoc to override this.

this is safe because here https://github.com/phpstan/phpstan-src/blob/624edbbd84f346bab195bbfebcf6e5993a180bb1/src/Testing/ErrorFormatterTestCase.php#L77 is a check for range of `$numFileErrors` so array_slice will always return a list<Error> considering input is a constant array, but in general `array_slice` shouldn't do dynamic types if offset or limit are variables as array_slice return value cannot be predicted, hence this PR.

ondrejmirtes: > will always return a list

The extension should reflect that. With your changes the function is now less usable than before. We shouldn't need `/** @var list<Error> */` here.

ondrejmirtes (changes requested):

dravnic:
@ondrejmirtes here a simple example of a false positive error as a result of using `array_slice` with a constant array and `$offset` that isn't a constant:

https://phpstan.org/r/83b542de-76ba-41e4-9be6-64f29a346534

VincentLanglet:
Current result is https://phpstan.org/r/852be7a1-9f16-41ba-9160-be965fc944cc

And your issue is solved @dravnic 
https://phpstan.org/r/20cbae06-d782-4c18-920b-3307c0db85b7

So this PR can be closed

PR #2581, opened 2023-08-23
Support dynamicConstantNames in ConstantRule

staabm:
closes https://github.com/phpstan/phpstan/issues/9794

### DIFF
--- a/conf/config.level1.neon
+++ b/conf/config.level1.neon
@@ -9,8 +9,17 @@ parameters:
 
 rules:
 	- PHPStan\Rules\Classes\UnusedConstructorParametersRule
-	- PHPStan\Rules\Constants\ConstantRule
 	- PHPStan\Rules\Functions\UnusedClosureUsesRule
 	- PHPStan\Rules\Variables\EmptyRule
 	- PHPStan\Rules\Variables\IssetRule
 	- PHPStan\Rules\Variables\NullCoalesceRule
+
+services:
+	-
+		class: PHPStan\Rules\Constants\ConstantRule
+		factory: @PHPStan\Rules\Constants\ConstantRuleFactory::create()
+		tags:
+			- phpstan.rules.rule
+
+	-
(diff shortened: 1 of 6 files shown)


conf/config.level1.neon
```diff
@@ -9,8 +9,17 @@ parameters:
...
-	- PHPStan\Rules\Constants\ConstantRule
 	- PHPStan\Rules\Functions\UnusedClosureUsesRule
 	- PHPStan\Rules\Variables\EmptyRule
 	- PHPStan\Rules\Variables\IssetRule
 	- PHPStan\Rules\Variables\NullCoalesceRule
+
+services:
+	-
+		class: PHPStan\Rules\Constants\ConstantRule
+		factory: @PHPStan\Rules\Constants\ConstantRuleFactory::create()
```
staabm, line 20: I took inspiration from `ConstantResolverFactory` to create this factory which uses the same way to retrieve `%dynamicConstantNames%`.

My first try was passing `%dynamicConstantNames%` directly via `arguments:` but this didn't work.

staabm:
I measured the memory consumption before/after the PR on my windows PC with `php bin/phpstan clear-result-cache -q && php bin/phpstan --debug -vvv`.

before this PR: 500 MB
after this PR: 504 MB

```
PHP 8.2.7 (cli) (built: Jun  7 2023 10:25:24) (NTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.2.7, Copyright (c) Zend Technologies
```

I don't think its worth optimizing

staabm:
I just [saw this same memory error on PHP7.2 on other PRs](https://github.com/phpstan/phpstan-src/actions/runs/5975444517/job/16211409905?pr=2585), which means the "memory increase" is likely a intermittent and not triggered by this PR

ondrejmirtes:
Closing because we need a different direction: https://github.com/phpstan/phpstan/issues/9794#issuecomment-1740797119

staabm:
thanks for the feedback.

PR #2585, opened 2023-08-25
Prevent most `$parentPhpDocBlock->getClassReflection()->isBuiltin()` calls

staabm:
micro optimizing - re-shuffle conditions, since we only need to call into the reflection class for constructors


ondrejmirtes:
Thank you!

PR #2586, opened 2023-08-25
Blackfire: Add distributed profilling support

staabm:
As per [blackfire docs](https://blackfire.io/docs/php/integrations/sdk#sub-profiles-php) this is what is necessary to use distributed profiles with blackfire.

I was not yet able to verify it works, since the plan I am on seems to not include distributed profilling.
I have to double check on my personal notebook, where I have more advanced features available.

In case PHPStan has a license with distributed profilling enabled, you may be able to test on your own, whether it works.

The goal is, to gather profiles which show the main and the worker processes at the same time like shown in 

subprofiles

https://blackfire.io/docs/profiling-cookbooks/distributed-profiling#distributed-profiling

until now, I was using blackfire only without parallel workers enabled, but this might help to also profile the parallel scenario which is what the PHPStan users are in the end using most likely.

### DIFF
--- a/src/Parallel/Process.php
+++ b/src/Parallel/Process.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Parallel;
 
+use BlackfireProbe;
 use Exception;
 use PHPStan\ShouldNotHappenException;
 use React\EventLoop\LoopInterface;
@@ -9,6 +10,7 @@ use React\EventLoop\TimerInterface;
 use React\Stream\ReadableStreamInterface;
 use React\Stream\WritableStreamInterface;
 use Throwable;
+use function class_exists;
 use function fclose;
 use function is_string;
 use function rewind;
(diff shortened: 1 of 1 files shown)


src/Parallel/Process.php
```diff
@@ -60,9 +60,16 @@ public function start(callable $onData, callable $onError, callable $onExit): vo
 		if ($tmpStdErr === false) {
 			throw new ShouldNotHappenException('Failed creating temp file for stderr.');
 		}
+
+		$env = $_ENV;
+		if (class_exists('BlackfireProbe') && \BlackfireProbe::isEnabled()) {
```
staabm: this class is available when the php process was started via blackfire CLI like

`blackfire run --ignore-exit-status php bin/phpstan analyse`

ondrejmirtes:
When I try a parallel run with this PR and Blackfire enabled, it gets stuck on:

```
phpstan pr2586 $ blackfire run php -d memory_limit=448M bin/phpstan
The profile will be stored in your Personal environment. The "--environment" option can be used to specify the target environment.
Note: Using configuration file /Users/ondrej/Development/phpstan/phpstan.neon.
    0/1559 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0%
```

And does not move.

It'd be nice if it worked but it's not a big deal, we can profile performance issues in a single thread.

ondrejmirtes:
Oh it moves, but it's very very slow. Much slower than profiling a single thread.

PR #2587, opened 2023-08-26
Fix: checkBenevolentUnionTypes false positive with templates

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9766


ondrejmirtes:
Thank you!

PR #2588, opened 2023-08-26
Add support for FloatRanges and non-zero-floats

PrinsFrank:


szepeviktor (approved):

ondrejmirtes (changes requested):
Hi, this is far from complete just by looking at the list of changed files. Can you please elaborate what additional work do you plan to do on this so that I can give you feedback what other things need to be done? Thanks.

PR #2589, opened 2023-08-26
Implement array shapes for `preg_match()` `$matches` by-ref parameter

staabm:
~~I tried solving the shapes with Hoa Regex AST, but I couldn't find the information about capturing groups on the ast nodes.
I don't see a way on how to identify where capturing groups are based on the AST.~~

therefore I tried the way suggested in https://github.com/phpstan/phpstan/issues/9502#issue-1771934831


staabm:
@mvorisek please provide feedback and if you could come up with a few more test-cases, that would be awesome.

whats the expected result for https://github.com/phpstan/phpstan/issues/9502 ?

tests/PHPStan/Analyser/data/preg_match_shapes.php
```diff
@@ -0,0 +1,44 @@
...
+	if (preg_match('/\w-(?P<num>\d+)-(\w)/', $s, $matches)) {
+		assertType('array{0: string, num?: string, 1?: string, 2?: string}', $matches);
+	}
+	assertType('array<string>', $matches);
+}
+
+function doOnlyNmedSubpattern(string $s): void {
+	// n modifier captures only named groups
+	if (preg_match('/(\w)-(?P<num>\d+)-(\w)/n', $s, $matches)) {
+		assertType('array<string>', $matches); // could be "array{0: string, num?: string, 1?: string}"
```
staabm: this case does not work with the trick from https://3v4l.org/sOXbn - therefore added it to show the current result as I had to early exit in the `RegexShapeMatcher` for it

mvorisek: same trick with last, but named, capture group works - https://3v4l.org/3SdDM

src/Type/Php/PregMatchTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	public function setTypeSpecifier(TypeSpecifier $typeSpecifier): void
+	{
+		$this->typeSpecifier = $typeSpecifier;
+	}
+
+	public function isFunctionSupported(FunctionReflection $functionReflection, FuncCall $node, TypeSpecifierContext $context): bool
+	{
+		return in_array(strtolower($functionReflection->getName()), ['preg_match'], true);
```
mvorisek: `preg_match_all` should be supported as well

also handle/test all relevant preg_ `$flags` params like `PREG_OFFSET_CAPTURE`, `PREG_SET_ORDER`, ... which have substantial impact on the result

staabm: lets do `preg_match` first and see what ondrey is thinking about it

staabm:
memo to me: I need to fix `PREG_UNMATCHED_AS_NULL` on PHP <= 7.3 
https://3v4l.org/3k0mr

PrinsFrank:
@staabm It would be really great to have this implemented now `reportPossiblyNonexistentConstantArrayOffset` is available! Can I help here?

staabm:
I have this PR on my todo list and will rebase it in the near future. 

You can help by verifying existing test expectations of this PR or help explore missing tests/cases

src/Type/Php/RegexShapeMatcher.php
```diff
@@ -0,0 +1,111 @@
...
+		if ($flags !== null) {
+			$trickFlags = PREG_UNMATCHED_AS_NULL | $flags;
+		} else {
+			$trickFlags = PREG_UNMATCHED_AS_NULL;
+		}
+
+		// add one capturing group to the end so all capture group keys
+		// are present in the $matches
+		// see https://3v4l.org/sOXbn, https://3v4l.org/3SdDM
+		$regex = preg_replace('~^(.)(.*)\K(\1\w*$)~', '|(?<phpstan_named_capture_group_last>)$3', $regex);
```
gharlan: I think that doesn't work for bracket style delimiters yet, like `'{regex}'`.
See https://www.php.net/manual/en/regexp.reference.delimiters.php.

TIL there can be leading whitespace before the delimiter: `'   /regex/'`

gharlan: Maybe this would work?

```suggestion
		$regex = preg_replace('~.[a-z\s]*$~i', '|(?<phpstan_named_capture_group_last>)$0', $regex);
```

I think it should work for all valid regexes. (And all invalid regexes should remain invalid.)

Note that I avoided to use `\w` for the modifiers at the end, because`\w` includes `_`, which can be used as delimiter.

staabm: @gharlan could you also suggest test-cases for this change?

gharlan: I think these test cases would fail at the moment and I hope they are fixed when using my suggested regex:

```php
	if (preg_match('  /Price: (£|€)\d+/ i u', $s, $matches)) {
		assertType('array{0: string, 1?: string}', $matches);
	}
	assertType('array<string>', $matches);

	if (preg_match('(Price: (.+))i', $s, $matches)) {
		assertType('array{0: string, 1?: string}', $matches);
	}
	assertType('array<string>', $matches);

	if (preg_match('_foo(.)\_i_i', $s, $matches)) {
		assertType('array{0: string, 1?: string}', $matches);
	}
	assertType('array<string>', $matches);
```

staabm: thanks

it does work for the first 2, but not the last example

staabm: (also interessting: with the new pattern, we get different errors per php version)

gharlan: Oh, I guess the `phpstan_named_capture_group_last` name is the problem there, because `_` is the delimiter in that regex. So it would be better to use `phpstanNamedCaptureGroupLast`.

staabm: looks better now. PHP 7.2/7.3 errors remain.

tests/PHPStan/Analyser/data/preg_match_shapes.php
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function doOffsetCaptureWithUnmatchedNull(string $s): void {
+	// see https://3v4l.org/07rBO#v8.2.9
+	if (preg_match('/(foo)(bar)(baz)/', $s, $matches, PREG_OFFSET_CAPTURE|PREG_UNMATCHED_AS_NULL)) {
+		assertType('array{0: array{null, -1}|array{string, int<0, max>}, 1?: array{null, -1}|array{string, int<0, max>}, 2?: array{null, -1}|array{string, int<0, max>}, 3?: array{null, -1}|array{string, int<0, max>}}', $matches);
+	}
+	assertType('array<array{null, -1}|array{string, int<0, max>}>', $matches);
+}
+
```
PrinsFrank: It is possible to have multiple groups with the same name, but only if the modifier '(?J)' is added, otherwise the result will always be null;

```suggestion

function doMultipleCaptureGroupsWithSameNameWithoutModifier(string $s): void {
    preg_match('/(?<Foo>[a-z]+)|(?<Foo>[0-9]+)/', $s, $matches);
    assertType('null', $matches);
}

function doMultipleAlternativeCaptureGroupsWithSameNameWithModifier(string $s): void {
    if (preg_match('/(?J)(?<Foo>[a-z]+)|(?<Foo>[0-9]+)/', $s, $matches)) {
        assertType('array{0: string, Foo: string, 1: string}', $matches);
    }
    assertType('array{}', $matches);
}

function doMultipleConsecutiveCaptureGroupsWithSameNameWithModifier(string $s): void {
    if (preg_match('/(?J)(?<Foo>[a-z]+)|(?<Foo>[0-9]+)/', $s, $matches)) {
        assertType('array{0: string, Foo: string, 1: string, 2: string}', $matches);
    }
    assertType('array{}', $matches);
}

```

Capture groups with the same name can be in two alternatives, but can also be consecutive, so this complicates things a bit.

tests/PHPStan/Analyser/data/preg_match_shapes.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+namespace PregMatchShapes;
+
+use function PHPStan\Testing\assertType;
+
```
PrinsFrank: When the return value of preg_match is false, the type of $matches is always null:

```suggestion

function invalid(string $s): void
{
    preg_match('', $s, $matches);

    assertType('null', $matches);
}

```

tests/PHPStan/Analyser/data/preg_match_shapes.php
```diff
@@ -0,0 +1,80 @@
...
+	assertType('array<string>', $matches);
+}
+
+function doNonCapturingGroup(string $s): void {
+	if (preg_match('/Price: (?:£|€)(\d+)/', $s, $matches)) {
+		assertType('array{0: string, 1?: string}', $matches);
+	}
+	assertType('array<string>', $matches);
+}
+
```
PrinsFrank: It is possible to turn off auto capturing of non named groups with the 'n' flag and modifier:

```suggestion

function doNonAutoCapturingFlag(string $s): void {
    if (preg_match('/(\d+)/n', $s)) {
        assertType('array{0: string}', $s);
    }

    assertType('array{}', $s);
}

function doNonAutoCapturingModifier(string $s): void {
    if (preg_match('/(?n)(\d+)/', $s)) {
        assertType('array{0: string}', $s);
    }

    assertType('array{}', $s);
}

```

tests/PHPStan/Analyser/data/preg_match_shapes.php
```diff
@@ -0,0 +1,80 @@
...
+		assertType('array{0: array{null, -1}|array{string, int<0, max>}, 1?: array{null, -1}|array{string, int<0, max>}, 2?: array{null, -1}|array{string, int<0, max>}, 3?: array{null, -1}|array{string, int<0, max>}}', $matches);
+	}
+	assertType('array<array{null, -1}|array{string, int<0, max>}>', $matches);
+}
+
+function doUnknownFlags(string $s, int $flags): void {
+	if (preg_match('/(foo)(bar)(baz)/', 'foobarbaz', $matches, $flags)) {
+		assertType('array<array{string|null, int<-1, max>}|string|null>', $matches);
+	}
+	assertType('array<array{string|null, int<-1, max>}|string|null>', $matches);
```
PrinsFrank: This turns into an empty array, and I think that is the case for all non-matches that aren't the result of a regex parse error, right?
```suggestion
	assertType('array{}', $matches);
```

PrinsFrank (changes requested):

tests/PHPStan/Analyser/data/preg_match_shapes.php
staabm: you are right as in the expectation is wrong. but the reason is, that the test-case contains a leftover typo from debugging. 

its not really interessting how the types behave when the preg-pattern and the subject is known at analysis time. thats super unlikely.

the interessting point is, that the type of `$matches` after the IF statement can still contain the matched type.

https://3v4l.org/gWuNV#v8.2.9

tests/PHPStan/Analyser/data/preg_match_shapes.php
staabm: adjusted and added the test, thanks

tests/PHPStan/Analyser/data/preg_match_shapes.php
staabm: adjusted and added, thanks

tests/PHPStan/Analyser/data/preg_match_shapes.php
staabm: these `null` expecting cases are interessting, but I feel out of scope of this PR.

the case of explicitly handling a possible syntax error in the pattern in application code usually does not happen.
even before this PR phpstan atm does not make `$matches` return null

src/Type/Php/RegexShapeMatcher.php
```diff
@@ -0,0 +1,111 @@
...
+			return new ArrayType(new MixedType(), new StringType());
+		}
+		unset($matches[array_key_last($matches)]);
+		unset($matches['phpstanNamedCaptureGroupLast']);
+
+		$builder = ConstantArrayTypeBuilder::createEmpty();
+		foreach (array_keys($matches) as $key) {
+			// atm we can't differentiate optional from mandatory groups based on the pattern.
+			// So we assume all are optional
+			$optional = true;
```
staabm: anyone has ideas on how we could identify which groups are optional and which not?

with the current PRs implementation most offsets are treated as optional, which makes the inferred type pretty weak. see e.g. https://github.com/phpstan/phpstan-src/pull/2589/files#diff-8a0ca036dacc90b9ccceaa1a9053a14f13a279e04687c683397085b0a3b7c573R9

mvorisek: I think you will have to parse the full regex.

staabm: @gharlan just pointed me to https://github.com/hoaproject/Regex/issues/38#issue-386915806 which seems to solve the problem in Hoa

mvorisek: Great finding! The fix seems correct based on https://github.com/hoaproject/Compiler/blob/b675e53267/README.md?plain=1#L134.

staabm:
In the end I combined both approaches. regex based retrieval of the capturing groups and the AST to identify how many non-optional groups are contained in the regex.

since the type inference now requires a regex which is parsable by Hoa\Regex I commented some type-assertions which currently don't work because the regex parser cannot parse it. I think these can be fixed in the future by applying fixes to the `Grammar.pp`.

I am neither a regex nor a grammer expert, therefore I left these things for other people more confident in this topic.
overall this means we get already a pretty good type narrowing based on a constant regex.


----

In a future PR I would improve the types to be even more precise for e.g.
`'/Price: (£|€)/i'` -> constant string £|€ instead of just string
`'/Price: (\d)/i'` -> numeric-string instead of string
`'/Price: (\s){2,10}/i'` -> non-falsey-string instead of string
...

src/Type/Php/RegexShapeMatcher.php
```diff
@@ -0,0 +1,176 @@
...
+			return $builder->getArray();
+		}
+
+		return $valueType;
+	}
+
+	private function countNonOptionalGroups(string $regex): ?int
+	{
+		/** @throws void */
+		$parser = Llk::load(new Read('hoa://Library/Regex/Grammar.pp'));
```
staabm: should this be moved into a external class, so the grammar is not read over and over?

tests/PHPStan/Analyser/data/preg_match_shapes.php
```diff
@@ -0,0 +1,118 @@
...
+function doNonCapturingGroup(string $s): void {
+	if (preg_match('/Price: (?:£|€)(\d+)/', $s, $matches)) {
+		assertType('array{string, string}', $matches);
+	}
+	assertType('array<string>', $matches);
+}
+
+function doNamedSubpattern(string $s): void {
+	if (preg_match('/\w-(?P<num>\d+)-(\w)/', $s, $matches)) {
+		// could be assertType('array{0: string, num: string, 1: string, 2: string, 3: string}', $matches);
```
staabm: here we can see one example of a regex which is not parsable with the current hoa grammar:

```
Fatal error: Uncaught Hoa\Compiler\Llk\Parser::parse(): (0) Unexpected token "-" (range) at line 1 and column 4:
/\w-(?P<num>\d+)-(\w)/
   ↑
in /Users/staabm/workspace/phpstan-src/vendor/hoa/compiler/Llk/Parser.php at line 1.
  thrown in /Users/staabm/workspace/phpstan-src/vendor/hoa/compiler/Llk/Parser.php on line 1
```

patches/Grammar.patch
```diff
@@ -0,0 +1,29 @@
+--- Grammar.pp  2024-05-18 12:15:53
++++ Grammar.pp.fix      2024-05-18 12:15:05
+@@ -109,7 +109,7 @@
+ // Please, see PCRESYNTAX(3), General Category properties, PCRE special category
+ // properties and script names for \p{} and \P{}.
+ %token character_type            \\([CdDhHNRsSvVwWX]|[pP]{[^}]+})
+-%token anchor                    \\(bBAZzG)|\^|\$
++%token anchor                    \\([bBAZzG])|\^|\$
```
staabm, line 8: unreleased hoa/regex upstream fix
https://github.com/hoaproject/Regex/commit/5f670af756d54d0d79565742cbfe03ad3ba8bdb2

patches/Grammar.patch
```diff
@@ -0,0 +1,29 @@
...
+ %token match_point_reset         \\K
+ %token literal                   \\.|.
+
+@@ -168,7 +168,7 @@
+         ::negative_class_:: #negativeclass
+       | ::class_::
+     )
+-    ( range() | literal() )+
++    ( <class_> | range() | literal() )+ <range>?
```
staabm: unreleased hoa/regex upstream fix
https://github.com/hoaproject/Regex/commit/ce7fd7b5a26bdeeee7c8385fd7cabd5686e9af37
https://github.com/hoaproject/Regex/commit/e770ada2e9f52b86c8c52f5eb12f3c4b3f9dbd0f

src/Type/Php/RegexShapeMatcher.php
mvorisek: Load once into static property?

patches/Grammar.patch
```diff
@@ -0,0 +1,29 @@
...
+     ::_class::
+
+ #range:
+@@ -178,7 +178,7 @@
+     capturing()
+   | literal()
+
+-capturing:
++#capturing:
```
staabm, line 26: upstream hoa fix
https://github.com/hoaproject/Regex/issues/38#issuecomment-443788516

tests/PHPStan/Analyser/data/preg_match_shapes.php
mvorisek: There are two issues:

a) `(?P<xxx>...` is equivalent of supported `(?<xxx>...` - spec https://www.pcre.org/original/doc/html/pcrepattern.html (search for ˙?p<˙), grammar: https://github.com/hoaproject/Regex/blob/master/Source/Grammar.pp#L73
b) `-` range - it is reserved character strictly if in `[` character group only (otherwise it is regular chanracter and very often used)

staabm: Thanks. Please provide code suggestions for fixes. 

[the example work in plain php](https://3v4l.org/VWvRk), so we either need to fix the grammar or leave the typing for these cases behind

mvorisek: Would you mind forking `hoaproject/Regex` and saving there a php test file I can run with all the issues you found?

staabm: look into this PR here and find all commented assertions. these are the cases which hoa cannot parse.

I have spent the last the 3 days of freetime for your feature request. take your time to do the tests yourself. Thanks

Seldaek:
This looks amazing, thanks @staabm. 

Now I'm just wondering how I can get it to recognize Composer\Pcre\Preg::match and co as well to get the same quality of type output there :s

If you can think of some way to make this more reusable by third party libs wrapping preg functions it'd be great.

staabm:
> Now I'm just wondering how I can get it to recognize Composer\Pcre\Preg::match and co as well to get the same quality of type output there :s
> 
> If you can think of some way to make this more reusable by third party libs wrapping preg functions it'd be great.

at first I want to see what @ondrejmirtes thinks about all this. he was not yet involved. 

After that: I can think of different ways to make this re-usable. for wrapper-libs which are 100% api compatible with the pcre_* native functions we could e.g. 
- rewrite the AST at analysis time
- create separate extensions which re-use the underlying implementation
- make it configurable via NEON config
- introduce a new phpdoc type
- ...

Seldaek:
Ok cool thanks. Yes composer/pcre is almost one to one.. Except it throws instead of returning false. For the other api returning object results there it's a bit more complex but anyway I see your point and I'll let you wrap things up here first. Just wanted to let you know in case it can influence design decisions along the way but it's not urgent at all.

tests/PHPStan/Analyser/data/preg_match_shapes.php
```diff
@@ -0,0 +1,161 @@
...
+namespace PregMatchShapes;
+
+use function PHPStan\Testing\assertType;
+
+
+function doMatch(string $s): void {
+	if (preg_match('/Price: /i', $s, $matches)) {
+		assertType('array{string}', $matches);
+	}
+	assertType('array<string>', $matches);
```
mvorisek:
```suggestion
	} else {
		assertType('array{}', $matches);
	}
	assertType('array{string}', $matches);
```

when pattern is not matched, the `$matches` is reset to `array{}` - https://3v4l.org/NE0bm

src/Type/Php/RegexShapeMatcher.php
```diff
@@ -0,0 +1,181 @@
...
+	public function matchType(string $regex, ?int $flags, TypeSpecifierContext $context): ?Type
+	{
+		// add one capturing group to the end so all capture group keys
+		// are present in the $matches
+		// see https://3v4l.org/sOXbn, https://3v4l.org/3SdDM
+		$captureGroupsRegex = preg_replace('~.[a-z\s]*$~i', '|(?<phpstanNamedCaptureGroupLast>)$0', $regex);
+
+		if (
+			$captureGroupsRegex === null
+			|| @preg_match($captureGroupsRegex, '', $matches, PREG_UNMATCHED_AS_NULL) === false
```
staabm: Use nette

ondrejmirtes:
Some initial thoughts:

* This should be only for bleeding edge. There is a big potential for bugs so we should limit the impact.
* There are more preg_ functions that would benefit from the implemented logic, right?
* * There's some regex syntax (`\Q...\E`) the Hoa library fails spectacularly with. We should test it in this PR: https://github.com/phpstan/phpstan/issues/10855#issuecomment-2044323638
* I don't like bending TypeSpecifyingExtension for this. This should either be hardcoded in NodeScopeResolver, or there should be some kind of new ParameterOutTypeExtension. Probably the latter, given we want to make this accessible to userland code.
* There are different flavours of custom PCRE wrappers. Some return matches by-ref as the original functions, some return the matches normally via returned value. The list of our options in https://github.com/phpstan/phpstan-src/pull/2589#issuecomment-2118797935 is comprehensive, I like "create separate extensions which re-use the underlying implementation" the most. This means that the actual implementation should be in an extension-agnostic service that other extensions can simply inject. This means the service has to be marked with `@api`.

tests/PHPStan/Analyser/data/preg_match_shapes.php
mvorisek: fix is here: https://github.com/mvorisek/Hoa-Regex/commit/0c8a5cbcf696df8f65168704af14ad5ebd48c641

and https://github.com/hoaproject/Regex/commit/e770ada2e9f52b86c8c52f5eb12f3c4b3f9dbd0f is not needed then :)

patches/Grammar.patch
mvorisek: 2nd patch was reverted, see https://github.com/phpstan/phpstan-src/pull/2589#discussion_r1605809712

src/Type/Php/RegexShapeMatcher.php
```diff
@@ -0,0 +1,181 @@
...
+		) {
+			return null;
+		}
+		unset($matches[array_key_last($matches)]);
+		unset($matches['phpstanNamedCaptureGroupLast']);
+
+		$remainingNonOptionalGroupCount = $this->countNonOptionalGroups($regex);
+		if ($remainingNonOptionalGroupCount === null) {
+			// regex could not be parsed by Hoa/Regex
```
mvorisek: I would prefer emitting a PHPStan error with the parse exception message - the PCRE grammar is actually not so complex, so we can fix the Hoa grammar based on issue reports.

This would effectively also validate constant regexps for syntax errors.

The https://3v4l.org/sOXbn parsing hack can be then removed in favor of Hoa parser.

tests/PHPStan/Analyser/data/preg_match_shapes.php
staabm: I tried your suggestion, but it did not fix any of the commented assertions.

parsing `'/\w-(?P<num>\d+)-(\w)/'` I still get an exception:
(but its a different one as in the 1st comment)

```
PHP Fatal error:  Uncaught Hoa\Compiler\Llk\Parser::parse(): (0) Unexpected token "?" (zero_or_one) at line 1 and column 6:
/\w-(?P<num>\d+)-(\w)/
     ↑
in /Users/staabm/workspace/phpstan-src/vendor/hoa/compiler/Llk/Parser.php at line 1.
  thrown in /Users/staabm/workspace/phpstan-src/vendor/hoa/compiler/Llk/Parser.php on line 1

```

ondrejmirtes:
An idea: send a separate PR with DynamicParameterOutType extensions, and once that is merged, you can base this PR on that.

tests/PHPStan/Analyser/data/preg_match_shapes.php
mvorisek: It is because of the `(?P<xxx>...)` part as explained above - the `-` in non-class is fixed. It should be present as a separate case.

staabm: ohh I see. I got the impression you are working on all these cases.

mvorisek: I can look into more cases later. This one was the most important. What about adding proper Hoa Regex unit testing to phpstan-src? We can assert the parsed regexes using https://github.com/mvorisek/Hoa-Regex/blob/0c8a5cbcf696df8f65168704af14ad5ebd48c641/test.php#L13-L25 function.

staabm: Sure, why not. Please send a PR to my branch/fork

mvorisek: I am not super familiar with phpstan-src codebase - where, what class can I base this unit test on?

staabm: I would create a new file in `tests/PHPStan/Type`. we can later on move it somewhere else if this doesn't work for ondrej.

staabm:
I don't know why the composer.lock is wrong as we can see here (in the CI jobs). I tried regenerating it, but I did not succeed.

on my local machine the PR is running the `make` and all the checks without any errors

---

this thing aside: I really like how the PR currently looks like. since I have implemented a `PregMatchParameterOutTypeExtension` I can just use the regular `PregMatchTypeSpecifyingExtension` to narrow the optional array offsets to the non-optional ones (so yes, I use both extensions in tandem). 

I am really happy with the type narrowing achieved with the current state

ondrejmirtes:
How are you updating the hash in composer.lock? You should have the same composer.lock as in 1.11.x and then run `composer update --lock`.

ondrejmirtes:
I'm not surprised Composer validation fails like it fails today. Huge chunks of the lock file are changed how they shouldn't be changed.

conf/config.neon
```diff
@@ -1465,6 +1466,23 @@ services:
 		tags:
 			- phpstan.dynamicFunctionThrowTypeExtension
 
+	-
+		class: PHPStan\Type\Php\PregMatchTypeSpecifyingExtension
+		arguments:
+			disabled: %featureToggles.narrowPregMatches%
```
ondrejmirtes: This extension does not need "disabled" parameter. Check out the "conditionalTags" sections.

conf/config.neon
```diff
@@ -1465,6 +1466,23 @@ services:
...
+		tags:
+			- phpstan.typeSpecifier.functionTypeSpecifyingExtension
+
+	-
+		class: PHPStan\Type\Php\PregMatchParameterOutTypeExtension
+		arguments:
+			disabled: %featureToggles.narrowPregMatches%
```
ondrejmirtes: dtto

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -0,0 +1,233 @@
...
+	)
+	{
+	}
+
+	public function matchType(Type $patternType, ?Type $flagsType, TrinaryLogic $wasMatched): ?Type
+	{
+		if (
+			!$this->phpVersion->returnsPregUnmatchedCapturingGroups()
+		) {
+			return null;
```
ondrejmirtes, line 55: We can't return anything useful on 7.2 and 7.3 at all?

ondrejmirtes (changes requested):

src/Type/Php/RegexArrayShapeMatcher.php
staabm: I have invested so much already in this PR, that I had not yet looked into 7.2/7.3 specifics. I am happy getting it running for the more modern stuff for now :)

conf/config.neon
staabm: I tried that with conditionaltags, but it did not work for `phpstan.functionParameterOutTypeExtension`.

staabm:
> I'm not surprised Composer validation fails like it fails today. Huge chunks of the lock file are changed how they shouldn't be changed.

I had the same feeling. the huge chunk was generated by `composer update --lock`.

I was able to fix it now, but updating composer to the latest version and doing the same steps I did before.

conf/config.neon
staabm:
```
	PHPStan\Type\Php\PregMatchParameterOutTypeExtension:
		phpstan.functionParameterOutTypeExtension: %featureToggles.narrowPregMatches%
```

leads to

```
In Compiler.php line 293:
                                                                                                                                        
  Unexpected item 'conditionalTags › PHPStan\Type\Php\PregMatchParameterOutTypeExtension › phpstan.functionParameterOutTypeExtension'.  
                                                                                                                                        
```

ondrejmirtes: Whenever we add new tags, they need to be added to the ConditionalTagsExtension schema as well.

staabm: thanks, thats it.

https://github.com/phpstan/phpstan-src/pull/3173

src/Type/Php/RegexArrayShapeMatcher.php
staabm: the array-shape matcher magic depends on named capturing groups.
in 7.2/7.3 these are not supported, so we can't provide a better array-shape for these php versions.

staabm:
atm we don't support array-shape inference on e.g.

```php
if (preg_match('~<?php\\s*\\/\\/\s*lint\s*([^\d\s]+)\s*([^\s]+)\s*~i', $firstLine, $m) === 1) {
	return version_compare(PHP_VERSION, $m[2], $m[1]) === false;
}
```

because we don't have a true-context because of the comparison with `=== 1`.
wdyt about rewriting a `if (preg_match(...) === 1)` AST to a `if (preg_match(...))` at analysis time, so these can also be supported?

tests/PHPStan/Analyser/data/preg_match_shapes.php
staabm: fixed

src/Type/Php/RegexArrayShapeMatcher.php
staabm: just added emtpy-array negative context inference, which also works for php 7

ondrejmirtes:
Hey, I'd like to merge this today. Before releasing it, I will write an extension for phpstan/phpstan-nette (`Nette\Utils\Strings` wraps preg functions).

I'd like you to write an extension for `composer/pcre`. It means that the library should probably ship with extension.neon that people can include in their own projects. `composer/composer` already does it: https://github.com/composer/composer/blob/11e5237ad9d9e8f29bdc57d946f87c816320d863/composer.json#L72

Although it sucks slightly we can't really enforce PHPStan version constraints there but hopefully it won't be much of an issue.

I have access to private projects that extensively use both of these. So I want to test this before creating a release.

There are also these additional wrappers that we might write extensions for in the future:

* https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/src/Preg.php
* https://github.com/thecodingmachine/safe
* https://packagist.org/packages/spatie/regex

tests/PHPStan/Analyser/data/preg_match_shapes.php
staabm: @PrinsFrank please open a new separate phpstan issue for these cases with a snippet and expectations

staabm:
:+1: 

how to require the non-merged phpstan-src PR in the external repo, so I can work with the unmerged api?

ondrejmirtes:
Wait after this is merged, then require `1.11.x-dev`.

ondrejmirtes:
Please fix the build and mark it as ready.

staabm:
I am not sure why we have these 7.4/7.2 remaining errors. are you fine with merging anyway?

staabm:
> I am not sure why we have these 7.4/7.2 remaining errors. are you fine with merging anyway?

made the expectations more php version specific so we can move on and get a green build.

we see some PHPStan extension builds failing because of the problem mentioned in https://github.com/phpstan/phpstan-src/pull/2589#issuecomment-2181186580

ondrejmirtes:
What about the errors here? https://github.com/phpstan/phpstan-src/actions/runs/9610380632

Like: https://github.com/phpstan/phpstan-symfony/blob/bca27f1701fc1a297749e6c2a1e3da4462c1a6af/src/Type/Symfony/ParameterDynamicReturnTypeExtension.php#L185-L188

I know you mentioned `=== 1` not being able to handle that in the type-specifying extension.

I don't know whether you solved that, based on the results probably not. My opinion is that we could hardcode this specifically in TypeSpecifier, so that extension is called correctly with the right context.

staabm:
> I don't know whether you solved that, based on the results probably not. My opinion is that we could hardcode this specifically in TypeSpecifier, so that extension is called correctly with the right context.

if you have an idea how this need to look like, please go ahead and implement it.

the only idea I have atm would be to add a AST visitor which detects a `if (preg_match(...) === 1)` and sets a attribute which later on can be detected by the TypeSpecifier.

hopefully you have something better in mind :)

mvorisek:
What exactly is the problem, why we cannot check if the context/preg_match result is surely non-falsy?

ondrejmirtes:
@staabm I meant this: https://github.com/phpstan/phpstan-src/pull/2589/commits/328fcb91de1605c1a578e45d72f20952b3312501

staabm:
I see, thanks. should we have the same for `==`?

ondrejmirtes:
@staabm Yeah, it's probably harmless, but do that in a next PR, not here :)

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

ondrejmirtes:
An extension can be as simple as this https://github.com/phpstan/phpstan-nette/commit/3e68a5d7f0be96bb97b3a8770391b23dfdc07c08

staabm:
the extension for `composer/pcre` is beeing worked on in https://github.com/composer/pcre/pull/24

westonruter:
I just noticed that this doesn't seem to play well with strict rules. 

This example without strict rules works: https://phpstan.org/r/6187bcca-8fc8-459b-9790-21a82e62b345

But when I enable strict rules, I need to cast the return value of `preg_match()` to a `bool` and this seems to block the types from passing through. For example: https://phpstan.org/r/f368fc07-6438-43f1-b80b-93dee22535f4

Same thing happens when I cast to an `int`: https://phpstan.org/r/0d011355-4446-4bd8-9b8d-0c44907f9a07

ondrejmirtes:
@westonruter Please open a new bug report about this.

westonruter:
Done: https://github.com/phpstan/phpstan/issues/11262

staabm:
extensions implemented for PHP-CS-Fixer `Preg::match` in https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/8103

staabm:
initial support for `composer/pcre` 2.x landed in https://github.com/composer/pcre/pull/25
coverage for more methods is beeing worked on in separate PRs

PR #2591, opened 2023-08-28
Fix type inference of `array_sum()`

zer0-star:
fix https://github.com/phpstan/phpstan/issues/9808

### DIFF
--- a/src/Type/Php/ArraySumFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ArraySumFunctionDynamicReturnTypeExtension.php
@@ -44,8 +44,12 @@ final class ArraySumFunctionDynamicReturnTypeExtension implements DynamicFunctio
 			if ($constantArray !== null) {
 				$node = new LNumber(0);
 
-				foreach ($constantArray->getValueTypes() as $type) {
-					$node = new Plus($node, new TypeExpr($type));
+				foreach ($constantArray->getValueTypes() as $i => $type) {
+					if ($constantArray->isOptionalKey($i)) {
+						$node = new Plus($node, new TypeExpr(TypeCombinator::union($type, new ConstantIntegerType(0))));
+					} else {
+						$node = new Plus($node, new TypeExpr($type));
+					}
 				}
 
 				$newItemType = $scope->getType($node);
--- a/src/Type/Php/ArraySumFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ArraySumFunctionDynamicReturnTypeExtension.php
@@ -15,6 +15,7 @@ use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\FloatType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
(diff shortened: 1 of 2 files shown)


zer0-star:
@ondrejmirtes Could you review this PR?

src/Type/Php/ArraySumFunctionDynamicReturnTypeExtension.php
```diff
@@ -35,21 +45,85 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			$nonScalarTypes = [];
+			$scalarValues = [];
 
-		if ($arrayType->isIterableAtLeastOnce()->yes()) {
-			if ($intUnionFloat->isSuperTypeOf($itemType)->yes()) {
-				return $itemType;
+			$types = $constantArray->getValueTypes();
+
+			foreach ($types as $type) {
+				$scalar = $type->getConstantScalarValues();
```
ondrejmirtes: Hi, I don't really love that this PR is working with specific scalar values and calls `array_sum` internally. It misses out on various properties of the type system, like working with IntegerRangeType.

I'd much rather be if for ConstantArrayType objects, like `[$one, $two, $three]`, it tried to do `$scope->getType(new Plus($one, new Plus($two, $three)))`. Because Plus expr handling in MutatingScope already knows how to work with al possible Types, we can just call it here like that.

Because ConstantArrayType contains Type and not Expr, you can take advantage of TypeExpr virtual node to build an Expr out of Type instances.

ondrejmirtes (changes requested):

ondrejmirtes (changes requested):
Yes, that's the idea :) And we don't have to limit ourselves to a single constant array, we can union results from all constant arrays. For example when something like `array{1}|array{2, 3}` is passed in.

zer0-star:
Thank you very much for your review! I will continue working it in tomorrow.

zer0-star:
@ondrejmirtes
Hello, can I ask you a question?
In the test case `foo4`, the inferred type is `BenevolentUnionType`, so the result becomes `(float|int)`. Do you think it's needed to fix this?

ondrejmirtes:
@zer0-star No, that's totally fine and actually a nice side-effect of the changes :)

zer0-star:
@ondrejmirtes
I think this PR is ready to be reviewed! I'm sorry to rush you, but could you review it for me?

ondrejmirtes (changes requested):
Hi, I really appreciate this, especially the thorough test suite :) I wanted to write a simplification and it helped me a lot to get it right.

I also pushed a failing test case - ConstantArrayType might have an optional key. This is how to check for that:

```php
foreach ($constantArray->getValueTypes() as $i => $type) {
    if ($constantArray->isOptionalKey($i)) { ... }
}
```

zer0-star:
Thanks so much for your help! I really appreciate it.
I will fix this problem later.

Is there something else to revise?

ondrejmirtes:
No, once the optional keys are supported, we are good to merge this :)

zer0-star:
I fixed it! I hope you like this change :hugs:

ondrejmirtes:
I don't really like any `instanceof *Type`, it's usually a code smell. Explained here: https://phpstan.org/blog/why-is-instanceof-type-wrong-and-getting-deprecated

There should be nothing special about `MixedType` and there were probably other types that needed a similar treatment. I was also able to get rid of other problems by doing `->toNumber()` on the resulting type. See: https://github.com/phpstan/phpstan-src/commit/291bb57979b3238d69b77c84d46fe6c47bd3d2b2?w=1

I'll merge the PR if the build gets green :) Thank you very much for your cooperation!

ondrejmirtes (approved):

zer0-star:
Thanks! I was happy to contribute to PHPStan.

I learned a lot from your review!

PR #2592, opened 2023-08-28
micro: use spl_object_id()

staabm:
`spl_object_id()` exists since php 7.2, I think we can use it over `spl_object_id()`.

rector and wordpress [micro benchmarks suggests](https://github.com/rectorphp/rector-src/pull/4876) its faster


samsonasik (approved):

ondrejmirtes:
Thank you.

PR #2594, opened 2023-08-30
micro: prevent unnecessary IO in PhpVersionFactoryFactory

staabm:
the `is_file` showed up in my blackfire profiles

grafik

### DIFF
--- a/src/Php/PhpVersionFactoryFactory.php
+++ b/src/Php/PhpVersionFactoryFactory.php
@@ -8,7 +8,6 @@ use PHPStan\File\CouldNotReadFileException;
 use PHPStan\File\FileReader;
 use function count;
 use function end;
-use function is_file;
 use function is_string;
 
 class PhpVersionFactoryFactory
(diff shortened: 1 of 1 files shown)


src/Php/PhpVersionFactoryFactory.php
```diff
@@ -29,17 +28,15 @@ public function create(): PhpVersionFactory
...
-				} catch (CouldNotReadFileException | JsonException) {
-					// pass
+			try {
+				$composerJsonContents = FileReader::read($composerJsonPath);
+				$composer = Json::decode($composerJsonContents, Json::FORCE_ARRAY);
+				$platformVersion = $composer['config']['platform']['php'] ?? null;
+				if (is_string($platformVersion)) {
+					$composerPhpVersion = $platformVersion;
 				}
+			} catch (CouldNotReadFileException | JsonException) {
```
staabm, line 38: we already handle the non-readable case via the try/catch, no need for the additional IO before calling into `FileReader::read`

PR #2595, opened 2023-08-30
Failing test for broken implicitThrows (Throwable returned)

janedbal:
I believe this behaviour is not correct, see failing test. The [condition that breaks it](https://github.com/phpstan/phpstan-src/blob/b551135ac79a532309675a3ac28fdae9d3fbb04e/src/Analyser/NodeScopeResolver.php#L3113).


ondrejmirtes:
Would be nice if this had an open bug report, otherwise it's gonna get lost.

janedbal:
I was hoping you tell me just remove that nonsense condition here :) I can create also an issue if you want, so should I?

janedbal:
Bugreport: https://github.com/phpstan/phpstan/issues/9826

PR #2598, opened 2023-09-01
MethodCall: mark virtual nullsafe call with attribute

janedbal:
Resolves underlying problem of https://github.com/phpstan/phpstan/issues/9830


ondrejmirtes:
Thank you!

PR #2599, opened 2023-09-03
More precise inotify related function signatures

thg2k:
# Changes

1) `inotify_add_watch`: Narrowed the returned value: It can only be a non-negative integer according to the [man page](https://linux.die.net/man/2/inotify_add_watch)
2) `inotify_add_watch`: Added `false` in case of failure, same concept as `fopen()`, in case the file does not exist or is not readable
3) `inotify_queue_len`: Narrowed the return value: It can only be zero or positive
4) `inotify_read`: Narrowed the return value: Added the shape of the returned array
5) `inotify_read`: Added `false` in case the stream is set a non-blocking.

**Question:** I have a doubt about point 5, i.e. adding `false` as possible return type to `inotify_read`. This can only happen if the inotify stream is set as non-blocking (tested). Should this go to the bleeding version of the signatures map?


ondrejmirtes:
Let's try this like that :) Thank you!

taka-oyama:
Hi, our CI started failing with the following on `inotify_read` after `1.10.33` and I'm assuming this is the culprit.

```php
Offset 'wd' does not exist on int<0, max>|string.    
Offset 'mask' does not exist on int<0, max>|string.  
Offset 'name' does not exist on int<0, max>|string. 
```

This is the code snippet where the error occurred.

```php
foreach (inotify_read($inotify) as $event) {
    $id = $event['wd'];
    $mask = $event['mask'];
    $name = $event['name'] ?? '';
    ...
}
```

PR #2603, opened 2023-09-04
Show custom tip for unfound Drupal symbols

bkline:
- add optional $className argument to discoveringSymbolsTip() method
- if that name starts with 'Drupal/' show link to phpstan-drupal project
- pass the symbol name to all 22 calls to discoveringSymbolsTip()

Closes #9837

### DIFF
--- a/src/Rules/Functions/CallToNonExistentFunctionRule.php
+++ b/src/Rules/Functions/CallToNonExistentFunctionRule.php
@@ -41,7 +41,7 @@ class CallToNonExistentFunctionRule implements Rule
 			}
 
 			return [
-				RuleErrorBuilder::message(sprintf('Function %s not found.', (string) $node->name))->identifier('function.notFound')->discoveringSymbolsTip()->build(),
+				RuleErrorBuilder::message(sprintf('Function %s not found.', (string) $node->name))->identifier('function.notFound')->discoveringSymbolsTip((string) $node->name)->build(),
 			];
 		}
 
--- a/src/Rules/Namespaces/ExistingNamesInGroupUseRule.php
+++ b/src/Rules/Namespaces/ExistingNamesInGroupUseRule.php
@@ -73,7 +73,7 @@ class ExistingNamesInGroupUseRule implements Rule
 	{
 		if (!$this->reflectionProvider->hasConstant($name, null)) {
 			return RuleErrorBuilder::message(sprintf('Used constant %s not found.', (string) $name))
-				->discoveringSymbolsTip()
+				->discoveringSymbolsTip((string) $name)
 				->line($name->getLine())
 				->identifier('constant.notFound')
 				->build();
@@ -86,7 +86,7 @@ class ExistingNamesInGroupUseRule implements Rule
 	{
 		if (!$this->reflectionProvider->hasFunction($name, null)) {
 			return RuleErrorBuilder::message(sprintf('Used function %s not found.', (string) $name))
-				->discoveringSymbolsTip()
+				->discoveringSymbolsTip((string) $name)
 				->line($name->getLine())
 				->identifier('function.notFound')
 				->build();
--- a/src/Rules/RuleErrorBuilder.php
+++ b/src/Rules/RuleErrorBuilder.php
@@ -10,6 +10,7 @@ use function count;
 use function implode;
 use function is_file;
 use function sprintf;
+use function substr;
 
 /**
  * @api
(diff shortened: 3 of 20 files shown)


src/Rules/Functions/CallToNonExistentFunctionRule.php
```diff
@@ -41,7 +41,7 @@ public function processNode(Node $node, Scope $scope): array
 			}
 
 			return [
-				RuleErrorBuilder::message(sprintf('Function %s not found.', (string) $node->name))->identifier('function.notFound')->discoveringSymbolsTip()->build(),
+				RuleErrorBuilder::message(sprintf('Function %s not found.', (string) $node->name))->identifier('function.notFound')->discoveringSymbolsTip((string) $node->name)->build(),
```
ondrejmirtes, line 44: Would be nice to do `(string) $node->name)` only once.

src/Rules/Namespaces/ExistingNamesInGroupUseRule.php
```diff
@@ -73,7 +73,7 @@ private function checkConstant(Node\Name $name): ?IdentifierRuleError
 	{
 		if (!$this->reflectionProvider->hasConstant($name, null)) {
 			return RuleErrorBuilder::message(sprintf('Used constant %s not found.', (string) $name))
-				->discoveringSymbolsTip()
+				->discoveringSymbolsTip((string) $name)
```
ondrejmirtes, line 76: Please remove the passed argument in cases where it's not a class name.

src/Rules/RuleErrorBuilder.php
```diff
@@ -182,8 +183,11 @@ public function addTip(string $tip): self
 	 * @phpstan-this-out self<T&TipRuleError>
 	 * @return self<T&TipRuleError>
 	 */
-	public function discoveringSymbolsTip(): self
+	public function discoveringSymbolsTip(?string $className = null): self
 	{
+		if (isset($className) && substr($className, 0, 7) === 'Drupal/') {
```
ondrejmirtes, line 188: A test would be nice, I'm pretty sure this code is wrong (It's `Drupal\`, not `Drupal/`) :) And please use `str_starts_with`.

The test should be written for one of the impacted rules, it doesn't matter which one. An error tip is tested with a third array item (https://phpstan.org/developing-extensions/rules#testing-the-rule) - besides the message and the line.

ondrejmirtes (changes requested):
And please rebase on top of 1.10.x, as the bot says.

src/Rules/Namespaces/ExistingNamesInGroupUseRule.php
bkline: So don't do what you asked me to do in the issue?

> Yes, all 22 calls should be updated.

src/Rules/Functions/CallToNonExistentFunctionRule.php
bkline: If I understand your comment below, you actually don't want the argument at all in these cases, since here the symbols are for functions, not classes. Right?

src/Rules/Namespaces/ExistingNamesInGroupUseRule.php
bkline: (I did ask about non-classes in a comment on the issue.)

> One more clarification. I assume that although we're calling the new argument $className, what we want to test is the name of the thing which was not found, even if that "thing" isn't a class (for example, an interface, contstant, trait, etc.). Right?

bkline:
> And please rebase on top of 1.10.x, as the bot says.

I did see what the bot said.

>  If your code is relevant on 1.10.x and you want it to be released sooner, please rebase your pull request and change its target to 1.10.x.

I decided that as long as it kicked in at some time in the future, it didn't really make any significant difference in the long run when that was. But I'll do the rebase if that's the only way the PR will be accepted. 😎

src/Rules/RuleErrorBuilder.php
bkline: You're right about the backslash. My mistake.

> And please use `str_starts_with`.

Hmm. I was under the impression that [PHPStan supports PHP >= 7.2](https://phpstan.org/user-guide/getting-started). The `str_starts_with()` function was only introduced in PHP 8.

I'll see what I can come up with for a test.

bkline:
Sorry, I'm going to drop this PR. I tried the rebase, but it failed, with a merge conflict. I could probably figure that out, but I also looked into what would be involved in creating a test, and I realized I had no idea where to start. I've created plenty of tests for other environments, but I don't know enough about this particular environment's expectations. I was willing to invest some effort into saving other developers from the difficulties I ran into trying to figure out why the tool wasn't working as advertised, but I've come to the realization that I'm probably not in the best position to accomplish that goal.

PR #2605, opened 2023-09-05
fix incorrect doc for inotify_read

taka-oyama:
https://github.com/phpstan/phpstan-src/pull/2599 attempted to improve the doc for inotify_read but the return type is incorrect (returns an inotify event but should return [array of inotify events](https://www.php.net/manual/en/function.inotify-read.php#refsect1-function.inotify-read-returnvalues)) and threw an error in our code.
This attempts to fix that.

Here is the snippet of our code.
https://phpstan.org/r/bb53cbfc-68cb-44d7-aed8-7efbd7438651


ondrejmirtes:
Thank you.

thg2k:
I'm deeply sorry about that. Unfortunately at the same time I was submitting this change I also stumbled upon phpstan/phpstan#8296 and I did not notice this trivial mistake. I just realised that by-ref parameters are **not** checked, and I was passing the inotify_read result to a function mangling the record, together with param-out.

PR #2607, opened 2023-09-07
Include intializations from parent scope for anonymous functions

fabacino:
Fixes bug [#9831](https://github.com/phpstan/phpstan/issues/9831)

### DIFF
--- a/src/Node/ClassPropertiesNode.php
+++ b/src/Node/ClassPropertiesNode.php
@@ -219,6 +219,9 @@ class ClassPropertiesNode extends NodeAbstract implements VirtualNode
 				}
 			} elseif (array_key_exists($propertyName, $initializedPropertiesMap)) {
 				$hasInitialization = $initializedPropertiesMap[$propertyName]->or($usageScope->hasExpressionType(new PropertyInitializationExpr($propertyName)));
+				if ($usageScope->isInAnonymousFunction() && $usageScope->getParentScope() !== null) {
+					$hasInitialization = $hasInitialization->or($usageScope->getParentScope()->hasExpressionType(new PropertyInitializationExpr($propertyName)));
+				}
 				if (
 					strtolower($function->getName()) !== '__construct'
 					&& array_key_exists($propertyName, $initializedInConstructor)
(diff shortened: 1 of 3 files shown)


src/Node/ClassPropertiesNode.php
```diff
@@ -219,6 +219,9 @@ public function getUninitializedProperties(
 				}
 			} elseif (array_key_exists($propertyName, $initializedPropertiesMap)) {
 				$hasInitialization = $initializedPropertiesMap[$propertyName]->or($usageScope->hasExpressionType(new PropertyInitializationExpr($propertyName)));
+				if ($usageScope->isInAnonymousFunction() && $usageScope->getParentScope() !== null) {
+					$hasInitialization = $hasInitialization->or($usageScope->getParentScope()->hasExpressionType(new PropertyInitializationExpr($propertyName)));
+				}
```
staabm: maybe only enter this IF, if the outer `!hasInitialization->yes()` ?

fabacino: Good point, thanks.

I also moved the initialization of `$hasInitialization` below the IF which ignores functions other than the constructor (https://github.com/phpstan/phpstan-src/pull/2607/files#diff-0e5b4f9339ad4cf44915b25c9ea53bf150de4b6fd060856247cc74b0fc0c54eaR225). Since the conditions in that IF do not depend on `$hasInitialization`, we can `continue` the loop earlier. I hope that's ok.

fabacino:
Is there anything I need to do to move this forward? The failing tests do not seem to be caused by the changes introduced in this PR.

ondrejmirtes:
I've been thinking about this, and the main thing here is that I'm not sure that this is the right fix. I just thought about it and I couldn't find a way how to break this so we're probably good to go. Time and feedback will tell us if there's something wrong with it :) Thank you!

fabacino:
I see, thanks for explaining.

ondrejmirtes:
This looks pretty similar to the problem you fixed here, can you please look into it? Thank you: https://github.com/phpstan/phpstan/issues/10048

PR #2609, opened 2023-09-07
Add support and tests for `$this` type assertions

ekisu:
Adds some tests for `$this` type assertion annotations.

As this depends on phpstan/phpdoc-parser#209, I've updated the dependency to point temporarily to that branch.


ondrejmirtes:
Please rebase this without the Composer change and on top of 1.10.x. phpdoc-parser is already updated with your change :)

ekisu:
@ondrejmirtes Done

ondrejmirtes:
Awesome, thank you!

PR #2608, opened 2023-09-07
Implement FailWithoutResultCacheErrorFormatter

staabm:
as described in https://github.com/phpstan/phpstan/pull/9857#discussion_r1318676916

use with
```
php bin/phpstan analyze --error-format=failWithoutResultCache -vvv
```

----

tested with

```
$ php bin/phpstan clear-result-cache
Note: Using configuration file C:\dvl\Workspace\phpstan-src-staabm\phpstan.neon.dist.
Result cache cleared from directory:
C:\dvl\Workspace\phpstan-src-staabm/tmp

mstaab@NB-COMPLEX-61 MINGW64 /c/dvl/Workspace/phpstan-src-staabm (err-formatter)
$ php bin/phpstan analyze --error-format=failWithoutResultCache -vvv
Note: Using configuration file C:\dvl\Workspace\phpstan-src-staabm\phpstan.neon.dist.
Result cache not used because the cache file does not exist.
 1563/1563 [============================] 100% 36 secs/36 secs

Result cache is saved.
Used memory: 3.63 GB

mstaab@NB-COMPLEX-61 MINGW64 /c/dvl/Workspace/phpstan-src-staabm (err-formatter)
$ echo $?
2

mstaab@NB-COMPLEX-61 MINGW64 /c/dvl/Workspace/phpstan-src-staabm (err-formatter)
$ php bin/phpstan analyze --error-format=failWithoutResultCache -vvv
Note: Using configuration file C:\dvl\Workspace\phpstan-src-staabm\phpstan.neon.dist.
 1563/1563 [============================] 100% < 1 sec/< 1 sec

Result cache is saved.


 [OK] No errors


Used memory: 130 MB

mstaab@NB-COMPLEX-61 MINGW64 /c/dvl/Workspace/phpstan-src-staabm (err-formatter)
$ echo $?
0
```

PR #2611, opened 2023-09-08
Added PHPSTAN_FAIL_WITHOUT_RESULT_CACHE env var

staabm:
to  allow testing of the result-cache ~~lets add a new env var~~ a new cli option, which we can use to instruct phpstan to exit with code=2, when the result cache was not used

see https://github.com/phpstan/phpstan/pull/9857#discussion_r1319346887

partly reverts https://github.com/phpstan/phpstan-src/pull/2608

----

```
➜  bug8778 git:(bug8778) ../../phpstan clear-result-cache -q; ../../phpstan analyse -q                                            
➜  bug8778 git:(bug8778) ../../../phpstan-src/bin/phpstan analyse -vvv --generate-baseline --fail-without-result-cache
Note: Using configuration file /Users/staabm/workspace/phpstan/e2e/bug8778/phpstan.neon.
Result cache not used because the metadata do not match.
 2/2 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%  1 sec/1 sec 

Result cache is saved.

                                                                                                                        
 [OK] Baseline generated with 1 error.                                                                                  
                                                                                                                        

Used memory: 78 MB
➜  bug8778 git:(bug8778) echo $?
2
```


ondrejmirtes:
I agree that the error formatter was the wrong direction. But instead of env variable, I'd prefer `--fail-without-result-cache` CLI option :)

staabm:
here you are

```
➜  bug8778 git:(bug8778) ../../phpstan clear-result-cache -q; ../../phpstan analyse -q                                            
➜  bug8778 git:(bug8778) ../../../phpstan-src/bin/phpstan analyse -vvv --generate-baseline --fail-without-result-cache
Note: Using configuration file /Users/staabm/workspace/phpstan/e2e/bug8778/phpstan.neon.
Result cache not used because the metadata do not match.
 2/2 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%  1 sec/1 sec 

Result cache is saved.

                                                                                                                        
 [OK] Baseline generated with 1 error.                                                                                  
                                                                                                                        

Used memory: 78 MB
➜  bug8778 git:(bug8778) echo $?
2

```

ondrejmirtes:
Thank you.

PR #2612, opened 2023-09-08
Uninitialized redeclared property on a class that parent initializes

jordisala1991:
Fix for: https://github.com/phpstan/phpstan/issues/9864

First PR for me here, probably will be doing stuff wrong 😅


jordisala1991:
I took a deep look at the code but I barely understand what is happening. The `getUninitializedProperties` from `ClassPropertiesNode` seems to not take into account properties initialized on parents. Should I try to fix that, or the fix should be done somewhere else?

PR #2613, opened 2023-09-09
Remove unnecessary IO

staabm:
re-submit less controversial part of https://github.com/phpstan/phpstan-src/pull/2596

----

we don't need this `is_file` call (and also not the one below for `is_dir`), because of

```
$classMapFiles = array_filter($classMapPaths, 'is_file');
$classMapDirectories = array_filter($classMapPaths, 'is_dir');
```

[above](https://github.com/phpstan/phpstan-src/pull/2613/files#diff-0fcf29f793b89662fb527a0ee50db9a7238efbb5c18122309ec8b275c14d7c87L76-L77)

### DIFF
--- a/src/Reflection/BetterReflection/SourceLocator/ComposerJsonAndInstalledJsonSourceLocatorMaker.php
+++ b/src/Reflection/BetterReflection/SourceLocator/ComposerJsonAndInstalledJsonSourceLocatorMaker.php
@@ -12,7 +12,6 @@ use PHPStan\File\CouldNotReadFileException;
 use PHPStan\File\FileReader;
 use PHPStan\Internal\ComposerHelper;
 use PHPStan\Php\PhpVersion;
-use function array_filter;
 use function array_key_exists;
 use function array_map;
 use function array_merge;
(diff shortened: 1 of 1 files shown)


src/Reflection/BetterReflection/SourceLocator/ComposerJsonAndInstalledJsonSourceLocatorMaker.php
```diff
@@ -112,18 +112,12 @@ public function create(string $projectInstallationPath): ?SourceLocator
 		);
 
 		foreach ($classMapDirectories as $classMapDirectory) {
-			if (!is_dir($classMapDirectory)) {
```
staabm, line 115: these calls are already done a few lines above at https://github.com/phpstan/phpstan-src/pull/2613/files#diff-0fcf29f793b89662fb527a0ee50db9a7238efbb5c18122309ec8b275c14d7c87L76-L77

src/Reflection/BetterReflection/SourceLocator/ComposerJsonAndInstalledJsonSourceLocatorMaker.php
```diff
@@ -111,19 +109,15 @@ public function create(string $projectInstallationPath): ?SourceLocator
...
-			}
 			$locators[] = $this->optimizedDirectorySourceLocatorRepository->getOrCreate($classMapDirectory);
 		}
 
 		$files = [];
 
+		$classMapFiles = array_filter($classMapPaths, 'is_file');
 		foreach (array_merge($classMapFiles, $filePaths) as $file) {
-			if (!is_file($file)) {
-				continue;
```
ondrejmirtes, line 125: What about checking `is_file` for `$filePaths` here?

ondrejmirtes (changes requested):

src/Reflection/BetterReflection/SourceLocator/ComposerJsonAndInstalledJsonSourceLocatorMaker.php
staabm: good catch, fixed.

ondrejmirtes: How is this change different to what already happened to previous version of the code?

staabm: before this PR, every path in `classMapPaths` was checked twice via `is_dir` / `is_file`. after the PR only once.

ondrejmirtes: Screenshot 2023-09-29 at 14 27 14

See the diff screenshot - I'd rather have the `is_file` / `is_dir` calls in foreaches than the `array_filter` calls.

staabm: oh I see. I read your initial comment differently.

just reworked the code to get rid of `array_filter()`.

ondrejmirtes:
Thank you.

PR #2614, opened 2023-09-09
Reduce unnecessary calls into reflection

staabm:
if the local properties are `true` we don't need to call reflection at all


ondrejmirtes:
Thank you.

PR #2618, opened 2023-09-12
Micro-optimize with `in_array()`

zonuexe:
These points are definitely not the bottleneck, but the chain of comparisons is compiled into fewer bytecode instructions by `in_array()`.

It appears to be valid not only for method calls and property accesses, but also for comparing variables twice.
https://3v4l.org/QQQMr/vld

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -974,7 +974,7 @@ class TypeSpecifier
 			if ($constantType->getValue() === 'boolean') {
 				$type = new BooleanType();
 			}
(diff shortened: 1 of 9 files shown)


thg2k:
Does this bring a measurable performance benefit? Because i have the feeling you are sacrificing readability for a negligible (or possibly negative) performance change.

ondrejmirtes:
This is fine, I like it 👍  Although PHP itself could optimize this by itself when it sees this code pattern.

ondrejmirtes:
Thank you!

PR #2620, opened 2023-09-12
Improve typing of filters that do not accept non-falsy strings

zonuexe:
The following four filters accept neither `''` nor `'0'`, so their values can be typed as `non-falsy-string`.

 * `FILTER_VALIDATE_EMAIL`
 * `FILTER_VALIDATE_IP`
 * `FILTER_VALIDATE_MAC`
 * `FILTER_VALIDATE_URL`

https://3v4l.org/vtE6a


ondrejmirtes:
Thank you!

PR #2621, opened 2023-09-15
fix incorrect argument type for Redis::setex / Redis::setnx

taka-oyama:
This will change argument type of `$value` from `string` to `mixed`.

This is needed because it's possible to set mixed values in PhpRedis when a serializer is set through `Redis::setOption(...)`.


taka-oyama:
Resubmit because I had to refork.

PR #2622, opened 2023-09-15
fix return type of parent calls for SplHeap

schlndh:
This is an attempt to fix https://github.com/phpstan/phpstan/issues/9867

The issue happened because the return type for `SplHeap` methods ignored the phpdoc (as you can see from the change the previous code only considered phpdoc from stubs). I discovered that similar code uses the `TypehintHelper` in these situations, so I used it as well.

But it broke a few tests due to benevolent unions in the `functionaMap.php`. So I added a workaround for it. I couldn't think of any issues, since I haven't found any code uses benevolent unions and which would need a phpdoc as well (e.g. due to generics).

-------

Previously, it worked for `SplMinHeap` because it took a different branch in `PhpClassReflectionExtension::createMethod`:

- Since `SplMinHeap` has the methods duplicated in [phpstorm-stubs](https://github.com/JetBrains/phpstorm-stubs/blob/ed9f6d8b529826bfb82b5d39b2d3697cbfc962bf/SPL/SPL_c1.php#L1499C33-L1499C33), the `$declaringClassName` is `SplMinHeap`.
- Then `$this->signatureMapProvider->hasMethodSignature($declaringClassName, $methodReflection->getName())` returns false, because the `functionMap.php` only has the methods for `SplHeap`. Since this is the branch that contains the incorrect code, the issue was avoided for `SplMinHeap`.
- For `SplMaxHeap` and `SplHeap` the declaring class is `SplHeap` (which is in the `functionMap.php`) and thus the branch was taken and the issue happened.

-----

Here are a few tests that didn't work without the change to `TypehintHelper::decideType` (there were a few more):

- DateTimeModifyReturnTypes.php
- pdo-prepare.php
- bug-6609.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -94,6 +94,7 @@
 			],
 			"jetbrains/phpstorm-stubs": [
 				"patches/PDO.patch",
+				"patches/ReflectionProperty.patch",
 				"patches/SessionHandler.patch"
 			],
 			"rector/rector": [
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a1b4d0008b053558b276997e38288b40",
+    "content-hash": "320291e3ee7da174c482bce9b0a0e86c",
     "packages": [
         {
             "name": "clue/ndjson-react",
@@ -567,12 +567,12 @@
                 }
             },
             "autoload": {
-                "psr-4": {
-                    "Hoa\\Consistency\\": "."
-                },
                 "files": [
                     "Prelude.php"
-                ]
+                ],
+                "psr-4": {
+                    "Hoa\\Consistency\\": "."
+                }
             },
             "notification-url": "https://packagist.org/downloads/",
             "license": [
@@ -974,12 +974,12 @@
                 }
             },
             "autoload": {
-                "psr-4": {
-                    "Hoa\\Protocol\\": "."
-                },
                 "files": [
                     "Wrapper.php"
-                ]
+                ],
+                "psr-4": {
+                    "Hoa\\Protocol\\": "."
+                }
             },
             "notification-url": "https://packagist.org/downloads/",
             "license": [
--- a/patches/ReflectionProperty.patch
+++ b/patches/ReflectionProperty.patch
@@ -1,0 +1,11 @@
+--- Reflection/ReflectionProperty.php	2023-09-07 12:59:56.000000000 +0200
++++ Reflection/ReflectionProperty.php	2023-09-15 13:24:07.900736741 +0200
+@@ -248,7 +248,7 @@
+      * Gets property type
+      *
+      * @link https://php.net/manual/en/reflectionproperty.gettype.php
+-     * @return ReflectionNamedType|ReflectionUnionType|null Returns a {@see ReflectionType} if the
++     * @return ReflectionType|null Returns a {@see ReflectionType} if the
+      * property has a type, and {@see null} otherwise.
+      * @since 7.4
+      */
--- a/src/Reflection/Php/PhpClassReflectionExtension.php
+++ b/src/Reflection/Php/PhpClassReflectionExtension.php
@@ -595,7 +595,7 @@ class PhpClassReflectionExtension
 								$throwType = $throwsTag->getType();
 							}
 							$returnTag = $phpDocBlock->getReturnTag();
-							if ($returnTag !== null) {
+							if ($returnTag !== null && count($methodSignatures) === 1) {
 								$phpDocReturnType = $returnTag->getType();
 							}
 							foreach ($phpDocBlock->getParamTags() as $name => $paramTag) {
@@ -860,10 +860,11 @@ class PhpClassReflectionExtension
 			);
 		}
 
-		$returnType = null;
 		if ($stubPhpDocReturnType !== null) {
 			$returnType = $stubPhpDocReturnType;
 			$phpDocReturnType = $stubPhpDocReturnType;
+		} else {
+			$returnType = TypehintHelper::decideType($methodSignature->getReturnType(), $phpDocReturnType);
 		}
 
 		return new FunctionVariantWithPhpDocs(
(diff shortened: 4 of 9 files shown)


patches/Memcache.patch
```diff
@@ -0,0 +1,11 @@
+--- memcache/memcache.php	2023-09-04 17:18:27.000000000 +0200
++++ memcache/memcache.php	2023-09-08 14:29:48.807174822 +0200
+@@ -217,7 +217,7 @@
+      * The lowest byte of the int is reserved for pecl/memcache internal usage (e.g. to indicate
+      * compression and serialization status).
+      * </p>
+-     * @return string|array|false <p>
```
schlndh: This broke `memcache-get.php` test. The explanation is here: https://github.com/phpstan/phpstan-src/pull/2323#issuecomment-1497231186

ondrejmirtes:
I'm gonna take a look at this soon but can you please look at this? https://github.com/phpstan/phpstan/discussions/9849#discussioncomment-6913474

I list two very common and very annoying issues, they're somewhat in the same area as this bug, and I think you're more than capable of addressing them 😊 Thank you very much.

ondrejmirtes (changes requested):
Can we solve this a different way instead of patching jetbrains/phpstorm-stubs? Like adding some entries to functionMap.php? Thanks!

patches/Memcache.patch
schlndh: This is not (easily) possible to fix via `functionMap.php`. There are two variants in function map: one which returns `mixed` and one which returns `array|false`. And as you can see, the phpdoc return type in phpstorm-stubs is `string|array|false`.

`decideType(mixed, string|array|false) = string|array|false`, which doesn't seem like something that could be changed.

Perhaps if `decideType` also considered where the two types come from. E.g. in this case we know that the first type comes from the function map. Since `string|array|false` could have been used in the function map, we could reach the conclusion that we used the broader type intentionally. However, there are issues in the function map as well, so it's though to say whether such approach would be more feasible.

ondrejmirtes: What about disregarding PhpStorm stubs type completely if functionMap has multiple variants? But the "golden test" for Reflection will come in handy in that scenario :)

schlndh:
Note to self: This might also get affected/fixed: https://github.com/phpstan/phpstan/issues/10057

```
2) PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD DOMDocument::load" ('METHOD DOMDocument::load', 'Has side-effects: Maybe\nVisi... mixed')
--- Expected
+++ Actual
@@ @@
     /**
      * @param string $filename
      * @param int $options
-     * @return mixed
+     * @return bool|DOMDocument
      */
     function load(string $filename, int $options = 0): mixed'
```

https://github.com/schlndh/phpstan-src/actions/runs/6586745404/job/17895674524?pr=1

patches/Memcache.patch
schlndh: Nice, that worked (for memcace::get). I limited it specifically to the return type. The stubs are still useful/necessary for positional parameter names, `@throws`, ...

But unfortunately, it doesn't cover the `ReflectionPropety` patch.

schlndh:
I removed two of the patches (the SplHeap patch was accepted to upstream, and Memcache was fixed by your idea).

I checked the differences in the reflection golden test - they seem mostly OK (though a lot of them are in niche, undocumented functions/methods, so I can't be 100% sure). Here are some of the wrong changes that I noticed:

```
12) PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD DOMDocument::load" ('METHOD DOMDocument::load', 'Is internal: Yes\nHas side-ef... mixed')
--- Expected
+++ Actual
@@ @@
     /**
      * @param string $filename
      * @param int $options
-     * @return mixed
+     * @return bool|DOMDocument
      */
     function load(string $filename, int $options = 0): mixed'

/home/schlndh/devel/custom/phpstan-src/tests/PHPStan/Reflection/ReflectionProviderGoldenTest.php:69

13) PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD DOMDocument::loadXML" ('METHOD DOMDocument::loadXML', 'Is internal: Yes\nHas side-ef... mixed')
--- Expected
+++ Actual
@@ @@
     /**
      * @param string $source
      * @param int $options
-     * @return mixed
+     * @return bool|DOMDocument
      */
     function loadXML(string $source, int $options = 0): mixed'
```

It seems that these methods can return `DOMDocument` only when called statically, which doesn't seem to be possible since PHP 8 (this result is from 8.2).

```
62) PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD MongoDB::createCollection" ('METHOD MongoDB::createCollection', 'Is internal: Yes\nHas side-ef... mixed')
--- Expected
+++ Actual
@@ @@
      * @param bool $options
      * @param int $size
      * @param int $max
-     * @return MongoCollection
+     * @return MongoCollection<p>
      */
     function createCollection(mixed $name, mixed $options, mixed $size, mixed $max): mixed'
```

There are several cases like this. The problem is this: `@return MongoCollection <p>Returns a collection object representing the new collection.</p>`. I'm not sure if the phpdoc parser is supposed to ignore the space between the type and the `<p>`. I found 5 such cases (all of them mongo related), so it's probably not a big deal for now.

```
107) PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD ReflectionUnionType::getTypes" ('METHOD ReflectionUnionType::getTypes', 'Is internal: Yes\nVisibility:... array')
--- Expected
+++ Actual
@@ @@
 Visibility: public
 Variants: 1
     /**
-     * @return array
+     * @return array<ReflectionIntersectionType|ReflectionNamedType>
      */
     function getTypes(): array'
```
I guess this should be just `array<ReflectionType>`.

ondrejmirtes:
As for the Mongo `@return MongoCollection <p>Returns a collection object representing the new collection.</p>` problem:

There's some logic in phpstan/phpdoc-parser that should avoid this problem: https://github.com/phpstan/phpdoc-parser/blob/c2b8bbfa971e25976e9c64ea30a4e944e2688cb4/src/Parser/TypeParser.php#L367-L395

Can you please try to write a failing test in the repository and maybe modify the logic so that it works for Mongo PHPDocs too? Thanks.

ondrejmirtes:
Thank you!

PR #2624, opened 2023-09-17
Bump to rector ^0.18

samsonasik:
Let see if the Rector 0.18 works on phpstan itself so we can fix early in case there is issue on it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
 		"phpstan/phpstan-phpunit": "^1.0",
 		"phpstan/phpstan-strict-rules": "^1.5.1",
 		"phpunit/phpunit": "^9.5.4",
-		"rector/rector": "^0.17",
+		"rector/rector": "^0.18",
 		"shipmonk/name-collision-detector": "^2.0"
 	},
 	"config": {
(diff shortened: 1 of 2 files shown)


samsonasik:
CI error seems unrelated:

```

php bin/phpstan clear-result-cache -q && php -d memory_limit=448M bin/phpstan

In Resolver.php line 469:
                                                                               
  Service (PHPStan\Analyser\RuleErrorTransformer::__construct()): syntax erro  
  r, unexpected '|', expecting variable (T_VARIABLE)                 
```

ondrejmirtes:
It's very much related.

It seems that Rector 0.18, even with `DowngradeUnionTypeDeclarationRector`, does not downgrade:

```php
public function __construct(Name|Expr $class, Identifier|Expr $name, Expr\StaticCall $originalNode)
```

The failures are most obvious here: https://github.com/phpstan/phpstan-src/actions/runs/6212770196

samsonasik:
Hm.., that can't be reproduced at https://getrector.com/demo/63390460-18c0-454c-b7a5-5fc183abd509

I probably missing something in code/config, could you try reproduce at https://getrector.com/demo ? Thank you.

ondrejmirtes:
Nope, sorry, you have the reproduction right here in phpstan-src :) If you can't upgrade the version here to Rector 0.18, please close this PR. It's not a priority for me right now.

samsonasik:
ok 👍

PR #2625, opened 2023-09-17
fix iterator_to_array, iterator_count for PHP 8.2

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/9793 (and the example given in https://github.com/phpstan/phpstan/issues/7760, though that issue is specified more broadly).

The problem happens here: https://github.com/phpstan/phpstan-src/blob/02ea303a16dab1cd9fb0a256826d5ac24c9c68e1/src/Reflection/SignatureMap/Php8SignatureMapProvider.php#L198 . We get the correct signature from php8-stubs, but the signature in `functionMap.php` was outdated and it overwrote the the correct signature.

### DIFF
--- a/resources/functionMap_php82delta.php
+++ b/resources/functionMap_php82delta.php
@@ -21,6 +21,8 @@
  */
 return [
 	'new' => [
+		'iterator_count' => ['0|positive-int', 'iterator'=>'iterable'],
+		'iterator_to_array' => ['array', 'iterator'=>'iterable', 'use_keys='=>'bool'],
 		'str_split' => ['list<string>', 'str'=>'string', 'split_length='=>'positive-int'],
 	],
 	'old' => [
--- a/tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
+++ b/tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
@@ -1512,4 +1512,41 @@ class CallToFunctionParametersRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/call-to-function-named-params-multivariant.php'], []);
 	}
 
+	public function testBug9793(): void
+	{
+		$errors = [];
+
+		if (PHP_VERSION_ID < 80200) {
+			$errors = [
+				[
+					'Parameter #1 $iterator of function iterator_to_array expects Traversable, array<stdClass> given.',
+					13,
+				],
+				[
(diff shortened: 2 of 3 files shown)


ondrejmirtes:
We could have done this, but it feels just like a workaround.

The facts:

* php-8-stubs have the correct native parameter type for PHP 8.2: https://github.com/phpstan/php-8-stubs/blob/main/stubs/ext/spl/iterator_count.php
* Types from functionMap.php are considered as PHPDoc types, and in case of a clash (PHPDoc type is not a subtype of the native type), the PHPDoc type is thrown away. This logic is in `TypehintHelper::decideType()` that's called in Php8SignatureMapProvider.
* There are other functions that have the wrong type in functionMap, the correct native type in php-8-stubs, and they work correctly. For example curl_init: https://phpstan.org/r/436b8a3d-4f12-401e-8974-266ea83eec3e

This is why we need the actual fix in the logic and not a workaround :) Because more functions might be impacted by this bug, if not now, then certainly in the future.

ondrejmirtes:
Thinking about the actual reason, I think the problem might be that functionMap has `Traversable` but the correct type is `iterable`. But `iterable`->isSuperTypeOf(`Traversable`) is `yes()`. At the same time PHPDoc types can actually be used to narrow down the type. Think of native `int` with PHPDoc's `positive-int`.

`iterable` is actually `array|Traversable`, but in PHPStan there's a special `IterableType` implementation.

So maybe we need to special-case this problem in TypehintHelper. If we have `iterable` in the native type and `Traversable` in the PHPDoc, `iterable` should still win.

schlndh:
@ondrejmirtes Ok, thanks for the pointers. I'll try look into the underlying problem. But I guess it wouldn't hurt to fix the functionMap.php regardless.

ondrejmirtes:
I'm looking forward to this one! The reflection golden test should make this a breeze, we'll see all the implication of the fix :)

schlndh:
I tried adding a hack to `decideType` instead of fixing the signature map. According the the reflection golden test the only two currently affected native functions/methods are these two. So it doesn't seem worth it to me to add such a weird edge-case to `decideType`. Also, the change broke existing test - i.e. it could have negative impact on non-native code.

You can see the experiment here: https://github.com/schlndh/phpstan-src/pull/3

ondrejmirtes:
Awesome, thank you! It's great we can now do these changes in informed way.

PR #2630, opened 2023-09-18
Result-Cache: Indicate key differences in cache metadata

staabm:
the goal of this PR is to ease debugging why/which result cache metadata is different.

before this PR its pretty hard to e.g. in CI get an idea why the result cache metadata did not match and which of the key-indicators is the cause.

After this change the console output looks like:
_note the new keys at the end of the error message_

after changes in the composer.lock

```
➜  phpstan-src git:(1.10.x) ✗ php bin/phpstan -vvv
Result cache not used because the metadata do not match: projectConfig, composerLocks
 1561/1561 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100% 21 secs/21 secs

Result cache is saved.
```

or when no cache is present:

```
➜  phpstan-src git:(1.10.x) ✗ php bin/phpstan -vvv
Result cache not used because the metadata do not match: phpstanVersion, projectConfig, composerLocks, composerInstalled
 1561/1561 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100% 25 secs/25 secs

Result cache is saved.
```

or when the level changes

```
php bin/phpstan -vvv
Result cache not used because the metadata do not match: projectConfig, level
    0/1561 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0% < 1 sec/< 1 sec
```

etc.

### DIFF
--- a/src/Analyser/ResultCache/ResultCacheManager.php
+++ b/src/Analyser/ResultCache/ResultCacheManager.php
@@ -28,6 +28,7 @@ use function array_unique;
 use function array_values;
 use function count;
 use function get_loaded_extensions;
+use function implode;
 use function is_array;
 use function is_file;
 use function is_string;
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

janedbal:
Thank you!

PR #2632, opened 2023-09-19
remove phpstan/phpstan-php-parser

staabm:
`nikic/PHP-Parser` meanwhile ships with all types defined in [stubs of phpstan-php-parser](https://github.com/phpstan/phpstan-php-parser/tree/1.2.x/stubs)


staabm:
I will followup in other phpstan/* repos

ondrejmirtes:
Thank you.

PR #2633, opened 2023-09-20
Detect array_rand second parameter

staabm:
closes https://github.com/phpstan/phpstan/issues/9803

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-9803.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-9803.php
@@ -4,17 +4,24 @@ declare(strict_types=1);
 
 namespace Bug9803;
 
+use function PHPStan\Testing\assertType;
+
 function doFoo() {
 	$random = rand(1, 5);
 	$array = array("one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten");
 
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Functions/data/bug-9803.php
```diff
@@ -0,0 +1,20 @@
...
+
+function doFoo() {
+	$random = rand(1, 5);
+	$array = array("one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten");
+
+	$keys = array();
+	if ($random == 1)
+		$keys = array(array_rand($array));
+	else
+		$keys = array_rand($array, $random);
```
ondrejmirtes, line 19: Would be nice to sprinkle some assertType calls here, and reference it from NodeScopeResolverTest.

And please add `{ ... }` around those `if` and `else`. Thanks!

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Functions/data/bug-9803.php
staabm: done

ondrejmirtes:
Thank you!

PR #2634, opened 2023-09-21
fix: incorrect argument type for Redis::setex / Redis::setnx

taka-oyama:
Resubmit of https://github.com/phpstan/phpstan-src/pull/2621 to rebase it with 1.10x.

This will change argument type of `$value` from `string` to `mixed`.

This is needed because it's possible to set mixed values in PhpRedis when a serializer is set through `Redis::setOption(...)`.


thg2k:
@ondrejmirtes is there a specific reason to keep both cases? IIRC they are strtolower-ed during the map build

taka-oyama:
Closing since it seems like it was fixed in https://github.com/phpstan/phpstan-src/pull/2706.

PR #2635, opened 2023-09-21
Support union-types in array_chunk()

staabm:


src/Type/Php/ArrayChunkFunctionReturnTypeExtension.php
```diff
@@ -46,7 +48,8 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 			$preserveKeys = false;
 		}
 
-		if ($lengthType instanceof ConstantIntegerType && $lengthType->getValue() < 1) {
```
staabm, line 49: primary motivation was getting rid of the `instanceof ConstantIntegerType` parts in this extension

src/Type/Php/ArrayChunkFunctionReturnTypeExtension.php
```diff
@@ -55,10 +58,17 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+		if (count($constantArrays) > 0 && $biggerOne->isSuperTypeOf($lengthType)->yes()) {
 			$results = [];
 			foreach ($constantArrays as $constantArray) {
-				$results[] = $constantArray->chunk($lengthType->getValue(), $preserveKeys);
+				foreach ($lengthType->getFiniteTypes() as $finiteType) {
+					if (!$finiteType instanceof ConstantIntegerType) {
+						throw new ShouldNotHappenException();
+					}
+
+					$results[] = $constantArray->chunk($finiteType->getValue(), $preserveKeys);
```
staabm: I think this might get slow when big integer ranges are involved for `$length`. any opinions about possible limits or different implementations - this might also be rare, as we only do it when `$preserveKey` is a known boolean and the array is constant at the same time.

tests/PHPStan/Analyser/data/array-chunk-php8.php
```diff
@@ -0,0 +1,18 @@
+<?php declare(strict_types = 1);
+
+namespace ArrayChunkPhp8;
```
staabm, line 3: failling tests before PR:

https://phpstan.org/r/467033a8-2d23-49f5-a20b-90871639ef6c

src/Type/Php/ArrayChunkFunctionReturnTypeExtension.php
```diff
@@ -46,22 +48,32 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-			foreach ($constantArrays as $constantArray) {
-				$results[] = $constantArray->chunk($lengthType->getValue(), $preserveKeys);
+		if ($preserveKeys !== null) {
+			$constantArrays = $arrayType->getConstantArrays();
+			$biggerOne = IntegerRangeType::fromInterval(1, null);
+			if (count($constantArrays) > 0 && $biggerOne->isSuperTypeOf($lengthType)->yes()) {
+				$results = [];
+				foreach ($constantArrays as $constantArray) {
+					foreach ($lengthType->getFiniteTypes() as $finiteType) {
+						if (!$finiteType instanceof ConstantIntegerType || $finiteType->getValue() < 1) {
```
ondrejmirtes, line 72: This is a courageous assumption. I wouldn't throw `ShouldNotHappenException`, it can break right after https://github.com/phpstan/phpstan-src/pull/2642/files gets merged.

So instead it'd be better to return the default return type, instead of crashing.

Also, I'd limit the possibilities to a small number of finite types, rather than have a performance problem on our hands. It can get rather big.

Also - you need to test this with `$lengthType` of infinite interval like `int<1, max>` or `int<50, max>`.

ondrejmirtes (changes requested):

src/Type/Php/ArrayChunkFunctionReturnTypeExtension.php
staabm: here you are

tests/PHPStan/Analyser/data/array-chunk.php
```diff
@@ -43,4 +43,36 @@ public function constantArraysWithOptionalKeys(array $arr): void
...
+	}
+
+	/**
+	 * @param int<1, 4> $oneToFour
+	 * @param int<1, 5> $tooBig
+	 */
+	function testLimits(array $arr, int $oneToFour, int $tooBig) {
+		/** @var array{a: 0, b?: 1, c: 2, d: 3} $arr */
+		assertType('array{0: array{0: 0, 1?: 1|2, 2?: 2|3, 3?: 3}, 1?: array{0?: 2|3, 1?: 3}}|array{array{0}, array{0?: 1|2, 1?: 2}, array{0?: 2|3, 1?: 3}, array{0?: 3}}', array_chunk($arr, $oneToFour));
+		assertType('non-empty-list<non-empty-list<0|1|2|3>>', array_chunk($arr, $tooBig));
```
staabm, line 74: here we can see the `FINITE_TYPES_LIMIT` in action

ondrejmirtes:
Thank you.

PR #2637, opened 2023-09-21
Allow @phpstan-use and @use tags to be declared into class phpdoc

baptistepillot:
This implements the feature request here: https://github.com/phpstan/phpstan/issues/7486.
I think this does the job. Hope this will help.

### DIFF
--- a/src/Rules/Generics/UsedTraitsRule.php
+++ b/src/Rules/Generics/UsedTraitsRule.php
@@ -11,6 +11,8 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\FileTypeMapper;
 use PHPStan\Type\Type;
 use function array_map;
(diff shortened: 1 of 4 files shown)


baptistepillot:
Seems that php 8.0+ ?-> nullsafe operator is not retro-translated for php 7.2+.
I am updating my commit and change request immediately...

baptistepillot:
Integration tests fail, but it does not seem to be caused by my changes, I guess.
I send another pull request proposal for branch 1.10.x : https://github.com/phpstan/phpstan-src/pull/2638
If this works with 1.10.x, should I close the current 1.11.x pull request, or rebase it regularly until checks work?

ondrejmirtes:
Just keep the 1.10.x one open :)

PR #2638, opened 2023-09-21
Allow @phpstan-use and @use tags to be declared into class phpdoc

baptistepillot:
This implements the feature request here: https://github.com/phpstan/phpstan/issues/7486.
I think this does the job. Hope this will help.

### DIFF
--- a/src/Rules/Generics/UsedTraitsRule.php
+++ b/src/Rules/Generics/UsedTraitsRule.php
@@ -11,6 +11,8 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\FileTypeMapper;
 use PHPStan\Type\Type;
 use function array_map;
+use function array_merge;
+use function is_null;
 use function sprintf;
 use function ucfirst;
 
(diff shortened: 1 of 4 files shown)


baptistepillot:
php 7.4's phpstan-doctrine fails because of Allowed memory size of 805306368 bytes exhausted (tried to allocate 20480 bytes).
The same as for a number of recent merge requests. phpstan eats a big bucket of memory for breakfast, seems that php 7.4 + Doctrine breaks the CI chain limited memory down.

baptistepillot:
Documentation update proposal regarding to `@use` : https://github.com/phpstan/phpstan/pull/9910

baptistepillot:
Hello there. What's up ? Do you envisage a merge date for this pull request ?
I am ready to rebase this before merging when you wish.

mxr576:
The current default branch is https://github.com/phpstan/phpstan-src/tree/1.11.x so the target should be changed for a start.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/pull/6115

PR #2640, opened 2023-09-22
Implement PregGrepDynamicReturnTypeExtension

staabm:
eleminates `false` in return-type when the pattern is valid


staabm:
ohh .. realized I did this once and failed in https://github.com/phpstan/phpstan-src/pull/2560

PR #2641, opened 2023-09-22
Fix resolving nested conditional types with union subjects

rvanvelzen:
Fixes phpstan/phpstan#9860


ondrejmirtes:
Thank you!

staabm:
@rvanvelzen thanks for this nice improvement.

this seems to have caused a regression, please see https://github.com/phpstan/phpstan/issues/9963

PR #2642, opened 2023-09-22
Make int ranges generic

rvanvelzen:
Fix phpstan/phpstan#6810

### DIFF
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -64,6 +64,7 @@ use PHPStan\Type\Enum\EnumCaseObjectType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\FloatType;
 use PHPStan\Type\Generic\GenericClassStringType;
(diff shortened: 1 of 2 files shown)


ondrejmirtes (changes requested):
I wonder - do we need a new `GenericIntegerRangeType` for this? Other types have a similar capability out of the box. Can we just change `IntegerRangeType` to support this?

rvanvelzen:
I did consider that option, but given that `IntegerRangeType` is public api and requires `?int` bounds, it didn't seem viable to me.

ondrejmirtes:
@rvanvelzen Can we make a new IntegerRangeType, deprecate the old one, and start using the new one everywhere? In 2.0 I'd just renamed the new one to the current `IntegerRangeType` name.

DaDeather:
Is anyone still working on this? Otherwise I'd possibly get to take care of this :+1:.

VincentLanglet:
Hi @DaDeather , feel free to work on it ;)

PR #2643, opened 2023-09-24
Fix 'getallheaders' return types

ghost:
The function 'getallheaders' can return both an array or false.

https://www.php.net/manual/en/function.getallheaders.php


stof:
@marv0g you need to rebase your commit to be based on 1.10.x, not just change the target branch. Otherwise, you are asking to merge all changes of 1.11.x into 1.10.x

PR #2644, opened 2023-09-24
Reduce memory consumption of TypeInferenceTestCase

staabm:
reduce the data beeing transfered between dataProvider and tests, by using just scalar values instead of objects with complex graphs which needs a lot of memory/time to serialize.

----

refs https://github.com/phpstan/phpstan/discussions/9914

repro https://github.com/gnutix/phpstan-typeinferencetest-memory-exhausted

before this PR
```
➜  memopt git:(main) ✗ vendor/bin/phpunit --no-results

PHPUnit 10.3.5 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.10
Configuration: /Users/staabm/workspace/memopt/phpunit.xml.dist

FFF....FF....FFFFFFFFFFFFFFFFFFFFF........FFF....F.F.....FFFFFF  63 / 168 ( 37%)
F......FFFFF.FFF................FFFFF.....F.....FFFFFFFFF...... 126 / 168 ( 75%)
FFFFFFFFF.......F.....FFFF............FF..                      168 / 168 (100%)

Time: 00:27.074, Memory: 5.38 GB
```

after this PR

```
 memopt git:(main) ✗ vendor/bin/phpunit --no-results

PHPUnit 10.3.5 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.10
Configuration: /Users/staabm/workspace/memopt/phpunit.xml.dist

FFF....FF....FFFFFFFFFFFFFFFFFFFFF........FFF....F.F.....FFFFFF  63 / 168 ( 37%)
F......FFFFF.FFF................FFFFF.....F.....FFFFFFFFF...... 126 / 168 ( 75%)
FFFFFFFFF.......F.....FFFF............FF..                      168 / 168 (100%)

Time: 00:00.042, Memory: 82.00 MB
```

=> ~98% memory improvement

PR #2647, opened 2023-09-25
Reproduce NodeScopeResolverTest memory issue

staabm:
reproduce https://github.com/phpstan/phpstan-src/pull/2644 in PHPStan codebase

with this changes I am able to reproduce the memory issue reported in https://github.com/phpstan/phpstan/discussions/9914 within the PHPStan codebase.

it seems to be a **phpunit10 only** issue, therefore was not reproducible with regular 1.10.x branch

just running
```
vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php --no-results
```

with this branch is running "endless" and takes gigabytes of memory.
(it finishes with enough RAM and time)

----

after picking the fix from https://github.com/phpstan/phpstan-src/pull/2644 into this PR it completes instantly and uses very little memory

```
$ vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php --no-results

FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 441 / 850 ( 51%)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 504 / 850 ( 59%)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 567 / 850 ( 66%)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 630 / 850 ( 74%)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 693 / 850 ( 81%)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 756 / 850 ( 88%)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 819 / 850 ( 96%)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF                                 850 / 850 (100%)

Time: 00:00.729, Memory: 126.00 MB
```

### DIFF
--- a/tests/PHPStan/Analyser/NodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/NodeScopeResolverTest.php
@@ -97,6 +97,8 @@ class NodeScopeResolverTest extends TypeInferenceTestCase
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/key-of.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/value-of.php');
 
+		return;
+
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/ext-ds.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/arrow-function-return-type.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/is-numeric.php');
--- a/composer.json
+++ b/composer.json
@@ -51,14 +51,13 @@
 		"phpstan/phpstan": "self.version"
 	},
 	"require-dev": {
-		"brianium/paratest": "^6.5",
 		"cweagans/composer-patches": "^1.7.3",
 		"php-parallel-lint/php-parallel-lint": "^1.2.0",
 		"phpstan/phpstan-deprecation-rules": "^1.0",
 		"phpstan/phpstan-nette": "^1.0",
 		"phpstan/phpstan-phpunit": "^1.0",
 		"phpstan/phpstan-strict-rules": "^1.5.1",
-		"phpunit/phpunit": "^9.5.4",
+		"phpunit/phpunit": "10.3.5",
 		"rector/rector": "^0.17",
 		"shipmonk/name-collision-detector": "^2.0"
 	},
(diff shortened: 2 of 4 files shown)


tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -97,6 +97,8 @@ public function dataFileAsserts(): iterable
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/key-of.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/value-of.php');
 
+		return;
```
staabm, line 100: return early as otherwise the reproducer takes endless time and RAM.

depending on your machine you might need to move this return a few lines above to reduce waiting time.

ondrejmirtes:
So why does this behaviour manifest only when there are top-level assertType calls, and not inside a closure?

Learning why could teach us something about optimization which could be applied in areas not related to testing.

staabm:
so I did a lot of measureing with different constellations.

using `"phpunit/phpunit": "10.3.5"` and `phpunit/phpunit@ccb711a`,

I can see a clear difference between tests beeing wrapped in functions or not.
this difference is only visible on `"phpunit/phpunit": "10.3.5"` though.
for some reasons it seems the "dataprovider serializing overhead" in phpunit is higher when the tests are not wrapped in a function.

the [~20% difference in memory consumption](https://github.com/phpstan/phpstan-src/pull/2644#issuecomment-1733662630) is just in the exporter

grafik

----

as soon as I use `phpunit/phpunit@ccb711a` there is no more difference between tests wrapped in functions or not.

PR #2648, opened 2023-09-25
Implement reportUncheckedExceptionDeadCatch

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9921

### DIFF
--- a/src/Rules/Exceptions/CatchWithUnthrownExceptionRule.php
+++ b/src/Rules/Exceptions/CatchWithUnthrownExceptionRule.php
@@ -7,6 +7,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Node\CatchWithUnthrownExceptionNode;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\TrinaryLogic;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\VerbosityLevel;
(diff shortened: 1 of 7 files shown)


src/Rules/Exceptions/CatchWithUnthrownExceptionRule.php
```diff
@@ -32,6 +40,14 @@ public function processNode(Node $node, Scope $scope): array
 			];
 		}
 
+		$isCheckedException = TrinaryLogic::createFromBoolean($this->reportUncheckedExceptionDeadCatch)->lazyOr(
```
ondrejmirtes: This code isn't very readable to me. We want to early-return only if `$this->reportUncheckedExceptionDeadCatch` is false. So please do all this logic inside `if (!$this->reportUncheckedExceptionDeadCatch) { ... } `.

Thank you.

ondrejmirtes (changes requested):

ondrejmirtes:
Much better, thank you. Please also send an update to the documentation and the blogpost.

ondrejmirtes:
Still waiting for that docs update :)

VincentLanglet:
> Still waiting for that docs update :)

Sorry for the delay, it's definetly in my TODO

VincentLanglet:
I did https://github.com/phpstan/phpstan/pull/9944 @ondrejmirtes

PR #2651, opened 2023-09-27
removed hardcoded DIRECTORY

oleg-andreyev:
Depends on new version of https://github.com/phpstan/php-8-stubs/pull/77

This is part 2 for https://github.com/phpstan/phpstan/issues/9930


ondrejmirtes:
As I said in https://github.com/phpstan/phpstan/issues/9930#issuecomment-1737126541, this is not a supported use case.

PR #2652, opened 2023-09-29
Fix subtracting types from union template types

rvanvelzen:
Fix phpstan/phpstan#9939

### DIFF
--- a/tests/PHPStan/Analyser/NodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/NodeScopeResolverTest.php
@@ -1281,6 +1281,7 @@ class NodeScopeResolverTest extends TypeInferenceTestCase
 
 		if (PHP_VERSION_ID >= 80100) {
 			yield from $this->gatherAssertTypes(__DIR__ . '/../Rules/Comparison/data/bug-9499.php');
+			yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-9939.php');
 		}
 
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-5998.php');
--- a/tests/PHPStan/Analyser/data/bug-9939.php
+++ b/tests/PHPStan/Analyser/data/bug-9939.php
@@ -1,0 +1,63 @@
+<?php declare(strict_types = 1); // lint >= 8.1
+
+namespace Bug9939;
+
+use function PHPStan\Testing\assertType;
+
+enum Combinator
+{
+	case NEXT_SIBLING;
+	case CHILD;
+	case FOLLOWING_SIBLING;
+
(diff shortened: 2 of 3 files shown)


ondrejmirtes:
FYI @rvanvelzen I described an issue here https://github.com/phpstan/phpstan/discussions/9925#discussioncomment-7137857, I think it's up your alley to fix it :)

`New_` has special behaviour in `MutatingScope::exactInstantiation` that looks at class-level `@template` and uses them for constructor. Maybe we would benefit from moving it into ParametersAcceptorSelector, maybe that would help us in part to solve that issue.

If you decide to look at it that would be great!

tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -1341,6 +1341,7 @@ public function dataFileAsserts(): iterable
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/nullsafe-vs-scalar.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-8517.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/../Rules/Functions/data/bug-9803.php');
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-9939.php');
```
ondrejmirtes: This needs to be put in `PHP_VERSION_ID` if.

tests/PHPStan/Analyser/data/bug-9939.php
```diff
@@ -0,0 +1,63 @@
+<?php declare(strict_types = 1);
```
ondrejmirtes: This needs `// lint >= 8.1`

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/NodeScopeResolverTest.php
ondrejmirtes: Also it's better to put in some random place in the middle, because at the end it's more likely to have conflict when merging into 1.11.x

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

thg2k:
Maybe a simplified test case instead of that bloated one? This is also PHP-version independent

```php

/**
 * @template T of string|array<string>
 */
class C {
  /**
   * @param T $a
   */
  public function foo($a): string {
    if (is_array($a)) {
      return "";
    }
    \PHPStan\Testing\assertType('T of string (class C, argument)', $a);
    return (string) $a;
  }
}
```

https://phpstan.org/r/bc6b8adf-a25e-4960-8f43-eca8399cf509

PR #2653, opened 2023-09-29
Implement @phpstan-trace

rvanvelzen:
Just a proof-of-concept. This requires a virtual node because it shows the value _after_ the following statement.

Please close if you're not interested :)


ondrejmirtes:
I think our `\PHPStan\dumpType()` is much better :) Every time I try to use `@psalm-trace` I get frustrated, because it does something else than I expect, exactly because it's not clear whether it should print the state before or after the statement, especially when we overwrite the same variable for example.

PR #2654, opened 2023-09-29
Restrict "Merge maintained branch" workflow to the main repository

thg2k:
This action is deemed to fail on all forks, plus even if fixed with a local secret token it would generate a desynced 1.11.x branch

### DIFF
--- a/.github/workflows/merge-maintained-branch.yml
+++ b/.github/workflows/merge-maintained-branch.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   merge:
     name: Merge branch
+    if: github.repository_owner == 'phpstan'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

BTW You shouldn't enable Actions on your fork anyway :) They're disabled by default.

thg2k:
> Thank you!
> 
> BTW You shouldn't enable Actions on your fork anyway :) They're disabled by default.

I have various branches with tweaks/customizations/wips that would benefit most of the tests. To this topic: Would it be a problem to lose the "branches: 1.10.x" limitation on the tests workflow? Cause my branch is called differently

ondrejmirtes:
Just open a PR here or in your fork and it's gonna get tested.

PR #2656, opened 2023-09-30
Drop wrong float comparison with filter_var()

mvorisek:
discussion https://github.com/phpstan/phpstan-src/pull/2358#discussion_r1174422409

`... < PHP_FLOAT_EPSILON` condition for small numbers always wrongly returned true, `===` is the only valid cmp as long as both left/right side of the operator are of type float

### DIFF
--- a/src/Type/Php/FilterFunctionReturnTypeHelper.php
+++ b/src/Type/Php/FilterFunctionReturnTypeHelper.php
@@ -32,7 +32,6 @@ use function is_int;
 use function octdec;
 use function preg_match;
 use function sprintf;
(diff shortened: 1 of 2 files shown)


mvorisek:
the failing tests are unrelated

ondrejmirtes:
These changes need tests to see what they’re sctually fixing.

mvorisek:
help welcomed with tests

staabm:
TIL that `'1e-50'` is a numeric string. I guess its super rare, but we might have some logic in the codebase which does not allow `-` in numeric strings.

ondrejmirtes:
Thank you.

PR #2657, opened 2023-09-30
Fix !isset() with ArrayDimFetch

staabm:
closes https://github.com/phpstan/phpstan/issues/9908
closes https://github.com/phpstan/phpstan/issues/8724
closes https://github.com/phpstan/phpstan/issues/8659
closes https://github.com/phpstan/phpstan/issues/8190
closes https://github.com/phpstan/phpstan/issues/5128
closes https://github.com/phpstan/phpstan/issues/9426
closes https://github.com/phpstan/phpstan/issues/9580


src/Analyser/MutatingScope.php
```diff
@@ -1996,16 +1996,16 @@ public function issetCheck(Expr $expr, callable $typeCallback, ?bool $result = n
 		} elseif ($expr instanceof Node\Expr\ArrayDimFetch && $expr->dim !== null) {
 			$type = $this->getType($expr->var);
 			$dimType = $this->getType($expr->dim);
-			$hasOffsetValue = $type->hasOffsetValueType($dimType);
 			if (!$type->isOffsetAccessible()->yes()) {
 				return $result ?? $this->issetCheckUndefined($expr->var);
 			}
 
+			$hasOffsetValue = $type->hasOffsetValueType($dimType);
```
staabm: unrelated cleanup / calcuate the type only when needed

staabm:
I can see in the issue bot, that the change as is would resolve a lot of open issues.

I have no idea yet, how to express the "nullability" in the !isset() case.

ondrejmirtes (changes requested):
So if we're going to handle `isset` in falsy context, we need to understand what `isset` does. With `!isset($foo)` it means:

1) The variable might not exist
2) The variable might be `null`

We need to test these scenarios:

1) Variable with certainty maybe and nullable
2) Variable with certainty maybe and not nullable
3) Variable with certainty yes and nullable
4) Variable with certainty yes and not nullable

For `!isset($foo['bar'])`, it means:

1) The key does not exist
2) The key exists but it's null

So various scenarious with `array{foo: int}` and `array{foo: int|null}` and also `array{foo?: int}` and `array{foo?: int|null}` need to be tested.

Plus for `!isset($foo['bar'])` it might also mean that anything in the chain might also not exist so the first 4) points about `!isset($foo)` apply to `!isset($foo['bar'])`.

I think the main hurdle here is that `SpecifiedTypes` currently cannot express a command to unset a variable. Because for `$foo` with certainty maybe and not-nullable, the only `!isset($foo)` is that the variable does not exist, so the certainty needs to be `no`.

But we can introduce some kind of `UnsetExpr` virtual node that would be in SpecifiedTypes and would instruct `MutatingScope::filterBySpecifiedTypes` to unset the variable from scope. That might or might not be the way to go.

tests/PHPStan/Analyser/data/falsy-isset.php
```diff
@@ -0,0 +1,115 @@
...
+	{
+		if (rand() % 2) {
+			$a = ['bar' => 'world'];
+			if (rand() % 3) {
+				$a = ['bar' => 'hello'];
+			}
+		}
+		assertVariableCertainty(TrinaryLogic::createMaybe(), $a);
+
+		assertType("array{bar?: 'hello'}|array{bar?: 'world'}", $a);
```
staabm: atm this errors with `"array{bar: 'hello'}|array{bar: 'world'}"` - [even before this PR](https://phpstan.org/r/304d20b1-4983-4cd8-8432-d394900e1a47).
not sure whether thats a separate pre existing problem or my expectation here is wrong

tests/PHPStan/Analyser/data/falsy-isset.php
```diff
@@ -0,0 +1,115 @@
...
+		if (isset($a['bar'])) {
+			assertVariableCertainty(TrinaryLogic::createYes(), $a);
+			assertType("array{bar: 'hello'}|array{bar: 'world'}", $a);
+			$a['bar'] = 1;
+			assertType("array{bar: 1}", $a);
+		} else {
+			assertVariableCertainty(TrinaryLogic::createNo(), $a);
+			assertType('*ERROR*', $a);
```
staabm, line 110: as requested, I am using certainty=no when certainty maybe & non-nullable -case.

this also turns the type after the if/else into a `*ERROR*` - I think because the type after if/else is determined by union of if- and else- branch, which because of union with NO

staabm: hmm maybe the NotIssetExpr should be handled like an early return, so it does not influence the type after the if/else?

src/Analyser/ExpressionTypeHolder.php
```diff
@@ -24,6 +24,11 @@ public static function createMaybe(Expr $expr, Type $type): self
 		return new self($expr, $type, TrinaryLogic::createMaybe());
 	}
 
+	public static function createNo(Expr $expr, Type $type): self
+	{
+		return new self($expr, $type, TrinaryLogic::createNo());
```
ondrejmirtes: This doesn't make sense. Instead of `no` just unset the entry in the array. This will also get rid of the error you commented about (the ErrorType problem).

src/Analyser/MutatingScope.php
```diff
@@ -3770,11 +3804,17 @@ public function filterByFalseyValue(Expr $expr): Scope
...
 	{
+		$scope = $this;
+
 		$typeSpecifications = [];
 		foreach ($specifiedTypes->getSureTypes() as $exprString => [$expr, $type]) {
 			if ($expr instanceof Node\Scalar || $expr instanceof Array_ || $expr instanceof Expr\UnaryMinus && $expr->expr instanceof Node\Scalar) {
 				continue;
 			}
+			if ($expr instanceof NotIssetExpr) {
+				$scope = $scope->notIssetExpression($expr);
```
ondrejmirtes: Just call `unsetExpression` method on MutatingScope.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
```diff
@@ -3803,11 +3806,32 @@ public function filterBySpecifiedTypes(SpecifiedTypes $specifiedTypes): self
...
 		});
 
-		$scope = $this;
 		$specifiedExpressions = [];
 		foreach ($typeSpecifications as $typeSpecification) {
 			$expr = $typeSpecification['expr'];
 			$type = $typeSpecification['type'];
+
+			if ($expr instanceof NotIssetExpr) {
+				$unsetExpr = $expr->getExpr();
```
ondrejmirtes: Do this only for 'sure'=true.

ondrejmirtes (changes requested):
Also please write regression tests for all the issues the issue bot found (after the first original iteration), to make sure we don't break them during the review.

src/Analyser/MutatingScope.php
staabm: moving it into the `'sure'=true` branch regresses some tests. I think what you want to achieve is already taken care of in the `TypeSpecifier` side of this PR, as `NotIssetExpr` is only passed thru in the [`!$context->true()`](https://github.com/phpstan/phpstan-src/pull/2657/files#diff-1076e38598f300a69c9dc998dae31374052f95bd69d0d71c11e9d0a4c188976dR678)-case

staabm:
I am bit a lost atm .. will take a break and look again at it later.

src/Analyser/TypeSpecifier.php
```diff
@@ -680,14 +680,81 @@ public function specifyTypesInCondition(
...
+						continue;
+					}
+
+					if (TypeCombinator::containsNull($offsetType)) {
+						$nullOffsetArray = ConstantArrayTypeBuilder::createEmpty();
+						$nullOffsetArray->setOffsetValueType($dimType, new NullType(), true);
+
+						$specifiedTypes = $specifiedTypes->unionWith($this->create(
+							$var->var,
+							TypeCombinator::union($type, $nullOffsetArray->getArray()),
```
staabm: ok, this type merging is plain wrong

staabm:
the only remaining errors are 2 test expectations, which I am not sure wheter my expectations are wrong, or there is a open todo:

```
1) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Analyser/data/falsy-isset.php:84" ('variableCertainty', '/home/runner/work/phpstan-src...et.php', PHPStan\TrinaryLogic Object (...), PHPStan\TrinaryLogic Object (...), 'a', 84)
Expected Maybe, actual certainty of variable $a is Yes in /home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Analyser/data/falsy-isset.php on line 84.

/home/runner/work/phpstan-src/phpstan-src/src/Testing/TypeInferenceTestCase.php:110
/home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:1369

2) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Analyser/data/falsy-isset.php:88" ('variableCertainty', '/home/runner/work/phpstan-src...et.php', PHPStan\TrinaryLogic Object (...), PHPStan\TrinaryLogic Object (...), 'a', 88)
Expected Maybe, actual certainty of variable $a is Yes in /home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Analyser/data/falsy-isset.php on line 88.
```

src/Analyser/TypeSpecifier.php
```diff
@@ -680,14 +681,100 @@ public function specifyTypesInCondition(
...
+								$builder->setOffsetValueType(
+									$dimType,
+									new NullType(),
+									true,
+								);
+								return $builder->getArray();
+							}
+
+							return $type->setOffsetValueType($dimType, new NullType());
+						});
```
staabm: instead of this `TypeTraverser` block, we could add a `optional` new arg to `Type->setOffsetValueType()`

src/Analyser/TypeSpecifier.php
```diff
@@ -680,14 +681,100 @@ public function specifyTypesInCondition(
...
+
+						$specifiedTypes = $specifiedTypes->unionWith($this->create(
+							$var->var,
+							$newType,
+							$context->negate(),
+							true,
```
staabm: I think there is no way arround `override=true` here, because we do not intersect/union the previous known array-shape, but just update a single offset of it.

src/Analyser/MutatingScope.php
```diff
@@ -3803,11 +3806,32 @@ public function filterBySpecifiedTypes(SpecifiedTypes $specifiedTypes): self
...
+					$arraySize = $type->getArraySize();
+
+					if (
+						$arraySize instanceof ConstantIntegerType
+						&& $arraySize->getValue() === 0
+					) {
+						$exprString = $this->getNodeKey($unsetExpr->var);
+
+						unset($scope->expressionTypes[$exprString]);
+						unset($scope->nativeExpressionTypes[$exprString]);
```
ondrejmirtes: I don't understand this logic at all. Here you're saying that if we `unset($var['foo'])` and the `$var` is an empty array, you're going to unset `$var` !?

ondrejmirtes (changes requested):
Please solve not only `ArrayDimFetch` but also standalone variables that I talk about here https://github.com/phpstan/phpstan-src/pull/2657#pullrequestreview-1651857463

src/Analyser/MutatingScope.php
staabm: this is the implementation part of [the tests which fail atm](https://github.com/phpstan/phpstan-src/pull/2657#issuecomment-1747523369)

the idea is to implement [what was described](https://github.com/phpstan/phpstan-src/pull/2657#pullrequestreview-1651857463) in

> Because for $foo with certainty maybe and not-nullable, the only !isset($foo) is that the variable does not exist, so the certainty needs to be no.

let me fix it

staabm:
just pushed the implementation for variables and a bunch of tests.

main problem is still the certainty stuff, and I think most - if not all - remaining test-failures feel related to certainty

---

and mooooore regression tests

src/Analyser/MutatingScope.php
staabm: I am doing this array stuff here, because I think:

a array shape which only has a single offset, which gets checked with isset() on said offset will  never have the variable defined in the falsey-branch

```php
$a = ['foo' => 'abc];
if (isset($a['foo'])) {
} else {
  // $a is always undefined 
}

```

staabm:
I have an idea how to fix the remaining problems, but just need a bit more time.
I will ping you when things are ready for another review

staabm:
@ondrejmirtes I did a lot of debugging locally and have some changes planned, but I think the main problem I am facing atm is, that when the TypeSpecifier is invoked for the `Isset_` expression, the falsey-scope no longer knows the `maybe` certainty of `$a`.

the scope in the falsey-branch already thinks `$a` has a `yes` certainty [because of processing happing in NodeScopeResolver](https://github.com/phpstan/phpstan-src/blob/2f2b07bb3b9ef979f48cd977075e9b005c60b14e/src/Analyser/NodeScopeResolver.php#L2637-L2645), before `TypeSpecifier Isset_` processing kicks in

example:

```php

function maybeCertainNull(): void
{
	if (rand() % 2) {
		$a = ['bar' => null];
		if (rand() % 3) {
			$a = ['bar' => 'hello'];
		}
	}

	assertVariableCertainty(TrinaryLogic::createMaybe(), $a);
	if (isset($a['bar'])) {
		assertVariableCertainty(TrinaryLogic::createYes(), $a);
	} else {
		assertVariableCertainty(TrinaryLogic::createMaybe(), $a); // Expected variable certainty Maybe, actual: Yes 
	}

	assertVariableCertainty(TrinaryLogic::createMaybe(), $a);
}
``` 

thats a problem we have already before this PR, and I don't know whether fixing it here with all the other changes needed would be a good thing as this PR will be already pretty large without this topic.

do you have an idea how to fix that problem? I don't have time right now to also dive into this new side-quest.

staabm:
after a break and a short walk, I realized how it could be fixed: see https://github.com/phpstan/phpstan-src/pull/2666

src/Analyser/MutatingScope.php
staabm: since we don't want to go from YES to NO certainty,  I think my above case is not possible and I should not try to support it?

ondrejmirtes:
```php
$a = ['foo' => 'abc'];
if (isset($a['foo'])) {
} else {
  // $a should be NeverType here
}
```

staabm:
I feel I am close to something useful :-)

[one TypeSpecifierTest](https://github.com/phpstan/phpstan-src/blob/cebf0d3b24eea6ac0a26db80a045883e6478fd06/tests/PHPStan/Analyser/TypeSpecifierTest.php#L615-L623) is failling, because [when merging left/right types](https://github.com/phpstan/phpstan-src/blob/cebf0d3b24eea6ac0a26db80a045883e6478fd06/src/Analyser/TypeSpecifier.php#L645-L648) the `SpecifiedTypes` intersection/union does not realize that `__phpstanIssetExpr($array)` and `$array` belong to the same expression and need to be merged

grafik

ondrejmirtes:
I''d say that when you're creating new IssetExpr in TypeSpecifier, you also need to say at the same time, what it means for the expression, and specify its type too in the same operation.

ondrejmirtes:
Maybe start a new PR instead of reviving this ancient one :)

PR #2665, opened 2023-10-08
Add more functions to register hasSideEffects => true

samsonasik:
The following functions has side effects:

```
stream_filter_append
ob_get_clean
openssl_sign
```

### DIFF
--- a/bin/functionMetadata_original.php
+++ b/bin/functionMetadata_original.php
@@ -101,6 +101,7 @@ return [
 	'rewind' => ['hasSideEffects' => true],
 	'rmdir' => ['hasSideEffects' => true],
 	'sprintf' => ['hasSideEffects' => false],
+	'stream_filter_append' => ['hasSideEffects' => true],
 	'symlink' => ['hasSideEffects' => true],
 	'tempnam' => ['hasSideEffects' => true],
 	'tmpfile' => ['hasSideEffects' => true],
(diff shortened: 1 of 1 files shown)


staabm:
could you add a NodeScopeResolverTest regression test?
(especially for `ob_get_clean` which is widely used I think)

samsonasik:
I will try when I have a chance.

I think between phpstan and psalm can complete each other side effect function lists, if we can have a script to include and mix both, that will be great for both sides.

bin/functionMetadata_original.php
```diff
@@ -114,6 +115,9 @@
 	'random_bytes' => ['hasSideEffects' => true],
 	'random_int' => ['hasSideEffects' => true],
 
+	'ob_get_clean' => ['hasSideEffects' => true],
```
staabm, line 118: @samsonasik you can use something like https://phpstan.org/r/9632df30-9ef5-4c50-b070-32972826e7d4 to test this change in a `NodeScopeResolverTest`

the dumped type should be `false|string`

samsonasik: `ob_start()` is required to make a `false|string` https://phpstan.org/r/1cff680a-416e-4ffb-adaf-db73a67ef759

staabm: ob_start() could be anywhere, its not relevant for type analysis. Its global state. 

https://phpstan.org/r/a07431d0-a49d-47ff-bd7e-e432fec503f7

samsonasik:
I am closing for now as I don't have a chance to update test yet.

PR #2666, opened 2023-10-08
Fix variable certainty for ArrayDimFetch in falsey isset context

staabm:
triggered by https://github.com/phpstan/phpstan-src/pull/2657#issuecomment-1751951154

----

btw: I realized that - across the PHPStan codebase - we are sometimes use `falsey` and `falsy` which is something we could fix with PHPStan 2.0

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3478,7 +3478,7 @@ class MutatingScope implements Scope
 							new HasOffsetValueType($dimType, $type),
 						),
 						$scope->getNativeType($expr->var),
-						$scope->hasExpressionType($expr->var), // keep certainty
+						$certainty,
 					);
 				}
 			}
@@ -3486,6 +3486,8 @@ class MutatingScope implements Scope
 
 		if ($certainty === null) {
 			$certainty = TrinaryLogic::createYes();
+		} elseif ($certainty->no()) {
+			throw new ShouldNotHappenException();
 		}
 
 		$exprString = $this->getNodeKey($expr);
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2340,6 +2340,10 @@ class MutatingScope implements Scope
 	/** @api */
 	public function hasExpressionType(Expr $node): TrinaryLogic
 	{
+		if ($node instanceof Variable && is_string($node->name)) {
+			return $this->hasVariableType($node->name);
+		}
+
 		$exprString = $this->getNodeKey($node);
 		if (!isset($this->expressionTypes[$exprString])) {
 			return TrinaryLogic::createNo();
@@ -3467,9 +3471,11 @@ class MutatingScope implements Scope
 						$types[] = new StringType();
 					}
 
-					$variableCertainty = null;
-					if ($expr->var instanceof Variable && is_string($expr->var->name)) {
-						$variableCertainty = $scope->hasVariableType($expr->var->name);
+					// keep certainty
+					$certainty = null;
+					$hasExpressionType = $scope->hasExpressionType($expr->var);
+					if (!$hasExpressionType->no()) {
+						$certainty = $hasExpressionType;
 					}
 
 					$scope = $scope->specifyExpressionType(
@@ -3479,7 +3485,7 @@ class MutatingScope implements Scope
 							new HasOffsetValueType($dimType, $type),
 						),
 						$scope->getNativeType($expr->var),
-						$variableCertainty, // keep variable certainty
+						$certainty,
 					);
 				}
 			}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3473,8 +3473,9 @@ class MutatingScope implements Scope
 
 					// keep certainty
 					$variableCertainty = null;
-					if ($expr->var instanceof Variable && is_string($expr->var->name)) {
-						$variableCertainty = $scope->hasVariableType($expr->var->name);
+					$hasExpressionType = $scope->hasExpressionType($expr->var);
+					if (!$hasExpressionType->no()) {
+						$variableCertainty = $hasExpressionType;
 					}
 
 					$scope = $scope->specifyExpressionType(
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3471,13 +3471,6 @@ class MutatingScope implements Scope
 						$types[] = new StringType();
 					}
 
-					// keep certainty
-					$variableCertainty = null;
-					$hasExpressionType = $scope->hasExpressionType($expr->var);
-					if (!$hasExpressionType->no()) {
-						$variableCertainty = $hasExpressionType;
-					}
-
 					$scope = $scope->specifyExpressionType(
 						$expr->var,
 						TypeCombinator::intersect(
@@ -3485,7 +3478,7 @@ class MutatingScope implements Scope
 							new HasOffsetValueType($dimType, $type),
 						),
 						$scope->getNativeType($expr->var),
-						$variableCertainty,
+						$scope->hasExpressionType($expr->var), // keep certainty
 					);
 				}
 			}
@@ -3493,8 +3486,6 @@ class MutatingScope implements Scope
 
 		if ($certainty === null) {
 			$certainty = TrinaryLogic::createYes();
-		} elseif ($certainty->no()) {
-			throw new ShouldNotHappenException();
 		}
 
 		$exprString = $this->getNodeKey($expr);
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3394,6 +3394,50 @@ class MutatingScope implements Scope
 		return $scope;
 	}
 
+	public function makeExpressionUncertain(Expr $expr): self
+	{
+		$exprString = $this->getNodeKey($expr);
+		if (!array_key_exists($exprString, $this->expressionTypes)
+			|| !array_key_exists($exprString, $this->nativeExpressionTypes)
+		) {
+			return $this;
+		}
+
+		$scope = $this;
+
+		$expressionTypes = $scope->expressionTypes;
+		$expressionTypes[$exprString] = new ExpressionTypeHolder(
+			$expressionTypes[$exprString]->getExpr(),
+			$expressionTypes[$exprString]->getType(),
+			TrinaryLogic::createMaybe(),
+		);
+		$nativeTypes = $scope->nativeExpressionTypes;
+		$nativeTypes[$exprString] = new ExpressionTypeHolder(
+			$nativeTypes[$exprString]->getExpr(),
+			$nativeTypes[$exprString]->getType(),
+			TrinaryLogic::createMaybe(),
+		);
+
+		return $this->scopeFactory->create(
+			$this->context,
+			$this->isDeclareStrictTypes(),
+			$this->getFunction(),
+			$this->getNamespace(),
+			$expressionTypes,
+			$nativeTypes,
+			$this->conditionalExpressions,
+			$this->inClosureBindScopeClasses,
+			$this->anonymousFunctionReflection,
+			$this->inFirstLevelStatement,
+			$this->currentlyAssignedExpressions,
+			$this->currentlyAllowedUndefinedExpressions,
+			$this->inFunctionCallsStack,
+			$this->afterExtractCall,
+			$this->parentScope,
+			$this->nativeTypesPromoted,
+		);
+	}
+
 	public function unsetExpression(Expr $expr): self
 	{
 		$scope = $this;
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3394,50 +3394,6 @@ class MutatingScope implements Scope
 		return $scope;
 	}
 
-	public function makeVariableUncertain(Variable $expr): self
-	{
-		$exprString = $this->getNodeKey($expr);
-		if (!array_key_exists($exprString, $this->expressionTypes)
-			|| !array_key_exists($exprString, $this->nativeExpressionTypes)
-		) {
-			return $this;
-		}
-
-		$scope = $this;
-
-		$expressionTypes = $scope->expressionTypes;
-		$expressionTypes[$exprString] = new ExpressionTypeHolder(
-			$expressionTypes[$exprString]->getExpr(),
-			$expressionTypes[$exprString]->getType(),
-			TrinaryLogic::createMaybe(),
-		);
-		$nativeTypes = $scope->nativeExpressionTypes;
-		$nativeTypes[$exprString] = new ExpressionTypeHolder(
-			$nativeTypes[$exprString]->getExpr(),
-			$nativeTypes[$exprString]->getType(),
-			TrinaryLogic::createMaybe(),
-		);
-
-		return $this->scopeFactory->create(
-			$this->context,
-			$this->isDeclareStrictTypes(),
-			$this->getFunction(),
-			$this->getNamespace(),
-			$expressionTypes,
-			$nativeTypes,
-			$this->conditionalExpressions,
-			$this->inClosureBindScopeClasses,
-			$this->anonymousFunctionReflection,
-			$this->inFirstLevelStatement,
-			$this->currentlyAssignedExpressions,
-			$this->currentlyAllowedUndefinedExpressions,
-			$this->inFunctionCallsStack,
-			$this->afterExtractCall,
-			$this->parentScope,
-			$this->nativeTypesPromoted,
-		);
-	}
-
 	public function unsetExpression(Expr $expr): self
 	{
 		$scope = $this;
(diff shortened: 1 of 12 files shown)


ondrejmirtes:
Please reproduce and fix the same problem with Empty_.

staabm:
hmm interessting.. it does not reproduce with `Empty_`

staabm:
Ohh even more interessting.. it does not reproduce in a NodeScopeResolverTest, but running the same code directly with phpstan in a separate file reproduces the problem for `Empty_` -> added a new AnalyserIntegrationTest

tests/PHPStan/Analyser/data/falsey-isset-certainty.php
```diff
@@ -0,0 +1,79 @@
+<?php
```
ondrejmirtes, line 1: This test is not referenced from anywhere

src/Analyser/NodeScopeResolver.php
```diff
@@ -2936,12 +2940,16 @@ static function (Node $node, Scope $scope) use ($nodeCallback): void {
 			$throwPoints = [];
 		}
 
+		if ($falsyScope === null) {
+			$falsyScope = $scope;
```
ondrejmirtes: It'd make more sense to just do `return new ExpressionResult(` in the `Isset_` branch instead of declare and use a variable thousands of lines from each other.

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1230,6 +1230,12 @@ public function testBug9690(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testFalseyEmpty(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/falsey-empty-certainty.php');
```
ondrejmirtes: I'd rather reference this in NodeScopeResolverTest instead.

ondrejmirtes (changes requested):

staabm:
after debugging deep in PHPStan internals, I think I am able to pinpoint the underlying issue.

while [handling `Isset_`](https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Analyser/NodeScopeResolver.php#L2637-L2651) or [`Empty_`](https://github.com/phpstan/phpstan-src/blob/a6742b70cb92e402236e14554829df1fb701b73a/src/Analyser/NodeScopeResolver.php#L2625-L2632) there is some expression types storing and re-storing going on.

while this process a variable which was "maybe" before automatically got a "yes"-certainty, because assigning a type to a variable automatically made it certain.

I fixed this mechanism to also remember the previous certainty of the variable, not just the type and native-type.

staabm:
regarding build errors

---

Carbon:

```
Error: Variable $date might not be defined.
 ------ -------------------------------------- 
  Line   tests/CarbonPeriod/ToArrayTest.php    
 ------ -------------------------------------- 
  100    Variable $date might not be defined.  
 ------ -------------------------------------- 
```

https://github.com/briannesbitt/Carbon/blob/0af369fba2435afe17a97cfa9a02a8e84d6c6640/tests/CarbonPeriod/ToArrayTest.php#L95-L100

I think this error could be valid, as we are probably not narrow the type of `$date` via `assertInstanceOfCarbon()` - which sounds like a carbon specific unit test utility to me

----

the `efabrica-team/phpstan-latte` error look wild.. I wonder whether this are related.

----

drupal

```
 ------ --------------------------------------------- 
  Line   core/modules/shortcut/shortcut.module        
 ------ --------------------------------------------- 
  294    Variable $shortcut_id might not be defined.  
  297    Variable $shortcut_id might not be defined.  
 ------ --------------------------------------------- 
 ```

https://github.com/drupal/drupal/blob/2700c5afb6c3936041db413872eea82dc0bd4fe4/core/modules/shortcut/shortcut.module#L280-L298

these look valid to me. it seems we are now detecting these mabye undefined while we were missing them before

```
------ ----------------------------------------------------------------------------------------------------------------------------- 
  Line   core/modules/views/src/Plugin/views/filter/NumericFilter.php                                                                 
 ------ ----------------------------------------------------------------------------------------------------------------------------- 
  426    Ignored error pattern #^Variable \$value might not be defined\.$# in                                                         
         path                                                                                                                         
         /home/runner/work/phpstan-src/phpstan-src/e2e/integration/repo/core/modules/views/src/Plugin/views/filter/NumericFilter.php  
         is expected to occur 2 times, but occurred 3 times.                                                                          
  432    Variable $value might not be defined.                                                                                        
 ------ ----------------------------------------------------------------------------------------------------------------------------- 
 ```

https://github.com/drupal/drupal/blob/2700c5afb6c3936041db413872eea82dc0bd4fe4/core/modules/views/src/Plugin/views/filter/NumericFilter.php#L423-L438

also looks valid to me

ondrejmirtes:
Did you look through the errors please?

* https://github.com/phpstan/phpstan-src/actions/runs/6453872440/job/17518378375
* https://github.com/phpstan/phpstan-src/actions/runs/6453872440/job/17518379195

staabm:
I did please see the comment above. I think our comments might be overlapped

ondrejmirtes:
Sorry, if there are 21 emails in the thread, I tend to overlook things :)

ondrejmirtes:
So if we're currently preserving the certainty, how does PHPStan go from "maybe" to "yes" in this case?

```php
// $foo is Maybe
if (isset($foo)) {
    // $foo is Yes
}
```

staabm:
excellent question. I had to debug further and look into it even more to answer it ;-).

we get from `maybe` to `yes` while we specify non-nullability in https://github.com/phpstan/phpstan-src/blob/6252d21705b2f863674a9596878aebe1fdb9f393/src/Analyser/TypeSpecifier.php#L748

just pushed a new commit which refactors this code a bit to make it more obvious

-----

another interessting find: it seems we stop specifying types in `isset(...)` as soon as we encouter the [first expression which is undefined](https://github.com/phpstan/phpstan-src/blob/6252d21705b2f863674a9596878aebe1fdb9f393/src/Analyser/TypeSpecifier.php#L720)

tests/PHPStan/Rules/Variables/IssetRuleTest.php
```diff
@@ -450,4 +450,21 @@ public function testObjectShapes(): void
 		$this->analyse([__DIR__ . '/data/isset-object-shapes.php'], []);
 	}
 
+	public function testBug3985(): void
+	{
+		$this->treatPhpDocTypesAsCertain = true;
+		$this->strictUnnecessaryNullsafePropertyFetch = true;
+
+		$this->analyse([__DIR__ . '/../../Analyser/data/bug-3985.php'], [
```
staabm, line 458: I decided to add new assertions for [this old bug](https://github.com/phpstan/phpstan/issues/3985), while in the process of understanding the internals better
(IMO the new assertions better describe what the end user expected)

ondrejmirtes:
Yeah, please clean this up as much as possible, and don't forget that both `empty` and `??` are very similar to `isset` and might have similar problems.

ondrejmirtes:
> we get from maybe to yes

I'd still like to get from `maybe` to `yes` for other specifiers, like `is_string($foo)` - if it's a string then we know the certainty is `yes`.

staabm:
I see. do you talk about new features or something I need to make sure we don't regress?

ondrejmirtes:
Not new features, but about this: https://phpstan.org/r/5f6093e0-0f6e-405a-9497-54263c6348b7

I worry that if you preserve certainty, there will be new error on line 12.

staabm:
should we not have the error on line 15 or should we have one on line8?

https://phpstan.org/r/d3ba81f8-188e-47ea-805e-54302006bc7a

ondrejmirtes:
We can't go from no to yes and from yes to no.

ondrejmirtes:
Also an idea: in the other PR you introduced `NotIssetExpr`.

I think it should really be `IssetExpr` and applied in both truthy and falsy contexts.

staabm:
> We can't go from no to yes and from yes to no.

lets see if I understood you right ;)

https://github.com/phpstan/phpstan-src/pull/2666/commits/b1981b022b919cd0bc291d8a93b1f2357a53c5bd

> Also an idea: in the other PR you introduced `NotIssetExpr`.
> 
> I think it should really be `IssetExpr` and applied in both truthy and falsy contexts.

I am fine with it. lets discuss it in the other PR when this one is merged

src/Analyser/NodeScopeResolver.php
```diff
@@ -1740,14 +1740,21 @@ private function ensureShallowNonNullability(MutatingScope $scope, Scope $origin
 		if ($isNull->yes()) {
 			return new EnsuredNonNullabilityResult($scope, []);
 		}
+
+		// keep variable certainty
+		$certainty = TrinaryLogic::createYes();
+		if ($exprToSpecify instanceof Variable && is_string($exprToSpecify->name)) {
+			$certainty = $originalScope->hasVariableType($exprToSpecify->name);
```
ondrejmirtes: There should be nothing special about `Variable`. Use `hasExpressionType`. And also if inside `hasExpressionType` if you detect that `$node` is `Variable`, call `hasVariableType` inside because there's some special handling there. But from the outside always call `hasExpressionType`.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
```diff
@@ -3466,23 +3470,37 @@ public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType):
 					if ($dimType instanceof ConstantIntegerType) {
 						$types[] = new StringType();
 					}
+
+					// keep certainty
+					$variableCertainty = null;
+					if ($expr->var instanceof Variable && is_string($expr->var->name)) {
+						$variableCertainty = $scope->hasVariableType($expr->var->name);
```
staabm: I tried using `hasExpressionType()` here, but it did not work for cases like

```php

use PHPStan\PhpDocParser\Ast\Type\GenericTypeNode;
use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;

class TypeNodeResolverRepro
{
	private function resolveGenericTypeNode(GenericTypeNode $typeNode): void
	{
		if ($typeNode->genericTypes[0] instanceof IdentifierTypeNode && $typeNode->genericTypes[0]->name === 'min') {
			$min = null;
		}

		if ($typeNode->genericTypes[1] instanceof IdentifierTypeNode && $typeNode->genericTypes[1]->name === 'max') {
		}
	}

}
```

which errored with

> Access to an undefined property PHPStan\PhpDocParser\Ast\Type\TypeNode::$name.

staabm: fixed this in the latest commit. mistakenly re-used the `$certainty` variable within this if block, which overwrote the same named parameter

tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
```diff
@@ -102,6 +102,10 @@ public function testDefinedVariables(): void
 				'Undefined variable: $variableInEmpty',
 				145,
 			],
+			[
+				'Undefined variable: $negatedVariableInEmpty',
```
staabm, line 106: @ondrejmirtes one last thing I am wondering:

https://phpstan.org/r/60a66a37-7ea8-45d6-a0db-3de562d5b88d

before this PR a undefied variable in `!empty()` had a YES but in `isset()` we kept the NO certainty (see error on line 9).
[since you said a NO can't get into YES and vice-versa](https://github.com/phpstan/phpstan-src/pull/2666#issuecomment-1752735049), I think this was a bug before, and the error we are now getting here is expected.


----

except this one thing, I think this PR is good to go.

src/Analyser/MutatingScope.php
```diff
@@ -3466,23 +3470,38 @@ public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType):
 					if ($dimType instanceof ConstantIntegerType) {
 						$types[] = new StringType();
 					}
+
+					// keep certainty
+					$variableCertainty = null;
+					$hasExpressionType = $scope->hasExpressionType($expr->var);
+					if (!$hasExpressionType->no()) {
+						$variableCertainty = $hasExpressionType;
```
ondrejmirtes: This looks a bit fishy. Why is the parameter handled different than for other expression types? I don't like that for ArrayDimFetch, you're basically throwing away what the user passed in when calling specifyExpressionType.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -948,7 +948,9 @@ public function testBug7581(): void
 	public function testBug7903(): void
 	{
 		$errors = $this->runAnalyse(__DIR__ . '/data/bug-7903.php');
-		$this->assertNoErrors($errors);
+		$this->assertCount(2, $errors);
+		$this->assertSame('Comparison operation ">" between (0|(non-empty-array<int|string, 0|(non-empty-array<literal-string&non-falsy-string, 0|float>)|float>)|float) and 0 results in an error.', $errors[0]->getMessage());
```
staabm: before [this commit](https://github.com/phpstan/phpstan-src/pull/2666/commits/5b5aa17b2c811aeb6aead3e86d5385d3969d5f24), the [`$ffffffjjjjjjccccc` variable at line 214](https://github.com/phpstan/phpstan-src/blob/b1d12a45af16a7c5df61069f713f9537fc6ac8c1/tests/PHPStan/Analyser/data/bug-7903.php#L214) was just mixed. after this commit we know more precisely its a oversized array

src/Analyser/MutatingScope.php
```diff
@@ -3466,23 +3470,29 @@ public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType):
...
 					}
+
 					$scope = $scope->specifyExpressionType(
 						$expr->var,
 						TypeCombinator::intersect(
 							TypeCombinator::intersect($exprVarType, TypeCombinator::union(...$types)),
 							new HasOffsetValueType($dimType, $type),
 						),
 						$scope->getNativeType($expr->var),
+						$scope->hasExpressionType($expr->var), // keep certainty
```
ondrejmirtes: Nope, this still isn't consistent. Just pass `$certainty` in there.

src/Analyser/MutatingScope.php
```diff
@@ -3466,23 +3470,29 @@ public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType):
...
 					);
 				}
 			}
 		}
 
+		if ($certainty === null) {
+			$certainty = TrinaryLogic::createYes();
```
ondrejmirtes, line 3488: Please reject `->no()` values, they don't make sense here. `assignVariable` also does it.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you! Please continue by rebasing and finishing https://github.com/phpstan/phpstan-src/pull/2657.

ondrejmirtes:
Hi @staabm this PR caused failures in Drupal and Carbon:

* https://github.com/phpstan/phpstan/actions/runs/6697009618/job/18196030092
* https://github.com/phpstan/phpstan/actions/runs/6697009618/job/18196030700

Fortunately it's not released yet. Please look into them otherwise I'll have to revert this. Thanks.

staabm:
I will have a look 👀

staabm:
I had a first look into drupal problem - [related source](https://github.com/drupal/drupal/blob/10.0.0/core/modules/book/book.module#L92-L128) - and reduced it to:

```php
<?php

function book_node_links_alter():void {
	if (isset($x)) {
		$book_links = 1;
	}

// Expected variable certainty Maybe, actual: No 
	assertVariableCertainty(\PHPStan\TrinaryLogic::createMaybe(), $book_links); 

	if (!empty($book_links)) {
		var_dump($book_links);
	}
}
```

I wonder why we have a NO certainty here.. is it expected?
its not something which regressed with this PR - its reproducible also with the latest stable release.

staabm:
another finding: in the latest 1.10 release in this snippet we went from No to Yes.

```php
function justEmpty(): void
{
	assertVariableCertainty(TrinaryLogic::createNo(), $foo);
	if (!empty($foo)) {
		assertVariableCertainty(TrinaryLogic::createYes(), $foo);
	}
	assertVariableCertainty(TrinaryLogic::createNo(), $foo);
}
```

this was changed with this PR here, but regressed the snippet in the above comment.

staabm:
to sum up my investigation of all [mentioned failures caused by this PR](https://github.com/phpstan/phpstan-src/pull/2666#issuecomment-1792687552):

IMO there is only 1 regression, and [thats the one reduced above](https://github.com/phpstan/phpstan-src/pull/2666#issuecomment-1793114923).
I think we need to question whether we are sure about "never go from No to Yes" in certainty, or we need to compensate NO certainty in the rule which emits "Undefined variable: $book_links" when nested in [empty()](https://phpstan.org/r/d0eaa827-e6d9-40ca-a02a-83cd5ea3d5a7) or [isset()](https://phpstan.org/r/76b6ed99-883d-41e0-9f59-c7ec176edced)

all remaining new failures are valid

PR #2670, opened 2023-10-09
Added text regarding patch issues during installation

mt-satak:
## Why did I submit this PR?
When performing `composer install`, the patch to `hoa/protocol` failed to apply.
```
  - Applying patches for hoa/protocol
    patches/Node.patch (0)
   Could not apply patch! Skipping. The error was: The process "patch '-p1' --no-backup-if-mismatch -d '/Users/satak/dev/phpstan-src/vendor/hoa/protocol' < '/Users/satak/dev/phpstan-src/patches/Node.patch'" exceeded the timeout of 300 seconds.

In Patches.php line 331:
                                              
  Cannot apply patch 0 (patches/Node.patch)! 
```

This error is caused by an out-of-date version of patch pre-installed on macOS, and can be resolved by doing `brew install gpatch` before running `composer install` to bring it up to date.

## Purpose of this PR
To add the above solution to the README to leave a clue for solving similar errors when they occur.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,7 @@ composer install
 
 If you encounter dependency problem, try using `export COMPOSER_ROOT_VERSION=1.10.x-dev`
 
-If you are using macOS and are using an older version of patch, you may have problems with patch application failure.
-Try using `brew install gpatch`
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

PR #2673, opened 2023-10-10
Respect proper case for namespaced functions

fredden:
Fixes https://github.com/phpstan/phpstan/issues/10003

### DIFF
--- a/src/Reflection/SignatureMap/NativeFunctionReflectionProvider.php
+++ b/src/Reflection/SignatureMap/NativeFunctionReflectionProvider.php
@@ -35,6 +35,7 @@ class NativeFunctionReflectionProvider
 	public function findFunctionReflection(string $functionName): ?NativeFunctionReflection
 	{
 		$lowerCasedFunctionName = strtolower($functionName);
+		$realFunctionName = $lowerCasedFunctionName;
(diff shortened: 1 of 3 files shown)


fredden:
It looks like the tests on Windows aren't working because the `mongodb` PHP extension isn't available there. Is there an  extension which can be used there which also demonstrates the bug?

ondrejmirtes:
Thank you, I reused your regression test but did a different and correct fix. See the commit history in this PR: https://github.com/phpstan/phpstan-src/pull/2673/commits/0db231783c725b3917df616e40a0930e535b4d3f

PHPStan cannot rely on runtime reflection because it's often wrong, and PHPStan also doesn't require PHP extensions to be loaded for successful analysis.

The fix I did will use the name from BetterReflection which takes it from jetbrains/phpstorm-stubs where it's correct: https://github.com/JetBrains/phpstorm-stubs/blob/master/mongodb/Monitoring/functions.php

fredden:
Thanks for the quick turn-around on this @ondrejmirtes! Yes, that fix looks a lot cleaner than what I came up with. I had looked through that function while investigating the bug, but hadn't spotted that the real name was retrievable from there.

PR #2675, opened 2023-10-12
Infer object type mysqli_fetch_object

staabm:


resources/functionMap.php
```diff
@@ -7472,7 +7472,7 @@
 'mysqli_fetch_field_direct' => ['object|false', 'result'=>'mysqli_result', 'fieldnr'=>'int'],
 'mysqli_fetch_fields' => ['array', 'result'=>'mysqli_result'],
 'mysqli_fetch_lengths' => ['array|false', 'result'=>'mysqli_result'],
-'mysqli_fetch_object' => ['object|null', 'result'=>'mysqli_result', 'class_name='=>'string', 'params='=>'?array'],
+'mysqli_fetch_object' => ['object|false|null', 'result'=>'mysqli_result', 'class_name='=>'class-string', 'params='=>'?array'],
```
staabm: seem we were missing the possible `false` type in php 7.2/7.3.

in newer versions it was already valid because of  php8-stubs

https://www.php.net/mysqli_fetch_object

ondrejmirtes: Don’t sneak in narrower parameter types like this 😊

ondrejmirtes (changes requested):

resources/functionMap.php
staabm: ahh good point. just dropped the param type change

staabm:
fixed conflicts

ondrejmirtes:
Thank you.

PR #2676, opened 2023-10-12
Don't collapse Traversable with php-src native traversable objects

staabm:
closes https://github.com/phpstan/phpstan/issues/10008

### DIFF
--- a/src/Type/Generic/GenericObjectType.php
+++ b/src/Type/Generic/GenericObjectType.php
@@ -23,9 +23,11 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeWithClassName;
 use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
+use Traversable;
 use function array_map;
 use function count;
 use function implode;
+use function in_array;
 use function sprintf;
 
 /** @api */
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/data/dba-traversables.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace DatabaseAccessTraversables;
+
+
+use mysqli_result;
+use PDOStatement;
+use Traversable;
+use function PHPStan\Testing\assertType;
+
+class HelloWorld
```
staabm, line 11: before this PR: https://phpstan.org/r/6c9c4317-0a4e-4c63-a06b-79865b9aaea3

src/Type/Generic/GenericObjectType.php
```diff
@@ -138,6 +140,17 @@ public function isSuperTypeOf(Type $type): TrinaryLogic
 			return $type->isSubTypeOf($this);
 		}
 
+		// don't collapse Traversable with classes which also implement Traversable to support phpstan-dba
+		// in inferring Traversable shapes onto php-src builtin classes
+		if ($this->getClassName() === Traversable::class) {
```
ondrejmirtes, line 145: You can't special-case this about `Traversable`, this needs to work generally.

ondrejmirtes (changes requested):

ondrejmirtes:
I tried a fix and it's hard: https://github.com/phpstan/phpstan/issues/10008#issuecomment-1761062790

PR #2680, opened 2023-10-14
Optimize memory consumption

staabm:
tested with 

`blackfire run --ignore-exit-status php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php`

before this PR: (1.10.x@7c4ef6b2a)
```
I/O Wait      171ms
CPU Time   2min 13s
Memory        368MB
Wall Time  2min 13s
Network         n/a     n/a     n/a
SQL             n/a     n/a
```

after this PR:
```
Wall Time  2min 16s
I/O Wait      183ms
CPU Time   2min 16s
Memory        345MB
Network         n/a     n/a     n/a
SQL             n/a     n/a
```

grafik

### DIFF
--- a/src/Analyser/ResultCache/ResultCacheManager.php
+++ b/src/Analyser/ResultCache/ResultCacheManager.php
@@ -10,8 +10,8 @@ use PHPStan\Collectors\CollectedData;
 use PHPStan\Command\Output;
 use PHPStan\Dependency\ExportedNodeFetcher;
 use PHPStan\Dependency\RootExportedNode;
+use PHPStan\File\CouldNotReadFileException;
 use PHPStan\File\FileFinder;
-use PHPStan\File\FileReader;
 use PHPStan\File\FileWriter;
 use PHPStan\Internal\ComposerHelper;
 use PHPStan\PhpDoc\StubFilesProvider;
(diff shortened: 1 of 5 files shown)


src/Reflection/BetterReflection/SourceLocator/OptimizedDirectorySourceLocatorFactory.php
```diff
@@ -42,12 +40,11 @@ public function createByDirectory(string $directory): NewOptimizedDirectorySourc
 		$files = $this->fileFinder->findFiles([$directory])->getFiles();
 		$fileHashes = [];
 		foreach ($files as $file) {
-			try {
-				$contents = FileReader::read($file);
```
ondrejmirtes, line 46: I think there are more places that do the same thing, please check ResultCacheManager. Or all the places that use FileReader. Thanks.

ondrejmirtes (changes requested):

src/Analyser/ResultCache/ResultCacheManager.php
```diff
@@ -789,10 +789,7 @@ private function getFileHash(string $path): string
 			return $this->fileHashes[$path];
 		}
 
-		$contents = FileReader::read($path);
-		$contents = str_replace("\r\n", "\n", $contents);
```
staabm, line 793: here was some normalization going on. removing it did not break tests.

I assume the result cache is not meant to be shared across different computers, therefore it seems not necessary

ondrejmirtes:
Thank you.

PR #2682, opened 2023-10-15
More efficient file finding

staabm:
symfony file-finder is constructing SplFileInfo under the hood which is costly.
using plain file iterators is more efficient

grafik

repro: `blackfire run --ignore-exit-status php bin/phpstan analyze src/Type/FileTypeMapper.php --debug`

before this PR 1.10.x@18a6723fa
```
Wall Time     45.9s
I/O Wait      110ms
CPU Time      45.8s
Memory        170MB
Network         n/a     n/a     n/a
SQL             n/a     n/a
```

after this PR
```
I/O Wait     53.5ms
CPU Time      35.4s
Memory        146MB
Wall Time     35.4s
Network         n/a     n/a     n/a
SQL             n/a     n/a
```

compare

grafik

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,6 @@
 		"react/socket": "^1.3",
 		"react/stream": "^1.1",
 		"symfony/console": "^5.4.3",
-		"symfony/finder": "^5.4.3",
 		"symfony/polyfill-intl-grapheme": "^1.23",
 		"symfony/polyfill-intl-normalizer": "^1.23",
 		"symfony/polyfill-mbstring": "^1.23",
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Hi, I think we're in the territory of diminishing returns, and I'm not willing to sacrifice code readability and maintainability for a few percent points. Also I'm a bit worried the new way is not going to behave exactly the same way, breaking peoples builds.

staabm:
I see. do you think reducing it to the very minimum parts which improve performance is the way to go, or should we even close right away?

ondrejmirtes:
If you want, you can try to improve symfony/finder itself, that would benefit the community a lot.

But I don't think the changes here are needed right now.

staabm:
while working thru the problem again and doing some blackfire profiles - and also measures without blackfire - I conclude there is no performance opportunity.

when running on plain php there is nearly no difference between symfony/finder and plain spl iterators.
[I wrote the case down](https://github.com/staabm/hungry-spl-file-info/tree/main) and reported it to blackfire.

PR #2687, opened 2023-10-20
fix switch(enum)

schlndh:
Fixes https://phpstan.org/r/da3c1077-5bcd-4ed7-80da-f8ce588f788e

Also fixes https://github.com/phpstan/phpstan/issues/3488#issuecomment-1372675975 and https://github.com/phpstan/phpstan/issues/9806

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -288,6 +288,7 @@ class TypeSpecifier
 				($leftType->isString()->yes() && $rightType->isString()->yes())
 				|| ($integerType->isSuperTypeOf($leftType)->yes() && $integerType->isSuperTypeOf($rightType)->yes())
 				|| ($floatType->isSuperTypeOf($leftType)->yes() && $floatType->isSuperTypeOf($rightType)->yes())
+				|| ($leftType->isEnum()->yes() && $rightType->isEnum()->yes())
 			) {
 				return $this->specifyTypesInCondition($scope, new Expr\BinaryOp\Identical($expr->left, $expr->right), $context, $rootExpr);
 			}
--- a/tests/PHPStan/Analyser/data/enums.php
+++ b/tests/PHPStan/Analyser/data/enums.php
@@ -309,6 +309,7 @@ class LooseComparisonWithEnums
 		assertType('false', false == Foo::ONE);
 		assertType('false', null == Foo::ONE);
 		assertType('false', Foo::ONE == null);
+		assertType('true', $foo == Foo::ONE || Foo::TWO == $foo);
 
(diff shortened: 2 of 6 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -288,6 +288,7 @@ public function specifyTypesInCondition(
 				($leftType->isString()->yes() && $rightType->isString()->yes())
 				|| ($integerType->isSuperTypeOf($leftType)->yes() && $integerType->isSuperTypeOf($rightType)->yes())
 				|| ($floatType->isSuperTypeOf($leftType)->yes() && $floatType->isSuperTypeOf($rightType)->yes())
+				|| ($leftType->isEnum()->yes() && $rightType->isEnum()->yes())
```
ondrejmirtes, line 291: Hi, I'm sorry I haven't reviewed this yet, but I saw this and have been thinking about it. I mostly like it but since the last time this code was touched, I introduced `Type::getFiniteTypes()` which is for these situations - when getFiniteTypes return a single object.

It works for enums, scalars, constant arrays. I wonder if this condition could be simplified with this method.

I'm also worried about many of the failures with the memory limit, but maybe it's not fault of this PR.

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
schlndh: Hi, thanks for the feedback - I didn't know about `getFiniteTypes`. However, I'm not sure that it is applicable here.

What I'm trying to express by this code is that `$a == $b` is equivalent to `$a === $b`, when both `$a` and `$b` are enums (actually it could be more general - https://3v4l.org/Huc7t). It doesn't matter whether it's an `EnumCaseObjectType` or `ObjectType` (maybe I'm just misunderstanding "when getFiniteTypes return a single object"). Furthermore, I want to avoid scalar types (e.g. `5 == "5"`).

I see that `count($t->getFiniteTypes()) === 1` is already used inside `resolveIdentical`. So my change does indirectly benefit from it.

schlndh: As for the memory limit: as far as I can see the memory limit is only exceeded for static analysis on PHP 7.2 (the other failures are due to a patch not applying on windows which I assume cannot be my fault). My guess is that it was borderline before and my adding a bit of code pushed it over the edge? I tried it with 500M and it worked: https://github.com/schlndh/phpstan-src/actions/runs/6603424059/job/17936514043

ondrejmirtes: You're right, it'd be bad for `5 == '5'`.

ondrejmirtes:
Thank you very much!

PR #2688, opened 2023-10-22
Remove Yaf framework signatures

staabm:
[Yaf is a framwork](https://github.com/sillydong/CZD_Yaf_Extension) with the last commit 2014 years ago and without a single release AFAIC

goal: reduce memory consumption

background analysis: https://github.com/phpstan/phpstan/issues/10039

----

phpstan-doctrine phpunit before

> Time: 00:35.267, Memory: 632.50 MB

after

> Time: 00:36.021, Memory: 626.50 MB


ondrejmirtes:
> [Yaf is a framwork](https://github.com/sillydong/CZD_Yaf_Extension) with the last commit 2014 years ago

You can be sure someone out there is using it, same as Zend 1 or Symfony 1. I'm gonna suggest something that will work even if the functionMap contains something only a handful of users use.

PR #2689, opened 2023-10-22
Remove swf extension signatures

staabm:
the swf extension is a tool for working with Adobe Flash files. it was used in php5 and early php7.
Support for Flash as a technology was cancelled by Adobe 31.12.2020

goal: reduce memory consumption

background analysis: https://github.com/phpstan/phpstan/issues/10039

----

phpstan-doctrine phpunit before

> Time: 00:36.021, Memory: 626.50 MB

after

> Time: 00:34.504, Memory: 618.00 MB


ondrejmirtes:
Closing for the same reason as in: https://github.com/phpstan/phpstan-src/pull/2688#issuecomment-1774069830

PR #2693, opened 2023-10-23
fix: remove `\` from DateTimeZone in functionMap

simPod:
This causes issue in thecodingmachine/safe as their generator takes phpstan's fcnmap and generates function signatures. However, now it generates e.g. `function date_create_immutable(string $datetime = "now", ?\\DateTimeZone $timezone = null): \DateTimeImmutable`. Therefore, this fix is needed. (https://github.com/thecodingmachine/safe/actions/runs/6613051619/job/17960125168?pr=432#step:18:1115)


ondrejmirtes:
> Therefore, this fix is needed.

Hi, I will merge this just to make everyone's life easier, but I disagree with the premise that the fix is really needed on PHPStan's side. PHPStan doesn't need this change and uses reflection data correctly.

Also other projects might benefit from combining multiple data sources as PHPStan does because `functionMap` does not always offer the full picture.

* https://github.com/phpstan/php-8-stubs
* https://github.com/phpstan/phpstan-src/tree/1.11.x/resources
* https://github.com/phpstan/phpstan-src/tree/1.11.x/stubs
* https://github.com/jetbrains/phpstorm-stubs

For example: PHPStan does not need to have delta files for PHP 8.3 because it already reads stuff that changed in PHP 8.3 from https://github.com/phpstan/php-8-stubs. So if `thecodingmachine/safe` reads these files and assumes that everything is correct for PHP 8.3, it's going to be wrong.

PR #2694, opened 2023-10-24
[Reflection] Update return types in some reflection methods

phansys:
Please, confirm I should I also update the existing stubs.
IIUC, stubs from PHPStorm already declare the proper [return types through attributes](https://github.com/JetBrains/phpstorm-stubs/blob/26462cb5c09c983cfd9beb78ec47b75f5cb8c162/Reflection/ReflectionFunctionAbstract.php#L248-L266) depending on the PHP version, but I guess they are not being parsed here.

See https://phpstan.org/r/85179264-493f-4b65-a8ff-4af5b45b31c1.


staabm:
I noticed you also [opened a PR at php-src](https://github.com/php/php-src/pull/12502).

just a note: the reported problem might get fixed, when php8-stubs include the upstream php-src change. we will see.

ondrejmirtes:
I disagree with these changes. These methods simply return `ReflectionType`. If you make your code work with that then you're better prepared for newer PHP versions.

PR #2695, opened 2023-10-24
Allow Redis::pconnect to have 7 parameters

Vaalyn:
This PR fixes a similar issue that was reported in https://github.com/phpstan/phpstan/issues/9413 and subsequently fixed in https://github.com/phpstan/phpstan-src/pull/2442.

The `Redis::pconnect` method also accepts 7 parameters like the `Redis::connect` method, currently using all 7 available parameters on the `pconnect` method will show the error:
```
Method Redis::pconnect() invoked with 7 parameters, 1-5 required.
```

As can be seen here on the [redis.stub.php](https://github.com/phpredis/phpredis/blob/a0c8fcc589bfcf8207ba3d8c7d065109f31cc1c6/redis.stub.php#L2243) the signature in the PHPStan function map is missing 2 parameters:
```
public function pconnect(string $host, int $port = 6379, float $timeout = 0, ?string $persistent_id = null, int $retry_interval = 0, float $read_timeout = 0, ?array $context = null): bool;
```

I've updated the function map to match the method signature.

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -9569,7 +9569,7 @@ return [
 'Redis::multi' => ['Redis', 'mode='=>'int'],
 'Redis::object' => ['string|long|false', 'info'=>'string', 'key'=>'string'],
 'Redis::open' => ['bool', 'host'=>'string', 'port='=>'int', 'timeout='=>'float', 'reserved='=>'null', 'retry_interval='=>'?int', 'read_timeout='=>'float'],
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I will merge this if you rebase it on top of 1.10.x.

Vaalyn:
@ondrejmirtes 👍 done

ondrejmirtes:
Thank you.

PR #572, opened 2021-07-06
Don't try to create a directory that already exists

PrinsFrank:
When creating directories, whenever those already exist mkdir is still called and reports an E_WARNING. Although it is supressed, those still end up in the error handler.

When this PR is merged, directories are checked first for existence. If so, mkdir is not called anymore, not triggering the E_WARNING.

PR #2697, opened 2023-10-26
[1.10.x] Create DirectoryCreator helper class

### LINKED ISSUE #992: [Bug]: Test fails on a PHPStan test
### What Happened

Hello!

I'm trying to write a test for a [custom PHPStan Rule](https://phpstan.org/developing-extensions/testing) and the test keeps yielding a warning when using `artisan test` / `pest`. However, the test passes just fine when using the `phpunit` test runner which makes me think this is an issue with pest.

Any ideas what might be causing this?

image

CC: @ondrejmirtes 

Thanks!

### How to Reproduce

- Install new laravel app
- Install PHPStan
- Create a custom PHPStan rule and test
- Try to run test with pest

### Sample Repository

_No response_

### Pest Version

2.24

### PHP Version

8.2.10

### Operation System

Linux

### Notes

_No response_

calebdw:
Hello!

This is similar to https://github.com/phpstan/phpstan-src/pull/572 --- every other `mkdir` call is guarded by an `is_dir` check so I updated this to be consistent.

Thanks!


ondrejmirtes:
As far as PHP is concerned, this code is equivalent. You don't need to check is_dir before calling mkdir, the current code is fine.

I suspect the problem has a different root cause in Pest which should be addressed instead.

calebdw:
Sure, I get that---likely having to do with how it's capturing warnings. But there's still three other calls to `mkdir` that are already guarded by an `is_dir` check so I feel either they should all be guarded or none of them should be guarded for consistency

ondrejmirtes:
If we wanted to make this consistent everywhere, I'd welcome a new class `DirectoryCreator` with a static method that would be used in place of these calls.

calebdw:
@ondrejmirtes done---please let me know if you'd like any other changes

src/Internal/FilesystemHelper.php
```diff
@@ -0,0 +1,27 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Internal;
+
+use function clearstatcache;
+use function is_dir;
+use function mkdir;
+
+final class FilesystemHelper
```
ondrejmirtes: Please name this class `DirectoryCreator` as I asked here: https://github.com/phpstan/phpstan-src/pull/2697#issuecomment-1784024859

ondrejmirtes (changes requested):

src/Internal/FilesystemHelper.php
calebdw: Done, I used `FilesystemHelper` because I figured other filesystem helpers could be added in the future

src/Cache/FileCacheStorage.php
```diff
@@ -28,20 +26,12 @@ public function __construct(private string $directory)
...
-			clearstatcache();
-			if (is_dir($directory)) {
-				return;
-			}
-
-			$error = error_get_last();
-			throw new InvalidArgumentException(sprintf('Failed to create directory "%s" (%s).', $this->directory, $error !== null ? $error['message'] : 'unknown cause'));
-		}
+		$error = error_get_last();
+		throw new InvalidArgumentException(sprintf('Failed to create directory "%s" (%s).', $this->directory, $error !== null ? $error['message'] : 'unknown cause'));
```
ondrejmirtes: Either we need to get rid of the specific error message here which relies on `error_get_last()` (leaks an implementation detail), or we need to throw a proper exception from DirectoryCreator, and surround each call with a try-catch.

I'd actually be in favour of the exception, because using return `bool` values to signal an error isn't great.

ondrejmirtes (changes requested):

src/Cache/FileCacheStorage.php
calebdw: Done

src/Internal/DirectoryCreator.php
```diff
@@ -0,0 +1,37 @@
...
+	/**
+	 * @throws InvalidArgumentException
+	 */
+	public static function ensureDirectoryExists(
+		string $directory,
+		int $mode = 0755,
+		bool $recursive = true,
+	): void
+	{
+		if (is_dir($directory)) {
```
staabm: since we have to use error suppressing in the below `mkdir()` calls, IMO we should not use `is_dir` here, which produces unnecessary IO

src/Internal/DirectoryCreator.php
```diff
@@ -0,0 +1,37 @@
...
+use InvalidArgumentException;
+use function clearstatcache;
+use function is_dir;
+use function mkdir;
+
+final class DirectoryCreator
+{
+
+	/**
+	 * @throws InvalidArgumentException
```
ondrejmirtes: Please create, throw and catch a specific exception, like `DirectoryCreatorException`.

src/Internal/DirectoryCreator.php
ondrejmirtes: Agreed, the first if can be removed

ondrejmirtes (changes requested):

src/Internal/DirectoryCreator.php
calebdw: That defeats the whole purpose of me submitting this PR in the first place... yes error suppression is used but an unnecessary warning is still generated which can cause issues for clients. There have been previous issues and PRs to add the `is_dir` check to avoid the warning being generated

ondrejmirtes: I've previously told you the root issue needs to be fixed in Pest. PHPStan's source code needs to make sense on its own. Otherwise it's not really good engineering and the code would be very brittle.

calebdw: I can remove, but I'm just pointing out that I'm not the first person this has caused issues with and who has tried to add the dir existence check

src/Internal/DirectoryCreator.php
calebdw: done

src/Internal/DirectoryCreator.php
calebdw: removed

src/Internal/DirectoryCreator.php
```diff
@@ -0,0 +1,32 @@
...
+	): void
+	{
+		if (@mkdir($directory, $mode, $recursive)) {
+			return;
+		}
+
+		clearstatcache();
+
+		if (!is_dir($directory)) {
+			throw new DirectoryCreatorException($directory);
```
ondrejmirtes, line 28: Would be great to actually use `error_get_last()` here and write a nice error message like the one you deleted in `makeDir`.

ondrejmirtes (changes requested):
After this it's good to merge 👍

src/Internal/DirectoryCreator.php
calebdw: done

ondrejmirtes:
Thank you.

ondrejmirtes:
Just did this to improve compatibility with PHPUnit and PHP-Parser 5: https://github.com/phpstan/phpstan-src/commit/9dac90dfd5fc809dd84fdac0532ecd51c6ae66bf

PR #2698, opened 2023-10-26
Micro-optimize with `in_array()`

zonuexe:
Same as https://github.com/phpstan/phpstan-src/pull/2618, but the effect is more partial.


ondrejmirtes:
Thank you!

PR #2700, opened 2023-10-27
fix !== with union of constants

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/10002

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1794,7 +1794,7 @@ class TypeSpecifier
 		$types = null;
 		if (
 			count($leftType->getFiniteTypes()) === 1
-			|| ($leftType->isConstantValue()->yes() && !$rightType->equals($leftType) && $rightType->isSuperTypeOf($leftType)->yes())
+			|| ($leftType->isConstantValue()->yes() && !$rightType->equals($leftType) && $rightType->isSuperTypeOf($leftType)->yes() && $context->true())
 		) {
 			$types = $this->create(
 				$rightExpr,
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/data/bug-10002.php
```diff
@@ -0,0 +1,26 @@
+<?php declare(strict_types=1);
+
+namespace Bug100002;
+
+use function PHPStan\Testing\assertType;
+
+class Foo
```
staabm, line 7: does this test cover bother changed execution paths?

schlndh: Good point - it doesn't. I'll add a test for the other direction as well.

schlndh:
Somehow this results in the following undesired change for https://phpstan.org/r/6f38a4ed-ff07-4959-9689-7d854de5f0ae :thinking: 

```
@@ @@
+PHP 7.4 – 8.3 (1 error)
+==========
+
+26: Expected type string, actual: bool|float|int|string
+
+PHP 7.2 – 7.3 (3 errors)
+==========
+
+24: Class PHPStan\Type\Constant\ConstantStringType not found.
+26: Call to method getValue() on an unknown class PHPStan\Type\Constant\ConstantStringType.
 26: Expected type string, actual: bool|float|int|string
 ```

**Comment by staabm** [comment]
I think its a unrelated side-effect. the playground sometime reports errors when you use PHPStan internal classes within the snippet

File: tests/PHPStan/Analyser/data/bug-10002.php [hunk 2]
```diff
@@ -0,0 +1,44 @@
+<?php declare(strict_types=1);
+
+namespace Bug100002;
```
ondrejmirtes: One extra zero :)

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
```diff
@@ -1792,7 +1792,7 @@ public function resolveIdentical(Expr\BinaryOp\Identical $expr, Scope $scope, Ty
 		$types = null;
 		if (
 			count($leftType->getFiniteTypes()) === 1
-			|| ($leftType->isConstantValue()->yes() && !$rightType->equals($leftType) && $rightType->isSuperTypeOf($leftType)->yes())
+			|| ($leftType->isConstantValue()->yes() && !$rightType->equals($leftType) && $rightType->isSuperTypeOf($leftType)->yes() && $context->true())
```
staabm: We could do the cheapest check (the newly added one) first

ondrejmirtes (approved):

ondrejmirtes:
Awesome, thank you!

PR #2702, opened 2023-10-28
Fix null coalesce assignment type

hirokinoue:
Fix: https://github.com/phpstan/phpstan/issues/9995

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1384,6 +1384,8 @@ class TypeSpecifier
 				$specifiedExprs[] = $expr->expr->var;
 				$expr = $expr->expr;
 			}
+		} elseif ($expr instanceof Expr\AssignOp\Coalesce) {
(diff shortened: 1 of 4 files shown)


hirokinoue:
Hi, I'm new to phptan-src. I created a pull request referred to several merged pull requests. Any advice is welcome. Thank you in advance.

ondrejmirtes:
Perfect, thank you very much!

ondrejmirtes:
I encourage you to look into more "easy fixes" issues (https://github.com/phpstan/phpstan/milestone/8) :) They might not be exactly "easy" to figure out, I mark issues as "easy fixes" where I'm sure it's actually fixable in the current feature set of PHPStan and doesn't require any fundamental changes or refactorings :)

The rule of thumb is - if the fix takes less than 50 lines of code, you're good to go, if it takes more and is tricker than that, I probably misqualified the issue and shouldn't be marked as easy to fix :)

hirokinoue:
Thank you so much for your review and advice :)

PR #2706, opened 2023-10-31
Fix Redis::keys function signature

Vaalyn:
This PR fixes the `Redis::keys` function signature to match the stub provided by the Redis extension: https://github.com/phpredis/phpredis/blob/a0c8fcc589bfcf8207ba3d8c7d065109f31cc1c6/redis.stub.php#L1948-L1949

Currently the functionMap says it can only return an `array<int,string>` but it can also return an instance of `Redis` when multimode is used or `false` when no server is reachable.


ondrejmirtes:
1) There are many functions like this in the Redis extension, please fix all of them.
2) Please wrap the return types in `__benevolent<...>` so that this change isn't too annoying on level 7+.

Vaalyn:
👍 will go through all the `Redis::` functions to fix the ones that are wrong and if applicable add missing ones.

Vaalyn:
I've gone through all `Redis::` entries in the `functionMap.php`, looking them up in the `redis.stub.php` from the Redis extension, fixing the return types, parameters, adding missing functions (as thorough as I could, maybe missed some but there are so many in there... 😅) and removing those that were removed from the extension.

ondrejmirtes:
Hello, thank you for this effort, the tests are failing.

> and removing those that were removed from the extension

Which methods did you remove and when they were removed from the extension?

Vaalyn:
> the tests are failing

Could you point me in the correct direction as to why they might be failing?

The error only says that one of the array keys for the parameters seems to be an integer but as far as I can see they are all strings:
```
TypeError: Argument 1 passed to PHPStan\Reflection\SignatureMap\SignatureMapParser::getParameterInfoFromName() must be of the type string, int given, called in /home/runner/work/phpstan-src/phpstan-src/src/Reflection/SignatureMap/SignatureMapParser.php on line 67
```

> Which methods did you remove and when they were removed from the extension?

```
'Redis::blPop\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::brPop\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::decrByFloat' => only available in RedisCluster, not in Redis
'Redis::del\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::delete\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::evaluate' => no idea, can't find anything about it on the repo
'Redis::exists\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::getKeys' => alias for "keys", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::getMultiple' => alias for "mget", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::lGet' => alias for "lindex", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::lGetRange' => alias for "lrange", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::listTrim' => alias for "ltrim", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::lSize' => alias for "lLen", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::renameKey' => alias for "rename", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::resetStat' => no idea, can't find anything about it on the repo
'Redis::sContains' => alias for "sismember", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::set\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::setTimeout' => alias for "expire", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::sGetMembers' => alias for "sMembers", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::slave' => it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::slave\'1' => no idea, see above
'Redis::sRemove' => alias for "srem", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::substr' => alias for "getRange", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::unlink\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::zAdd\'1' => as far as I could see there are is no alternative function signature in the extension
'Redis::zDelete' => alias for "zRem", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::zDeleteRangeByRank' => alias for "zRemRangeByRank", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::zDeleteRangeByScore' => alias for "zRemRangeByScore", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::zRemove' => alias for "zRem", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
'Redis::zSize' => alias for "zCard", readme says it will be removed in future versions but it's not in their stub file for at least 1 year no idea when exactly it got yeeted from there
```

ondrejmirtes:
> The error only says that one of the array keys for the parameters seems to be an integer but as far as I can see they are all strings:

If you named a parameter like `'1'`, it's going to be cast by PHP to an int.

You should be able to debug this by running the test and print what it's currently running around that line.

Vaalyn:
Found the cause of the error by var_dumping while running the tests locally.

It was due to one of the parameters missing the parameter name.

ondrejmirtes:
Please return the removed functions back to the file, someone might be still using them.

Vaalyn:
@ondrejmirtes added them back into the map 👍 

Is there any guide on when to delete removed functions?

ondrejmirtes:
> when to delete removed functions

There isn't. Basically there is never a good reason to remove functions that might still be in use.

resources/functionMap.php
```diff
@@ -9467,222 +9467,269 @@
...
-'Redis::_unserialize' => ['mixed', 'value'=>'string'],
-'Redis::append' => ['int', 'key'=>'string', 'value'=>'string'],
-'Redis::auth' => ['bool', 'password'=>'string|string[]'],
-'Redis::bgRewriteAOF' => ['bool'],
-'Redis::bgSave' => ['bool'],
-'Redis::bitCount' => ['int', 'key'=>'string'],
-'Redis::bitOp' => ['int', 'operation'=>'string', '...args'=>'string'],
-'Redis::bitpos' => ['int', 'key'=>'string', 'bit'=>'int', 'start='=>'int', 'end='=>'int'],
-'Redis::blPop' => ['array', 'keys'=>'string[]', 'timeout'=>'int'],
+'Redis::__construct' => ['void', 'options'=>'?array{host?:string,port?:int,connectTimeout?:float,auth?:list{string|null|false,string}|list{string},ssl?:array<string, mixed>,backoff?:array<string, mixed>}'],
```
ondrejmirtes: The `options` parameter should be optional

ondrejmirtes (changes requested):

resources/functionMap.php
Vaalyn: 👍 done

ondrejmirtes:
Thank you! You might also be interested in cleaning this up in https://github.com/JetBrains/phpstorm-stubs/tree/master/redis :)

RobiNN1:
Probably after this PR there is issue `Method Redis::restore() invoked with 3 parameters, 4 required.`

4th parameter is optional `$options = null`

https://phpstan.org/r/00184a69-6093-4fef-a5fe-abf0be19b79f

ondrejmirtes:
@RobiNN1 Please open a bug report.

PR #2707, opened 2023-10-31
Fix `class-string<*>` being supertype of any string literal

jrmajor:
Fixes phpstan/phpstan#10076.

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -1252,6 +1252,12 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug10076(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-10076.php');
+		$this->assertNoErrors($errors);
+	}
+
 	/**
 	 * @param string[]|null $allAnalysedFiles
 	 * @return Error[]
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -1252,12 +1252,6 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertNoErrors($errors);
 	}
 
-	public function testBug10076(): void
-	{
-		$errors = $this->runAnalyse(__DIR__ . '/data/bug-10076.php');
-		$this->assertNoErrors($errors);
-	}
-
 	/**
 	 * @param string[]|null $allAnalysedFiles
 	 * @return Error[]
--- a/tests/PHPStan/Analyser/data/bug-10076.php
+++ b/tests/PHPStan/Analyser/data/bug-10076.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Bug10076;
+
+/**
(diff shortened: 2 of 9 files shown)


ondrejmirtes:
Hi, changes to isSuperTypeOf need to be verified in TypeCombinatorTest::testUnion and testIntersect. So please add some cases to the data providers there.

You need to test expectations with different types that should hit different branches in that method.

Thank you.

jrmajor:
Hi, thank you for such a quick reply :)

> You need to test expectations with different types that should hit different branches in that method.

Are the tests that I've added sufficient, or do I need to add some for the branches that I didn't touch too?

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1252,6 +1252,12 @@ public function testBug10049(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug10076(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-10076.php');
```
ondrejmirtes: This should be tested in the corresponding rule - CallToFunctionParametersRule

ondrejmirtes: AnalyserIntegrationTest is reserved for crashes, infinite loops etc.

tests/PHPStan/Analyser/data/bug-10076.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+function result($type = 'object'): array
+{
+	return [];
+}
+
+result();
+result('object');
+result('array');
+result(\DateTime::class);
```
ondrejmirtes: Please do not mix function declarations and side-effects such as function calls. Please put these calls in an anonymous function.

tests/PHPStan/Type/TypeCombinatorTest.php
```diff
@@ -4134,6 +4162,19 @@ public function dataIntersect(): iterable
...
+				new ConstantStringType(Exception::class),
+				new GenericClassStringType(TemplateTypeFactory::create(
+					TemplateTypeScope::createWithFunction('foo'),
+					'T',
+					null,
+					TemplateTypeVariance::createInvariant(),
+				)),
+			],
+			ConstantStringType::class,
+			"'Exception'",
```
ondrejmirtes, line 4189: Please test intersection with `'array'` too - should be NeverType.

ondrejmirtes (changes requested):
After these changes it will be good to merge 👍

jrmajor:
This broke the test for phpstan/phpstan#6697. Not sure what to do here. Fixing phpstan/phpstan#10076 requires assuming that constant string is not a class-string unless PHPStan knows about that class, while phpstan/phpstan#6697 relies on the assumption that constant string may be a name of a class that PHPStan doesn't know about.

tests/PHPStan/Type/TypeCombinatorTest.php
jrmajor: Fixing the test for phpstan/phpstan#6697 would make it `'array'&class-string<T (function foo(), parameter)>` instead of `never`, because we would need to assume that `'array'` may be name of a class PHPStan doesn't know about. What should I do about that?

jrmajor:
@ondrejmirtes Friendly ping, I need your decision here.

PR #2708, opened 2023-11-01
More precise falsey isset()

staabm:
split off parts out from https://github.com/phpstan/phpstan-src/pull/2657 since it gets too complicated.

lets do it one step at a time instead

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -67,11 +67,9 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\UnionType;
-use function array_filter;
 use function array_key_exists;
 use function array_map;
 use function array_merge;
-use function array_reduce;
 use function array_reverse;
 use function count;
 use function in_array;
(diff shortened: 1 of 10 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -681,14 +679,35 @@ public function specifyTypesInCondition(
...
+				foreach ($expr->vars as $var) {
+					$isset = $scope->issetCheck($var, static fn () => true);
+
+					if ($isset !== true) {
+						continue;
+					}
+
+					$type = $scope->getType($var);
+					if (
+						!$var instanceof Expr\Variable ||
```
ondrejmirtes: I'm not sure what's special about `Variable` that it needs to be mentioned in this condition, it's usually a red flag.

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
staabm: Its the first step to get this better as before. 

Isset on variable is a very common case. I will make this more general in followups

ondrejmirtes: Please make this correct now, it's the right time to do it. The logic is correct when there is no special handling of a specific expr types.

src/Analyser/TypeSpecifier.php
```diff
@@ -681,14 +679,34 @@ public function specifyTypesInCondition(
...
+				foreach ($expr->vars as $var) {
+					$isset = $scope->issetCheck($var, static fn () => true);
+
+					if ($isset !== true) {
+						continue;
+					}
+
+					$type = $scope->getType($var);
+					if (
+						!$type instanceof MixedType && !TypeCombinator::containsNull($type)
```
ondrejmirtes: `!$type instanceof MixedType && ` doesn't need to be there as `TypeCombinator::containsNull` will never return `true` for `MixedType`

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
staabm: without this explicit `MixedType` check the 

```
function mixedIsset(mixed $m): void
{
	if (isset($m)) {
		assertType("mixed~null", $m);
	} else {
		assertType("null", $m);
	}
}
```

case will not succeed. I turned the conditions arround, maybe its more readable now.

ondrejmirtes:
I knew deep down the condition really wasn't needed. `instanceof` of a specific type is also usually a red flag: https://github.com/phpstan/phpstan-src/pull/2708/commits/001ad45ab5f11aba6baffba512aad56f9ccc740a

ondrejmirtes:
Thank you.

staabm:
awesome, thanks!

PR #2709, opened 2023-11-02
Rewrite multi param isset() to and-chain

staabm:
motivation: ease impl of https://github.com/phpstan/phpstan-src/pull/2657 by moving the expr-var loop before the context true/false branches which simplifies the branch cases

closes https://github.com/phpstan/phpstan/issues/8366
closes https://github.com/phpstan/phpstan/issues/10064

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -675,8 +675,8 @@ class TypeSpecifier
 			&& count($expr->vars) > 0
 			&& !$context->null()
 		) {
+			// rewrite multi param isset() to and-chained single param isset()
 			if (count($expr->vars) > 1) {
-				// rewrite multi param isset() to and-chained single param isset()
 				$issets = [];
 				foreach ($expr->vars as $var) {
 					$issets[] = new Expr\Isset_([$var], $expr->getAttributes());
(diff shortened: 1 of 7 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -674,59 +675,75 @@ public function specifyTypesInCondition(
...
+					new NullType(),
+					$context->negate(),
+					false,
+					$scope,
+					$rootExpr,
+				);
 			}
 
-			$types = null;
+			$tmpVars = [$issetExpr];
```
staabm, line 726: I had to rename this var, because the previous `$var` variable after this PR is no longer within a foreach and therefore PHPStan would otherwise report a overlap with the foreach-var below

(no logic change contained otherwise in this block)

src/Analyser/TypeSpecifier.php
```diff
@@ -674,59 +675,75 @@ public function specifyTypesInCondition(
...
 			&& !$context->null()
 		) {
-			if (!$context->true()) {
-				if (!$scope instanceof MutatingScope) {
-					throw new ShouldNotHappenException();
-				}
-
-				$specifiedTypes = new SpecifiedTypes();
+			// rewrite multi param isset() to and-chained single param isset()
+			if (count($expr->vars) > 1) {
```
staabm, line 679: the actual change is here.

Implemented the idea expressed in https://github.com/phpstan/phpstan/issues/8366#issuecomment-1314755824

staabm:
ahh, let me add a regression test for https://github.com/phpstan/phpstan/issues/10064

tests/PHPStan/Analyser/data/bug-8366.php
```diff
@@ -0,0 +1,26 @@
...
+use function PHPStan\Testing\assertType;
+
+function validateParams(
+	?\DateTimeImmutable $datePeriod,
+	?\DateTimeImmutable $untilDate,
+	?int $count
+): void {
+
+	if (isset($untilDate, $count)) {
+		throw new \InvalidArgumentException('Too much params, choose between until and count.');
```
ondrejmirtes, line 17: Additional assertType calls inside this condition would be great

ondrejmirtes (changes requested):
Excellent idea, thank you

ondrejmirtes:
Thank you.

PR #2710, opened 2023-11-02
Fix !isset() with Variable

staabm:
Another split out of https://github.com/phpstan/phpstan-src/pull/2657

implements the 

> Because for $foo with certainty maybe and not-nullable, the only !isset($foo) is that the variable does not exist, so the certainty needs to be no.

part of https://github.com/phpstan/phpstan-src/pull/2657#pullrequestreview-1651857463

---

closes https://github.com/phpstan/phpstan/issues/3632
closes https://github.com/phpstan/phpstan/issues/8190
closes https://github.com/phpstan/phpstan/issues/8366
closes https://github.com/phpstan/phpstan/issues/8659
closes https://github.com/phpstan/phpstan/issues/9580
closes https://github.com/phpstan/phpstan/issues/10064
closes https://github.com/phpstan/phpstan/issues/10088

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3850,22 +3850,18 @@ class MutatingScope implements Scope
 			if ($expr instanceof IssetExpr) {
 				$issetExpr = $expr;
 				$expr = $issetExpr->getExpr();
-				$certainty = $issetExpr->getCertainty();
 
-				if ($certainty->no()) {
-					$scope = $scope->unsetExpression($expr);
-
-					if ($expr instanceof Variable) {
-						$exprString = $this->getNodeKey($expr);
-
-						unset($scope->expressionTypes[$exprString]);
-						unset($scope->nativeExpressionTypes[$exprString]);
-					}
-				} else {
+				if ($typeSpecification['sure']) {
 					$scope = $scope->setExpressionCertainty(
 						$expr,
-						$certainty,
+						TrinaryLogic::createMaybe(),
 					);
+				} else {
+					$scope = $scope->unsetExpression($expr);
+					$exprString = $this->getNodeKey($expr);
+
+					unset($scope->expressionTypes[$exprString]);
+					unset($scope->nativeExpressionTypes[$exprString]);
 				}
 
 				continue;
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3891,10 +3891,11 @@ class MutatingScope implements Scope
 
 		foreach ($conditions as $conditionalExprString => $expressions) {
 			$certainty = TrinaryLogic::lazyExtremeIdentity($expressions, static fn (ConditionalExpressionHolder $holder) => $holder->getTypeHolder()->getCertainty());
-			$type = TypeCombinator::union(...array_map(static fn (ConditionalExpressionHolder $holder) => $holder->getTypeHolder()->getType(), $expressions));
 			if ($certainty->no()) {
 				unset($scope->expressionTypes[$conditionalExprString]);
 			} else {
+				$type = TypeCombinator::union(...array_map(static fn (ConditionalExpressionHolder $holder) => $holder->getTypeHolder()->getType(), $expressions));
+
 				$scope->expressionTypes[$conditionalExprString] = array_key_exists($conditionalExprString, $scope->expressionTypes)
 					? new ExpressionTypeHolder(
 						$scope->expressionTypes[$conditionalExprString]->getExpr(),
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3840,6 +3840,7 @@ class MutatingScope implements Scope
 			$specifiedExpressions[$this->getNodeKey($expr)] = ExpressionTypeHolder::createYes($expr, $scope->getType($expr));
 		}
 
+		$conditions = [];
 		foreach ($scope->conditionalExpressions as $conditionalExprString => $conditionalExpressions) {
 			foreach ($conditionalExpressions as $conditionalExpression) {
 				foreach ($conditionalExpression->getConditionExpressionTypeHolders() as $holderExprString => $conditionalTypeHolder) {
@@ -3848,18 +3849,24 @@ class MutatingScope implements Scope
 					}
 				}
 
-				if ($conditionalExpression->getTypeHolder()->getCertainty()->no()) {
-					unset($scope->expressionTypes[$conditionalExprString]);
-				} else {
-					$scope->expressionTypes[$conditionalExprString] = array_key_exists($conditionalExprString, $scope->expressionTypes)
-						? new ExpressionTypeHolder(
-							$scope->expressionTypes[$conditionalExprString]->getExpr(),
-							TypeCombinator::intersect($scope->expressionTypes[$conditionalExprString]->getType(), $conditionalExpression->getTypeHolder()->getType()),
-							TrinaryLogic::maxMin($scope->expressionTypes[$conditionalExprString]->getCertainty(), $conditionalExpression->getTypeHolder()->getCertainty()),
-						)
-						: $conditionalExpression->getTypeHolder();
-					$specifiedExpressions[$conditionalExprString] = $conditionalExpression->getTypeHolder();
-				}
+				$conditions[$conditionalExprString][] = $conditionalExpression;
+				$specifiedExpressions[$conditionalExprString] = $conditionalExpression->getTypeHolder();
+			}
+		}
+
+		foreach ($conditions as $conditionalExprString => $expressions) {
+			$certainty = TrinaryLogic::lazyExtremeIdentity($expressions, static fn (ConditionalExpressionHolder $holder) => $holder->getTypeHolder()->getCertainty());
+			$type = TypeCombinator::union(...array_map(static fn (ConditionalExpressionHolder $holder) => $holder->getTypeHolder()->getType(), $expressions));
+			if ($certainty->no()) {
+				unset($scope->expressionTypes[$conditionalExprString]);
+			} else {
+				$scope->expressionTypes[$conditionalExprString] = array_key_exists($conditionalExprString, $scope->expressionTypes)
+					? new ExpressionTypeHolder(
+						$scope->expressionTypes[$conditionalExprString]->getExpr(),
+						TypeCombinator::intersect($scope->expressionTypes[$conditionalExprString]->getType(), $type),
+						TrinaryLogic::maxMin($scope->expressionTypes[$conditionalExprString]->getCertainty(), $certainty),
+					)
+					: $expressions[0]->getTypeHolder();
 			}
 		}
 
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2348,6 +2348,10 @@ class MutatingScope implements Scope
 	/** @api */
 	public function hasExpressionType(Expr $node): TrinaryLogic
 	{
+		if ($node instanceof Variable && is_string($node->name)) {
+			return $this->hasVariableType($node->name);
+		}
+
 		$exprString = $this->getNodeKey($node);
 		if (!isset($this->expressionTypes[$exprString])) {
 			return TrinaryLogic::createNo();
@@ -3440,7 +3444,7 @@ class MutatingScope implements Scope
 		return $scope->invalidateExpression($expr);
 	}
 
-	public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType): self
+	public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType, ?TrinaryLogic $certainty = null): self
 	{
 		if ($expr instanceof ConstFetch) {
 			$loweredConstName = strtolower($expr->name->toString());
@@ -3474,6 +3478,7 @@ class MutatingScope implements Scope
 					if ($dimType instanceof ConstantIntegerType) {
 						$types[] = new StringType();
 					}
+
 					$scope = $scope->specifyExpressionType(
 						$expr->var,
 						TypeCombinator::intersect(
@@ -3481,16 +3486,23 @@ class MutatingScope implements Scope
 							new HasOffsetValueType($dimType, $type),
 						),
 						$scope->getNativeType($expr->var),
+						$certainty,
 					);
 				}
 			}
 		}
 
+		if ($certainty === null) {
+			$certainty = TrinaryLogic::createYes();
+		} elseif ($certainty->no()) {
+			throw new ShouldNotHappenException();
+		}
+
 		$exprString = $this->getNodeKey($expr);
 		$expressionTypes = $scope->expressionTypes;
-		$expressionTypes[$exprString] = ExpressionTypeHolder::createYes($expr, $type);
+		$expressionTypes[$exprString] = new ExpressionTypeHolder($expr, $type, $certainty);
 		$nativeTypes = $scope->nativeExpressionTypes;
-		$nativeTypes[$exprString] = ExpressionTypeHolder::createYes($expr, $nativeType);
+		$nativeTypes[$exprString] = new ExpressionTypeHolder($expr, $nativeType, $certainty);
 
 		return $this->scopeFactory->create(
 			$this->context,
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3720,6 +3720,23 @@ class MutatingScope implements Scope
 		);
 	}
 
+	private function setExpressionCertainty(Expr $expr, TrinaryLogic $certainty): self
+	{
+		if ($this->hasExpressionType($expr)->no()) {
+			throw new ShouldNotHappenException();
+		}
+
+		$originalExprType = $this->getType($expr);
+		$nativeType = $this->getNativeType($expr);
+
+		return $this->specifyExpressionType(
+			$expr,
+			$originalExprType,
+			$nativeType,
+			$certainty,
+		);
+	}
+
 	private function addTypeToExpression(Expr $expr, Type $type): self
 	{
 		$originalExprType = $this->getType($expr);
@@ -3844,6 +3861,11 @@ class MutatingScope implements Scope
 						unset($scope->expressionTypes[$exprString]);
 						unset($scope->nativeExpressionTypes[$exprString]);
 					}
+				} else {
+					$scope = $scope->setExpressionCertainty(
+						$expr,
+						$certainty,
+					);
 				}
 
 				continue;
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -37,6 +37,7 @@ use PHPStan\Node\Expr\OriginalPropertyTypeExpr;
 use PHPStan\Node\Expr\PropertyInitializationExpr;
 use PHPStan\Node\Expr\SetOffsetValueTypeExpr;
 use PHPStan\Node\Expr\TypeExpr;
+use PHPStan\Node\IssetExpr;
 use PHPStan\Node\Printer\ExprPrinter;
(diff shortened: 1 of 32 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -709,18 +710,48 @@ public function specifyTypesInCondition(
...
 				);
+				if ($isset === true) {
+					return $exprType;
+				}
+
+				if ($issetExpr instanceof Expr\Variable && is_string($issetExpr->name)) {
+					$type = $scope->getType($issetExpr);
+
+					$nullType = new NullType();
+					if (!$nullType->isSuperTypeOf($type)->no()) {
```
staabm: I am not sure about this condition

src/Analyser/TypeSpecifier.php
```diff
@@ -709,18 +710,48 @@ public function specifyTypesInCondition(
...
 					$context->negate(),
 					false,
 					$scope,
 					$rootExpr,
 				);
+				if ($isset === true) {
+					return $exprType;
+				}
+
+				if ($issetExpr instanceof Expr\Variable && is_string($issetExpr->name)) {
```
staabm, line 609: hopefully its OK to do the variable stuff in a separate PR.
whats missing is `ArrayDimFetch`, `PropertyFetch`, `StaticPropertyFetch` which could be done in a followup?

staabm:
the carbon error seems valid

the larastan errors seem unrelated, as they also show up in other open PRs

the presta shop error looks fishy

src/Analyser/MutatingScope.php
```diff
@@ -3820,6 +3838,31 @@ public function filterBySpecifiedTypes(SpecifiedTypes $specifiedTypes): self
 		foreach ($typeSpecifications as $typeSpecification) {
 			$expr = $typeSpecification['expr'];
 			$type = $typeSpecification['type'];
+
+			if ($expr instanceof IssetExpr) {
+				$issetExpr = $expr;
+				$expr = $issetExpr->getExpr();
+				$certainty = $issetExpr->getCertainty();
```
ondrejmirtes: This shouldn't be done through a property on IssetExpr, but rather this should work through `$typeSpecification['sure']` being true or false same as for all the other things.

src/Analyser/MutatingScope.php
```diff
@@ -3820,6 +3838,31 @@ public function filterBySpecifiedTypes(SpecifiedTypes $specifiedTypes): self
...
+
+				if ($certainty->no()) {
+					$scope = $scope->unsetExpression($expr);
+
+					if ($expr instanceof Variable) {
```
ondrejmirtes: Also I'm bothered by `instanceof Variable` again. I'd say we don't need this at all because `unsetExpression` should take care of everything. When NodeScopeResolver handles Unset_, it works even for variables without any ifs. See https://phpstan.org/r/889ea0c0-1b67-4894-b3f4-2f76ed4ed48e

ondrejmirtes (changes requested):
Also this is the place to test all the different scenarios like I described in the original PR: https://github.com/phpstan/phpstan-src/pull/2657#pullrequestreview-1651857463

So if we're doing `!isset($a)`:

1) If the variable is nullable, it means that after this PR it might be null or it might not exist
2) If the variable is not nullable, it means that after this PR it does not exist anymore

etc...

src/Analyser/MutatingScope.php
```diff
@@ -3820,6 +3838,27 @@ public function filterBySpecifiedTypes(SpecifiedTypes $specifiedTypes): self
...
+				if ($typeSpecification['sure']) {
+					$scope = $scope->setExpressionCertainty(
+						$expr,
+						TrinaryLogic::createMaybe(),
+					);
+				} else {
+					$scope = $scope->unsetExpression($expr);
+					$exprString = $this->getNodeKey($expr);
+
+					unset($scope->expressionTypes[$exprString]);
```
ondrejmirtes: This shouldn't be needed as it's already handled in unsetExpression

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: that was a great hint. I would have never thought this should be done via `$typeSpecification['sure']`

staabm:
I have more tests and a small fix open todo

staabm:
one interessting failling drupal case:

https://github.com/drupal/drupal/blob/2700c5afb6c3936041db413872eea82dc0bd4fe4/core/modules/user/tests/src/Unit/UserAccessControlHandlerTest.php#L138C84-L154

```
------ ------------------------------------------------------------------- 
  Line   core/modules/user/tests/src/Unit/UserAccessControlHandlerTest.php  
 ------ ------------------------------------------------------------------- 
  152    Variable $result might not be defined.                             
 ------ ------------------------------------------------------------------- 
 ```

reduced it to 

```php
function doBar(bool $view, bool $edit): void {
	foreach (['view' => $view, 'edit' => $edit] as $operation => $result) {
		assertVariableCertainty(TrinaryLogic::createYes(), $result);

		$result_text = !isset($result) ? 'null' : ($result ? 'true' : 'false'); // Variable $result in isset() always exists and is not nullable.

		assertVariableCertainty(TrinaryLogic::createYes(), $result); // Expected variable certainty Yes, actual: Maybe
		assertType('bool', $result); // Variable $result might not be defined.
	}
}
```

-> I think its great that PHPStan realized that since `$result` can never be null and therefore the `isset()` is actually dead code
-> since we turn the certainty to NO via !isset() we get a misleading consequential false-positve error

ondrejmirtes:
Hi, I'm putting this on hold for a while until this problem is resolved: https://github.com/phpstan/phpstan-src/pull/2666#issuecomment-1792687552

ondrejmirtes:
Alright, one more problem this PR should try to fix is:

```
Expected No, actual certainty of variable $shortcut_id is Maybe in /Users/ondrej/Development/phpstan/tests/PHPStan/Analyser/data/bug-10088.php on line 22.
```

Not sure why it does not work now.

ondrejmirtes:
The problem in Drupal should be fixed, the code in Carbon should still fail because it was wrong from the start.

ondrejmirtes:
Did my commit in forward fashion on 1.10.x, only added more tests here.

ondrejmirtes:
Results look better but there's something wrong with `!empty` in Drupal: https://github.com/phpstan/phpstan-src/actions/runs/6756144616/job/18365350671

Source: https://github.com/drupal/drupal/blob/2700c5afb6c3936041db413872eea82dc0bd4fe4/core/modules/book/book.module#L123

staabm:
I guess the problem is related that the variable is only defined via dim fetches

(Maybe I can have a look tonight)

ondrejmirtes:
That shouldn't matter, right? Also I found out it already fails on 1.10.x so maybe I broke it with my recent commits.

staabm:
It shouldn't - but its a notable difference of the case at hand vs working unit tests

staabm:
Another notable difference of the drupal case: the dim-fetch based declarations are wrapped in a property-fetch based isset()

staabm:
added a reduced test for the drupal case. looks like a certainty issue of regular variables (which was already contained in 1.10.40)

ondrejmirtes:
Hi, my plan for today is to make 1.10.x releasable again by reverting all related changes and picking those commits into this PR. Right now it’s pretty broken and we need to do this properly in order not to endanger the stability. So we’re gonna focus the changes here and make sure the CI is green.

I also need to work on other stuff like PHP 8.3 and PHPStan 1.11.

staabm:
I see thanks. If there is anything I can do to support you in other areas, feel free to note.

staabm:
> Results look better but there's something wrong with `!empty` in Drupal: [phpstan/phpstan-src/actions/runs/6756144616/job/18365350671](https://github.com/phpstan/phpstan-src/actions/runs/6756144616/job/18365350671)
> 
> Source: [drupal/drupal@`2700c5a`/core/modules/book/book.module#L123](https://github.com/drupal/drupal/blob/2700c5afb6c3936041db413872eea82dc0bd4fe4/core/modules/book/book.module#L123)

I also found the cause for this problem:

```php
function book_node_links_alter(array &$links, NodeInterface $node) {
  \PHPStan\dumpType(isset($node->book['depth'])); // Dumped type: false 
    if (isset($node->book['depth'])) {
          $book_links['book_printer'] = [
          ];
    }
    \PHPStan\Testing\assertVariableCertainty(\PHPStan\TrinaryLogic::createMaybe(), $book_links); // Expected variable certainty Maybe, actual: No

    if (!empty($book_links)) {
      $links['book'] = [
        '#links' => $book_links,
      ];
    }
}
```

I think everything is working as expected - as far as this PR is a concern. we conclude the isset() expression is always FALSE, therefore we don't take the IF-truethy-scope into the IF-final-scope which means the variable has a certainty of NO after the `if (isset($node->book['depth'])) {` block

staabm:
> Alright, one more problem this PR should try to fix is:
> 
> ```
> Expected No, actual certainty of variable $shortcut_id is Maybe in /Users/ondrej/Development/phpstan/tests/PHPStan/Analyser/data/bug-10088.php on line 22.
> ```
> 
> Not sure why it does not work now.

I think the cause for this problem is, that the scope before entering the `if ($link_mode === "add") {` statement is

```php
$scope->debug();

array (
  '$shortcut_id (Maybe)' => '1',
  '$link_mode (Yes)' => '\'add\'|\'remove\'',
  'native $shortcut_id' => '1',
  'native $link_mode' => '\'add\'|\'remove\'',
  'condition about $shortcut_id #1' => 'if $link_mode=\'remove\' then $shortcut_id is 1 (Yes)',
)
```

which means we don't know about a conditional type and certainty for the ELSE case of the ternary `$link_mode = isset($shortcut_id) ? "remove" : "add";`

in the end it feels like a possible future improvement..?

ondrejmirtes:
I hope you don't mind the forcepushes. If you have some commits locally and I forcepush the base under your feet, can solve it easily with `git rebase --onto`.

staabm:
summarizing/interpreting the current errors

### nikic/PHP-Parser

[origin code](https://github.com/nikic/PHP-Parser/blob/da65ae474d2db681b0c37f6d3ff97dba8e14cc5f/lib/PhpParser/PrettyPrinterAbstract.php#L853-L854)

```
------ ------------------------------------- 
  Line   PhpParser/PrettyPrinterAbstract.php  
 ------ ------------------------------------- 
  854    If condition is always true.         
 ------ ------------------------------------- 
 ```

reduced: 

```php
<?php

use PhpParser\Node;

/**
 * @param positive-int $itemStartPos
 */
function doFoo(Node $arrItem , Node $origArrItem , $itemStartPos, int $pos): void {
	$itemStartPos = $origArrItem->getStartTokenPos();
	$itemEndPos = $origArrItem->getEndTokenPos();
	\assert($itemStartPos >= 0 && $itemEndPos >= 0 && $itemStartPos >= $pos);

	$comments = $arrItem->getComments();
	$origComments = $origArrItem->getComments();
	assertType('array<PhpParser\Comment>', $origComments);
	$commentStartPos = $origComments ? $origComments[0]->getStartTokenPos() : $itemStartPos;
	assertType('array<PhpParser\Comment>', $origComments);
	\assert($commentStartPos >= 0);

	assertType('array<PhpParser\Comment>', $comments);
	assertType('array<PhpParser\Comment>', $origComments); // Expected type array<PhpParser\Comment>, actual: array{}  
}
```

-> false positive -> fixed in [ce6f25b](https://github.com/phpstan/phpstan-src/pull/2710/commits/ce6f25bb1ff7d237e6f5145b931641b079999655)

### briannesbitt/Carbon

[origin code](https://github.com/briannesbitt/Carbon/blob/0af369fba2435afe17a97cfa9a02a8e84d6c6640/tests/CarbonPeriod/ToArrayTest.php#L99-L100)

```
------ -------------------------------------- 
  Line   tests/CarbonPeriod/ToArrayTest.php    
 ------ -------------------------------------- 
  100    Variable $date might not be defined.  
 ------ -------------------------------------- 
 ```

looks like a valid error IMO since carbon did not annotate `assertInstanceOfCarbon` in a way which narrows the arg.
this changed just recently https://github.com/briannesbitt/Carbon/commit/f0dcc1c4a895f710c758229a99388d87ef1def26

### drupal/drupal

```
------ --------------------------------- 
  Line   core/modules/book/book.module    
 ------ --------------------------------- 
  123    Undefined variable: $book_links  
 ------ --------------------------------- 

 ------ ------------------------------------------------------------------- 
  Line   core/modules/user/tests/src/Unit/UserAccessControlHandlerTest.php  
 ------ ------------------------------------------------------------------- 
  152    Variable $result might not be defined.                             
 ------ ------------------------------------------------------------------- 
```

[origin source book.module ](https://github.com/drupal/drupal/blob/2700c5afb6c3936041db413872eea82dc0bd4fe4/core/modules/book/book.module#L120-L126)

-> false positive

[origin source UserAccessControlHandlerTest](https://github.com/drupal/drupal/blob/2700c5afb6c3936041db413872eea82dc0bd4fe4/core/modules/user/tests/src/Unit/UserAccessControlHandlerTest.php#L149-L153)

-> false positive -> fixed in https://github.com/phpstan/phpstan-src/pull/2710/commits/425c581746775d65148c541c2d6ccc9b4fc09001

reduced

```php
<?php

class DrupalTest
{
	public function assertFieldAccess(bool $view, bool $edit): void {
		foreach (['view' => $view, 'edit' => $edit] as $operation => $result) {
			assertVariableCertainty(\PHPStan\TrinaryLogic::createYes(), $result);
			$result_text = !isset($result) ? 'null' : ($result ? 'true' : 'false');
			assertVariableCertainty(\PHPStan\TrinaryLogic::createYes(), $result); // Expected variable certainty Yes, actual: Maybe
			echo $result;
		}
	}
}

```

### PrestaShop/PrestaShop

```
------ ---------------------------------------------------------- 
  Line   controllers/admin/AdminModulesController.php              
 ------ ---------------------------------------------------------- 
  832    Undefined variable: $modules_list_save                    
  832    Variable $modules_list_save in isset() is never defined.  
 ------ ---------------------------------------------------------- 
 ```

[origin source](https://github.com/PrestaShop/PrestaShop/blob/ca9d81aae0bb17f3e767bb5d835348ed144a3ab5/controllers/admin/AdminModulesController.php#L832)

-> works as expected -> `modules_list_save` is either isset and non-null, or non-existant. therefore the last branch cannot be taken

**Comment by staabm** [comment]
the certainty of `$expr` in `isset($expr)` after ternary currently fails, because scopes are merged and NO merged with YES turns into MAYBE at

https://github.com/phpstan/phpstan-src/blob/28dab12d14f60f0b30b307433b9fd60a99587ba2/src/Analyser/NodeScopeResolver.php#L2795

[see currently failling test-case](https://github.com/phpstan/phpstan-src/pull/2710/files#diff-341ec38542af7441ace821df0645cfa728e2b47b51d7b29beda6cf48fb3cbe21R140-R152)

I think we somehow need to make sure the NO certainty given thru `IssetExpr` does not influence the final scope.


~this need to cover though that the ternary might contain a regular `unset()` which still should have an impact.~ 
update: [thats a syntax error](https://3v4l.org/JPnUY), so we are fine for `unset` in ternary.

**Comment by staabm** [comment]
I worked on fixing remaining bugs. I think I fixed the remaining ones and added lots of more tests.

errors in CI seem valid.

would be great to have another round of reviews, when time allows.

**Comment by ondrejmirtes** [comment]
Just tested this on a real-world project.

PHAR from this branch takes 17m24s to complete, stable PHAR takes 15m18s to complete (in a single thread).

But this might just be caused by a single file, will need to investigate further.

**Comment by staabm** [comment]
I tried this PR on several real world projects and could not see a slowdown. if you can provide a reproducer I can look into it

**Comment by ondrejmirtes** [comment]
There is no performance difference anymore 🎉 

But there are some false positives:

https://phpstan.org/r/5808b352-1128-4a68-ae93-28990774c56f - reports `Variable $matches might not be defined.`

https://phpstan.org/r/8c8d7bcc-1b43-40d0-ab26-8776f17a51bc - broken type assert

There might be more but they're probably connected to these. I'm gonna take a look again after you fix these :) Thanks.

**Comment by staabm** [comment]
awesome - thank you. I will have a look in the upcoming days.

**Comment by staabm** [comment]
@ondrejmirtes just pushed fixes for the snippets you mentioned

**Comment by ondrejmirtes** [comment]
Hi, I have one more bug for you: https://phpstan.org/r/e4755dda-74e3-44f3-bb09-b4311cb7835c

On your PR this snippet reports:

```
 ------ ----------------------------------------------------------------------------------------------
  Line   test.php
 ------ ----------------------------------------------------------------------------------------------
  25     Call to function array_key_exists() with 'nearest_premise' and array will always evaluate to
         true.
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting
            treatPhpDocTypesAsCertain: false in your phpstan.neon.dist.
 ------ ----------------------------------------------------------------------------------------------
```

**Comment by staabm** [comment]
thanks for bringing these examples up.

I think the bug is in building the `$finalScope` after ternary.

atm we merge the following 2 scopes:

```
$ifTrueScope->debug();
array (
  '$innerHits (Yes)' => 'array&hasOffset(\'nearest_premise\')',
  '$this (Yes)' => '$this(HelloWorld)',
  'native $innerHits' => 'array&hasOffset(\'nearest_premise\')',
  'native $this' => '$this(HelloWorld)',
  'condition about $innerHits #1' => 'if $innerHits=array then $innerHits is array&hasOffset(\'nearest_premise\') (Yes)',
)

$ifFalseScope->debug();
array (
  '$innerHits (Yes)' => '*NEVER*',
  '$this (Yes)' => '$this(HelloWorld)',
  'native $innerHits' => 'array',
  'native $this' => '$this(HelloWorld)',
  'condition about $innerHits #1' => 'if $innerHits=array then $innerHits is array&hasOffset(\'nearest_premise\') (Yes)',
)
```

I wonder whether merging `'$innerHits (Yes)' => 'array&hasOffset(\'nearest_premise\')',` and `'$innerHits (Yes)' => '*NEVER*',` should lead to unsetting the variable-type from scope? 

atm these are merged together and the resulting scope looks like

```
$finalScope->debug();
array (
  '$innerHits (Yes)' => 'array&hasOffset(\'nearest_premise\')',
  '$this (Yes)' => '$this(HelloWorld)',
  'native $innerHits' => 'array',
  'native $this' => '$this(HelloWorld)',
  'condition about $innerHits #1' => 'if $innerHits=array then $innerHits is array&hasOffset(\'nearest_premise\') (Yes)',
)
```

wdyt? do you agree?

**Comment by clxmstaab** [comment]
ohh.. I think the bug is that we have `'$innerHits (Yes)' => '*NEVER*',`  but it should be `array` instead in Falsey-scope

**Comment by ondrejmirtes** [comment]
Yes, `'$innerHits (Yes)' => '*NEVER*',` is one issue but also we should not probably create conditionals about the same variable, it doesn't make sense:

```
'condition about $innerHits #1' => 'if $innerHits=array then $innerHits is ...
```

ondrejmirtes:
Ohh I fixed it :)

staabm:
really nice, thank you. I was looking at it for hours and coulnd't make sense of it.

clxmstaab:
I have a test-case locally for the conditon bug. let me push it into the PR

ondrejmirtes:
One part didn't make much sense to me so I changed it:

* https://github.com/phpstan/phpstan-src/pull/2710/commits/6e72a32b80aff83f7b9d383c57c5cfc5e18df48f
* https://github.com/phpstan/phpstan-src/pull/2710/commits/e718636917b395eedca8562bcf7881cb81f3d1b4

The change about PHPDoc certainty is probably a bug in merging scopes, I don't care much about that right now.

I will be merging this soon. Thank you!

ondrejmirtes (approved):

staabm:
thanks for your support!

PR #2715, opened 2023-11-06
fix: star projection is always within template type bounds

jiripudil:
closes phpstan/phpstan#10097


ondrejmirtes:
Thank you very much!

PR #2718, opened 2023-11-08
Add `ConstantsInTraitsRule`

paulbalandan:
This PR adds a rule to flag usage of constants inside traits, which feature is only available on PHP 8.2 and later.

Ref: https://php.watch/versions/8.2/constants-in-traits

### DIFF
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -88,6 +88,7 @@ rules:
 	- PHPStan\Rules\Properties\MissingReadOnlyPropertyAssignRule
 	- PHPStan\Rules\Properties\PropertyAttributesRule
 	- PHPStan\Rules\Properties\ReadOnlyPropertyRule
+	- PHPStan\Rules\Traits\ConstantsInTraitsRule
 	- PHPStan\Rules\Variables\UnsetRule
 	- PHPStan\Rules\Whitespace\FileWhitespaceRule
 
--- a/src/Php/PhpVersion.php
+++ b/src/Php/PhpVersion.php
@@ -232,4 +232,9 @@ class PhpVersion
 		return $this->versionId >= 80300;
 	}
 
+	public function supportsConstantsInTraits(): bool
+	{
+		return $this->versionId >= 80200;
+	}
+
 }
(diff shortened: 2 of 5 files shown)


ondrejmirtes (changes requested):
Hi, this really does not have to be a Collector+Rule, just a simple Rule.

paulbalandan:
> Hi, this really does not have to be a Collector+Rule, just a simple Rule.

Initially, I tried using rule only targeting the `ClassConst` node but not getting any luck. Upon dumping the nodes, I found the node for traits is `CollectedDataNode` so I tried working on that direction. Did I miss anything?

ondrejmirtes:
Yes, ClassConst rule should be fine, please push your earlier attempt and I'll see what's wrong.

ondrejmirtes:
You need to use the trait by a class in order to have it inspected.

Alternatively (maybe a better option) is to hook your rule onto Trait_ node and inspect its stmts.

ondrejmirtes:
Looks pretty okay now. FYI there are dozens of similar rules that could be implemented that PHPStan does not check but PHP crashes on it. Would you be interested in looking at the list and maybe implementing some of them? Thank you!

paulbalandan:
> Looks pretty okay now. FYI there are dozens of similar rules that could be implemented that PHPStan does not check but PHP crashes on it. Would you be interested in looking at the list and maybe implementing some of them? Thank you!

Sure. Maybe you have a list or something?

paulbalandan:
The larastan tests failures are not mine as I see from the larastan repo their build is failing. Not sure for phpstan-doctrine, though.

ondrejmirtes:
Thank you very much. The failures are not your fault, you certainly couldn't break that with a new rule :)

As for the other rule ideas. Here's one list:

* https://github.com/phpstan/phpstan/milestone/2

And I also have a pretty comprehensive list in a private  Trello board. If you give me your email address, I can invite you to it.

paulbalandan:
Thanks! Here's my email: paulbalandan[at]gmail.com

ondrejmirtes:
Just invited you, thanks!

PR #2722, opened 2023-11-09
Fix return value of mailparse_msg_get_part()

jlherren:
It may return false when the part was not found.

https://github.com/php/pecl-mail-mailparse/blob/7f58ef4910fdc6a41f15527a4aa1e16cbdb6cdfd/mailparse.c#L1563


ondrejmirtes:
Thank you.

PR #2723, opened 2023-11-10
fix reflection golden test on push

schlndh:
I noticed that the reflection golden test passed on [this commit](https://github.com/phpstan/phpstan-src/commit/c86d4af618446d48fee79f035daa6c349ab8f012). It turns out that I had the ref wrong for push - it generated the golden data from the commit itself.


ondrejmirtes:
Thank you!

PR #2724, opened 2023-11-10
Merge InvalidPHPStanDocTagRule into InvalidPhpDocTagValueRule.

mad-briller:
I noticed that the code in `InvalidPHPStanDocTagRule` and `InvalidPhpDocTagValueRule` is very similar, and phpstan is parsing all relevant phpdocs twice in these rules due to this duplication of code.

An alternative considered was using `FileTypeMapper` to get a cached phpdoc instead, but because `->getResolvedPhpDoc` does not have an expression level scope it was not feasible.

I understand it may be preferable to keep the old structure, so feel free to decline this pr if this is undesired.


mad-briller:
not sure what's going on with CI but the failures seem unrelated

ondrejmirtes:
Hi, I'd like to understand what is the goal here: is it performance, is it code deduplication, is it something else?

mad-briller:
The goal is performance, but after further investigation i think the impact is not as big as i initially believed.

When profiling phpstan analyzing a single file, i saw that these 2 rules were taking a significant portion of the execution time (~6% iirc, lost the profiles), so i looked into what they were doing to reduce the cost and this stood out, as the profiler reported most of the time is spent in the rules is in the phpdoc `parse()` method.

However i realised that profiling using `bin/phpstan test.php` to analyse a file in the root of the `phpstan-src` repo influenced the results, as the `StubValidator` is running these rules over all the stub files in phpstan, skewing the results and over-inflating their costs relative to what is being analysed, and causing me to believe this change to had a bigger overall impact.

Instead, i analysed a ~50 file repo and the InvalidPHPStanDocTagRule takes 0.2% of time and InvalidPhpDocTagValueRule 0.6% of time. 

After this change the InvalidPhpDocTagValueRule still only takes 0.6% of time, so this is still a performance win as the cost of the InvalidPHPStanDocTagRule is mostly absorbed into the InvalidPhpDocTagValueRule, but in absolute terms it is not as massive as i first believed so it's at your discretion whether to accept it.

ondrejmirtes:
Yeah, I'm gonna close this. I prefer the code to look how it looks now. One rule is for reporting parse errors in PHPDocs, the other for reporting unknown `@phpstan-` tags.

PR #2727, opened 2023-11-12
Change iptcparse return type

ManuelHu:
`iptcparse` should only return string-indexed arrays with string[]-data associated, if I am reading the correctly: https://github.com/php/php-src/blob/524b13460752fba908f88e3c4428b91fa66c083a/ext/standard/iptc.c#L355-L365.


ondrejmirtes:
Thank you.

PR #2728, opened 2023-11-12
Consider last `else if` of an `if` / `else if` combination to be the last condition

herndlm:
Closes https://github.com/phpstan/phpstan/issues/10113

### DIFF
--- a/src/Parser/LastConditionVisitor.php
+++ b/src/Parser/LastConditionVisitor.php
@@ -26,6 +26,17 @@ class LastConditionVisitor extends NodeVisitorAbstract
 			}
 		}
 
+		if (
+			$node instanceof Node\Stmt\If_
+			&& $node->else instanceof Node\Stmt\Else_
+			&& count($node->else->stmts) === 1
+			&& $node->else->stmts[0] instanceof Node\Stmt\If_
+			&& $node->else->stmts[0]->else === null
+		) {
+			// consider last else if of an if / else if combination to be the last condition
+			$node->else->stmts[0]->cond->setAttribute(self::ATTRIBUTE_NAME, true);
+		}
+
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Would this error go away too? https://phpstan.org/r/6e31b9d4-2f6b-4c02-9c8c-72f62dc720f8

herndlm:
> Would this error go away too? https://phpstan.org/r/6e31b9d4-2f6b-4c02-9c8c-72f62dc720f8

no, but looks like the related rule (`UnreachableIfBranchesRule`) is not using the `isLastCondition` attribute yet.

herndlm:
> > Would this error go away too? https://phpstan.org/r/6e31b9d4-2f6b-4c02-9c8c-72f62dc720f8
> 
> no, but looks like the related rule (`UnreachableIfBranchesRule`) is not using the `isLastCondition` attribute yet.

did a deeper check, the visitor is not handling `else` at all, so it will never be set. if this should go away that specific rule needs an adaption I'd say, this PR is basically unrelated IMO.

ondrejmirtes:
I'd accept a PR that runs a special visitor in RichParser before all the visitors that would transform AST with many `else if`s into a flat `elseif` structure :)

herndlm:
superseded by https://github.com/phpstan/phpstan-src/pull/2799

herndlm:
@ondrejmirtes since https://github.com/phpstan/phpstan-src/pull/2799 didn't work out, would you consider this solution after all?

ondrejmirtes:
This probably needs a different solution than transforming the AST for everyone.

herndlm:
> This probably needs a different solution than transforming the AST for everyone.

yeah. so, https://github.com/phpstan/phpstan-src/pull/2799 was the AST transformation which didn't work. this here was the less invasive `LastConditionVisitor` which worked fine :)

PR #2730, opened 2023-11-13
Detect non-abstract methods with no body

paulbalandan:
Fixes https://github.com/phpstan/phpstan/issues/4244

### DIFF
--- a/src/Rules/Methods/AbstractMethodInNonAbstractClassRule.php
+++ b/src/Rules/Methods/AbstractMethodInNonAbstractClassRule.php
@@ -27,17 +27,29 @@ class AbstractMethodInNonAbstractClassRule implements Rule
 		}
 
 		$class = $scope->getClassReflection();
-		if ($class->isAbstract()) {
-			return [];
+
+		if (!$class->isAbstract() && $node->isAbstract()) {
+			return [
+				RuleErrorBuilder::message(sprintf(
+					'%s %s contains abstract method %s().',
+					$class->isInterface() ? 'Interface' : 'Non-abstract class',
+					$class->getDisplayName(),
+					$node->name->toString(),
+				))->nonIgnorable()->build(),
+			];
 		}
 
-		if (!$node->isAbstract()) {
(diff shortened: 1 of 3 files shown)


src/Rules/Methods/AbstractMethodInNonAbstractClassRule.php
```diff
@@ -20,24 +20,39 @@ public function getNodeType(): string
 		return Node\Stmt\ClassMethod::class;
 	}
 
+	/**
+	 * @param Node\Stmt\ClassMethod $node
```
ondrejmirtes: This PHPDoc is not needed. Both PHPStan and PhpStorm know what's going into the method.

tests/PHPStan/Rules/Methods/AbstractMethodInNonAbstractClassRuleTest.php
```diff
@@ -20,15 +20,22 @@ protected function getRule(): Rule
 
 	public function testRule(): void
 	{
-		$this->analyse([__DIR__ . '/data/abstract-method.php'], [
+		$this->analyse([
+			__DIR__ . '/data/abstract-method.php',
+			__DIR__ . '/data/bug-4244.php',
```
ondrejmirtes: Please add a separate test method for testing a new file.

ondrejmirtes (changes requested):

src/Rules/Methods/AbstractMethodInNonAbstractClassRule.php
paulbalandan: Done.

tests/PHPStan/Rules/Methods/AbstractMethodInNonAbstractClassRuleTest.php
paulbalandan: Done.

ondrejmirtes:
Awesome, thank you!

PR #2731, opened 2023-11-13
Detect properties declared in interface

paulbalandan:
Fixes https://github.com/phpstan/phpstan/issues/8915

### DIFF
--- a/tests/PHPStan/Rules/Properties/PropertiesInInterfaceRuleTest.php
+++ b/tests/PHPStan/Rules/Properties/PropertiesInInterfaceRuleTest.php
@@ -1,0 +1,33 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Properties;
+
+use PHPStan\Rules\Rule;
+use PHPStan\Testing\RuleTestCase;
+
+/**
+ * @extends RuleTestCase<PropertiesInInterfaceRule>
+ */
+class PropertiesInInterfaceRuleTest extends RuleTestCase
+{
+
+	protected function getRule(): Rule
+	{
+		return new PropertiesInInterfaceRule();
+	}
+
+	public function testRule(): void
+	{
+		$this->analyse([__DIR__ . '/data/properties-in-interface.php'], [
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Lint fails - path to this file needs to be added as `--exclude` in the Makefile.

tests/PHPStan/Rules/Properties/PropertiesInInterfaceRuleTest.php
```diff
@@ -0,0 +1,29 @@
...
+{
+
+	protected function getRule(): Rule
+	{
+		return new PropertiesInInterfaceRule();
+	}
+
+	public function testRule(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-8915.php'], [
```
mvorisek: https://3v4l.org/nstX1 in another test file would be good to test too

tests/PHPStan/Rules/Properties/PropertiesInInterfaceRuleTest.php
```diff
@@ -0,0 +1,40 @@
...
+	protected function getRule(): Rule
+	{
+		return new PropertiesInInterfaceRule();
+	}
+
+	public function testRule(): void
+	{
+		$this->analyse([
+			__DIR__ . '/data/properties-in-interface.php',
+			__DIR__ . '/data/bug-8915.php',
```
ondrejmirtes: Same point here - separate test method pleaase :) (And FYI personally I wouldn't add this file here at all, the rule is straightforward and the test does not really test anything different :))

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Properties/PropertiesInInterfaceRuleTest.php
paulbalandan: Thanks for the pointers! I retained the first test.

ondrejmirtes:
Thank you!

PR #2733, opened 2023-11-13
Add VariadicParametersDeclarationRule

paulbalandan:
Closes https://github.com/phpstan/phpstan/issues/3802

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/variadic-parameters-declaration.php
+++ b/tests/PHPStan/Rules/Functions/data/variadic-parameters-declaration.php
@@ -1,0 +1,23 @@
+<?php declare(strict_types=1);
+
+namespace VariadicParametersDeclaration;
+
+class HelloWorld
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Functions/data/variadic-parameters-declaration.php
```diff
@@ -0,0 +1,19 @@
+<?php declare(strict_types=1);
+
+namespace VariadicParametersDeclaration;
+
+class HelloWorld
+{
+	public function multipleVariadicParams(int ...$p1, int ...$p2): void
```
ondrejmirtes, line 7: It's always a good idea to include also an example where the rule is not supposed to report anything.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #2736, opened 2023-11-14
Fix ExponentiateHelper throwing TypeError for string operands.

mad-briller:
closes https://github.com/phpstan/phpstan/issues/10125


mad-briller:
hmm perhaps this isn't the best approach looking at ci...

PR #2738, opened 2023-11-15
VarTagTypeRuleHelper: fix widening array shapes

janedbal:
Resolves https://github.com/phpstan/phpstan/issues/10130

### DIFF
--- a/src/Rules/PhpDoc/VarTagTypeRuleHelper.php
+++ b/src/Rules/PhpDoc/VarTagTypeRuleHelper.php
@@ -9,7 +9,9 @@ use PHPStan\Node\Expr\GetOffsetValueTypeExpr;
 use PHPStan\PhpDoc\Tag\VarTag;
 use PHPStan\Rules\RuleError;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeUtils;
@@ -128,7 +130,8 @@ class VarTagTypeRuleHelper
 	private function shouldVarTagTypeBeReported(Node\Expr $expr, Type $type, Type $varTagType): bool
 	{
 		if ($expr instanceof Expr\Array_ && $expr->items === []) {
-			return false;
+			$type = new ArrayType(new MixedType(), new MixedType());
+			return $type->isSuperTypeOf($varTagType)->no(); // allow annotating empty array with anything that can accept it (e.g. int[]|null)
 		}
 
 		if ($expr instanceof Expr\New_) {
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/PhpDoc/VarTagChangedExpressionTypeRuleTest.php
```diff
@@ -48,4 +48,9 @@ public function testAssignOfDifferentVariable(): void
 		]);
 	}
 
+	public function testBug10130(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-10130.php'], []);
```
ondrejmirtes: Your expected output is "Report even third widening." but your change silences all.

BTW the reasoning behind the original code is that sometimes you want to do this:

```php
/** @var int|null */
$a = null;
foreach ($list as $v) {
    // sometimes assign $a
}
```

So I'd accept a change that still does not report this (including array shapes etc.):

```php
/** @var int|null */
$a = null;
```

But reports this:

```php
// $null is null
/** @var int|null */
$a = $null;
```

It could be achieved by allowing this only for `ConstFetch`, `Scalar` nodes, and `Array_` node.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/PhpDoc/VarTagChangedExpressionTypeRuleTest.php
janedbal: > Your expected output is "Report even third widening." but your change silences all.

I did not, just the test was wrong. Fixed

> It could be achieved by allowing this only for ConstFetch, Scalar nodes, and Array_ node.

Yeah something like that is needed. Let me try.

src/Rules/PhpDoc/VarTagTypeRuleHelper.php
```diff
@@ -127,6 +127,10 @@ private function containsPhpStanType(Type $type): bool
 
 	private function shouldVarTagTypeBeReported(Node\Expr $expr, Type $type, Type $varTagType): bool
 	{
+		if ($expr instanceof Expr\Array_ && $expr->items === []) {
+			return false;
```
ondrejmirtes: This is wrong. It's gonna silence this case:

```php
/** @var int */
$a = []; // nonsense
```

ondrejmirtes (changes requested):

src/Rules/PhpDoc/VarTagTypeRuleHelper.php
janedbal: Just fixed that.

janedbal:
This starts to find real bugs: https://github.com/phpstan/phpstan-src/actions/runs/6877930111/job/18706649361?pr=2738#step:6:289

tests/PHPStan/Rules/PhpDoc/VarTagChangedExpressionTypeRuleTest.php
janedbal: The only issue was Array_ so added exception for that.

ondrejmirtes:
Thank you!

janedbal:
Thank you!

PR #2739, opened 2023-11-15
Update array_filter signature in PHP 8.0 to allow null as callback

andyexeter:
The callback parameter of array_filter has been nullable since PHP 8.0 but PHPStan doesn't currently allow it:

https://phpstan.org/r/307000c9-843b-42af-9910-baa02ec72f2f


andyexeter:
Closing in favour of #2740

PR #2740, opened 2023-11-15
Update array_filter signature in PHP 8.0 to allow null as callback

andyexeter:
The callback parameter of array_filter has been nullable since PHP 8.0 but PHPStan doesn't currently allow it:

https://phpstan.org/r/307000c9-843b-42af-9910-baa02ec72f2f

### DIFF
--- a/resources/functionMap_php80delta.php
+++ b/resources/functionMap_php80delta.php
@@ -159,7 +159,6 @@ return [
 	],
 	'old' => [
 		'array_combine' => ['associative-array|false', 'keys'=>'string[]|int[]', 'values'=>'array'],
-		'array_filter' => ['array', 'input'=>'array', 'callback='=>'callable(mixed,mixed):bool|callable(mixed):bool|null', 'flag='=>'int'],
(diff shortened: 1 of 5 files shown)


andyexeter:
I'm not sure why the tests are failing. All I did was copy the signature from the main functionMap.php file and add `|null` to the `callback` signature.

ondrejmirtes:
Please reproduce your problem in CallToFunctionParametersRuleTest. You'll find that it will still most likely fail.

`array_filter` is a special function. You need to add `NullType` here too: https://github.com/phpstan/phpstan-src/blob/d846b0226665cb4e781a4fa0b47602922fc52766/src/Reflection/ParametersAcceptorSelector.php#L174-L180 (create `new UnionType` of that `CallableType` and `new NullType`).

Please disregard the irrelevant failures.

andyexeter:
@ondrejmirtes I think this is ready for review, the failing tests appear unrelated.

ondrejmirtes:
Thank you!

PR #2741, opened 2023-11-15
Implement MagicClassConstantRule

staabm:
closes https://github.com/phpstan/phpstan/issues/10099

background: https://www.php.net/manual/en/language.constants.magic.php


conf/config.level0.neon
```diff
@@ -56,6 +56,7 @@ rules:
 	- PHPStan\Rules\Classes\NonClassAttributeClassRule
 	- PHPStan\Rules\Classes\TraitAttributeClassRule
 	- PHPStan\Rules\Constants\FinalConstantRule
+	- PHPStan\Rules\Constants\MagicClassConstantRule
```
ondrejmirtes: This should be enabled only with bleeding edge. The code does not error, it just returns an empty string AFAIK.

ondrejmirtes (changes requested):

src/Rules/Constants/MagicClassConstantRule.php
```diff
@@ -0,0 +1,33 @@
...
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if ($scope->isInClass()) {
+			return [];
+		}
+
+		return [
+			RuleErrorBuilder::message(
+				sprintf('Magic constant %s cannot be used outside a class.', $node->getName()),
```
ondrejmirtes: The message should be different. It just returns an empty string.

ondrejmirtes (changes requested):

conf/config.level0.neon
staabm: I have used a feature flag with a more general name, so I can use it for similar rules for `__FUNCTION__` and `__METHOD__`

ondrejmirtes:
Can you also please go through other magic constants and report them when used in a wrong way? Maybe it can all be done inside this single rule.

staabm:
there is one interessting leftover case:

```php 
// test.php
function doFooParams(
	string $file = __FILE__,
	int $line = __LINE__,
	string $class = __CLASS__,
	string $dir = __DIR__,
	string $namespace = __NAMESPACE__,
	string $method = __METHOD__,
	string $function = __FUNCTION__,
	string $trait = __TRAIT__
): void
{
}
```

since used in the parameter-list and not the function body it false-positives with

https://3v4l.org/BbZWn

```
$ php bin/phpstan analyze test.php --debug
Note: Using configuration file C:\dvl\Workspace\phpstan-src\phpstan.neon.dist.
C:\dvl\Workspace\phpstan-src\test.php
 ------ -----------------------------------------------------------------------------
  Line   test.php
 ------ -----------------------------------------------------------------------------
  6      Magic constant __CLASS__ is always empty when used outside a class.
  8      Magic constant __NAMESPACE__ is always empty when used in global namespace.
  9      Magic constant __METHOD__ is always empty when used outside a function.
  10     Magic constant __FUNCTION__ is always empty when used outside a function.
 ------ -----------------------------------------------------------------------------
```

staabm:
we could fix the above case by moving `enterFunction` before the `processParamNode` call in 

https://github.com/phpstan/phpstan-src/blob/231ed3e1feff8306318b1ba36c7d5ebf76542136/src/Analyser/NodeScopeResolver.php#L451-L459

wdyt?

staabm (commented):
Instead of changing NodeScopeResolver I went with adding a dedicated new NodeVisitor to prevent the false positives

src/Rules/Constants/MagicConstantContextRule.php
```diff
@@ -0,0 +1,71 @@
...
+			return [
+				RuleErrorBuilder::message(
+					sprintf('Magic constant %s is always empty when used outside a function.', $node->getName()),
+				)->build(),
+			];
+		} elseif ($node instanceof MagicConst\Namespace_) {
+			if ($scope->getNamespace() === null) {
+				return [
+					RuleErrorBuilder::message(
+						sprintf('Magic constant %s is always empty when used in global namespace.', $node->getName()),
```
ondrejmirtes: Please deleted "when used" from all messages. They'll still make sense.

src/Rules/Constants/MagicConstantContextRule.php
```diff
@@ -0,0 +1,71 @@
...
+				)->build(),
+			];
+		} elseif ($node instanceof MagicConst\Trait_) {
+			if ($scope->isInTrait()) {
+				return [];
+			}
+
+			return [
+				RuleErrorBuilder::message(
+					sprintf('Magic constant %s is always empty when used outside a trait-using-class.', $node->getName()),
```
ondrejmirtes: "outside a trait"

ondrejmirtes (changes requested):

staabm:
(btw: interessting that issuebot did not realize that the snippet from issue https://github.com/phpstan/phpstan/issues/10099 gets a new error with this PR - locally it works as expected though)

src/Rules/Constants/MagicConstantContextRule.php
```diff
@@ -0,0 +1,71 @@
...
+				return [];
+			}
+
+			return [
+				RuleErrorBuilder::message(
+					sprintf('Magic constant %s is always empty outside a trait.', $node->getName()),
+				)->build(),
+			];
+		} elseif ($node instanceof MagicConst\Method || $node instanceof MagicConst\Function_) {
+			// https://3v4l.org/3CAHm
```
ondrejmirtes: This doesn't seem right. What about `private $foo = __METHOD__;`?

Also a false positive here - you can have an anonymous function outside a class/function and it's going to report `'{closure}'`.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #2742, opened 2023-11-15
RuleTestCase: add analyseFileWithErrorsAsComments

janedbal:
I'd like to suggest this feature I'm using for about 4-5 years in my projects. E.g. [here](https://github.com/shipmonk-rnd/phpstan-rules/blob/master/tests/RuleTestCase.php#L20).

Benefits:
- **immediately visible where the error is raised** and where not
- **errors float with code**, so you can easily add code in between existing errors without modifying expected errors (currently even a single use statement added breaks everything)

Downsides:
- no support for multiple errors on a single line
- when traits come into play, error lines are not always in ascending order so it mismatches

-------------------

I believe that at least 90% tests could benefit from this. Other can stick with the old approach.

### DIFF
--- a/tests/PHPStan/Rules/Cast/data/invalid-cast.php
+++ b/tests/PHPStan/Rules/Cast/data/invalid-cast.php
@@ -4,7 +4,7 @@ function (
 	string $str
 ) {
 	(string) $str;
-	(string) new \stdClass();
+	(string) new \stdClass(); // error: Cannot cast stdClass to string.
 	(string) new \Test\ClassWithToString();
 
 	(object) new \stdClass();
@@ -20,8 +20,8 @@ function (
 	(int) "123"; // ok
 	(int) "blabla";
 
-	(int) new \stdClass();
-	(float) new \stdClass();
+	(int) new \stdClass(); // error: Cannot cast stdClass to int.
+	(float) new \stdClass(); // error: Cannot cast stdClass to float.
 
 	(string) fopen('php://memory', 'r');
 	(int) fopen('php://memory', 'r');
@@ -32,20 +32,20 @@ function (
 ) {
 	/** @var object $object */
 	$object = doFoo();
-	(string) $object;
+	(string) $object; // error: Cannot cast object to string.
 
 	if (method_exists($object, '__toString')) {
 		(string) $object;
 	}
 
-	(string) $foo;
+	(string) $foo; // error: Cannot cast Test\Foo to string.
 	if (method_exists($foo, '__toString')) {
 		(string) $foo;
 	}
 
 	/** @var array|float|int $arrayOrFloatOrInt */
 	$arrayOrFloatOrInt = doFoo();
-	(string) $arrayOrFloatOrInt;
+	(string) $arrayOrFloatOrInt; // error: Cannot cast array|float|int to string.
 };
 
 function (
--- a/src/Testing/RuleTestCase.php
+++ b/src/Testing/RuleTestCase.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Testing;
 
+use LogicException;
 use PhpParser\Node;
 use PHPStan\Analyser\Analyser;
 use PHPStan\Analyser\Error;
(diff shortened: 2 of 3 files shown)


tests/PHPStan/Rules/Cast/data/invalid-cast.php
```diff
@@ -4,7 +4,7 @@ function (
 	string $str
 ) {
 	(string) $str;
-	(string) new \stdClass();
+	(string) new \stdClass(); // error: Cannot cast stdClass to string.
```
staabm, line 7: what comes to mind: maybe we can do it similar to `assertType()` and `assertVariableCertainty()`

```suggestion
	assertErrorMessage('Cannot cast stdClass to string.', (string) new \stdClass());
```

?

ondrejmirtes: Yes, I'd like something like that too. Perhaps something like `assertErrorOnNextLine('message');`

janedbal: That is much less flexible. You cannot call method between class methods etc.

janedbal: Error can be anywhere, in phpdoc, at class level, at trait usage. Comment is the only thing you can place anywhere.

janedbal: Also, the implemented version works even within phpdocs (to be supported in https://github.com/phpstan/phpstan-src/pull/2807)

DaveLiddament:
I've created a similar package to do help with testing custom rules: https://github.com/DaveLiddament/phpstan-rule-test-helper.  This provides `AbstractRuleTestCase` to extend for testing custom rules. 

This allows you to either hard code the error message in the fixture file (similar to @janedbal's suggestion) . E.g. 
```php
    $item->updateName("world"); // ERROR can not call method
```

--------

If the same error message is repeated multiple times, then you can also override the `getErrorFormatter` with the error message. This means error messages need updating only in one place in the test: 
```php 
  protected function getErrorFormatter(): string 
  {
    return "Can not call method";
  } 
```

The fixture simplifies to: 
```php
    $item->updateName("world"); // ERROR 
```

--------


Finally you can provide context in the error messages:
```php 
  protected function getErrorFormatter(): string 
  {
    return "Can not call method {0} from {1}";
  } 
```

The fixture error has additional contextual information, if there is more than 1 bit it is separated by `|` character.
```php
    $item->updateName("world"); // ERROR Item::updateName|Foo
```

In the above example the expected error message would be: `Can not call method Item::updateName from Foo`
 
--------
 
 
 It be nice to either bring similar functionality into PHPStan, or highlight helper packages that can help with testing.

ondrejmirtes:
Some thoughts after letting this sit for a while:

* Expecting errors in comments is very similar to ignoring errors
* If we implement a similar system I want it to work like `assertType` - that it can be tested both by running PHPUnit and simply by analysing code with PHPStan
* So what about `@phpstan-error An exact error message` that works the same way as `@phpstan-ignore` with an error identifier in 1.11.x?

DaveLiddament:
> Some thoughts after letting this sit for a while:
> 
> * Expecting errors in comments is very similar to ignoring errors
> * If we implement a similar system I want it to work like `assertType` - that it can be tested both by running PHPUnit and simply by analysing code with PHPStan
> * So what about `@phpstan-error An exact error message` that works the same way as `@phpstan-ignore` with an error identifier in 1.11.x?

I like `@phpstan-error`. 

I love the fact you'd then not even need to create PHPUnit tests, you can just point PHPStan at the code snippet.

janedbal:
> So what about @phpstan-error An exact error message that works the same way as @phpstan-ignore with an error identifier in 1.11.x?

I'm not sure that the next-line magic is needed here. Imo not.

janedbal:
For the meantime, we open-sourced our solution to a standalone package with just this functionality (inline error asserts in tested code): https://github.com/shipmonk-rnd/phpstan-dev

PR #2743, opened 2023-11-15
add named arguments variants to reflection golden test

schlndh:
I'm adding the named arguments variants to the reflection golden test as promised. I added a condition to exclude some cases where the names in the function map are already correct. There are some more duplicates due to the parameter names in the positional variant being fixed later via phpstorm stubs. 

I also found that the phpstorm stubs are sometimes incorrect (at least `XSLTProcessor::setParameter` and `PDOStatement::setFetchMode`). So I'd remove the rename for named params in next PR.


ondrejmirtes:
Thank you!

PR #2744, opened 2023-11-15
Fully qualify the name of the `#[\Override]` attribute in error messages

TimWolla:
As previously discussed on https://twitter.com/OndrejMirtes/status/1724492688932688014. Can you point me to the place where I can find the documentation to send a PR?

----------

This makes the attribute definition from the error message copy and paste safe for namespaced classes.

As PHP attributes do not need to be backed by a class, using the non-qualified `#[Override]` variant without a matching `use Override;` appears to look correct at a quick glance, but will not actually do anything useful.


ondrejmirtes:
Thank you very much!

PR #2745, opened 2023-11-17
Add InvalidCallablePropertyTypeRule

paulbalandan:
Closes phpstan/phpstan#4424

### DIFF
--- a/src/Rules/Properties/InvalidCallablePropertyTypeRule.php
+++ b/src/Rules/Properties/InvalidCallablePropertyTypeRule.php
@@ -1,0 +1,65 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Properties;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\ClassPropertyNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\CallableType;
+use PHPStan\Type\IntersectionType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeTraverser;
+use PHPStan\Type\UnionType;
+use function sprintf;
+
+/**
+ * @implements Rule<ClassPropertyNode>
+ */
+class InvalidCallablePropertyTypeRule implements Rule
+{
+
+	public function getNodeType(): string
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/Properties/data/invalid-callable-property-type.php
```diff
@@ -0,0 +1,59 @@
...
+	public callable $a;
+
+	/** @var callable(): void|null */
+	public ?callable $b;
+
+	/** @var callable(): void|string */
+	public callable|string $c;
+
+	/** @var Foo&callable(): void */
+	public callable&Foo $d;
```
ondrejmirtes: This isn't valid for a different reason that should be a separate rule: https://3v4l.org/bM6Ap#v8.2.11 (because it's not valid anywhere, not just in properties)

src/Rules/Properties/InvalidCallablePropertyTypeRule.php
```diff
@@ -0,0 +1,120 @@
...
+	 */
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$nativeType = $node->getNativeType();
+
+		if ($nativeType === null) {
+			return [];
+		}
+
+		if ($this->isExplicitCallableType($nativeType)) {
```
ondrejmirtes: This rule can be simplified a lot:

1) Please still keep `if ($nativeType === null) {`.
2) Get `$node->getClassReflection()` and get property reflection via `->getNativeProperty()`.
3) Get `PhpPropertyReflection::getNativeType()`.
4) Use `TypeTraverser::map` to extract CallableType:

```php
TypeTraverser::map($classType, static function (Type $type, callable $traverse) use (&$callableTypes): Type {
		return $traverse($type);
	}
	if ($type instanceof CallableType) {
		$callableTypes[] = $type;
	}
	
	return $type;
});
```

We don't need different error message for different situations.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Properties/data/invalid-callable-property-type.php
paulbalandan: Removed.

ondrejmirtes:
Thank you!

PR #2748, opened 2023-11-19
fix named arguments for few multi-variant methods

schlndh:
There are a few cases where PHPStorm stubs have the wrong parameter names. So I'd remove the parameter rename for named params. Since named params are only available from PHP 8, we should have the correct names from php-8-stubs.

### DIFF
--- a/src/Reflection/Php/PhpClassReflectionExtension.php
+++ b/src/Reflection/Php/PhpClassReflectionExtension.php
@@ -593,6 +593,10 @@ class PhpClassReflectionExtension
 
 								$phpDocParameterNameMapping[$signatureParameters[$paramI]->getName()] = $reflectionParameter->getName();
 							}
+
+							if ($signatureType === 'named') {
+								$phpDocParameterNameMapping = [];
+							}
 						}
 					}
 					$variantsByType[$signatureType][] = $this->createNativeMethodVariant($methodSignature, $stubPhpDocParameterTypes, $stubPhpDocParameterVariadicity, $stubPhpDocReturnType, $phpDocParameterTypes, $phpDocReturnType, $phpDocParameterNameMapping, $stubPhpParameterOutTypes, $phpDocParameterOutTypes);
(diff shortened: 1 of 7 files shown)


src/Reflection/Php/PhpClassReflectionExtension.php
```diff
@@ -593,6 +593,10 @@ private function createMethod(
 
 								$phpDocParameterNameMapping[$signatureParameters[$paramI]->getName()] = $reflectionParameter->getName();
 							}
+
+							if ($signatureType === 'named') {
+								$phpDocParameterNameMapping = [];
```
ondrejmirtes: What is the significance of this change?

ondrejmirtes (changes requested):
Please tell me where the "named argument variants" are being read from phpstorm-stubs. I'd say that PHP functions in php-8-stubs should always be read from there.

schlndh:
> Please tell me where the "named argument variants" are being read from phpstorm-stubs. I'd say that PHP functions in php-8-stubs should always be read from there.

It's not the "named argument variants" that are being read from phpstorm-stubs. But the arguments are renamed based on the phpstorm-stubs. It's exactly the change that you highlighted (`$phpDocParameterNameMapping = []`) that turns this off for named arguments.

ondrejmirtes:
So let's say that in phpstorm-stubs there's:

```php
/**
 * @param int $a
 * @param string $b
 */
public function foo($a, $b): void
```

And in reality the named arguments are named:

```php
public function foo($c, $d): void
```

I'd still expect them to be typed with int and string based on the PHPDoc from phpstorm-stubs.

Is that what this PR changes? Or something else?

schlndh:
You're right - the `$phpDocParameterNameMapping` is still needed inside `createNativeMethodVariant` to match the phpdoc types etc. Currently it makes no difference since the named variants are only used for multi-variant methods. But if they were to expand, it might have caused some bugs. I changed it accordingly.

ondrejmirtes:
Thank you!

PR #152, opened 2020-03-02
Update setcookie parameters

spl1nes:
Add setcookie parameters 'samesite' & 'url_encode' according to: https://github.com/php/php-src/blob/b9c157726bf4229d694ec840daf4dc6ccb1c3898/ext/standard/head.h#L38 and https://github.com/php/php-src/blob/b9c157726bf4229d694ec840daf4dc6ccb1c3898/ext/standard/head.c#L77

Whether setrawcookie also supports this is not clear from the `php-src` code.

PR #2752, opened 2023-11-20
remove fake params from setcookie/setrawcookie

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/10171

It turns out that this is wrong: https://github.com/phpstan/phpstan-src/pull/152 When I try to call `setcookie` with all 9 parameters it says that max 7 are accepted (PHP 8.2), named arguments such as `samesite: 'lax'` also don't work. I'm guessing that the arguments in the `php_setcookie` C function are there to cover the different variants + `setcookie` vs `setrawcookie`, but I haven't checked it.

https://3v4l.org/eN9lM


schlndh:
> However, the documentation states that the second signature, which takes an array as the third parameter, is not usable with named parameters.

The documentation is wrong about this - both variants seem to work with named arguments (at least on PHP 8.2).

schlndh:
The wrong signature from functionMap triggered this condition: https://github.com/phpstan/phpstan-src/blob/49f1667d416f59f472a4b6acd297fc36efa86c03/src/Reflection/SignatureMap/Php8SignatureMapProvider.php#L255 that's why it was excluded from the named variants.

ondrejmirtes:
Wow, these are such tricky problems 😊 Thank you!

PR #2753, opened 2023-11-20
Narrow ReflectionEnum::getBackingType after ReflectionEnum::isBacked

RobertMe:
Fixes phpstan/phpstan#10167

### DIFF
--- a/stubs/ReflectionEnum.stub
+++ b/stubs/ReflectionEnum.stub
@@ -18,4 +18,8 @@ class ReflectionEnum extends ReflectionClass
 	 */
 	public function getCase(string $name): ReflectionEnumUnitCase {}
 
+	/**
+	 * @phpstan-assert-if-true !null $this->getBackingType()
+	 */
+	public function isBacked(): bool {}
 }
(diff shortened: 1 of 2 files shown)


RobertMe:
I'm guessing the tests could be "improved" (not sure whether it should be at root of the file, if it actually can be added in this same file).
Furthermore I "discovered" that the return type actually is dependent on the PHP version (PHP 8.2+ returns `ReflectionNamedType|null` while lower versions return `ReflectionType|null`) so that might require some fix/improvement too.

staabm:
Why did you close the PR?

RobertMe:
> Why did you close the PR?

Because it's not finished (failing tests); I was unsure how to fix the 8.2+ vs 8.1 differences, also because of a lack of input (/"support"); I didn't even continue using the code which lead to this "issue" (so in the end I personally didn't even need it anymore, although I normally would finish PRs that I started, even when I wouldn't need them anymore);
And, to be honest, foremost because of how my more recent [issue](https://github.com/phpstan/phpstan/issues/10298) and [accompanying PR](https://github.com/phpstan/phpstan-src/pull/2825) were handled and I'm not even sure whether I like to continue using PHPStan, or whether to switch to Psalm (although I might regret that if Psalm doesn't validate and infer types from SQL queries 😉).

staabm:
I see, thx. I will have a look at the issue

PR #2755, opened 2023-11-21
Add RedefinedParametersRule

paulbalandan:
Closes https://github.com/phpstan/phpstan/issues/8025

### DIFF
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -75,6 +75,7 @@ rules:
 	- PHPStan\Rules\Functions\InnerFunctionRule
 	- PHPStan\Rules\Functions\ParamAttributesRule
 	- PHPStan\Rules\Functions\PrintfParametersRule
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you!

ondrejmirtes:
Did you please look into this?

> Research what PHP version considers this as an error. I remember a situation that such code was present in a library (React.PHP maybe) and it started failing only on PHP 8.0 or so.

paulbalandan:
> Did you please look into this?
> 
> > Research what PHP version considers this as an error. I remember a situation that such code was present in a library (React.PHP maybe) and it started failing only on PHP 8.0 or so.

I did not look really at reactphp. I just tried redefined parameters at 3v4l.org and changed the PHP version from 7.2 up. All are failing.

ondrejmirtes:
Alright 😊

PR #2756, opened 2023-11-21
Implement AbstractPrivateMethodRule

staabm:
Abstract function cannot be declared private

https://3v4l.org/9V2aN#veol

----

as a followup I would do similar stuff for interfaces: https://3v4l.org/3VBl2

### DIFF
--- a/src/Rules/Methods/AbstractPrivateMethodRule.php
+++ b/src/Rules/Methods/AbstractPrivateMethodRule.php
@@ -48,7 +48,7 @@ class AbstractPrivateMethodRule implements Rule
 				'Private method %s::%s() cannot be abstract.',
 				$method->getDeclaringClass()->getDisplayName(),
 				$method->getName(),
-			))->build(),
+			))->nonIgnorable()->build(),
 		];
 	}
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
1) This has to be allowed in traits
2) Please test interfaces too

src/Rules/Methods/AbstractPrivateMethodRule.php
```diff
@@ -0,0 +1,55 @@
...
+
+		if (!$classReflection->isAbstract() && !$classReflection->isInterface()) {
+			return [];
+		}
+
+		return [
+			RuleErrorBuilder::message(sprintf(
+				'Private method %s::%s() cannot be abstract.',
+				$method->getDeclaringClass()->getDisplayName(),
+				$method->getName(),
```
ondrejmirtes, line 50: Non-ignorable please

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you. Please do final private next.

staabm:
> Please do final private next.

what do you have in mind what https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Rules/Methods/FinalPrivateMethodRule.php is not already doing?

ondrejmirtes:
oh sorry, I didn't know we already have it

PR #2757, opened 2023-11-21
Introduce InvalidLexicalVariablesInClosureUseRule

paulbalandan:
Detect uses of `$this`, superglobals, and same name with parameters for variables used in `use` of closures.

Closes https://github.com/phpstan/phpstan/issues/1855

### DIFF
--- a/src/Rules/Functions/InvalidLexicalVariablesInClosureUseRule.php
+++ b/src/Rules/Functions/InvalidLexicalVariablesInClosureUseRule.php
@@ -1,0 +1,95 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Functions;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function array_filter;
+use function array_map;
+use function in_array;
+use function is_string;
+use function sprintf;
+
+/**
+ * @implements Rule<Node\Expr\Closure>
+ */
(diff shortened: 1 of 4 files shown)


src/Rules/Functions/InvalidLexicalVariablesInClosureUseRule.php
```diff
@@ -0,0 +1,87 @@
...
+					->nonIgnorable()
+					->build();
+				continue;
+			}
+
+			if (!in_array($var, $params, true)) {
+				continue;
+			}
+
+			$errors[] = RuleErrorBuilder::message(sprintf('Cannot use lexical variable $%s as a parameter name.', $var))
```
ondrejmirtes: I think this needs a better message. Something like "Cannot use variable with the same name that's already in parameters".

src/Rules/Functions/InvalidLexicalVariablesInClosureUseRule.php
```diff
@@ -0,0 +1,87 @@
...
+
+		foreach ($node->uses as $use) {
+			if (!is_string($use->var->name)) {
+				continue;
+			}
+
+			$var = $use->var->name;
+
+			if ($var === 'this') {
+				$errors[] = RuleErrorBuilder::message('Cannot use $this as lexical variable.')->nonIgnorable()->build();
```
ondrejmirtes: To mark the exact line in multiline closure declarations, please use `->line` on the RuleErrorBuilder. In all reported errors here. Thanks.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #2758, opened 2023-11-21
Implement AbstractProtectedMethodRule

staabm:
interfaces cannot contain non-public methods

https://3v4l.org/3VBl2


src/Rules/Methods/AbstractProtectedMethodRule.php
```diff
@@ -0,0 +1,51 @@
...
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\InClassMethodNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function sprintf;
+
+/** @implements Rule<InClassMethodNode> */
+class AbstractProtectedMethodRule implements Rule
```
ondrejmirtes: I'd make it clear this is only about interface, both in rule class name, and the error message

ondrejmirtes (changes requested):

src/Rules/Methods/AbstractProtectedMethodRule.php
staabm: I also though about that. 

DX wise there are a lot of error situations. I think we need to make sure, that we don't report too much errors at a time.

- methods must be `public` in interfaces https://3v4l.org/tmQX8
- methods are not allowed to be declared `abstract` in interfaces https://3v4l.org/rlVKF
- the combination of the 2: `private abstract` or `protected abstract` in interfaces https://3v4l.org/LR2DP
- final on abstract is not allowed https://phpstan.org/r/f7056d2d-412e-4698-910f-e5e9439e0648 -> `final private abstract`

thats also the reason why I did not cover interfaces in `AbstractPrivateMethodRule` in the first place, as if we use a different error message in this rule here it would be kind of inconsistent to have

```
interface fooInterface {
	abstract private function sayPrivate() : void; // 'Private method fooInterface::sayPrivate() cannot be abstract.'
	abstract protected function sayProtected() : void; // 'Protected method fooInterface::sayPrivate() cannot be abstract in interface.'
	abstract public function sayPublic() : void;
}
```

ondrejmirtes: I'm fine with saying `Protected method fooInterface::sayProtected() cannot be abstract in interface.`.

staabm: I changed the PR to error about method visibility in interfaces (no matter abstract method or not)

ondrejmirtes:
Thank you.

ondrejmirtes:
FYI just noticed a typo in the class name `MethodVisibitiliyInInterfaceRule`

staabm:
ohh good catch. do you take care of it, or should I PR a name-change?

ondrejmirtes:
Please send a PR

PR #2759, opened 2023-11-21
Implement DeclarePositionRule

staabm:
closes https://github.com/phpstan/phpstan/issues/6195

### DIFF
--- a/src/Rules/Keywords/DeclareStrictTypesRule.php
+++ b/src/Rules/Keywords/DeclareStrictTypesRule.php
@@ -33,8 +33,10 @@ class DeclareStrictTypesRule implements Rule
 				continue;
 			}
 
-			$validValues = new UnionType([new ConstantIntegerType(0), new ConstantIntegerType(1)]);
-			if (!$validValues->isSuperTypeOf($scope->getType($declare->value))->yes()) {
+			if (
+				!$declare->value instanceof Node\Scalar\LNumber
+				|| !in_array($declare->value->value, [0, 1], true)
+			) {
 				return [
 					RuleErrorBuilder::message(sprintf(
 						'Declare strict_types must have 0 or 1 as its value.',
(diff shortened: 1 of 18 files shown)


mad-briller:
does this ignore `declare` used for [purposes other than](https://phpstan.org/r/afae2682-8ec3-4f74-b555-92daf62bad46) `strict_types`?

staabm:
good catch - fixed thanks

ondrejmirtes:
Please name this accordingly "DeclareStrictTypesRule" and also check nonsense in the directive, like `'foo'` instead of 0/1.

src/Rules/Keywords/DeclareStrictTypesRule.php
```diff
@@ -0,0 +1,69 @@
...
+	{
+		$declaresStrictTypes = false;
+		foreach ($node->declares as $declare) {
+			if (
+				$declare->key->name !== 'strict_types'
+			) {
+				continue;
+			}
+
+			$validValues = new UnionType([new ConstantIntegerType(0), new ConstantIntegerType(1)]);
```
ondrejmirtes: Please don't use the type system for this, just the AST. I'm pretty sure some stuff is not allowed that would slip through this.

ondrejmirtes (changes requested):

src/Rules/Keywords/DeclareStrictTypesRule.php
```diff
@@ -0,0 +1,70 @@
...
+				continue;
+			}
+
+			if (
+				!$declare->value instanceof Node\Scalar\LNumber
+				|| !in_array($declare->value->value, [0, 1], true)
+			) {
+				return [
+					RuleErrorBuilder::message(sprintf(
+						'Declare strict_types must have 0 or 1 as its value.',
```
ondrejmirtes: This message could be improved with `'Declare strict_types must have 0 or 1 as its value, %s given.'`

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

ondrejmirtes:
Hey @staabm this rule now causes a false positive: https://phpstan.org/r/adec71c6-2f0f-41b0-afc2-5df3f9110bce

Please look into it as it's currently blocking a release. Thank you.

staabm:
I am on vacation and can have a look on sunday

PR #2760, opened 2023-11-21
VarTagTypeRuleHelper: fix widening list<array{id: int}> to mixed[]

janedbal:
Missing case of https://github.com/phpstan/phpstan/issues/10130


ondrejmirtes:
Thank you. I'll see what the real-world impact is.

PR #2761, opened 2023-11-21
DynamicReturnTypeExtensions: allow hooking on object

janedbal:
To support stuff like https://github.com/phpstan/phpstan-doctrine/issues/499

### DIFF
--- a/tests/PHPStan/Analyser/data/TestDynamicReturnTypeExtensions.php
+++ b/tests/PHPStan/Analyser/data/TestDynamicReturnTypeExtensions.php
@@ -268,9 +268,9 @@ class Bug7391BDynamicStaticMethodReturnTypeExtension implements DynamicStaticMet
 
 class ObjectHookedDynamicStaticMethodReturnTypeExtension implements DynamicStaticMethodReturnTypeExtension {
 
-	public function getClass(): string
+	public function getClass(): ?string
 	{
-		return 'object';
+		return null;
 	}
 
 	public function isStaticMethodSupported(MethodReflection $methodReflection): bool
--- a/tests/PHPStan/Analyser/data/TestDynamicReturnTypeExtensions.php
+++ b/tests/PHPStan/Analyser/data/TestDynamicReturnTypeExtensions.php
@@ -13,6 +13,7 @@ use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ResolvedMethodReflection;
 use PHPStan\Reflection\Type\CalledOnTypeUnresolvedMethodPrototypeReflection;
 use PHPStan\Reflection\Type\UnionTypeMethodReflection;
+use PHPStan\Type\BooleanType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/data/TestDynamicReturnTypeExtensions.php
```diff
@@ -263,3 +264,23 @@ public function getTypeFromStaticMethodCall(
 		return $scope->getType(new New_($methodCall->class));
 	}
 }
+
+
+class ObjectHookedDynamicStaticMethodReturnTypeExtension implements DynamicStaticMethodReturnTypeExtension {
+
+	public function getClass(): string
+	{
+		return 'object';
```
ondrejmirtes: Magic strings like this are a no-go :) We need a new specific extension type. I'd imagine the interface can look the same, without the `getClass` method.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/TestDynamicReturnTypeExtensions.php
janedbal: Ok, how do we name it?

Or how about returning null here?

janedbal: ~~Heh, `object::class` works, so it is not a magic string anymore :D  https://3v4l.org/rrvKH~~



nvm, that is nonsense

janedbal: Personally, I'd prefer `null`

janedbal: Implemented null version.

ondrejmirtes: The name isn't that important 😊 It can be "ClasslessDynamicMethodReturnTypeExtension".

ondrejmirtes:
I've got a completely different and much more universal idea how to solve this, stay tuned 😊

tests/PHPStan/Analyser/data/TestDynamicReturnTypeExtensions.php
janedbal: Classless is a good prefix! But I assume I should wait for [this](https://github.com/phpstan/phpstan-src/pull/2761#issuecomment-1823617901)

ondrejmirtes:
So I worry that we might need to add many more extensions to be able to describe universally all these situations.

What I propose is a new `ExpressionTypeResolverExtension` with method:

`getType(Expe $expe): ?Type`

It would be called right at the beginning of MutatingScope::resolveType.

Returning null means "I'm not interested in this expression".

This mechanism would also be able to replace not-really-flexible overriding of scopeClass, because we can only have one for the whole analysis.

Feel free to prototype it 😊

janedbal:
Added extension as requested, but I think it needs to be `getType(Expr $expr, Scope $scope): ?Type`, see https://github.com/phpstan/phpstan-src/pull/2789

janedbal:
Closing in favor of https://github.com/phpstan/phpstan-src/pull/2789

PR #2763, opened 2023-11-22
Introduce InvalidTypesInUnionRule

paulbalandan:
Closes https://github.com/phpstan/phpstan/issues/9185

### DIFF
--- a/tests/PHPStan/Rules/Types/data/invalid-union-with-mixed.php
+++ b/tests/PHPStan/Rules/Types/data/invalid-union-with-mixed.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types=1);
+
+namespace InvalidUnionWithMixed;
+
+class Foo
+{
+
+    public ?string $bar = null;
+    public ?mixed $baz = null;
+
+    public int|null $lorem = null;
+    public mixed|string $ipsum = '';
+
+    public function dolor(
+        string|int $sit,
+        bool|mixed $amet
+    ): mixed|string
+    {
+        return '';
+    }
+
+    public function lorem(): mixed|string
+    {
+        return '';
+    }
+
+}
+
+function funcWithMixed(mixed $a, ?mixed $b): ?mixed
+{
+    return $a;
+}
+
+static fn (int $a): ?mixed => $a;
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
I love it, thank you very much!

herndlm:
Nice, thx @paulbalandan !

tests/PHPStan/Rules/Types/data/invalid-union-with-mixed.php
```diff
@@ -0,0 +1,34 @@
+<?php declare(strict_types=1);
+
+namespace InvalidUnionWithMixed;
+
+class Foo
```
staabm, line 5: what about applying the same rules for phpdoc types?

https://phpstan.org/r/9c3d8fed-52b9-4285-b877-c34ec751ae4c

ondrejmirtes: We'd have to give it some more thought, thanks to type normalization `mixed|null` is totally fine in PHPDocs, I don't want to mark that as an error.

herndlm: a special rule only for phpdocs invalid types maybe, maybe hidden behind bleeding edge? hmm

ondrejmirtes: Invalid types in PHPDocs like `string&int` are already reported. If you think there is some scenario that isn't covered, open a new feature requests, thanks.

herndlm: As mentioned before, we have https://github.com/phpstan/phpstan/issues/6720 already, so it might be a special thing for void mostly :)

ondrejmirtes: Void in return types should be transformed into null after a call.

herndlm: would you say only for phpdoc or in general? currently thinking were the best place is to transform types 🤔

ondrejmirtes: In MutatingScope::resolveType where MethodCall/StaticCall/FuncCall are handled.

herndlm: Ok thx, I did exactly that already 😊 looked mostly good with a couple of question marks. Will open a PR when ready.

PR #2764, opened 2023-11-22
Narrow enum_exists() arg to UnitEnum

staabm:
closes https://github.com/phpstan/phpstan/issues/10169

### DIFF
--- a/src/Type/Php/ClassExistsFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/ClassExistsFunctionTypeSpecifyingExtension.php
@@ -55,20 +55,14 @@ class ClassExistsFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyi
 			);
 		}
 
+		$narrowedType = new ClassStringType();
 		if ($functionReflection->getName() === 'enum_exists') {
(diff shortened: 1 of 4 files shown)


src/Type/Php/ClassExistsFunctionTypeSpecifyingExtension.php
```diff
@@ -54,6 +55,17 @@ public function specifyTypes(FunctionReflection $functionReflection, FuncCall $n
 			);
 		}
 
+		if ($functionReflection->getName() === 'enum_exists') {
+			return $this->typeSpecifier->create(
```
ondrejmirtes: Please less code duplication here. The only thing different is the type argument.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

ZejdCicak:
This makes `enum_exists` always treat an enum as UnitEnum, and never as BackedEnum, do you think there is a way to deal with this in library?

ondrejmirtes:
@ZejdCicak Please open a new issue with a reproducing snippet of your problem.

PR #2765, opened 2023-11-22
Fix #10176 - Optional 4th parameter for Redis::restore

Vaalyn:
This PR fixes the 4th parameter of the Redis::restore method to be optional.

See issue [#10176](https://github.com/phpstan/phpstan/issues/10176)


ondrejmirtes:
Thank you!

PR #2767, opened 2023-11-22
Fix for bug #9654: Unsafe use of new static when trait enforces constructor via abstract method

maurice2k:
Fixes https://github.com/phpstan/phpstan/issues/9654

### DIFF
--- a/src/Rules/Classes/NewStaticRule.php
+++ b/src/Rules/Classes/NewStaticRule.php
@@ -74,6 +74,19 @@ class NewStaticRule implements Rule
 			}
 		}
 
+		if ($scope->isInTrait()) {
+			$traitReflection = $scope->getTraitReflection();
+			if ($traitReflection->hasConstructor()) {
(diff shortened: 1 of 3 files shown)


src/Rules/Classes/NewStaticRule.php
```diff
@@ -74,6 +74,19 @@ public function processNode(Node $node, Scope $scope): array
 			}
 		}
 
+		if ($scope->isInTrait()) {
+			$traitReflection = $scope->getTraitReflection();
+			if ($traitReflection->hasConstructor()) {
+				$traitConstructor = $traitReflection->getConstructor();
+
+				if ($traitConstructor instanceof PhpMethodReflection) {
```
ondrejmirtes, line 82: This still needs to error before PHP 8.0: https://3v4l.org/cOUYD

ondrejmirtes (changes requested):

PR #2769, opened 2023-11-23
Classname in instanceof cannot be null

staabm:
closes https://github.com/phpstan/phpstan/issues/10036

### DIFF
--- a/src/Rules/Classes/ImpossibleInstanceOfRule.php
+++ b/src/Rules/Classes/ImpossibleInstanceOfRule.php
@@ -36,6 +36,18 @@ class ImpossibleInstanceOfRule implements Rule
 
 	public function processNode(Node $node, Scope $scope): array
 	{
+		if (!$node->class instanceof Node\Name) {
+			$classType = $this->treatPhpDocTypesAsCertain ? $scope->getType($node->class) : $scope->getNativeType($node->class);
+			if (TypeCombinator::containsNull($classType)) {
+				return [
+					RuleErrorBuilder::message(sprintf(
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Rules/Classes/data/bug-10036.php
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	public function sayString(string $class): void
+	{
+		var_dump(new stdClass instanceof $class);
+	}
+
+	public function sayMixed(mixed $class): void
+	{
+		var_dump(new stdClass instanceof $class);
```
staabm, line 27: was not sure whether we want the rule error on mixed

src/Rules/Classes/ImpossibleInstanceOfRule.php
```diff
@@ -36,6 +36,18 @@ public function getNodeType(): string
 
 	public function processNode(Node $node, Scope $scope): array
 	{
+		if (!$node->class instanceof Node\Name) {
```
staabm: ~~maybe this needs to be BleedingEdge only?~~ reworked into a level 8 rule instead

ondrejmirtes:
There are many other invalid things that can't be in `instanceof`: https://3v4l.org/6h0tt

This can't just be a level 8 rule. Stuff like this needs to be reported sooner: https://phpstan.org/r/caf262f8-6927-4288-8ce0-09da00a1dd1d

`mixed` should be reported on level 9.

To help with that you need to use RuleLevelHelper.

PR #2770, opened 2023-11-23
Replace with in_array() instead of redundant property

zonuexe:
Since this extension does not require any properties, this change should reduce memory usage by a very very very small amount.


zonuexe:
In my opinion, constants are better than properties, so I'm considering removing properties in other places as well, but are there any chances that these changes will be accepted? https://github.com/zonuexe/phpstan-src/commit/642244902f0747e6c8c7cbac261fc26cbd1407af

I'm torn between arguing that PHPStan's processes are likely to run frequently, so it makes sense to make micro optimizations, and that these changes have no performance impact, so they should not be changed unnecessarily.

ondrejmirtes:
Yeah I'll accept this but not because of performance (which is not going to be mesurable) but because it makes sense code-wise and design-wise.

ondrejmirtes:
Thank you!

PR #2771, opened 2023-11-23
Detect Closure and WeakReference/WeakMap calls when unused

mvorisek:
fix https://github.com/phpstan/phpstan/issues/10181

PR to phpstorm-stubs: https://github.com/JetBrains/phpstorm-stubs/pull/1589

repro https://phpstan.org/r/f006bb3f-fe74-4d64-a0ba-946bf6eade04 - let me know where to put this test if wanted


staabm:
the test needs to go into the rule-class-test which emits the error

ondrejmirtes:
If the phpstorm-stubs PR gets merged then we don't need to do anything here.

mvorisek:
I expect the phpstorm stub merged

mvorisek:
@ondrejmirtes the phpstom stub PR was merged, can this https://github.com/phpstan/phpstan-src/blob/1.10.x/composer.json#L17 dep be updated?

PR #2772, opened 2023-11-24
Use dedicated Type methods over isSuperTypeOf()

staabm:
while reviewing https://github.com/phpstan/phpstan-src/pull/2710 I noticed that we can simplify stuff by using dedicated Type methods

### DIFF
--- a/src/Rules/IssetCheck.php
+++ b/src/Rules/IssetCheck.php
@@ -8,6 +8,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Rules\Properties\PropertyDescriptor;
 use PHPStan\Rules\Properties\PropertyReflectionFinder;
 use PHPStan\Type\MixedType;
+use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
 use function is_string;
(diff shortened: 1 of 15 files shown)


tests/PHPStan/Rules/Variables/IssetRuleTest.php
```diff
@@ -281,7 +281,7 @@ public function testVariableCertaintyInIsset(): void
 				116,
 			],
 			[
-				'Variable $variableInSecondCase in isset() always exists and is always null.',
+				'Variable $variableInSecondCase in isset() always exists and is not nullable.',
```
staabm: this one is actually interessting, as I think it actually fixes a bug - which is suprising.

https://github.com/phpstan/phpstan-src/blob/3a5328b15858b1e484b011bfc61075655cc3574b/tests/PHPStan/Rules/Variables/data/variable-certainty-isset.php#L117

staabm:
the doctrine error is valid: upstream fix in https://github.com/doctrine/orm/pull/11085

tests/PHPStan/Rules/Variables/IssetRuleTest.php
staabm: having another look at the test-case I think our assertions could be more precise:

https://github.com/phpstan/phpstan-src/blob/3a5328b15858b1e484b011bfc61075655cc3574b/tests/PHPStan/Rules/Variables/data/variable-certainty-isset.php#L108-L124

in case 1 and 2 the case expression is always true and we are only assigning values. as the switch-case is constructed we will always run into `break` on line 114, which effectively means the cases below line 114 are dead code.

staabm: found the reason for the change in behaviour:

$neverType->isNull() -> no
$nullType->isSuperTypeOf($neverType) -> yes

src/Rules/IssetCheck.php
```diff
@@ -44,11 +45,14 @@ public function check(Expr $expr, Scope $scope, string $operatorDescription, cal
...
-						$typeMessageCallback,
-					);
+					$type = $this->treatPhpDocTypesAsCertain ? $scope->getType($expr) : $scope->getNativeType($expr);
+					if (!$type instanceof NeverType) {
+						return $this->generateError(
+							$type,
+							sprintf('Variable $%s %s always exists and', $expr->name, $operatorDescription),
+							$typeMessageCallback,
+						);
+					}
```
staabm, line 55: this changes address the comments in https://github.com/phpstan/phpstan-src/pull/2772#discussion_r1404243129

everything else in this PR was a mechnical change from
`((new ConstantBooleanType(false))->isSuperTypeOf($type)->yes())` to `$type->isFalse()->yes()` etc.

staabm:
the drupal error is caused by invalid parameter types: https://github.com/phpstan/phpstan/issues/10189

therefore I think this is good to go.

ondrejmirtes:
Thank you!

PR #2774, opened 2023-11-24
Reject extra keys in ConstantArrayType::acceptsWithReason

schlndh:
This PR attempts to make acceptsWithReason reject extra keys in ConstantArrayType. I discovered the issue with the options array in `setcookie`:

https://phpstan.org/r/a55864e0-64be-431a-b40a-64a01c6d708c
https://3v4l.org/v6AL2

But even in functions which ignore the extra keys it seems generally undesirable to allow them. E.g. a programmer could pass a key thinking that it will do something, but then it wouldn't. Though I guess in some cases it could be a bit annoying/controversial.

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -309,19 +309,18 @@ class ConstantArrayType extends ArrayType implements ConstantType
 		$result = AcceptsResult::createYes();
 		foreach ($this->keyTypes as $i => $keyType) {
 			$valueType = $this->valueTypes[$i];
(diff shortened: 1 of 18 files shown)


ondrejmirtes:
Hi, this is a huge topic and cannot be done in a minor/patch version like this. See: https://github.com/phpstan/phpstan/issues/8438

We also have to implement some syntax for unsealed array specifically, and also implement "array shape with known type of extra keys" to solve https://github.com/phpstan/phpstan/issues/4703 (it's not gonna be an "intersection" of arrays but that's how the issue is called now).

All of this is connected together.

schlndh:
Thank you for the explanation. This is much more complicated than I realized.

PR #2775, opened 2023-11-24
add filename and line number to phpunit output

oadam:


ondrejmirtes:
Please link some proof this is part of the JUnit standard.

oadam:
I amended the commit to try and justify the inclusion. Seems like there is no agreed upon junit standard :cry: !

no xsd here :
https://github.com/sebastianbergmann/phpunit/blob/main/src/Logging/JUnit/JunitXmlLogger.php#L298

discussion of the absence of proper junit xsd here :
https://github.com/sebastianbergmann/phpunit/issues/2964

Doing the same as phpunit should be safe ?

oadam:
Some of the attributes I have added are added here in PHPUnit : https://github.com/sebastianbergmann/phpunit/blob/main/src/Logging/JUnit/JunitXmlLogger.php#L428

src/Command/ErrorFormatter/JunitErrorFormatter.php
```diff
@@ -27,7 +27,7 @@ public function formatErrors(
 
 		$result = '<?xml version="1.0" encoding="UTF-8"?>';
 		$result .= sprintf(
-			'<testsuite failures="%d" name="phpstan" tests="%d" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/junit-team/junit5/r5.5.1/platform-tests/src/test/resources/jenkins-junit.xsd">',
```
ondrejmirtes, line 30: Why do you remove this? Maybe it's important

tests/PHPStan/Command/ErrorFormatter/junit-schema.xsd
```diff
@@ -1,118 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" ?>
```
ondrejmirtes, line 1: Maybe you can find a newer version of this schema that includes file and line?

src/Command/ErrorFormatter/JunitErrorFormatter.php
```diff
@@ -65,9 +67,16 @@ public function formatErrors(
 	 *
 	 *
 	 */
-	private function createTestCase(string $reference, string $type, ?string $message = null): string
+	private function createTestCase(string $reference, string $type, ?string $message = null, ?string $fileName = null, ?int $lineNumber = null): string
 	{
-		$result = sprintf('<testcase name="%s">', $this->escape($reference));
+		$result = sprintf('<testcase name="%s"', $this->escape($reference));
+		if ($fileName !== null) {
+			$result .= sprintf(' file="%s"', $this->escape($fileName));
```
ondrejmirtes, line 74: What about absolute vs. relative path to the file? Typically you don't want to print the whole absolute path, but trim the beginning. But I'm not sure what PHPUnit does and what tools that read the output expect.

Anyway, this is usually done with the help of `RelativePathHelper`. Check the other formatters.

ondrejmirtes (changes requested):

oadam:
@ondrejmirtes I'm removing any mention of a XSD because there is no official schema.

PHPUnit does not have it in its output as well : https://github.com/sebastianbergmann/phpunit/blob/main/tests/end-to-end/logging/log-junit-to-file.phpt#L26

If I add the `RelativePathHelper`, would you be open to merge this even with the XSD removal ?

My objective is to add file and linenumber so that, in my private repo, I can better annotate failed pull requests with https://github.com/marketplace/actions/junit-report-action (it annotates PR with [annotations on offending lines](https://github.com/mikepenz/action-junit-report/blob/main/.github/images/annotated.png))

oadam:
I just realized that you have a github output format ! You can disregard this PR is removing the XSD seems too much

ondrejmirtes:
Yeah, PHPStan provides nice experience out of the box :) Thanks anyway.

PR #2776, opened 2023-11-25
Impl StrIncrementDecrementFunctionReturnTypeExtension

zonuexe:
Add types to [`str_increment()`](https://php.net/str_increment) and [`str_decrement()`](https://php.net/str_decrement) added in PHP 8.3.

However, unlike something like `strtolower()`, I can't think of many use cases for incrementing/decrementing constants, so it makes sense to reject this PR.

### DIFF
--- a/src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
+++ b/src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
@@ -1,0 +1,136 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\ErrorType;
+use PHPStan\Type\Type;
+use ValueError;
+use function chr;
+use function count;
+use function function_exists;
+use function implode;
+use function in_array;
+use function is_float;
+use function is_int;
+use function is_numeric;
+use function ord;
+use function preg_match;
+use function str_split;
+use function stripos;
+
+class StrIncrementDecrementFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+{
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return in_array($functionReflection->getName(), ['str_increment', 'str_decrement'], true);
+	}
+
+	public function getTypeFromFunctionCall(
+		FunctionReflection $functionReflection,
+		FuncCall $functionCall,
+		Scope $scope,
+	): Type
+	{
+		$fnName = $functionReflection->getName();
+		$args = $functionCall->getArgs();
+
+		if (count($args) !== 1) {
+			return ParametersAcceptorSelector::selectSingle($functionReflection->getVariants())->getReturnType();
+		}
+
+		$argType = $scope->getType($args[0]->value);
+		if (count($argType->getConstantScalarValues()) === 0) {
+			return ParametersAcceptorSelector::selectSingle($functionReflection->getVariants())->getReturnType();
+		}
+
+		$value = $argType->getConstantScalarValues()[0];
+		if (!(is_string($value) || is_int($value) || is_float($value))) {
+			return new ErrorType();
+		}
+		$string = (string) $value;
+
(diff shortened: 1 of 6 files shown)


src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,154 @@
...
+					'F' => 'E',
+					'g' => 'f',
+					'G' => 'F',
+				][$s[$offset]];
+
+				return $s;
+			}
+		}
+
+		return (string)++$s;
```
zonuexe: refs https://github.com/phpstan/phpstan/issues/10187

resources/functionMetadata.php
```diff
@@ -1473,8 +1475,10 @@
 	'sqrt' => ['hasSideEffects' => false],
 	'stat' => ['hasSideEffects' => false],
 	'str_contains' => ['hasSideEffects' => false],
+	'str_decrement' => ['hasSideEffects' => false],
```
ondrejmirtes: Please see the top of this file:

https://github.com/phpstan/phpstan-src/blob/1bb4a34455dcca9239e7c6a50c36614daf718204/resources/functionMetadata.php#L3-L8

resources/functionMap_php83delta.php
```diff
@@ -0,0 +1,30 @@
...
+ *   Functions are expected to be removed only in major releases of php.
+ *
+ * @see FunctionSignatureMap.php
+ *
+ * @phan-file-suppress PhanPluginMixedKeyNoKey (read by Phan when analyzing this file)
+ */
+return [
+	'new' => [
+		'str_decrement' => ['non-empty-string'],
+		'str_increment' => ['non-falsy-string'],
```
ondrejmirtes: This signature is wrong. The functions also have parameters.

src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,154 @@
...
+use function count;
+use function function_exists;
+use function implode;
+use function in_array;
+use function is_float;
+use function is_int;
+use function is_numeric;
+use function ord;
+use function preg_match;
+use function str_decrement;
```
ondrejmirtes: You need to ignore the errors in a new file referenced from here (for `< 80300`) https://github.com/phpstan/phpstan-src/blob/1.11.x/build/ignore-by-php-version.neon.php

src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,154 @@
...
+		return $value === null
+			? new ErrorType()
+			: new ConstantStringType($value);
+	}
+
+	private function increment(string $s): ?string
+	{
+		if (function_exists('str_increment')) {
+			try {
+				return str_increment($s);
```
ondrejmirtes: Or maybe we can always call our own implementation and not have a different one here.

ondrejmirtes (changes requested):

src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,136 @@
...
+		FunctionReflection $functionReflection,
+		FuncCall $functionCall,
+		Scope $scope,
+	): Type
+	{
+		$fnName = $functionReflection->getName();
+		$args = $functionCall->getArgs();
+
+		if (count($args) !== 1) {
+			return ParametersAcceptorSelector::selectSingle($functionReflection->getVariants())->getReturnType();
```
ondrejmirtes: You can just return `null` in these cases

src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		$argType = $scope->getType($args[0]->value);
+		if (count($argType->getConstantScalarValues()) === 0) {
+			return ParametersAcceptorSelector::selectSingle($functionReflection->getVariants())->getReturnType();
+		}
+
+		$value = $argType->getConstantScalarValues()[0];
```
ondrejmirtes: You should account for union of values, don't just read the first one

ondrejmirtes (changes requested):

src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php
zonuexe: I see, I misunderstood `UnionType::getConstantScalarValues()`.

ondrejmirtes:
Thank you!

PR #2778, opened 2023-11-26
Transform `void` return to `null` after call

herndlm:
Closes https://github.com/phpstan/phpstan/issues/6720

What I meant with question marks yesterday
- I'm not a 100% sure about this yet, we can see that it turns off many reports and e.g. `UsageOfVoidMatchExpressionRule` seems to be not doing anything anymore I guess
- On the other hand it is how PHP behaves
- I noticed that Generator void returns are not handled

Also thought of: only transform void in union/intersection types, but that does feel inconsistent..

WDYT?

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -678,7 +678,7 @@ class MutatingScope implements Scope
 			$key .= '/*' . $node->getAttribute('startFilePos') . '*/';
 		}
 
-		if ($node->hasAttribute(self::KEEP_VOID_ATTRIBUTE_NAME)) {
+		if ($node->getAttribute(self::KEEP_VOID_ATTRIBUTE_NAME) === true) {
 			$key .= '/*' . self::KEEP_VOID_ATTRIBUTE_NAME . '*/';
 		}
 
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -676,6 +676,10 @@ class MutatingScope implements Scope
 			$key .= '/*' . $node->getAttribute('startFilePos') . '*/';
 		}
 
+		if ($node->hasAttribute('keepVoid')) {
+			$key .= '/*keepVoid*/';
+		}
+
 		return $key;
 	}
 
@@ -1492,6 +1496,9 @@ class MutatingScope implements Scope
 			$matchScope = $this;
 			foreach ($node->arms as $arm) {
 				if ($arm->conds === null) {
+					if ($node->hasAttribute('keepVoid')) {
+						$arm->body->setAttribute('keepVoid', $node->getAttribute('keepVoid'));
+					}
 					$types[] = $matchScope->getType($arm->body);
 					continue;
 				}
@@ -1521,6 +1528,9 @@ class MutatingScope implements Scope
 
 				if (!$filteringExprType->isFalse()->yes()) {
 					$truthyScope = $matchScope->filterByTruthyValue($filteringExpr);
+					if ($node->hasAttribute('keepVoid')) {
+						$arm->body->setAttribute('keepVoid', $node->getAttribute('keepVoid'));
+					}
 					$types[] = $truthyScope->getType($arm->body);
 				}
 
@@ -1946,7 +1956,7 @@ class MutatingScope implements Scope
 				}
 			}
 
-			return $this->transformVoidToNull($parametersAcceptor->getReturnType());
+			return $this->transformVoidToNull($parametersAcceptor->getReturnType(), $node);
 		}
 
 		return new MixedType();
@@ -1986,8 +1996,12 @@ class MutatingScope implements Scope
 		return $type;
 	}
 
-	private function transformVoidToNull(Type $type): Type
+	private function transformVoidToNull(Type $type, Node $node): Type
 	{
+		if ($node->getAttribute('keepVoid') === true) {
+			return $type;
+		}
+
 		return TypeTraverser::map($type, static function (Type $type, callable $traverse): Type {
 			if ($type instanceof UnionType || $type instanceof IntersectionType) {
 				return $traverse($type);
@@ -5118,7 +5132,7 @@ class MutatingScope implements Scope
 			$normalizedMethodCall = ArgumentsNormalizer::reorderStaticCallArguments($parametersAcceptor, $methodCall);
 		}
 		if ($normalizedMethodCall === null) {
-			return $this->transformVoidToNull($parametersAcceptor->getReturnType());
+			return $this->transformVoidToNull($parametersAcceptor->getReturnType(), $methodCall);
 		}
 
 		$resolvedTypes = [];
@@ -5157,10 +5171,10 @@ class MutatingScope implements Scope
 		}
 
 		if (count($resolvedTypes) > 0) {
-			return $this->transformVoidToNull(TypeCombinator::union(...$resolvedTypes));
+			return $this->transformVoidToNull(TypeCombinator::union(...$resolvedTypes), $methodCall);
 		}
 
-		return $this->transformVoidToNull($parametersAcceptor->getReturnType());
+		return $this->transformVoidToNull($parametersAcceptor->getReturnType(), $methodCall);
 	}
 
 	/** @api */
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -150,6 +150,8 @@ class MutatingScope implements Scope
 
 	private const BOOLEAN_EXPRESSION_MAX_PROCESS_DEPTH = 4;
 
+	private const KEEP_VOID_ATTRIBUTE_NAME = 'keepVoid';
+
 	/** @var Type[] */
 	private array $resolvedTypes = [];
 
@@ -676,8 +678,8 @@ class MutatingScope implements Scope
 			$key .= '/*' . $node->getAttribute('startFilePos') . '*/';
 		}
 
-		if ($node->hasAttribute('keepVoid')) {
-			$key .= '/*keepVoid*/';
+		if ($node->hasAttribute(self::KEEP_VOID_ATTRIBUTE_NAME)) {
+			$key .= '/*' . self::KEEP_VOID_ATTRIBUTE_NAME . '*/';
 		}
 
 		return $key;
@@ -1240,9 +1242,7 @@ class MutatingScope implements Scope
 						new VoidType(),
 					]);
 				} else {
-					$expr = clone $node->expr;
-					$expr->setAttribute('keepVoid', true);
-					$returnType = $arrowScope->getType($expr);
+					$returnType = $arrowScope->getKeepVoidType($node->expr);
 					if ($node->returnType !== null) {
 						$returnType = TypehintHelper::decideType($this->getFunctionType($node->returnType, false, false), $returnType);
 					}
@@ -1498,8 +1498,8 @@ class MutatingScope implements Scope
 			$matchScope = $this;
 			foreach ($node->arms as $arm) {
 				if ($arm->conds === null) {
-					if ($node->hasAttribute('keepVoid')) {
-						$arm->body->setAttribute('keepVoid', $node->getAttribute('keepVoid'));
+					if ($node->hasAttribute(self::KEEP_VOID_ATTRIBUTE_NAME)) {
+						$arm->body->setAttribute(self::KEEP_VOID_ATTRIBUTE_NAME, $node->getAttribute(self::KEEP_VOID_ATTRIBUTE_NAME));
 					}
 					$types[] = $matchScope->getType($arm->body);
 					continue;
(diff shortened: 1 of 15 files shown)


tests/PHPStan/Analyser/data/reflection-type.php
```diff
@@ -9,7 +9,7 @@ function test(
 	\ReflectionFunctionAbstract $reflectionFunctionAbstract,
 	\ReflectionParameter $reflectionParameter
 ){
-	assertType('ReflectionType|null', $reflectionProperty->getType());
+	assertType('ReflectionNamedType|ReflectionUnionType|null', $reflectionProperty->getType());
```
herndlm: looks like an unwanted behaviour change? I'm not sure what it means.

herndlm:
```suggestion
	assertType('ReflectionType|nulll', $reflectionProperty->getType());
```

tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php
```diff
@@ -104,70 +104,6 @@ public function testCallMethods(): void
 				63,
 				'Learn more at https://phpstan.org/user-guide/discovering-symbols',
 			],
-			[
```
ondrejmirtes: Do not remove tests from rules affected by this change, instead fix the rules so that these problems are still reported ;)

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php
herndlm: But isn't this the expected behaviour change?

ondrejmirtes: No, you still want to know that you're calling a `void` function and you are not supposed to read the result.

herndlm: OK, that makes more sense to me and kind of answers my open questions. Will need to think about the implementation again then though :)

tests/PHPStan/Analyser/data/reflection-type.php
```diff
@@ -9,7 +9,7 @@ function test(
 	\ReflectionFunctionAbstract $reflectionFunctionAbstract,
 	\ReflectionParameter $reflectionParameter
 ){
-	assertType('ReflectionType|null', $reflectionProperty->getType());
+	assertType('ReflectionType|nulll', $reflectionProperty->getType());
```
herndlm:
```suggestion
	assertType('ReflectionType|null', $reflectionProperty->getType());
```

ondrejmirtes:
I think the problem with existing rules might not be that hard to fix.

Most of the failures are from the error that's supposed to be reported here: https://github.com/phpstan/phpstan-src/blob/3c7860c8d8bcca5de1d36322dc52e9f6a2f00387/src/Rules/FunctionCallParametersCheck.php#L233-L241

We could for example clone the node and add an attribute like `'keepVoid' => true` that will not transform void into null while resolving the type in MutatingScope.

herndlm:
> I think the problem with existing rules might not be that hard to fix.
> 
> Most of the failures are from the error that's supposed to be reported here:
> 
> https://github.com/phpstan/phpstan-src/blob/3c7860c8d8bcca5de1d36322dc52e9f6a2f00387/src/Rules/FunctionCallParametersCheck.php#L233-L241
> 
> We could for example clone the node and add an attribute like `'keepVoid' => true` that will not transform void into null while resolving the type in MutatingScope.

I had a similar thought. Basically, we'd need to either a) keep the information somewhere that this was a void before, either e.g. in the type or the node as you said or b) modify the void type to behave more like null, since it's a special return-only type it always means that it must have came from a return. I guess the "remember in node" approach is the cleanest and one that makes most sense :) I also quickly checked psalm yesterday and looks like it's transforming void to null too basically.

tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php
mvorisek: Yes, return type of `void` should be kept reported but converted to `null`. See https://github.com/phpstan/phpstan/issues/9388, when `void` is unioned with another type, it might probably be not reported, at least not for level <=6.

herndlm:
hmm, this keeps being tricky, I tried to clone the node in rules and add an attribute as you suggested. problem: there were cases where the type was already resolved and it would get it from the internal cache via key again. this could also be circumvented I guess? it's a bit unfortunate hmm.
I'd like to have a wasVoid() method on the type that tells me this, but feels like way too much overhead just for this.. :)

ondrejmirtes:
@herndlm Change the cache key based on the attribute in `MutatingScope::getNodeKey()`.

src/Analyser/MutatingScope.php
```diff
@@ -674,6 +674,10 @@ private function getNodeKey(Expr $node): string
 			$key .= '/*' . $node->getAttribute('startFilePos') . '*/';
 		}
 
+		if ($node->hasAttribute('keepVoid')) {
```
staabm: would be great if we could put this string into a class-constant somewhere.

that way it would ease navigation with the IDE and make dependent parts obvious

herndlm: yeah, I wanted to check first if all is fine and then ask about how to properly refactor this thing ☺️ I hate that it is sprinkled x times around the code base now..

herndlm:
got it working, this is looking not bad, did only quickly go through the failures, the ones that worry me
- `Method call return value that should be used, but is not` from simplify/simplify, no clue what that is yet.
- the `Parameter #2 .. of class ..  expects Closure(): void, Closure(): null given. ` of PMMP, could be a sign that more adaptions are needed hmm

ondrejmirtes:
> Closure(): void, Closure(): null given.

Stop early in the callback:

```php
if ($type instanceof CallableType || $type instanceof ClosureType) {
    return $type;
}
```

About simplify - don't worry about it, they'll have to fix their code.

staabm:
symplify/symplify is abandoned. parts of it [moved to symplify/phpstan-rules](https://packagist.org/packages/symplify/phpstan-rules)

ondrejmirtes:
I've update the tests in phpstan/phpstan to [symplify/phpstan-rules](https://packagist.org/packages/symplify/phpstan-rules), you can try again :)

herndlm:
sorry, still trying to recreate one of the failures of PMMP to properly test the fix..
currently e.g.
```php
class Bar
{

	/** @phpstan-param \Closure(): void $handler */
	public function doFoo(\Closure $handler): void
	{
		\PHPStan\dumpType($handler()); // void
		\PHPStan\dumpType($this->handler()); // null
	}

	public function doBar(): void
	{
		$handler = $this->handler(...);
		\PHPStan\dumpType($handler); // Closure(): void
		$this->doFoo($handler);
	}

	private function handler(): void
	{
	}

}
```

is working just fine in an AnalyserIntegrationTest which I don't understand. I tried to replicate the CI failure I saw via https://github.com/pmmp/PocketMine-MP/blob/0984aa670d5c04e4714853dbf738f51bfa4f39e1/src/network/mcpe/handler/SessionStartPacketHandler.php#L39 and https://github.com/pmmp/PocketMine-MP/blob/0984aa670d5c04e4714853dbf738f51bfa4f39e1/src/network/mcpe/NetworkSession.php#L201

but I can of course just try to adapt the traverse callback 😅

ondrejmirtes:
This passes on 1.10.x but fails on your PR: https://phpstan.org/r/f8d2decc-2d58-4ce4-a2b3-d811700abe9f

ondrejmirtes:
I'd say we need to do the same `keepVoid` thing for inferring the return type of arrow function in MutatingScope around here: https://github.com/phpstan/phpstan-src/blob/7e7cc9cb3daebc5450daf76ac5bd4e848532630b/src/Analyser/MutatingScope.php#L1230-L1233

herndlm:
bingo! thx again :)

what do you think about the previous code example I pasted, it looks like those closure returns are not transformed to null yet. I guess that could be fixed?

ondrejmirtes:
> it looks like those closure returns are not transformed to null

That's okay and actually correct behaviour. We don't want to transform `void` to `null` in a return type there.

Your TypeTransformer callback only transforms standalone `void` + `void` in union and intersection 👍

herndlm:
sorry for the many force pushes, should be good now. do you not think we should refactor it somehow? e.g. put the `'keepVoid'` in some constant, or maybe even have some static helper methods to add / preserve it or so.

src/Analyser/MutatingScope.php
ondrejmirtes: Could be a new method on Scope, just don't start it with `getType`, I'd hate to confuse autocompletion with this niche use case

ondrejmirtes:
Please see https://github.com/phpstan/phpstan-src/pull/2778#discussion_r1410754985

src/Analyser/MutatingScope.php
herndlm: still need to do this refactor, missed the comment. currently working on it

herndlm: looks like I'm prone to commenting race conditions today.. :) did something, not sure about the name of course. this is now in Scope after all and should be stable. WDYT?

src/Analyser/MutatingScope.php
```diff
@@ -674,6 +676,10 @@ private function getNodeKey(Expr $node): string
 			$key .= '/*' . $node->getAttribute('startFilePos') . '*/';
 		}
 
+		if ($node->hasAttribute(self::KEEP_VOID_ATTRIBUTE_NAME)) {
+			$key .= '/*' . self::KEEP_VOID_ATTRIBUTE_NAME . '*/';
```
ondrejmirtes, line 682: This should also check the value of the attribute and do it only if it's true.

src/Analyser/MutatingScope.php
```diff
@@ -1977,6 +1989,25 @@ private function getNullsafeShortCircuitingType(Expr $expr, Type $type): Type
 		return $type;
 	}
 
+	private function transformVoidToNull(Type $type, Node $node): Type
+	{
+		if ($node->getAttribute(self::KEEP_VOID_ATTRIBUTE_NAME) === true) {
```
ondrejmirtes, line 2003: I don't get this. So if it's true then it doesn't transform the type? It should be the opposite, right?

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
herndlm: No, that should be correct. It is named "keepVoid" -> don't transform it to null if true

ondrejmirtes: Oh alright

src/Analyser/MutatingScope.php
```diff
@@ -674,6 +676,10 @@ private function getNodeKey(Expr $node): string
 			$key .= '/*' . $node->getAttribute('startFilePos') . '*/';
 		}
 
+		if ($node->hasAttribute(self::KEEP_VOID_ATTRIBUTE_NAME)) {
```
herndlm:
```suggestion
		if ($node->getAttribute(self::KEEP_VOID_ATTRIBUTE_NAME) === true) {
```
Let's see if I can do some phone fixing again.. :)

src/Analyser/MutatingScope.php
herndlm: Good catch, thx

staabm:
@herndlm this PR here will render the PR at https://github.com/phpstan/phpstan-src/pull/2494 obsolete, right?

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
```diff
@@ -1584,6 +1584,11 @@ public function testBug10171(): void
 		]);
 	}
 
+	public function testBug6720(): void
```
mvorisek, line 1587: please add https://github.com/phpstan/phpstan/issues/9388 test too

ondrejmirtes:
Thank you!

ondrejmirtes:
@herndlm Can you please look at this bug? https://github.com/WordPress/WordPress-Coding-Standards/pull/2416/files

We should allow returning `null` from `void` method when it's typed in a PHPDoc. Here's the code: https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WhiteSpace/ObjectOperatorSpacingSniff.php

herndlm:
I dislike those WP void unions.. but sure. https://phpstan.org/r/12a90bd8-64db-4620-855c-23579d0085f6 seems to be a simple reproducer too.

src/Analyser/MutatingScope.php
```diff
@@ -1236,7 +1242,7 @@ private function resolveType(string $exprString, Expr $node): Type
 						new VoidType(),
 					]);
 				} else {
-					$returnType = $arrowScope->getType($node->expr);
+					$returnType = $arrowScope->getKeepVoidType($node->expr);
```
staabm, line 1245: does it make sense to handle a "void" return type of a arrow function? can arrow functions have a void return type?

ondrejmirtes: Try to change it and see what fails. I think it's valid, at least for a throw expr or calling never function.

staabm: just found https://github.com/phpstan/phpstan-src/pull/2778#issuecomment-1833752355 which answers this question, I guess

staabm: hmm locally I did not get a test failure after changing it. lets see https://github.com/phpstan/phpstan-src/pull/2873

PR #2779, opened 2023-11-26
Ignore shebang in DeclareStrictTypesRule

staabm:
refs https://github.com/phpstan/phpstan-src/pull/2759#issuecomment-1826276219

### DIFF
--- a/src/Parser/DeclarePositionVisitor.php
+++ b/src/Parser/DeclarePositionVisitor.php
@@ -4,6 +4,7 @@ namespace PHPStan\Parser;
 
 use PhpParser\Node;
 use PhpParser\NodeVisitorAbstract;
+use function str_starts_with;
 
 class DeclarePositionVisitor extends NodeVisitorAbstract
 {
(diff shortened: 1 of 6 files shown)


src/Parser/DeclarePositionVisitor.php
```diff
@@ -20,6 +21,15 @@ public function beforeTraverse(array $nodes): ?array
 
 	public function enterNode(Node $node): ?Node
 	{
+		// ignore shebang
+		if (
+			$this->isFirstStatement
+			&& $node instanceof Node\Stmt\InlineHTML
+			&& str_starts_with($node->value, '#!')
```
ondrejmirtes, line 28: Please add a test that tests `str_starts_with($node->value, '#!')`. So InlineHTML with different contents should still throw an error.

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #2782, opened 2023-11-26
Make extensions that return default types simply `return null;`

zonuexe:
Prevent copy-pasting by referencing existing extensions by simply replacing it with `return null;`.

refs https://github.com/phpstan/phpstan-src/pull/2776#discussion_r1405419842


ondrejmirtes:
Thank you.

PR #2787, opened 2023-11-27
Use dedicated Type methods over `isSuperTypeOf()`

staabm:
fixes https://github.com/phpstan/phpstan/issues/9778
fixes https://github.com/phpstan/phpstan/issues/9723
fixes https://github.com/phpstan/phpstan/issues/6407


staabm:
the issue bot summary is interessting. I need to check whether these are actually fixed by this PR or whether these were fixed in one of the other commits.. I did not plan to fix these 😎

ondrejmirtes:
Thank you.

herndlm:
The type refactorings that unintentionally fixes issues are the best 😊

PR #2788, opened 2023-11-27
Use dedicated Type methods over `isSuperTypeOf()`

staabm:
I think this will be the last one of this kind. I worked thru all search patterns I could think of ;-)


ondrejmirtes:
Thank you.

PR #2789, opened 2023-11-27
ExpressionTypeResolverExtension

janedbal:
See https://github.com/phpstan/phpstan-src/pull/2761#issuecomment-1824576419

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -86,6 +86,7 @@ use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\ConstantTypeHelper;
 use PHPStan\Type\DynamicReturnTypeExtensionRegistry;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\ExpressionTypeResolverExtensionRegistry;
 use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\Generic\GenericObjectType;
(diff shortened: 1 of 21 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -678,6 +680,13 @@ private function getNodeKey(Expr $node): string
 
 	private function resolveType(string $exprString, Expr $node): Type
 	{
+		foreach ($this->expressionTypeResolverExtensionRegistry->getExtensions() as $extension) {
+			$type = $extension->getType($node);
```
janedbal: I think I need to pass `$this` here, because for [my usecase](https://github.com/phpstan/phpstan-doctrine/issues/499) I need to be able to detect the return type of the method call.

The [original solution](https://github.com/phpstan/phpstan-doctrine/pull/500/files#diff-8ee6d9610948c4e8345cb982486a7bc2a6546b7d2451966fd9df79e07042ae9eR37) used MethodReflection, but that is not possible here as the caller is unknown.

WDYT @ondrejmirtes ?

ondrejmirtes: Yeah sure, let's add Scope.

ondrejmirtes (changes requested):

ondrejmirtes:
The failures might or might not be some Rector chicanery, don't worry about it.

src/Analyser/MutatingScope.php
janedbal: Hm, that easily leads to infinite recursion :thinking:

janedbal: That being said, I start to like more the old idea.

Or do you have any smart idea how to redesign this without this infinite recursion risk?

ondrejmirtes: Well you shouldn't run into infinite recursion. For the same reason this one https://apiref.phpstan.org/1.11.x/PHPStan.PhpDoc.TypeNodeResolverExtension.html also does not run into infinite recursion, with basically the same design.

You should `return null;` for the nodes you are not interested in.

For method calls you should do `$scope->getMethodReflection($scope->getType($methodCall->var), $methodCall->name)`. And continue from there.

I don't think you will run into infinite recursion that way.

janedbal: Well, ok, that works. But I think most people would just use sth like `$scope->getType($methodCall)`.

So I changed the test to use this approach.

PR #2792, opened 2023-11-28
Fix int-range return type for range()

dantleech:
Fixes the return type of `range` when passed `IntegerRange`

Fixes https://github.com/phpstan/phpstan/issues/10213

### DIFF
--- a/src/Type/Php/RangeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/RangeFunctionReturnTypeExtension.php
@@ -112,7 +112,7 @@ class RangeFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExten
 		$isStepInteger = $stepType->isInteger()->yes();
 
 		if ($isInteger && $isStepInteger) {
-			return AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new IntegerType()));
+			return AccessoryArrayListType::intersectWith(new ArrayType($argType, $argType));
 		}
 
 		if ($argType->isFloat()->yes()) {
--- a/src/Type/Php/RangeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/RangeFunctionReturnTypeExtension.php
@@ -113,7 +113,7 @@ class RangeFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExten
 
 		if ($isInteger && $isStepInteger) {
 			if ($argType instanceof IntegerRangeType) {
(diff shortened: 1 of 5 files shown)


dantleech:
out of curiosity - is there a way to eliminate the delay before the test runs? or to filter a specific test case (the `range-int-range.php`)

ondrejmirtes:
@dantleech When I want to debug something NodeScopeResolverTest, I just run PHPStan on that file. Both TypeInferenceTestCase run through PHPUnit, and PHPStan analysis itself, will report discrepancies in `assertType` etc.

dantleech:
makes sense! thanks

src/Type/Php/RangeFunctionReturnTypeExtension.php
```diff
@@ -112,6 +112,9 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 		$isStepInteger = $stepType->isInteger()->yes();
 
 		if ($isInteger && $isStepInteger) {
+			if ($argType instanceof IntegerRangeType) {
+				return AccessoryArrayListType::intersectWith(new ArrayType($argType, $argType));
+			}
 			return AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new IntegerType()));
```
staabm: can it be simplified to?

```suggestion
			return AccessoryArrayListType::intersectWith(new ArrayType($argType, $argType));
```

dantleech: I did that originally, it won't work with scalar types, as you'll get a RT as `array{1,10}` for `range(1, 10)` (or similar)

dantleech: which also makes me wonder if the `IntegerRange` should be inferred from scalar types instead of `IntegerType` :thinking:

ondrejmirtes (changes requested):
Sorry for a late review.

1) Look at the issue bot: https://github.com/phpstan/phpstan-src/actions/runs/7020506328 This also fixes https://github.com/phpstan/phpstan/issues/9573 so please write a regression test.
2) Please test different ranges passed into the `range` function. For example for `int<5, 10>` and `int<5, 8>`, the resulting range is `int<5, 8>` I think? Please come up with more test cases to make sure the math is correct.

Thank you.

dantleech:
Updated!

ondrejmirtes:
Thank you.

PR #2793, opened 2023-11-28
Fix #10215 - Redis::hMset parameter too strict

Vaalyn:
This PR fixes the 2nd problem reported in the issue [#10215](https://github.com/phpstan/phpstan/issues/10215) where the `fieldvals` parameter for the `Redis::hMset` is too strict since int keys are automatically casted to strings in PhpRedis.


ondrejmirtes:
Thank you.

PR #2796, opened 2023-11-29
Non-numeric strings in pow() lead to error

staabm:
closes https://github.com/phpstan/phpstan/issues/10125

### DIFF
--- a/src/Type/ExponentiateHelper.php
+++ b/src/Type/ExponentiateHelper.php
@@ -87,13 +87,13 @@ final class ExponentiateHelper
 			if ($exponent->getMin() !== null) {
 				$min = self::pow($base->getValue(), $exponent->getMin());
 				if ($min === null) {
-					return null;
+					return new ErrorType();
 				}
 			}
 			if ($exponent->getMax() !== null) {
 				$max = self::pow($base->getValue(), $exponent->getMax());
 				if ($max === null) {
-					return null;
+					return new ErrorType();
 				}
 			}
 
@@ -104,9 +104,8 @@ final class ExponentiateHelper
 
 		if ($exponent instanceof ConstantScalarType) {
 			$result = self::pow($base->getValue(), $exponent->getValue());
-
 			if ($result === null) {
-				return null;
+				return new ErrorType();
 			}
 
 			if (is_int($result)) {
--- a/src/Type/ExponentiateHelper.php
+++ b/src/Type/ExponentiateHelper.php
@@ -2,7 +2,6 @@
 
 namespace PHPStan\Type;
 
-use PHPStan\Type\Accessory\AccessoryNumericStringType;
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/data/pow.php
```diff
@@ -177,3 +177,23 @@ function doFoo(int $intA, int $intB, string $s, bool $bool, $numericS, float $fl
 	assertType('*ERROR*', $bool ** $arr);
 	assertType('*ERROR*', $bool ** []);
 };
+
+function invalidConstantOperands(): void {
+	assertType('*ERROR*', 'a' ** 1);
```
staabm: php fatal errors on regular (non-numeric strings). I think we should too

https://3v4l.org/iTXRN

src/Type/ExponentiateHelper.php
```diff
@@ -77,16 +66,37 @@ public static function exponentiate(Type $base, Type $exponent): Type
 		]);
 	}
 
-	private static function exponentiateConstantScalar(ConstantScalarType $base, Type $exponent): ?Type
+	private static function exponentiateConstantScalar(Type $base, Type $exponent): ?Type
 	{
+		$allowedOperandTypes = new UnionType([
+			new IntegerType(),
+			new FloatType(),
+			new IntersectionType([
```
staabm: limited the allowed types to numeric-strings (instead of just string before)

src/Type/ExponentiateHelper.php
```diff
@@ -105,4 +115,12 @@ private static function exponentiateConstantScalar(ConstantScalarType $base, Typ
 		return null;
 	}
 
+	/**
+	 * @return float|int
+	 */
+	private static function pow(mixed $base, mixed $exp)
```
staabm: I had to wrap the pow operation into a separate method, as otherwise phpstan errored about invalid args to `pow()` / `**`?

src/Type/ExponentiateHelper.php
```diff
@@ -77,16 +66,37 @@ public static function exponentiate(Type $base, Type $exponent): Type
...
+			new FloatType(),
+			new IntersectionType([
+				new StringType(),
+				new AccessoryNumericStringType(),
+			]),
+			new BooleanType(),
+			new NullType(),
+		]);
+		if (!$allowedOperandTypes->isSuperTypeOf($exponent)->yes()) {
+			return new ErrorType();
```
ondrejmirtes: I disagree with returning ErrorType here.
`string` might still be a `numeric-string` so it's wrong to return ErrorType for that case.

We just want to prevent the internal error the issue is about.

ondrejmirtes (changes requested):

staabm:
reduced the PR to a minimal fix

ondrejmirtes:
Thank you.

PR #2797, opened 2023-11-29
Incremented numeric-string should change to int/float

staabm:
closes https://github.com/phpstan/phpstan/issues/10122
closes https://github.com/phpstan/phpstan/issues/10187

https://3v4l.org/7J17g


tests/PHPStan/Analyser/data/bug-10122.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Bug10122;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @param string $s
+ * @param numeric-string $ns
+ */
+function doFoo(string $s, $ns, float $f) {
+	assertType('string', ++$s);
```
zonuexe: `++$s` will be an int even if the input value is a string. https://3v4l.org/i4rru

refs https://github.com/phpstan/phpstan/issues/10187

staabm: great catch, thx

staabm: please have another look, I think it should look better now

staabm:
I have tested locally that the error in https://github.com/phpstan/phpstan/issues/10187 no longer happens as requested.
if required I can send a strict-rules `UselessCastRuleTest` regression test, but it feels not necessary.

ondrejmirtes (changes requested):
Please add regression test for 10187.

resources/functionMap_php80delta.php
```diff
@@ -102,7 +102,7 @@
 		'str_ends_with' => ['bool', 'haystack'=>'string', 'needle'=>'string'],
 		'str_starts_with' => ['bool', 'haystack'=>'string', 'needle'=>'string'],
 		'strchr' => ['string|false', 'haystack'=>'string', 'needle'=>'string', 'before_needle='=>'bool'],
-		'stripos' => ['int|false', 'haystack'=>'string', 'needle'=>'string', 'offset='=>'int'],
+		'stripos' => ['0|positive-int|false', 'haystack'=>'string', 'needle'=>'string', 'offset='=>'int'],
```
staabm, line 105: changed to have the same return type across php versions, so we don't need to ignore different errors per php version

staabm:
took me a few attempts to get this work across php versions.. but it should work now

ondrejmirtes:
Thank you.

PR #2799, opened 2023-11-30
Transform `if` / `else if` into flat `if` / `elseif` when possible

herndlm:
Closes https://github.com/phpstan/phpstan/issues/10113

Alternative to https://github.com/phpstan/phpstan-src/pull/2728

### DIFF
--- a/src/Parser/RichParser.php
+++ b/src/Parser/RichParser.php
@@ -78,6 +78,9 @@ class RichParser implements Parser
 		$traitCollectingVisitor = new TraitCollectingVisitor();
 		$nodeTraverser->addVisitor($traitCollectingVisitor);
 
+		$transformElseIfToFlatElseIfVisitor = new TransformElseIfToFlatElseIfVisitor();
+		$nodeTraverser->addVisitor($transformElseIfToFlatElseIfVisitor);
+
 		foreach ($this->container->getServicesByTag(self::VISITOR_SERVICE_TAG) as $visitor) {
 			$nodeTraverser->addVisitor($visitor);
 		}
--- a/src/Parser/RichParser.php
+++ b/src/Parser/RichParser.php
@@ -72,14 +72,15 @@ class RichParser implements Parser
 			throw new ShouldNotHappenException();
 		}
 
+		$transformNodeTraverser = new NodeTraverser();
+		$transformNodeTraverser->addVisitor(new TransformElseIfToFlatElseIfVisitor());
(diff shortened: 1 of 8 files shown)


herndlm:
Looking good, but apparently I made Rector on Windows angry somehow?

herndlm:
any idea what the (rector) issue might be here? it kept happening after a rebase, I don't want to rebase unnecessarily. wonder if I need to bump some rector cache key?

src/Parser/RichParser.php
```diff
@@ -66,6 +66,9 @@ public function parseString(string $sourceCode): array
 		$traitCollectingVisitor = new TraitCollectingVisitor();
 		$nodeTraverser->addVisitor($traitCollectingVisitor);
 
+		$transformElseIfToFlatElseIfVisitor = new TransformElseIfToFlatElseIfVisitor();
+		$nodeTraverser->addVisitor($transformElseIfToFlatElseIfVisitor);
```
ondrejmirtes: I'm a bit surprised this works. I'm not sure about the order and effects the visitors are called in. I'd expect the other visitors not being able to see effects from this one.

I'd be a little bit calmer if the AST was transformed with `TransformElseIfToFlatElseIfVisitor` by a separate NodeTraverser and only after that it was passed to the current `$nodeTraverser`.

ondrejmirtes (changes requested):
Also: Please write a custom rule in `tests` that just verifies the AST is transformed as expected.

So it should hook onto `If_` and just report how many `elseifs` it sees.

The Rector issue might or might not be fixed by other suggestion :)

herndlm:
Got inspired by `CleaningParserTest` and ended up writing the test a bit different. WDYT? the (stolen) core part could be useful for more traverse/transform tests if needed, but it felt not right to try to make it generic now.
Rector still doesn't like me here apparently :/

herndlm:
I believe this one is ready (for review) again. I was just waiting for you to switch to the simplified downgrader :)

ondrejmirtes:
The failures in Rector worry me a bit: https://github.com/phpstan/phpstan-src/actions/runs/7420725636/job/20192723240

herndlm:
I don't know Rector well enough but I guess the errors related to that if/else rector mean that the change here is not possible. Is https://github.com/phpstan/phpstan-src/pull/2728 a better alternative after all, maybe? Or something else? 🤔

PR #2803, opened 2023-12-05
ResolvedPhpDocBlock: fix parent return tag merging

janedbal:
This should resolve:
- https://github.com/phpstan/phpstan/issues/6462 
- https://github.com/phpstan/phpstan-doctrine/pull/503
- https://github.com/phpstan/phpstan/issues/10208
- https://github.com/phpstan/phpstan/issues/3580
- https://github.com/phpstan/phpstan/issues/4396

Current failure: https://phpstan.org/r/f8871cfa-7e74-48a3-854f-cb64495aabea

### DIFF
--- a/src/PhpDoc/ResolvedPhpDocBlock.php
+++ b/src/PhpDoc/ResolvedPhpDocBlock.php
@@ -22,10 +22,13 @@ use PHPStan\PhpDoc\Tag\TypedTag;
 use PHPStan\PhpDoc\Tag\UsesTag;
 use PHPStan\PhpDoc\Tag\VarTag;
 use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocNode;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\ConditionalTypeForParameter;
 use PHPStan\Type\Generic\TemplateTypeHelper;
 use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\Generic\TemplateTypeVariance;
+use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeTraverser;
 use function array_key_exists;
@@ -57,6 +60,8 @@ class ResolvedPhpDocBlock
 	private array $templateTags;
 
 	private PhpDocNodeResolver $phpDocNodeResolver;
+
+	private ReflectionProvider $reflectionProvider;
 
 	/** @var array<(string|int), VarTag>|false */
 	private array|false $varTags = false;
@@ -138,6 +143,7 @@ class ResolvedPhpDocBlock
 		TemplateTypeMap $templateTypeMap,
 		array $templateTags,
 		PhpDocNodeResolver $phpDocNodeResolver,
+		ReflectionProvider $reflectionProvider,
 	): self
 	{
 		// new property also needs to be added to createEmpty() and merge()
@@ -150,6 +156,7 @@ class ResolvedPhpDocBlock
 		$self->templateTypeMap = $templateTypeMap;
 		$self->templateTags = $templateTags;
 		$self->phpDocNodeResolver = $phpDocNodeResolver;
+		$self->reflectionProvider = $reflectionProvider;
 
 		return $self;
 	}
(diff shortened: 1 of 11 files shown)


src/PhpDoc/ResolvedPhpDocBlock.php
```diff
@@ -824,6 +837,21 @@ private static function mergeOneParentReturnTag(?ReturnTag $returnTag, self $par
 
 		$parentType = $parentReturnTag->getType();
 
+		if ($classReflection !== null) {
+			$parentType = TypeTraverser::map(
+				$parentType,
+				static function (Type $type, callable $traverse) use ($classReflection): Type {
+					if ($type instanceof StaticType) {
+						return $type->changeBaseClass($classReflection);
```
ondrejmirtes, line 845: Great, I'd fix this the same way so 👍 

I just hate adding new properties to this class but I didn't figure out a better solution. yet. Merging this.

ondrejmirtes (approved):

staabm:
it seems the change fixes more stuff, see https://github.com/phpstan/phpstan-src/actions/runs/7100699219?pr=2803

staabm:
Regression tests for all fixed issues would be good. Sorry I did not mention it earlier

janedbal:
@staabm All those look the same. But we can add them to be sure..

janedbal:
@staabm Added

tests/PHPStan/Rules/Methods/MethodSignatureRuleTest.php
```diff
@@ -474,4 +474,21 @@ public function testBug10184(): void
...
+		}
+
+		$this->reportMaybes = true;
+		$this->reportStatic = true;
+
+		$this->analyse([
+			__DIR__ . '/data/bug-10208.php',
+			__DIR__ . '/data/bug-6462.php',
+			__DIR__ . '/data/bug-4396.php',
+			__DIR__ . '/data/bug-3580.php',
```
ondrejmirtes: I need separate test method for each of these files. Allowing passing multiple files into `analyse` method was a mistake and I eventually want to get rid of it. Thanks.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Methods/MethodSignatureRuleTest.php
janedbal: Yeah, it felt weird, but I could instantly see 4 errors on 1.10.x instead of failing on first one.

Splitted.

janedbal: > I eventually want to get rid of it

What about emitting deprecation notice there?

ondrejmirtes: I'd accept removing a native type from the parameter and putting a `string` type in a bleeding-edge PHPDoc stub file for RuleTestCase :)

tests/PHPStan/Rules/Methods/MethodSignatureRuleTest.php
```diff
@@ -474,4 +474,19 @@ public function testBug10184(): void
...
+	public function testParentReturnTypeMergingFix(): void
+	{
+		if (PHP_VERSION_ID < 70400) {
+			$this->markTestSkipped('Test requires PHP 7.4.');
+		}
+
+		$this->reportMaybes = true;
+		$this->reportStatic = true;
+
+		$this->analyse([__DIR__ . '/data/bug-10208.php'], []);
```
ondrejmirtes, line 486: One `analyse` call per PHPUnit `test` method please 😂  That way we learn all of them are wrong instead of one at a time.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Methods/MethodSignatureRuleTest.php
janedbal: Oh god, some AI should read this and fix it instantly for me 😁

tests/PHPStan/Rules/Methods/MethodSignatureRuleTest.php
janedbal: Ok, that is for another MR for sure. Resolving this thread.

tests/PHPStan/Rules/Methods/MethodSignatureRuleTest.php
janedbal: Done

ondrejmirtes:
Thank you!

PR #2804, opened 2023-12-05
VarTagTypeRuleHelper: disallow widening list<int> to array<int>

janedbal:
This looks intentional in [this commit](https://github.com/phpstan/phpstan-src/commit/4e8dea569fc918a6b19230f0b077f6c9dda2c1ed), but it is imo just wrong. No widening should be allowed.


janedbal:
Please see #2805, I believe that one supersedes this.

ondrejmirtes:
See https://github.com/phpstan/phpstan-src/pull/2805#issuecomment-1841153347 and also the CI results that prove me right :)

PR #2805, opened 2023-12-05
VarTagTypeRuleHelper: remove all useless/wrong complexity

janedbal:
This is a better version of https://github.com/phpstan/phpstan-src/pull/2804, adding more and more conditions for edgecases will always leave rooms for errors. Lets keep it simple & correct.

Maybe we can enable this strict & proper behaviour only for bleedingEdge?

### DIFF
--- a/tests/PHPStan/Rules/PhpDoc/WrongVariableNameInVarTagRuleTest.php
+++ b/tests/PHPStan/Rules/PhpDoc/WrongVariableNameInVarTagRuleTest.php
@@ -270,6 +270,10 @@ class WrongVariableNameInVarTagRuleTest extends RuleTestCase
 				29,
 			],
 			[
+				'PHPDoc tag @var with type array<int> is not subtype of type list<int>.',
+				32,
+			],
+			[
 				'PHPDoc tag @var with type array<string> is not subtype of type list<int>.',
 				35,
 			],
@@ -280,6 +284,10 @@ class WrongVariableNameInVarTagRuleTest extends RuleTestCase
 			[
 				'PHPDoc tag @var with type Iterator<mixed, string> is not subtype of type Iterator<int, int>.',
 				44,
+			],
+			[
+				'PHPDoc tag @var with type array<int> is not subtype of type array<int, int>.',
+				47,
 			],
 			/*[
 				// reported by VarTagChangedExpressionTypeRule
@@ -293,6 +301,10 @@ class WrongVariableNameInVarTagRuleTest extends RuleTestCase
 			[
 				'PHPDoc tag @var with type int is not subtype of native type string.',
 				109,
+			],
+			[
+				'PHPDoc tag @var with type array<string> is not subtype of type array<int, string>.',
+				122,
 			],
 			[
 				'PHPDoc tag @var with type array<int> is not subtype of type array<int, string>.',
--- a/tests/PHPStan/Rules/PhpDoc/WrongVariableNameInVarTagRuleTest.php
+++ b/tests/PHPStan/Rules/PhpDoc/WrongVariableNameInVarTagRuleTest.php
@@ -323,6 +323,14 @@ class WrongVariableNameInVarTagRuleTest extends RuleTestCase
 				148,
 			],
 			[
+				'PHPDoc tag @var with type array<array<string>> is not subtype of type array<list<string|null>>.',
+				154,
+			],
+			[
+				'PHPDoc tag @var with type array<array<string>> is not subtype of type array<list<string|null>>.',
+				157,
+			],
+			[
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/PhpDoc/WrongVariableNameInVarTagRuleTest.php
```diff
@@ -356,10 +376,18 @@ public function testEmptyArrayInitWithWiderPhpDoc(bool $checkTypeAgainstNativeTy
...
+					6,
+				],
 				[
 					'PHPDoc tag @var with type int is not subtype of native type array{}.',
 					24,
 				],
+				[
+					'PHPDoc tag @var with type array is not subtype of native type array{}.',
+					29,
+				],
```
janedbal, line 390: All constant assignments are working fine. Those is imo invalid "init variable" approaches and should be reported. 

But, this is probably solvable in shouldVarTagTypeBeReported if you really insist on previous behaviour.

ondrejmirtes:
The complexity is there precisely because I want to allow:

```php
/** @var int[] $foo */
$foo = returnListOfIntegers();
```

Because it's harmless and PHPStan cannot be too annoying.

After one of your changes (https://github.com/phpstan/phpstan-src/pull/2738) I had to spend some time iterating so that the rule isn't too annoying:

* https://github.com/phpstan/phpstan-src/commit/42f6c2e64bc06dd60e2c3bc44c911170731e6b4c
* https://github.com/phpstan/phpstan-src/commit/67893d4b5cd7c82c83a9e9296c106bee7a52c4e9
* https://github.com/phpstan/phpstan-src/commit/256e8dcfb629265e8c1ecac75a6348a11a630738

As it stands right now when someone comes to me with a question regarding "PHPStan complains about my `@var`, what I should do?" I know how to help them and how they should improve their code.

After https://github.com/phpstan/phpstan-src/pull/2738 I found situations in real-world code that I myself didn't know how to solve (aside from deleting `@var` which led to other issues), that's why I needed to adjust it.

Inline `@var` PHPDoc is a horrible multi-faceted beast and we need to have the current heuristics so that PHPStan strikes the right balance between being useful and too pedantic.

janedbal:
I believe there should be some option to have it strictly based on type system, not based on [your taste](https://github.com/phpstan/phpstan-src/commit/42f6c2e64bc06dd60e2c3bc44c911170731e6b4c). This way (with all this complexity), there will be many more holes - just found [another issue](https://phpstan.org/r/5d5ff4cf-a639-42d6-b212-fb3019343cfe) (`array{id: int}|null` can be widened to `mixed[]|null`)

Also, [allowing list to array widening](https://github.com/phpstan/phpstan-src/pull/2804) may seem ok-ish for one, but we utilize lists a lot and this is degrading our efforts.

All those holes are hard to spot, I invested quite some time to find all those I reported ([1](https://github.com/phpstan/phpstan-src/pull/2738), [2](https://github.com/phpstan/phpstan-src/pull/2760), [3](https://github.com/phpstan/phpstan-src/pull/2804), [4](https://phpstan.org/r/5d5ff4cf-a639-42d6-b212-fb3019343cfe)). But if you insist keeping this as-is, I can always add the strict & proper behaviour to https://github.com/shipmonk-rnd/phpstan-rules

Last thing to consider: maybe if there was some way to do "declare variable type" in PHPStan (like some `@type`), people would stop abusing vardoc for that and many of those edgecases you want to adjust may disappear.

ondrejmirtes:
An idea: `@phpstan-var` where we can afford to be a bit stricter.

janedbal:
Not really a fan of that idea. I think people are used to that 

- `@phpstan-param` and `@psalm-param` and `@param`
- `@phpstan-return` and `@psalm-return` and `@return`

are almost synonyms. With some exceptions in IDEs.

janedbal:
@ondrejmirtes What about feature flag `strictVarDocChecks` ?

ondrejmirtes:
@janedbal Feature flags are for things we eventually want to enable for everybody (in the next major version). Which already applies to this rule (it's only in bleeding edge).

But I don't want this proposed behaviour in the next major version.

ondrejmirtes:
If you mean a new permanent option like listed here https://phpstan.org/config-reference#stricter-analysis, then sure, I'd merge that, there's already `reportWrongPhpDocTypeInVarTag` that makes it a bit more strict (and is enabled in phpstan-strict-rules).

janedbal:
Yeah, I mean regular option. Thx, I'll prepare that.

janedbal:
I suggest to name it `stricterPhpDocTypeCheckInVarTag` to correlate with existing one:

```neon
- reportWrongPhpDocTypeInVarTag: true
- stricterPhpDocTypeCheckInVarTag: true
```

PR #2806, opened 2023-12-06
Narrow ReflectionEnum::getName() after ReflectionEnum::isBacked()

franmomu:
Closes https://github.com/phpstan/phpstan/issues/10192


ondrejmirtes:
Thank you!

PR #2807, opened 2023-12-07
PhpDocParser: add config for lines in its AST & enable ignoring errors within phpdocs

janedbal:
This allows custom rules to properly report exact lines of e.g. `@param` tags when needed. 

image

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -1827,6 +1827,7 @@ services:
 		arguments:
 			parser: @currentPhpVersionPhpParser
 			lexer: @currentPhpVersionLexer
+			enableIgnoreErrorsWithinPhpDocs: %featureToggles.bleedingEdge%
 		autowired: no
 
 	currentPhpVersionSimpleParser:
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -52,6 +52,7 @@ parameters:
 		readOnlyByPhpDoc: false
 		phpDocParserRequireWhitespaceBeforeDescription: false
 		phpDocParserIncludeLines: false
+		enableIgnoreErrorsWithinPhpDocs: false
 		runtimeReflectionRules: false
 		notAnalysedTrait: false
 		curlSetOptTypes: false
@@ -1827,7 +1828,7 @@ services:
 		arguments:
 			parser: @currentPhpVersionPhpParser
 			lexer: @currentPhpVersionLexer
-			enableIgnoreErrorsWithinPhpDocs: %featureToggles.bleedingEdge%
+			enableIgnoreErrorsWithinPhpDocs: %featureToggles.enableIgnoreErrorsWithinPhpDocs%
 		autowired: no
 
 	currentPhpVersionSimpleParser:
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -51,6 +51,7 @@ parameters:
 		checkUnresolvableParameterTypes: false
 		readOnlyByPhpDoc: false
 		phpDocParserRequireWhitespaceBeforeDescription: false
+		phpDocParserIncludeLines: false
 		runtimeReflectionRules: false
 		notAnalysedTrait: false
 		curlSetOptTypes: false
@@ -389,6 +390,8 @@ services:
 		arguments:
 			requireWhitespaceBeforeDescription: %featureToggles.phpDocParserRequireWhitespaceBeforeDescription%
 			preserveTypeAliasesWithInvalidTypes: true
+			usedAttributes:
+				lines: %featureToggles.phpDocParserIncludeLines%
 
 	-
 		class: PHPStan\PhpDoc\ConstExprParserFactory
--- a/src/Rules/PhpDoc/PhpDocLineHelper.php
+++ b/src/Rules/PhpDoc/PhpDocLineHelper.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\PhpDoc;
+
+use PhpParser\Node;
+use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
+
+class PhpDocLineHelper
+{
+
+	/**
+	 * This method returns exact line of e.g. `@param` tag in PHPDoc so that it can be used for presice error reporting
+	 * - exact position is available only when bleedingEdge is enabled
+	 * - otherwise, it falls back to given node start line
+	 */
+	public static function detectTagLine(Node $node, PhpDocTagNode $phpDocTagNode): int
+	{
+		$phpDocTagLine = $phpDocTagNode->getAttribute('startLine');
+		$phpDoc = $node->getDocComment();
+
+		if ($phpDocTagLine === null || $phpDoc === null) {
+			return $node->getLine();
+		}
+
+		return $phpDoc->getStartLine() + $phpDocTagLine - 1;
+	}
+
+}
--- a/src/Rules/PhpDoc/PhpDocLineHelper.php
+++ b/src/Rules/PhpDoc/PhpDocLineHelper.php
@@ -2,8 +2,8 @@
 
 namespace PHPStan\Rules\PhpDoc;
 
-use PhpParser\Node;
-use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
+use PhpParser\Node as PhpParserNode;
+use PHPStan\PhpDocParser\Ast\Node as PhpDocNode;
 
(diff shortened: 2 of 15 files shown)


ondrejmirtes:
I'm looking forward to this! I'd like a proof of concept of using this in some 1st party PHPStan rules here that report things about PHPDocs. It should probably be done through some helper class so that we have logic in a single place that makes sure "Do we have the lines? Report precise lines. Otherwise report the line with the method declaration" that would apply this only to bleeding edge.

One of the challenges we'll face is ignoring errors I think. Right now you can ignore errors like this: https://phpstan.org/r/31387e59-5309-44f9-8a54-eeff03762cc7

But if line 6 gets reported instead of line 8 then we need to do something about ignoring errors too.

src/Rules/PhpDoc/PhpDocLineHelper.php
```diff
@@ -0,0 +1,28 @@
...
+namespace PHPStan\Rules\PhpDoc;
+
+use PhpParser\Node;
+use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
+
+class PhpDocLineHelper
+{
+
+	/**
+	 * This method returns exact line of e.g. `@param` tag in PHPDoc so that it can be used for presice error reporting
```
staabm:
```suggestion
	 * This method returns exact line of e.g. `@param` tag in PHPDoc so that it can be used for precise error reporting
```

tests/PHPStan/Analyser/data/ignore-next-line.php
```diff
@@ -17,16 +17,6 @@ public function doFoo(): void
...
 		/** @phpstan-ignore-next-line */
 		fail();
-
-		/*
-		 * @phpstan-ignore-next-line
-		 */
-		fail();
-
-		/**
-		 * @phpstan-ignore-next-line
```
janedbal, line 27: This is tricky BC break. In my POV, this was always wrong since it was not ignoring next line. It could even be N + 10 line.

```
/*
 * @phpstan-ignore-next-line
 * @param $notIgnored
 * @throws StillNotIgnored
 */
ignored();
```

But it will most probably break stuff in some codebases, so we may use this fixed behaviour only for bleedingEdge.

tests/PHPStan/Rules/PhpDoc/data/invalid-phpstan-doc.php
```diff
@@ -14,10 +14,10 @@ class Boo extends Baz
 	function foo(string $a){}
 
 	/**
-	 * @phpstan-ignore-next-line
```
janedbal: If you want, I can make this rule to use proper lines too and this remains working.

tests/PHPStan/Analyser/data/ignore-next-line.php
janedbal: Yeah, it occurs from time to time: https://github.com/search?q=%40phpstan-ignore-next-line&type=code

janedbal: So I kept the old behaviour for non bleeding edge.

conf/config.neon
```diff
@@ -1822,6 +1825,7 @@ services:
 		arguments:
 			parser: @currentPhpVersionPhpParser
 			lexer: @currentPhpVersionLexer
+			enableIgnoreErrorsWithinPhpDocs: %featureToggles.bleedingEdge%
```
ondrejmirtes: `phpDocParserIncludeLines` would make more sense here

src/Rules/PhpDoc/PhpDocLineHelper.php
```diff
@@ -0,0 +1,28 @@
...
+
+class PhpDocLineHelper
+{
+
+	/**
+	 * This method returns exact line of e.g. `@param` tag in PHPDoc so that it can be used for precise error reporting
+	 * - exact position is available only when bleedingEdge is enabled
+	 * - otherwise, it falls back to given node start line
+	 */
+	public static function detectTagLine(Node $node, PhpDocTagNode $phpDocTagNode): int
```
ondrejmirtes: Any https://phpstan.github.io/phpdoc-parser/PHPStan.PhpDocParser.Ast.Node.html should be accepted here, not just tag.

tests/PHPStan/Rules/PhpDoc/data/invalid-phpstan-doc.php
ondrejmirtes: Yes, please update this rule the same way the others, it's also about errors in PHPDocs.

ondrejmirtes (changes requested):
Also - after we merge this, please also checkout 1.11.x and make sure that `@phpstan-ignore` works as expected. It's a new way to ignore errors with identifiers and it automatically decides if the error to ignore should be on the current line, or the next line.

So somehow it should work inside PHPDocs too.

conf/config.neon
janedbal: I dont think so, those two are not really bounded. This is just fixed behaviour of ignore-error. Unrelated to phpDocParser setup.

We can name it `%featureToggles.ignoreErrorsWithinPhpDocs%` if you want

ondrejmirtes: Yeah please do

janedbal:
Should I update docs somehow? The broken edgecase was never documented: https://phpstan.org/user-guide/ignoring-errors

```
/**
 * @phpstan-ignore-next-line
 *
 * This is the legacy behaviour, the next line is meant as next-non-comment-line
 */
fail();
```

I'd say it even states the fixed one :D

> Both @phpstan-ignore-line and @phpstan-ignore-next-line ignore all errors on the corresponding line.

tests/PHPStan/Rules/PhpDoc/data/ignore-line-within-phpdoc.php
```diff
@@ -0,0 +1,28 @@
...
+ *  // @phpstan-ignore-next-line
+ * @param ( $invalid2
+ *
+ * @param ( $invalid1 // @phpstan-ignore-line
+ *
+ * @param ( $invalid2 // @phpstan-ignore-line
+ *
+ *
+ * // @phpstan-ignore-next-line
+ * @param ( $invalid2
```
ondrejmirtes, line 19: Does this also work? Please add a test:

```
 * @phpstan-ignore-next-line
 * @param ( $invalid2
 * ```

**Review by ondrejmirtes — CHANGES_REQUESTED** [review]

**Comment by ondrejmirtes** [comment]
> Should I update docs somehow?

No, it's okay, with that test I was just testing the behaviour to prove the fact that "next line" means "next line after this PHPDoc"

File: tests/PHPStan/Rules/PhpDoc/data/ignore-line-within-phpdoc.php (hunk 6 above)
Comment (janedbal) [reply tests/PHPStan/Rules/PhpDoc/data/ignore-line-within-phpdoc.php:19]: Sure it does, adjusted.

**Comment by ondrejmirtes** [comment]
@janedbal If you rebase now the build should not fail because of Rector anymore (after I wrote 1.5k lines in simple-downgrader, 3.5k if you include tests 😂 )

**Comment by janedbal** [comment]
Rebased.

**Comment by ondrejmirtes** [comment]
Awesome, thank you!

**Comment by gnutix** [comment]
> No, it's okay, with that test I was just testing the behaviour to prove the fact that "next line" means "next line after this PHPDoc"

I think this behavior is now broken @ondrejmirtes. I just upgraded and started getting : 

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/.../SomeClass.php                                                                         
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------- 
  167    No error to ignore is reported on line 167.                                                                                                         
  168    Method App\...\SomeClass::getFunctions() return type has no signature specified for Closure.  
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------- 
```

For code : 

```php
    /**
     * @phpstan-ignore-next-line I have no idea how this Closure is typed, that's the point...
     *
     * @return array<Closure|DeferredEvaluation>
     */
    private function getFunctions(Context $context): array
    {
         // ...
    }
```

If I change it to the following, then it works again : 

```php
    /**
     * @return array<Closure|DeferredEvaluation>
     */
    /** @phpstan-ignore-next-line I have no idea how this Closure is typed, that's the point... */
    private function getFunctions(Context $context): array
```

janedbal:
@gnutix That is intentional. Now (with bleeding edge), next line means actual next line. See https://github.com/phpstan/phpstan/discussions/10374

PR #2808, opened 2023-12-08
Update return type information for SimpleXMLElement::addChild

DaveLiddament:
Fix for [#10269](https://github.com/phpstan/phpstan/issues/10269)


ondrejmirtes:
Thank you! :)

PR #2810, opened 2023-12-08
Update return type of `spl_autoload_functions` on PHP8.0+

ManuelHu:
The [PHP documentation](https://www.php.net/manual/en/function.spl-autoload-functions.php) states that `spl_autoload_functions` might only return `false` on PHP versions up to 7.4; it will return an empty array on PHP8+.

* `functionMap.php` contains the PHP 7.4 signature. **This is unchanged. Is this correct?**
* The updated signature was **only** added to `functionMap_php80delta.php`
* The PHP8.x stubs already have the (only) correct native return type `array` specified. Before this PR, PHPStan assumes a return type of `array` (as `list<...>|false` in functionMap.php is not a subtype of the native type `array`)
* with this change the functionMap type should be a proper subtype of the native return type from the subs.

This PR also fixes some static analysis errors in PHPStan's source code that resulted from the functionMap changes.

### DIFF
--- a/build/baseline-8.0.neon
+++ b/build/baseline-8.0.neon
@@ -1,7 +1,7 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Strict comparison using \\=\\=\\= between array and false will always evaluate to false\\.$#"
+			message: "#^Strict comparison using \\=\\=\\= between list<callable\\(string\\): void> and false will always evaluate to false\\.$#"
 			count: 1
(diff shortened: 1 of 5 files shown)


ManuelHu:
@ondrejmirtes This (already very old) PR should now be ready for review and merging. I hope I did everythng right for the version specific signatures, at least it seems to work.

This might break some custom `@var` type annotations (like here in PHPStan), should this maybe go into bleeding edge for now?

ondrejmirtes:
This is okay, thank you.

PR #2811, opened 2023-12-08
phpstan-assert with count results in type loss

staabm:
closes https://github.com/phpstan/phpstan/issues/10264

PR #2812, opened 2023-12-08
phpstan-assert with recursive count() results in type loss

staabm:
simliar to https://github.com/phpstan/phpstan-src/pull/2811

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Analyser;
 
+use Countable;
 use PhpParser\Node;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\ArrayDimFetch;
@@ -79,6 +80,7 @@ use function in_array;
 use function is_string;
 use function strtolower;
 use function substr;
+use const COUNT_NORMAL;
 
 class TypeSpecifier
 {
@@ -208,7 +210,7 @@ class TypeSpecifier
 
 			if (
 				$expr->left instanceof FuncCall
-				&& count($expr->left->getArgs()) === 1
+				&& count($expr->left->getArgs()) >= 1
 				&& $expr->left->name instanceof Name
 				&& in_array(strtolower((string) $expr->left->name), ['count', 'sizeof', 'strlen', 'mb_strlen'], true)
 				&& (
@@ -237,7 +239,7 @@ class TypeSpecifier
 			if (
 				!$context->null()
 				&& $expr->right instanceof FuncCall
-				&& count($expr->right->getArgs()) === 1
+				&& count($expr->right->getArgs()) >= 1
 				&& $expr->right->name instanceof Name
 				&& in_array(strtolower((string) $expr->right->name), ['count', 'sizeof'], true)
 				&& $leftType->isInteger()->yes()
@@ -247,6 +249,39 @@ class TypeSpecifier
 					|| ($context->false() && (new ConstantIntegerType(1 - $offset))->isSuperTypeOf($leftType)->yes())
 				) {
 					$argType = $scope->getType($expr->right->getArgs()[0]->value);
+
+					if ($context->truthy() && $argType->isArray()->maybe()) {
+						$countables = [];
+						if ($argType instanceof UnionType) {
+							$countableInterface = new ObjectType(Countable::class);
+							foreach ($argType->getTypes() as $innerType) {
+								if (
+									$innerType->isArray()->yes()
+								) {
+									$innerType = TypeCombinator::intersect(new NonEmptyArrayType(), $innerType);
+									if ($innerType->isList()->yes()) {
+										$innerType = AccessoryArrayListType::intersectWith($innerType);
+									}
+									$countables[] = $innerType;
+								}
+
+								if (
+									!$countableInterface->isSuperTypeOf($innerType)->yes()
+								) {
+									continue;
+								}
+
+								$countables[] = $innerType;
+							}
+						}
+
+						if (count($countables) > 0) {
+							$countableType = TypeCombinator::union(...$countables);
+
+							return $this->create($expr->right->getArgs()[0]->value, $countableType, $context, false, $scope, $rootExpr);
+						}
+					}
+
 					if ($argType->isArray()->yes()) {
 						$newType = new NonEmptyArrayType();
 						if ($context->true() && $argType->isList()->yes()) {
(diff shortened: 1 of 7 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -237,7 +237,7 @@ public function specifyTypesInCondition(
 			if (
 				!$context->null()
 				&& $expr->right instanceof FuncCall
-				&& count($expr->right->getArgs()) === 1
+				&& count($expr->right->getArgs()) >= 1
```
ondrejmirtes, line 242: There are more places in `TypeSpecifier` where `count` FuncCall is handled and the number of args gets checked.

Please write failing tests for these places and then try to fix them, thanks.

ondrejmirtes (changes requested):

ondrejmirtes:
There's a conflict, please rebase :)

tests/PHPStan/Analyser/data/list-count.php
```diff
@@ -22,3 +22,22 @@ function foo(array $items) {
...
+function bar(array $items) {
+	assertType('list<int>', $items);
+	if (count($items, COUNT_RECURSIVE) === 3) {
+		assertType('array{int, int, int}', $items);
+		array_shift($items);
+		assertType('array{int, int}', $items);
+	} elseif (count($items, COUNT_RECURSIVE) === 0) {
+		assertType('array{}', $items);
+	} elseif (count($items, COUNT_RECURSIVE) === 5) {
+		assertType('array{int, int, int, int, int}', $items);
```
staabm: hmm.. I copied this over from the above tests, but I am not sure whether thats correct in the recursive case..

staabm: aahh since the type of the list is `int` its correct (count cannot find more elements by recursing).

ondrejmirtes: Yeah but it's not going to be correct if one of the elements is an array in which case this assumption should not proceed.

tests/PHPStan/Analyser/data/list-count.php
```diff
@@ -22,3 +22,41 @@ function foo(array $items) {
...
+	assertType('list<int>', $items);
+}
+
+/**
+ * @param list<int|int[]> $items
+ */
+function recursiveCount(array $items):void {
+	assertType('list<array<int>|int>', $items);
+	if (count($items, COUNT_RECURSIVE) === 3) {
+		assertType('non-empty-list<array<int>|int>', $items);
```
staabm, line 90: we can't judge for the constant array case when recursive - therefore separate expectations here

tests/PHPStan/Analyser/data/list-count.php
staabm: do you think its worth checking for the array case, or is https://github.com/phpstan/phpstan-src/pull/2812/commits/c5beb9ca46f3cfb3dbcff6fac2332fef19b5edb1 good enough?

staabm: instead of adding inference of the `count("non-recursive-type", COUNT_RECURSIVE)`-case I think it would make more sense to create a new rule which errors when `COUNT_RECURSIVE` is used on a type which is not recursive?

src/Analyser/TypeSpecifier.php
```diff
@@ -900,8 +902,19 @@ private function specifyTypesForConstantBinaryExpression(
...
 				$argType = $scope->getType($exprNode->getArgs()[0]->value);
 				if ($argType->isArray()->yes()) {
+					if (count($exprNode->getArgs()) === 1) {
+						$isNormalCount = true;
+					} else {
+						$mode = $scope->getType($exprNode->getArgs()[1]->value);
+						$isNormalCount = (new ConstantIntegerType(COUNT_NORMAL))->isSuperTypeOf($mode)->yes();
+
+						if (!$isNormalCount) {
+							$isNormalCount = $argType->getIterableValueType()->isScalar()->yes();
```
ondrejmirtes: This is too restrictive. For example objects cannot be counted recursively, not even countables: https://3v4l.org/3GCpt

So you need to ask `isArray` and work with uncertainty - TrinaryLogic. Think about the result when it's definitely not an array, when it can be an array and when it's definitely an array.

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
```diff
@@ -924,8 +926,19 @@ private function specifyTypesForConstantBinaryExpression(
...
 				$argType = $scope->getType($exprNode->getArgs()[0]->value);
 				if ($argType->isArray()->yes()) {
+					if (count($exprNode->getArgs()) === 1) {
+						$isNormalCount = true;
+					} else {
+						$mode = $scope->getType($exprNode->getArgs()[1]->value);
+						$isNormalCount = (new ConstantIntegerType(COUNT_NORMAL))->isSuperTypeOf($mode)->yes();
+
+						if (!$isNormalCount) {
+							$isNormalCount = $argType->getIterableValueType()->isArray()->no();
```
ondrejmirtes, line 1006: I think this still misses some scenarios.

We should have different results for `$isNormalCount` yes/maybe/no.
And also different results for `$argType->getIterableValueType()->isArray()` yes/maybe/no.

If you create a matrix:

no + no
no + maybe
no + yes
maybe + no
maybe + maybe
maybe + yes
yes + no
yes + maybe
yes + yes

You'll see that some scenarios aren't covered correctly.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/count-maybe.php
```diff
@@ -0,0 +1,115 @@
...
+		assertType('float', $notCountable);
+	} else {
+		assertType('float', $notCountable);
+	}
+}
+
+function doBar3(float $notCountable, float $invalidMode): void
+{
+	if (count($notCountable, $invalidMode) > 0) {
+		assertType('float', $notCountable);
```
staabm, line 34: should we resolve to a `*ERROR*` on invalid mode? I was not sure, as we already have a error at the count level because of the invalid argument type...?

src/Analyser/TypeSpecifier.php
staabm: thanks, I added tests covering this testmatrix in 

`tests/PHPStan/Analyser/data/count-maybe.php`

staabm:
squashed, rebased against 1.11 and re-targetted the PR. not sure why CI  thinks there is a merge commit contained

ondrejmirtes:
Thank you.

thg2k:
@staabm @ondrejmirtes I don't know if you are aware of this, but this exact commit also fixes this bug:

```php
/** @param array<mixed>|false $a */
function foo($a): void
{
  while (count($a) > 0) {
    array_shift($a);
  }
}
```

Before this commit:
NO ERROR (wrong)

After this commit:
`Parameter #1 $value of function count expects array|Countable, array|false given.`

I thought it was worth bringing this up in case it's a side effect. Maybe we should add it as a test case?

staabm:
I will have a look. Thanks for noticing

PR #2813, opened 2023-12-08
Fix count(list) regression

staabm:
refs https://github.com/phpstan/phpstan-src/pull/2811#issuecomment-1847296674


ondrejmirtes:
Thank you!

PR #2814, opened 2023-12-08
Introduce ForbiddenInterfacesInClassImplementsRule

paulbalandan:


paulbalandan:
I really don't know what to do with the pmmp errors, so initial review is appreciated.

ondrejmirtes:
What about checking just the literal `extends` and `implements` in the class AST instead of reflection which can pull in weird stuff from `vendor` and PHP extensions which we can't fix?

PR #2815, opened 2023-12-09
Use const instead of properties

zonuexe:


src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
```diff
@@ -91,8 +89,8 @@ private function getPreliminarilyResolvedTypeFromFunctionCall(
 			return TypeUtils::toBenevolentUnion($defaultReturnType);
 		}
 
-		if ($subjectArgumentType->isNonEmptyString()->yes() && array_key_exists($functionReflection->getName(), $this->functionsReplacePosition)) {
-			$replaceArgumentPosition = $this->functionsReplacePosition[$functionReflection->getName()];
+		if ($subjectArgumentType->isNonEmptyString()->yes() && isset(self::FUNCTIONS_REPLACE_POSITION[$functionReflection->getName()])) {
```
staabm: why the change from `array_key_exists` to `isset()`?

zonuexe: Certainly there is no rational reason to replace `array_key_exists()` with `isset` to shave off a few bytes. Although both compile to a `ZEND_ARRAY_KEY_EXISTS` instruction rather than a function call, `array_key_exists()` has a very slight performance advantage since it avoids the cost of a null check.

ondrejmirtes:
Thank you!

PR #2816, opened 2023-12-09
Simplify default-return path in Extensions

staabm:
similar to https://github.com/phpstan/phpstan-src/pull/2782


ondrejmirtes:
Thank you.

PR #2817, opened 2023-12-10
Narrow Closure::bind $newScope param

mvorisek:
Same change as in https://github.com/JetBrains/phpstorm-stubs/pull/1591.

Am I right the phpstorm-stubs are used for `Pure` metadata only?

The default type of `newScope` arg is `'static'` which is not subtype of `class-string`, does phpstan imply that from the real reflection?

### DIFF
--- a/resources/functionMap_bleedingEdge.php
+++ b/resources/functionMap_bleedingEdge.php
@@ -2,6 +2,8 @@
 
 return [
 	'new' => [
+		'Closure::bind' => ['Closure', 'old'=>'Closure', 'to'=>'?object', 'scope='=>'object|class-string|\'static\'|null'],
+		'Closure::bindTo' => ['Closure', 'new'=>'?object', 'newscope='=>'object|class-string|\'static\'|null'],
 		'error_log' => ['bool', 'message'=>'string', 'message_type='=>'0|1|2|3|4', 'destination='=>'string', 'extra_headers='=>'string'],
 		'SplFileObject::flock' => ['bool', 'operation'=>'int-mask<LOCK_SH, LOCK_EX, LOCK_UN, LOCK_NB>', '&w_wouldblock='=>'0|1'],
 		'Imagick::adaptiveBlurImage' => ['bool', 'radius'=>'float', 'sigma'=>'float', 'channel='=>'Imagick::CHANNEL_*'],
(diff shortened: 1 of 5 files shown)


staabm:
> The default type of newScope arg is 'static' which is not subtype of class-string, does phpstan imply that from the real reflection?

You could use NodeScopeResolverTest to check your assumptions

ondrejmirtes (changes requested):
Please do not rename parameter names

mvorisek:
> > The default type of newScope arg is 'static' which is not subtype of class-string, does phpstan imply that from the real reflection?
> 
> You could use NodeScopeResolverTest to check your assumptions

Did you mean `tests/PHPStan/Analyser/NodeScopeResolverTest.php`? How can I assert a phpstan error message there if some non-class-string is passed?

staabm:
In these tests you can only assert types. If you need to assert error messages you have to go with a rule-test

mvorisek:
@ondrejmirtes I can added a test to `tests\PHPStan\Rules\Functions\CallToFunctionParametersRuleTest.php`, there is one simillar `testBug4413` test. Please let me know how to name the testcase and if I should improve also other types like https://github.com/phpstan/phpstan-src/blob/1.10.57/resources/functionMap.php#L9929 which needs `class-string` too.

resources/functionMap.php
```diff
@@ -996,8 +996,8 @@
 'closelog' => ['bool'],
 'Closure::__construct' => ['void'],
 'Closure::__invoke' => ['', '...args='=>''],
-'Closure::bind' => ['Closure', 'old'=>'Closure', 'to'=>'?object', 'scope='=>'object|string|null'],
-'Closure::bindTo' => ['Closure', 'new'=>'?object', 'newscope='=>'object|string|null'],
+'Closure::bind' => ['Closure', 'old'=>'Closure', 'to'=>'?object', 'scope='=>'object|class-string|null'],
+'Closure::bindTo' => ['Closure', 'new'=>'?object', 'newscope='=>'object|class-string|null'],
```
ondrejmirtes: We typically want to narrow parameter types only in bleedingEdge. This change should be done in resources/functionMap_bleedingEdge.php.

That's because this is basically a BC break. You're forcing users that pass a `string` in there to add/rewrite their PHPDoc types.

ondrejmirtes (changes requested):
Add whatever test you think is your best idea, I'll review if it's okay or not.

resources/functionMap.php
mvorisek: done

ondrejmirtes:
Thank you!

PR #2820, opened 2023-12-12
shipmonk/name-collision-detector: update to 2.1.0

janedbal:
- just better output there: https://github.com/shipmonk-rnd/name-collision-detector/releases/tag/2.1.0


ondrejmirtes:
Thank you.

PR #2822, opened 2023-12-13
array_udiff() comparator type is too strict

staabm:
closes https://github.com/phpstan/phpstan/issues/9697

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-9697.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-9697.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Bug9697;
+
+
+function doFoo(): void
+{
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Functions/data/bug-9697.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Bug9697;
+
+
+function doFoo(): void
+{
+	$oldItems = [1,2,3];
+	$newItems = [1,2];
+
+	$comparator = fn (int $a, int $b):int => $a - $b;
```
staabm, line 10: interessting.. I would expect the php 7.2/7.3 lint jobs to error on this file

https://3v4l.org/FrKmK

ondrejmirtes:
Thank you.

PR #2823, opened 2023-12-13
InvalidPromotedPropertiesRule - use FunctionLike as node type

paulbalandan:
The nodes of interest in this rule all implement `Node\FunctionLike` so I guessed we can use that as the `getNodeType`.


ondrejmirtes:
Awesome!

PR #2825, opened 2023-12-13
Fix mixing property and param attributes on promoted property

RobertMe:
Simplify the ParamAttributesRule by merging AttributeCheck::check calls and using an int mask to check for both TARGET_PARAMETER and TARGET_PROPERTY in one pass.

Fixes phpstan/phpstan#10298

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-10298.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-10298.php
@@ -1,0 +1,18 @@
+<?php declare(strict_types = 1);
+
+namespace Bug10298;
+
+#[\Attribute(\Attribute::TARGET_PROPERTY)]
+class PropAttr {}
+
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+class ParamAttr {}
+
+class Test
+{
+	public function __construct(
+		#[PropAttr]
+		#[ParamAttr]
+		public string $test
+	) {}
+}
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
There is no bug to fix: https://github.com/phpstan/phpstan/issues/10298#issuecomment-1854459122

tests/PHPStan/Rules/Functions/data/bug-10298.php
```diff
@@ -0,0 +1,18 @@
...
+
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+class ParamAttr {}
+
+class Test
+{
+	public function __construct(
+		#[PropAttr]
+		#[ParamAttr]
+		public string $test
```
darenas31415, line 16: Coming from this issue [phpstan/phpstan#10385](https://github.com/phpstan/phpstan/issues/10385) maybe you could improve the test suite by adding another test case to cover when an attribute has both property and parameter targets like:

```php
#[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_PARAMETER)]
class PropAndParamAttr {}

class Test
{
	public function __construct(
		#[PropAndParamAttr]
		public string $test
	) {}
}
```

ondrejmirtes: I believe there is already such a test. Feel free to check that and send a PR in case it isn't 😊

PR #2826, opened 2023-12-13
Added regression test

staabm:
this [was fixed](https://github.com/phpstan/phpstan/issues/5008#issuecomment-992642621) somewhen in the past

closes https://github.com/phpstan/phpstan/issues/5008


ondrejmirtes:
Thank you.

PR #2830, opened 2023-12-16
Narrow ReflectionEnum::getBackingType() after ReflectionEnum::isBacked()

staabm:
Fixes https://github.com/phpstan/phpstan/issues/10167


ondrejmirtes:
Thank you!

PR #2833, opened 2023-12-18
Hide internal commands

rvanvelzen:
Showing internal commands in the command listing may be confusing and isn't necessary.


ondrejmirtes:
Perfect, I needed this! Thank you.

PR #2834, opened 2023-12-18
Fix coalescing on an optional array offset

rvanvelzen:
Fix phpstan/phpstan#10317


ondrejmirtes:
Thank you very much!

shadowhand:
Thank you, this fix just found a couple of undiscovered issues!

PR #2835, opened 2023-12-18
More precise `RecursiveIteratorIterator::__construct()` types

staabm:
I had initially implemented it with a separate `StubFilesExtension` to load this more precise signature only in bleedingEdge.
later on I realized most iterator implemetations already use pretty narrow $flags/$mode types, therefore I went just with this more precise types.

closes https://github.com/phpstan/phpstan/issues/10324


stubs/iterable.stub
```diff
@@ -157,15 +157,19 @@ class ArrayIterator implements SeekableIterator, ArrayAccess, Countable
...
+	const LEAVES_ONLY=0;
+	const SELF_FIRST=1;
+	const CHILD_FIRST=2;
+	const CATCH_GET_CHILD=16;
 
 	/**
 	 * @param T $iterator
-	 * @param int $mode
-	 * @param int $flags
+	 * @param self::LEAVES_ONLY|self::SELF_FIRST|self::CHILD_FIRST $mode
```
ondrejmirtes: I'd put this in `resources/functionMap_bleedingEdge.php`.

ondrejmirtes (changes requested):

stubs/iterable.stub
staabm: initially I tried that, too. turns out I need a mix of both, as the constants were missing on the stub.

staabm:
phpstan-symfony errors are fixed with https://github.com/phpstan/phpstan-symfony/pull/375

stubs/iterable.stub
ondrejmirtes: I don't understand.

staabm: in the first iteration of this PR I tried adding it to `resources/functionMap_bleedingEdge.php` and did not succeed (without the additional change in iterable stub). now with the PR as is it works

ondrejmirtes: I'm interested in why you didn't succeeed. It should be fine.

stubs/iterable.stub
```diff
@@ -157,15 +157,12 @@ class ArrayIterator implements SeekableIterator, ArrayAccess, Countable
  */
 class RecursiveIteratorIterator
 {
-
 	/**
 	 * @param T $iterator
-	 * @param int $mode
-	 * @param int $flags
```
staabm, line 164: it seems as soon as I keep these phpdocs here, the signature from `resources/functionMap_bleedingEdge.php` does not work, as it starts failling

```
There was 1 failure:

1) PHPStan\Rules\Classes\InstantiationRuleTest::testBug10324
--- Expected
+++ Actual
@@ @@
-'23: Parameter #3 $flags of class RecursiveIteratorIterator constructor expects 0|16, 2 given.
+'
 '

/Users/staabm/workspace/phpstan-src/src/Testing/RuleTestCase.php:159
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Rules/Classes/InstantiationRuleTest.php:466

```

ondrejmirtes:
Makes sense now, thank you.

PR #2838, opened 2023-12-20
class_exists(Interface::class) is always false

janedbal:
Recently found [bug caused by this in phpstan-doctrine](https://github.com/phpstan/phpstan-doctrine/blob/1.3.53/src/Type/Doctrine/Query/QueryResultTypeWalker.php#L815).

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -75,14 +75,11 @@ class ImpossibleCheckTypeHelper
 					return $assertValue->getValue();
 				}
 				if (in_array($functionName, [
+					'class_exists',
 					'interface_exists',
 					'trait_exists',
 					'enum_exists',
 				], true)) {
-					return null;
-				}
-
-				if ($functionName === 'class_exists') {
 					if (count($node->getArgs()) < 1) {
 						return null;
 					}
@@ -95,9 +92,23 @@ class ImpossibleCheckTypeHelper
 						return null;
 					}
 					$reflection = $this->reflectionProvider->getClass($argConstantString->getValue());
-					if ($reflection->isInterface()) {
+
+					if ($functionName === 'class_exists' && ($reflection->isInterface() || $reflection->isTrait())) {
 						return false;
 					}
+
+					if ($functionName === 'interface_exists' && ($reflection->isClass() || $reflection->isTrait() || $reflection->isEnum())) {
+						return false;
+					}
+
+					if ($functionName === 'trait_exists' && ($reflection->isClass() || $reflection->isInterface() || $reflection->isEnum())) {
+						return false;
+					}
+
+					if ($functionName === 'enum_exists' && ($reflection->isClass() || $reflection->isInterface() || $reflection->isTrait())) {
+						return false;
+					}
+
 					return null;
 				}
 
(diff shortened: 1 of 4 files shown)


staabm:
I tried similar things in https://github.com/phpstan/phpstan-src/pull/2277
I think you can get inspiration about a few more cases which need to be handled here in the linked PR.

beware: ondrej might have different opinions.. feedback on the linked PR was not discussed yet

---

and we should have an eye on issuebot, as IIRC there should be more open issues regarding this topic.

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -80,8 +80,38 @@ public function findSpecifiedType(
...
+						return null;
+					}
+					$argType = $scope->getType($node->getArgs()[0]->value);
+					if (count($argType->getConstantStrings()) !== 1) {
+						return null;
+					}
+					$argConstantString = $argType->getConstantStrings()[0];
+					if (!$this->reflectionProvider->hasClass($argConstantString->getValue())) {
+						return null;
+					}
```
staabm, line 93: you are missing things like

```
$c = _Enum::class;
if (rand(0,1)) {
  $c = _Class::class;
}
class_exists($c);
```

janedbal: Yeah, I dont want to cover all edgecases..

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -80,8 +80,38 @@ public function findSpecifiedType(
...
+					if ($functionName === 'interface_exists' && ($reflection->isClass() || $reflection->isTrait() || $reflection->isEnum())) {
+						return false;
+					}
+
+					if ($functionName === 'trait_exists' && ($reflection->isClass() || $reflection->isInterface() || $reflection->isEnum())) {
+						return false;
+					}
+
+					if ($functionName === 'enum_exists' && ($reflection->isClass() || $reflection->isInterface() || $reflection->isTrait())) {
+						return false;
```
ondrejmirtes, line 109: I think these conditions could be improved. Name what it cannot be instead of what it can be, for example:

```php
if ($functionName === 'interface_exists' && !$reflection->isInterface()) {
```

ondrejmirtes (changes requested):
I'd rather be if this was somehow done in ClassExistsFunctionTypeSpecifyingExtension instead of ImpossibleCheckTypeHelper. 

ClassExistsFunctionTypeSpecifyingExtension is "closer to the core" and can influence the correctness in more places, for example `class_exists($foo)` for `$foo` being `class-string<SomeInterface>` should lead to `NeverType` for `$foo`.

ImpossibleCheckTypeHelper is only informing the rules right at the end of the analysis.

I understand this can lead to some more work because telling that `Scalar\String_` should be `NeverType` can break assumptions in some places.

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
janedbal: But that can break if some future PHP version onboards new struct type. But that would probably lead to many changes everywhere, so why not...

janedbal:
> I'd rather be if this was somehow done in ClassExistsFunctionTypeSpecifyingExtension

I agree, but cannot we do it as next step and merge this?

----

*Note: I dont feel implementing that in near future.*

ondrejmirtes:
We're in no rush here and doing this correctly can always teach us something new :)

janedbal:
> We're in no rush here and doing this correctly can always teach us something new :)

Ok, then we can probably close this PR.

staabm:
seems this is not beeing worked on. [extracted a new issue from the tests](https://github.com/phpstan/phpstan/issues/14683), so we don't loose them.

thank you

PR #2840, opened 2023-12-22
Add option reportAnyTypeWideningInVarTag

janedbal:
See https://github.com/phpstan/phpstan-src/pull/2805

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -122,6 +122,7 @@ parameters:
 	reportMaybesInPropertyPhpDocTypes: false
 	reportStaticMethodSignatures: false
 	reportWrongPhpDocTypeInVarTag: false
+	stricterPhpDocTypeCheckInVarTag: false
 	checkMissingOverrideMethodAttribute: false
 	mixinExcludeClasses: []
 	scanFiles: []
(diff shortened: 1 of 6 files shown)


conf/config.neon
```diff
@@ -121,6 +121,7 @@ parameters:
 	reportMaybesInPropertyPhpDocTypes: false
 	reportStaticMethodSignatures: false
 	reportWrongPhpDocTypeInVarTag: false
+	stricterPhpDocTypeCheckInVarTag: false
```
ondrejmirtes: `stricter` isn't really descriptive. It should say something that it always requires a subtype...

ondrejmirtes (changes requested):

conf/config.neon
janedbal: what about 

- `reportAnyWideningInPhpDocTypeInVarTag` (the new one)
- `reportWrongPhpDocTypeInVarTag`

ondrejmirtes: What about `reportAnyTypeWideningInVarTag`? `PhpDocType` is redundant and not always true, because the behaviour depends on `reportWrongPhpDocTypeInVarTag` being true or false.

janedbal: Ok then, `reportAnyTypeWideningInVarTag`

janedbal: renamed

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

janedbal:
Thank you!

PR #2841, opened 2023-12-23
Use `str_starts_with()` and `str_contains()` instead of `strpos()`

zonuexe:
These codes have been clarified by replacing them with functions introduced in PHP 8.x, and are only slightly micro-optimized.


ondrejmirtes:
Thank you! Yes, I prefer this, but to be clear, this is enabled only thanks to the fact that PHPStan includes Symfony polyfills :) But on newer PHP versions the native functions will be used.

PR #2842, opened 2023-12-23
Fix return type of DOMXPath::query

staabm:
> This can also return namespace nodes, which are not a child class of DOMNode.

analog https://github.com/vimeo/psalm/pull/10443

### DIFF
--- a/stubs/dom.stub
+++ b/stubs/dom.stub
@@ -54,7 +54,7 @@ class DOMElement extends DOMNode
 }
 
 /**
- * @template-covariant TNode as DOMNode
+ * @template-covariant TNode as DOMNode|DOMNameSpaceNode
  * @implements Traversable<int, TNode>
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Please show an example on 3v4l.org with examples of using this method and the cases where it returns different objects as stated in this PR. I'm worried of typical problems like that in 99 % of cases this is going to return what's already stated in the file, and this change is going to report these 99 % of cases for the sake of correctness for the 1 %.

staabm:
https://3v4l.org/GpOJA

ondrejmirtes:
So only a query that starts with `namespace` is going to return that?

ndossche:
> So only a query that starts with `namespace` is going to return that?

Only queries that explicitly request namespace nodes return that.

ondrejmirtes:
Yeah so this PR isn't really worth it in the current form.

PR #2845, opened 2024-01-01
Remove line from "nikic/php-parser" error message

mvorisek:
related discussion/issue: https://github.com/phpstan/phpstan/issues/10358#issuecomment-1873356412

### DIFF
--- a/src/Parser/ParserErrorsException.php
+++ b/src/Parser/ParserErrorsException.php
@@ -22,7 +22,7 @@ class ParserErrorsException extends Exception
 		private ?string $parsedFile,
 	)
 	{
-		parent::__construct(implode(', ', array_map(static fn (Error $error): string => $error->getMessage(), $errors)));
+		parent::__construct(implode(', ', array_map(static fn (Error $error): string => $error->getRawMessage(), $errors)));
 		if (count($errors) > 0) {
 			$this->attributes = $errors[0]->getAttributes();
 		} else {
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -1290,6 +1290,7 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$errors = $this->runAnalyse(__DIR__ . '/data/bug-10358.php');
 		$this->assertCount(1, $errors);
 		$this->assertSame('Cannot use Ns\Foo2 as Foo because the name is already in use', $errors[0]->getMessage());
+		$this->assertSame(6, $errors[0]->getLine());
 	}
 
 	public function testBug10509(): void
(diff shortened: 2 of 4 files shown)


mvorisek:
Please advise where can I put https://phpstan.org/r/11e0c7af-bd52-4add-8de0-98397c5d492e test and on what test can I base it - I need to assert the phpstan error message.

ondrejmirtes:
AnalyserIntegrationTest

mvorisek:
The CI failures should be unrelated.

src/Parser/ParserErrorsException.php
```diff
@@ -22,7 +22,7 @@ public function __construct(
 		private ?string $parsedFile,
 	)
 	{
-		parent::__construct(implode(', ', array_map(static fn (Error $error): string => $error->getMessage(), $errors)));
+		parent::__construct(implode(', ', array_map(static fn (Error $error): string => $error->getRawMessage(), $errors)));
```
ondrejmirtes, line 25: I think one of the places are not tested. Please see the path where this is actually used, perhaps by totally breaking this piece of code and seeing where it fails.

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1270,6 +1270,13 @@ public function testBug10147(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug10358(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-10358.php');
+		$this->assertCount(1, $errors);
+		$this->assertSame('Cannot use Ns\Foo2 as Foo because the name is already in use', $errors[0]->getMessage());
```
ondrejmirtes, line 1292: Please assert the line too.

ondrejmirtes (changes requested):

src/Parser/ParserErrorsException.php
mvorisek: `src/Analyser/FileAnalyser.php` change is currently tested by `tests/PHPStan/Analyser/AnalyserIntegrationTest.php::testBug10358`

mvorisek: The other exception is consumed in https://github.com/phpstan/phpstan-src/blob/1.10.67/src/Analyser/FileAnalyser.php#L211 but the message is not used - so there is currently not easy way to test the message, but it is tested in https://github.com/phpstan/phpstan-src/blob/1.10.67/tests/PHPStan/Analyser/AnalyserTest.php#L408.

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
mvorisek: done

ondrejmirtes:
Thank you!

PR #2848, opened 2024-01-02
fix(DirectoryCreator): Check if directory exists before creating

aldilop:
Suppressing the error with '@' still calls custom error handlers.
We can check if the directory exists before calling mkdir.

### DIFF
--- a/src/Internal/DirectoryCreator.php
+++ b/src/Internal/DirectoryCreator.php
@@ -18,6 +18,9 @@ final class DirectoryCreator
 		bool $recursive = true,
 	): void
 	{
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I disagree. You should fix your custom error handler instead.

ondrejmirtes:
Just did this to improve compatibility with PHPUnit and PHP-Parser 5: https://github.com/phpstan/phpstan-src/commit/9dac90dfd5fc809dd84fdac0532ecd51c6ae66bf

aldilop:
Thanks!

PR #2853, opened 2024-01-04
Proper merge 1.10 to 1.11

janedbal:
Supersedes https://github.com/phpstan/phpstan-src/pull/2852 as it should contain proper merge commit


ondrejmirtes:
Thank you!

PR #2856, opened 2024-01-05
Implement RequireExtendsPropertiesClassReflectionExtension

staabm:


src/Reflection/RequireExtension/RequireExtendsPropertiesClassReflectionExtension.php
```diff
@@ -0,0 +1,74 @@
...
+namespace PHPStan\Reflection\RequireExtension;
+
+use PHPStan\Analyser\OutOfClassScope;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\PropertiesClassReflectionExtension;
+use PHPStan\Reflection\PropertyReflection;
+use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\VerbosityLevel;
+
+class RequireExtendsPropertiesClassReflectionExtension implements PropertiesClassReflectionExtension
```
staabm, line 11: one extendsion for properties, one for methods? or combined in one class?

and another pair for require-implements?

I gues too many extension could make it slow - because of exzessive reflection lookups

ondrejmirtes: I think we can combine extends/implements into a single extension, but keep methods and properties as separate extensions.

ondrejmirtes (changes requested):
Please add some regression tests that use these tags to make sure this actually solves the problems:

* https://github.com/phpstan/phpstan/issues/10302
* https://github.com/phpstan/phpstan/issues/9899
* https://github.com/phpstan/phpstan/issues/8550

staabm:
(just pushed a intermediate state to allow me switching to a different computer over the weekend)

tests/PHPStan/Rules/Properties/data/require-implements.php
```diff
@@ -0,0 +1,37 @@
...
+
+interface MyInterface
+{
+	public function doSomething(): string;
+
+	static public function doSomethingStatic(): int;
+}
+
+/**
+ * @phpstan-require-implements MyInterface
```
staabm, line 13: in this case here, the userland code needs to be

```
class MyClass implements MyInterface
{
	use MyTrait;
}
```

which means for `require-implements` I don't need a class-reflection extension, because the default-extension will already allow access to the interface methods, right?

we only need additional rules to verify the `@phpstan-require-implements MyInterface` is used properly, right?

staabm: turned differently: my thinking is we need the class-reflection extension only to allow access to properties/methods for the `@phpstan-require-extends` case?

src/PhpDoc/ResolvedPhpDocBlock.php
```diff
@@ -94,6 +96,11 @@ class ResolvedPhpDocBlock
 	/** @var array<MixinTag>|false */
 	private array|false $mixinTags = false;
 
+	/** @var array<RequireExtendsTag>|false */
+	private array|false $requireExtendsTags = false;
+	/** @var array<RequireImplementsTag>|false */
+	private array|false $requireImplementsTags = false;
```
ondrejmirtes: Needs to be added to createEmpty

src/Reflection/ClassReflection.php
```diff
@@ -425,7 +427,7 @@ public function hasProperty(string $propertyName): bool
 		}
 
 		foreach ($this->propertiesClassReflectionExtensions as $i => $extension) {
-			if ($i > 0 && !$this->allowsDynamicPropertiesExtensions()) {
+			if ($i > 1 && !$this->allowsDynamicPropertiesExtensions()) {
```
ondrejmirtes: This looks like some magic. I'd want to inject the new extensions into ClassReflection and call them regardless of `allowsDynamicPropertiesExtensions`.

src/Reflection/RequireExtension/RequireExtendsOrImplementsMethodsClassReflectionExtension.php
```diff
@@ -0,0 +1,81 @@
...
+			$static = $method->isStatic();
+
+			if (
+				!$static
+				&& $classReflection->hasNativeMethod('__callStatic')
+			) {
+				$static = true;
+			}
+
+			return new RequireExtendsOrImplementsMethodReflection($method, $static);
```
ondrejmirtes: You don't need to create new class here, just get the method reflection from the type.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Properties/data/require-implements.php
staabm: I still think we only need the class-reflection extension for the case where we use `@phpstan-require-extends` on a interface.

when using `@phpstan-require-extends` or `@phpstan-require-implements` on a trait, we only need to create a few rules to sanity check the annotations context and params are valid

src/Reflection/ClassReflectionExtensionRegistry.php
```diff
@@ -53,4 +57,14 @@ public function getAllowedSubTypesClassReflectionExtensions(): array
 		return $this->allowedSubTypesClassReflectionExtensions;
 	}
 
+	public function getExtendsPropertyClassReflectionExtension(): RequireExtendsPropertiesClassReflectionExtension
```
staabm:
```suggestion
	public function getRequireExtendsPropertyClassReflectionExtension(): RequireExtendsPropertiesClassReflectionExtension
```

src/Reflection/ClassReflectionExtensionRegistry.php
```diff
@@ -53,4 +57,14 @@ public function getAllowedSubTypesClassReflectionExtensions(): array
...
+	{
+		return $this->requireExtendsPropertiesClassReflectionExtension;
+	}
+
+	public function getExtendsMethodsClassReflectionExtension(): RequireExtendsMethodsClassReflectionExtension
```
staabm:
```suggestion
	public function getRequireExtendsMethodsClassReflectionExtension(): RequireExtendsMethodsClassReflectionExtension
```

src/PhpDoc/ResolvedPhpDocBlock.php
```diff
@@ -181,6 +189,8 @@ public static function createEmpty(): self
 		$self->returnTag = null;
 		$self->throwsTag = null;
 		$self->mixinTags = [];
+		$self->requireExtendsTags = [];
+		$self->requireImplementsTags = [];
```
ondrejmirtes: The properties assignment is also missing from `merge` method.

src/Reflection/ClassReflection.php
```diff
@@ -424,6 +430,10 @@ public function hasProperty(string $propertyName): bool
 			return $this->hasNativeProperty($propertyName);
 		}
 
+		if ($this->requireExtendsPropertiesClassReflectionExtension->hasProperty($this, $propertyName)) {
```
ondrejmirtes: Maybe do it below the foreach?

src/Reflection/ClassReflection.php
```diff
@@ -438,6 +448,10 @@ public function hasProperty(string $propertyName): bool
 
 	public function hasMethod(string $methodName): bool
 	{
+		if ($this->requireExtendsMethodsClassReflectionExtension->hasMethod($this, $methodName)) {
```
ondrejmirtes: Maybe do it below the foreach?

src/Reflection/ClassReflection.php
```diff
@@ -453,6 +467,15 @@ public function getMethod(string $methodName, ClassMemberAccessAnswerer $scope):
 		if ($scope->isInClass()) {
 			$key = sprintf('%s-%s', $key, $scope->getClassReflection()->getCacheKey());
 		}
+
+		if ($this->requireExtendsMethodsClassReflectionExtension->hasMethod($this, $methodName)) {
```
ondrejmirtes: Maybe do it below the foreach? If the method already exists the native one should have precedence.

src/Reflection/ClassReflection.php
```diff
@@ -453,6 +467,15 @@ public function getMethod(string $methodName, ClassMemberAccessAnswerer $scope):
...
+			$method = $this->wrapExtendedMethod($this->requireExtendsMethodsClassReflectionExtension->getMethod($this, $methodName));
```
ondrejmirtes: We should not need to wrap it. The extension should already return the native method.

src/Reflection/ClassReflection.php
```diff
@@ -576,10 +599,21 @@ public function getProperty(string $propertyName, ClassMemberAccessAnswerer $sco
 			$key = sprintf('%s-%s', $key, $scope->getClassReflection()->getCacheKey());
 		}
 		if (!isset($this->properties[$key])) {
+			if ($this->requireExtendsPropertiesClassReflectionExtension->hasProperty($this, $propertyName)) {
+				$property = $this->requireExtendsPropertiesClassReflectionExtension->getProperty($this, $propertyName);
+				if ($scope->canAccessProperty($property)) {
+					return $this->properties[$key] = $property;
```
ondrejmirtes: Do it below the foreach

src/Reflection/ClassReflection.php
ondrejmirtes: Also you don't need to ask about "canCallMethod".

src/Reflection/ClassReflection.php
ondrejmirtes: Also you don't need to ask about "canAccessProperty".

src/Reflection/ClassReflectionExtensionRegistry.php
ondrejmirtes: Yes please

src/Reflection/ClassReflectionExtensionRegistry.php
ondrejmirtes: Yes please.

src/Reflection/RequireExtension/RequireExtendsMethodsClassReflectionExtension.php
```diff
@@ -0,0 +1,73 @@
...
+
+	/** @var array<string, array<string, true>> */
+	private array $inProcess = [];
+
+	public function hasMethod(ClassReflection $classReflection, string $methodName): bool
+	{
+		return $this->findMethod($classReflection, $methodName) !== null;
+	}
+
+	public function getMethod(ClassReflection $classReflection, string $methodName): MethodReflection
```
ondrejmirtes, line 23: This can have ExtendedMethodReflection return type.

src/Reflection/RequireExtension/RequireExtendsMethodsClassReflectionExtension.php
```diff
@@ -0,0 +1,73 @@
...
+
+			$typeDescription = $type->describe(VerbosityLevel::typeOnly());
+			if (isset($this->inProcess[$typeDescription][$methodName])) {
+				continue;
+			}
+
+			$this->inProcess[$typeDescription][$methodName] = true;
+
+			if (!$type->hasMethod($methodName)->yes()) {
+				unset($this->inProcess[$typeDescription][$methodName]);
```
ondrejmirtes: Is this recursion guard tested?

src/Reflection/RequireExtension/RequireExtendsMethodsClassReflectionExtension.php
```diff
@@ -0,0 +1,73 @@
...
+			throw new ShouldNotHappenException();
+		}
+
+		return $method;
+	}
+
+	private function findMethod(ClassReflection $classReflection, string $methodName): ?MethodReflection
+	{
+		if (!$classReflection->isInterface()) {
+			return null;
```
ondrejmirtes, line 39: This is a point for the rules - `@...-require-extends` can only be above an interface.

src/Reflection/RequireExtension/RequireExtendsPropertiesClassReflectionExtension.php
```diff
@@ -0,0 +1,72 @@
...
+		$requireExtendsTags = $classReflection->getRequireExtendsTags();
+		foreach ($requireExtendsTags as $requireExtendsTag) {
+			$type = $requireExtendsTag->getType();
+
+			$typeDescription = $type->describe(VerbosityLevel::typeOnly());
+			if (isset($this->inProcess[$typeDescription][$propertyName])) {
+				continue;
+			}
+
+			$this->inProcess[$typeDescription][$propertyName] = true;
```
ondrejmirtes: Is this recursion guard tested?

tests/PHPStan/Rules/Properties/data/require-implements.php
ondrejmirtes: Yeah, looks like it 👍

ondrejmirtes (changes requested):

src/Reflection/RequireExtension/RequireExtendsMethodsClassReflectionExtension.php
staabm: so do you mean we should still try to infer the property across the type hierarchy, even though the `require-extends` is placed on a invalid location?

ondrejmirtes: Sorry, I don't get it. This comment was just a reminder for you to check whether `require-extends` is always above an interface.

src/Reflection/ClassReflection.php
```diff
@@ -465,6 +480,11 @@ public function getMethod(string $methodName, ClassMemberAccessAnswerer $scope):
 				}
 				$this->methods[$key] = $method;
 			}
+
+			if ($this->requireExtendsMethodsClassReflectionExtension->hasMethod($this, $methodName)) {
+				$method = $this->requireExtendsMethodsClassReflectionExtension->getMethod($this, $methodName);
+				$this->methods[$key] = $method;
```
ondrejmirtes, line 488: Why are you overriding a method that might have already been set in the foreach above?

src/Reflection/ClassReflection.php
```diff
@@ -590,6 +613,11 @@ public function getProperty(string $propertyName, ClassMemberAccessAnswerer $sco
 				}
 				$this->properties[$key] = $property;
 			}
+
+			if ($this->requireExtendsPropertiesClassReflectionExtension->hasProperty($this, $propertyName)) {
+				$property = $this->requireExtendsPropertiesClassReflectionExtension->getProperty($this, $propertyName);
+				$this->properties[$key] = $property;
```
ondrejmirtes, line 622: Why are you overriding a property that might have already been set in the foreach above?

ondrejmirtes (changes requested):

src/Reflection/ClassReflection.php
staabm: it seems we cannot use the more specialized return type, because PHP 7.2/7.3 don't support covariance on the return type?

https://github.com/phpstan/phpstan-src/actions/runs/7431936254/job/20223170712?pr=2856

https://www.php.net/manual/en/language.oop5.variance.php

ondrejmirtes: Just write it in the PHPDoc.

src/Reflection/RequireExtension/RequireExtendsMethodsClassReflectionExtension.php
staabm: I tried thinking of a way this could get recursive, but wasn't able to come up with a example.

do you think its impossible and the recursion-guard should be removed?

ondrejmirtes: I'd expect something like this to end up in infinite recursion https://phpstan.org/r/aa768c71-4596-4edf-a900-60e4db58582b

staabm: thx for the inspiration. I tried similar things before and I can't force them to endless recurse.

I think this might not happen, because the new extension early exits for non-interfaces and the `phpstan-require-extends` requires a class as argument.

since I was still not able to force a recursion, I have removed the recursion guards for now

src/Reflection/RequireExtension/RequireExtendsMethodsClassReflectionExtension.php
staabm: rules will be topic for a followup PR. thanks for noticing.

ondrejmirtes:
I merged part of this pull request: https://github.com/phpstan/phpstan-src/commit/53a61dc8674fe5c61fcc08efe08221e919661132

You should be able to rebase this PR cleanly (maybe after you squash all commits).

The reason why I did this is that you can start working on the rules PR in parallel to this PR.

I want to keep 1.10.x releasable and I don't want to merge this PR and then wait several releases for the rules PR to be finished, because I see them as an atomic feature.

You should be able to work on rules in a separate branch/PR without any conflicts.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

ondrejmirtes:
FYI this was a missing part: https://github.com/phpstan/phpstan-src/commit/9ac6fd1a4325adcf9e2ebdc2738e177bd459032f

ondrejmirtes:
About generics - I'd like to support things like:

```php
/**
 * @template T
 * @phpstan-require-extends Model<T>
 */
```

1) Which would inform about what `T` is in case the `Model` typehints it in the accessed property/called method. I think it might already work with the existing code.

2) Also when doing `class Foo implements Bar extends Model`, it should enforce that the same type in `@implements Bar<X>` is also used in `@extends Model<X>`.

3) Also if `Model` is generic but the user does not declare what `T` is in `@phpstan-require-extends`, PHPStan should complain.

4) It should for example also complain when `Model` has two `@template`  but the number of type variables isn't 2.

For 3) and 4) there are already existing checks which we should just call.

Do you feel up to the task? If you start working on this and hit a roadblock don't worry, open a draft PR and we'll figure out why it doesn't work. Thanks!

staabm:
> FYI this was a missing part: [9ac6fd1](https://github.com/phpstan/phpstan-src/commit/9ac6fd1a4325adcf9e2ebdc2738e177bd459032f)

great catch, thank you.

staabm:
@ondrejmirtes should we add the require-extends and require-implements types to `$dependenciesReflections` similar how its done for `MixinTypes`?

https://github.com/phpstan/phpstan-src/blob/9c4508ded60ee320752f454e0a24029fcfc76908/src/Dependency/DependencyResolver.php#L489-L496

ondrejmirtes:
Yes, definitely! Ideally try to simulate in e2e-tests.yml a situation where the result cache gets stale. There arr several examples already.

PR #2857, opened 2024-01-05
Remove unused DirectClassReflectionExtensionRegistryProvider

staabm:
noticed because I need to adjust `ClassReflectionExtensionRegistryProvider` 

looks like this class is unused?


ondrejmirtes:
Thank you.

PR #2858, opened 2024-01-06
Added UnusedBinaryOperationResultRule

lulco:
Resolves https://github.com/phpstan/phpstan/issues/10389

### DIFF
--- a/conf/config.level4.neon
+++ b/conf/config.level4.neon
@@ -6,6 +6,7 @@ rules:
 	- PHPStan\Rules\DeadCode\UnreachableStatementRule
 	- PHPStan\Rules\DeadCode\UnusedPrivateConstantRule
 	- PHPStan\Rules\DeadCode\UnusedPrivateMethodRule
+	- PHPStan\Rules\DeadCode\UnusedBinaryOperationResultRule
 	- PHPStan\Rules\Exceptions\OverwrittenExitPointByFinallyRule
 	- PHPStan\Rules\Functions\CallToFunctionStatementWithoutSideEffectsRule
 	- PHPStan\Rules\Methods\CallToConstructorStatementWithoutSideEffectsRule
--- a/src/Parser/ParentNodeTypeOfBinaryOperationsVisitor.php
+++ b/src/Parser/ParentNodeTypeOfBinaryOperationsVisitor.php
@@ -1,0 +1,36 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Parser;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\BinaryOp;
+use PhpParser\NodeVisitorAbstract;
+use function array_pop;
+use function count;
+use function get_class;
+
+final class ParentNodeTypeOfBinaryOperationsVisitor extends NodeVisitorAbstract
+{
+
+	public const ATTRIBUTE_NAME = 'parentNodeTypeOfBinaryOperation';
+
+	/** @var array<int, class-string<Node>> */
+	private array $typeStack = [];
+
+	public function beforeTraverse(array $nodes): ?array
+	{
+		$this->typeStack = [];
+		return null;
+	}
+
+	public function enterNode(Node $node): ?Node
+	{
+		if ($node instanceof BinaryOp && count($this->typeStack) > 0) {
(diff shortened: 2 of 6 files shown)


conf/config.level4.neon
```diff
@@ -6,6 +6,7 @@ rules:
 	- PHPStan\Rules\DeadCode\UnreachableStatementRule
 	- PHPStan\Rules\DeadCode\UnusedPrivateConstantRule
 	- PHPStan\Rules\DeadCode\UnusedPrivateMethodRule
+	- PHPStan\Rules\DeadCode\UnusedBinaryOperationResultRule
```
lulco, line 9: added to level 4 to other dead code rules

src/Parser/ParentNodeTypeOfBinaryOperationsVisitor.php
```diff
@@ -0,0 +1,36 @@
...
+use PhpParser\Node\Expr\BinaryOp;
+use PhpParser\NodeVisitorAbstract;
+use function array_pop;
+use function count;
+use function get_class;
+
+final class ParentNodeTypeOfBinaryOperationsVisitor extends NodeVisitorAbstract
+{
+
+	public const ATTRIBUTE_NAME = 'parentNodeTypeOfBinaryOperation';
```
lulco, line 15: I'm not sure about the attribute name

src/Rules/DeadCode/UnusedBinaryOperationResultRule.php
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	/**
+	 * @param BinaryOp $node
+	 */
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if ($node->getAttribute(ParentNodeTypeOfBinaryOperationsVisitor::ATTRIBUTE_NAME) === Expression::class) {
+			return [
+				RuleErrorBuilder::message('Result of binary operation is not used.')
```
lulco, line 31: is this message enough?

tests/PHPStan/Rules/DeadCode/data/unused-binary-operation-result.php
```diff
@@ -0,0 +1,64 @@
...
+
+$foo = new Foo();
+$foo->doInt() + $foo->doInt();
+$foo->doInt() - $foo->doInt();
+$foo->doInt() * $foo->doInt();
+$foo->doInt() ^ $foo->doInt();
+pow($foo->doInt(), $foo->doInt());
+$foo->doInt() - 5 + $foo->doInt();
+'No' . $foo->doString();
+echo 'Yes' . $foo->doString();
```
staabm, line 64: in test-cases we should verify that the rule will not error on function/method calls like

```php
func(5+ 5);
func($foo->doInt() + $foo->doInt());
func('No'. $foo->getString());
```

lulco: function pow is already used, I can add som method calls, but should I add also static method, var_dump, print, condition and other nodes?

staabm:
please have a look at the failling jobs in CI, e.g. 

https://github.com/phpstan/phpstan-src/actions/runs/7433088846/job/20225724445?pr=2858
https://github.com/pmmp/PocketMine-MP/blob/e0b07ff3087b652407439a29c941f3b66ca92c86/src/pocketmine/inventory/CraftingManager.php#L147

is a false positive

staabm:
another false positive here

https://github.com/phpstan/phpstan-src/actions/runs/7433088846/job/20225730624?pr=2858

https://github.com/PrestaShop/PrestaShop/blob/ca9d81aae0bb17f3e767bb5d835348ed144a3ab5/install-dev/controllers/console/process.php#L115
https://github.com/PrestaShop/PrestaShop/blob/ca9d81aae0bb17f3e767bb5d835348ed144a3ab5/install-dev/controllers/http/process.php#L79

and here

https://github.com/phpstan/phpstan-src/actions/runs/7433088846/job/20225730781?pr=2858
https://github.com/phpstan/phpstan/blob/d63096687c708a2089c6b9e4b11988256195483b/e2e/bug8054/public/fare.de/boot.php#L3

ondrejmirtes:
This does not need a new rule since there's already NoopRule for that purpose. Also please see https://github.com/phpstan/phpstan/issues/10389#issuecomment-1879771657 for more context.

PR #2859, opened 2024-01-07
Implement require-extends rules

staabm:


src/Rules/Classes/RequireExtendsRule.php
```diff
@@ -0,0 +1,76 @@
...
+				$type = $extendsTag->getType();
+				$typeName = $type->describe(VerbosityLevel::typeOnly());
+
+				if (in_array($typeName, $parentNames, true)) {
+					continue;
+				}
+
+				$errors[] = RuleErrorBuilder::message(
+					sprintf(
+						'%s requires using class to extend %s, but %s does not.',
```
staabm: error message inspired by psam errors: https://psalm.dev/r/c29b6e8c15

tests/PHPStan/Rules/PhpDoc/IncompatibleRequireExtendsTypeRuleTest.php
```diff
@@ -0,0 +1,62 @@
...
+				'PHPDoc tag @require-extends contains invalid type IncompatibleRequireExtends\SomeEnum.',
+				18,
+			],
+			[
+				'PHPDoc tag @require-extends contains unknown class IncompatibleRequireExtends\TypeDoesNotExist.',
+				23,
+				'Learn more at https://phpstan.org/user-guide/discovering-symbols',
+			],
+			[
+				'PHPDoc tag @require-extends cannot contain generic type.',
```
staabm: this error messge is a bit misleading. my idea was that in `@phpstan-require-extends SomeClass` we only name the class by its name, without generics params (generic classes can still be used at the using-class). 

that way we don't need to validate the generics and therefore safe phpstan from validating them twice.

.. I think we just need a better error message.. wdyt?

ondrejmirtes:
Before I look at the source code, here's what I'd expect these rules to do:

A rule to check the definition of `@phpstan-require-extends`
-----------

* That it's only above an interface or a trait (it's fine to hook onto InClassNode and check that, we don't need to care about cases where it's defined for example above a function).
* That it's only an ObjectType, to reject `@phpstan-require-extends int` for example.
* That it's an existing and non-final class

A rule to check the definition of `@phpstan-require-implements`
-----------

* That it's only above a trait - other uses don't make sense IMHO.
* That it's only an ObjectType
* That it's an existing interface

A rule to check a class that directly implements an interface or directly uses a trait with `@phpstan-require-extends`
-------------

* This class needs to directly or indirectly extend the class from the tag

A rule to check a class that directly uses a trait with `@phpstan-require-implements`
-------------

* This class needs to directly or indirectly implement the interface from the tag

Please confirm the rules do all of these things. I think 4 rule classes as outlined above would be ideal.

staabm:
I have implemented 2 rules for `require-extends`, which I think fullfill your description. if you agree with them I would do the other 2 `require-implements` in a similar fashion.

one thing I noticed: since we use `InClassNode` for `IncompatibleRequireExtendsTypeRule` errors in phpdoc trait definitions will not show up. Should we work with `ClassLike` instead or maybe need a separate rules for checking the phpdoc on traits directly?

src/Rules/PhpDoc/IncompatibleRequireImplementsTypeRule.php
```diff
@@ -0,0 +1,122 @@
...
+		private ClassCaseSensitivityCheck $classCaseSensitivityCheck,
+		private UnresolvableTypeHelper $unresolvableTypeHelper,
+		private bool $checkClassCaseSensitivity,
+	)
+	{
+	}
+
+	public function getNodeType(): string
+	{
+		return ClassLike::class;
```
ondrejmirtes: Ask for InClassNode instead, it contains reflection already. You won't have to get the reflection from ReflectionProvider, and it works for anonymous classes too.

src/Rules/PhpDoc/IncompatibleRequireImplementsTypeRule.php
```diff
@@ -0,0 +1,122 @@
...
+			&& count($implementsTags) > 0
+		) {
+			return [
+				RuleErrorBuilder::message('PHPDoc tag @require-implements is only valid on trait.')->build(),
+			];
+		}
+
+		$errors = [];
+		foreach ($implementsTags as $implementsTag) {
+			$type = $implementsTag->getType();
```
ondrejmirtes: This whole foreach is too complicated. Just check that the type is `ObjectType`.

src/Rules/PhpDoc/IncompatibleRequireImplementsTypeRule.php
```diff
@@ -0,0 +1,122 @@
...
+			if (!$type->canCallMethods()->yes() || !$type->canAccessProperties()->yes()) {
+				$errors[] = RuleErrorBuilder::message(sprintf('PHPDoc tag @require-implements contains non-object type %s.', $type->describe(VerbosityLevel::typeOnly())))->build();
```
ondrejmirtes: `@require-implements` -> `@phpstan-require-implements` everywhere

conf/config.level2.neon
```diff
@@ -64,6 +64,29 @@ services:
 			checkClassCaseSensitivity: %checkClassCaseSensitivity%
 		tags:
 			- phpstan.rules.rule
+
+	-
+		class: PHPStan\Rules\Classes\RequireExtendsRule
+		tags:
+			- phpstan.rules.rule
```
ondrejmirtes: Rules that just have the tag should be in the `rules` section above.

src/Rules/Classes/RequireExtendsRule.php
```diff
@@ -0,0 +1,76 @@
...
+		return InClassNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$classReflection = $node->getClassReflection();
+		$parentNames = $classReflection->getParentClassesNames();
+
+		$errors = [];
+		foreach ($classReflection->getInterfaces() as $interface) {
```
ondrejmirtes: `getImmediateInterfaces`

src/Rules/Classes/RequireExtendsRule.php
```diff
@@ -0,0 +1,76 @@
...
+			$extendsTags = $interface->getRequireExtendsTags();
+			foreach ($extendsTags as $extendsTag) {
+				$type = $extendsTag->getType();
+				$typeName = $type->describe(VerbosityLevel::typeOnly());
+
+				if (in_array($typeName, $parentNames, true)) {
```
ondrejmirtes: You can't really compare type description with a class name, that's going to 💣

src/Rules/Classes/RequireExtendsRule.php
```diff
@@ -0,0 +1,76 @@
...
+			}
+		}
+
+		foreach ($classReflection->getTraits() as $trait) {
+			$extendsTags = $trait->getRequireExtendsTags();
+			foreach ($extendsTags as $extendsTag) {
+				$type = $extendsTag->getType();
+				$typeName = $type->describe(VerbosityLevel::typeOnly());
+
+				if (in_array($typeName, $parentNames, true)) {
```
ondrejmirtes: dtto

src/Rules/Classes/RequireImplementsRule.php
```diff
@@ -0,0 +1,57 @@
...
+		$interfaceNames = array_map(static fn (ClassReflection $interface): string => $interface->getName(), $classReflection->getInterfaces());
+
+		$errors = [];
+		foreach ($classReflection->getTraits() as $trait) {
+			$implementsTags = $trait->getRequireImplementsTags();
+			foreach ($implementsTags as $implementsTag) {
+				$type = $implementsTag->getType();
+				$typeName = $type->describe(VerbosityLevel::typeOnly());
+
+				if (in_array($typeName, $interfaceNames, true)) {
```
ondrejmirtes: There should be an easier way to write this. What about `$classReflection->isSubclassOf()`?

src/Rules/PhpDoc/IncompatibleRequireExtendsTypeRule.php
```diff
@@ -0,0 +1,125 @@
...
+		) {
+			return [];
+		}
+
+		$classReflection = $this->reflectionProvider->getClass($node->namespacedName->toString());
+		$extendsTags = $classReflection->getRequireExtendsTags();
+
+		if (
+			!$classReflection->isTrait()
+			&& ! $classReflection->isInterface()
```
ondrejmirtes: No space after `!`

src/Rules/PhpDoc/IncompatibleRequireExtendsTypeRule.php
```diff
@@ -0,0 +1,125 @@
...
+			&& count($extendsTags) > 0
+		) {
+			return [
+				RuleErrorBuilder::message('PHPDoc tag @require-extends is only valid on trait or interface.')->build(),
+			];
+		}
+
+		$errors = [];
+		foreach ($extendsTags as $extendsTag) {
+			$type = $extendsTag->getType();
```
ondrejmirtes: Again, this whole section should just ask about `$type instanceof ObjectType`.

ondrejmirtes (changes requested):

src/Rules/PhpDoc/IncompatibleRequireImplementsTypeRule.php
staabm: but it doesn't work for trait defintions. it would not error on 

```
/**
 * @phpstan-require-implements SomeTrait
 */
trait InvalidTrait1 {}
```

ondrejmirtes: Two separate rules then please.

ondrejmirtes: For traits there's InTraitNode.

staabm: `InTraitNode` will only error for traits when a class exists which uses them. are you fine with that?

src/Rules/PhpDoc/IncompatibleRequireExtendsTypeTraitRule.php
```diff
@@ -0,0 +1,73 @@
...
+	public function __construct(
+		private ClassCaseSensitivityCheck $classCaseSensitivityCheck,
+		private bool $checkClassCaseSensitivity,
+	)
+	{
+	}
+
+	public function getNodeType(): string
+	{
+		return InTraitNode::class;
```
ondrejmirtes: You're right, this one is better served by `Node\Stmt\Trait_` - even when the trait is used multiple times, we just want to report the declaration once.

ondrejmirtes (changes requested):

ondrejmirtes (changes requested):
Hey, I just pushed some cosmetic changes.

What still needs to be done here:

1) Support `@psalm-` prefixes in PhpDocNodeResolver along with some tests: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-require-extends
2) There is a lot of literal duplication between RequireExtendsDefinitionTraitRule and RequireExtendsDefinitionClassRule. We usually solve this by introducing a "check" class that gets injected into rule classes to reuse the common code.

Also please note that after we merge the current two PRs and release them, I'd like you to work on properly introducing generics support for the new tags. But I want that to happen later and to not complicate the current PRs. Thanks.

tests/PHPStan/Rules/Classes/RequireImplementsRuleTest.php
```diff
@@ -56,6 +56,18 @@ public function testRule(): void
 				'Trait IncompatibleRequireImplements\InvalidTrait4 requires using class to implement IncompatibleRequireImplements\SomeClass<T>, but IncompatibleRequireImplements\InvalidTraitUse4 does not.',
 				137,
 			],
+			[
+				'Trait IncompatibleRequireImplements\ValidPsalmTrait requires using class to implement IncompatibleRequireImplements\RequiredInterface2, but AnonymousClass4fea61018e98554a605d5c87523a32cc does not.',
```
ondrejmirtes: this message is wrong, you need to use getDisplayName on ClassReflection

ondrejmirtes (changes requested):

ondrejmirtes:
Please use it consistently everywhere on ClassReflection instead of getName when building user-facing messages, thanks.

staabm:
sorry missed one occurence

src/Rules/PhpDoc/RequireExtendsDefinitionClassRule.php
```diff
@@ -0,0 +1,47 @@
...
+	{
+		return InClassNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$classReflection = $node->getClassReflection();
+		$extendsTags = $classReflection->getRequireExtendsTags();
+
+		if (count($extendsTags) === 0) {
```
staabm, line 34: I think we should error when a class defines 2 or more `@*-require-extends`

ondrejmirtes: Not error, but support and check all of them.

ondrejmirtes: Oh sorry, I misunderstood you. Please add a test and fix the rule.

src/Rules/PhpDoc/RequireExtendsCheck.php
```diff
@@ -30,6 +31,11 @@ public function __construct(
 	public function checkExtendsTags(Node $node, array $extendsTags): array
 	{
 		$errors = [];
+
+		if (count($extendsTags) > 1) {
+			$errors[] = RuleErrorBuilder::message(sprintf('PHPDoc tag @phpstan-require-extends can only be used once.'))->build();
```
ondrejmirtes, line 36: Oh right, this makes sense because you can't have multiple parents :)

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2860, opened 2024-01-07
optimize repeated container creation in tests

schlndh:
I looked into the performance of creating containers in tests and I found out that creating SourceLocators in `ComposerJsonAndInstalledJsonSourceLocatorMaker` is by far the most expensive part (on my machine about 1.1s out of 1.4s for the whole uncached `getContainer`). So here I try to cache it statically for the tests. As far as I can tell, it only depends on a few parameters (at least for now), which seem likely to be the same for most of the tests.

Quick benchmark with `rm -rf /tmp/phpstan-tests/; time make tests` gives me about 1m 22s with this change vs about 1m 46s without it.

Motivation: This is a preliminary step before I attempt to make it possible for individual test methods to use their own config for the container, instead of just relying on `getAdditionalConfigFiles` which is static. That would be useful for https://github.com/phpstan/phpstan-src/pull/2809 and possibly other things as well.

### DIFF
--- a/src/Testing/TestCaseSourceLocatorFactory.php
+++ b/src/Testing/TestCaseSourceLocatorFactory.php
@@ -12,6 +12,7 @@ use PHPStan\BetterReflection\SourceLocator\Type\EvaledCodeSourceLocator;
 use PHPStan\BetterReflection\SourceLocator\Type\MemoizingSourceLocator;
 use PHPStan\BetterReflection\SourceLocator\Type\PhpInternalSourceLocator;
 use PHPStan\BetterReflection\SourceLocator\Type\SourceLocator;
+use PHPStan\Php\PhpVersion;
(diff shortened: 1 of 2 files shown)


src/Testing/TestCaseSourceLocatorFactory.php
```diff
@@ -52,10 +73,11 @@ public function create(): SourceLocator
 				if ($composerSourceLocator === null) {
 					continue;
 				}
-				$locators[] = $composerSourceLocator;
+				self::$composerSourceLocatorsCache[$cacheKey][] = $composerSourceLocator;
```
ondrejmirtes: One tiny detail - please first create a local variable and only after the foreach is done, assign it to the static property.

With the current approach if the 3rd iteration fails, the property still gets populated but next time it will just return two locators.

ondrejmirtes (changes requested):

ondrejmirtes:
Awesome, thank you!

PR #2861, opened 2024-01-10
require-extends should not error on interfaces

staabm:
There are two issues:

1) Interface should not complain about not extending a class
2) When the interface is being implemented by the class required, the rule should not complain

Both are reproduced here: 
https://phpstan.org/r/25925129-a872-4300-8437-a8edd6dc825a

### DIFF
--- a/src/Rules/Classes/RequireExtendsRule.php
+++ b/src/Rules/Classes/RequireExtendsRule.php
@@ -26,16 +26,16 @@ class RequireExtendsRule implements Rule
 	{
 		$classReflection = $node->getClassReflection();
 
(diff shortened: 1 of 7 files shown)


src/Rules/Classes/RequireExtendsRule.php
```diff
@@ -26,16 +26,20 @@ public function processNode(Node $node, Scope $scope): array
 	{
 		$classReflection = $node->getClassReflection();
 
+		if ($classReflection->isInterface()) {
```
staabm, line 29: I think we should also skip abstract classes

ondrejmirtes:
Thank you.

PR #2862, opened 2024-01-10
Fix transliterator function return types

PrinsFrank:
This PR fixes the incorrect return types for ext-intl transliterator functions

source:
https://www.php.net/manual/en/transliterator.createinverse.php
https://www.php.net/manual/en/transliterator.geterrorcode.php
https://www.php.net/manual/en/transliterator.geterrormessage.php
https://www.php.net/manual/en/transliterator.listids.php


ondrejmirtes:
Thank you.

PR #2866, opened 2024-01-11
Fix stale result cache for require-extends/require-implements

staabm:
refs https://github.com/phpstan/phpstan-src/pull/2856#issuecomment-1887220761


.github/workflows/e2e-tests.yml
```diff
@@ -114,6 +114,26 @@ jobs:
...
+              echo -n > phpstan-baseline.neon
+              ../../bin/phpstan -vvv
+              patch -b src/Baz.php < patch-1.patch
+              cat baseline-1.neon > phpstan-baseline.neon
+              ../../bin/phpstan -vvv
+              mv src/Baz.php.orig src/Baz.php
+              echo -n > phpstan-baseline.neon
+              ../../bin/phpstan -vvv
+          - script: |
+              cd e2e/result-cache-8
```
ondrejmirtes: Where did 7 go? :)

ondrejmirtes (changes requested):

.github/workflows/e2e-tests.yml
staabm: I dropped it after I realized I need just one test for each annotation :-)

.github/workflows/e2e-tests.yml
```diff
@@ -114,6 +114,26 @@ jobs:
...
+              echo -n > phpstan-baseline.neon
+              ../../bin/phpstan -vvv
+              patch -b src/Baz.php < patch-1.patch
+              cat baseline-1.neon > phpstan-baseline.neon
+              ../../bin/phpstan -vvv
+              mv src/Baz.php.orig src/Baz.php
+              echo -n > phpstan-baseline.neon
+              ../../bin/phpstan -vvv
```
ondrejmirtes, line 136: I tried `result-cache-8` after removing changes from `src/` and it does not fail for me locally. Please first submit a draft PR with just the E2E tests so it's obvious they fail.

ondrejmirtes (changes requested):

.github/workflows/e2e-tests.yml
staabm: after having slept about it, I think the situation is as follows:

we only need to support `require-extends` for result-cache invalidation because thats the only new part in the [reflection hierarchy](https://github.com/phpstan/phpstan-src/pull/2856/files#diff-9007a149012fc95c1311cb1346c8e75f59786c0aafb5390f2e56d28f0bc2ff8a). 

`require-implements` works with reflection information which was available before, we just added some rules arround it.

[see failling test](https://github.com/phpstan/phpstan-src/actions/runs/7499547666/job/20416549891?pr=2866)
[and the fix](https://github.com/phpstan/phpstan-src/actions/runs/7499761471/job/20417169014?pr=2866)

ondrejmirtes (changes requested):
I just figured out when `@phpstan-require-implements` has to be handled in DependencyResolver.

```php
/** @phpstan-require-implements Bar */
trait Foo
{
}
```

When the file with `Bar` changes, we need to reanalyse trait `Foo`, because the RequireImplementsRule might find some new error, or get rid of an existing error. Like for example when `Foo` changes type from interface to class and so on.

Please write a failing E2E test number 7 with this use-case, make sure it fails, then fix it in DependencyResolver. Thanks :)

staabm:
Thank you, I will have a look until monday

staabm:
took me a few attempts.. :-/.

failling test in https://github.com/phpstan/phpstan-src/actions/runs/7518819436/job/20466608449?pr=2866

I think its ready for another round of review

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2868, opened 2024-01-12
Simplify default-return path in Extensions

staabm:
similar to https://github.com/phpstan/phpstan-src/pull/2782

Not sure why my previous search pattern did not match these. 

I think there is no more occurrences left which can be simplified in a similar fashion


ondrejmirtes:
Conflict here.

ondrejmirtes:
Thank you.

PR #2869, opened 2024-01-12
strtok() always returns a non-empty-string when it does not return false

staabm:
inspired by https://github.com/vimeo/psalm/pull/10497

analog https://phpstan.org/r/795bc4fc-d3ab-4833-a75a-ba3ae9897aaa


ondrejmirtes:
Thank you.

PR #2870, opened 2024-01-14
Faster NullableType resolving

staabm:
running phpstan on `microsoft/microsoft-graph` shows a bottleneck on `TypeCombinator::addNull` in `ParserNodeTypeToPHPStanType->resolve`.

grafik

since we are in a very low level method we can just return `new UnionType` and prevent calls to [`isSuperTypeOf` and `union`](https://github.com/phpstan/phpstan-src/blob/44f6b0c8eb5325d31e67cd18c00198f9cc1814b4/src/Type/TypeCombinator.php#L42-L51) which shaves of a few seconds


ondrejmirtes:
You can't do this for some types, for example `?mixed`. Yeah the type is wrong but it'd break something down the line in PHPStan.

ondrejmirtes:
Looks like the method is called a lot of times. The number of times should correspond to the number of functions/methods inside the project. Maybe the type could be cached in reflection layer in more places.

staabm:
> Looks like the method is called a lot of times. The number of times should correspond to the number of functions/methods inside the project.

the method is called once per parameter and once per return type. so it is called way more often then methods/functions contained in the project beeing scanned.

ondrejmirtes:
Yeah that's what I meant of course, but we should verify it's not called more often (I suspect it is, definitely for closures and arrow functions, maybe even normal functions and methods).

PR #2871, opened 2024-01-14
Faster NullableType resolving

staabm:
similar to https://github.com/phpstan/phpstan-src/blob/7edb6fc21776dd2105032103bd689c92f61fdf7d/src/PhpDoc/TypeNodeResolver.php#L525

origin investigation in https://github.com/phpstan/phpstan-src/pull/2870#issue-2080665602


ondrejmirtes:
How faster really is this? From the looks of it it doesn't look like an optimization at all. I'd rather investigate https://github.com/phpstan/phpstan-src/pull/2870#issuecomment-1890970421

PR #2872, opened 2024-01-14
Reduce unnecessary calls to getFunctionType()

staabm:
another take on https://github.com/phpstan/phpstan-src/pull/2870


ondrejmirtes:
Thank you.

PR #2873, opened 2024-01-14
Simplify arrow-function return type inference

staabm:
Arrow-functions can't have a void return type

refs https://github.com/phpstan/phpstan-src/pull/2778#discussion_r1451778501

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -1242,7 +1242,7 @@ class MutatingScope implements Scope
 						new VoidType(),
 					]);
 				} else {
-					$returnType = $arrowScope->getKeepVoidType($node->expr);
+					$returnType = $arrowScope->getType($node->expr);
 					if ($node->returnType !== null) {
 						$returnType = TypehintHelper::decideType($this->getFunctionType($node->returnType, false, false), $returnType);
 					}


src/Analyser/MutatingScope.php
```diff
@@ -1242,7 +1242,7 @@ private function resolveType(string $exprString, Expr $node): Type
 						new VoidType(),
 					]);
 				} else {
-					$returnType = $arrowScope->getKeepVoidType($node->expr);
+					$returnType = $arrowScope->getType($node->expr);
 					if ($node->returnType !== null) {
 						$returnType = TypehintHelper::decideType($this->getFunctionType($node->returnType, false, false), $returnType);
```
staabm, line 1247: I came to the conclusion that `getKeepVoidType` might not be necessary because the following IF did not keep void when infered from `$node->returnType` either

staabm:
//cc @herndlm

ondrejmirtes:
If no tests were failing before, and this doesn't fix anything, what is the purpose of this change?

staabm:
Its a line of code which did not make sense to me while looking thru functionType stuff.

I have no concrete case in mind - feel free to close

ondrejmirtes:
If it ain't broke don't fix it :)

But when testing a piece of code I found a false negative on 1.10.x:

* https://phpstan.org/r/d7783f5d-73c9-4d04-b65a-7979c165ad21
* https://3v4l.org/6ddr0

PR #2875, opened 2024-01-15
Add composer-dependency-analyser to CI (fix shadow & misplaced dependencies)

janedbal:
- Documentation: https://github.com/shipmonk-rnd/composer-dependency-analyser
- **Unused dependencies** are ignored for now, it is more tricky (will try in next PR)
- **Misplaced dependency** was only [nette/finder](https://github.com/phpstan/phpstan-src/pull/2875#discussion_r1486293822) and [jetbrains/phpstorm-stubs](https://github.com/phpstan/phpstan-src/pull/2875#discussion_r1491192795) (unused in src, used only in tests)
- See **shadow dependencies** report before the fix:

image


.github/workflows/lint.yml
```diff
@@ -127,3 +127,26 @@ jobs:
 
       - name: "Name Collision Detector"
         run: "make name-collision"
+
+
+  composer-dependency-analysis:
```
janedbal: I kept `composer-require-checker` even though `composer-dependency-analyser` found more real shadow dependencies than `composer-require-checker` in every package I tried.

build/composer-dependency-analyser.php
```diff
@@ -0,0 +1,15 @@
...
+
+use ShipMonk\ComposerDependencyAnalyser\Config\Configuration;
+use ShipMonk\ComposerDependencyAnalyser\Config\ErrorType;
+
+$config = new Configuration();
+
+return $config
+	->addPathToScan(__DIR__ . '/../bin', true)
+	->ignoreErrors([ErrorType::UNUSED_DEPENDENCY])
+	->ignoreErrorsOnPackage('phpunit/phpunit', [ErrorType::DEV_DEPENDENCY_IN_PROD]) // prepared test tooling
```
janedbal, line 11: All violations are within `src/Testing`, would you like to ignore this error only there?

*(that might force people to place all PHPUnit-related stuff there)*

janedbal:
@ondrejmirtes Any comments on this one? I believe it found real issues and thus make sense to keep it in CI to guard that. Also, the package is stable now.

ondrejmirtes:
I agree, we could remove composer-require-checker if this one does a better job. But my priority now is to finish and release 1.11 (and I'm travelling to PHP UK this werk) so I'll most likely start processing the backlog of emails (currently 130) after that.

.github/workflows/lint.yml
janedbal: Removed require checker as per https://github.com/phpstan/phpstan-src/pull/2875#issuecomment-1938516595

build/composer-dependency-analyser.php
janedbal: Personally, I'd keep it as is. Resolving thread.

composer.json
```diff
@@ -13,11 +13,11 @@
 		"fidry/cpu-core-counter": "^0.5.0",
 		"hoa/compiler": "3.17.08.08",
 		"hoa/exception": "^1.0",
+		"hoa/file": "1.17.07.11",
 		"hoa/regex": "1.17.01.13",
 		"jetbrains/phpstorm-stubs": "dev-master#8ea3bf44722f76a3a32c89fe769b6144c253f8a6",
 		"nette/bootstrap": "^3.0",
 		"nette/di": "3.1.10",
-		"nette/finder": "^2.5",
```
janedbal, line 20: Used only in tests with single usage:
- `tests/PHPStan/Rules/Properties/data/efabrica-latte-bug.php`

So I moved it to require-dev.

janedbal:
@ondrejmirtes Ok, I removed composer-require-checker, updated the tool and fixed [one more issue](https://github.com/phpstan/phpstan-src/pull/2875#discussion_r1486293822). It is also rebased so it could be merged now if you feel so.

composer.json
```diff
@@ -53,13 +53,16 @@
 	"require-dev": {
 		"brianium/paratest": "^6.5",
 		"cweagans/composer-patches": "^1.7.3",
+		"jetbrains/phpstorm-stubs": "dev-master#8ea3bf44722f76a3a32c89fe769b6144c253f8a6",
```
janedbal: Latest version found also this miscategorization (moved from require to require-dev). Looks valid to me as it is used only in `ReflectionProviderGoldenTest`, no usage in src.

(but there is string reference in `compiler/src/Console/PrepareCommand.php`, might be worth checking)

composer.json
ondrejmirtes: The tool is wrong here. PHPStan depends on ondrejmirtes/BetterReflection which depends on jetbrains/phpstorm-stubs in `require`. So `jetbrains/phpstorm-stubs` is part of production dependencies of PHPStan.

The line here in `require` in `composer.json` is so that PHPStan gets this newer commit from dev-master as opposed to the one ondrejmirtes/BetterReflection depends on.

ondrejmirtes (changes requested):

composer.json
janedbal: I see, fixed (moved back where it was).

But such intention cannot be ever detected by any rule, so I ignored it with your explanation there.

Thank you

composer.json
janedbal: Reverted. [Reason](https://github.com/phpstan/phpstan-src/pull/2875#discussion_r1495886870)

composer.json
janedbal: Squashed and rebased, should be mergeable again.

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "bd93faf2a75150f9ba8e41fb86b8c632",
+    "content-hash": "2723c5a59596e6ccf2430f75a1fcf36b",
```
ondrejmirtes, line 7: What command did you please use to update composer.lock here? I need it so I can replay the changes on 1.11.x

ondrejmirtes:
Thank you very much!

composer.lock
janedbal: `composer update --lock`, see [docs](https://getcomposer.org/doc/03-cli.md#update-u-upgrade)

ondrejmirtes: Thanks!

ondrejmirtes: Just realized this advice is wrong.

Example:

1) 1.10.x updates `phpstan/phpdoc-parser` from 1.25.0 to 1.26.0.
2) I merge 1.10.x into 1.11.x and get conflict in composer.lock
3) I run `git checkout --ours composer.lock` to get it look like on top of 1.11.x.

Now I need to rerun the commands that make the lock file being changed in 1.10.x. Running `composer update --lock` is not sufficient, it's going to keep phpdoc-parser on version 1.25.0 in composer.lock. But I want it to be 1.26.0.

The correct solution is to run `composer update phpdoc-parser`.

janedbal: My working flow:
- I merge all non-conflicting stuff of `composer.lock` in PHPStorm merge tool.
- The only conflict is the hash, I pick any.
   - _If not, some package got update on both sides._
- I fix that hash by `composer update --lock`

ondrejmirtes: I see!

PR #2878, opened 2024-01-17
Remove dead cache.nodesByFileCountMax

janedbal:
- obviously, this might break for those who use it


ondrejmirtes:
I have this noted for 2.0, don't worry :)

PR #2879, opened 2024-01-17
CachedParser: use LRU cache

janedbal:


src/Parser/LRUCache.php
```diff
@@ -0,0 +1,77 @@
...
+	{
+		if (!isset($this->cache[$key])) {
+			throw new LogicException(sprintf('Key %s was not found in the cache, use ->has() first', $key));
+		}
+
+		$this->list->rewind();
+		while ($this->list->current() !== $key) {
+			$this->list->next();
+		}
+		$this->list->rewind();
```
staabm: I wonder why you rewind after the loop? Doesn't it make the previous loop useless?

staabm:
> But the time difference is unmeasurable here. Who knows if some codebase with complex huge files might cause measurable differences or not.

did you compare before and after with the lowered capacity?
Did you have a look at memory consumption of both impls?

janedbal:
> Did you habe a look at memory consumption of both impls?

Only by seeing what PHPStan reports in -vvv mode. No changes.

> Did you compare before and after with the lowered capacity?

PHPStan itself uses lowered capacity (128 instead of default 256) and that is what I tested (results above). I did not do any tests on our huge codebase yet.

src/Parser/LRUCache.php
janedbal: No idea how this ended up here, definitely doesnt make sense :D

staabm:
Yeah… my point was more about: in case we can reduce the capacity of the cache with this PR, it might be useful because even if it is equal fast, it might save memory

src/Parser/LRUCache.php
janedbal: So the results are useless, this needs proper tests and rerunning the analysis. I'm surprised the cache-misses improved with such error here. Thx

src/Parser/LRUCache.php
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @template TValue
+ * @implements ParserCache<TValue>
+ */
+class LRUCache implements ParserCache
+{
+
+	/** @var array<string, TValue> */
+	private array $cache;
```
staabm, line 19: it might make sense to try SplFixedArray here

staabm: Hmm our capacity is too small for it beeing useful

https://stackoverflow.com/a/11831076

janedbal:
After all, it seems like time & cache-miss & memory difference is very small. 

----------

So the only benefit of this MR might be the code readability & test improvement. I can replace the LRUCache with old logic. Would you merge it after that @ondrejmirtes? Or just close it.

ondrejmirtes:
Yeah, I don't think this fixes a real problem :)

PR #2880, opened 2024-01-18
Update `Locale` signatures

devnix:
I'm not sure about documenting all optional parameters of `Locale::composeLocale()` as it has no effect when introducing non-documented parameters to the function, but I've reviewed all the functions, even the undocumented ones and AFAIK these are the correct signatures 😊


devnix:
Not sure why this is failing: https://github.com/phpstan/phpstan-src/actions/runs/7568063545/job/20608638825?pr=2880

resources/functionMap.php
```diff
@@ -5953,41 +5953,41 @@
...
-'Locale::getDisplayName' => ['string', 'locale'=>'string', 'in_locale='=>'string'],
-'Locale::getDisplayRegion' => ['string', 'locale'=>'string', 'in_locale='=>'string'],
+'Locale::acceptFromHttp' => ['non-empty-string|false', 'header'=>'string'],
+'Locale::canonicalize' => ['non-empty-string|null', 'locale'=>'string'],
+'Locale::composeLocale' => ['string|false', 'subtags'=>'array{language:string, script?:string, region?:string, variant?:array<array-key,string>, private?:array<array-key,string>, extlang?:array<array-key,string>, variant0?:string, variant1?:string, variant2?:string, variant3?:string, variant4?:string, variant5?:string, variant6?:string, variant7?:string, variant8?:string, variant9?:string, variant10?:string, variant11?:string, variant12?:string, variant13?:string, variant14?:string,  private0?:string, private1?:string, private2?:string, private3?:string, private4?:string, private5?:string, private6?:string, private7?:string, private8?:string, private9?:string, private10?:string, private11?:string, private12?:string, private13?:string, private14?:string, extlang0?:string, extlang1?:string, extlang2?:string}'],
```
ondrejmirtes: Narrowing down parameter types is a job for this file instead https://github.com/phpstan/phpstan-src/blob/1.11.x/resources/functionMap_bleedingEdge.php

Because it's essentially a BC break. You're forcing someone that calls this with a general `array` to update typehints in their whole call chain.

ondrejmirtes (changes requested):

ondrejmirtes:
That build failure is not the fault of this PR.

ondrejmirtes:
Also please get rid of the merge commit, the history should be linear (use rebase instead).

resources/functionMap.php
devnix: In my tests, passing more parameters than the documented ones didn't reported anything, but I can move that to the bleeding edge 😄

**Edit**: Sorry, you're probably referring to parameters like `non-empty-string`, consider it done!

devnix:
I think this should do it 😄

resources/functionMap_bleedingEdge.php
```diff
@@ -125,6 +125,22 @@
...
 		'extract' => ['0|positive-int', '&rw_var_array'=>'array', 'extract_type='=>'EXTR_OVERWRITE|EXTR_SKIP|EXTR_PREFIX_SAME|EXTR_PREFIX_ALL|EXTR_PREFIX_INVALID|EXTR_IF_EXISTS|EXTR_PREFIX_IF_EXISTS|EXTR_REFS', 'prefix='=>'string|null'],
 		'RecursiveIteratorIterator::__construct' => ['void', 'iterator'=>'RecursiveIterator|IteratorAggregate', 'mode='=>'RecursiveIteratorIterator::LEAVES_ONLY|RecursiveIteratorIterator::SELF_FIRST|RecursiveIteratorIterator::CHILD_FIRST', 'flags='=>'0|RecursiveIteratorIterator::CATCH_GET_CHILD'],
+		'Locale::acceptFromHttp' => ['non-empty-string|false', 'header'=>'string'],
```
ondrejmirtes: If you make return types more specific it can be put into functionMap.php, not bleeding edge. Only more specific parameter types need to be put into bleeding edge.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

lazka:
Did this fix https://github.com/phpstan/phpstan/issues/7855 ?

PR #2881, opened 2024-01-18
Verify json serializability of collector data in single process run

staabm:
refs https://github.com/phpstan/phpstan/discussions/10453#discussioncomment-8166904

### DIFF
--- a/src/Analyser/Analyser.php
+++ b/src/Analyser/Analyser.php
@@ -5,11 +5,14 @@ namespace PHPStan\Analyser;
 use Closure;
 use PHPStan\Collectors\CollectedData;
 use PHPStan\Collectors\Registry as CollectorRegistry;
+use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\Registry as RuleRegistry;
 use Throwable;
 use function array_fill_keys;
 use function array_merge;
 use function count;
+use function json_decode;
+use function json_encode;
 use function memory_get_peak_usage;
 use function sprintf;
 
@@ -22,6 +25,7 @@ class Analyser
 		private CollectorRegistry $collectorRegistry,
 		private NodeScopeResolver $nodeScopeResolver,
 		private int $internalErrorsCountLimit,
+		private ReflectionProvider $reflectionProvider,
 	)
 	{
 	}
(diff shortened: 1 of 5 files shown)


src/Analyser/Analyser.php
```diff
@@ -104,6 +109,42 @@ public function analyse(
...
+			}
+
+			$restored = json_decode($json);
+			if ($restored === $collected->getData()) {
+				continue;
+			}
+
+			$errors[] = $this->buildCollectedDataError(
+				sprintf(
+					'Data collected by %s is not JSON decodable and will not work in parallel analysis.',
```
staabm, line 140: I think we need to be more precise here. I could not come up with a better error message.

the most important point is, that a datastructure needs to fulfill `json_decode(json_encode($data)) === $data`

mad-briller: perhaps a better term would be JSON serializable because that's the name of the [interface](https://www.php.net/manual/en/class.jsonserializable.php) php uses to allow encoding an object as json?

ondrejmirtes (changes requested):
What I'd like here too is a new PHPStan rule in Api namespace that would check this too.

Best way is a new method on Type - isJsonEncodable.

src/Analyser/Analyser.php
```diff
@@ -104,6 +109,42 @@ public function analyse(
...
+					),
+					$collected,
+				);
+				$alreadyReported[$collected->getCollectorType()] = true;
+
+				break;
+			}
+
+			$restored = json_decode($json);
+			if ($restored === $collected->getData()) {
```
mad-briller, line 134: thinking about this, is this actually the correct check?

if there is an object that implements JsonSerializable returned from a collector it will encode fine, but when decoded it will be an `stdClass` meaning the `===` check would fail

if it's intended that you can use JsonSerializable in collectors then the check needs to be more robust

staabm: you describe exactly the case, why this rule was created. the behaviour of phpstan is different depending on whether phpstan is running with parallel worker or with a single worker (e.g. in `--debug`), since the collected data is only json encoded in the the worker and decoded in the main thread in parallel, which means when you put a `JsonSerializable` object into collected data it will work differently depending on the worker scenario.

in single worker mode the collector data is passed thru the php process 1:1 and does not get influenced by json encode/decode

mad-briller: ah my misunderstanding, when Ondrej said 

> This type would not be precise. Anything that is JSON-serializable works.

i thought the intention was that JsonSerializable should be permitted, thanks for clearing that up

staabm:
> What I'd like here too is a new PHPStan rule in Api namespace that would check this too.

@ondrejmirtes would the new rule check the generic template type of `TValue` of each collector? or do you have something different in mind?


——-

> Best way is a new method on Type - isJsonEncodable.

we need two way json encode/decodability. Encode-ability alone - as we have it before this PR - only works in single worker mode, which actually means JsonSerializable objects are hard to work with in real world use-cases.

see https://github.com/phpstan/phpstan-src/pull/2881#discussion_r1457699423

PR #2882, opened 2024-01-19
curl_multi_init() cannot return false

staabm:
even in php 7.4 it could not return false. it used to return `resource` and this changed with PHP8 to `CurlMultiHandle`:
https://www.php.net/manual/en/function.curl-multi-init.php

https://phpstan.org/r/3d26ccc3-01a7-424c-8dee-f5c4cac8db9d

https://github.com/php/php-src/blob/a0f6c5474ecba5693c5b5e45ffe802f375a8c0e4/ext/curl/multi.c#L62-L69


ondrejmirtes:
Thank you!

PR #2884, opened 2024-01-21
error_log() is impure

staabm:
closes https://github.com/phpstan/phpstan/issues/10437

### DIFF
--- a/tests/PHPStan/Analyser/NodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/NodeScopeResolverTest.php
@@ -1423,6 +1423,7 @@ class NodeScopeResolverTest extends TypeInferenceTestCase
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10302-interface-extends.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10302-trait-extends.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10302-trait-implements.php');
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/impure-error-log.php');
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -1423,6 +1423,7 @@ public function dataFileAsserts(): iterable
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10302-interface-extends.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10302-trait-extends.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10302-trait-implements.php');
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/impure-error-log.php');
```
ondrejmirtes: Please don't put at the end, but somewhere random in the middle. This is to prevent conflicts when merging into 1.11.x

tests/PHPStan/Analyser/data/impure-error-log.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace ImpureErrorLog;
+
+use function PHPStan\Testing\assertType;
+
+$message = 'foo';
+$logfile = 'bar/baz.txt';
+if (!error_log($message, 3, $logfile)) {
+	assertType('bool', error_log($message, 3, $logfile));
+}
```
ondrejmirtes, line 13: Please always put the code you're testing into an anonymous function. The root scope behaves a little bit differently (see `Scope::canAnyVariableExist()`) and when we are not testing something specific to the root scope I like to avoid issues with that.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/impure-error-log.php
staabm: right, forgot about that. thanks for the headsup.

ondrejmirtes:
Thank you.

PR #2886, opened 2024-01-21
fix spaceship operator for large constant unions

schlndh:
This PR fixes a false positive encountered by https://github.com/phpstan/phpstan-src/pull/2885 . I simplified it to smaller example: https://phpstan.org/r/892eed14-49b8-4a98-b591-90cf5741a99e

This scenario only happens with bleeding edge enabled. Without it the constant union doesn't make it into `getSpaceshipType` due to https://github.com/phpstan/phpstan-src/blob/4e1bfacafb0cd4a4dd89946d211205fb0ba9cef5/src/Type/Generic/TemplateTypeTrait.php#L269

As far as I can tell, the spaceship operator should always return -1, 0, or 1: https://3v4l.org/FVeq7


ondrejmirtes:
Thank you!

PR #2890, opened 2024-01-27
InvalidThrowsPhpDocValueRule: support @require-extends, @require-implements

RobertMe:
Handle / support `@throws` types which are annotated with `@phpstan-require-extends` or `@phpstan-require-implements`.

Fixes phpstan/phpstan#10475

Currently this is a WIP as support for intersections and unions is lacking and this is more like a "support request" on how I can support this.

### DIFF
--- a/src/Rules/PhpDoc/InvalidThrowsPhpDocValueRule.php
+++ b/src/Rules/PhpDoc/InvalidThrowsPhpDocValueRule.php
@@ -9,6 +9,8 @@ use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Type\FileTypeMapper;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
 use Throwable;
 use function sprintf;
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Fixed the implementation, see my commits.

ondrejmirtes:
Thank you.

RobertMe:
Thank you. I more or less had the same in mind for handling unions (check of `UnionType`, iterate over its types and make a recursive call returning `false` if one is invalid / `true` if their all valid). But wanted to do to same for intersections (but obviously being `true` if any is valid, instead of all). And then "falling back" to what I had in case it was neither of those. But the resolved types (from the annotation) can indeed be combined into an intersection with the given type, and then doing the `isSuperTypeOf` check on that, dropping the need for an `IntersectionType` specific path.

staabm:
does this PR also needs support in `DependencyResolver`, so types referenced via `require-extends`/`require-implements` from a throw-type will invalidate the result cache properly?

(similar how it was done with https://github.com/phpstan/phpstan-src/pull/2866)

ondrejmirtes:
I don't know. Maybe when the class in require-extends starts/stops being a Throwable we get a stale cache.

You can always verify your hypothesis with a failing test 😊

PR #2891, opened 2024-01-27
`sort()`, `rsort()` and `usort()` converts an array to list

takaram:
Closes phpstan/phpstan#6467
Closes phpstan/phpstan#10423
Closes phpstan/phpstan#3312

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -129,6 +129,7 @@ use PHPStan\Reflection\SignatureMap\SignatureMapProvider;
 use PHPStan\Rules\Properties\ReadWritePropertiesExtensionProvider;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
+use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\ClosureType;
@@ -2130,6 +2131,21 @@ class NodeScopeResolver
 
 			if (
 				$functionReflection !== null
+				&& (
+					(in_array($functionReflection->getName(), ['sort', 'rsort'], true) && count($expr->getArgs()) >= 1)
+					|| ($functionReflection->getName() === 'usort' && count($expr->getArgs()) >= 2)
+				)
+			) {
+				$arrayArg = $expr->getArgs()[0]->value;
+				$scope = $scope->assignExpression(
+					$arrayArg,
+					$this->getArraySortFunctionType($scope->getType($arrayArg)),
+					$this->getArraySortFunctionType($scope->getNativeType($arrayArg)),
+				);
+			}
+
+			if (
+				$functionReflection !== null
 				&& $functionReflection->getName() === 'extract'
 			) {
 				$extractedArg = $expr->getArgs()[0]->value;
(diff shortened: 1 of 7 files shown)


staabm:
Please have a look at https://github.com/phpstan/phpstan/issues/3312

It seems also fixed by your PR. If so, please add a test

takaram:
Yes, this PR will fix it too. I added a test.
Thank you!

src/Analyser/NodeScopeResolver.php
```diff
@@ -3079,6 +3095,24 @@ static function (?Type $offsetType, Type $valueType, bool $optional) use (&$arra
 		return $arrayType;
 	}
 
+	private function getArraySortFunctionType(Type $type): Type
+	{
+		if (!$type->isArray()->yes()) {
+			return new ErrorType();
```
ondrejmirtes: Just return the same `$type` instead

src/Analyser/NodeScopeResolver.php
```diff
@@ -3079,6 +3095,24 @@ static function (?Type $offsetType, Type $valueType, bool $optional) use (&$arra
...
+		}
+
+		$isIterableAtLeastOnce = $type->isIterableAtLeastOnce();
+		if ($isIterableAtLeastOnce->no()) {
+			return new ConstantArrayType([], []);
```
ondrejmirtes: Just return the same `$type` instead.

src/Analyser/NodeScopeResolver.php
```diff
@@ -2128,6 +2129,21 @@ static function (): void {
 				);
 			}
 
+			if (
+				$functionReflection !== null
+				&& (
+					(in_array($functionReflection->getName(), ['sort', 'rsort'], true) && count($expr->getArgs()) >= 1)
+					|| ($functionReflection->getName() === 'usort' && count($expr->getArgs()) >= 2)
```
ondrejmirtes: We don't need to complicate this by checking that the number of args is `>= 2`. Just check that the first arg is always there and join the condition into a single `in_array`.

ondrejmirtes (changes requested):
Otherwise perfect 👍

takaram:
@ondrejmirtes 
Thank you. I fixed the code.

takaram:
https://github.com/phpstan/phpstan-src/actions/runs/7682388532/job/20936712149
This failure looks to be caused by this PR, but I'm not sure why it fails🤔
I need some more investigation

takaram:
> https://github.com/phpstan/phpstan-src/actions/runs/7682388532/job/20936712149

These errors were reported before, but they were in the [baseline](https://github.com/doctrine/persistence/blob/e7cf418cafe83a75f7a7096cc9fa890b201d28f1/phpstan-baseline.neon#L33-L41).
This PR changes the type descriptions, so baseline no longer suppresses the errors.

So I think we can ignore the CI error.

src/Analyser/NodeScopeResolver.php
takaram: Fixed

src/Analyser/NodeScopeResolver.php
takaram: Fixed

src/Analyser/NodeScopeResolver.php
takaram: Fixed

ondrejmirtes:
> So I think we can ignore the CI error.

Yes we can :)

ondrejmirtes:
Thank you!

PR #2892, opened 2024-01-28
Improve parameter type of SplObjectStorage::removeAll/Except

schlndh:
See https://phpstan.org/r/bff03687-1f39-496b-a23e-0386bcb75362 . Ideally, `removeNoIntersect` should be reported as an issue since it is not possible for the removes to do anything. However, AFAIK there is no way to express such constraint in the current type system (i.e. I'd want to make sure that the key types of the two storages have a non-empty intersection).

I considered using `\SplObjectStorage<covariant TObject, *>|\SplObjectStorage<contravariant TObject, *>`, which would report `removeNoIntersect`, but it would lead to a false positive in `removePossibleIntersect`. And `\SplObjectStorage<*, *>` seems to have been the intention of the previous stub.

Fixes https://github.com/phpstan/phpstan/issues/9411

### DIFF
--- a/stubs/SplObjectStorage.stub
+++ b/stubs/SplObjectStorage.stub
@@ -42,12 +42,12 @@ class SplObjectStorage implements Countable, Iterator, Serializable, ArrayAccess
     public function getInfo() { }
 
     /**
-     * @param \SplObjectStorage<object, mixed> $storage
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Nice, thanks a lot!

PR #2894, opened 2024-01-30
Fix result-cache with phpstan-types on New_ references

staabm:
closes https://github.com/phpstan/phpstan/issues/10449


staabm:
~~I think the fix is not correct, as we have the same problem in more places. let me just have another look~~

refactored with [45307b9](https://github.com/phpstan/phpstan-src/pull/2894/commits/45307b9fb29a343683c3536a8550f3e4b162e7a0) so the fix also works for phpstan-types referenced via Catch_, Instanceof_ or similar references

PR #2897, opened 2024-02-01
Reproduce Bug 10523

staabm:
closes https://github.com/phpstan/phpstan/issues/10523
closes https://github.com/phpstan/phpstan/issues/10822

### DIFF
--- a/src/Node/ClassPropertiesNode.php
+++ b/src/Node/ClassPropertiesNode.php
@@ -152,14 +152,14 @@ class ClassPropertiesNode extends NodeAbstract implements VirtualNode
 			return [$uninitializedProperties, [], []];
 		}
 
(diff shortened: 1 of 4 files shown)


src/Node/ClassPropertiesNode.php
```diff
@@ -355,6 +359,12 @@ private function getMethodsCalledFromConstructor(
 				continue;
 			}
 
+			if ($inMethod->getName() !== '__construct') {
```
staabm, line 360: at this point we take the properties which have been assigned in `__construct` over to the `additionalConstructor` methods as described in https://github.com/phpstan/phpstan/issues/10523#issuecomment-1923616806.

staabm:
I think this one is ready for a review

ondrejmirtes:
This also fixes https://github.com/phpstan/phpstan/issues/10822 please add a regression test for that.

staabm:
@ondrejmirtes here you are

ondrejmirtes:
Thank you.

PR #2899, opened 2024-02-01
Simplify `ClassPropertiesNode->getUninitializedProperties()`

staabm:
on my way trying to understand this thing.. a small initial cleanup which stood out.

this code is not required to be called within the loop. once before is enough.


ondrejmirtes:
Thank you.

PR #2901, opened 2024-02-02
Make `array_map` argument take into account unpack

ryium:
fixes https://github.com/phpstan/phpstan/issues/10527


tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1315,6 +1315,16 @@ public function testBug10509(): void
 		$this->assertSame('PHPDoc tag @return contains unresolvable type.', $errors[1]->getMessage());
 	}
 
+	public function testBug10527(): void
```
staabm: You should move this test to the rule-test which tests the rule emitting the error you are trying to fix. 

AnalyserIntegrationTest is reserved for out-of-memory or fatal error issues

ondrejmirtes:
Thank you!

PR #2902, opened 2024-02-02
Make array_map understand union of callables.

mad-briller:
Fixes https://github.com/phpstan/phpstan/issues/10442

### DIFF
--- a/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
@@ -39,10 +39,11 @@ class ArrayMapFunctionReturnTypeExtension implements DynamicFunctionReturnTypeEx
 		$callableIsNull = $callableType->isNull()->yes();
 
 		if ($callableType->isCallable()->yes()) {
-			$valueType = new NeverType();
+			$valueTypes = [new NeverType()];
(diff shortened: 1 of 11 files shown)


ondrejmirtes:
Awesome! This fixes a bunch of issues (https://github.com/phpstan/phpstan-src/actions/runs/7974759948):

* https://github.com/phpstan/phpstan/issues/6633
* https://github.com/phpstan/phpstan/issues/3818
* https://github.com/phpstan/phpstan/issues/9594
* https://github.com/phpstan/phpstan/issues/9614
* https://github.com/phpstan/phpstan/issues/10283

Please add regression tests for them - I think type inference tests in NodeScopeResolverTest should be fine. Please put the new files in dataFileAsserts in random places and not at the end, it makes the merge conflict to be less probable :)

mad-briller:
@ondrejmirtes i've investigated the those issues and found that https://github.com/phpstan/phpstan/issues/3818 appears to be an issue regarding the application of `@var` phpdocs, and is not fixed by this. However there is an additional try link provided in that issue that is resolved so i've added that as a regression test but this change doesn't close the issue. 

however all the others are solved by this change! :tada: you love to see it

ondrejmirtes:
Awesome, thank you very much!

PR #2903, opened 2024-02-02
Fix ArrayAccess destructuring assignment producing mixed.

mad-briller:
Fixes https://github.com/phpstan/phpstan/issues/9456

### DIFF
--- a/src/Type/ObjectType.php
+++ b/src/Type/ObjectType.php
@@ -1129,12 +1129,12 @@ class ObjectType implements TypeWithClassName, SubtractableType
 
 	public function getOffsetValueType(Type $offsetType): Type
 	{
-		if (!$this->isExtraOffsetAccessibleClass()->no()) {
-			return new MixedType();
-		}
-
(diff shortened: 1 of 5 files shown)


mad-briller:
It appears changing the order of these operations breaks analysis of PrestaShop here:
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/LocalizationPack.php#L495

I can't seem to recreate an `*ERROR*` type from the `$attributes['install'] == 1` expression in a NodeScopeResolver test or a Rule test for `InvalidComparisonOperationRule`, and `(SimpleXMLElement|null)` isn't instanceof `ArrayAccess` so i can't see how it would be affected anyway.

what's interesting is 2 lines above the expression is an offset access to the `$attributes` value which isn't reported as an issue.

not sure how to proceed; i've added a second commit which shows the rule test for the operation that i tried to recreate, a little advice or direction would be a big help, thanks

VincentLanglet:
This seems to be solved

PR #2904, opened 2024-02-03
Simplify function-name determination

staabm:
determine function names based on the FuncCall node. also doing less reflection which is faster.


ondrejmirtes:
These can't be simplified. The point of ReflectionProvider is that it finds whether the called function is implemented in userland (and might do something different) or whether function is from PHP and should be handled by these rules.

See: https://3v4l.org/AMksn

PR #2909, opened 2024-02-09
Phpdoc inheritance with generics

RobertMe:
This fixes the issue with inherited template types not being resolved in the context of `@phpstan-assert*` usage.
For `@phpstan-self-out` I added some checks to the existing test to validate the (template) type

Fixes phpstan/phpstan#10037

### DIFF
--- a/src/PhpDoc/ResolvedPhpDocBlock.php
+++ b/src/PhpDoc/ResolvedPhpDocBlock.php
@@ -931,8 +931,12 @@ class ResolvedPhpDocBlock
 			$phpDocBlock = $parentPhpDocBlocks[$i];
 
 			return array_map(
-				static fn (AssertTag $assertTag) => $assertTag->withParameter(
-					$phpDocBlock->transformAssertTagParameterWithParameterNameMapping($assertTag->getParameter()),
+				static fn (AssertTag $assertTag) => self::resolveTemplateTypeInTag(
+					$assertTag->withParameter(
+						$phpDocBlock->transformAssertTagParameterWithParameterNameMapping($assertTag->getParameter()),
+					),
+					$phpDocBlock,
+					TemplateTypeVariance::createCovariant(),
 				),
 				$result,
 			);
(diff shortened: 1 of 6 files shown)


RobertMe:
I struggled to get something to work for `@throws`, so maybe you can help out with that @ondrejmirtes :) I'm unsure where to add the test (file or specific method) and following that how to properly implement it.

ondrejmirtes:
Looks like https://github.com/phpstan/phpstan/issues/9123 is also fixed, can you please add a regression test? Thanks!

RobertMe:
> Looks like [phpstan/phpstan#9123](https://github.com/phpstan/phpstan/issues/9123) is also fixed, can you please add a regression test? Thanks!

I'm not sure I'm seeing what should be tested here which isn't the case already? The only difference in that example is the explicit `/** @phpstan-assert-if-true MyEvent $event */` on the second implementation. But that already worked because it was the work-a-round you gave me as well :) And for the mentioned expected result, of both being `Event`, I think that's incorrect and should be `MyEvent`?

Bu obviously I could just copy/paste the linked example into a `bug-...` test file if you want me to. But personally I don't see how that ticket differs from the test I already wrote (but I might be missing something).

ondrejmirtes:
> I'm not sure I'm seeing what should be tested here

Every fixed issue gets a regression test, regardless of whether the issue was exactly the same or not. These code snippets are very valuable and we should not throw them away.

Please use this snippet https://phpstan.org/r/a9d9027b-da35-44f2-a757-de78dab88386 and do `assertType` instead of `dumpType`. Thank you!

RobertMe:
> Every fixed issue gets a regression test, regardless of whether the issue was exactly the same or not. These code snippets are very valuable and we should not throw them away.

Okay, just adding an explicit test case for every fixed ticket 👍🏻

> Please use this snippet https://phpstan.org/r/a9d9027b-da35-44f2-a757-de78dab88386 and do `assertType` instead of `dumpType`. Thank you!

Done, and added both of your examples from phpstan/phpstan#10037 as well

ondrejmirtes:
Perfect, thank you!

PR #2910, opened 2024-02-10
DynamicFunctionReturnTypeExtension for the get_debug_type function.

patrickkusebauch:
Closes: https://github.com/phpstan/phpstan/issues/10558
Closes: https://github.com/phpstan/phpstan/issues/10585

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -1361,6 +1361,11 @@ services:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
 
 	-
+		class: PHPStan\Type\Php\GetDebugTypeFunctionReturnTypeExtension
+		tags:
+			- phpstan.broker.dynamicFunctionReturnTypeExtension
+
+	-
 		class: PHPStan\Type\Php\GetParentClassDynamicFunctionReturnTypeExtension
 		tags:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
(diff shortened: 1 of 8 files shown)


patrickkusebauch:
Looking at the failing checks, none of them seem to be genuinely caused by this change.

staabm:
Please add a pattern-matching example-test and use it in rule-tests which emit errors before this PR and are expected to be fixed now

patrickkusebauch:
> Please add a pattern-matching example-test and use it in rule-tests which emit errors before this PR and are expected to be fixed now

Updated the PR.

Error emitted before this PR: `'20: Match expression does not handle remaining value: string'`.

patrickkusebauch:
Updated to also specify the type in condition branches. This closes https://github.com/phpstan/phpstan/issues/10585 for me.

I have a question about `TypeTraverser::map` and `TypeCombinator::union`. I tried to replicate the code in `GettypeFunctionReturnTypeExtension` as it is a similar feature. However, when trying to combine:
 - `'bool'|string` (constant string `bool` or any string) it gets normalized to `string`
 -  `'Namespace\ClassA'|'Namespace\ClassB'` (constant strings with FQCN) it gets normalized to `Namespace\ClassA|Namespace\ClassB`
 
 Is this expected? Can this be turned off? Are those considered equivalent?

ondrejmirtes:
Thank you!

PR #2914, opened 2024-02-13
Keep numeric-strings in str_repeat()

staabm:
closes https://github.com/phpstan/phpstan/issues/10572

### DIFF
--- a/src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
+++ b/src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
@@ -8,6 +8,7 @@ use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\Accessory\AccessoryLiteralStringType;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
+use PHPStan\Type\Accessory\AccessoryNumericStringType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
@@ -72,6 +73,10 @@ class StrRepeatFunctionReturnTypeExtension implements DynamicFunctionReturnTypeE
 
 		if ($inputType->isLiteralString()->yes()) {
 			$accessoryTypes[] = new AccessoryLiteralStringType();
+		}
+
+		if ($inputType->isNumericString()->yes()) {
+			$accessoryTypes[] = new AccessoryNumericStringType();
 		}
 
 		if (count($accessoryTypes) > 0) {
--- a/src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
+++ b/src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type\Php;
 
+use Nette\Utils\Strings;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/data/literal-string.php
```diff
@@ -34,6 +34,8 @@ public function doFoo($literalString, string $string)
 			str_repeat('a', 99)
 		);
 		assertType('literal-string&non-falsy-string', str_repeat('a', 100));
+		assertType('literal-string&non-empty-string&numeric-string', str_repeat('0', 100));
```
mad-briller: might be a bit pedantic but this could be `non-falsy-string` too as `'00'` is truthy

staabm: agree. still I think its "good" enough and not worth another If case

ondrejmirtes (changes requested):
Please put AccessoryNumericStringType in `ConstantStringType::generalize()` too.

src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
```diff
@@ -74,6 +75,10 @@ public function getTypeFromFunctionCall(
 			$accessoryTypes[] = new AccessoryLiteralStringType();
 		}
 
+		if ($inputType->isNumericString()->yes()) {
+			$accessoryTypes[] = new AccessoryNumericStringType();
```
ondrejmirtes: I'm afraid this can be a bit tricky. Repeating a numeric type multiple times can lead to a non-numeric type: https://3v4l.org/aRBdZ

ondrejmirtes: So we should probably do it only with a constant (literal) types with no decimal points.

ondrejmirtes (changes requested):

src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
staabm: good catch. fixed, thanks.

gharlan: I think it should only be done for strings that only contain digits. Because of these numeric strings: https://3v4l.org/U4bM2

It should also be ensured that the multiplier is not 0.

staabm: ohh my, its even different across php versions ;)

https://3v4l.org/m6et3

staabm:
I had a look into blackfire profiles and did not see any bottlenecks/resource hogs

tests/PHPStan/Analyser/data/literal-string.php
```diff
@@ -34,6 +37,25 @@ public function doFoo($literalString, string $string)
...
+
+		// see https://3v4l.org/U4bM2
+		assertType("non-empty-string", str_repeat($numericString, 1)); // could be numeric-string
+		assertType("non-empty-string", str_repeat($numericString, 2));
+		assertType("literal-string", str_repeat($literalString, 1));
+		$x = rand(1,2);
+		assertType("literal-string&non-falsy-string", str_repeat('    1   ', $x));
+		assertType("literal-string&non-falsy-string", str_repeat('+1', $x));
+		assertType("literal-string&non-falsy-string", str_repeat('1e9', $x));
+		assertType("literal-string&non-falsy-string&numeric-string", str_repeat('19', $x));
```
gharlan, line 57: Maybe add something similar to this (don't know how it should look exactly):

```php
$x = rand(0,2);
assertType("literal-string&non-falsy-string&numeric-string|''", str_repeat('19', $x));
// or just "string"?
```

staabm: good catch. I forgot to handle a maybe zero multiplyer.

the type as you suggested wouldn't work though, because while type normalization it gets less precise and will just lead to a `literal-string`.

ondrejmirtes:
Thank you.

PR #2916, opened 2024-02-14
POC: Speed optimize class-loading for levels tests

staabm:
level tests are the slowest part of the CI process and take a considerable amount of time.
[they usually take arround something like 10 minutes](https://github.com/phpstan/phpstan-src/actions/runs/7884519183/job/21513678177?pr=2914)

locally I have reduced rule level tests to a small subset to easy reproducing the problem.
my reduced set of rule-level tests before this PR run like

```

Time: 01:12.819, Memory: 32.00 MB
```

after the change in autoload rules it runs like

```

Time: 00:46.306, Memory: 24.00 MB
```

which is ~37% faster. 

lets see wheter this speedup reproduces in CI and the change itself is acceptable

----

~~side-note: the PR as is currently yields to 2 notices when running phpunit, which obviously should be fixed~~
 fixed by [11ffaa8](https://github.com/phpstan/phpstan-src/pull/2916/commits/11ffaa8afdd1a38f78b402ab238614125bd939d4)
```
$ time vendor/bin/phpunit tests/PHPStan/Generics/GenericsIntegrationTest.php --group levels
PHP Warning:  Class "ReturnTypes\Foo" not found in C:\dvl\Workspace\phpstan-src-staabm\tests\phpstan-bootstrap.php on line 8

Warning: Class "ReturnTypes\Foo" not found in C:\dvl\Workspace\phpstan-src-staabm\tests\phpstan-bootstrap.php on line 8
PHP Warning:  Class "TestAccessProperties\FooAccessProperties" not found in C:\dvl\Workspace\phpstan-src-staabm\tests\phpstan-bootstrap.php on line 9

Warning: Class "TestAccessProperties\FooAccessProperties" not found in C:\dvl\Workspace\phpstan-src-staabm\tests\phpstan-bootstrap.php on line 9
```


staabm:
In CI it seems we save ~2 minutes with this PR in Tests / Levels tests (reduced from previously ~10 minutes).

the approach of this PR is faster, because `ComposerJsonAndInstalledJsonSourceLocatorMaker` analyses PHPStan composer.json and needs to verify/check sha1 checksums for all files configured in composer autoloading. In RuleLevelTests we trigger a lot of phpstan processes which all have to re-verify the hash-sums of all files and therefore it gets slow.

With every autoloadable symbol added via a new test to PHPStan the rule-level test gets a tiny bit slower.

.github/workflows/test-levels-composer.patch
```diff
@@ -0,0 +1,18 @@
...
+index 6dde20b5f..7ea244ee1 100644
+--- a/composer.json
++++ b/composer.json
+@@ -116,8 +116,11 @@
+ 			]
+ 		},
+ 		"classmap": [
+-			"tests/e2e",
+-			"tests/PHPStan"
++			"tests/PHPStan/Rules/Methods/data/returnTypes.php",
```
ondrejmirtes: I understand the intention, but I don't love it. It's going to be really hard to maintain this list and debug problems as time goes on.

I'd be more interested in parallelizing running the tests. Basically each "topic" in LevelsTestCase could have its own GitHub Actions job.

It's not exactly the same but the idea is here: https://rias.be/blog/running-phpunit-tests-in-parallel-using-github-actions

.github/workflows/tests-levels-matrix.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+$testList = shell_exec('php vendor/bin/phpunit --list-tests');
```
staabm: currently the `--list-tests` phpunit feature does not allow filtering by test-groups.

I have opened a feature request about that, so we may can get rid/simplify this script in the future.

https://github.com/sebastianbergmann/phpunit/issues/5702

.github/workflows/test-levels-composer.patch
staabm: I have adjusted the PR. It now filters the existing tests based on class-name and builds a dynamic matrix with it.

this means we run now 40 tests in parallel (which each takes ~1m 15s) instead of one large job which runs ~10 minutes

grafik


the `tests.yml` github action workflow for this PR finished in 4m 15s, while it takes ~10m 30s before this PR 🚀

bendavies: > I understand the intention, but I don't love it. It's going to be really hard to maintain this list and debug problems as time goes on.
> 
> I'd be more interested in parallelizing running the tests. Basically each "topic" in LevelsTestCase could have its own GitHub Actions job.
> 
> It's not exactly the same but the idea is here: [rias.be/blog/running-phpunit-tests-in-parallel-using-github-actions](https://rias.be/blog/running-phpunit-tests-in-parallel-using-github-actions)

This is very similar to what i did in psalm years ago:
https://github.com/vimeo/psalm/pull/4985

But in psalm, each job also runs test in parallel.

staabm: @bendavies the tests jobs - even before this PR - are already highly parallelized - within the github action jobs and also across several jobs. 

This PR splits up one very slow part of the pre-existing tests into even more jobs because these were particular slow

bendavies: 👍  apologies.
i couldn't see (still can't) where tests are split across jobs.

staabm: >  apologies

no worries.

> i couldn't see (still can't) where tests are split across jobs.

see the screen in my above comment. there you can see the "levels test" matrix, with several jobs.
if you compare to other existing PR, this step is just a single job.

the PR implements this by inspecting the test-suite level-tests test-cases. Because of the used dataprovider, this means the `testLevels` method represents several test-cases. each of these test-cases is run in a separate github action job, because I feed this list of test-cases as a json array into the "test-levels" github action job matrix.

.github/workflows/tests-levels-matrix.php
```diff
@@ -0,0 +1,33 @@
...
+foreach(explode("\n", $testList) as $line) {
+	$cleanedLine = trim($line, ' -');
+
+	if ($cleanedLine === '') {
+		continue;
+	}
+
+	if (
+		!str_contains($cleanedLine, 'PHPStan\Generics\GenericsIntegrationTest') &&
+		!str_contains($cleanedLine, 'PHPStan\Levels\LevelsIntegrationTest')
```
ondrejmirtes: We should check the `@group` inside the docblock instead of hardcoding these two classes.

.github/workflows/tests-levels-matrix.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
ondrejmirtes: declare strict types please

.github/workflows/tests.yml
```diff
@@ -98,11 +98,45 @@ jobs:
...
+
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v3
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "8.1"
```
ondrejmirtes: Let's do 8.3 here :)

.github/workflows/tests.yml
```diff
@@ -121,7 +155,7 @@ jobs:
         run: "composer install --no-interaction --no-progress"
 
       - name: "Tests"
-        run: "make tests-levels"
+        run: "php vendor/bin/phpunit --filter '${{ matrix.test-filter }}' --group levels"
```
ondrejmirtes: I just did some local testing and in this format PHPUnit will still execute all data providers from all tests. It would be good to also add the file in which the filtered test is. I think we can get even faster that way.

See the comparison:

```
time vendor/bin/phpunit --filter 'PHPStan\\Type\\UnionTypeTest::testSorting'
15.7s

time vendor/bin/phpunit tests/PHPStan/Type/UnionTypeTest.php --filter 'PHPStan\\Type\\UnionTypeTest::testSorting'
1.1s
```

ondrejmirtes (changes requested):
I really love this! Just found a few minor things :)

.github/workflows/tests-levels-matrix.php
staabm: to make this work based on the phpdoc, I would have needed to reflect on classes and read all class-comments.

instead I updated phpunit to 10.x for the matrix job and used `--list-tests-xml` which returns a xml containing group information. its a option which does not exist in phpunit 9.x

.github/workflows/tests.yml
staabm: indeed, it got even faster now :)

grafik

.github/workflows/tests-levels-matrix.php
```diff
@@ -0,0 +1,32 @@
...
+		}
+
+		$testCaseName = (string) $testCaseMethod['id'];
+
+		[$className, $testName] = explode('::', $testCaseName, 2);
+		$fileName = 'tests/'. str_replace('\\', DIRECTORY_SEPARATOR, $className) . '.php';
+
+		$filter = str_replace('\\', '\\\\', $testCaseName);
+
+		$testFilters[] = sprintf("%s --filter '%s'", $fileName, $filter);
```
ondrejmirtes: Why is doubling backslashes needed? I’d say that instead of that we need escapeshellarg() for both fileName and filter.

ondrejmirtes (changes requested):

.github/workflows/tests-levels-matrix.php
staabm: Running the plain command on the CLI only works only with double backslashes (in the filter arg)

See your command in https://github.com/phpstan/phpstan-src/pull/2916#discussion_r1493341854

staabm: I guess we can add escapeshellarg(), but we still need this manual escaping

staabm: lets give it a try and we will see

staabm: https://github.com/phpstan/phpstan-src/actions/runs/7944377066/job/21689826256 shows that it doesn't work when we only use `escapeshellarg` (no tests executed).

ondrejmirtes: Why is that? I noticed it doesn't work but I don't know what eats the characters, whether it's bash or PHP. Doubling the backslashes seems very peculiar to me, maybe there are other things we need to fix too but we're currently unaware of that.

staabm: Phpunit examples in docs also mention double backslashes

https://docs.phpunit.de/en/9.6/textui.html#textui-examples-filter-patterns

staabm: Its interpreted as regex. Let mit try preg_quote()

ondrejmirtes: I know. It's probably a bash thing.

ondrejmirtes:
Thank you.

.github/workflows/tests-levels-matrix.php
staabm: preg_quote would escape the following characters: `. \ + * ? [ ^ ] $ ( ) { } = ! < > | : - #`

the only one of these which can show up in a class/method name is `\`. I think we can use `preg_quote` if you prefer but it should be enough to escape the backslash IMO

edit: [I had an idea](https://github.com/phpstan/phpstan-src/pull/2923)

PR #2917, opened 2024-02-14
Introduce array_values rule

kamil-zacek:


src/Rules/Functions/ArrayValuesRule.php
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		$args = $node->getArgs();
+
+		if (count($args) !== 1) {
+			return [];
+		}
+
+		if ($this->treatPhpDocTypesAsCertain) {
+			$arrayType = $scope->getType($args[0]->value);
```
ondrejmirtes, line 75: To account for named argument call like `array_values(unused: 'foo', array: $array)` you need to normalize this first. You should call `ArgumentsNormalizer::reorderFuncArguments()`. The `$parametersAcceptor` argument you can get by fetching `FunctionReflection` from ReflectionProvider and then doing `ParametersAcceptorSelector::selectFromArgs()` with `$function->getVariants()`. And write a test for this too :)

src/Rules/Functions/ArrayValuesRule.php
```diff
@@ -0,0 +1,88 @@
...
+			return [];
+		}
+
+		$args = $node->getArgs();
+
+		if (count($args) !== 1) {
+			return [];
+		}
+
+		if ($this->treatPhpDocTypesAsCertain) {
```
ondrejmirtes: Please test this with true/false.

src/Rules/Functions/ArrayValuesRule.php
```diff
@@ -0,0 +1,88 @@
...
+
+			return [
+				RuleErrorBuilder::message(sprintf(
+					$message,
+					$arrayType->describe(VerbosityLevel::value()),
+				))->build(),
+			];
+		}
+
+		if ($arrayType->isList()->yes() && $arrayType->isIterable()->yes()) {
```
ondrejmirtes: When is something a list but not iterable?

ondrejmirtes (changes requested):

src/Rules/Functions/ArrayValuesRule.php
kamil-zacek: I hope it is correct now. But when there is unsupported named argument (that unused/unsupported argument) - this rule doesn't return any error, because I hope another rule check whether function call is called with correct argument count / names.

ondrejmirtes: Yeah sure there is 😊

ondrejmirtes:
Awesome, thank you!

staabm:
Nice one 🤙

PR #2918, opened 2024-02-15
Fixes incorrect types reported by parse_url

ChrisBrenton:
Fixes parse_url return type when $component is -1 returning more types than were valid. Adds missing types when $component is not a constant value.

Fixes: https://github.com/phpstan/phpstan/issues/4754


ondrejmirtes:
Thank you.

PR #2920, opened 2024-02-16
Adds type specifier for settype.

ChrisBrenton:
I'm fairly new to PHPStan's code base and I've been looking through some of the issues marked "Easy fixes" for a way to get started. I'm not certain if I am going about this the right way or if its completely mad - particularly the test.

I don't expect this to be accepted as is, but it'd be good to get some pointers to how I might otherwise go about it if/when you have time?

### DIFF
--- a/src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
@@ -5,6 +5,8 @@ namespace PHPStan\Type\Php;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Analyser\SpecifiedTypes;
+use PHPStan\Analyser\TypeSpecifier;
+use PHPStan\Analyser\TypeSpecifierAwareExtension;
 use PHPStan\Analyser\TypeSpecifierContext;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\ErrorType;
@@ -15,9 +17,12 @@ use PHPStan\Type\TypeCombinator;
 use stdClass;
 use function count;
 use function strtolower;
+use function var_dump;
 
-class SetTypeFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyingExtension
+class SetTypeFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyingExtension, TypeSpecifierAwareExtension
 {
+
+	private TypeSpecifier $typeSpecifier;
 
 	public function isFunctionSupported(FunctionReflection $functionReflection, FuncCall $node, TypeSpecifierContext $context): bool
 	{
@@ -32,40 +37,63 @@ class SetTypeFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyingEx
 		$castType = $scope->getType($node->getArgs()[1]->value);
 
 		$constantStrings = $castType->getConstantStrings();
+		if (count($constantStrings) < 1) {
+			return new SpecifiedTypes();
+		}
+
 		$types = [];
 
 		foreach ($constantStrings as $constantString) {
 			switch ($constantString->getValue()) {
 				case 'bool':
 				case 'boolean':
+					var_dump('converting to bool');
 					$types[] = $valueType->toBoolean();
 					break;
 				case 'int':
 				case 'integer':
+					var_dump('converting to int');
 					$types[] = $valueType->toInteger();
 					break;
 				case 'float':
 				case 'double':
+					var_dump('converting to float');
 					$types[] = $valueType->toFloat();
 					break;
 				case 'string':
+					var_dump('converting to string');
 					$types[] = $valueType->toString();
 					break;
 				case 'array':
+					var_dump('converting to array');
 					$types[] = $valueType->toArray();
 					break;
 				case 'object':
+					var_dump('converting to object');
 					$types[] = new ObjectType(stdClass::class);
 					break;
 				case 'null':
+					var_dump('converting to null');
 					$types[] = new NullType();
 					break;
 				default:
+					var_dump('defaulting');
 					$types[] = new ErrorType();
 			}
 		}
 
-		return new SpecifiedTypes(['$value' => [$value, TypeCombinator::union(...$types)]], [], true);
+		return $this->typeSpecifier->create(
+			$value,
+			TypeCombinator::union(...$types),
+			TypeSpecifierContext::createTruthy(),
+			true,
+			$scope,
+		);
+	}
+
+	public function setTypeSpecifier(TypeSpecifier $typeSpecifier): void
+	{
+		$this->typeSpecifier = $typeSpecifier;
 	}
 
 }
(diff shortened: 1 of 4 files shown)


src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,53 @@
...
+	{
+		$value = $node->getArgs()[0]->value;
+		$valueType = $scope->getType($value);
+		$castType = $scope->getType($node->getArgs()[1]->value);
+
+		$constantStrings = $castType->getConstantStrings();
+		$types = [];
+
+		foreach ($constantStrings as $constantString) {
+			$types[] = match ($constantString->getValue()) {
```
staabm: IIRC the phpstan downgrade release process does not support the match expression. You need to rephrase the logic with other means

tests/PHPStan/Analyser/data/set-type-type-specifying.php
```diff
@@ -0,0 +1,511 @@
...
+
+namespace SetTypeTypeSpecifying;
+
+use stdClass;
+use function PHPStan\Testing\assertType;
+
+/**
+ * @param 'string'|'int'|'integer'|'float'|'double'|'bool'|'boolean'|'array'|'object'|'null' $castTo
+ */
+function setTypeSpecifying(string $value, string $castTo): void
```
staabm: I think a 2nd test function which works with a non-constant $x would be helpful. 

Like

`function doFoo(int $i, string $s ...)`

And assert similar to what you did in this method

ChrisBrenton: If I caught the intention of your message correctly, this is just a mistake on my part when writing this test initially. The function was supposed to receive a mixed type for the first argument to match the signature of `settype`. I've changed this now, but if that wasn't the intention, please correct me.

staabm: I would add a test-functions like

```
function doBoolean(int $i, string $s) {
	settype($i, 'boolean');
	assertType('bool', $x);

	settype($s, 'boolean');
	assertType('bool', $x);
}
```

since for this non-constant values, different types are returned.
there are lots of combinations and I think it doesn't make sense to test all combinations.. but at least a few most common ones should be tested

ChrisBrenton: Updated - I've added the functions as you described to the existing test cases.

src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,71 @@
...
+			&& count($node->getArgs()) > 1;
+	}
+
+	public function specifyTypes(FunctionReflection $functionReflection, FuncCall $node, Scope $scope, TypeSpecifierContext $context): SpecifiedTypes
+	{
+		$value = $node->getArgs()[0]->value;
+		$valueType = $scope->getType($value);
+		$castType = $scope->getType($node->getArgs()[1]->value);
+
+		$constantStrings = $castType->getConstantStrings();
```
ondrejmirtes, line 39: If the constant string is empty, we should return empty `new SpecifiedTypes` here.

src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,71 @@
...
+					break;
+				case 'null':
+					$types[] = new NullType();
+					break;
+				default:
+					$types[] = new ErrorType();
+			}
+		}
+
+		return new SpecifiedTypes(['$value' => [$value, TypeCombinator::union(...$types)]], [], true);
```
ondrejmirtes: `$value` is wrong here, it cannot work. It'd only work if the call looked like this: `settype($value, 'int')`.

You need to call `TypeSpecifier::create()`, as all the other type-specifying extensions do. That will create the correct `SpecifiedTypes` object for the right expression.

ondrejmirtes (changes requested):

src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
ChrisBrenton: I've made the change as requested, but I am not confident that I have implemented it correctly. The tests pass as I'd expect but I am very unsure as to the purpose/behaviour of $context. I would appreciate feedback on this PR if possible, and if you have time and are willing, an explanation of $context - for me this is one of the biggest unknowns I have come across so far.

src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,99 @@
...
+
+class SetTypeFunctionTypeSpecifyingExtension implements FunctionTypeSpecifyingExtension, TypeSpecifierAwareExtension
+{
+
+	private TypeSpecifier $typeSpecifier;
+
+	public function isFunctionSupported(FunctionReflection $functionReflection, FuncCall $node, TypeSpecifierContext $context): bool
+	{
+		return strtolower($functionReflection->getName()) === 'settype'
+			&& count($node->getArgs()) > 1;
```
ondrejmirtes: This needs to check that `$context->null()`.

src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,99 @@
...
+			return new SpecifiedTypes();
+		}
+
+		$types = [];
+
+		foreach ($constantStrings as $constantString) {
+			switch ($constantString->getValue()) {
+				case 'bool':
+				case 'boolean':
+					var_dump('converting to bool');
```
ondrejmirtes: Forgot var_dump calls there

src/Type/Php/SetTypeFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,99 @@
...
+					var_dump('defaulting');
+					$types[] = new ErrorType();
+			}
+		}
+
+		return $this->typeSpecifier->create(
+			$value,
+			TypeCombinator::union(...$types),
+			TypeSpecifierContext::createTruthy(),
+			true,
```
ondrejmirtes, line 81: No overwriting - this needs to be `false`.

ondrejmirtes (changes requested):
Otherwise it looks fine.

ondrejmirtes:
Yeah, and a lot of tests are failing, please fix them :)

Also, I've got an idea for a future PR: There could be a rule that checks whether the performed `settype` makes sense - e.g. whether it doesn't lead to ErrorType.

ondrejmirtes:
Oh I'm sorry, `$overwrite` needs to be true in this case, it's a very rare case where this value makes sense :)

ChrisBrenton:
No worries. Fixed. My understanding is this is `true` in this scenario because it is effecting the passed in value by reference?

ondrejmirtes:
Let's say that the original type is `non-empty-string|int`. When the extension specified the type to `StringType`, then:

* With overwrite=false the resulting type is `non-empty-string`
* With overwrite=true the resulting type `string`

Most type-specifying functions are about type-checking and narrowing, not about casting (changing) the type. That's why overwrite=false is there in 99 % of cases.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

ondrejmirtes:
FYI I thought you meant to fix https://github.com/phpstan/phpstan/issues/3250 but I don't think it does that. Please try that with a regression test in a new PR.

PR #2922, opened 2024-02-17
Adds dynamic return type extension for hash_hmac.

ChrisBrenton:


ChrisBrenton:
I've converted this to draft state temporarily. It seems that tests on PHP versions < 8.0 are failing because of the order of how types are described? It seems that in another test, the types are expected to be parenthesised too. Is there a way to handle this?

src/Type/Php/HashHmacFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,64 @@
...
+		$args = $functionCall->getArgs();
+
+		if (count($args) < 3) {
+			return null;
+		}
+
+		$algo = $scope->getType($args[0]->value);
+
+		if (!$algo instanceof ConstantStringType) {
+			$types = [new AccessoryNonEmptyStringType()];
```
mad-briller, line 43: Accessory* types should always be an intersection with the type they are accessories of:

```php
new IntersectionType([new StringType(), new AccessoryNonEmptyStringType()]);
```

additionally this could be `non-falsy-string` instead of just `non-empty-string`

mad-briller:
> It seems that in another test, the types are expected to be parenthesised too. Is there a way to handle this?

parenthesised union types are what's known as a [BenevolentUnionType](https://apiref.phpstan.org/1.10.x/PHPStan.Type.BenevolentUnionType.html). They are a [more forgiving](https://phpstan.org/config-reference#checkbenevolentuniontypes) version of a union. If there are tests already expecting a benevolent union that should be respected after your change.

src/Type/Php/HashHmacFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,64 @@
...
+	{
+		$args = $functionCall->getArgs();
+
+		if (count($args) < 3) {
+			return null;
+		}
+
+		$algo = $scope->getType($args[0]->value);
+
+		if (!$algo instanceof ConstantStringType) {
```
mad-briller, line 42: you shouldn't use `instanceof ConstantStringType` as it is [deprecated](https://phpstan.org/blog/why-is-instanceof-type-wrong-and-getting-deprecated). The [article](https://phpstan.org/blog/why-is-instanceof-type-wrong-and-getting-deprecated#constantstringtype) goes into details about what you should do instead and why.

tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -1431,6 +1431,14 @@ public function dataFileAsserts(): iterable
...
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-9123.php');
 		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10037.php');
+		if (PHP_VERSION_ID >= 80000) {
+			yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-3771-php8.php');
+		}
+		if (PHP_VERSION_ID >= 80000) {
+			return;
+		}
+
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-3771-php7.php');
```
staabm, line 1441: if you don't add this new test at the very end, but one line before the CS-fixer will not do this ugly rewrite

```suggestion
		if (PHP_VERSION_ID >= 80000) {
			yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-3771-php8.php');
		}
		if (PHP_VERSION_ID < 80000) {
			yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-3771-php7.php');
		}
		yield from $this->gatherAssertTypes(__DIR__ . '/data/bug-10037.php');
```

src/Type/Php/HashHmacFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,64 @@
...
+			$types = [new AccessoryNonEmptyStringType()];
+
+			if (!$this->phpVersion->throwsValueErrorForInternalFunctions()) {
+				$types[] = new ConstantBooleanType(false);
+			}
+
+			return TypeCombinator::union(...$types);
+		}
+
+		$knownAlgos = hash_hmac_algos();
```
staabm, line 52: this might require a `function_exists` check, in case the extension is not loaded

ChrisBrenton: Aren't `hash_hmac` and `hash_hmac_algos` both provided by the same extension? Wouldn't this call be protected by the implementation of `isFunctionSupported`?

staabm: Phpstan can analyse code for extensions without the extensions beeing installed. It knows the signatures/return types.

Phpstan does not execute the code which is analyzed. 

IsFunctionSupported() only tells whether this extension here wants to handle a given call.

paulbalandan:
Isn't this covered already by `HashFunctionsReturnTypeExtension`?

staabm:
> Isn't this covered already by `HashFunctionsReturnTypeExtension`?

good point. it might make more sense to adjust the existing extension

ChrisBrenton:
Will look into this more.

PR #2926, opened 2024-02-20
Fixed incorrect return type of PDO::quote()

rasmus0201:
Fixes https://github.com/phpstan/phpstan/issues/10512


VincentLanglet:
Friendly ping @ondrejmirtes you might have missed this "easy" PR to review or do you also want a non-regression test for such changes ?

ondrejmirtes:
I didn't merge this yet because I'm not sure about this change.

The documentation states:

> Returns false if the driver does not support quoting in this way.

So a benevolent union type might be the better choice here - `__benevolent<string|false>`.

Also the usage of this function at all is not recommended in favor of prepared statements.

That's why I didn't rush to merge this.

PR #495, opened 2021-04-15
Add AlwaysUsedClassConstantsExtension interface

jdecool:
This PR should fix https://github.com/phpstan/phpstan/issues/4859

As discussed there is no mechanism like `ReadWritePropertiesExtension` but for class constants.

This PR introduce an `AlwaysUsedClassConstantsExtension` to fix this.

PR #2927, opened 2024-02-20
Add always used method extension

axlon:
This PR adds a new extension type to PHPStan. This new extension type allows developers to tell PHPStan when a private method is used.

My personal use case:
I make use of a library that exposes a trait that dynamically calls methods on my objects, since I have no other uses for these methods I typically make them private. Since there are no direct references to these methods PHPStan reports them as unused.

I used #495 as a blueprint for this PR.

Lastly, one thing I'm somewhat unsure about is the positioning of the extension check within the rule, maybe moving it up or down somewhat could be beneficial to performance.

### DIFF
--- a/src/Rules/Methods/AlwaysUsedMethodExtension.php
+++ b/src/Rules/Methods/AlwaysUsedMethodExtension.php
@@ -1,0 +1,27 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Methods;
+
+use PHPStan\Reflection\MethodReflection;
+
+/**
+ * This is the extension interface to implement if you want to describe an always-used class method.
+ *
+ * To register it in the configuration file use the `phpstan.methods.alwaysUsedMethodExtension` service tag:
+ *
+ * ```
+ * services:
+ * 	-
+ *		class: App\PHPStan\MyExtension
+ *		tags:
+ *			- phpstan.methods.alwaysUsedMethodExtension
+ * ```
+ *
+ * @api
+ */
+interface AlwaysUsedMethodExtension
+{
+
+	public function isAlwaysUsed(MethodReflection $methodReflection): bool;
+
+}
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -975,6 +975,9 @@ services:
 		class: PHPStan\Rules\Constants\LazyAlwaysUsedClassConstantsExtensionProvider
 
 	-
+		class: PHPStan\Rules\Methods\LazyAlwaysUsedMethodExtensionProvider
+
+	-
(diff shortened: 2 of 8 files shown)


axlon:
@ondrejmirtes is this something you would consider adding to PHPStan? If something is missing or needs clarification please let me know, happy to look at it!

PS I noticed 2 actions are failing, but I don't know why or if it even related to this PR

ondrejmirtes:
@axlon Don't worry, I really like this PR, it's well thought-out, I just didn't have much time to look into this thorougly.

src/Rules/Methods/AlwaysUsedMethodExtension.php
```diff
@@ -0,0 +1,12 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Methods;
+
+use PHPStan\Reflection\MethodReflection;
+
+interface AlwaysUsedMethodExtension
```
ondrejmirtes, line 22: This needs `@api` and also description similar to https://apiref.phpstan.org/1.11.x/source-src.PhpDoc.StubFilesExtension.html#5.

ondrejmirtes (changes requested):
Besides one little thing this works for me :) Thank you!

ondrejmirtes:
Perfect, thank you!

thg2k:
I like the use case, I ran into something similar in the past. But I was wondering: Wouldn't an annotation be more suitable for this job rather than a plugin type?

```php

/**
 * @phpstan-always-used
 */
private function my_method(): void {
}

```

ondrejmirtes:
> Wouldn't an annotation be more suitable for this job

You can write an annotation - an ignore annotation :)

The thing with extensions is that you can describe the logic generally. Instead of adding thousands of annotations in your code, you can write a single extension that describes the logic to mark the right methods as always used.

thg2k:
> You can write an annotation - an ignore annotation :)
> 

That's surely a solution, but it would ignore every possible error, not only the intended one. I think I solved with a ignoreErrors template in the config file, like a regexp ignoring all kind of private unused errors, targeting the specific directory where the problematic classes reside.

> The thing with extensions is that you can describe the logic generally. Instead of adding thousands of annotations in your code, you can write a single extension that describes the logic to mark the right methods as always used.

That's very true.

Thx!

ondrejmirtes:
> but it would ignore every possible error, not only the intended one

This is going to be solved in PHPStan 1.11 with error identifiers and new `@phpstan-ignore` tag. See: https://phpstan.org/user-guide/ignoring-errors#ignoring-in-code-using-phpdocs

PR #2928, opened 2024-02-21
ArrayType - string offset might exist as integer offset

michalbundyra:
Fixes phpstan/phpstan#10610

I am not sure yet what changes to do in `phpstan-baseline.neon` (if any), and if we need any `assertType` in test case 🤔


michalbundyra:
@ondrejmirtes pretty unsure how to fix it now. Not completely sure if my changes are good too. Feel free to push change or if you could suggest something I am happy to try :) Thanks

ondrejmirtes:
Looks like you fixed it. You just need to update the existing tests that are failing because it should no longe error.

You should assertType `self::MAP[$k][$value]` to make sure that `getOffsetValueType` behaves the same way `	hasOffsetValueType` does.

michalbundyra:
@ondrejmirtes 

Thanks for your help here.

> Looks like you fixed it. You just need to update the existing tests that are failing because it should no longe error.

It's done now - https://github.com/phpstan/phpstan-src/pull/2928/commits/72c18e868ed86ee8dc0e8ea3fe4281ea4dc8780d


> You should assertType self::MAP[$k][$value] to make sure that getOffsetValueType behaves the same way 	hasOffsetValueType does.

Not sure how I can get that? And where it is needed as atm I cannot see any test failure 🤔 


Thanks

ondrejmirtes:
Basically do the same thing I did here https://github.com/phpstan/phpstan-src/commit/0b78c55f42f4e097eeb7be80d5aff59fb84c4b67

michalbundyra:
@ondrejmirtes 

> Basically do the same thing I did here https://github.com/phpstan/phpstan-src/commit/0b78c55f42f4e097eeb7be80d5aff59fb84c4b67

I do not really understand the changes in there - https://github.com/phpstan/phpstan-src/commit/0b78c55f42f4e097eeb7be80d5aff59fb84c4b67, as it is:

```php
		$value = self::MAP[$value] ?? $value;

		assertType("'Test1'|'Test2'", self::MAP[$value]);
```

but as you see `$value` is already changed. So I do not really know what this assertType does there and why it has this value. For me it looks wrong.

I've pushed some changes. Hope this is what we need 🤞

tests/PHPStan/Rules/Variables/data/bug-10610.php
```diff
@@ -0,0 +1,87 @@
...
+		'bar' => [
+			'30' => 'Test3',
+		],
+	];
+
+	public function validate(string $k, string $value): void
+	{
+		$res = self::MAP[$k][$value] ?? '';
+
+		assertType("'582'|'689'|'817'|'1050'|'2923'|'4057'|'4078'|'4078'|'4079'|'4080'|'4080'|'4091'|'4111'|'4112'|'4113'|'4117'|'4118'|'4126'|'4129'|'4309'|'4904'|'5222'|'5223'|'5242'|'5250'|'5252'|'5255'|'5261'|'5264'|'5268'|'5269'|'5270'|'5271'|'5271'|'5274'|'5277'|'5278'|'5279'|'5279'|'5280'|'5281'|'5281'|'5281'|'5283'|'5283'|'5285'|'5285'|'5298'|'5305'|'5310'|'5310'|'5352'|'5364'|'5375'|'5460'|'7581'|'7624'|'7672'|'7953'|'7953'|'8206'|'9697'|'Test3'", self::MAP[$k][$value]);
```
ondrejmirtes: You need to do this in a file that's referenced from NodeScopeResolverTest

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Variables/data/bug-10610.php
ondrejmirtes: Never trust a test you haven't seen fail 😊

michalbundyra: I am very lost here, I have added this: https://github.com/phpstan/phpstan-src/pull/2928/files#diff-1e9b47d3e7051caa11f9d76ff7bd6ff2afb430014dd93dad916cbb8c2148f545R171

Sorry 🤦‍♂️

ondrejmirtes: The CI results show that the `getOffsetValueType` needs updating.

ondrejmirtes: Also feel free to make the array shorter, we don't need such a big array to verify this functionality. Thanks.

michalbundyra: Sorry @ondrejmirtes , i do not follow, the error message is:
```diff
--- Expected
+++ Actual
@@ @@
-''582'|'689'|'817'|'1050'|'2923'|'4057'|'4078'|'4078'|'4079'|'4080'|'4080'|'4091'|'4111'|'4112'|'4113'|'4117'|'4118'|'4126'|'4129'|'4309'|'4904'|'5222'|'5223'|'5242'|'5250'|'5252'|'5255'|'5261'|'5264'|'5268'|'5269'|'5270'|'5271'|'5271'|'5274'|'5277'|'5278'|'5279'|'5279'|'5280'|'5281'|'5281'|'5281'|'5283'|'5283'|'5285'|'5285'|'5298'|'5305'|'5310'|'5310'|'5352'|'5364'|'5375'|'5460'|'7581'|'7624'|'7672'|'7953'|'7953'|'8206'|'9697'|'Test3''
+'*ERROR*'
```

I bet you do not want me replace all these types with `'*ERROR*'` ❓

ondrejmirtes: The code in the ArrayType::getOffsetValueType() method needs a similar fix you did in hasOffsetValueType().

michalbundyra:
@ondrejmirtes it should be sorted now. Thanks 🙏 

> Also feel free to make the array shorter

I've tried. But for mysterious reason I cannot replicate the issue with shorter array 🤦‍♂️

tests/PHPStan/Rules/Variables/data/bug-10610.php
michalbundyra: Updated also types, just unique values.

michalbundyra:
I only hope you squash PRs ;-)

ondrejmirtes:
Thank you!

PR #2929, opened 2024-02-21
Fix return values of mysqli get_warnings

janedbal:
- https://www.php.net/manual/en/mysqli.get-warnings.php
- https://www.php.net/manual/en/mysqli-stmt.get-warnings.php


ondrejmirtes:
Thank you!

PR #2930, opened 2024-02-21
Fix false issue when passing enum case to generic class

EmilMassey:
Fixes https://github.com/phpstan/phpstan/issues/10484


ondrejmirtes:
See https://github.com/phpstan/phpstan/issues/10484#issuecomment-2290984517

PR #2931, opened 2024-02-22
Add resolution of generic @method tags

mad-briller:
Attempt to close: https://github.com/phpstan/phpstan/issues/6371#issuecomment-1032316511

This is as far as i've managed to get today on this issue; thanks to your help i've managed to forward the `TemplateTag`s through and correctly build the template type map, but now i can't get it to resolve the return types correctly :sweat_smile: 

The output i'm getting from the test is:

```
There were 2 failures:

1) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/home/brad.miller/code/phpstan-src/tests/PHPStan/Analyser/data/generic-method-tags.php:21" ('type', '/home/brad.miller/code/phpsta...gs.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\ObjectType Object (...), 21)
Expected type int, got type GenericMethodTags\TVal in /home/brad.miller/code/phpstan-src/tests/PHPStan/Analyser/data/generic-method-tags.php on line 21.
--- Expected
+++ Actual
@@ @@
-'int'
+'GenericMethodTags\TVal'

/home/brad.miller/code/phpstan-src/src/Testing/TypeInferenceTestCase.php:102
/home/brad.miller/code/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:1450

2) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/home/brad.miller/code/phpstan-src/tests/PHPStan/Analyser/data/generic-method-tags.php:22" ('type', '/home/brad.miller/code/phpsta...gs.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\ObjectType Object (...), 22)
Expected type string, got type GenericMethodTags\TVal in /home/brad.miller/code/phpstan-src/tests/PHPStan/Analyser/data/generic-method-tags.php on line 22.
--- Expected
+++ Actual
@@ @@
-'string'
+'GenericMethodTags\TVal'
```

i suspect it's because the `$returnType` passed to `AnnotationMethodReflection` is resolved with only the class-level templates and not both class and method level templates in `TemplateTypeHelper::resolveTemplateTypes`. `resolveTemplateTypes()` is returning an `ObjectType()` which seems incorrect, but i'm not sure what it's supposed to be to cause it to be templated properly.

To resolve this i've tried:

1. union the `$classReflection->getActiveTemplateTypeMap()` and my generated `$templateTypeMap` before passing it in to `resolveTemplateTypes()`
2. removed the call to `TemplateTypeHelper::resolveTemplateTypes`, as i suspected it may be resolving too soon? 

Sorry to keep asking, but is there any further pointers you have for me? Thanks.

### DIFF
--- a/src/PhpDoc/PhpDocNodeResolver.php
+++ b/src/PhpDoc/PhpDocNodeResolver.php
@@ -185,12 +185,24 @@ class PhpDocNodeResolver
 					);
 				}
 
+				$templates = [];
+				foreach ($tagValue->templateTypes as $templateType) {
+					$templates[$templateType->name] = new TemplateTag(
+						$templateType->name,
+						$templateType->bound !== null
+							? $this->typeNodeResolver->resolve($templateType->bound, $nameScope)
+							: new MixedType(),
+						TemplateTypeVariance::createInvariant()
+					);
+				}
+
 				$resolved[$tagValue->methodName] = new MethodTag(
 					$tagValue->returnType !== null
 						? $this->typeNodeResolver->resolve($tagValue->returnType, $nameScope)
 						: new MixedType(),
 					$tagValue->isStatic,
 					$parameters,
+					$templates
 				);
 			}
 		}
--- a/src/PhpDoc/PhpDocNodeResolver.php
+++ b/src/PhpDoc/PhpDocNodeResolver.php
@@ -165,7 +165,7 @@ class PhpDocNodeResolver
 			foreach ($phpDocNode->getMethodTagValues($tagName) as $tagValue) {
 				$templateTags = [];
 
-				if (count($tagValue->templateTypes) > 0) {
+				if (count($tagValue->templateTypes) > 0 && $nameScope->getClassName() !== null) {
 					foreach ($tagValue->templateTypes as $templateType) {
 						$templateTags[$templateType->name] = new TemplateTag(
 							$templateType->name,
--- a/src/PhpDoc/PhpDocNodeResolver.php
+++ b/src/PhpDoc/PhpDocNodeResolver.php
@@ -172,7 +172,7 @@ class PhpDocNodeResolver
 							$templateType->bound !== null
 								? $this->typeNodeResolver->resolve($templateType->bound, $nameScope)
 								: new MixedType(),
-							TemplateTypeVariance::createInvariant()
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
Thi has something to do with NameScope (which contains information about what T refers to in PHPDoc) and might not be actually trivial to fix, I'll think about it 😊

src/PhpDoc/PhpDocNodeResolver.php
```diff
@@ -185,12 +185,24 @@ public function resolveMethodTags(PhpDocNode $phpDocNode, NameScope $nameScope):
 					);
 				}
 
+				$templates = [];
```
ondrejmirtes: I figured out what needs to be done.

We need to look at `$tagValue->templateTypes` first and when it's non empty, we need to call `$nameScope->withTemplateTypeMap(...)`. And pass that object instead of currently passing `$nameScope` basically to all `$this->typeNodeResolver->resolve(...)` inside this foreach in `resolveMethodTags`.

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
I think I did it right.

mad-briller:
ah nice! beat me to it, was just finishing breakfast and then going to take a look at this

ondrejmirtes:
Anyway, there's a very similar feature request - to do the same thing for callables and Closures. So first we need to finish https://github.com/phpstan/phpdoc-parser/pull/199 and then to a very similar PR in phpstan-src to this one.

ondrejmirtes:
Thank you!

staabm:
Nice!

ondrejmirtes:
I just realized we're missing some rules about this. Basically the equivalents of these: https://phpstan.org/r/ca12c484-b4f1-4e5b-a418-3069ec569411

The best way to implement them is to create a new rule similar to https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Rules/Generics/MethodTemplateTypeRule.php#L78 that also calls TemplateTypeCheck.

Do you have time to implement this? Thanks .)

PR #2932, opened 2024-02-22
Error on references of PHPStan internal class names

staabm:
As discussed in https://github.com/phpstan/phpstan/issues/10587#issuecomment-1956826459 this PR adds a new `ClassForbiddenNameCheck` which reports errors when code references a PHPStan prefixed class from a phpstan.phar file


staabm:
Similar problems can happen on other popular tools, e.g. Rector. 
Should we also hardcode a rector prefix?

ondrejmirtes:
Thank you, I love this!

I wonder if we could detect it automatically but then we're in the https://github.com/shipmonk-rnd/composer-dependency-analyser area so probably not worth it.

I'd hardcode:

* PHPStan prefix
* Rector prefix
* The default PHP-Scoper prefix (https://github.com/humbug/php-scoper) - please research what it is.

staabm:
Added support for rector

https://github.com/rectorphp/rector-src/blob/f5a45a5aeb6e5e1264d546e55dbd8880046b36ce/scoper.php#L31

… and php-scoper

https://github.com/humbug/php-scoper/blob/cd909d4bb53e6668a0284fea08458428c28af697/src/Configuration/ConfigurationFactory.php#L427C17-L427C27

src/Rules/ClassNameCheck.php
```diff
@@ -0,0 +1,31 @@
...
+	/**
+	 * @param ClassNameNodePair[] $pairs
+	 * @return RuleError[]
+	 */
+	public function checkClassNames(array $pairs, bool $checkClassCaseSensitivity = true): array
+	{
+		$errors = [];
+
+		if ($checkClassCaseSensitivity) {
+			$errors += $this->classCaseSensitivityCheck->checkClassNames($pairs);
```
ondrejmirtes: I don't really like doing `+` here. I'd rather iterate over the array and add the error one-by-one with `$errors[] = ...`.

src/Rules/ClassForbiddenNameCheck.php
```diff
@@ -0,0 +1,49 @@
...
+
+			if ($projectName === null) {
+				continue;
+			}
+
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'Internal %s Class cannot be referenced: %s.',
+				$projectName,
+				$className,
+			))->line($pair->getNode()->getLine())->build();
```
ondrejmirtes: Should be non-ignorable.

src/Rules/ClassForbiddenNameCheck.php
```diff
@@ -0,0 +1,49 @@
...
+					break;
+				}
+			}
+
+			if ($projectName === null) {
+				continue;
+			}
+
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'Internal %s Class cannot be referenced: %s.',
```
ondrejmirtes: Please rewrite this as: "Referencing prefixed PHPStan class: _PHPStan_d5242a5\Symfony/Finder"

And add a tip without the prefix: "This is most likely unintentional. Did you mean to type Symfony\Finder?"

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #2935, opened 2024-02-23
Add rule to check @method template tags don't clash with class templates, type aliases or existing classes.

mad-briller:
follow up from: https://github.com/phpstan/phpstan-src/pull/2931#issuecomment-1961157975


ondrejmirtes:
Perfect, thank you!

PR #2936, opened 2024-02-23
composer-dependency-analyser: detect also unused dependencies

janedbal:
- followup of #2875 
  - **enable unused dependency checking**

### DIFF
--- a/build/composer-dependency-analyser.php
+++ b/build/composer-dependency-analyser.php
@@ -5,9 +5,33 @@ use ShipMonk\ComposerDependencyAnalyser\Config\ErrorType;
 
 $config = new Configuration();
 
+$polyfills = [
+	'symfony/polyfill-intl-grapheme',
+	'symfony/polyfill-intl-normalizer',
+	'symfony/polyfill-mbstring',
+	'symfony/polyfill-php73',
+	'symfony/polyfill-php74',
+	'symfony/polyfill-php80',
+	'symfony/polyfill-php81',
+];
+
+$pinnedToSupportPhp72 = [
+	'symfony/process',
+	'symfony/service-contracts',
+	'symfony/string',
+];
+
 return $config
 	->addPathToScan(__DIR__ . '/../bin', true)
-	->ignoreErrors([ErrorType::UNUSED_DEPENDENCY])
+	->ignoreErrorsOnPackages(
+		[
+			'hoa/regex', // used only via stream wrapper hoa://
+			'react/async', // function usage (https://github.com/shipmonk-rnd/composer-dependency-analyser/issues/67)
+			...$pinnedToSupportPhp72, // those are unused, but we need to pin them to support PHP 7.2
+			...$polyfills, // not detected by composer-dependency-analyser
+		],
+		[ErrorType::UNUSED_DEPENDENCY],
+	)
 	->ignoreErrorsOnPackage('phpunit/phpunit', [ErrorType::DEV_DEPENDENCY_IN_PROD]) // prepared test tooling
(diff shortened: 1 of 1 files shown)


composer.lock
```diff
@@ -4127,97 +4129,31 @@
...
-                    "url": "https://tidelift.com/funding/github/packagist/symfony/symfony",
-                    "type": "tidelift"
-                }
-            ],
-            "time": "2023-08-07T10:36:04+00:00"
-        },
         {
             "name": "symfony/service-contracts",
-            "version": "v2.5.2",
+            "version": "v3.4.1",
```
ondrejmirtes: You can't simply update these dependencies :)

janedbal: Why? There is no usage in the codebase, or is it? 

So this became transitive dependency after doing `composer remove symfony/service-contracts` (as it is still needed by some other package). So it got upgraded to latest version by composer (as any other transitive dependency here).

janedbal: Ok, you [partially explained it here](https://github.com/ondrejmirtes/simple-downgrader/issues/5#issuecomment-1961364675)

ondrejmirtes: > So this became transitive dependency

This is in composer.json `require` for the same reason jetbrains/phpstorm-stubs is there. To force a specific version.

You worked around it just by ignoring it, but I think that composer-dependency-analyser shouldn't mark dependencies in composer.json as unused if they're already a transitive dependency.

janedbal: > composer-dependency-analyser shouldn't mark dependencies in composer.json as unused if they're already a transitive dependency

This is hard to say in general. Currently, any dependency without a usage is marked as unused. I'd rather keep it as-is and let maintainers decide if they need it in composer.json (due to some version pinning like here) or they just forgot it there. Ignoring the error for specific packages is super easy and when commented, it documents the purpose why it is there even when unused.

janedbal: Ok, I removed only things that didnt break CI. 

So if your coverage is solid, upgrading `symfony/service-contracts` by removing direct dependency is safe here.

composer.json
```diff
@@ -14,7 +14,6 @@
 		"hoa/compiler": "3.17.08.08",
 		"hoa/exception": "^1.0",
 		"hoa/file": "1.17.07.11",
-		"hoa/regex": "1.17.01.13",
```
ondrejmirtes: `hoa/regex` is used and I want to keep it here (I know your tool cannot possibly detect it): https://github.com/phpstan/phpstan-src/blob/fd0ded1b5422acb941201fd3dd42e72d94386598/src/DependencyInjection/ValidateIgnoredErrorsExtension.php#L59

composer.json
```diff
@@ -45,7 +44,6 @@
 		"symfony/polyfill-php80": "^1.23",
 		"symfony/polyfill-php81": "^1.27",
 		"symfony/process": "^5.4.3",
-		"symfony/service-contracts": "^2.5.0",
```
ondrejmirtes: This is going to install `3.4.1` which is `php >= 8.1`... Besides relying on CI which detects these mistakes I also manually check what versions these packages depend on...

ondrejmirtes (changes requested):

composer.json
janedbal: Ok, I'll revert that too. 

It must be hard for you to maintain it like this :) Quick check of composer.lock reveals few more:

- symfony/deprecation-contracts: >=8.1
- composer/pcre: ^7.4 || ^8.0
- psr/container: >=7.4.0

Maybe you want to pin those too.

composer.json
janedbal: Thanks, ignored too.

ondrejmirtes:
Thank you!

PR #2938, opened 2024-02-23
Add support for generic CallableType.

mad-briller:
Trying to follow up the `phpdoc-parser` pr with implementing generic callables and closures. About as far as i can get tonight so putting this up to get some feedback; not sure if this is the right direction at all, just trying to apply what i learned from the other prs.

Right now the test is outputting:

```
1) PHPStan\Analyser\NodeScopeResolverTest::testFileAsserts with data set "/home/brad.miller/code/phpstan-src/tests/PHPStan/Analyser/data/generic-callables.php:12" ('type', '/home/brad.miller/code/phpsta...es.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\Generic\TemplateMixedType Object (...), 12)
Expected type int, got type TRet (anonymous function, argument) in /home/brad.miller/code/phpstan-src/tests/PHPStan/Analyser/data/generic-callables.php on line 12.
--- Expected
+++ Actual
@@ @@
-'int'
+'TRet (anonymous function, argument)'
```

which seems somewhat familiar to the `$nameScope` thing you mentioned in the generic `@method` pr, but i am calling `->withTemplateTypeMap()` like you suggested over there.

One major thing that i can't seem to find is where the template types are replaced with the actual real types; as i feel like with the addition of `$resolvedTemplateTypeMap` on `CallableType` i'm going to have to do that somewhere to then pass in the resolved type map, but where i should do that escapes me.

### DIFF
--- a/src/Type/Generic/TemplateTypeHelper.php
+++ b/src/Type/Generic/TemplateTypeHelper.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type\Generic;
 
+use PHPStan\Reflection\ParametersAcceptor;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\NonAcceptingNeverType;
@@ -85,8 +86,35 @@ class TemplateTypeHelper
 	 */
 	public static function toArgument(Type $type): Type
 	{
+		$ownedTemplates = [];
+
 		/** @var T */
-		return TypeTraverser::map($type, static function (Type $type, callable $traverse): Type {
+		return TypeTraverser::map($type, static function (Type $type, callable $traverse) use (&$ownedTemplates): Type {
+			if ($type instanceof ParametersAcceptor) {
+				$templateTypeMap = $type->getTemplateTypeMap();
+
+				foreach ($type->getParameters() as $parameter) {
+					$parameterType = $parameter->getType();
+					if (!($parameterType instanceof TemplateType) || !$templateTypeMap->hasType($parameterType->getName())) {
+						continue;
+					}
+
+					$ownedTemplates[] = $parameterType;
+				}
+
+				$returnType = $type->getReturnType();
+
+				if ($returnType instanceof TemplateType && $templateTypeMap->hasType($returnType->getName())) {
+					$ownedTemplates[] = $returnType;
+				}
+			}
+
+			foreach ($ownedTemplates as $ownedTemplate) {
+				if ($ownedTemplate === $type) {
+					return $traverse($type);
+				}
+			}
+
 			if ($type instanceof TemplateType) {
 				return $traverse($type->toArgument());
 			}
--- a/src/Type/Generic/TemplateTypeScope.php
+++ b/src/Type/Generic/TemplateTypeScope.php
@@ -7,6 +7,11 @@ use function sprintf;
 class TemplateTypeScope
 {
 
+	public static function createWithAnonymousFunction(): self
+	{
+		return new self(null, null);
+	}
+
 	public static function createWithFunction(string $functionName): self
 	{
 		return new self(null, $functionName);
@@ -48,6 +53,10 @@ class TemplateTypeScope
 	/** @api */
 	public function describe(): string
 	{
+		if ($this->className === null && $this->functionName === null) {
+			return 'anonymous function';
+		}
+
 		if ($this->className === null) {
 			return sprintf('function %s()', $this->functionName);
 		}
--- a/tests/PHPStan/Analyser/data/generic-callables.php
+++ b/tests/PHPStan/Analyser/data/generic-callables.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace GenericCallables;
+
+use Closure;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @template TFuncRet of mixed
+ * @param TFuncRet $mixed
+ *
+ * @return Closure(): TFuncRet
+ */
+function testFuncClosure(mixed $mixed): Closure
+{
+}
+
+/**
+ * @template TFuncRet of mixed
+ * @param TFuncRet $mixed
+ *
+ * @return Closure<TClosureRet of mixed>(TClosureRet $val): (TClosureRet|TFuncRet)
+ */
+function testFuncClosureMixed(mixed $mixed)
+{
+}
+
+/**
+ * @template TFuncRet of mixed
+ * @param TFuncRet $mixed
+ *
+ * @return callable(): TFuncRet
+ */
+function testFuncCallable(mixed $mixed): callable
+{
+}
+
+/**
+ * @param Closure<TRet of mixed>(TRet $val): TRet $callable
+ * @param non-empty-list<Closure<TRet of mixed>(TRet $val): TRet> $callables
+ */
+function testClosure(Closure $callable, int $int, string $str, array $callables): void
+{
+	assertType('Closure<TRet of mixed>(TRet): TRet', $callable);
+	assertType('int', $callable($int));
+	assertType('string', $callable($str));
+	assertType('string', $callables[0]($str));
+	assertType('Closure(): 1', testFuncClosure(1));
+}
+
+function testClosureMixed(int $int, string $str): void
+{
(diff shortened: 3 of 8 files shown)


src/Type/Generic/TemplateTypeScope.php
```diff
@@ -7,6 +7,11 @@
 class TemplateTypeScope
 {
 
+	public static function createWithAnonymousFunction(): self
```
mad-briller, line 10: not sure at all about this but wasn't sure of an alternative

ondrejmirtes (changes requested):
Yeah, I'd do the same thing here, if it doesn't work you have to debug why :)

The transformation happens in `CallableType::traverse()`. So basically it should be automatic.

You should also pass TemplateTypeMap into ClosureType.

As an exercise you should make the CallableType constructor parameters required, and make sure they're passed in everywhere it's relevant. Most importantly in all `new self(...)` instances inside the class.

What gives me a little pause is `$this->resolvedTemplateTypeMap`. I'm not sure what should happen there. Maybe we can just continue using `createEmpty` there.

ondrejmirtes:
The main workflow that influences generics is through ParametersAcceptorSelector, GenericParametersAcceptorResolver specifically. So the code that resolves these types will always go through there, so you can debug what's going on.

ondrejmirtes:
I realized that ClosureType already supports generics (https://phpstan.org/r/eebaf81b-333c-4e96-97fb-801436690236) so if something does not work for CallableType you can debug what's going on in ClosureType and get inspired by that :)

src/Type/Generic/TemplateTypeHelper.php
```diff
@@ -85,8 +86,35 @@ public static function resolveToBounds(Type $type): Type
 	 */
 	public static function toArgument(Type $type): Type
 	{
+		$ownedTemplates = [];
+
+		if ($type instanceof ParametersAcceptor) {
```
mad-briller: really not sure about this;

the reason i've added it is because in `MutatingScope->enterFunctionLike` calls all phpdoc types are turned to into arguments instead of parameters, which would prevent the templating from occurring, leading to the failing test i posted here yesterday.

the idea of this code is to not transform any templates owned by the `callable` or `Closure` phpdoc to arguments, which made my tests and other failing NodeScopeResolver generic tests pass aswell so i think it's correct?

mad-briller:
thanks for your wisdom there @ondrejmirtes it guided me to where to begin looking, and great tip about `ClosureType` too it got the ball rolling way faster than i would've myself; i was just starting with callable and was do closure after i got it working, sods law that i chose the more awkward way around :sweat_smile:

src/Type/Generic/TemplateTypeHelper.php
ondrejmirtes: I don't think this works that well. This will only work if the generic callable is the top-most type passed in here. When you have a union like `null|callable<T>(T): T` it falls apart.

The pattern I usually use to solve an issue like this is to do this inside the `TypeTraverser` callback:

```php
if ($type instanceof CallableType) {
    return TypeTraverser::map($type, static function (Type $type, callable $traverser) {
        // do different logic to callables based on their templates
    });
}
```

Example of where I did this: https://github.com/phpstan/phpstan-src/blob/f87269f2d652eb1494b95d3312a211ca2929ea37/src/Reflection/ResolvedFunctionVariant.php#L221

ondrejmirtes (changes requested):
We also need to catch the things MethodTagValueNode catches for these types in these places:

* Method parameters
* Method `@param-out`
* Method return types
* Function parameters
* Function `@param-out`
* Function return types
* Property types
* Inline `@var`s
* All the class-level PHPDocs like `@method` and `@property`

You can write the common logic in a rule helper that you'll inject into these rules:

* IncompatiblePhpDocTypeRule (covers everything for functions and methods)
* IncompatiblePropertyPhpDocTypeRule
* IncompatibleSelfOutTypeRule
* InvalidPhpDocVarTagTypeRule
* MethodTagTemplateTypeRule (maybe rename this one)
* a new rule for property tags - nothing is currently checked about them!

This could also be an opportunity to check more about `@method` parameters and return types, like classes that do not exist (this is typically done in `ExistingClassesIn*` rules), unresolvable types etc.

Do you want to do this here or in a follow-up PR? :) Thanks.

ondrejmirtes:
Also please inspect the failures in https://github.com/phpstan/phpstan-src/actions/runs/8031291020, they look very interesting!

ondrejmirtes:
Oh sorry, most of them are fault of a previous PR :)

mad-briller:
> Oh sorry, most of them are fault of a previous PR :)

yeah i think the rector ones are caused by this pr so i'll look into them when i get more time; working on this has cooked my brain so i need a break!

src/Type/Generic/TemplateTypeHelper.php
mad-briller: had another go at this and it's a can of worms, for example:

```
/**
 * @param Closure<TRetFirst of mixed>(TRetFirst $valone): (Closure<TRetSecond of mixed>(TRetSecond $valtwo): (TRetFirst|TRetSecond)) $closure
 */
function testNestedClosures(Closure $closure, string $str, int $int): void
{
	assertType('Closure<TRetFirst of mixed>(TRetFirst): (Closure<TRetSecond of mixed>(TRetSecond): (TRetFirst|TRetSecond))', $closure);
	$closure1 = $closure($str);
	assertType('Closure<TRetSecond of mixed>(TRetSecond): (string|TRetSecond)', $closure1);
	$result = $closure1($int);
	assertType('string|int', $result);
}
```

can't figure out at all how to handle this in `TypeTraverser::map`. The complexity comes from the fact there are multiple lists of `$ownedTemplates` depending on where in the tree walk we are at, and simply doing 

```
if ($type instanceof CallableType) {
    return TypeTraverser::map($type, static function (Type $type, callable $traverser) {
        // do different logic to callables based on their templates
    });
}
```
at the top of the first `TypeTraverser::map` means that we fail to walk the second tree of ClosureType and it's owned templates... i've tried many different approaches and failed at them all, gonna sit on it for a bit and try again tomorrow

mad-briller:
The rector failures are caused by the action installing `1.25.0` of `phpdoc-parser` and not `1.26.0`, which means `$typeNode->templateTypes` is null. rector has it's own requirement for this package and so it's the old problem of phar vs local autoloading order:

https://github.com/rectorphp/rector-src/blob/main/composer.json#L26

is it worth me handling this by adding a `$typeNode->templateTypes !== null &&` check on the if?

tests/PHPStan/Analyser/data/generic-callables.php
```diff
@@ -0,0 +1,80 @@
...
+	assertType('string', $callable($str));
+	assertType('callable(): 1', testFuncCallable(1));
+}
+
+/**
+ * @param Closure<TRetFirst of mixed>(TRetFirst $valone): (Closure<TRetSecond of mixed>(TRetSecond $valtwo): (TRetFirst|TRetSecond)) $closure
+ */
+function testNestedClosures(Closure $closure, string $str, int $int): void
+{
+	assertType('Closure<TRetFirst of mixed>(TRetFirst): (Closure<TRetSecond of mixed>(TRetSecond $valtwo): (TRetFirst|TRetSecond))', $closure);
```
mad-briller, line 75: the return type `Closure` does not have the parameter names removed as the removal is done in `describe()` not `toPhpDocNode()`, and describe is only called for the top-level. not sure if this is intended or not so thought i'd mention it

mad-briller:
https://github.com/phpstan/phpstan-src/pull/2938#pullrequestreview-1899509254

i can investigate adding these in a follow up pr if you'd like, though i'm a bit overwhelmed by the list :sweat_smile:

ondrejmirtes:
About https://github.com/phpstan/phpstan-src/pull/2938#pullrequestreview-1899509254 - yeah, sure, this can be addressed in multiple small follow-up PRs :) I'd really appreciate if you continued working on this :) (I have even more similar stuff for you if you want :))

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2939, opened 2024-02-23
Issue 10626: Do type checks on pass by reference parameters for non-builtin functions and methods.

ljmaskey:
Do type checks on pass by reference parameters for non-builtin functions and methods.

Initially, all pass by reference parameters were skipped for type checking. This change continues to skip those checks for builtin PHP methods, but does check parameters for userland methods.

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -278,7 +278,7 @@ class FunctionCallParametersCheck
 			if ($this->checkArgumentTypes) {
 				$parameterType = TypeUtils::resolveLateResolvableTypes($parameter->getType());
 
-				if (!$parameter->passedByReference()->createsNewVariable()) {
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Yes, please write a regression test for 8296 😊 Thanks.

ljmaskey:
Just to be a pain, I've added the test for 8296, but opened the PR against 1.10.x instead. That's at https://github.com/phpstan/phpstan-src/pull/2941. If it makes sense to keep this one open, I can add the additional test to this PR, or we can just close this one and go with the new one against 1.10.x.

ondrejmirtes:
Hey, just a tip - you don't need to close a PR when you just want to change the base. The correct thing to do here would be to change the base in the GitHub UI to 1.10.x, rebase your commits locally on top of 1.10.x (git rebase --onto is a great way to do it) and force-push to the same branch. But you already opened a new PR so I'm going to close this one :)

PR #2940, opened 2024-02-24
check T of mixed&Foo and T of mixed|Foo

schlndh:
Fixes https://phpstan.org/r/a3548a66-d83b-46d4-b418-7dbc038cbeb1 (all var_dumps should be reported).
Fixes https://phpstan.org/r/7ec28c62-5411-412a-a4c8-c7f9253b918d (there should be actual type instead of stdClass).

### DIFF
--- a/src/Rules/RuleLevelHelper.php
+++ b/src/Rules/RuleLevelHelper.php
@@ -10,6 +10,7 @@ use PHPStan\Type\CallableType;
 use PHPStan\Type\ClosureType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\Generic\TemplateMixedType;
+use PHPStan\Type\IntersectionType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\NullType;
@@ -303,6 +304,20 @@ class RuleLevelHelper
 			return new FoundTypeResult(new ErrorType(), [], [], null);
 		}
 		$type = $scope->getType($var);
+
+		return $this->findTypeToCheckImplementation($scope, $var, $type, $unknownClassErrorPattern, $unionTypeCriteriaCallback, true);
+	}
+
+	/** @param callable(Type $type): bool $unionTypeCriteriaCallback */
+	private function findTypeToCheckImplementation(
+		Scope $scope,
+		Expr $var,
+		Type $type,
+		string $unknownClassErrorPattern,
+		callable $unionTypeCriteriaCallback,
+		bool $isTopLevel = false,
+	): FoundTypeResult
+	{
 		if (!$this->checkNullables && !$type->isNull()->yes()) {
 			$type = TypeCombinator::removeNull($type);
 		}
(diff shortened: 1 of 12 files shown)


src/Rules/RuleLevelHelper.php
```diff
@@ -345,27 +360,33 @@ public function findTypeToCheck(
 		if ($type instanceof MixedType || $type instanceof NeverType) {
 			return new FoundTypeResult(new ErrorType(), [], [], null);
 		}
-		if ($type instanceof StaticType) {
-			$type = $type->getStaticObjectType();
+		if (!$this->newRuleLevelHelper) {
+			if ($isTopLevel && $type instanceof StaticType) {
+				$type = $type->getStaticObjectType();
```
schlndh, line 365: This line hid the following issue (`CallMethodsRuleTest`):
```
Parameter #1 $other of method Test\CollectionWithStaticParam::add() expects static(Test\AppleCollection), Test\AppleCollection given.
```

ondrejmirtes:
1) Looks like this fixed https://github.com/phpstan/phpstan/issues/9009, please add a regression test :)
2) There's a levels test failure https://github.com/phpstan/phpstan-src/actions/runs/8030898804/job/21938683114, you can just re-run that locally and commit the updated JSON file :)

Thank you very much!

tests/PHPStan/Rules/Methods/CallStaticMethodsRuleTest.php
```diff
@@ -250,7 +250,7 @@ public function testCallStaticMethods(): void
 				'Learn more at https://phpstan.org/user-guide/discovering-symbols',
 			],
 			[
-				'Call to an undefined static method static(CallStaticMethods\CallWithStatic)::nonexistent().',
```
schlndh, line 253: I tried to minimize breaking the previous error messages, and this is the result that I ended up with. IMO `static(...)` is unnecessary information here - the `nonexistent` method has to exist on `CallWithStatic` for the issue to go away.

Should I limit this change to bleeding edge only (e.g. via `newRuleLevelHelper`)?

ondrejmirtes: I think it's fine as it is 😊

schlndh:
I thought that I caused a false positive here: https://github.com/drupal/drupal/blob/5c30729a8365dc7e7d649d597a11f170eecff26c/core/modules/migrate/tests/src/Kernel/MigrateTestBase.php#L173
https://github.com/phpstan/phpstan-src/actions/runs/8031005259/job/21938938357?pr=2940

```
  ------ ------------------------------------------------------------------ 
  Line   core/modules/migrate/tests/src/Kernel/MigrateTestBase.php         
 ------ ------------------------------------------------------------------ 
  173    Call to an undefined static method                                
         Drupal\Tests\migrate\Kernel\MigrateTestBase::migrateDumpAlter().  
 ------ ------------------------------------------------------------------ 
```

But it looks like the issue is already present in phpstan, I just changed the error message: https://phpstan.org/r/1475efe6-5aaf-4922-bab9-c4ebacc8d24c

ondrejmirtes:
Thank you!

PR #2941, opened 2024-02-24
Check types when argument is passed by reference

ljmaskey:
Do type checks on pass by reference parameters for non-builtin functions and methods.

Initially, all pass by reference parameters were skipped for type checking. This change continues to skip those checks for builtin PHP methods, but does check parameters for userland methods.

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-10626.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-10626.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace PassedByReference;
+
+function intByValue(int $value): void
+{
+
+}
+
+function intByReference(int &$value): void
+{
+
+}
+
+$notAnInt = 'not-an-int';
+intByValue($notAnInt);
+intByReference($notAnInt);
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/Functions/data/bug-10626.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace PassedByReference;
```
ondrejmirtes: This needs to have "Bug10626" namespace

tests/PHPStan/Rules/Methods/data/bug-8296.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+class VerifyLoginTask{
```
ondrejmirtes, line 5: This needs to have Bug8296 namespace

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2942, opened 2024-02-24
Use `Type::isNull()` in `Nullsafe(MethodCall|PropertyFetch)Rule`

takaram:
This makes the logic simpler.


ondrejmirtes:
Thank you!

PR #2945, opened 2024-02-26
Add rule to check for existing classes, type aliases and shadowing of CallableType and ClosureType templates on methods and functions.

mad-briller:
follow up to: https://github.com/phpstan/phpstan-src/pull/2938#pullrequestreview-1899509254

covers `@param` `@param-out` and `@return` on functions / methods

### DIFF
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -927,13 +927,13 @@ class TypeNodeResolver
 		$returnType = $this->resolve($typeNode->returnType, $nameScope);
 
 		if ($mainType instanceof CallableType) {
-			return new CallableType($parameters, $returnType, $isVariadic, $templateTypeMap);
+			return new CallableType($parameters, $returnType, $isVariadic, $templateTypeMap, null, $templateTags);
 
 		} elseif (
 			$mainType instanceof ObjectType
 			&& $mainType->getClassName() === Closure::class
 		) {
-			return new ClosureType($parameters, $returnType, $isVariadic, $templateTypeMap);
+			return new ClosureType($parameters, $returnType, $isVariadic, $templateTypeMap, null, null, $templateTags);
 		}
 
 		return new ErrorType();
--- a/src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
+++ b/src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
@@ -31,6 +31,7 @@ class IncompatiblePhpDocTypeRule implements Rule
 		private FileTypeMapper $fileTypeMapper,
 		private GenericObjectTypeCheck $genericObjectTypeCheck,
 		private UnresolvableTypeHelper $unresolvableTypeHelper,
+		private GenericCallableRuleHelper $genericCallableRuleHelper,
 	)
 	{
 	}
@@ -137,6 +138,16 @@ class IncompatiblePhpDocTypeRule implements Rule
 						),
 					));
 
+					$errors = array_merge($errors, $this->genericCallableRuleHelper->check(
+						$node,
+						$scope,
+						sprintf('%s for parameter $%s', $escapedTagName, $escapedParameterName),
+						$phpDocParamType,
+						$functionName,
+						$resolvedPhpDoc->getTemplateTags(),
+						$scope->isInClass() ? $scope->getClassReflection() : null,
+					));
+
 					if ($phpDocParamTag instanceof ParamOutTag) {
 						if (!$byRefParameters[$parameterName]) {
 							$errors[] = RuleErrorBuilder::message(sprintf(
@@ -215,6 +226,16 @@ class IncompatiblePhpDocTypeRule implements Rule
 
 					$errors[] = $errorBuilder->build();
 				}
+
+				$errors = array_merge($errors, $this->genericCallableRuleHelper->check(
+					$node,
+					$scope,
+					'@return',
+					$phpDocReturnType,
+					$functionName,
+					$resolvedPhpDoc->getTemplateTags(),
+					$scope->isInClass() ? $scope->getClassReflection() : null,
+				));
 			}
 		}
 
--- a/src/Type/ClosureType.php
+++ b/src/Type/ClosureType.php
@@ -5,6 +5,7 @@ namespace PHPStan\Type;
 use Closure;
 use PHPStan\Analyser\OutOfClassScope;
 use PHPStan\Php\PhpVersion;
+use PHPStan\PhpDoc\Tag\TemplateTag;
 use PHPStan\PhpDocParser\Ast\PhpDoc\TemplateTagValueNode;
 use PHPStan\PhpDocParser\Ast\Type\CallableTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\CallableTypeParameterNode;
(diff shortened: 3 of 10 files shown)


src/PhpDoc/TypeNodeResolver.php
```diff
@@ -927,13 +927,13 @@ function (CallableTypeParameterNode $parameterNode) use ($nameScope, &$isVariadi
 		$returnType = $this->resolve($typeNode->returnType, $nameScope);
 
 		if ($mainType instanceof CallableType) {
-			return new CallableType($parameters, $returnType, $isVariadic, $templateTypeMap);
+			return new CallableType($parameters, $returnType, $isVariadic, $templateTypeMap, null, $templateTags);
```
mad-briller, line 930: without passing the original `$templateTags` into callable / closure type, it was impossible to report unsupported bounds as they are turned into `TemplateMixedType` by `$this->resolve()` in the TypeNodeResolver

src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
```diff
@@ -137,6 +140,20 @@ public function processNode(Node $node, Scope $scope): array
 						),
 					));
 
+					if ($phpDocParamType instanceof CallableType || $phpDocParamType instanceof ClosureType) {
```
mad-briller: these lines cause phpstan to report an error when analysing itself but i wasn't sure of a safe alternative, given i've had to add `getTemplateTags` to these 2 types, and it's not safe to assume that `ParametersAcceptor` will have these methods so i cannot use `getCallableParametersAcceptors()`

ondrejmirtes: Yes, I agree with this

ondrejmirtes: But you need to regenerate the baseline :)

mad-briller: thought as much, just wanted to give you a chance to tell me if i was wrong

src/Type/ClosureType.php
```diff
@@ -75,6 +76,7 @@ public function __construct(
 		?TemplateTypeMap $templateTypeMap = null,
 		?TemplateTypeMap $resolvedTemplateTypeMap = null,
 		?TemplateTypeVarianceMap $callSiteVarianceMap = null,
```
ondrejmirtes, line 78: Little suspicious here - why do we need `$callSiteVarianceMap` in ClosureType but not in CallableType? Looks like an omission :)

src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
ondrejmirtes: Hey sorry, I was on my phone and thought we're in TemplateTypeHelper::toArgument (there's also `if ($type instanceof ParametersAcceptor) {`).

This has the same problem we've had there - this will only work with top-most type. We need to employ TypeTraverser again. (In TypeTraverser calls `instanceof CallableType` is allowed so baseline can be cleared again).

For example if you check what `$this->unresolvableTypeHelper->containsUnresolvableType($phpDocReturnType)` does, it looks for types recursively.

We should do a similar thing here :)

ondrejmirtes (changes requested):

src/Type/ClosureType.php
mad-briller: good question; can't say i understand the intention of the `TemplateTypeVarianceMap` to be honest, but i can add it as an accepted parameter to CallableType

ondrejmirtes: Please check the commit where it was added, and what it does in ClosureType, that should give you some context - and do the same in CallableType :)

ondrejmirtes: (I'm also not sure, this stuff was done by Jiří Pudil, which I contact every time I need help with variance :))

src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
mad-briller: ah yeah of course, great catch

src/Type/ClosureType.php
mad-briller: reading https://phpstan.org/blog/guide-to-call-site-generic-variance i'm not sure call-site variance would apply to ClosureType / CallableType unless ClosureType is created as a firstClassCallable, so it's correct for the variance map to be empty, unless @jiripudil has wisdom for us? :angel: 

regardless, i've added the variance map as a constructor arg to the CallableType aswell for completeness

src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
```diff
@@ -137,6 +141,24 @@ public function processNode(Node $node, Scope $scope): array
 						),
 					));
 
+					TypeTraverser::map($phpDocParamType, function (Type $type, callable $traverse) use ($escapedTagName, $escapedParameterName, $node, $scope, &$errors, $functionName, $resolvedPhpDoc): Type {
+						if ($type instanceof CallableType || $type instanceof ClosureType) {
+							$classReflection = $scope->isInClass() ? $scope->getClassReflection() : null;
+							$errors = array_merge($errors, $this->genericCallableRuleHelper->check(
```
ondrejmirtes: This could just be this with the TypeTraverser call done inside, right?

```php

							$errors = array_merge($errors, $this->genericCallableRuleHelper->check(
								$node,
								$scope,
								sprintf('%s for parameter $%s', $escapedTagName, $escapedParameterName),
								$type,
								$functionName,
								$resolvedPhpDoc->getTemplateTags(),
								$classReflection,
								$classReflection !== null ? $classReflection->getTemplateTags() : [],
							));
```

src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
```diff
@@ -215,6 +237,24 @@ public function processNode(Node $node, Scope $scope): array
 
 					$errors[] = $errorBuilder->build();
 				}
+
+				TypeTraverser::map($phpDocReturnType, function (Type $type, callable $traverse) use ($node, $scope, &$errors, $functionName, $resolvedPhpDoc): Type {
+					if ($type instanceof CallableType || $type instanceof ClosureType) {
+						$classReflection = $scope->isInClass() ? $scope->getClassReflection() : null;
+						$errors = array_merge($errors, $this->genericCallableRuleHelper->check(
```
ondrejmirtes: Same here

ondrejmirtes (changes requested):

src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php
mad-briller: yeah it can, short sighted of me :sweat_smile:

src/Type/ClosureType.php
jiripudil: Hi, that's right, ClosureType is only created with the call-site variance map when constructed from callable, and even then it's more for the sake of completeness than anything else.

In hindsight, it might actually be an oversight that it remained there: since I had to move the getCallSiteVarianceMap() method to a different interface for BC reasons (https://github.com/phpstan/phpstan-src/pull/2485/commits/107716730a1c1b1dd76fdc9b0aaa2950dd593d40), it will only be passed around the ClosureTypes but will never be called to do anything useful :|

ondrejmirtes: @jiripudil Feel free to clean it up :) Also I'd like to pick your brain on how generic closures should be accepted by other generic and non-generic closures. We currently don't have this logic because we didn't need it...

mad-briller: should i revert the change to add the variance map to CallableType then ?

ondrejmirtes: Yes, looks like it :)

ondrejmirtes:
Perfect, thank you! Looking forward to next steps :)

mad-briller:
> Perfect, thank you! Looking forward to next steps :)

Already got native properties covered just need to rebase when I'm at home

PR #2946, opened 2024-02-27
Add rule to check for existing classes, type aliases and shadowing of generic callables in properties.

mad-briller:
another follow up to: https://github.com/phpstan/phpstan-src/pull/2938#pullrequestreview-1899509254

covers native properties

### DIFF
--- a/tests/PHPStan/Rules/PhpDoc/data/generic-callables-incompatible.php
+++ b/tests/PHPStan/Rules/PhpDoc/data/generic-callables-incompatible.php
@@ -1,4 +1,4 @@
-<?php declare(strict_types=1);
+<?php declare(strict_types=1); // lint > 8.0
 
 namespace GenericCallablesIncompatible;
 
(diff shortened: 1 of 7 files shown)


mad-briller:
is there a way to disable the `Lint` action for a file on a specific php version or do i have to exclude it for all / make the test file compatible with all?

tests/PHPStan/Rules/PhpDoc/data/generic-callables-incompatible.php
```diff
@@ -191,3 +191,14 @@ function shadowsParamOutArray(array &$existingClasses): void
 function shadowsReturnArray(): array
 {
 }
+
```
ondrejmirtes, line 194: Do `// lint >= 7.4` at the top of the file, see many comments like that in tests files :)

mad-briller: good to know, dunno how i never noticed those before

ondrejmirtes:
Thank you!

mad-briller:
Currently without a dev env to continue on with the other steps, I'll try to follow up when I get one sorted

PR #2947, opened 2024-02-28
Correct return type of date_sun_info() function

zonuexe:
Prior to PHP 8.0, [`date_sun_info()`](https://www.php.net/manual/en/function.date-sun-info.php) function could return `false`. However, this only applies if the function is called with the wrong arguments. Therefore, the PHP manual only described the return value prototype as `array`.

- https://3v4l.org/l5IFm

It seems that the use cases to which benevolent union should be applied in PHPStan are not yet documented, but this function may be one of them.

refs https://github.com/php/doc-en/pull/3207


ondrejmirtes:
Thank you!

PR #2948, opened 2024-02-28
Use narrowed conditional type if/else types for subtype checks

rvanvelzen:
Fix phpstan/phpstan#10622


ondrejmirtes:
Perfect, thank you!

PR #2949, opened 2024-02-28
Fix `isSuperTypeOf` of callable/Closure when parameter count differ

takaram:
Fixes phpstan/phpstan#10594


ondrejmirtes:
Awesome, thank you!

PR #2950, opened 2024-02-29
Add support for constructor assertions

axlon:
Enables the following:

```php
class Person
{
    /** @phpstan-assert non-empty-string $firstName */
    public function __construct(string $firstName) {}
}

$person = new Person($firstName);
// $firstName can no longer be empty
```

Fixes https://github.com/phpstan/phpstan/issues/10645 (partially)

### DIFF
--- a/tests/PHPStan/Analyser/data/assert-constructor.php
+++ b/tests/PHPStan/Analyser/data/assert-constructor.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace AssertConstructor;
+
+use function PHPStan\Testing\assertType;
+
+class Person
+{
+	/**
+	 * @phpstan-assert non-empty-string $firstName
+	 */
+	public function __construct(
+		public string $firstName,
+	) {
+		assert($firstName !== '');
+	}
+}
+
+function (string $firstName) {
+	new Person($firstName);
+	assertType('non-empty-string', $firstName);
+};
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/data/assert-constructor.php
```diff
@@ -0,0 +1,24 @@
...
+	 * @phpstan-assert non-empty-string $firstName
+	 */
+	public function __construct(
+		public string $firstName,
+	) {
+		assert($firstName !== '');
+	}
+}
+
+/**
```
ondrejmirtes: Move this into anonymous function and use a string parameter coming into the function.

ondrejmirtes (changes requested):
What do you mean by "partially"?

axlon:
> What do you mean by "partially"?

@ondrejmirtes This PR currently does not tackle assertions on promoted properties, so for example:

```php
class Person
{
    /** @phpstan-assert non-empty-string $this->firstName */
    public function __construct(public string $firstName) {}
}

$person = new Person($firstName);
dumpType($person->firstName); // non-empty-string
```

I'm working on this as well but it needs a little more time

ondrejmirtes:
This use case does not make sense to me, you can type the property to be non-empty-string from the start, no need to assert it.

axlon:
The use case is being able to annotate promoted props to be a narrowed version of the variable going in. I skipped some statements in my original example for brevity but there was supposed to be an assert call in the constructor asserting the promoted prop was not empty.

Essential this, but with promoted props:

```php
class Person
{
    /**
     * @var non-empty-string
     */
    public string $firstName;

    public function __construct(string $firstName)
    {
        assert($firstName !== '');
        $this->firstName = $firstName;
    }
}
```

I am aware I could also annotate the class with `@property` as well, so its not really a big deal if asserting `$this->firstName` from a constructor is a no go 👍

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #2951, opened 2024-03-01
Prevent ShouldNotHappenException in filter_* extensions

staabm:
test in https://github.com/phpstan/phpstan/pull/10661

closes https://github.com/phpstan/phpstan/issues/10483


src/Type/Php/FilterFunctionReturnTypeHelper.php
```diff
@@ -240,7 +242,7 @@ private function getConstant(string $constantName): int
 		$constant = $this->reflectionProvider->getConstant(new Node\Name($constantName), null);
 		$valueType = $constant->getValueType();
 		if (!$valueType instanceof ConstantIntegerType) {
-			throw new ShouldNotHappenException(sprintf('Constant %s does not have integer type.', $constantName));
+			return self::UNKNOWN_CONSTANT;
```
staabm, line 243: I thought about returning null, or catching this exception (or a new exception) at the call-sites, but this seemed too much for this case. instead I decided to return a negative constant value, which does not overlap with other existing FILTER_* constants

ondrejmirtes (changes requested):
What's the effect of this on the type inference? Please include the failing test here in this PR too. So we can build on it and verify the actual returned type by this extension.

staabm:
(I need to verify on a different computer with PHP 7.2 installed whether the test is working as intended)

.github/workflows/e2e-tests.yml
```diff
@@ -202,6 +202,8 @@ jobs:
 
     strategy:
       matrix:
+        php-version: ["8.1"]
```
ondrejmirtes: This doesn't do what you think it does. See: https://github.com/phpstan/phpstan-src/actions/runs/8159444153/job/22303754862 - you basically disabled all of the tests.

Normal run: https://github.com/phpstan/phpstan-src/actions/runs/8170798487

ondrejmirtes (changes requested):

.github/workflows/e2e-tests.yml
staabm: good catch. fixed it.

thank you

e2e/bug10483/src/bug10483.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+function doFoo(mixed $filter): void {
+    \PHPStan\Testing\assertType('non-falsy-string|false', filter_var("no", FILTER_VALIDATE_REGEXP));
```
ondrejmirtes, line 4: Alright so I found out that the usual result on PHP versions where this constant exists is the same as here. Which makes me a bit suspicious. Can you explain that? We surely must be losing information when we return the unknown constant somewhere, right?

ondrejmirtes (changes requested):

ondrejmirtes:
Alright, and now add the same test to the usual NodeScopeResolverTest, so that the difference is obvious 😊 You'll most likely need different files for different PHP versions?

staabm:
reverted the previous test commit.

I thnk there is no type inference difference to observe, because phpstan internally works with the same "wrong" constant value as the code beeing analyzed (because the constant is defined in a bootstrap.php script, therefore mangels the global scope of the whole process).

the only thing we need to prevent is the crash

staabm:
I got one idea for which I am not sure whether its worth testing. the type inference might be different when we define the constant in bootstrap with a integer value of another existing FILTER_ constant. that way we would trigger an overlap of 2 constants.

but thats not something the initial case we try to fix triggers

PR #2952, opened 2024-03-01
Fix param-out generics via stubs

staabm:
triggered by discussion in https://github.com/phpstan/phpstan-src/commit/ee737e6df6cc73c845515ffb863aa5c62b59a34f#commitcomment-138984141

### DIFF
--- a/tests/PHPStan/Analyser/data/param-out.php
+++ b/tests/PHPStan/Analyser/data/param-out.php
@@ -501,3 +501,34 @@ function ($s, $t): void {
 	assertType('string', $s);
 	assertType('string', $t);
 };
+
+// typed via phpdoc in stub file
+function rsort(array &$array, int $flags = SORT_REGULAR): bool
+{
+}
+
+function ($a): void {
+	\ParamOut\rsort($a);
+	assertType('list<mixed>', $a);
+};
+/**
+ * @param string[] $a
+ */
+function ($a): void {
+	\ParamOut\rsort($a);
+	assertType('list<string>', $a);
+};
+/**
+ * @param non-empty-array<string> $a
+ */
+function ($a): void {
+	\ParamOut\rsort($a);
+	assertType('non-empty-list<string>', $a);
+};
+/**
+ * @param non-empty-list<string> $a
+ */
+function ($a): void {
+	\ParamOut\rsort($a);
+	assertType('non-empty-list<string>', $a);
+};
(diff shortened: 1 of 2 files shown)


staabm:
@rvanvelzen may I ask you for your opinion on these conditional types involved, as I think you are the expert :)?

---

I am wondering why we resolve `@param` differently compared to `@param-out` 

https://github.com/phpstan/phpstan-src/blob/39dc97bd9f64bfd16bd974d510b090ae4776aeaf/src/Reflection/ResolvedFunctionVariant.php#L75-L83

vs.

https://github.com/phpstan/phpstan-src/blob/39dc97bd9f64bfd16bd974d510b090ae4776aeaf/src/Reflection/ResolvedFunctionVariant.php#L85-L93

harmonizing this does not yet make a difference though, as the conditional types involved are not `ConditionalTypeForParameter` for `param-out`. so I am wondering whether phpdoc-parser needs adjustments to produce `ConditionalTypeForParameter` instead of `ConditionalType` for `param-out`...?

rvanvelzen:
The reason the added tests fail is because anonymous functions don't support phpdocs, so the parameter type is never set correctly.

Why `sort` etc. didn't work before is because coming from the signature map they're defined as `PassedByReference::readsArgument()` instead of `PassedByReference::createsNewVariable()`, so the out-type is never used.

As for the conditional parameters: it probably hasn't been a use-case for now. I didn't implement param-out types 😉

VincentLanglet:
What was the issue you were trying to use @staabm 
rsort seems to handle list/non-empty https://phpstan.org/r/3a1f2bc6-5d8d-4662-937a-f3b10c84fe43
should we close this ?

Also param out seems to support conditional https://github.com/phpstan/phpstan-src/blob/d393a1d7f99cb74c42fced0be94e5baf69c40d81/src/Reflection/ResolvedFunctionVariantWithOriginal.php#L89

staabm:
@VincentLanglet I think motivation is the comments in https://github.com/phpstan/phpstan-src/commit/ee737e6df6cc73c845515ffb863aa5c62b59a34f#commitcomment-138984141

VincentLanglet:
> @VincentLanglet I think motivation is the comments in [ee737e6#commitcomment-138984141](https://github.com/phpstan/phpstan-src/commit/ee737e6df6cc73c845515ffb863aa5c62b59a34f#commitcomment-138984141)

Then I think it can be close because the issue is solved for both rsort and custom method with generic conditional param out 
https://phpstan.org/r/95c69276-87cf-4a5b-b436-fd571c5b0255

PR #2953, opened 2024-03-03
Report uses of deprecated constants

staabm:
implement reflection for constants, to support use of `@deprecated` on const definitions

### DIFF
--- a/src/Reflection/Constant/RuntimeConstantReflection.php
+++ b/src/Reflection/Constant/RuntimeConstantReflection.php
@@ -10,6 +10,9 @@ use PHPStan\Type\Type;
 class RuntimeConstantReflection implements GlobalConstantReflection
 {
 
+	/**
+	 * @param non-empty-string|null $docComment
+	 */
 	public function __construct(
 		private string $name,
 		private Type $valueType,
--- a/src/Reflection/Constant/RuntimeConstantReflection.php
+++ b/src/Reflection/Constant/RuntimeConstantReflection.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Reflection\Constant;
 
+use PHPStan\BetterReflection\Reflection\Annotation\AnnotationHelper;
 use PHPStan\Reflection\GlobalConstantReflection;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Type;
@@ -13,6 +14,7 @@ class RuntimeConstantReflection implements GlobalConstantReflection
 		private string $name,
 		private Type $valueType,
 		private ?string $fileName,
+		private ?string $docComment,
 	)
 	{
 	}
@@ -34,12 +36,18 @@ class RuntimeConstantReflection implements GlobalConstantReflection
 
 	public function isDeprecated(): TrinaryLogic
 	{
-		return TrinaryLogic::createNo();
+		return TrinaryLogic::createFromBoolean(AnnotationHelper::isDeprecated($this->getDocComment()));
 	}
 
 	public function getDeprecatedDescription(): ?string
 	{
 		return null;
+	}
+
+	/** @return non-empty-string|null */
+	public function getDocComment(): ?string
+	{
+		return $this->docComment;
 	}
 
 	public function isInternal(): TrinaryLogic
--- a/src/Reflection/Constant/RuntimeConstantReflection.php
+++ b/src/Reflection/Constant/RuntimeConstantReflection.php
@@ -2,13 +2,18 @@
 
 namespace PHPStan\Reflection\Constant;
 
-use PHPStan\BetterReflection\Reflection\Annotation\AnnotationHelper;
+use PHPStan\PhpDoc\ResolvedPhpDocBlock;
 use PHPStan\Reflection\GlobalConstantReflection;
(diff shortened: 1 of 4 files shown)


staabm:
build errors should be adressed with https://github.com/phpstan/phpstan-deprecation-rules/pull/112

I was not yet able to test it properly, as I did not yet found a way to use this phpstan-src patch in the deprecation-rules repo

tests/PHPStan/Reflection/Constant/RuntimeConstantReflectionTest.php
```diff
@@ -0,0 +1,47 @@
...
+				TrinaryLogic::createYes(),
+				"don't use it!",
+			],
+		];
+	}
+
+	/**
+	 * @dataProvider dataDeprecatedConstants
+	 */
+	public function testDeprecatedConstants(Name $constName, TrinaryLogic $isDeprecated, ?string $deprecationMessage): void
```
herndlm, line 48: hi, the dataprovider usage is confusing for me. there is explicitly only one file and if that is changed the single array entry in the dataprovider would need to be updated, right? but is there ever a use case where a second array entry would be added? it feels weird to me, or I'm just not getting it :)

staabm: I had no dataprovider before and refactored it to what it is now, as it felt repetitive.

With the current structure one could add a new const definition into the test-file and add expectations for this file in the dataprovider

herndlm: Ah, I got it. I'm so used to using yield in those that the array nesting confused me I think and I believed it would execute the test only once. Makes sense.

src/Reflection/Constant/RuntimeConstantReflection.php
```diff
@@ -2,17 +2,28 @@
...
+	private ?string $deprecatedDescription = null;
+
+	/**
+	 * @param non-empty-string|null $docComment
+	 */
 	public function __construct(
 		private string $name,
 		private Type $valueType,
 		private ?string $fileName,
+		private FileTypeMapper $fileTypeMapper,
```
ondrejmirtes: I'd rather be if this constructor only accepted `bool $isDeprecated` and `?string $deprecatedDescription`. They should be resolved where this object is created.

tests/PHPStan/Reflection/Constant/RuntimeConstantReflectionTest.php
```diff
@@ -0,0 +1,47 @@
...
+				TrinaryLogic::createNo(),
+				null,
+			],
+			[
+				new Name('\DeprecatedConst\MY_CONST'),
+				TrinaryLogic::createYes(),
+				null,
+			],
+			[
+				new Name('\DeprecatedConst\MY_CONST2'),
```
ondrejmirtes: You should test some deprecated constants from phpstorm-stubs, like `CURLOPT_FTP_SSL`.

ondrejmirtes: Also test `FILTER_SANITIZE_STRING` which is deprecated on 8.1+.

ondrejmirtes (changes requested):

tests/PHPStan/Reflection/Constant/RuntimeConstantReflectionTest.php
```diff
@@ -0,0 +1,59 @@
...
+class RuntimeConstantReflectionTest extends PHPStanTestCase
+{
+
+	public function dataDeprecatedConstants(): iterable
+	{
+		if (PHP_VERSION_ID >= 80100) {
+			yield [
+				new Name('\FILTER_SANITIZE_STRING'),
+				TrinaryLogic::createYes(),
+				'8.1', // deprecation message used in e.g. https://github.com/JetBrains/phpstorm-stubs/blob/9608c953230b08f07b703ecfe459cc58d5421437/filter/filter.php#L478
```
ondrejmirtes: Would be great to get rid of it, right. If the message === X.Y or X.Y.Z, we can remove it using regex

ondrejmirtes (changes requested):

tests/PHPStan/Reflection/Constant/RuntimeConstantReflectionTest.php
staabm: filtered it out using regex. 

looking into the CI errors https://github.com/phpstan/phpstan-src/actions/runs/8152892522/job/22283354469?pr=2953 I got the idea, we could also turn the deprecation-message into `Use of constant %s is deprecated since PHP %s.` instead of removing it completly. wdyt?

ondrejmirtes: >  we could also turn the deprecation-message into Use of constant %s is deprecated since PHP %s. instead of removing it completly. wdyt?

We don't do this for any other feature, I don't think it's worth it. Passing different `phpVersion` or simply analysing it on phpstan.org/try makes it obvious what's going on.

ondrejmirtes:
Thank you.

PR #2957, opened 2024-03-06
implement strlen in the same way as mb_strlen

is-hoku:
resolve https://github.com/phpstan/phpstan/issues/10677


ondrejmirtes:
Thank you!

PR #2958, opened 2024-03-07
Fix return type of mysqli_result::fetch_array + fetch_assoc

paranoiq:
https://www.php.net/manual/en/mysqli-result.fetch-array.php 
https://www.php.net/manual/en/mysqli-result.fetch-assoc.php


janedbal:
Just met this, should be easy to merge imo :)

ondrejmirtes:
There's a conflict, can you please fix it?

Also: note that the type on PHP 8+ is already correct :) https://phpstan.org/r/7868d592-23fb-4a4b-8145-ebc89ca74bb4

janedbal:
Looks like this was fixed in https://github.com/phpstan/phpstan-src/pull/3279

ondrejmirtes:
Yeah :)

PR #2959, opened 2024-03-07
Ensure http_response_header is a list<string>

Seldaek:
Not sure if that's the proper/best use of AccessoryArrayListType, but anyway I hope the intent is clear :)

### DIFF
--- a/tests/PHPStan/Analyser/data/http-response-header.php
+++ b/tests/PHPStan/Analyser/data/http-response-header.php
@@ -1,0 +1,6 @@
+<?php
+
+use function PHPStan\Testing\assertType;
+
+file_get_contents('https://example.org');
+assertType('list<string>', $http_response_header);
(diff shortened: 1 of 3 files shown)


staabm:
since no test is failling after the change it seems we are missing a NodeScopeResolverTest with a type inference assertion for this variable :)

staabm (approved):
lgtm - thank you

tests/PHPStan/Analyser/data/http-response-header.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+use function PHPStan\Testing\assertType;
+
+file_get_contents('https://example.org');
+assertType('list<string>', $http_response_header);
```
staabm, line 7: ohh I think you are missing another assert for `assertNativeType`

Seldaek: Ok hopefully I did this correctly :)

ondrejmirtes:
Thank you!

PR #2962, opened 2024-03-09
Fix Closure::fromCallable and first-class callables not propagating templates.

mad-briller:
Also improves the `array_map` return type extension to be smarter about generic callables, specifically for single argument constant arrays.

This improves the situation of https://github.com/phpstan/phpstan/issues/10685 but i hesitate to say it solves it, there are some underlying issues that i'll add to the relevant test locations on this pr

### DIFF
--- a/tests/PHPStan/Analyser/data/generic-callables.php
+++ b/tests/PHPStan/Analyser/data/generic-callables.php
@@ -78,3 +78,146 @@ function testNestedClosures(Closure $closure, string $str, int $int): void
 	$result = $closure1($int);
 	assertType('int|string', $result);
 }
+
+/**
+ * @template T
+ * @param T $arg
+ * @return T
+ */
+function foo(mixed $arg): mixed {}
+
+class Foo
+{
+	/**
+	 * @template T
+	 * @param T $arg
+	 * @return T
+	 */
+	public function foo(mixed $arg): mixed {}
+}
+
+function test(): void
+{
+	assertType('Closure<T of mixed>(T): T', foo(...));
+	assertType('1', foo(...)(1));
+
+	$foo = new Foo();
+	$closure = Closure::fromCallable([$foo, 'foo']);
+	assertType('Closure<T of mixed>(T): T', $closure);
+	assertType('1', $closure(1));
+}
+
+/**
+ * @template A
+ * @param A $value
+ * @return A
+ */
+function identity(mixed $value): mixed
+{
+	return $value;
+}
+
+/**
+ * @template B
+ * @param B $value
+ * @return B
+ */
+function identity2(mixed $value): mixed
+{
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Analyser/data/generic-callables.php
```diff
@@ -78,3 +78,146 @@ function testNestedClosures(Closure $closure, string $str, int $int): void
...
+
+	assertType('list<array{1, 2}>', $fn([1], [2]));
+}
+
+/**
+ * @template X
+ * @template Y
+ * @template Z
+ * @param callable(X $x, Y $y): Z $fn
+ * @return callable(Y $xy, X $yx): Z
```
mad-briller: without parameter names on the callable then the generic type resolution fails, because `GenericParametersAcceptorResolver` uses `$nativeParameterReflection->getName()` which returns `''` for unnamed parameters, and uses those names as array keys. If multiple parameters are unnamed this results in assigning to `$namedArgTypes['']` repeatedly and it falls apart

I've tried taking a stab at making `GenericParametersAcceptorResolver` care less about parameter names, but it's getting the better of me. Still trying to have a go at solving it but thought i'd put this up to possibly get some wisdom

tests/PHPStan/Analyser/data/generic-callables.php
```diff
@@ -78,3 +78,146 @@ function testNestedClosures(Closure $closure, string $str, int $int): void
...
+/**
+ * @template L
+ * @template M
+ * @template N
+ * @template O
+ * @param callable(L): M $ab
+ * @param callable(N): O $cd
+ * @return Closure(array{L, N}): array{M, O}
+ */
+function compose2(callable $ab, callable $cd): Closure
```
mad-briller, line 208: i had to change the generic type names of this function from `A` `B` `C` `D` to something else, as `identity()` and `identity2()` are also using `A` and `B` as generic type names, which made the type resolution fail. Been trying to dig into it a bit but as for why i'm really not sure right now,

tests/PHPStan/Analyser/data/generic-callables.php
mad-briller: scratch this, in true [rubber duck debugging](https://rubberduckdebugging.com/) fashion as i look at the code to describe the issue i had a wave of inspiration...

tests/PHPStan/Analyser/data/generic-callables.php
mad-briller: could this be related to `TemplateTypeScope`? i'm yet to properly understand it's role in the type resolution but it seems like it could be pertinent

VincentLanglet:
https://github.com/phpstan/phpstan/issues/10685 is now closed

PR #2963, opened 2024-03-09
Resolve PHPStan issue with deprecated parameter order in PHP 8.1

sayuprc:
Since PHP 8.1, a nullable, optional parameter with a default value of null followed by a required parameter will generate a warning.
The current PHPStan did not recognize such code as deprecated.
This PR updates PHPStan to properly handle such cases.
Please review.
Thank you.

resolve phpstan/phpstan#6668

### DIFF
--- a/src/Php/PhpVersion.php
+++ b/src/Php/PhpVersion.php
@@ -61,6 +61,11 @@ class PhpVersion
 		return $this->versionId >= 80100;
 	}
 
+	public function deprecatesRequiredParameterAfterOptionalUnionOrMixed(): bool
+	{
+		return $this->versionId >= 80300;
+	}
+
 	public function supportsLessOverridenParametersWithVariadic(): bool
 	{
 		return $this->versionId >= 80000;
--- a/src/Php/PhpVersion.php
+++ b/src/Php/PhpVersion.php
@@ -56,6 +56,11 @@ class PhpVersion
 		return $this->versionId >= 80000;
 	}
 
+	public function deprecatesRequiredParameterAfterOptionalNullableAndDefaultNull(): bool
+	{
+		return $this->versionId >= 80100;
+	}
+
 	public function supportsLessOverridenParametersWithVariadic(): bool
 	{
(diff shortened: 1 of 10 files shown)


ondrejmirtes (changes requested):
I'm worried this is a lot more complex topic. For example: Why is this reported only on PHP 8.3+? https://3v4l.org/e6n6Q

Also - your message says "Deprecated in 8.0" but in code your condition says `>= 80100`.

sayuprc:
Thank you for the review.

I was aware that this code was generating deprecated warnings from PHP 8.1. https://3v4l.org/0VAUk
It seems that there are differences in behavior with patterns like `int|null`, so I will investigate it again.

sayuprc:
@ondrejmirtes 
Since PHP 8.3, the deprecated conditions have changed, and warnings for union types and mixed types are also issued.

For this update, I am considering issuing warnings for nullable types deprecated since PHP 8.1 and for union types and mixed types deprecated since PHP 8.3.

I would like the message format to only replace the version number.
For example, in the case of `mixed $a = null`, it would display "Deprecated in PHP 8.3".

ondrejmirtes:
@sayuprc Yes, that would be okay 👍

ondrejmirtes:
Awesome, thank you!

ondrejmirtes:
I just noticed this doesn't make any tests fail - is that right? https://github.com/phpstan/phpstan-src/commit/6512a847f91e0dc905c1b434fbbdc9cd78e57122

sayuprc:
@ondrejmirtes 
If you remove `$targetPhpVersion = null;`, you won't be able to display the correct PHP version in the warning message for code like `fn (int|null $a = null, int $b, int $c = 0, $d): int => 1;`.

ondrejmirtes:
@sayuprc Please send a fix with a test as a new PR. Thank you.

sayuprc:
@ondrejmirtes 
I have created a new pull request. #2971

PR #2967, opened 2024-03-13
Constants deprecation depends on php version in comment

staabm:
since support for [`@deprecated` on constants was added into BetterReflection](https://github.com/Roave/BetterReflection/pull/1396) with this PR we now take care of interpreting the phpdoc comment and decide per phpversion when a constant is deprecated or not.

this also makes sure that the deprecation message depends on phpstan.neon `phpVersion`.

closes https://github.com/phpstan/phpstan/issues/10700

### DIFF
--- a/src/Reflection/BetterReflection/BetterReflectionProvider.php
+++ b/src/Reflection/BetterReflection/BetterReflectionProvider.php
@@ -373,9 +373,17 @@ class BetterReflectionProvider implements ReflectionProvider
 			if ($resolvedPhpDoc->isDeprecated() && $resolvedPhpDoc->getDeprecatedTag() !== null) {
 				$deprecatedMessage = $resolvedPhpDoc->getDeprecatedTag()->getMessage();
 
-				// filter raw version number messages like in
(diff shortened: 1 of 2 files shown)


src/Reflection/BetterReflection/BetterReflectionProvider.php
```diff
@@ -373,9 +373,17 @@ public function getConstant(Node\Name $nameNode, ?NamespaceAnswerer $namespaceAn
 			if ($resolvedPhpDoc->isDeprecated() && $resolvedPhpDoc->getDeprecatedTag() !== null) {
 				$deprecatedMessage = $resolvedPhpDoc->getDeprecatedTag()->getMessage();
 
-				// filter raw version number messages like in
-				// https://github.com/JetBrains/phpstorm-stubs/blob/9608c953230b08f07b703ecfe459cc58d5421437/filter/filter.php#L478
-				if (Strings::match($deprecatedMessage ?? '', '#^\d+\.\d+(\.\d+)?$#') === null) {
+				$matches = Strings::match($deprecatedMessage ?? '', '#^(\d+)\.(\d+)(?:\.(\d+))?$#');
```
staabm, line 376: this duplicates the logic which should already have happened in BetterReflection..  I think it is wrong

ondrejmirtes:
Thank you.

PR #2968, opened 2024-03-13
Utilize composer min-php version to narrow PHP_VERSION_ID

staabm:
idea is to catch errors like https://github.com/Roave/BetterReflection/pull/1406#discussion_r1523065225 in which bogus conditions like `if (PHP_VERSION_ID >= 80100) {}` have been used in tests. 

it's a bug in the linked case, because the project itself only supports `"php": "~8.2.0 || ~8.3.2",`.
Since CI only is running against in composer.json declared supported versions this error was not spotted before.


ondrejmirtes (changes requested):
Do the upper bound too - so that we know `^8.1` cannot be 9.

staabm:
@ondrejmirtes what do you think about all these 

> Comparison operation "<" between int<80100, max> and 80100 is always false.

errors? should we ignore them via phpstan.neon or should we add e.g. a feature flag, so we can disable the composer based min/max-php version?

ondrejmirtes:
It's not really generally usable. In PHPStan and many other apps, the constraint in composer.json does nor say anything about production environment, or can be outdated.

It could be:

* On in bleeding edge
* But at the same time there should be a way to turn it off, or make it configurable manually, like reading the range from phpstan.neon instead of composer.json. phpVersion behaves similarly, when it's null, we try to detect it.

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -671,7 +671,7 @@ public function testBug8068(): void
 
 	public function testBug6243(): void
 	{
-		if (PHP_VERSION_ID < 704000) {
```
staabm: found a bug in a pre-existing test -> https://github.com/phpstan/phpstan-src/pull/3254

build/ignore-by-php-version.neon.php
```diff
@@ -37,5 +37,7 @@
 $config = [];
 $config['includes'] = $includes;
 $config['parameters']['phpVersion'] = PHP_VERSION_ID;
+$config['parameters']['minPhpVersion'] = 70100;
+$config['parameters']['maxPhpVersion'] = 90000;
```
staabm: I put the min/max php version here, so we have a fixed upper/lower bound in phpstan-src, but all projects per default still use composer.json based auto-php-version bounds. I think even projects using bleeding-edge would benefit from automatic php version bounds.

the case that projects need to opt-out from this check will be rare I guess - but is possible via the new `phpstan.neon`  parameters by defining custom versions

staabm:
I think the errors we see in CI make sense. not sure what we can/should do about them. 

it make sense to check for php-versions smaller then the composer.json declared supported ones, to be super safe.
maybe I should use `composer-min-version - 1` as a lower bound?

_edit: did just that with [0d398ab](https://github.com/phpstan/phpstan-src/pull/2968/commits/0d398ab0c2355934716854f5402729f2bad175b1)_

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,114 @@
...
+
+		$parser = new VersionParser();
+		$constraint = $parser->parseConstraints($composerPhpVersion);
+
+		if ($this->minVersion === null && !$constraint->getLowerBound()->isZero()) {
+			$minVersion = $this->buildVersion($constraint->getLowerBound()->getVersion());
+			// allow checks with < below lower bound to prevent errors like
+			// 'Comparison operation "<" between int<70205, 90000> and 70205 is always false.'
+			if ($minVersion !== null) {
+				$this->minVersion = new PhpVersion($minVersion->getVersionId() - 1);
```
Seldaek: This subtraction seems very dodgy to me. If I get it correctly this is to fix warnings where `if (PHP_VERSION_ID < 70100)` where I have a require of php `>=7.1` for example. But why should it not warn? Because clearly per the requirement the code should not be allowed to run with a version lower than 70100.

Seldaek: Re-reading your comment above:

> it make sense to check for php-versions smaller then the composer.json declared supported ones, to be super safe.

Yes it may make sense in some cases, but then you want to explicitly override PHPStan and add this in the baseline IMO. In most cases I'd think you will want PHPStan to tell you that this check can be removed when you bump your PHP version requirement, which suddenly would not work anymore here due to the -1.

staabm: relevant examples from previous CI failures are

- https://github.com/phpstan/phpstan-src/actions/runs/8267804971/job/22619260364?pr=2968 (in composer)
- https://github.com/phpstan/phpstan-src/actions/runs/8267804971/job/22619308636?pr=2968 (in PrestaShop)

I am fine with leaving this error and let the projects put them into a baseline, but my thinking was, that it might be pretty common that code tries to detect whether it is invoked from a not supported php-version (so whether PHP_VERSION is below the lower limit of composer.json requirements)

Seldaek: Right yeah those two examples for Composer are definitely things I'd be ok ignoring if it gets me visibility on other things I forgot to delete.

staabm:
@ondrejmirtes keeping this one up2date is a bit cumbersome, because of the added composer dependency (and the conflicts I need to resolve on every other dependency update).

would be great if you could find the time to move it forward.

ondrejmirtes (changes requested):
Some thoughts about this PR:

1) There are more constants related to PHP versions: https://www.php.net/manual/en/reserved.constants.php
2) To be really useful, this should also influence `version_compare`: https://www.php.net/manual/en/function.version-compare.php A dynamic return type extension would be sufficient.
3) I'd like `phpVersion` to be either: `null`, a single number, or a structure with min/max

If it's null, resolve it to a single number in PHPStan 1.x (according to config.platform.php in composer.json), and to min/max structure with bleeding edge (according to require.php in composer.json).

If it's a single number, keep today's behaviour. Meaning PHP_VERSION_ID is not going to be limited.

If it's min/max, clamp PHP_VERSION_ID and other constants to be an IntegerRangeType.

Also we should have some kind of warning if the values don't make sense. max should be >= min.

staabm:
thanks for the feedback.

> If it's null, resolve it to a single number in PHPStan 1.x (according to config.platform.php in composer.json), and to min/max structure with bleeding edge (according to require.php in composer.json).
> 
> If it's a single number, keep today's behaviour. Meaning PHP_VERSION_ID is not going to be limited.
> 
> If it's min/max, clamp PHP_VERSION_ID and other constants to be an IntegerRangeType.

this means in case of a configured min/max version, we don't have a concrete version anymore in `PhpVersion`.
so all the support-methods don't work like they do today. since `PhpVersion` is tagged `@api` I cannot change it to return a `Trinary`..

https://github.com/phpstan/phpstan-src/blob/3e2eed92d7d3fb0b010e9b150dd56dcd96a39925/src/Php/PhpVersion.php#L29-L32

ondrejmirtes:
That's right. Changing all these methods to a TrinaryLogic would be a costly refactoring that would eventually lead to `ReflectionClass::hasClass()` also returning `TrinaryLogic`. I'd like that but not at this point.

Please try using the `min` value as the origin for these methods to decide whether to return true or false.

ondrejmirtes:
If that's not going to work for some reason then I'm not interested in this PR at this point, sorry. When investing in an area we need to do something that makes sense, not just a minimal amount of work that would change the interpretation of a single constant.

staabm:
Sure, lets see how it goes. 

Alternatively/Additionally I will look into a narrowed PhpVersion based on scope like we brainstormed in https://github.com/phpstan/phpstan/discussions/4212#discussioncomment-8803159

staabm:
I think all the feedback is adressed

e2e/composer-min-max-version/test.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+\PHPStan\Testing\assertType('int<80099, 80299>', PHP_VERSION_ID);
+\PHPStan\Testing\assertType('8', PHP_MAJOR_VERSION);
+\PHPStan\Testing\assertType('int<0, 2>', PHP_MINOR_VERSION);
```
mvorisek:
```suggestion
\PHPStan\Testing\assertType('int<80100, 80299>', PHP_VERSION_ID);
\PHPStan\Testing\assertType('8', PHP_MAJOR_VERSION);
\PHPStan\Testing\assertType('int<1, 2>', PHP_MINOR_VERSION);
```

AFAK `PHP_VERSION_ID` starts with the major/minor even if run on alpha release

`PHP_MINOR_VERSION` should be narrowed if `PHP_MAJOR_VERSION` is one constant number

staabm: See https://github.com/phpstan/phpstan-src/pull/2968#discussion_r1524840873

mvorisek: I saw it. But I do not support it. The reason is composer already check this during runtime and if the check is really, really wanted, the user should ignore that error instead.

staabm:
squashed, rebased against 1.11 and re-targetted the PR.

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,126 @@
...
+		bool $bleedingEdge,
+	)
+	{
+		if (is_int($phpVersion)) {
+			return;
+		}
+
+		if (is_array($phpVersion)) {
+			if ($phpVersion['max'] < $phpVersion['min']) {
+				throw new ShouldNotHappenException('Invalid PHP version range: phpVersion.max should be greater or equal to phpVersion.min.');
```
ondrejmirtes, line 43: I'd like to validate this sooner. ContainerFactory::validateParameters is the best place to do this.

We can keep this here but when someone puts the wrong value in phpstan.neon, PHPStan should output a nice error message like it does today with invalid parameter values.

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,126 @@
...
+	 * @param int|array{min: int, max: int}|null $phpVersion
+	 */
+	public function __construct(
+		private array $composerAutoloaderProjectPaths,
+		int|array|null $phpVersion,
+		bool $bleedingEdge,
+	)
+	{
+		if (is_int($phpVersion)) {
+			return;
```
ondrejmirtes, line 38: I think this if can be just removed.

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,126 @@
...
+		// fallback to composer.json based php-version constraint
+		$composerPhpVersion = $this->getComposerRequireVersion();
+		if ($composerPhpVersion === null) {
+			return;
+		}
+
+		$parser = new VersionParser();
+		$constraint = $parser->parseConstraints($composerPhpVersion);
+
+		if ($this->minVersion === null && !$constraint->getLowerBound()->isZero()) {
```
ondrejmirtes: How can `$this->minVersion` can be something else than null at this point? It doesn't make much sense to me.

src/Php/ComposerPhpVersionFactory.php
ondrejmirtes: Yeah, please remove the `-1`.

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,126 @@
...
+			$minVersion = $this->buildVersion($constraint->getLowerBound()->getVersion());
+			// allow checks with < below lower bound to prevent errors like
+			// 'Comparison operation "<" between int<70205, 90000> and 70205 is always false.'
+			if ($minVersion !== null) {
+				$this->minVersion = new PhpVersion($minVersion->getVersionId() - 1);
+			}
+		}
+		if ($this->maxVersion !== null || $constraint->getUpperBound()->isPositiveInfinity()) {
```
ondrejmirtes: How can `maxVersion` can be something else than `null` at this point?

src/Php/PhpVersionFactoryFactory.php
```diff
@@ -43,7 +49,20 @@ public function create(): PhpVersionFactory
...
 
-		return new PhpVersionFactory($this->versionId, $composerPhpVersion);
+		$versionId = null;
+
+		if (is_int($this->phpVersion)) {
+			$versionId = $this->phpVersion;
+		}
+
+		if ($this->bleedingEdge && is_array($this->phpVersion)) {
+			if (!array_key_exists('min', $this->phpVersion)) {
```
ondrejmirtes: This check doesn't have to be here, it's already enforced by the constructor array shape.

ondrejmirtes (changes requested):

staabm:
All feedback should be addressed.

> Yeah, please remove the `-1`.

~~because of this change we are now seeing a lot of errors in the test-suite, e.g.:~~

> Comparison operation "<" between int<80100, 90000> and 80100 is always false.

~~should these be ignored globally? any other idea on how to handle these?~~

**resolved with** [edde1b6](https://github.com/phpstan/phpstan-src/pull/2968/commits/edde1b6c4c7478d1b908be9183a7b19806cbfde8), see https://github.com/phpstan/phpstan-src/pull/2968#discussion_r1623337877

----

random idea: use a separate error-identifier for these errors?

src/Php/ComposerPhpVersionFactory.php
staabm: re-added this line, which fixed the build.

errors before the fix can be seen in e.g. https://github.com/phpstan/phpstan-src/actions/runs/9331440117/job/25686106757?pr=2968

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,125 @@
...
+					}
+				} catch (CouldNotReadFileException | JsonException) {
+					// pass
+				}
+			}
+		}
+		return $composerPhpVersion;
+	}
+
+	private function buildVersion(string $minVersion): ?PhpVersion
```
mvorisek, line 110:
```suggestion
	private function buildVersion(string $version): ?PhpVersion
```

src/Php/PhpVersion.php
```diff
@@ -17,11 +17,26 @@ public function getVersionId(): int
 		return $this->versionId;
 	}
 
+	public function getMajor(): int
+	{
+		return (int) floor($this->versionId / 10000);
```
mvorisek, line 44: `intdiv` would be better

src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
```diff
@@ -77,4 +85,24 @@ public function getTypeFromFunctionCall(
 		return TypeCombinator::union(...$types);
 	}
 
+	/**
+	 * @return ConstantStringType[]
+	 */
+	private function getVersionStrings(Expr $expr, Scope $scope): array
+	{
+		if (
+			$expr instanceof Expr\ConstFetch
```
mvorisek, line 94: [PHP_EXTRA_VERSION](https://www.php.net/manual/en/reserved.constants.php#constant.php-extra-version) should be probably support as well.

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,125 @@
...
+	private function getComposerRequireVersion(): ?string
+	{
+		$composerPhpVersion = null;
+		if (count($this->composerAutoloaderProjectPaths) > 0) {
+			$composerJsonPath = end($this->composerAutoloaderProjectPaths) . '/composer.json';
+			if (is_file($composerJsonPath)) {
+				try {
+					$composerJsonContents = FileReader::read($composerJsonPath);
+					$composer = Json::decode($composerJsonContents, Json::FORCE_ARRAY);
+					$requiredVersion = $composer['require']['php'] ?? null;
```
mvorisek, line 98: https://github.com/phpstan/phpstan/issues/9256 should be probably implemented in this PR as well and tested.

ondrejmirtes: That's unrelated and can be done later.

staabm:
rebased for 2.x in https://github.com/phpstan/phpstan-src/pull/3584

PR #2969, opened 2024-03-13
Extension / Rule change will not invalidate the cache

marcelthole:
Hey 👋 
we have the problem that the cache file is not invalidated if we change the code inside of a custom rule.

```neon
services:
    -
        class: CustomPHPStanRule\CustomRuleA
parameters:
	inferPrivatePropertyTypeFromConstructor: true
```

```bash
$ cat cache/phpstan/resultCache.php  | grep -A1 "projectExtensionFiles"
        'projectExtensionFiles' => array (
        ),
```

I expected the CustomRuleA in the projectExtensionFile. if this is there the normal cache workflow will also check the sha1 file sum and invalidates the cache.

With the current change the current service is always added as projectExtensionFile and every change on that file will invalidate the cache

### DIFF
--- a/e2e/result-cache-8/CustomRule.php
+++ b/e2e/result-cache-8/CustomRule.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace TestResultCache8;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
+class CustomRule implements Rule
+{
+	public function getNodeType(): string
+	{
+		return Node\Stmt\ClassMethod::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		return [];
+	}
+}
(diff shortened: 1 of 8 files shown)


ondrejmirtes:
Please try to reproduce the problem first. Here you can see how we typically test result cache-related problems https://github.com/phpstan/phpstan-src/blob/1.11.x/.github/workflows/e2e-tests.yml

e2e/result-cache-8/CustomRule.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace TestResultCache8;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
+class CustomRule implements Rule
```
ondrejmirtes, line 10: Your custom rule should be part of your analysed paths. If you fix that I don't think you'll have this problem anymore.

marcelthole: > Result cache not used because extension file /var/www/e2e/result-cache-8/src/CustomRule.php hash does not match.

👍 nice thanks a lot for the hint

PR #2970, opened 2024-03-14
`Type->getReferencedClasses()` returns a `class-string`

staabm:
declaring the more precise types makes working in phpstan extension easier, as I can use more precise typing in e.g. method parameters to which I pass the result of `getReferencedClasses()` like

```php

class myext {
  function doFoo() {
        foreach($classType->getReferencedClasses() as $referencedClass) {
            $this->myMethod($referencedClass);
        }
  }

  /** @param class-string $class */
  private function myMethod($class) {}
}

```

we could do similar stuff for `Type->getObjectClassNames()`

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1304,6 +1304,11 @@ parameters:
 			path: src/Type/ObjectType.php
 
 		-
+			message: "#^Method PHPStan\\\\Type\\\\ObjectType\\:\\:getReferencedClasses\\(\\) should return array\\<class\\-string\\> but returns array\\<int, string\\>\\.$#"
+			count: 1
+			path: src/Type/ObjectType.php
+
+		-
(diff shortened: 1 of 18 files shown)


phpstan-baseline.neon
```diff
@@ -1303,6 +1303,11 @@ parameters:
 			count: 3
 			path: src/Type/ObjectType.php
 
+		-
+			message: "#^Method PHPStan\\\\Type\\\\ObjectType\\:\\:getReferencedClasses\\(\\) should return array\\<class\\-string\\> but returns array\\<int, string\\>\\.$#"
```
staabm, line 1307: baselined this error, as I would otherwise have to add a `@param class-string` to `ObjectType->construct()`, which would be a BC break

ondrejmirtes:
You can't rely on this method to return an array of `class-string`, precisely because of the error you baselined. That is never going away. TypeNodeResolver will happily create ObjectType for non-existent classes, and there are rules that iterate over `getReferencedClasses` and check if they exist or not.

PR #2971, opened 2024-03-16
Fix deprecated warning version display

sayuprc:
I modified it to display the appropriate version when the conditions for displaying the deprecated warning are different in the code, and added tests.

ref #2963


ondrejmirtes:
Thank you, I thought I merged it 😊

PR #2972, opened 2024-03-17
Narrow down static and $this together

schlndh:
This is an attempt to fix https://phpstan.org/r/7df004e9-cbcd-41b8-a882-363914a91c1a
It also fixes the 1st part of https://github.com/phpstan/phpstan/issues/9465
Fixes https://github.com/phpstan/phpstan/issues/5987

I encountered this use-case here: https://github.com/phpstan/phpstan-src/pull/2940#issuecomment-1962383225 . IMO it's a bit of an edge-case, but it feels like phpstan should understand it.

----

I found one case where this change replaces one bug with another one:

> In non-static contexts, the called class will be the class of the object instance. Since $this-> will try to call private methods from the same scope, using static:: may give different results. Another difference is that static:: can only refer to static properties. 

https://www.php.net/manual/en/language.oop5.late-static-bindings.php
https://3v4l.org/SuYOd#v8.3.4
https://phpstan.org/r/763ec51a-49b0-4ba1-a037-b9939ef3cbd4

Previously phpstan thought that `static::fooPrivate()` inside the ifs would still be string (private), but it should be int (public). With my changes it presumably gets affected by the same bug that affects `$this` in the first if, so it now thinks that it's `*NEVER*`.

Since it is already bugged and probably non-trivial to fix, I decided to ignore it for now.

----

Some open questions:

- Should something be done with `resolveTypeByName`? It returns `TypeWithClassName`, so it cannot return an `IntersectionType`. Therefore, any usage with `static` could result in the kind of bugs that this PR tries to fix.
- Should there be a shortcut method for `$scope->getType(new Expr\ClassConstFetch($className, 'class'))->getClassStringObjectType();`?

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-instanceof-static-vs-this-prop-assign-scope.php
+++ b/tests/PHPStan/Analyser/data/bug-instanceof-static-vs-this-prop-assign-scope.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types=1);
+
+namespace BugInstanceofStaticVsThisPropAssignScope;
+
+class FooBase
+{
+	public function run(): void
+	{
+		if ($this instanceof FooChild) {
+			$this::$prop = 5;
+			\PHPStan\Testing\assertType('5', $this::$prop);
+
+			static::$prop = 6;
+			\PHPStan\Testing\assertType('6', static::$prop);
+		}
+
+		if (is_a(static::class, FooChild::class, true)) {
+			$this::$prop = 5;
+			\PHPStan\Testing\assertType('5', $this::$prop);
+
+			static::$prop = 6;
+			\PHPStan\Testing\assertType('6', static::$prop);
+		}
+	}
+}
+
+class FooChild extends FooBase
+{
+	public static $prop;
+}
--- a/tests/PHPStan/Analyser/data/bug-instanceof-static-vs-this-type-specifier.php
+++ b/tests/PHPStan/Analyser/data/bug-instanceof-static-vs-this-type-specifier.php
@@ -6,6 +6,8 @@ interface FooInterface
 {
 	/** @phpstan-assert-if-true null $v */
 	public static function isNull(?int $v): bool;
+
+	public static function foo(): int;
 }
 
 class FooBase
@@ -16,12 +18,20 @@ class FooBase
 			if ($this->isNull($v)) \PHPStan\Testing\assertType('null', $v);
 			if ($this::isNull($v)) \PHPStan\Testing\assertType('null', $v);
 			if (static::isNull($v)) \PHPStan\Testing\assertType('null', $v);
+
+			if ($this::foo() === 5) \PHPStan\Testing\assertType('5', $this::foo());
+			if ($this->foo() === 5) \PHPStan\Testing\assertType('5', $this->foo());
+			if (static::foo() === 5) \PHPStan\Testing\assertType('5', static::foo());
 		}
 
 		if (is_a(static::class, FooInterface::class, true)) {
 			if ($this->isNull($v)) \PHPStan\Testing\assertType('null', $v);
 			if ($this::isNull($v)) \PHPStan\Testing\assertType('null', $v);
 			if (static::isNull($v)) \PHPStan\Testing\assertType('null', $v);
+
+			if ($this::foo() === 5) \PHPStan\Testing\assertType('5', $this::foo());
+			if ($this->foo() === 5) \PHPStan\Testing\assertType('5', $this->foo());
+			if (static::foo() === 5) \PHPStan\Testing\assertType('5', static::foo());
 		}
 	}
 }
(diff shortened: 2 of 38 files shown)


tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
```diff
@@ -181,30 +181,30 @@ public function testImpossibleCheckTypeFunctionCall(): void
 					631,
 				],
 				[
-					'Call to function method_exists() with \'CheckTypeFunctionCall\\\\MethodExistsWithTrait\' and \'method\' will always evaluate to true.',
+					'Call to function method_exists() with class-string<static(CheckTypeFunctionCall\MethodExistsWithTrait)> and \'method\' will always evaluate to true.',
```
schlndh, line 184: I think the previous message was better, but since there is already `$this(...)` in the messages above, I decided to ignore it.

ondrejmirtes:
I'm sorry, I don't fully understand this. I don't get why `class-string<static(CheckTypeFunctionCall\MethodExistsWithTrait)>` should be better than `\'CheckTypeFunctionCall\\\\MethodExistsWithTrait\'`.

schlndh:
@ondrejmirtes It's not better. See https://github.com/phpstan/phpstan-src/pull/2972#discussion_r1527450851

schlndh:
I added a fix for https://github.com/phpstan/phpstan/issues/5987 (because previously it was a bit more broken compared to 1.10.x). But I can't say that I'm confident that I didn't break something else as a result.

My understanding is that the closure in `Closure::bind` should be analyzed as if it was a method in `$newScope`, and it's contents was wrapped in `if ($this instanceof $newThis::class)` (and similarly for other possible usages). So that's what I tried to implement. But it's not really possible for me to understand all the consequences. I encountered several surprises (e.g. the phpdoc resolving), which I (hopefully) fixed. But chances are that things which are not covered by tests and issue bot, may have gotten broken.

Additionally, there are likely many bugs/inconsistencies remaining. For example https://phpstan.org/r/8a986f5e-dc04-4fdb-8ac7-aae2c820e0dc gives the following result in my branch (i.e. inconsistent `static(...)` and `$this(...)` types):

```
 ------ ----------------------------------------------------------------------------------------- 
  22     Expected type $this(NodeScopeResolverClosureBind\Bar), actual:                           
         NodeScopeResolverClosureBind\Bar                                                         
  23     Expected type class-string<static(NodeScopeResolverClosureBind\Bar)>, actual:            
         class-string<NodeScopeResolverClosureBind\Bar&static(NodeScopeResolverClosureBind\Foo)>  
  32     Expected type $this(NodeScopeResolverClosureBind\Bar), actual:                           
         NodeScopeResolverClosureBind\Bar                                                         
  33     Expected type class-string<static(NodeScopeResolverClosureBind\Bar)>, actual:            
         class-string<NodeScopeResolverClosureBind\Bar&static(NodeScopeResolverClosureBind\Foo)>  
 ------ ----------------------------------------------------------------------------------------- 
```

But I'm hoping that it's nevertheless an improvement over the previous implementation.

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
```diff
@@ -1646,4 +1646,38 @@ public function testArgon2PasswordHash(): void
...
 
+	public function testBugInstanceofStaticVsThis(): void
+	{
+		if (PHP_VERSION_ID < 80100) {
+			$this->markTestSkipped('Test requires PHP 8.1.');
+		}
+
+		$this->analyse([__DIR__ . '/../Properties/data/bug-instanceof-static-vs-this-property-assign.php'], [
+			[
+				'Parameter #1 $var is passed by reference so it does not accept readonly property BugInstanceofStaticVsThisPropertyAssign\FooChild::$nativeReadonlyProp.',
```
schlndh, line 1657: Unrelated to this PR: It should probably complain about `@readonly` properties as well.

tests/PHPStan/Analyser/data/bug-instanceof-static-vs-this-prop-assign-scope.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace BugInstanceofStaticVsThisPropAssignScope;
+
+class FooBase
+{
+	public function run(): void
+	{
+		if ($this instanceof FooChild) {
+			$this::$prop = 5;
+			\PHPStan\Testing\assertType('5', $this::$prop);
```
schlndh, line 11: Ideally, this should work even if the write is via `$this::` and the read is via `static::`, and vice-versa, but for now it doesn't.

tests/PHPStan/Analyser/data/bug-instanceof-static-vs-this-type-specifier.php
```diff
@@ -0,0 +1,37 @@
...
+class FooBase
+{
+	public function bar(?int $v): void
+	{
+		if ($this instanceof FooInterface) {
+			if ($this->isNull($v)) \PHPStan\Testing\assertType('null', $v);
+			if ($this::isNull($v)) \PHPStan\Testing\assertType('null', $v);
+			if (static::isNull($v)) \PHPStan\Testing\assertType('null', $v);
+
+			if ($this::foo() === 5) \PHPStan\Testing\assertType('5', $this::foo());
```
schlndh, line 22: Ideally, this should work for any combination of `$this::`, `static::` and `$this::` (private methods might be an issue) for read/write. But for now it doesn't.

PR #2973, opened 2024-03-17
Avoid duplicate in Union type description

VincentLanglet:
This is related to https://github.com/phpstan/phpstan-doctrine/pull/552#issuecomment-1991750862

When a custom Type has
- templates (like index and values)
- other internal metadata (like DQL here)

The union Type of this custom Type encounter a weird situation
- Type CustomType('A') and CustomType('B') cannot be merged or the metadata will be lost
- description of the union will be CustomType|CustomType

A simple solution is to call array_unique on the described types.


ondrejmirtes:
The root issue is in type normalization, not describing UnionType.

If two types with the same description come up in a UnionType, it's because they haven't been properly normalized (due to isSuperTypeOf being implemented in a wrong way). See https://phpstan.org/developing-extensions/type-system#type-normalization

VincentLanglet:
> The root issue is in type normalization, not describing UnionType.
> 
> If two types with the same description come up in a UnionType, it's because they haven't been properly normalized (due to isSuperTypeOf being implemented in a wrong way). See [phpstan.org/developing-extensions/type-system#type-normalization](https://phpstan.org/developing-extensions/type-system#type-normalization)

But in case of custom Type with metadata like DQL here, we have the situation
```
$queryTypeA = new QueryType($dqlA);
$queryTypeB = new QueryType($dqlB);

$queryTypeA->isSuperType($queryTypeB); // FALSE
$queryTypeB->isSuperType($queryTypeA); // FALSE
```

To me it's because the QueryType is templated as `QueryType<index, value>` when it should/could be `QueryType<index, value, DQL>`.
The union type would now be
```
'Doctrine\ORM\Query<mixed, mixed, 'foo'>'|'Doctrine\ORM\Query<mixed, mixed, 'bar'>'
```
instead of
```
'Doctrine\ORM\Query<mixed, mixed>'|'Doctrine\ORM\Query<mixed, mixed>'
```

But I assume this template was not added because it's too complicated for phpdoc, people would end with
```
'Doctrine\ORM\Query<mixed, mixed, 'SELECT * from ...'>'
```

ondrejmirtes:
Alright, I understand it now. But I want a slightly different behaviour - for the most-detailed VerbosityLevel (precise I think) I don't want to unique the types, I want to keep them separate and mark them with a number, so the result is going to be:

`int#1|int#2`

Thanks

VincentLanglet:
> Alright, I understand it now. But I want a slightly different behaviour - for the most-detailed VerbosityLevel (precise I think) I don't want to unique the types, I want to keep them separate and mark them with a number, so the result is going to be:
> 
> `int#1|int#2`
> 
> Thanks

Done in https://github.com/phpstan/phpstan-src/pull/2973/commits/851f7c32a7d09cd482dae839e6e21eea6d3b1280

tests/PHPStan/Type/UnionTypeTest.php
```diff
@@ -923,10 +944,12 @@ public function dataDescribe(): array
 	 */
 	public function testDescribe(
 		Type $type,
+		string $expectedDescribeDescription,
```
staabm, line 947:
```suggestion
		string $expectedPreciseDescription,
```

tests/PHPStan/Type/UnionTypeTest.php
```diff
@@ -923,10 +944,12 @@ public function dataDescribe(): array
...
 		string $expectedValueDescription,
 		string $expectedTypeOnlyDescription,
 	): void
 	{
+		$this->assertSame($expectedDescribeDescription, $type->describe(VerbosityLevel::precise()));
```
staabm:
```suggestion
		string $expectedPreciseDescription,
		string $expectedValueDescription,
		string $expectedTypeOnlyDescription,
	): void
	{
		$this->assertSame($expectedPreciseDescription, $type->describe(VerbosityLevel::precise()));
```

staabm:
(lint job errors)

VincentLanglet:
> (lint job errors)

Should be ok now

tests/PHPStan/Type/UnionTypeTest.php
```diff
@@ -912,9 +926,16 @@ public function dataDescribe(): array
 					),
 					new NullType(),
 				),
+				'TFoo of TBar (class foo, parameter) (class foo, parameter)|null',
 				'(TFoo of TBar)|null',
 				'(TFoo of TBar)|null',
 			],
+			[
+				new UnionType([new QueryType('foo'), new QueryType('bar')]),
+				'Doctrine\ORM\Query<mixed, mixed>#1|Doctrine\ORM\Query<mixed, mixed>#2',
```
ondrejmirtes: You don't need to add QueryType class here, you can just add the same type twice into UnionType constructor.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #2975, opened 2024-03-17
Fix 10057: DomDocument load return type

VincentLanglet:
I dunno if it closes https://github.com/phpstan/phpstan/issues/10057

because of the php-8 issue, cf https://github.com/phpstan/phpstan/issues/10057#issuecomment-2002551335


staabm:
> I dunno if it closes [phpstan/phpstan#10057](https://github.com/phpstan/phpstan/issues/10057)

if you are not sure, you could add a NodeScopeResolverTest to verify your assumption

VincentLanglet:
> if you are not sure, you could add a NodeScopeResolverTest to verify your assumption

I'm pretty sure it won't close the PHP 8 part (and only fix the issue on php 7).

But I dunno how to fix it for PHP 8 since it's in phpstan/php-8-stubs and https://github.com/phpstan/php-8-stubs/blob/main/extractor/extract.php is not easy to understand at first read.

ondrejmirtes:
Because the type in phpstan/php-8-stubs is only in a PHPDoc then changing functionMap.php here actually works, as can be seen in the Reflection golden test: https://github.com/phpstan/phpstan-src/actions/runs/8317200612/job/22757684538

ondrejmirtes:
Thank you.

PR #2976, opened 2024-03-18
`ReflectionProvider->hasClass()` narrows `$className` arg

staabm:
similar motivation as in https://github.com/phpstan/phpstan-src/pull/2970


ondrejmirtes:
Thank you.

PR #2979, opened 2024-03-19
Introduce ForbiddenClassNameExtension for append additional forbidden…

kamil-zacek:
… class prefixes

### DIFF
--- a/src/Classes/ForbiddenClassNameExtension.php
+++ b/src/Classes/ForbiddenClassNameExtension.php
@@ -1,0 +1,32 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Classes;
+
+/**
+ * This is the extension interface to implement if you want to dynamically
+ * add forbidden class prefixes to the ClassForbiddenNameCheck rule.
+ *
+ * The idea is that you want to report usages of classes that you're not supposed to use in application.
+ * For example: Generated Doctrine proxies from their configured namespace.
+ *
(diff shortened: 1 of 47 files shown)


staabm:
Could you give us an idea what your use-case is?

Is it php-scoper related or something completely different?

ondrejmirtes:
@staabm This was discussed outside of GitHub. The idea is that you want to report usages of classes that you're not supposed to use in application. For example: Generated Doctrine proxies from their configured namespace.

src/Classes/ForbiddenClassNameExtension.php
```diff
@@ -0,0 +1,13 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Classes;
+
+interface ForbiddenClassNameExtension
```
ondrejmirtes, line 24: Please put `@api` here and explanation what it is for and how to register it. Example: https://apiref.phpstan.org/1.11.x/source-src.PhpDoc.StubFilesExtension.html#5

ondrejmirtes (changes requested):
The only other problem is that this isn't tested. Please create a new RuleTestCase for one of the rules affected by this, add an example new neon file with a new extension implementing this interface, and reference it in `public static function getAdditionalConfigFiles()` in that test case. That way this extension will be added to the container.

ondrejmirtes (approved):

ondrejmirtes:
Perfect, thank you!

PR #2981, opened 2024-03-20
Test invalidated object in static closure

staabm:
looks like its a case we do not test yet.

followup to https://github.com/phpstan/phpstan-src/commit/9bb4ef961960a1056060209b3ed7e09b93c06f36


ondrejmirtes:
Thank you.

PR #2985, opened 2024-03-22
Fix slow enum-union with lots of cases

staabm:
closes https://github.com/phpstan/phpstan/issues/10772

this effectively reverts https://github.com/phpstan/phpstan-src/commit/0a39b42f84c3b8436b79b2df678b0483735b5366 but fixes a nasty perf problem

the reproducer of #10772 took >2 minutes before this change and finishes instantly after it

```
$ time ../phpstan-src/bin/phpstan analyse --debug
Note: Using configuration file C:\dvl\Workspace\stan-long\phpstan.neon.
C:\dvl\Workspace\stan-long\src\test.php


 [OK] No errors



real    0m2.669s
user    0m0.000s
sys     0m0.000s
```

Todo: regression test


staabm:
> Todo: regression test

not sure where to put the regression test, since we need a composer.json/lock. in the past we used `AnalyzerIntegration` test, but this doesn't have composer.json project level.

should I do a end-2-end test? if so, how/where to put a "timeout" for it?

-----

alternatively I can try to reduce the existing test to something without external composer dependencies

staabm:
(added regression test)

src/Type/TypeCombinator.php
```diff
@@ -200,8 +201,7 @@ public static function union(Type ...$types): Type
 			if ($types[$i] instanceof StringType && !$types[$i] instanceof ClassStringType) {
 				$hasGenericScalarTypes[ConstantStringType::class] = true;
 			}
-			$enumCases = $types[$i]->getEnumCases();
```
VincentLanglet, line 203: If the issue is with `getEnumCases` be slow.

Shouldn't be introduced a new method `isSomething`, implemented by all types in order to avoid the error message
```
Doing instanceof PHPStan\\\\Type\\\\Enum\\\\EnumCaseObjectType is error\\-prone and deprecated\\
``` 
?

Then
```
if ($types[$i]->isSomething()->yes()) {
```

ondrejmirtes:
Thank you!

src/Type/TypeCombinator.php
staabm: Not sure its worth a new type-method just for this bottleneck

PR #2986, opened 2024-03-22
Modify functionMap and add __benevolent<>

zonuexe:
Add `__benevolent<>` to functions whose documentation says "return false", which return false when passed a parameter of the wrong type, or which do not return false upon normal use.

Modify the functionMap for other functions, including those already covered by DynamicFunctionReturnTypeExtension.


resources/functionMap.php
```diff
@@ -2991,7 +2991,7 @@
 'finfo_set_flags' => ['bool', 'finfo'=>'resource', 'options'=>'int'],
 'floatval' => ['float', 'var'=>'scalar|array|resource|null'],
 'flock' => ['bool', 'fp'=>'resource', 'operation'=>'int', '&w_wouldblock='=>'int'],
-'floor' => ['float|false', 'number'=>'float'],
+'floor' => ['__benevolent<float|false>', 'number'=>'float'],
```
zonuexe, line 2994: Same as `RoundFunctionReturnTypeExtension`.

https://github.com/phpstan/phpstan-src/blob/f51d90c3d2fb577e690ddeed951a4a9d98a64628/src/Type/Php/RoundFunctionReturnTypeExtension.php#L48-L56

resources/functionMap.php
```diff
@@ -3311,7 +3311,7 @@
 'get_extension_funcs' => ['list<callable-string>|false', 'extension_name'=>'string'],
 'get_headers' => ['array|false', 'url'=>'string', 'format='=>'int', 'context='=>'resource'],
 'get_html_translation_table' => ['array', 'table='=>'int', 'flags='=>'int', 'encoding='=>'string'],
-'get_include_path' => ['string|false'],
+'get_include_path' => ['__benevolent<string|false>'],
```
zonuexe, line 3314: Since the situation where the ini `include_path` entry does not exist is an abnormal situation, it seems that `__benevolent<>` can be added.

https://github.com/php/php-src/blob/df017cd0ef9e89d97b64dcf01efc5395fab30f6c/ext/standard/basic_functions.c#L2084-L2099

resources/functionMap.php
```diff
@@ -6494,8 +6494,8 @@
 'MemcachePool::set' => ['bool', 'key'=>'string', 'var'=>'mixed', 'flag='=>'int', 'expire='=>'int'],
 'MemcachePool::setCompressThreshold' => ['bool', 'threshold'=>'int', 'min_savings='=>'float'],
 'MemcachePool::setServerParams' => ['bool', 'host'=>'string', 'port='=>'int', 'timeout='=>'int', 'retry_interval='=>'int', 'status='=>'bool', 'failure_callback='=>'callable'],
-'memory_get_peak_usage' => ['int', 'real_usage='=>'bool'],
-'memory_get_usage' => ['int', 'real_usage='=>'bool'],
+'memory_get_peak_usage' => ['positive-int', 'real_usage='=>'bool'],
+'memory_get_usage' => ['positive-int', 'real_usage='=>'bool'],
```
zonuexe, line 6498: These functions return only positive values.

https://3v4l.org/ZTQKN

resources/functionMap.php
```diff
@@ -6525,7 +6525,7 @@
 'ming_useconstants' => ['void', 'use'=>'int'],
 'ming_useswfversion' => ['void', 'version'=>'int'],
 'mkdir' => ['bool', 'pathname'=>'string', 'mode='=>'int', 'recursive='=>'bool', 'context='=>'resource'],
-'mktime' => ['int|false', 'hour='=>'int', 'min='=>'int', 'sec='=>'int', 'mon='=>'int', 'day='=>'int', 'year='=>'int'],
+'mktime' => ['__benevolent<int|false>', 'hour='=>'int', 'min='=>'int', 'sec='=>'int', 'mon='=>'int', 'day='=>'int', 'year='=>'int'],
```
zonuexe, line 6528: This function returns false only when it receives a value outside the integer range.

https://github.com/php/php-src/blob/df017cd0ef9e89d97b64dcf01efc5395fab30f6c/ext/date/php_date.c#L1209-L1213

PHP Manual says:

> `mktime()` returns the Unix timestamp of the arguments given, or false if the timestamp doesn't fit in a PHP integer.

It seems that even if `PHP_INT_MAX` is passed for all parameters, this value will never be exceeded.

https://3v4l.org/dnSu2

src/Type/Php/RoundFunctionReturnTypeExtension.php
```diff
@@ -39,21 +38,18 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
-			// PHP 8 can either return a float or fatal.
-			$defaultReturnType = new FloatType();
 		} else {
 			// PHP 7 returns null with a missing parameter.
 			$noArgsReturnType = new NullType();
-			// PHP 7 can return either a float or false.
-			$defaultReturnType = new BenevolentUnionType([
-				new FloatType(),
-				new ConstantBooleanType(false),
-			]);
```
zonuexe, line 56: This extension change is unnecessary, but it removes the duplicate definition of the type with functionMap.

ondrejmirtes:
Thank you, changes make sense 😊

src/Type/Php/RoundFunctionReturnTypeExtension.php
staabm: just stumbled over https://github.com/phpstan/phpstan-src/blob/cbdc933b68f42dacf44143441595b9df36ce7f2b/src/Type/Php/StrtotimeFunctionReturnTypeExtension.php#L39-L41 which might benefit from a similar simplification.

PR #2994, opened 2024-03-27
Prevent unnecessary calls into reflection from JSON extensions

staabm:
In todays debugging session I realized that the json throw extensions call into reflection even for non `json_*` calls.
accidentally spotted while working on https://github.com/Roave/BetterReflection/issues/1415

move the cheaper check in front to prevent this unnecessary reflection stuff.

see

grafik

I looked arround for similar problems in other extensions but could not see additional ones.


ondrejmirtes:
Thank you.

PR #2996, opened 2024-03-29
More precise `gc_status()` signature for PHP8.3+

staabm:
Noticed while working thru PHPUnit

> gc_status() now returns the following additional fields: "running", "protected", "full", "buffer_size", "application_time", "collector_time", "destructor_time", and "free_time". 

see 
- https://www.php.net/manual/en/function.gc-status.php
- https://3v4l.org/Ok8K6#v8.3.4


ondrejmirtes:
Thank you.

PR #3000, opened 2024-04-03
Support `@phan-` prefixes on recognized phpdoc tags

anomiex:
Per request at https://github.com/phpstan/phpdoc-parser/pull/235#issuecomment-2035354495

Hopefully I added the right tests in the right places, mostly I grepped for "psalm" then tried to add similar tests. Feel free to update the PR if necessary.

### DIFF
--- a/src/PhpDoc/PhpDocNodeResolver.php
+++ b/src/PhpDoc/PhpDocNodeResolver.php
@@ -68,7 +68,7 @@ class PhpDocNodeResolver
 	{
 		$resolved = [];
 		$resolvedByTag = [];
-		foreach (['@var', '@psalm-var', '@phpstan-var', '@phan-var'] as $tagName) {
+		foreach (['@var', '@phan-var', '@psalm-var', '@phpstan-var'] as $tagName) {
 			$tagResolved = [];
 			foreach ($phpDocNode->getVarTagValues($tagName) as $tagValue) {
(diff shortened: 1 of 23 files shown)


ondrejmirtes:
The implementation is wrong. You should test what happens which prefix has priority if more of them are present. @phpstan- should always win.

anomiex:
I just copied what seemed to be being done for `@psalm-`, for which the only place I saw priority logic is in `PhpDocNodeResolver::resolveTemplateTags()`, for which I put `@phan-` at a lower priority than both `@psalm-` and `@phpstan-`. The test case [here](https://github.com/phpstan/phpstan-src/pull/3000/files#diff-e889e99ecbffcbf5ef459b1394d207df94d660bb8f7f32a93ccb2cbc402c2f41R1213-R1223) seems to confirm that.

As I mentioned, feel free to update the PR. Or, if you want to point out specific test cases where `@phan-` is overriding `@phpstan-` (and where existing `@psalm-` doesn't have the same problem), I'll try to take a look when I have time.

ondrejmirtes (changes requested):
InvalidPhpDocTagValueRule needs to ignore `@phan-` tags, same as it ignores `@psalm-`.

ondrejmirtes:
Thanks a lot!

PR #3001, opened 2024-04-05
PHPDoc: address multiple variables defined in one @param

Muqsit:
This commit addresses instances where multiple $vars were defined in `@param` PHPDoc.


ondrejmirtes:
Thank you.

PR #3002, opened 2024-04-05
Forbidd PHPUnitPhar prefixed classes

staabm:
PHPUnit uses this unique prefix since the latest release, so we can easily detect prefixed classes.

see https://github.com/sebastianbergmann/phpunit/commit/4d829e0740db0cbcbcd39ec5656161b4326004a4#diff-840c8b2ea07ab3c49b8c509c41e0de93231d36254f4f4ac621bcf2d21063ce5b

regular PHPUnit classes will have a `PHPUnit` namespace (these are not prefixed)


ondrejmirtes:
Thank you!

PR #2905, opened 2024-02-04
chore: Update Box

theofidry:
**Not ready for review**

PR #3008, opened 2024-04-09
Test PHAR - same

ondrejmirtes:


thg2k:
Are you testing phar build reproducibility? I had a similar need in the past and i wrote my own routine to patch the manifests' timestamps 😆

ondrejmirtes:
The phar.yml workflow already checks this at least for a few years. When the checksum is the same we can afford not to run some workflows from phpstan/phpstan.

You can see the amount of checks saved here: https://github.com/phpstan/phpstan-src/actions/runs/8612810842 vs. https://github.com/phpstan/phpstan-src/actions/runs/8612805237

This time I just wanted to verify this commit https://github.com/phpstan/phpstan/commit/a370d5461dbdda49f61201486a37ff3dbad1c47d was correct.

ondrejmirtes:
We actually have a workflow here for this exact purpose (PHAR reproducibility) https://github.com/phpstan/phpstan-src/blob/1.11.x/.github/workflows/checksum-phar.yml

thg2k:
Nice, I suppose box.phar takes care of patching the phar manifest, as PHP's phar extension does not provide a reproducible way to build phars, as it embeds the current timestamp of the files (at least last time i checked in 7.x).

Digging into your CI is in my todo list 😄, it's really cool case study. Also i'm more a gitlab person so i'm not very familiar with github's details, so I can learn from this.

ondrejmirtes:
Yeah, some option for that was added to Box but I'm on a quite old version (the author tries to make the latest version work and remove some hacks from phpstan-src here https://github.com/phpstan/phpstan-src/pull/2905) and I use this script https://github.com/phpstan/phpstan-src/blob/1.11.x/compiler/build/resign.php to fix the timestamps - Composer's PHAR file is built the same way.

thg2k:
Ahah `Seld\PharUtils\Timestamps` is exactly the same library I started with, but I wasn't happy with the detection of the stub file parsing, now I no longer remember the details but I was using a custom stub file and it had some issues so I ended up rewriting it from scratch. If you are interested, I can dig it up and donate it to the project so you drop that dependency.

PR #3011, opened 2024-04-11
composer-dependency-analyser: update to 1.5.0 (support functions)

janedbal:
- [composer-dependency-analyser 1.5.0](https://github.com/shipmonk-rnd/composer-dependency-analyser/?tab=readme-ov-file#composer-dependency-analyser)


ondrejmirtes:
Thank you!

PR #3013, opened 2024-04-12
Fix fatal error on `constant('')`

staabm:
closes https://github.com/phpstan/phpstan/issues/10867

there are only 2 other callers for this very same method in the codebase, and there we already early exit on `''`:

e.g.

https://github.com/phpstan/phpstan-src/blob/5c3a711ae7435aa7ba972ec9cacafb7fdb131bcc/src/Type/Php/DefinedConstantTypeSpecifyingExtension.php#L50-L60


ondrejmirtes (changes requested):
I'd like to have a stub for Name constructor and all Name subclasses constructors that enforces non-empty-string.

staabm:
using non-empty-string names everywhere would spread across the whole codebase. I fixed the places which worked without ugly catches and baselined the remaining ones

staabm:
tested locally with regarding `__construct()` in all known Name classes

```
➜  phpstan-src git:(bug10867) ✗ cat test4.php 
<?php

$n = new \PhpParser\Node\Name('');

$r = new \PhpParser\Node\Name\Relative('');

$fq = new \PhpParser\Node\Name\FullyQualified('');
➜  phpstan-src git:(bug10867) ✗ php bin/phpstan analyze test4.php --debug
/Users/staabm/workspace/phpstan-src/test4.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------ 
  Line   test4.php                                                                                                                                                   
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------ 
  3      Parameter #1 $name of class PhpParser\Node\Name constructor expects non-empty-array<string>|PhpParser\Node\Name|non-empty-string, '' given.                 
         💡 Type #1 from the union: '' is empty.                                                                                                                     
  5      Parameter #1 $name of class PhpParser\Node\Name\Relative constructor expects non-empty-array<string>|PhpParser\Node\Name|non-empty-string, '' given.        
         💡 Type #1 from the union: '' is empty.                                                                                                                     
  7      Parameter #1 $name of class PhpParser\Node\Name\FullyQualified constructor expects non-empty-array<string>|PhpParser\Node\Name|non-empty-string, '' given.  
         💡 Type #1 from the union: '' is empty.                                                                                                                     
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------ 


                                                                                                                        
 [ERROR] Found 3 errors                                                                                                 
                                                                                                                        
```

also created a upstream PR for PHPParser: https://github.com/nikic/PHP-Parser/pull/993

ondrejmirtes:
> using non-empty-string names everywhere would spread across the whole codebase.

Yes, that's the idea. I'm not looking for a quick fix but a correct fix. Please do not baseline anything. Baseline is not supposed to grow.

staabm:
looks like this will be a breaking change and can only be merged for 2.0?
should the crash fix be separated?

conf/config.neon
```diff
@@ -175,6 +175,7 @@ parameters:
 		- ../stubs/ext-ds.stub
 		- ../stubs/ImagickPixel.stub
 		- ../stubs/PDOStatement.stub
+		- ../stubs/PhpParserName.stub
```
ondrejmirtes: Please register this stub only in `build/phpstan.neon`. It's only for internal PHPStan purposes.

ondrejmirtes: And move the stub elsewhere, probably to `build/stubs`.

src/Type/Php/ArrayFilterFunctionReturnTypeReturnTypeExtension.php
```diff
@@ -245,7 +245,16 @@ private function processKeyAndItemType(MutatingScope $scope, Type $keyType, Type
 	private static function createFunctionName(string $funcName): Name
 	{
 		if ($funcName[0] === '\\') {
-			return new Name\FullyQualified(substr($funcName, 1));
+			$fqcn = ltrim($funcName, '\\');
+			if ($fqcn === '') {
```
ondrejmirtes: You can definitely write some userland code that will execute this condition

src/Type/Php/ArrayFilterFunctionReturnTypeReturnTypeExtension.php
```diff
@@ -245,7 +245,16 @@ private function processKeyAndItemType(MutatingScope $scope, Type $keyType, Type
...
+				throw new ShouldNotHappenException();
+			}
+
+			return new Name\FullyQualified($fqcn);
+		}
+
+		if ($funcName === '') {
+			throw new ShouldNotHappenException();
```
ondrejmirtes: You can definitely write some userland code that will execute this condition

src/Type/Php/ConstantHelper.php
```diff
@@ -8,18 +8,27 @@
...
+	 * @param non-empty-string $constantName
+	 */
 	public function createExprFromConstantName(string $constantName): Expr
 	{
 		$classConstParts = explode('::', $constantName);
 		if (count($classConstParts) >= 2) {
-			$classConstName = new FullyQualified(ltrim($classConstParts[0], '\\'));
+			$fqcn = ltrim($classConstParts[0], '\\');
+			if ($fqcn === '') {
+				throw new ShouldNotHappenException();
```
ondrejmirtes: What if you can `constant('::aa')`? That will execute this condition

src/Rules/Generics/TemplateTypeCheck.php
```diff
@@ -66,6 +67,10 @@ public function check(
 	{
 		$messages = [];
 		foreach ($templateTags as $templateTag) {
+			if ($templateTag->getName() === '') {
```
ondrejmirtes: This should be tackled in TemplateTag by typehinting `non-empty-string`.

src/Reflection/InitializerExprContext.php
```diff
@@ -127,6 +138,10 @@ public static function fromStubParameter(
 
 	public static function fromGlobalConstant(ReflectionConstant $constant): self
 	{
+		if ($constant->getNamespaceName() === '') {
+			throw new ShouldNotHappenException('Namespace cannot be empty.');
```
ondrejmirtes: This could be tackled by a better typehint in BetterReflection.

ondrejmirtes (changes requested):
Please be more careful - in some places where we gain type safety thanks to the new PHPDocs, you instantly throw it away by throwing `ShouldNotHappenException`.

ondrejmirtes:
It'd be great if you could finish this one. Thanks.

staabm:
I am on it right in this moment :)

sorting out dependencies atm

src/Reflection/InitializerExprContext.php
staabm: fix in https://github.com/Roave/BetterReflection/pull/1428

staabm:
I think this one is ready to land. the last remaining error should be fixed with https://github.com/Roave/BetterReflection/pull/1428

ondrejmirtes:
Thank you!

build/stubs/PhpParserName.stub
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser\Node;
+
+use PhpParser\NodeAbstract;
+
+class Name extends NodeAbstract
```
staabm, line 7: we should be able to remove this stub after the next php-parser release because of https://github.com/nikic/PHP-Parser/pull/993

PR #3015, opened 2024-04-16
TypeInferenceTestCase: ability to disable implicit throws

janedbal:
Copied from https://github.com/phpstan/phpstan-src/pull/2576/files


ondrejmirtes:
Thank you!

PR #3016, opened 2024-04-16
Immediately called function without implicit throw broken for first class callable

janedbal:
Looks like it got broken in [this commit](https://github.com/phpstan/phpstan-src/commit/8a8b712ff038df9ebea57f16a9e81f777870fe49). Checked by:

```bash
wget https://raw.githubusercontent.com/janedbal/phpstan-src/fd32899a00a58fea6e8db47356f97619c5906b11/tests/PHPStan/Analyser/data/immediately-called-function-without-implicit-throw.php

git checkout 925c0b7343eef8e86d4c65c5d2cd3fe5a65aca9b
bin/phpstan analyse -l 9 -c tests/PHPStan/Rules/Exceptions/data/ability-to-disable-implicit-throws.neon immediately-called-function-without-implicit-throw.php
# no errors

git checkout 8a8b712ff038df9ebea57f16a9e81f777870fe49
bin/phpstan analyse -l 9 -c tests/PHPStan/Rules/Exceptions/data/ability-to-disable-implicit-throws.neon immediately-called-function-without-implicit-throw.php
# error at line 11: Expected variable certainty Yes, actual: Maybe
```

Commits around that: https://github.com/phpstan/phpstan-src/commits/1.11.x/?after=f280c25a28e8fdf9853165324e8672c06065f8a8+139

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2232,7 +2232,11 @@ class NodeScopeResolver
 					$throwPoints = array_merge($throwPoints, $invokeResult->getThrowPoints());
 					$impurePoints = array_merge($impurePoints, $invokeResult->getImpurePoints());
 				} elseif ($parametersAcceptor instanceof CallableParametersAcceptor) {
-					$throwPoints = array_merge($throwPoints, array_map(static fn (SimpleThrowPoint $throwPoint) => $throwPoint->isExplicit() ? ThrowPoint::createExplicit($scope, $throwPoint->getType(), $expr, $throwPoint->canContainAnyThrowable()) : ThrowPoint::createImplicit($scope, $expr), $parametersAcceptor->getThrowPoints()));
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Fixed, thanks

janedbal:
Looks like this fix didnt make it to stable 1.11, why?

ondrejmirtes:
The log here in this PR says:

> [ondrejmirtes](https://github.com/ondrejmirtes) merged commit [e956fad](https://github.com/phpstan/phpstan-src/commit/e956fad2e405d3e8952536a340571a7ec30bd564)

And when I click on that commit it shows it's part of 1.11.0. What's the problem?

janedbal:
It still behaves the old way here: https://github.com/shipmonk-rnd/phpstan-rules/pull/237/files#diff-0a900f4844a6aa21b30f50de6ce5b92fb31bbf3fbf928757e409d06584d89d53R243

janedbal:
My bad, the behaviour stays only when implicitThrows is disabled. It is changed in 1.11 for default setup (implicitThrows being true). Ref: https://github.com/shipmonk-rnd/phpstan-rules/pull/239

PR #3017, opened 2024-04-17
Fix `Imagick::identifyImage()` return type

SVillette:
Fix https://github.com/phpstan/phpstan/issues/10888


ondrejmirtes:
Thank you!

PR #3018, opened 2024-04-17
Make AccessoryNonFalsyStringType::toInteger() return IntegerType instead of non-zero

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/10893


ondrejmirtes:
Thank you!

PR #3020, opened 2024-04-17
Detect unused function-call on a separate line with possibly pure function

staabm:

### DIFF
--- a/src/Rules/DeadCode/FunctionWithoutImpurePointsCollector.php
+++ b/src/Rules/DeadCode/FunctionWithoutImpurePointsCollector.php
@@ -26,6 +26,9 @@ class FunctionWithoutImpurePointsCollector implements Collector
 		if (!$function->isPure()->maybe()) {
 			return null;
 		}
+		if (!$function->hasSideEffects()->maybe()) {
+			return null;
+		}
 
 		if (count($node->getImpurePoints()) !== 0) {
 			return null;
--- a/src/Rules/DeadCode/FunctionWithoutImpurePointsCollector.php
+++ b/src/Rules/DeadCode/FunctionWithoutImpurePointsCollector.php
@@ -1,0 +1,54 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\DeadCode;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Collectors\Collector;
+use PHPStan\Node\FunctionReturnStatementsNode;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use function count;
+
+/**
+ * @implements Collector<FunctionReturnStatementsNode, string>
+ */
+class FunctionWithoutImpurePointsCollector implements Collector
+{
+
+	public function getNodeType(): string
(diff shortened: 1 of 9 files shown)


src/Rules/DeadCode/FunctionWithoutImpurePointsCollector.php
```diff
@@ -0,0 +1,54 @@
...
+
+	public function getNodeType(): string
+	{
+		return FunctionReturnStatementsNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope)
+	{
+		$function = $node->getFunctionReflection();
+		if (!$function->isPure()->maybe()) {
```
ondrejmirtes, line 26: We should also eliminate `void` function here.

src/Rules/DeadCode/PossiblyPureFuncCallCollector.php
```diff
@@ -0,0 +1,44 @@
...
+	{
+		if (!$node->name instanceof Node\Name) {
+			return null;
+		}
+
+		if (!$this->reflectionProvider->hasFunction($node->name, $scope)) {
+			return null;
+		}
+
+		$functionReflection = $this->reflectionProvider->getFunction($node->name, $scope);
```
ondrejmirtes, line 36: We should also have this condition:

```php
if (!$function->hasSideEffects()->maybe() {
    return null;
}
```

Because:

1) If it's `no()` it's already covered by CallToFunctionStatementWithoutSideEffectsRule
2) if it's `yes()` then it's a valid call and should not be reported in this rule.

The distinction between `hasSideEffects()` and `isPure` essentially boils down to function being `void` and function being marked as `@phpstan-pure` and `@phpstan-impure`.

Here we're dealing with the uncertainty so we need both of these to be `maybe()`.

ondrejmirtes (changes requested):

src/Rules/DeadCode/FunctionWithoutImpurePointsCollector.php
ondrejmirtes: Thinking about it, we should simply ask about `hasSideEffects()`.

ondrejmirtes:
Awesome, thank you!

staabm:
Cool. Will do method call variants tomorrow

ondrejmirtes:
Needed to fix this: https://github.com/phpstan/phpstan-src/commit/57b0ae06e2fa4e2a43e9deb9f27afb8b24166bc9

If we looked at issue-bot it would tell us, but no worries :)

ondrejmirtes:
This is a false positive: https://phpstan.org/r/eac46a37-248f-41a1-88c8-510241e06b6b

Because it works: https://3v4l.org/5XkR6

Please fix it before moving forward, thanks :)

ondrejmirtes:
Oh god, we don't have to care about that. It's some kind of weird array loophole. This isn't supported on PHP since 5.4.

staabm:
To make sure we are on the same page: there is no more open todo and I can start with methods node-type?

ondrejmirtes:
Yes 😊

PR #3022, opened 2024-04-18
Detect unused method call on a separate line with possibly pure method

staabm:

### DIFF
--- a/src/Rules/DeadCode/CallToMethodStatementWithoutImpurePointsRule.php
+++ b/src/Rules/DeadCode/CallToMethodStatementWithoutImpurePointsRule.php
@@ -35,28 +35,26 @@ class CallToMethodStatementWithoutImpurePointsRule implements Rule
 		}
 
 		$errors = [];
-		foreach ($node->get(PossiblyPureMethodCallCollector::class) as $filePath => $collected) {
-			foreach ($collected as $data) {
-				foreach ($data as [$className, $method, $line]) {
-					if (!array_key_exists($className, $methods)) {
-						continue;
-					}
-
-					$lowerMethod = strtolower($method);
-					if (!array_key_exists($lowerMethod, $methods[$className])) {
-						continue;
-					}
-
-					$originalMethodName = $methods[$className][$lowerMethod];
-
-					$errors[] = RuleErrorBuilder::message(sprintf(
-						'Call to method %s() on a separate line has no effect.',
-						$originalMethodName,
-					))->file($filePath)
-						->line($line)
-						->identifier('method.resultUnused')
-						->build();
+		foreach ($node->get(PossiblyPureMethodCallCollector::class) as $filePath => $data) {
+			foreach ($data as [$className, $method, $line]) {
+				if (!array_key_exists($className, $methods)) {
+					continue;
 				}
+
+				$lowerMethod = strtolower($method);
+				if (!array_key_exists($lowerMethod, $methods[$className])) {
+					continue;
+				}
+
+				$originalMethodName = $methods[$className][$lowerMethod];
+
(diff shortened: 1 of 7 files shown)


src/Rules/DeadCode/PossiblyPureMethodCallCollector.php
```diff
@@ -0,0 +1,60 @@
...
+
+			$methodReflection = $classReflection->getMethod($methodName, $scope);
+			if (!$methodReflection->isPure()->maybe()) {
+				return null;
+			}
+			if (!$methodReflection->hasSideEffects()->maybe()) {
+				return null;
+			}
+
+			return [$methodReflection->getDeclaringClass()->getName(), $methodReflection->getName(), $node->getStartLine()];
```
staabm: this needs some love in case of multiple class-reflections

src/Rules/DeadCode/CallToMethodStatementWithoutImpurePointsRule.php
```diff
@@ -0,0 +1,64 @@
...
+	{
+		return CollectedDataNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$methods = [];
+		foreach ($node->get(MethodWithoutImpurePointsCollector::class) as $collected) {
+			foreach ($collected as [$className, $methodName, $classDisplayName]) {
+				if (!array_key_exists($className, $methods)) {
```
ondrejmirtes, line 32: class names are also case insensitive

src/Rules/DeadCode/PossiblyPureMethodCallCollector.php
```diff
@@ -0,0 +1,60 @@
...
+			return null;
+		}
+
+		$methodName = $node->expr->name->toString();
+		if (strtolower($methodName) === '__construct') {
+			return null;
+		}
+
+		$classType = $scope->getType($node->expr->var);
+		foreach ($classType->getObjectClassReflections() as $classReflection) {
```
ondrejmirtes: Please use `$scope->getMethodReflection()` instead of all of this.

UnionTypeMethodReflection returns `$this->methods[0]->getDeclaringClass()` as declaring class but we can solve that later. Thanks.

src/Rules/DeadCode/PossiblyPureMethodCallCollector.php
```diff
@@ -0,0 +1,60 @@
...
+			if (!$classReflection->hasMethod($methodName)) {
+				continue;
+			}
+
+			$methodReflection = $classReflection->getMethod($methodName, $scope);
+			if (!$methodReflection->isPure()->maybe()) {
```
ondrejmirtes: Skip if method is not final. Child class (polymorphic call) might have side effects.

ondrejmirtes (changes requested):

src/Rules/DeadCode/PossiblyPureMethodCallCollector.php
ondrejmirtes: If method is not final, it's sufficient that `$classReflection` is final. Also please write test cases for these.

staabm:
I cannot reproduce the stale-result-cache errors locally.

remaining build errors seem unrelated

ondrejmirtes:
FYI I added a commit https://github.com/phpstan/phpstan-src/pull/3022/commits/024c32a5154e49eae303f70ca9b8cefa8743a16a - I think these cases can be reported as well

ondrejmirtes:
(FYI We won't need to care about finality for StaticCall as we know exactly the class+method we're calling, there's no polymorphism at hand.)

ondrejmirtes:
Thank you!

PR #3023, opened 2024-04-18
Detect unused static method call on a separate line with possibly pure method

staabm:

### DIFF
--- a/src/Rules/DeadCode/CallToStaticMethodStatementWithoutImpurePointsRule.php
+++ b/src/Rules/DeadCode/CallToStaticMethodStatementWithoutImpurePointsRule.php
@@ -53,11 +53,11 @@ class CallToStaticMethodStatementWithoutImpurePointsRule implements Rule
 				$originalMethodName = $methods[$className][$lowerMethod];
 
 				$errors[] = RuleErrorBuilder::message(sprintf(
-					'Call to method %s() on a separate line has no effect.',
+					'Call to %s() on a separate line has no effect.',
 					$originalMethodName,
 				))->file($filePath)
 					->line($line)
-					->identifier('method.resultUnused')
+					->identifier('staticMethod.resultUnused')
 					->build();
 			}
 		}
--- a/src/Rules/DeadCode/CallToStaticMethodStatementWithoutImpurePointsRule.php
+++ b/src/Rules/DeadCode/CallToStaticMethodStatementWithoutImpurePointsRule.php
@@ -1,0 +1,68 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\DeadCode;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\CollectedDataNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function array_key_exists;
+use function sprintf;
+use function strtolower;
+
+/**
+ * @implements Rule<CollectedDataNode>
+ */
+class CallToStaticMethodStatementWithoutImpurePointsRule implements Rule
+{
+
+	public function getNodeType(): string
+	{
+		return CollectedDataNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$methods = [];
+		foreach ($node->get(MethodWithoutImpurePointsCollector::class) as $collected) {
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/DeadCode/data/call-to-static-method-without-impure-points.php
```diff
@@ -0,0 +1,104 @@
...
+function (): void {
+	X::myFunc();
+	X::myFUNC();
+	X::throwingFUNC();
+	X::throwingFunc();
+	X::funcWithRef();
+	X::impureFunc();
+	X::callingImpureFunc();
+
+	$a = X::myFunc();
```
staabm, line 14: I did not yet find a way to make the rule error on this `Expr`. the current `PossiblyPureStaticCallCollector` is based on `Expression` which this line is not.

Should I implement a separate Collector for `Expr`?

src/Rules/DeadCode/PossiblyPureStaticCallCollector.php
```diff
@@ -0,0 +1,56 @@
...
+			return null;
+		}
+		if (!$node->expr->name instanceof Node\Identifier) {
+			return null;
+		}
+
+		if ($node->expr->class instanceof Node\Name) {
+			$calledOnType = $scope->resolveTypeByName($node->expr->class);
+		} else {
+			$calledOnType = $scope->getType($node->expr->class);
```
ondrejmirtes: We shouldn't report anything if it's Expr. Keep it simple. We'd have to deal with the polymorphism again.

src/Rules/DeadCode/CallToStaticMethodStatementWithoutImpurePointsRule.php
```diff
@@ -0,0 +1,68 @@
...
+					$methods[$className] = [];
+				}
+				$methods[$className][strtolower($methodName)] = $classDisplayName . '::' . $methodName;
+			}
+		}
+
+		$errors = [];
+		foreach ($node->get(PossiblyPureStaticCallCollector::class) as $filePath => $data) {
+			foreach ($data as [$className, $method, $line]) {
+				$className = strtolower($className);
```
ondrejmirtes: $lowerClassName

ondrejmirtes (changes requested):

ondrejmirtes:
Also - the rules are not registered in config

tests/PHPStan/Rules/DeadCode/data/call-to-static-method-without-impure-points.php
ondrejmirtes: Why do you think this line should error?

staabm: I thought it should, because we do a similar error for functions: https://github.com/phpstan/phpstan-src/commit/57b0ae06e2fa4e2a43e9deb9f27afb8b24166bc9

ondrejmirtes: Nope, that was a fix so that this is NOT reported.

src/Rules/DeadCode/PossiblyPureStaticCallCollector.php
```diff
@@ -0,0 +1,55 @@
...
+			return null;
+		}
+
+		if (!$node->expr->class instanceof Node\Name) {
+			return null;
+		}
+
+		$methodName = $node->expr->name->toString();
+		$calledOnType = $scope->resolveTypeByName($node->expr->class);
+		$methodReflection = $scope->getMethodReflection($calledOnType, $methodName);
```
ondrejmirtes, line 40: We don't need to go through the type system at all. We have a class name and a method name as strings at hand, so we should go through ReflectionProvider and `ClassReflection::getNativeMethod()` instead. Thanks.

ondrejmirtes (changes requested):

src/Rules/DeadCode/PossiblyPureStaticCallCollector.php
ondrejmirtes: Oh scratch that, this actually avoids some other problems.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #3024, opened 2024-04-18
 Make the url key of metadata returned by stream_get_meta_data() optional

hirokinoue:
Fix: https://github.com/phpstan/phpstan/issues/10887


ondrejmirtes:
Thank you!

PR #3028, opened 2024-04-23
Fail when offset on array does not exist

ttomdewit:
Add initial step to fail when offset on an array does not exist.

All credit to https://github.com/ondrejmirtes

Fixes https://github.com/phpstan/phpstan/issues/7553


ondrejmirtes:
Thank you!

PR #3029, opened 2024-04-23
Fix 10922: array type in foreach

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10922


src/Analyser/NodeScopeResolver.php
```diff
@@ -951,7 +951,10 @@ private function processStmtNode(
 						]),
 					)));
 				} else {
```
VincentLanglet: I'm not sure about the purpose of 
```
if ($this->polluteScopeWithAlwaysIterableForeach) {
     $finalScope = $finalScope->mergeWith($scope->filterByTruthyValue(new BooleanOr(
	     new BinaryOp\Identical(
		     $stmt->expr,
			 new Array_([]),
	     ),
		 new FuncCall(new Name\FullyQualified('is_object'), [
			 new Arg($stmt->expr),
		 ]),
	)));
} else {
	$finalScope = $finalScope->mergeWith($scope);
}
```

But I soon as I make this change, the issue is solved.

I dunno
- if it has a bad impact (I can add tests if you had some in mind)
- if I should just remove the else part and the if condition.

src/Analyser/NodeScopeResolver.php
```diff
@@ -951,7 +951,10 @@ private function processStmtNode(
...
-					$finalScope = $finalScope->mergeWith($scope);
+					$finalScope = $finalScope->mergeWith($scope->filterByTruthyValue(new BinaryOp\Identical(
+						$stmt->expr,
+						new Array_([]),
+					)));
```
ondrejmirtes: Please remove the if-else as both these branches look like should be the same.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/Bug10922Test.php
```diff
@@ -0,0 +1,36 @@
...
+		...$args,
+	): void
+	{
+		$this->assertFileAsserts($assertType, $file, ...$args);
+	}
+
+	public static function getAdditionalConfigFiles(): array
+	{
+		return [
+			__DIR__ . '/../../../conf/bleedingEdge.neon',
```
ondrejmirtes, line 31: Merging with parent would be better (I think bleedingEdge is there already)

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/Bug10922Test.php
VincentLanglet: Parent is
```
/**
	 * @return string[]
	 */
	public static function getAdditionalConfigFiles(): array
	{
		return [];
	}
```

VincentLanglet: NodeScopeResolverTest use
```
	public static function getAdditionalConfigFiles(): array
	{
		return [
			__DIR__ . '/../../../conf/bleedingEdge.neon',
			__DIR__ . '/typeAliases.neon',
		];
	}
```
I followed the same idea

src/Analyser/NodeScopeResolver.php
VincentLanglet: done

ondrejmirtes:
Thank you.

PR #3031, opened 2024-04-24
ExponentiateHelper: fix internal error and internal warning

janedbal:
Probably fixes even https://github.com/phpstan/phpstan/issues/10125

```
Warning: A non-numeric value encountered on phar:///home/honza/Development/shipmonk/phpstan-rules/vendor/phpstan/phpstan/phpstan.phar/src/Type/ExponentiateHelper.php:72
```

Found in https://github.com/shipmonk-rnd/phpstan-rules tests when run with new https://github.com/phpstan/phpstan-src/commit/204ab2787f88bc31c4b2d30db0715997f3da4549

### DIFF
--- a/src/Type/ExponentiateHelper.php
+++ b/src/Type/ExponentiateHelper.php
@@ -4,8 +4,10 @@ namespace PHPStan\Type;
 
 use PHPStan\Type\Constant\ConstantFloatType;
 use PHPStan\Type\Constant\ConstantIntegerType;
+use Throwable;
 use function is_float;
 use function is_int;
+use function pow;
 
 final class ExponentiateHelper
 {
@@ -95,7 +97,12 @@ final class ExponentiateHelper
 		}
 
 		if ($exponent instanceof ConstantScalarType) {
-			$result = $base->getValue() ** $exponent->getValue();
+			try {
+				// @ to avoid "Warning: A non-numeric value encountered"
+				$result = @pow($base->getValue(), $exponent->getValue()); // @phpstan-ignore-line
+			} catch (Throwable) {
+				return new ErrorType();
+			}
 			if (is_int($result)) {
 				return new ConstantIntegerType($result);
 			}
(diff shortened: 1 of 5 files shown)


src/Type/ExponentiateHelper.php
```diff
@@ -95,7 +97,12 @@ private static function exponentiateConstantScalar(ConstantScalarType $base, Typ
 		}
 
 		if ($exponent instanceof ConstantScalarType) {
-			$result = $base->getValue() ** $exponent->getValue();
+			try {
+				// @ to avoid "Warning: A non-numeric value encountered"
+				$result = @pow($base->getValue(), $exponent->getValue()); // @phpstan-ignore-line
```
ondrejmirtes, line 102: I'd rather prevent the problem by checking the type of the values instead of silencing the error.

ondrejmirtes (changes requested):

src/Type/ExponentiateHelper.php
janedbal: That would require some crazy regexes. I dont think there is a simple way to simulate the magical numeric-string behaviour: https://3v4l.org/ECZjp#v8.3.6

This way we are sure we get exactly what PHP evalutes.

ondrejmirtes: Looks like `is_numeric` would do the job: https://3v4l.org/sYolO#v8.3.6

janedbal: Not really, those are both numeric, but results in different output type: https://3v4l.org/aDslS#v8.3.6

staabm: There is a similar open PR you might can get inspiration from

https://github.com/phpstan/phpstan-src/pull/2796

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/pull/2796

PR #3035, opened 2024-04-30
Make rule errors a little bit more DRY

axlon:
Noticed this message is used a lot throughout the codebase, and it could probably use a bit of DRYness, this would make it easier to change the message down the line it ever became necessary.

Additionally, this also makes it easier for PHPStan extensions' rules to use the exact same tip as PHPStan itself, if they do respect this setting being on or off


ondrejmirtes:
Thank you! 👍

PR #3037, opened 2024-05-03
Fix `count($x) > $n === true` negation

rvanvelzen:
Fix phpstan/phpstan#10952

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -243,13 +243,13 @@ class TypeSpecifier
 				&& $leftType->isInteger()->yes()
 			) {
 				if (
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Awesome, thank you!

rvanvelzen:
The same problem actually applies to `strlen()`, maybe one day I'll submit a PR for that too :)

staabm:
> The same problem actually applies to `strlen()`, maybe one day I'll submit a PR for that too :)

@rvanvelzen  I tried replicating this problem for `strlen()` but did not succeed:
https://phpstan.org/r/fa21633d-7311-406a-8959-cbb4a488a61d

rvanvelzen:
@staabm the problem occurs when wrapping the condition with `(...) === true`: https://phpstan.org/r/8adcd9e6-501f-4e37-a175-001c753a4153

PR #3039, opened 2024-05-03
Support multi-byte string function variants

staabm:


src/Type/Php/SubstrDynamicReturnTypeExtension.php
```diff
@@ -25,7 +26,7 @@ class SubstrDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExten
 
 	public function isFunctionSupported(FunctionReflection $functionReflection): bool
 	{
-		return $functionReflection->getName() === 'substr';
+		return in_array($functionReflection->getName(), ['substr', 'mb_substr'], true);
```
rvanvelzen, line 30: The `substr()` calls below should be changed to use the correct function as well.

rvanvelzen:
Some tests with multibyte characters would be good as well 👍

src/Type/Php/SubstrDynamicReturnTypeExtension.php
staabm: great catch, thx.

ondrejmirtes:
Thank you.

PR #3040, opened 2024-05-03
Fix `strlen($x) > $n === true` negation

staabm:
as noted in https://github.com/phpstan/phpstan-src/pull/3037#issuecomment-2092539870


ondrejmirtes:
Thank you!

PR #1791, opened 2022-10-04
Make array access on possible false tolerant with isset

ahmedash95:
Closes https://github.com/phpstan/phpstan/issues/8068.

Hey Ondrej,

As discussed in the bug. PHP is tolerant with `isset`.  

The removed line `Cannot access offset \'a\' on array{a: 1, b: 1}|(Closure(): void).` from the other test is reasonable because closure key access in isset would always be reported.

appreciate your help on getting the fix for this one.

PR #3042, opened 2024-05-05
Revert "Make array access on possible false tolerant with isset"

rajyan:
Reverts phpstan/phpstan-src#1791


rajyan:
As explained in

https://github.com/phpstan/phpstan/issues/8393

we need to handle isOffsetAccessible differently in `isset`.
I'm first reverting this, because offset access on 'Closure' raises an error.
https://github.com/phpstan/phpstan-src/pull/1791#discussion_r1025871491

We have to handle isOffsetAccsible->maybe of 'accessibleType|false' and 'accessibleType|closure' differently.

PR #3045, opened 2024-05-05
feat: implement isOffsetAccessLegal

rajyan:
Details about isOffsetAccessLegal are explained in https://github.com/phpstan/phpstan/issues/8393

ref https://github.com/phpstan/phpstan-src/pull/1791#discussion_r1025871491
closes https://github.com/phpstan/phpstan/issues/8393
resolves https://github.com/phpstan/phpstan/issues/10926

### DIFF
--- a/src/Rules/Arrays/NonexistentOffsetInArrayDimFetchRule.php
+++ b/src/Rules/Arrays/NonexistentOffsetInArrayDimFetchRule.php
@@ -64,7 +64,7 @@ class NonexistentOffsetInArrayDimFetchRule implements Rule
 			return [];
 		}
 
-		if ($scope->isUndefinedExpressionAllowed($node) && !$isOffsetAccessible->no()) {
+		if ($scope->isUndefinedExpressionAllowed($node) && $isOffsetAccessibleType->isOffsetAccessLegal()->yes()) {
 			return [];
 		}
 
--- a/src/Type/NonexistentParentClassType.php
+++ b/src/Type/NonexistentParentClassType.php
@@ -154,7 +154,7 @@ class NonexistentParentClassType implements Type
 
 	public function isOffsetAccessLegal(): TrinaryLogic
 	{
-		return TrinaryLogic::createYes();
+		return TrinaryLogic::createNo();
 	}
 
 	public function isScalar(): TrinaryLogic
--- a/src/Type/NonexistentParentClassType.php
+++ b/src/Type/NonexistentParentClassType.php
@@ -152,6 +152,11 @@ class NonexistentParentClassType implements Type
 		return new ErrorType();
 	}
 
+	public function isOffsetAccessLegal(): TrinaryLogic
+	{
+		return TrinaryLogic::createYes();
+	}
+
 	public function isScalar(): TrinaryLogic
 	{
 		return TrinaryLogic::createNo();
--- a/src/Type/Traits/NonOffsetAccessibleTypeTrait.php
+++ b/src/Type/Traits/NonOffsetAccessibleTypeTrait.php
@@ -14,11 +14,6 @@ trait NonOffsetAccessibleTypeTrait
 		return TrinaryLogic::createNo();
 	}
 
-	public function isOffsetAccessLegal(): TrinaryLogic
-	{
-		return TrinaryLogic::createYes();
-	}
-
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
 		return TrinaryLogic::createNo();
--- a/src/Type/Traits/NonOffsetAccessibleTypeTrait.php
+++ b/src/Type/Traits/NonOffsetAccessibleTypeTrait.php
@@ -14,6 +14,11 @@ trait NonOffsetAccessibleTypeTrait
 		return TrinaryLogic::createNo();
 	}
 
+	public function isOffsetAccessLegal(): TrinaryLogic
+	{
+		return TrinaryLogic::createYes();
+	}
+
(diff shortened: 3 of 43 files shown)


tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -115,6 +115,10 @@ public function testRule(): void
 				'Cannot access offset \'a\' on Closure(): void.',
 				253,
 			],
+			[
+				'Cannot access offset \'a\' on array{a: 1, b: 1}|(Closure(): void).',
+				258,
+			],
```
rajyan, line 121: This is the behavior I wanted to revert.
ClosureType (Object not implementing ArrayAcces) throws fatal error.

src/Type/Traits/NonOffsetAccessibleTypeTrait.php
```diff
@@ -14,6 +14,11 @@ public function isOffsetAccessible(): TrinaryLogic
 		return TrinaryLogic::createNo();
 	}
 
+	public function isOffsetAccessLegal(): TrinaryLogic
+	{
+		return TrinaryLogic::createYes();
+	}
```
rajyan: The most important difference between `isOffsetAccessible` is this line. Types like integer, boolean are not offset accessible but doesn't throw a fatal error on access in `isset`.

src/Rules/Arrays/NonexistentOffsetInArrayDimFetchRule.php
```diff
@@ -64,7 +64,7 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
-		if ($scope->isUndefinedExpressionAllowed($node) && !$isOffsetAccessible->no()) {
+		if ($scope->isUndefinedExpressionAllowed($node) && $isOffsetAccessibleType->isOffsetAccessLegal()->yes()) {
```
rajyan, line 67: Maybe it's better to add a tip for `!$isOffsetAccessibleType->isOffsetAccessLegal()->yes()` case.

* The `isOffsetAccessibleType` should be narrowed down to a type excluding `ObjectType` that do not implement `ArrayAccess::class`
* The class implementing `ArrayAccess::Class` should be marked `final`

src/Type/Traits/NonOffsetAccessibleTypeTrait.php
ondrejmirtes: I don't like this being added to a trait with a name like this. I'd like a separate trait for this method that would be used in all the places where relevant. Or maybe no trait at all, and just put this implementation in all the places where it applies.

src/Rules/Arrays/NonexistentOffsetInArrayDimFetchRule.php
ondrejmirtes: > The class implementing ArrayAccess::Class should be marked final

I don't get this part. It doesn't matter if a class is final when it already implements `ArrayAccess`. What a finality of such class would achieve?

ondrejmirtes (changes requested):
Otherwise awesome! 👍

src/Rules/Arrays/NonexistentOffsetInArrayDimFetchRule.php
rajyan: This was just a mistake. I was mixed up with the fact that we cant say `isOffsetAccessLegal()->no()` if the class is not final, because child class might implement ArrayClass.

https://github.com/phpstan/phpstan/issues/10926#issuecomment-2090771360

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -678,6 +678,10 @@ public function testBug8068(): void
 				"Cannot access offset 'path' on Closure.",
 				18,
 			],
+			[
+				"Cannot access offset 'path' on iterable<int|string, object>.",
+				26,
+			],
```
rajyan, line 684: ref https://3v4l.org/hIimN

src/Type/NonexistentParentClassType.php
```diff
@@ -152,6 +152,11 @@ public function toArrayKey(): Type
 		return new ErrorType();
 	}
 
+	public function isOffsetAccessLegal(): TrinaryLogic
+	{
+		return TrinaryLogic::createYes();
```
ondrejmirtes: I think this should be `no`, and I think we can test it with `$p = new parent();` and trying offset access on that.

ondrejmirtes (changes requested):
One levels test needs updating (just re-run it and commit the generated JSON), thanks

src/Type/NonexistentParentClassType.php
rajyan: You are definitely right! Thank you.

Also, found an interesting result that NonExistentParentClassType behavior has changed from 7.4.

We cannot even declare a NonExistentParentClassType from PHP 8.0. 
https://3v4l.org/7TR10

Calling array access is an error across all versions.
https://3v4l.org/ClM1B

ondrejmirtes: We can just skip the file from linting in Makefile https://github.com/phpstan/phpstan-src/blob/78d3d99d9306b045a5464c19c9afcca5e5d7b1e5/Makefile#L76

ondrejmirtes:
Awesome, thank you!

rajyan:
Thank you for your swift responses as always!

PR #3049, opened 2024-05-06
Feature check redeclared readonly properties v2

schlndh:
This is a relaxed (and reimplemented) version of the rule from: https://github.com/phpstan/phpstan-src/pull/3041#issuecomment-2094153967

> Considering this example: https://phpstan.org/r/cf781e9e-cfdd-4aae-855d-ead64bf16afa - if (a subclass redeclares readonly property in promoted fashion OR assigns the redeclared readonly property) AND calls parent constructor AND parent constructor declares the readonly property in promoted fashion, then we should always report an error. This can be achieved with local analysis only, without performance impact. PHPStan should see that as duplicate assignment with existing rules in place. So MissingReadOnlyPropertyAssignRule should report "is already assigned" just by modifying the data that is fed into it.

Fixes https://github.com/phpstan/phpstan/issues/8101
Fixes https://github.com/phpstan/phpstan/issues/9863
Fixes https://github.com/phpstan/phpstan/issues/9864

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -177,6 +177,7 @@ use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\TypeUtils;
 use PHPStan\Type\UnionType;
+use ReflectionProperty;
 use Throwable;
 use Traversable;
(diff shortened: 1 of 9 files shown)


ondrejmirtes:
Awesome job! Thank you.

PR #3050, opened 2024-05-06
Introduce `new` type

ruudk:
This makes it possible that a new instance of a class-string will be returned.

```php
/**
 * @var array<string, class-string>
 */
private const TYPES = [
	'foo' => DateTime::class,
	'bar' => DateTimeImmutable::class,
];

/**
 * @template T of key-of<self::TYPES>
 * @param T $type
 *
 * @return new<self::TYPES[T]>
 */
public static function get(string $type) : ?object
{
	$class = self::TYPES[$type];
	return new $class('now');
}
```

See https://github.com/phpstan/phpstan/issues/9704

The work was [done](https://github.com/phpstan/phpstan/issues/9704#issuecomment-2093128708) by @rvanvelzen in a gist. I just created the PR for it.

### DIFF
--- a/src/Type/NewObjectType.php
+++ b/src/Type/NewObjectType.php
@@ -1,0 +1,104 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type;
+
+use PHPStan\PhpDocParser\Ast\Type\GenericTypeNode;
+use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
+use PHPStan\PhpDocParser\Ast\Type\TypeNode;
+use PHPStan\Type\Generic\TemplateTypeVariance;
+use PHPStan\Type\Traits\LateResolvableTypeTrait;
+use PHPStan\Type\Traits\NonGeneralizableTypeTrait;
+use function sprintf;
+
+/** @api */
+class NewObjectType implements CompoundType, LateResolvableType
+{
+
+	use LateResolvableTypeTrait;
+	use NonGeneralizableTypeTrait;
+
+	public function __construct(private Type $type)
+	{
+	}
+
+	public function getType(): Type
+	{
+		return $this->type;
+	}
+
+	public function getReferencedClasses(): array
+	{
+		return $this->type->getReferencedClasses();
+	}
+
+	public function getReferencedTemplateTypes(TemplateTypeVariance $positionVariance): array
+	{
+		return $this->type->getReferencedTemplateTypes($positionVariance);
+	}
+
+	public function equals(Type $type): bool
+	{
+		return $type instanceof self
+			&& $this->type->equals($type->type);
+	}
+
+	public function describe(VerbosityLevel $level): string
+	{
+		return sprintf('new<%s>', $this->type->describe($level));
+	}
+
+	public function isResolvable(): bool
+	{
+		return !TypeUtils::containsTemplateType($this->type);
+	}
+
+	protected function getResult(): Type
+	{
+		return $this->type->getObjectTypeOrClassStringObjectType();
+	}
+
+	/**
+	 * @param callable(Type): Type $cb
+	 */
+	public function traverse(callable $cb): Type
+	{
+		$type = $cb($this->type);
+
+		if ($this->type === $type) {
+			return $this;
+		}
+
+		return new self($type);
+	}
+
+	public function traverseSimultaneously(Type $right, callable $cb): Type
+	{
+		if (!$right instanceof self) {
+			return $this;
+		}
+
+		$type = $cb($this->type, $right->type);
+
+		if ($this->type === $type) {
+			return $this;
+		}
+
+		return new self($type);
+	}
+
+	public function toPhpDocNode(): TypeNode
+	{
+		return new GenericTypeNode(new IdentifierTypeNode('new'), [$this->type->toPhpDocNode()]);
+	}
+
+	/**
+	 * @param mixed[] $properties
+	 */
+	public static function __set_state(array $properties): Type
+	{
+		return new self(
+			$properties['type'],
+		);
+	}
+
+}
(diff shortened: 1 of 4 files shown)


src/Type/NewObjectType.php
```diff
@@ -0,0 +1,104 @@
...
+
+	public function equals(Type $type): bool
+	{
+		return $type instanceof self
+			&& $this->type->equals($type->type);
+	}
+
+	public function describe(VerbosityLevel $level): string
+	{
+		return sprintf('object<%s>', $this->type->describe($level));
```
ondrejmirtes: `new<%s>`

src/Type/NewObjectType.php
```diff
@@ -0,0 +1,104 @@
...
+		if ($this->type === $type) {
+			return $this;
+		}
+
+		return new self($type);
+	}
+
+	public function toPhpDocNode(): TypeNode
+	{
+		return new GenericTypeNode(new IdentifierTypeNode('object'), [$this->type->toPhpDocNode()]);
```
ondrejmirtes: `new`

ondrejmirtes (changes requested):
Nice!

ruudk:
Made some progress.

Interestingly this works:
```php
/**
 * @template M of self::TYPES
 * @template T of key-of<M>
 * @param T $type
 *
 * @return new<M[T]>
 */
public static function get(string $type) : object
{
	$class = self::TYPES[$type];
	return new $class('now');
}
```

But this doesn't:
```php
/**
 * @template T of key-of<self::TYPES>
 * @param T $type
 *
 * @return new<self::TYPES[T]>
 */
public static function get2(string $type) : object
{
	$class = self::TYPES[$type];
	return new $class('now');
}
/**
 * @template T of class-string
 * @param T $type
 *
 * @return new<T>
 */
public static function create(string $type) : object
{
	return new $type();
}
```

I have the feeling this has something to do with \PHPStan\Type\NewObjectType::isResolvable:
```php
public function isResolvable(): bool
{
	return !TypeUtils::containsTemplateType($this->type);
}
```

Maybe we should allow `class-string` there...

ruudk:
This is already possible, without `new`:
```php
/**
 * @template T of object
 * @param class-string<T> $type
 *
 * @return T
 */
public static function create(string $type) : object
{
	return new $type();
}
```

ondrejmirtes:
> have the feeling this has something to do with \PHPStan\Type\NewObjectType::isResolvable:

I don't think it does. Other late-resolvable types look the same. You need to launch Xdebug and see what's going on :)

ruudk:
Updated the PR.  I think this is ready now.

`new` is only needed when you are working with templated types like:
```php
/**
 * @template M of self::TYPES
 * @template T of key-of<M>
 * @param T $type
 *
 * @return new<M[T]>
 */
public static function get(string $type) : object
{
	$class = self::TYPES[$type];
	return new $class('now');
}
```

For simple cases, you can use:
```php
/**
 * @template T of object
 * @param class-string<T> $type
 *
 * @return T
 */
public static function create(string $type) : object
{
	return new $type();
}
```

ondrejmirtes:
What was the problem described here? https://github.com/phpstan/phpstan-src/pull/3050#issuecomment-2095548138

ruudk:
I guess writing it like this just will not be possible:
```php
/**
 * @template T of key-of<self::TYPES>
 * @param T $type
 *
 * @return new<self::TYPES[T]>
 */
public static function get2(string $type) : object
{
	$class = self::TYPES[$type];
	return new $class('now');
}
```

When I use [your suggested syntax](https://github.com/phpstan/phpstan/issues/9704#issuecomment-1659049214) (with another template), it works.

ondrejmirtes:
@ruudk Can you please send this working example with a description to this article? https://phpstan.org/blog/generics-by-examples

Source is here: https://github.com/phpstan/phpstan/blob/1.11.x/website/src/_posts/generics-by-examples.md

rvanvelzen:
`self::TYPES[T]` is a parse error currently which is why it doesn't work. Needs to be fixed in the phpdoc-parser.

Writing `(self::TYPES)[T]` is a workaround.

ruudk:
@rvanvelzen Nice! I added this example also to the test.

ruudk:
> Can you please send this working example with a description to this article? https://phpstan.org/blog/generics-by-examples

Done in https://github.com/phpstan/phpstan/pull/10971

ruudk:
> `self::TYPES[T]` is a parse error currently which is why it doesn't work. Needs to be fixed in the phpdoc-parser.
> 
> Writing `(self::TYPES)[T]` is a workaround.

@rvanvelzen Interesting,

Maybe [that fix](https://github.com/phpstan/phpdoc-parser/pull/237) will make this possible without `new`?

```
@template T of key-of<self::TYPES>
@template O of class-string<self::TYPES[T]>
@param T $type
@return O
```

Edit: tried it, doesn't work.

ondrejmirtes:
The improvement to phpdoc-parser has been merged here, please rebase this branch and try out what it improves (add tests for previously impossible use-cases).

You can also try it out in the playground https://phpstan.org/try, it has been updated too.

ruudk:
@ondrejmirtes Rebased. With the fix in phpdoc-parser we can remove the parenthesis. This simplifies the syntax. Updated the [example](https://github.com/phpstan/phpstan/pull/10971) too.

ondrejmirtes:
Awesome, thank you!

ondrejmirtes:
Did we just remove the need for `class-string-map`? This looks awfully similar:

```php
	/**
	 * @template T of key-of<self::TYPES>
	 * @param T $type
	 *
	 * @return new<self::TYPES[T]>
	 */
```

ondrejmirtes:
No, probably not, see https://github.com/phpstan/phpstan/issues/9521

PR #3052, opened 2024-05-07
fix: stdClass is not a universal base class

rajyan:
https://www.php.net/manual/en/class.stdclass.php

I think we can subtract more than `stdClass` on json_decode when force associative array option is enabled.


rajyan:
I wanted to test this behavior on `NonexistentOffsetInArrayDimFetchRule`, but noticed that we cannot subtract `StrictMixedType`.

ondrejmirtes:
Thank you very much!

PR #3053, opened 2024-05-08
Specify types on match() condition correctly

rvanvelzen:
Fix phpstan/phpstan#10974
Fix phpstan/phpstan#9694


ondrejmirtes:
Thank you very much!

PR #3054, opened 2024-05-08
Improve type for predefined constant

zonuexe:
All these constants are strings that are neither `''` nor `'0'` and are replaced by `AccessoryNonFalsyStringType`.

### DIFF
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -6,7 +6,7 @@ use PhpParser\Node\Name;
 use PHPStan\Reflection\NamespaceAnswerer;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Reflection\ReflectionProvider\ReflectionProviderProvider;
-use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
+use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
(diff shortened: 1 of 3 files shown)


src/Analyser/ConstantResolver.php
```diff
@@ -232,7 +232,7 @@ public function resolvePredefinedConstant(string $resolvedConstantName): ?Type
 			return IntegerRangeType::fromInterval(1, null);
 		}
 		if ($resolvedConstantName === '__COMPILER_HALT_OFFSET__') {
-			return IntegerRangeType::fromInterval(0, null);
+			return IntegerRangeType::fromInterval(1, null);
```
zonuexe, line 235: Zero-byte PHP code itself cannot output `__COMPILER_HALT_OFFSET__`, so it is always `>=1`.

The minimum value is probably 48, but it looks unnatural, so leave it above 1.

* https://php-play.dev/?c=DwfgpgxgFg9gBAfQQYQPIFkAKBJAMgUQCUEAJAQVwBUFUAxWgZX2oQG4koBDAGwBcEIMALYAHAJbcwAJwAUASlZA&v=8.3&f=html

herndlm:
I added those in what I think was my first ever phpstan contribution 😊

ondrejmirtes:
Thank you!

PR #3055, opened 2024-05-08
Make ResultCacheEndToEndTest pass level 9

AJenbo:
This takes care of 26 out of 817 reported issues at level 9. It might take some time to get everything to this level, but it has to start some where right? Also using this to try out the waters on this path, let me know what you think?


ondrejmirtes:
Thank you! I like it, I merged it but it doesn't mean I'll ever use level 9 here 😀

PR #3056, opened 2024-05-09
IgnoreLexer: drop support for multiline identifiers

janedbal:


ondrejmirtes (changes requested):
1) Please keep the EOL token in the array. It should be used in `parseIdentifiers` to stop parsing.
2) There should be a test with `@phpstan-ignore` followed by a new line to cause a parse error.
3) We should stop supporting trailing comma, there should be a test for that.

janedbal:
Added as you want. I also added more tests for current behaviour, but I think it should not be that lax and it should report any unsupported usages. Maybe, I'll try to send another MR that adjusts that.

janedbal:
Removed the snapshot, will create separate MR with that.

ondrejmirtes:
Thank you!

PR #3057, opened 2024-05-09
Fixes bug 10980

greew:
When while loop condition is a `while(true)`, the `continue` exit points are never used, and shouldn't be calculated in the final result.

Fixes [#phpstan/phpstan/10980](https://github.com/phpstan/phpstan/issues/10980)


ondrejmirtes:
Perfect, thank you! Even more awesome given this is your first contribution :) Here's to many more! 🍻

greew:
Well... I'm already sponsoring you, but might as well chip in with the code as well :grin:

PR #3059, opened 2024-05-10
Lazy create ClassReflection in enum-cases

staabm:
the enum-case object will lazy create the class-reflection when used the first time.
in our slow repro case this means we don't need to create class-reflection for millions of objects over and over, since we are only check `isSuperTypeOf`.

refs https://github.com/phpstan/phpstan/issues/10979

before this PR

```
time php bin/phpstan analyze --debug LibraryMasterEnum.php  
23.83s user 0.15s system 99% cpu 24.065 total
```

after this PR:

```
time php bin/phpstan analyze --debug LibraryMasterEnum.php  
17.21s user 0.16s system 98% cpu 17.610 total
```

not perfect yet, but still good progress

### DIFF
--- a/src/Type/Enum/EnumCaseObjectType.php
+++ b/src/Type/Enum/EnumCaseObjectType.php
@@ -79,7 +79,7 @@ class EnumCaseObjectType extends ObjectType
 			return $type->isSubTypeOf($this);
 		}
 
-		$parent = new parent($this->getClassName(), $this->getSubtractedType());
+		$parent = new parent($this->getClassName(), $this->getSubtractedType(), $this->getClassReflectionOrNull());
 
 		return $parent->isSuperTypeOf($type)->and(TrinaryLogic::createMaybe());
 	}
@@ -165,7 +165,7 @@ class EnumCaseObjectType extends ObjectType
 
 	public function generalize(GeneralizePrecision $precision): Type
 	{
-		return new parent($this->getClassName());
+		return new parent($this->getClassName(), null, $this->getClassReflectionOrNull());
 	}
 
 	public function isSmallerThan(Type $otherType): TrinaryLogic
--- a/src/Type/Enum/EnumCaseObjectType.php
+++ b/src/Type/Enum/EnumCaseObjectType.php
@@ -165,7 +165,7 @@ class EnumCaseObjectType extends ObjectType
 
 	public function generalize(GeneralizePrecision $precision): Type
 	{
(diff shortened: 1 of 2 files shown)


staabm:
compared against the latest 1.11.x for the 10979 case

before 
```
php bin/phpstan analyze --debug LibraryMasterEnum.php  1.99s user 0.13s system 98% cpu 2.145 total

php bin/phpstan analyze --debug LibraryMasterEnum.php  1.98s user 0.13s system 99% cpu 2.127 total

php bin/phpstan analyze --debug LibraryMasterEnum.php  1.96s user 0.13s system 99% cpu 2.106 total
```

after

```
php bin/phpstan analyze --debug LibraryMasterEnum.php  1.94s user 0.14s system 99% cpu 2.096 total

php bin/phpstan analyze --debug LibraryMasterEnum.php  1.93s user 0.13s system 99% cpu 2.067 total

php bin/phpstan analyze --debug LibraryMasterEnum.php  1.92s user 0.13s system 99% cpu 2.070 total
```

blackfire diff

grafik

src/Type/Enum/EnumCaseObjectType.php
```diff
@@ -79,7 +79,7 @@ public function isSuperTypeOf(Type $type): TrinaryLogic
 			return $type->isSubTypeOf($this);
 		}
 
-		$parent = new parent($this->getClassName(), $this->getSubtractedType(), $this->getClassReflection());
+		$parent = new parent($this->getClassName(), $this->getSubtractedType());
```
ondrejmirtes: The problem with this is loss of precision. ObjectType often has ClassReflection given be the current scope. It makes analysis more precise if there are multiple classes with the same name - it's going to use reflection from the current file for real.

I have an idea: use classReflection if it's been set but don't fetch it from ReflectionProvider.

ondrejmirtes (changes requested):

src/Type/Enum/EnumCaseObjectType.php
staabm: it seems we are missing test-coverage for the case you mentioned.

just added a new method to get but not create the class-reflection of the base-class

ondrejmirtes:
Is it still faster than before?

staabm:
hmm .. you are right.. the perf benefit nearly went away with the lastest change.

it seems EnumCaseObjectType is slower when the class-reflection is known.

since the perf benefit is not that much, I think we can stop here.

PR #3060, opened 2024-05-10
Enum-case comparison: cheap checks first

staabm:
calls to `getClassName()` show up in the profile of the https://github.com/phpstan/phpstan/issues/10979 repro.

grafik

not super important but an easy low hanging fruit

### DIFF
--- a/src/Type/Enum/EnumCaseObjectType.php
+++ b/src/Type/Enum/EnumCaseObjectType.php
@@ -53,8 +53,8 @@ class EnumCaseObjectType extends ObjectType
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


staabm:
after the root cause patch landed with #3062 this PR still shows a 14% improvement on the repro case of #10979

grafik

before
```
php bin/phpstan analyze --debug LibraryMasterEnum.php
  2.06s user 0.13s system 99% cpu 2.216 total
```

after
```
php bin/phpstan analyze --debug LibraryMasterEnum.php
  2.00s user 0.13s system 98% cpu 2.157 total
```

ondrejmirtes:
Thank you.

PR #3062, opened 2024-05-10
Added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/10979

~~reverting the fix for https://github.com/phpstan/phpstan-src/pull/2985 does not regress the test of said PR.
I need to dive deeper why the test no longer reproduces the perf problem we had at the time #2985 was created~~

I had a typo locally .. investigating..

### DIFF
--- a/src/Type/ObjectType.php
+++ b/src/Type/ObjectType.php
@@ -94,6 +94,9 @@ class ObjectType implements TypeWithClassName, SubtractableType
 
 	private ?string $cachedDescription = null;
 
+	/** @var array<string, list<EnumCaseObjectType>> */
+	private static array $enumCases = [];
+
 	/** @api */
 	public function __construct(
 		private string $className,
@@ -114,6 +117,7 @@ class ObjectType implements TypeWithClassName, SubtractableType
 		self::$methods = [];
 		self::$properties = [];
 		self::$ancestors = [];
+		self::$enumCases = [];
 	}
 
 	private static function createFromReflection(ClassReflection $reflection): self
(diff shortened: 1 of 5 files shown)


src/Type/TypeCombinator.php
```diff
@@ -201,7 +200,8 @@ public static function union(Type ...$types): Type
 			if ($types[$i] instanceof StringType && !$types[$i] instanceof ClassStringType) {
 				$hasGenericScalarTypes[ConstantStringType::class] = true;
 			}
-			if ($types[$i] instanceof EnumCaseObjectType) {
+			$enumCases = $types[$i]->getEnumCases();
```
staabm, line 203: reverts https://github.com/phpstan/phpstan-src/pull/2985

staabm:
after this PR we now have both cases fast:

```
php bin/phpstan analyze --debug bug-10772.php  1.30s user 0.14s system 98% cpu 1.451 total


php bin/phpstan analyze --debug LibraryMasterEnum.php  3.16s user 0.14s system 99% cpu 3.328 total
```

staabm:
will be AFK for a few hours

src/Type/ObjectType.php
```diff
@@ -94,6 +94,9 @@ class ObjectType implements TypeWithClassName, SubtractableType
 
 	private ?string $cachedDescription = null;
 
+	/** @var array<class-string, list<EnumCaseObjectType>> */
+	private static array $enumCasesCache = [];
```
ondrejmirtes: Needs to be reset in `resetCaches()`

src/Type/ObjectType.php
```diff
@@ -1215,23 +1218,32 @@ public function getEnumCases(): array
 			return [];
 		}
 
-		$subtracted = [];
-		if ($this->subtractedType !== null) {
-			foreach ($this->subtractedType->getEnumCases() as $enumCase) {
-				$subtracted[$enumCase->getEnumCaseName()] = true;
+		$className = $classReflection->getName();
+		if (array_key_exists($className, self::$enumCasesCache)) {
```
ondrejmirtes: The cache key needs to be `$this->describeCache()`.

ondrejmirtes (changes requested):
Yeah this looks good 👍

staabm:
this one is good to go. I will have a look at the other remaining open PRs after this one is merged, whether there is still meaningful impact

```
php bin/phpstan analyze --debug LibraryMasterEnum.php  
  2.02s user 0.13s system 99% cpu 2.160 total

php bin/phpstan analyze --debug bug-10772.php
  1.26s user 0.14s system 98% cpu 1.423 total
```

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1338,6 +1338,16 @@ public function testBug10772(): void
...
 	}
 
+	public function testBug10979(): void
+	{
+		if (PHP_VERSION_ID < 80100) {
+			$this->markTestSkipped('Test requires PHP 8.1.');
+		}
+
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-10979.php');
+		$this->assertCount(80, $errors);
```
ondrejmirtes: Can you please fix the example so that it's valid and does not report any errors? And make sure it still reproduces the performance problem.

I'm worrying that the number 80 is going to fluctuate and is going to fail quite often as we work on PHPStan.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
staabm: we now have a zero errors repro

ondrejmirtes:
Thank you!

PR #3063, opened 2024-05-12
NULL and null as a property default value are treated differently when overriding a parent property

staabm:
closes https://github.com/phpstan/phpstan/issues/10987

I looked thru the codebase but could not see any other occurence with the same problem


src/Rules/Properties/DefaultValueTypesAssignedToPropertiesRule.php
```diff
@@ -39,7 +39,7 @@ public function processNode(Node $node, Scope $scope): array
 		$propertyReflection = $classReflection->getNativeProperty($node->getName());
 		$propertyType = $propertyReflection->getWritableType();
 		if ($propertyReflection->getNativeType() instanceof MixedType) {
-			if ($default instanceof Node\Expr\ConstFetch && (string) $default->name === 'null') {
+			if ($default instanceof Node\Expr\ConstFetch && $default->name->toLowerString() === 'null') {
```
staabm, line 42: btw: the origin reporter mentioned he expected 2 errors instead. since we already have this special case for native `mixed` I went with 0 errors for consistency.

ondrejmirtes:
Awesome, thank you!

PR #3064, opened 2024-05-12
Fix parent-keyword case-sensitivity

staabm:
similar to https://github.com/phpstan/phpstan-src/pull/3063 but for the `parent` keyword.

this time I did not look only for `'null'` but all types of hardcoded keyword comparisons 😅


ondrejmirtes:
Thank you!

PR #3065, opened 2024-05-12
Fix bug #10862

greew:
This fix ensures that we use correct array indicies in >= PHP8.3. 

(See https://www.php.net/manual/en/migration83.incompatible.php#migration83.incompatible.core.negative-index-to-empty-array).

We ensure, that if
- PHP version is 8.3+
- AND the current element to add is the first element 
- AND the key is less than 0 (it's negative)

then we set the newAutoIndexes to the negative value + 1.

Added tests to show correctness for both < PHP8.3 and >= PHP8.3

This should fix bug https://github.com/phpstan/phpstan/issues/10862

---

I'm not sure that I quite understood the `$optional` part - also, if you prefer having tests created in another way than each in their separate file (or something else), please let me know! :)


src/Type/Constant/ConstantArrayTypeBuilder.php
```diff
@@ -181,6 +182,14 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $opt
 							$this->nextAutoIndexes[] = $newAutoIndex;
 						}
 					}
+					if (PHP_VERSION_ID >= 80300 && count($this->keyTypes) === 1 && $offsetType->getValue() < 0) {
```
staabm: Usually php feature detects are implemented in PhpVersion class

greew: Great!! It's exactly these kinds of comments, I need! 

I'll try and figure out how that's working! 😊

greew:
I'm not sure this is the right solution, but since `ConstantArrayTypeBuilder` is not created by DI, I don't know how else to insert the `PhpVersion` class into the class.

herndlm:
> I'm not sure this is the right solution, but since `ConstantArrayTypeBuilder` is not created by DI, I don't know how else to insert the `PhpVersion` class into the class.

```php
$phpVersion = PhpVersionStaticAccessor::getInstance();
```

just directly where you need it apparently. It's already used somewhere in `ConstantArrayType`

greew:
> > I'm not sure this is the right solution, but since `ConstantArrayTypeBuilder` is not created by DI, I don't know how else to insert the `PhpVersion` class into the class.
> 
> ```
> $phpVersion = PhpVersionStaticAccessor::getInstance();
> ```
> 
> just directly where you need it apparently. It's already used somewhere in `ConstantArrayType`

Greeeeat - thank you very much!! :smiley:

ondrejmirtes:
I feel like this fix should be done differently.

Because PHPStan has a wrong assumption even on previous PHP versions: https://phpstan.org/r/492b583a-dc4b-4be2-a4c9-2232e039c19c (2nd key in `$a` should be `-3` on all PHP versions, not just PHP 8.3).

Only on PHP 8.3 `$b` should be `array{-4: 1, -3: 2}`, otherwise `array{-4: 1, 0: 2}`.

My suggestion: Fix the behaviour in ConstantArrayTypeBuilder regardless of PHP version. And then in `InitializerExprTypeResolver::getArrayType()` (DI is available there, hooray!) do the condition for PHP < 8.3 (so that the 2nd key is 0, not -3).

greew:
> I feel like this fix should be done differently.
> 
> Because PHPStan has a wrong assumption even on previous PHP versions: https://phpstan.org/r/492b583a-dc4b-4be2-a4c9-2232e039c19c (2nd key in `$a` should be `-3` on all PHP versions, not just PHP 8.3).
> 
> Only on PHP 8.3 `$b` should be `array{-4: 1, -3: 2}`, otherwise `array{-4: 1, 0: 2}`.
> 
> My suggestion: Fix the behaviour in ConstantArrayTypeBuilder regardless of PHP version. And then in `InitializerExprTypeResolver::getArrayType()` (DI is available there, hooray!) do the condition for PHP < 8.3 (so that the 2nd key is 0, not -3).

Great - thanks! And nice caught with the direct initialization difference in the old PHP versions!

Anyhow - I can't seem to figure out how to implement the change in the `InitializerExprTypeResolver::getArrayType()` method. It is still the `ConstantArrayTypeBuilder` that handles the calculation of which `nextAutoIndexes` to use, and I can't change this data from within `InitializerExprTypeResolver`.

If I'm not to change the visibility of the methods of `ConstantArrayTypeBuilder`, can someone give me a hint of how to do this?

ondrejmirtes:
In `InitializerExprTypeResolver::getArrayType()` you can call a new method on `ConstantArrayTypeBuilder` after creation that changes how it computes next indexes...

greew:
OK, I've added tests for both pre-8.0, 8.0+ and 8.3+ as requested in https://github.com/phpstan/phpstan/issues/10862#issuecomment-2156376121.

As requested earlier on, I've also moved the PHP version check to the `InitializerExprTypeResolver` class, but also had to add a PHP version check in `ConstantArrayType` to take care of the PHP >=8.0 < 8.3 findings.

greew:
Seems like the 7.3 and 7.4 errors test errors were introduced in https://github.com/phpstan/phpstan-src/pull/3128 ??

staabm:
> Seems like the 7.3 and 7.4 errors test errors were introduced in #3128 ??

fixed the build error with https://github.com/phpstan/phpstan-src/pull/3137

tests/PHPStan/Analyser/Bug10862Test.php
```diff
@@ -0,0 +1,36 @@
...
+	public function dataFileAsserts(): iterable
+	{
+		$path = 'bug-10862-php8.3';
+		if (PHP_VERSION_ID < 80300) {
+			$path = 'bug-10862-php8.0';
+		}
+		if (PHP_VERSION_ID < 80000) {
+			$path = 'bug-10862';
+		}
+		yield from self::gatherAssertTypes(__DIR__ . '/data/' . $path . '.php');
```
ondrejmirtes, line 20: I'd prefer if the conditions contained the whole `yield from` call instead of concatenating the path dynamically like this. It will make the path clickable in IDE.

src/Type/Constant/ConstantArrayTypeBuilder.php
```diff
@@ -181,6 +181,14 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $opt
 							$this->nextAutoIndexes[] = $newAutoIndex;
 						}
 					}
+					if (count($this->keyTypes) === 1 && $offsetType->getValue() < 0) {
```
ondrejmirtes, line 184: Changes to the builder certainly need new tests in ConstantArrayTypeBuilderTest.

src/Type/Constant/ConstantArrayTypeBuilder.php
```diff
@@ -314,4 +322,12 @@ public function isList(): bool
 		return $this->isList->yes();
 	}
 
+	public function resetNextAutoIndexToZeroIfNegative(): void
```
ondrejmirtes, line 325: I'd thought about this method that it should be set once after creation and not before every `setOffsetValueType` call. It should change the behaviour for all the calls, not just for the next one. Instead of this logic, this method should switch a `bool` property so that the computation of `nextAutoIndexes` changes based on that.

ondrejmirtes (changes requested):

VincentLanglet:
Hi @greew, are you still interested in finishing in PR (and doing it on 2.1.x now) ?

greew:
@VincentLanglet Uhmm... I'll see if I can get the time within the next weeks :)

I had pretty much forgotten about this PR :)

greew:
Ok, I guess that I have to face, that I haven't got the time to do it.

Sorry..

ondrejmirtes:
Please send a new PR if you decide to finish this. Thank you.

PR #3066, opened 2024-05-12
Add regression test

herndlm:
Closes https://github.com/phpstan/phpstan/issues/3300

See also https://phpstan.org/r/f16bc296-b26d-4081-a308-7dd0416def71

Not sure if more precision for constant arrays is needed? For such an edge case the current result might be enough? My PR https://github.com/phpstan/phpstan-src/pull/2117 adds some complexity, including changes on `Type::getKeysArray()` and rebasing those changes on latest 1.11.x is not trivial after https://github.com/phpstan/phpstan-src/pull/2516 and I'm worried they would make things even more complex..

The reason why this is fixed is the early exit in https://github.com/phpstan/phpstan-src/blob/d5a4746e91feacdceffe66d00d4976ecaa4a9673/src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php#L30-L32 when more than one argument is given which did not exist yet when the bug was reported.

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-3300.php
+++ b/tests/PHPStan/Analyser/data/bug-3300.php
@@ -1,0 +1,44 @@
+<?php declare(strict_types=1);
+
+namespace Bug3300;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * Class FormTypeHelper.
+ */
+class FormTypeHelper
+{
+	private const TYPE_TO_CLASS_MAP = [
+		'text' => 'TextType::class',
+		'group' => 'EntityManagerFormType::class',
+		'number' => 'IntegerType::class',
+		'select' => 'ChoiceType::class',
+		'radio' => 'ChoiceType::class',
+		'checkbox' => 'ChoiceType::class',
+		'bool' => 'CheckboxType::class',
+	];
+
+	/**
+	 * @param string $class
+	 *
+	 * @return string
+	 *
+	 * @throws \Exception
+	 */
+	public static function getTypeFromClass(string $class): string
+	{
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
How does this compare to your older PR? https://github.com/phpstan/phpstan-src/pull/2117

Is this issue really fixed?

herndlm:
> How does this compare to your older PR? #2117
> 
> Is this issue really fixed?

before, when the issue was reported `ArrayKeysFunctionDynamicReturnTypeExtension` was incorrectly specifying types also when `array_keys` was called with more than only one arg -> even if people wanted to filter for specific keys we were always returning all constant array keys, see https://github.com/phpstan/phpstan-src/blob/1c6ce5dfe4d16ad25ddef96a88541f7e44a3204f/src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php#L32

Then I did some refactoring which didn't change behaviour.

Then @staabm added in https://github.com/phpstan/phpstan-src/commit/89cfc493b786ede4eb85c476dcbca2773f05f500#diff-a408f91196cfe8e476bde977d3ceab1867fddee0b39a9bf1bb4e94ff7c1a2eb3 the check where the extension would only handle calls of `array_keys` with a single arg anymore.

And that basically fixed it, I'd say. Since then the extension didn't wrongfully use the constant array keys anymore, the type specification must be falling back to the default function definition, where the types are wider and that doesn't make problems.

Difference to my solution? I tried to add support to `array_keys` with more than 1 arg and properly use the array keys to have a more specific / narrower type. But that's a nice-to-have thing which adds most likely only unnecessary complexity at the moment, especially after https://github.com/phpstan/phpstan-src/pull/2516.

ondrejmirtes:
Thank you.

PR #3071, opened 2024-05-13
stream_get_meta_data return type

aleswita:
Return array key `uri` is not optional

https://www.php.net/manual/en/function.stream-get-meta-data.php
https://phpstan.org/r/21664a97-94a0-4ac8-b592-37d8a160bdd4


ondrejmirtes:
It is optional https://github.com/phpstan/phpstan/issues/10887

aleswita:
hm.. thanks!

PR #3072, opened 2024-05-13
Function containing inlineHtml has side-effects

staabm:
closes https://github.com/phpstan/phpstan/issues/10998

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1809,7 +1809,7 @@ class NodeScopeResolver
 			$hasYield = false;
 			$throwPoints = [];
 			$impurePoints = [
-				new ImpurePoint($scope, $stmt, 'inlineHtml', 'inlineHtml', true),
+				new ImpurePoint($scope, $stmt, 'betweenPhpTags', 'output between PHP opening and closing tags', true),
 			];
 		} elseif ($stmt instanceof Node\Stmt\Nop) {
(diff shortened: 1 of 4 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -1804,6 +1805,12 @@ static function (Node $node, Scope $scope) use ($nodeCallback): void {
 				$exprResult = $this->processExprNode($stmt, $stmt->expr, $scope, $nodeCallback, ExpressionContext::createDeep());
 				$impurePoints = $exprResult->getImpurePoints();
 			}
+		} elseif ($stmt instanceof InlineHTML) {
+			$hasYield = false;
+			$throwPoints = [];
+			$impurePoints = [
+				new ImpurePoint($scope, $stmt, 'inlineHtml', 'inlineHtml', true),
```
ondrejmirtes: Please test the error message when this is done inside `@phpstan-pure` function. It should be nicer than `inlineHtml`.

I'd even say that the identifier should be renamed to something like `betweenPhpTags` or something.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: I did not realize these strings are taken into the rule error message.

I still have to learn the new classes and their implications :)

ondrejmirtes:
Thank you.

PR #3075, opened 2024-05-13
Support integer ranges in sprintf() constant-result inference

staabm:


src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -88,11 +91,23 @@ public function getTypeFromFunctionCall(
...
 		foreach ($args as $arg) {
 			$argType = $scope->getType($arg->value);
-			if (count($argType->getConstantScalarValues()) === 0) {
+			$constantScalarValues = $argType->getConstantScalarValues();
+
+			if (count($constantScalarValues) === 0) {
+				if ($argType instanceof IntegerRangeType) {
+					foreach ($argType->getFiniteTypes() as $finiteType) {
+						if (!$finiteType instanceof ConstantIntegerType) {
+							throw new ShouldNotHappenException();
```
ondrejmirtes: Instead of this we should add PHPDoc above `IntegerRangeType::getFiniteTypes`

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3077, opened 2024-05-14
`fgetcsv` accepts `null` for `$length`

staabm:
fixes 

> Parameter #2 $length of function fgetcsv expects int<0, max>, null given.

on https://phpstan.org/r/856a86eb-7d7b-4265-9aa5-baf3c83d4321
for PHP 7.x see https://www.php.net/manual/en/function.fgetcsv.php


sreichel:
Can you please re-open?

ondrejmirtes:
Thank you.

PR #3078, opened 2024-05-15
More concrete return types for some database functions

devnix:
I'm not 100% sure about how hard can we trust on this, but I don't find a case where those arrays should ever be empty.

This suggestion is related to https://github.com/doctrine/dbal/pull/6287


devnix:
I have a couple of questions. Changing the return of `SQLite3Result::fetchArray()` to a conditional one depending on the `mode` parameter would be considered a breaking change and should be on bleeding edge? Should we propose all these changes to bleeding edge?

ondrejmirtes:
> Changing the return 

I don't think that's a BC break. It's simply improving the analysis to be more precise.

ondrejmirtes:
Thank you.

PR #3079, opened 2024-05-15
RichParser: stricter identifiers parser

janedbal:


tests/PHPStan/Parser/RichParserTest.php
```diff
@@ -267,47 +267,171 @@ public function dataLinesToIgnoreParseErrors(): iterable
...
+			'<?php' . PHP_EOL .
+			'test(); // @phpstan-ignore mečoun two' . PHP_EOL,
+			[
+				2 => ["Unexpected token type T_OTHER 'čoun' after T_IDENTIFIER, expected T_COMMA or T_END or T_OPEN_PARENTHESIS"],
+			],
+		];
+
+		yield [
+			'<?php' . PHP_EOL .
+			'test(); // @phpstan-ignore mečoun' . PHP_EOL,
```
mvorisek, line 380: The identifier grammar should be the same as any other PHP identifier grammar, ie. allow `\x80-\xff` to be valid identifier character - https://www.php.net/manual/en/language.variables.basics.php#language.variables.basics.

tests/PHPStan/Parser/RichParserTest.php
```diff
@@ -267,47 +267,171 @@ public function dataLinesToIgnoreParseErrors(): iterable
...
+			'<?php' . PHP_EOL .
+			'test(); // @phpstan-ignore return.ref, return.non čičí',
+			[
+				2 => ["Unexpected token type T_OTHER 'čičí' after T_IDENTIFIER, expected T_COMMA or T_END or T_OPEN_PARENTHESIS"],
+			],
+		];
+
+		yield [
+			'<?php' . PHP_EOL .
+			'test(); // @phpstan-ignore identifier -- comment' . PHP_EOL, // phpcs comment style
```
mvorisek, line 396: Just a question - if the identifiers have to be delimited by comma, what are the reasons to not allow free-text comment, ie. comment not wrapped in parenthesis?

ex: `@phpstan-ignore return.reg https://github.com/doctrine/dbal/pull/6407`

Such grammar is unambiguous as after `@phpstan-ignore` and `,\s*` one identifier will be still required.

I find this more natural and it will also URL issue we currently have - `@phpstan-ignore return.reg (https://github.com/doctrine/dbal/pull/6407)` - is opened as `https://github.com/doctrine/dbal/pull/6407)`.

tests/PHPStan/Parser/RichParserTest.php
ondrejmirtes: No, it's very simple ASCII for good reasons https://github.com/phpstan/phpstan-src/blob/16d5b010c61b5149b1b12970a70f3bde1dcc8358/src/Analyser/Error.php#L19

tests/PHPStan/Parser/RichParserTest.php
janedbal: With that, you could not use comma in comment, right?

`@phpstan-ignore identifier.one ignored, because of PHPStan bug, identifier.two`

Now you can (and you can even use parenthesis inside):

- `@phpstan-ignore identifier.one (ignored, because of PHPStan bug), identifier.two`
- `@phpstan-ignore identifier.one (ignored (because of PHPStan bug)), identifier.two`

janedbal: > @phpstan-ignore return.reg (https://github.com/doctrine/dbal/pull/6407) - is opened as https://github.com/doctrine/dbal/pull/6407).

Most widely used IDE (PHPStorm) does not do that (luckily for us).

mvorisek: 99% ignores are consisting of a single identifier. So I mean to parse anything after the current `\w+( \(...\))?(, \w+( \(...\))?)*` (simplified) grammar as a free-text comment.

ondrejmirtes:
Thank you!

PR #3080, opened 2024-05-16
Conditional return type for `SQLite3Result::fetchArray()`

devnix:
I'm aware that this change can be wrong or incomplete. I was wondering if there is a way in this project to do tests for types, for example https://github.com/azjezz/psl/blob/next/tests/static-analysis


ondrejmirtes:
You mean tests for types? Sure: https://phpstan.org/developing-extensions/testing#type-inference

Here most of them are in NodeScopeResolverTest.

VincentLanglet:
Hi @devnix, are you still interested by this PR ? If so, please now target 2.1.x instead. Thanks

resources/functionMap.php
```diff
@@ -11714,7 +11714,7 @@
 'SQLite3Result::__construct' => ['void'],
 'SQLite3Result::columnName' => ['string', 'column_number'=>'int'],
 'SQLite3Result::columnType' => ['int', 'column_number'=>'int'],
-'SQLite3Result::fetchArray' => ['array|false', 'mode='=>'int'],
+'SQLite3Result::fetchArray' => ['($mode is SQLITE3_NUM ? non-empty-list<mixed>|false : ($mode is SQLITE3_ASSOC ? non-empty-array<string,mixed>|false : false))', 'mode='=>'int'],
```
staabm: its the first "conditional return type"-style signature in this file.
I think we should add a nsrt test to make sure it actually works

VincentLanglet (commented):
Is a similar return type possible for `fetchAll` ?
https://github.com/JetBrains/phpstorm-stubs/blob/34a487bd374e42f4ce58606473d1b6882bfec2b7/sqlite3/sqlite3.php#L642
(will need to be in php85delta since it was introduced in php 8.5).

Also might be worth to add fetchArray has having sideEffect to avoid something like 
https://phpstan.org/r/7281fb29-1861-4297-8c50-4a708cbebf1c

devnix:
> Is a similar return type possible for `fetchAll` ? https://github.com/JetBrains/phpstorm-stubs/blob/34a487bd374e42f4ce58606473d1b6882bfec2b7/sqlite3/sqlite3.php#L642 (will need to be in php85delta since it was introduced in php 8.5).
> 
> Also might be worth to add fetchArray has having sideEffect to avoid something like https://phpstan.org/r/7281fb29-1861-4297-8c50-4a708cbebf1c

Great catch @VincentLanglet, done!

VincentLanglet (approved):

tests/PHPStan/Analyser/nsrt/sqlite3-result-fetch-all-php85.php
```diff
@@ -0,0 +1,31 @@
+<?php // php >= 8.5
```
VincentLanglet:
```suggestion
<?php // lint >= 8.5
```

VincentLanglet (changes requested):

tests/PHPStan/Analyser/nsrt/sqlite3-result-fetch-all-php85.php
devnix: Sorry, fixed 🤦🏼‍♂️

VincentLanglet (approved):

staabm (approved):

staabm:
thank you

PR #3081, opened 2024-05-17
Fix substr() type narrowing for maybe one-length result

staabm:
closes https://github.com/phpstan/phpstan/issues/11035
closes https://github.com/phpstan/phpstan/issues/9804

### DIFF
--- a/tests/PHPStan/Analyser/data/non-empty-string-substr.php
+++ b/tests/PHPStan/Analyser/data/non-empty-string-substr.php
@@ -54,7 +54,7 @@ class Foo
 		assertType('string', mb_substr($s, 0, $positiveInt));
 		assertType('non-empty-string', mb_substr($nonEmpty, 0, $positiveInt));
 
-		assertType('non-falsy-string', mb_substr("déjà_vu", 0, $positiveInt));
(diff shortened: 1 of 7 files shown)


tests/PHPStan/Analyser/data/non-empty-string-substr.php
```diff
@@ -54,7 +54,7 @@ public function doMbSubstr(string $s, $nonEmpty, $positiveInt, $postiveRange, $n
 		assertType('string', mb_substr($s, 0, $positiveInt));
 		assertType('non-empty-string', mb_substr($nonEmpty, 0, $positiveInt));
 
-		assertType('non-falsy-string', mb_substr("déjà_vu", 0, $positiveInt));
+		assertType('non-empty-string', mb_substr("déjà_vu", 0, $positiveInt));
```
VincentLanglet, line 57: We could keep `non-falsy-string` here since there is no `0` in the string constant @staabm

staabm: I also though about that. I think such a case is pretty unlikely and supporting it would mean a lot of complication of the extension.

therefore I decided against supporting that

ondrejmirtes:
Thank you.

PR #3082, opened 2024-05-17
make assigning property via reference impure

schlndh:
Fixes https://phpstan.org/r/cbe5c7f9-f5bd-4101-8cf6-37b7732946d4

I set `certain = false`, because theoretically the referenced variable may already be set, and the reference may never be written. In which case it would be pure. But I can't imagine why the reference would be used in such a way in the first place.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2146,8 +2146,8 @@ class NodeScopeResolver
 							$impurePoints[] = new ImpurePoint(
 								$scope,
 								$expr,
-								'propertyAssign',
-								'property assignment',
+								'propertyAssignByRef',
+								'property assignment by reference',
 								false,
 							);
 						}
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2135,7 +2135,23 @@ class NodeScopeResolver
 				$nodeCallback,
 				$context,
 				function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context): ExpressionResult {
+					$impurePoints = [];
 					if ($expr instanceof AssignRef) {
+						$referencedExpr = $expr->expr;
+						while ($referencedExpr instanceof ArrayDimFetch) {
+							$referencedExpr = $referencedExpr->var;
+						}
+
+						if ($referencedExpr instanceof PropertyFetch || $referencedExpr instanceof StaticPropertyFetch) {
+							$impurePoints[] = new ImpurePoint(
+								$scope,
+								$expr,
+								'propertyAssign',
+								'property assignment',
+								false,
+							);
+						}
+
 						$scope = $scope->enterExpressionAssign($expr->expr);
 					}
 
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Pure/data/impure-assign-ref.php
```diff
@@ -0,0 +1,61 @@
...
+		$value = &$other->value;
+		$value = 1;
+	}
+
+	private function bar6(): void
+	{
+		$value = &$this->objectArr[0]->foo;
+		$value = 1;
+	}
+
```
staabm, line 51: I thnk you are missing `NullsafePropertyFetch` variants

```suggestion
```
	private function bar7(?self $other): void
	{
		$value = &$other?->value;
		$value = 1;
	}
```
```

src/Analyser/NodeScopeResolver.php
```diff
@@ -2135,7 +2135,23 @@ public function processExprNode(Node\Stmt $stmt, Expr $expr, MutatingScope $scop
...
+						$referencedExpr = $expr->expr;
+						while ($referencedExpr instanceof ArrayDimFetch) {
+							$referencedExpr = $referencedExpr->var;
+						}
+
+						if ($referencedExpr instanceof PropertyFetch || $referencedExpr instanceof StaticPropertyFetch) {
+							$impurePoints[] = new ImpurePoint(
+								$scope,
+								$expr,
+								'propertyAssign',
```
ondrejmirtes: I think this should be a new identifier. Something like `propertyAssignByRef` or something.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
```diff
@@ -2135,7 +2135,23 @@ public function processExprNode(Node\Stmt $stmt, Expr $expr, MutatingScope $scop
...
+						while ($referencedExpr instanceof ArrayDimFetch) {
+							$referencedExpr = $referencedExpr->var;
+						}
+
+						if ($referencedExpr instanceof PropertyFetch || $referencedExpr instanceof StaticPropertyFetch) {
+							$impurePoints[] = new ImpurePoint(
+								$scope,
+								$expr,
+								'propertyAssignByRef',
+								'property assignment by reference',
```
ondrejmirtes, line 2150: This deserves a test :) In PureFunctionRule or PureMethodRule.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Meaning: test the scenario where this error message is shown.

src/Analyser/NodeScopeResolver.php
```diff
@@ -2135,7 +2135,23 @@ public function processExprNode(Node\Stmt $stmt, Expr $expr, MutatingScope $scop
...
+								false,
```
ondrejmirtes, line 2151: Should be `true`, right? There isn't a case where this might still be pure.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
schlndh: It could be pure:

> I set certain = false, because theoretically the referenced variable may already be set, and the reference may never be written. In which case it would be pure. But I can't imagine why the reference would be used in such a way in the first place.

```php
<?php

class Foo
{
    public ?int $foo;
    
    public function bar(): void
    {
        $x = &$this->foo;
    }
}

$foo = new Foo;
$foo->foo = 5;
$foo->bar();
var_dump($foo->foo);
```

ondrejmirtes: Alright, cool :)

ondrejmirtes:
Thank you!

PR #3083, opened 2024-05-19
Implement ParameterOutTypeExtensions

staabm:
refs https://github.com/phpstan/phpstan-src/pull/2589#issuecomment-2119139532


staabm:
I am not yet sure this feature is useful for the preg_match inference, because the type depends on context: https://github.com/phpstan/phpstan-src/pull/2589/files#diff-65e7da030813493bd5e549d34d89e84b206ed1f44cfc490ffaa52dcba136bd9bR128

staabm:
Thinking more about it: maybe we need a param out type which depends on the function call return type for the preg_match case

ondrejmirtes:
My question is: How important that is? Can it work without this condition in the first release? I have some ideas but it'd require some exploration (conditional types).

src/Type/DynamicFunctionParameterOutTypeExtension.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParameterReflection;
+
+interface DynamicFunctionParameterOutTypeExtension
```
ondrejmirtes: These interfaces need to be marked with `@api` and a description how to register them.

ondrejmirtes (changes requested):
Otherwise 👍

staabm:
> My question is: How important that is? Can it work without this condition in the first release? I have some ideas but it'd require some exploration (conditional types).

we will see when looking into it. I think this PR as is also makes sense in general.

I think conditional param-out types could work.

src/Analyser/NodeScopeResolver.php
```diff
@@ -4522,8 +4524,22 @@ private function processArgs(
 			}
 
 			if ($assignByReference) {
+				if ($currentParameter === null) {
+					throw new ShouldNotHappenException();
+				}
+
 				$argValue = $arg->value;
 				if ($argValue instanceof Variable && is_string($argValue->name)) {
```
staabm, line 4551: why does this code path only exec on variables and not e.g. on `PropertyFetch`?

src/Analyser/NodeScopeResolver.php
```diff
@@ -4522,8 +4524,22 @@ private function processArgs(
...
+
+					$paramOutTypes = $this->getParameterOutExtensionsTypes($callLike, $calleeReflection, $currentParameter, $scope);
+					if ($paramOutTypes !== []) {
+						if (!$byRefType instanceof MixedType) {
+							$paramOutTypes[] = $byRefType;
```
ondrejmirtes: I feel like extensions should override the byRefType, it should not be added in any case. Meaning all this logic should be in `else` below line 4521.

src/Analyser/NodeScopeResolver.php
```diff
@@ -4522,8 +4524,22 @@ private function processArgs(
...
 
 			if ($assignByReference) {
+				if ($currentParameter === null) {
+					throw new ShouldNotHappenException();
+				}
+
 				$argValue = $arg->value;
 				if ($argValue instanceof Variable && is_string($argValue->name)) {
+
+					$paramOutTypes = $this->getParameterOutExtensionsTypes($callLike, $calleeReflection, $currentParameter, $scope);
```
ondrejmirtes: Will this realistically ever return multiple types? I feel like the union should be done inside this method, and the method should return `?Type`.

src/Analyser/NodeScopeResolver.php
ondrejmirtes: You can explore that in another PR.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: I did it like that, because return-type extension work the same way. will adjust.

ondrejmirtes: It doesn't. it throws away the original return type when an extension returns a non-null value.

staabm: oh sorry, my reply above was meant for this comment: https://github.com/phpstan/phpstan-src/pull/3083#discussion_r1627254141 - in the end it doesn't matter though

src/Type/DynamicFunctionParameterOutTypeExtension.php
```diff
@@ -0,0 +1,32 @@
...
+ * services:
+ * 	-
+ *		class: App\PHPStan\MyExtension
+ *		tags:
+ *			- phpstan.dynamicFunctionReturnTypeExtension
+ * ```
+ *
+ * @api
+ */
+interface DynamicFunctionParameterOutTypeExtension
```
ondrejmirtes: We've recently started dropping the "dynamic" part of the name from an extension. See `FunctionParameterClosureTypeExtension` for example.

Since you're also applying it here inconsistently (you're not mentioning "dynamic" in the PR title, and also `LazyParameterOutTypeExtensionProvider` doesn't contain it), please drop it from all the names in this PR. Thanks.

ondrejmirtes (changes requested):

src/Type/DynamicFunctionParameterOutTypeExtension.php
staabm: Fixed

ondrejmirtes:
Thank you!

PR #3087, opened 2024-05-21
[junit formatter] Do not escape the error message twice

mousetail:
Currently, the `junit` formatter escapes the error message twice. This makes it hard to read.

The message is escaped once inside `JunitErrorFormatter::createTextCase` and one in `JunitErrorFormatter::formatErrors`. Only of of these is needed.

IMHO this should be the responsibility of `createTestCase` so I removed it from `formatErrors`

## Previous Output

```xml
<failure type="ERROR"
    message="Parameter #4 $options of method App\Auth\Provider::createRequest() expects array&amp;lt;string, mixed&amp;gt;, array&amp;lt;int, string&amp;gt; given."/>
```

## New Output

```xml
<failure type="ERROR"
    message="Parameter #4 $options of method App\Auth\Provider::createRequest() expects array&lt;string, mixed&gt;, array&lt;int, string&gt; given."/>
```

### DIFF
--- a/src/Command/ErrorFormatter/JunitErrorFormatter.php
+++ b/src/Command/ErrorFormatter/JunitErrorFormatter.php
@@ -37,7 +37,7 @@ class JunitErrorFormatter implements ErrorFormatter
 			$result .= $this->createTestCase(
 				sprintf('%s:%s', $fileName, (string) $fileSpecificError->getLine()),
 				'ERROR',
-				$this->escape($fileSpecificError->getMessage()),
+				$fileSpecificError->getMessage(),
 			);
 		}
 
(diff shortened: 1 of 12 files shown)


ondrejmirtes:
Please show this problem in a test that fails without the changes in src/.

mousetail:
@ondrejmirtes Tests added

mousetail:
Not exactly sure why generic errors in the table formatter are not rendering properly in PHP <8.1 but row specific errors are. The code between the two cases seems identical

https://github.com/mousetail/phpstan-src/blob/patch-1/src/Command/ErrorFormatter/TableErrorFormatter.php#L131

There is some odd printing HTML to console stuff going on? That might explain the missing generic if it's interpreted as a HTML tag. Doesn't explain why it works properly for the file specific error case or in PHP 8.1-8.3 though.

In either case it doesn't seem related to this issue, though it will need to be fixed before this can be merged.

mousetail:
Tests are fixed but not sure why some of the the integration tests are failing. The failures seem unrelated to the changes

mousetail:
@ondrejmirtes Anything more necessary to get this merged?

ondrejmirtes:
Thank you!

PR #3089, opened 2024-05-23
Add new *ParameterClosureTypeExtension

canvural:
Hello 👋🏽 

This PR adds new `ClosureTypeChangingExtension` extension that enables changing the closure type on the receiving side when necessary.

There are couple of issues I want solve before merging it though so I'm making a draft first:
- Naming 😄 Not sure any of the names I picked. Also I see `Dynamic` prefix in other extensions, not sure if we should use it here too.
- The extension finding loop stops as soon as it finds a matching extension. That made sense to me. But we can also gather every type and then union them at the end.
- I think I tested it enough, but not sure if there is any edge case that I should test also.


src/Type/FunctionClosureTypeChangingExtension.php
```diff
@@ -0,0 +1,16 @@
...
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+
+interface FunctionClosureTypeChangingExtension
+{
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool;
+
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type;
```
ondrejmirtes: I feel like this method signature is not sufficient because the parameter where the closure is passed to is also significat, right? So we also need to know `ParameterReflection` here. Also probably add it to the `*Supported` method too.

src/Analyser/NodeScopeResolver.php
```diff
@@ -3928,6 +3930,49 @@ private function processClosureNode(
 		?Type $passedToType,
 	): ProcessClosureResult
 	{
+		if ($stmt instanceof Node\Stmt\Expression && $stmt->expr instanceof Expr\CallLike) {
```
ondrejmirtes: This needs to work for nested expressions to so to eliminate everything that isn't `Node\Stmt\Expression` is not ideal.

ondrejmirtes: We also need to do the same thing for `processArrowFunctionNode`.

src/Analyser/NodeScopeResolver.php
```diff
@@ -3928,6 +3930,49 @@ private function processClosureNode(
...
+			if ($stmt->expr instanceof FuncCall && $stmt->expr->name instanceof Name && $this->reflectionProvider->hasFunction($stmt->expr->name, $scope)) {
```
ondrejmirtes: We could get rid of these `if` statements if this logic was implemented in `processArgs` (where `processClosureNode` is called) it could be simplified a lot. You wouldn't have to look for the right reflection again, because it's available there as `$calleeReflection`.

ondrejmirtes (changes requested):
Let's solve the names as the last thing, now is time to solve the fundamentals :)

src/Type/FunctionClosureTypeChangingExtension.php
canvural: > I feel like this method signature is not sufficient because the parameter where the closure is passed to is also significat, right?

hmm, not sure about that. We want to overwrite that parameter by supplying our own. And for my use case I did not need it. The test case I wrote here is basically the use case I needed.

src/Analyser/NodeScopeResolver.php
canvural: > This needs to work for nested expressions to so to eliminate everything that isn't Node\Stmt\Expression is not ideal.

How to do that?

src/Analyser/NodeScopeResolver.php
canvural: Sure. I can try that.

src/Type/FunctionClosureTypeChangingExtension.php
ondrejmirtes: Let's say you have a method call with three parameters. Each parameter is a callable. You want this extension to override each callable parameter with its own type.

Your current design only works for methods with a single callable parameter, not multiple. Right?

src/Analyser/NodeScopeResolver.php
ondrejmirtes: I think my other suggestion (to move to processArgs) will take care of this too.

src/Type/FunctionClosureTypeChangingExtension.php
canvural: ooh yeah, now I understood. We want to be able to select which parameter we want to modify, right? By checking the parameter name or something.

ondrejmirtes: Yes 😊

canvural: Done ea4494680a8aaee57c03530e69375660aea7999e

src/Analyser/NodeScopeResolver.php
canvural: ea4494680a8aaee57c03530e69375660aea7999e

Yeah it simplified a lot. But the test case with calling static method on a class string doesn't work. And maybe it doesn't need to work 🤷🏽

ondrejmirtes: Yeah, I'm not surprised by that. $calleeReflecrion is probably null for that case now.

canvural: Removed it for now.

Also I forgot to mention, there was no easy way to access the expression in `processArgs`, so had to refactor a bit so it accepts the expression (CallLike) now and extracts the args inside.

ondrejmirtes (changes requested):
This is fine by me 👍 

Let's think about the name:

1) I think we can omit "changing". All extensions "change" something.
2) I think the extension names should include "parameter".

So something like this:

* (Function|Method|StaticMethod)ParameterClosureTypeExtension (it communicates that it's about a function parameter and its type when you pass in closure)

Maybe we can also prefix the name with "Dynamic" but it might not be necessary.

Let me know your ideas :)

canvural:
I can't find anything better than `(Function|Method|StaticMethod)ParameterClosureTypeExtension` actually. `Dynamic` also feels unnecessary cause every extension is dynamic 😄  But we can put it for consistency if you want. So I'll make the change 👍🏽 

> The extension finding loop stops as soon as it finds a matching extension. That made sense to me. But we can also gather every type and then union them at the end.

Do you have any thoughts for this one also?

ondrejmirtes:
Yeah, "dynamic" is not necessary.

I'm fine with the current logic, if someone is going to need adjusting, we can change it in the future.

I'm going to merge it after the rename, thanks :)

ondrejmirtes:
I'm on my phone and remembered one more thing - make sure the extension interfaces are marked with `@api` and also have a description like other interfaces, with a way to register them in .neon. Thanks!

ondrejmirtes:
Thank you!

You can now require `phpstan/phpstan:^1.11.3` in Larastan. Right now it's going to download a dev version that already contains this PR. Try to create the extensions you have in mind, and if it works for you, I can release PHPStan 1.11.3 and you can release a new Larastan version :)

canvural:
Yeah I already tried it with downloading the PHAR file from the PR 😄  And tweeted the results. Everything looks good so far. I'll do more testing.

Thanks for the idea and your help! 😊

ondrejmirtes:
Alright 😊 Since there already has been a release today, I'm going to release 1.11.3 at some point next week (while in Berlin at IPC).

canvural:
Sure. No rush from my side.

PR #3092, opened 2024-05-24
add api annotation to AcceptsResult::__construct

canvural:
This PR:
- adds `@api` to the ` AcceptsResult::__construct` method
- also makes ` AcceptsResult::createNo` to accept reasons


ondrejmirtes:
Thank you!

PR #3093, opened 2024-05-24
Add new header functions for PHP 8.4

Seldaek:
See https://wiki.php.net/rfc/http-last-response-headers


ondrejmirtes:
Thank you!

PR #3094, opened 2024-05-25
DateTimeImmutable::modify() can no longer return false since PHP 8.3

JanTvrdik:
Since PHP 8.3, `DateTimeImmutable::modify()` can no longer return `false`, and throws `DateMalformedStringException` instead.

Relevant RFC: https://wiki.php.net/rfc/datetime-exceptions

Relevant php-src commit: https://github.com/php/php-src/commit/b7860cd56476d737fcf2e1fdda109b042312e3fa

Relevant code change:

obrazek

### DIFF
--- a/build/baseline-8.3.neon
+++ b/build/baseline-8.3.neon
@@ -1,0 +1,6 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Strict comparison using === between DateTime and false will always evaluate to false\\.$#"
+			count: 1
+			path: ../src/Type/Php/DateTimeModifyReturnTypeExtension.php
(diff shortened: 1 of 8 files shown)


ondrejmirtes:
Some tests failed, otherwise 👍 

Btw what about `@throws` for the method?

JanTvrdik:
> Btw what about @throws for the method?

It throws `DateMalformedStringException`. But you can't put `@throws` into function map, can you?

ondrejmirtes:
There would have to be conditionally loaded stub file for 8.4+.

janedbal:
Can we fix `@throws` in next step? I believe this already improves things.

ondrejmirtes:
I feel like fixing the return type and adding `@throws` is an atomic change. We know it should be done, let's do it while we have the context.

JanTvrdik:
Looks to be fixed as part of https://github.com/phpstan/phpstan-src/pull/3209.

ondrejmirtes:
I totally forgot about this PR, sorry! I was forced to fix this as part of update to jetbrains/phpstorm-stubs about exception-throwing on 8.3.

PR #3095, opened 2024-05-25
Support new PHP8.4 mb_lcfirst, mb_ucfirst function variants

staabm:
see 
- https://php.watch/versions/8.4/mb_ucfirst-mb_ucfirst
- https://github.com/php/php-src/pull/13161


ondrejmirtes:
Thank you.

PR #3096, opened 2024-05-26
Improve `explode` return type precision

staabm:
inspired by comments in https://github.com/phpstan/phpstan-src/commit/00691075acb1db2d1a6fbdbf607e0b52271815bb#r43325405

### DIFF
--- a/src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
@@ -6,7 +6,6 @@ use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Php\PhpVersion;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\Reflection\ParametersAcceptorSelector;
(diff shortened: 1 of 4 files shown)


src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
```diff
@@ -41,32 +40,34 @@ public function getTypeFromFunctionCall(
...
+		$returnType = AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new StringType()));
+		if (
+			!isset($args[2])
+			|| IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($scope->getType($args[2]->value))->yes()
+		) {
+			$returnType = TypeCombinator::intersect($returnType, new NonEmptyArrayType());
+		}
 
-			return $arrayType;
+		if ($this->phpVersion->getVersionId() <= 80000 && $isEmptyString->maybe()) {
```
ondrejmirtes: Do not use phpVersion like that, create a new method that describes the thing. Or use an existing method.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3102, opened 2024-05-27
Subtract zero-float when comparing against zero-int

staabm:
closes https://github.com/phpstan/phpstan/issues/5309


ondrejmirtes:
Thank you.

PR #3103, opened 2024-05-27
Fix wrong return type for SplFileObject::fgets()

DanielSiepmann:
The method does not return false, in contrast to function fgets(). The method will throw an exception instead.

Resolves: https://github.com/phpstan/phpstan/issues/11091


ondrejmirtes:
Thank you!

PR #3105, opened 2024-05-29
Fix wrong fread() length parameter type and return type

thg2k:
This PR addresses two issues with the `fread()` signature:

1) The `length` parameter requires `int<1, max>` at engine level instead of `int<0, max>`.

2) The return type could never be `false` until php 7.4.0 (unless bogus parameters are passed, but that does not count). Even in case of stream failure. Since php 7.4.0, it can indeed return `false` for example when reading from a valid stream, i.e. a failed socket.

About the return type, i'm not completely sure we should fix this. For forward compatibility, in case of language level < 7.4.x, it might be good that phpstan warns about the possibility of false return value.

Thoughts?


thg2k:
For the test failure, I believe I cannot use PHP_VERSION_ID inside the data files to determine the correct assertType, right? So instead of duplicating the whole data file from filesystem-function.php to filesystem-function74.php, how about I simply skip the test with PHP < 7.4?

ondrejmirtes:
Thank you.

PR #3106, opened 2024-05-29
CallToFunctionStatementWithoutSideEffectsRule checks the purity of `array_filter()`, `array_map()`, and `array_reduce()`

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/11101

The following functions can also cause side effects via callbacks, but since they are used infrequently and the processing is more complex, it is sufficient to follow the existing functionMetadata.

* [`array_udiff_assoc()`](https://www.php.net/array_udiff_assoc)
* [`array_udiff_uassoc()`](https://www.php.net/array_udiff_uassoc)
* [`array_uintersect()`](https://www.php.net/array_uintersect)
* [`array_uintersect_assoc()`](https://www.php.net/array_uintersect_assoc)
* [`array_uintersect_uassoc()`](https://www.php.net/array_uintersect_uassoc)

https://github.com/phpstan/phpstan-src/blob/66a2fd00ff85f489a1bc1b6e8798cb6b0ac209df/resources/functionMetadata.php#L721-L728


staabm:
Another case which should be tested: https://github.com/phpstan/phpstan/issues/11100

tests/PHPStan/Rules/Functions/data/function-call-statement-no-side-effects-8.0.php
```diff
@@ -19,6 +19,11 @@ public function noEffect(string $url, $resourceOrNull)
 		var_export([], return: true);
 		print_r([], return: true);
 		highlight_string($url, return: true);
+		array_filter([], callback: 'is_string');
```
staabm, line 22: We should test cases in which the callback is impure, e.g. because of calls to impure functions or access to global variables or similar from within the callable

ondrejmirtes:
It's not sufficient to handle this in a rule. It needs to be considered in NodeScopeResolver, so that the statements produce or not produce impurePoints.

ondrejmirtes:
This is a wrong direction. We need something like `@phpstan-pure-unless-callable-is-impure $callable`.

PR #3108, opened 2024-05-31
make method names non-falsy-string

schlndh:
Currently it's `string` (https://phpstan.org/r/377d039c-f54b-45c0-afa5-a8d0721c3f5a). I'm not aware of any way of creating methods with falsy names.

It's covered by reflection golden test (it doesn't currently work for properties due to a bug - https://github.com/phpstan/phpstan-src/pull/3109), so I think a separate test is unnecessary.


ondrejmirtes:
Thank you!

PR #3109, opened 2024-05-31
fix property descriptions in ReflectionProviderGoldenTest

schlndh:
I noticed that all properties in the reflection golden test look like this:

```
-----
PROPERTY ReflectionMethod::$name
-----
MISSING
```

It turns out to be due to a simple mismatch between the property names in the test input (start with `$`) and the names expected by the reflection provider (no `$`).


ondrejmirtes:
Thank you!

PR #3110, opened 2024-06-01
`array_fill_keys` expects array of array-key

staabm:
refs https://github.com/phpstan/phpstan/issues/11111

----

the also issue mentions

> We could also check other invalid accesses:
>
>    Reading access: var_dump($array[Language::GER]); --> Cannot access offset of type Language on array
>    Write: $array[Language::ITA] = 'foobar'; --> Cannot access offset of type Language on array
>    Unset: unset($array[Language::GER]); --> Cannot unset offset of type Language on array

which makes sense, but a separate problem for a separate PR IMO


resources/functionMap.php
```diff
@@ -261,7 +261,7 @@
 'array_change_key_case' => ['array', 'input'=>'array', 'case='=>'int'],
 'array_chunk' => ['list<array>', 'input'=>'array', 'size'=>'positive-int', 'preserve_keys='=>'bool'],
 'array_column' => ['array', 'array'=>'array', 'column_key'=>'mixed', 'index_key='=>'mixed'],
-'array_combine' => ['array|false', 'keys'=>'array', 'values'=>'array'],
+'array_combine' => ['array|false', 'keys'=>'array<array-key>', 'values'=>'array'],
```
staabm, line 264: while here I used the same fix here - I think `array-key` better conveys the idea of the type beeing required

this change is here just for consistency with `array_fill_keys`

ondrejmirtes:
It's not okay to change the functionMap for this requirement. This needs a custom rule.

Some types outside of int|string can be cast to array key.

resources/functionMap.php
```diff
@@ -271,7 +271,7 @@
 'array_diff_ukey' => ['array', 'arr1'=>'array', 'arr2'=>'array', 'key_comp_func'=>'callable(mixed,mixed):int'],
 'array_diff_ukey\'1' => ['array', 'arr1'=>'array', 'arr2'=>'array', 'arr3'=>'array', 'arg4'=>'array|callable(mixed,mixed):int', '...rest='=>'array|callable(mixed,mixed):int'],
 'array_fill' => ['array', 'start_key'=>'int', 'num'=>'int', 'val'=>'mixed'],
-'array_fill_keys' => ['array', 'keys'=>'array', 'val'=>'mixed'],
+'array_fill_keys' => ['array', 'keys'=>'array<array-key>', 'val'=>'mixed'],
```
staabm, line 274: we could also move this into bleeding-edge only, if BC is a concern

ondrejmirtes:
As for:

> We could also check other invalid accesses:

I believe PHPStan already does all of that but if it does not, open a feature request.

staabm:
> I believe PHPStan already does all of that but if it does not, open a feature request.

https://phpstan.org/r/a2c1cff4-0464-4021-bb32-04680554d9bf

you are right, it works already

staabm:
> It's not okay to change the functionMap for this requirement. This needs a custom rule.
> 
> Some types outside of int|string can be cast to array key.

would it make sense to create a new "virtual type", like we have with `__stringandstringable` ?

https://github.com/phpstan/phpstan-src/blob/04b840344a5f403e9a48120d7b4dccb278cab2d2/src/PhpDoc/TypeNodeResolver.php#L431-L432

ondrejmirtes:
I'd say creating a new custom rule is easier than modifying the type system.

PR #3112, opened 2024-06-02
Add error identifier to RawErrorFormatter

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/11125

### DIFF
--- a/src/Command/ErrorFormatter/RawErrorFormatter.php
+++ b/src/Command/ErrorFormatter/RawErrorFormatter.php
@@ -19,13 +19,20 @@ class RawErrorFormatter implements ErrorFormatter
 			$output->writeLineFormatted('');
 		}
 
+		$outputIdentifiers = $output->isVerbose();
 		foreach ($analysisResult->getFileSpecificErrors() as $fileSpecificError) {
+			$identifier = '';
+			if ($outputIdentifiers && $fileSpecificError->getIdentifier() !== null) {
+				$identifier = " [identifier:{$fileSpecificError->getIdentifier()}]";
+			}
+
 			$output->writeRaw(
 				sprintf(
-					'%s:%d:%s',
+					'%s:%d:%s%s',
 					$fileSpecificError->getFile(),
 					$fileSpecificError->getLine() ?? '?',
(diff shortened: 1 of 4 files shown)


src/Command/ErrorFormatter/RawErrorFormatter.php
```diff
@@ -19,13 +19,20 @@ public function formatErrors(
 			$output->writeLineFormatted('');
 		}
 
+		$outputIdentifiers = $output->isVerbose();
 		foreach ($analysisResult->getFileSpecificErrors() as $fileSpecificError) {
+			$identifier = '';
+			if ($outputIdentifiers && $fileSpecificError->getIdentifier() !== null) {
+				$identifier = sprintf(' [identifier:%s]', $fileSpecificError->getIdentifier());
```
ondrejmirtes: I'd rather not include the `:` here. If we omit that then we're just changing the format of the error message which is fine by me. But I' worried this `raw` error formatter is used by some parsers and this would be a BC break for them.

BTW I think this `raw` error formatter is really inferior - you always get absolute paths, and you don't get tips about errors.

ondrejmirtes (changes requested):

src/Command/ErrorFormatter/RawErrorFormatter.php
zonuexe: You're right, I fixed it!

src/Command/ErrorFormatter/RawErrorFormatter.php
```diff
@@ -19,13 +19,20 @@ public function formatErrors(
...
+			}
+
 			$output->writeRaw(
 				sprintf(
-					'%s:%d:%s',
+					'%s:%d:%s%s',
 					$fileSpecificError->getFile(),
 					$fileSpecificError->getLine() ?? '?',
 					$fileSpecificError->getMessage(),
+					$identifier,
```
ondrejmirtes, line 35: And some tests please!

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3117, opened 2024-06-03
Consider numeric-string types after string concat

staabm:
Improving concat types related to https://github.com/phpstan/phpstan/issues/11129

while implementing I realized that the actual [root-cause for 11129 is likely related to some loop related types](https://phpstan.org/r/5fe1a560-a9af-48dc-9ebd-324fcdc0014c), not the concat typing alone.

I currently have no plans to look into the loop related parts, therefore submitting this typing enhancements alone. this means this PR will __not__ close Bug 11129.

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -83,6 +83,7 @@ use function is_numeric;
 use function max;
 use function min;
 use function sprintf;
+use function str_contains;
 use function strtolower;
 use const INF;
 
@@ -485,7 +486,11 @@ class InitializerExprTypeResolver
 					continue;
 				}
 
-				if (!is_numeric($rightConstantString->getValue()) || $rightConstantString->getValue() < 0) {
+				if (
+					!is_numeric($rightConstantString->getValue())
+					|| str_contains($rightConstantString->getValue(), '.')
+					|| $rightConstantString->getValue() < 0
+				) {
 					$allRightConstantsZeroOrMore = false;
 					break;
 				}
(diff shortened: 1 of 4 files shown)


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -475,6 +477,29 @@ public function resolveConcatType(Type $left, Type $right): Type
...
+				if ($rightConstantString->getValue() === '') {
+					continue;
+				}
+
+				if (!is_numeric($rightConstantString->getValue()) || $rightConstantString->getValue() < 0) {
+					$allRightConstantsZeroOrMore = false;
+					break;
+				}
+
+				$allRightConstantsZeroOrMore = true;
```
ondrejmirtes, line 497: This is going to pass for floats too. `$numericString . '0.2'` will not be `numeric-string`. https://3v4l.org/BCS2K

ondrejmirtes (changes requested):

src/Reflection/InitializerExprTypeResolver.php
staabm: good catch, added more tests and a fix

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -475,6 +478,33 @@ public function resolveConcatType(Type $left, Type $right): Type
...
+		if ($leftNumericStringNonEmpty->isNumericString()->yes()) {
+			$allRightConstantsZeroOrMore = false;
+			foreach ($rightConstantStrings as $rightConstantString) {
+				if ($rightConstantString->getValue() === '') {
+					continue;
+				}
+
+				if (
+					!is_numeric($rightConstantString->getValue())
+					|| str_contains($rightConstantString->getValue(), '.')
```
ondrejmirtes: I'd rather only include int-like strings rather than exclude strings with `.`. Look at these fuckers: https://3v4l.org/Ia4r0

ondrejmirtes (changes requested):

src/Reflection/InitializerExprTypeResolver.php
staabm: another great catch. fixed - thanks.

ondrejmirtes:
Thank you!

PR #3118, opened 2024-06-03
fix gnupg_init function

pistej:
Fixed function: https://www.php.net/manual/en/function.gnupg-init.php


ondrejmirtes (changes requested):
Would be nice to provide a precise array shape for those options.

ondrejmirtes:
Thank you!

PR #3119, opened 2024-06-03
Implement StringReplaceSubjectParameterRule

staabm:
closes https://github.com/phpstan/phpstan/issues/11118

### DIFF
--- a/src/Rules/Functions/StringReplaceSubjectParameterRule.php
+++ b/src/Rules/Functions/StringReplaceSubjectParameterRule.php
@@ -1,0 +1,89 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Functions;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function array_key_exists;
+use function count;
+use function sprintf;
+
+/**
+ * @implements Rule<Node\Expr\FuncCall>
+ */
+class StringReplaceSubjectParameterRule implements Rule
+{
+
+	private const SUBJECT_PARAMETER_POSITION = [
+		'strtr' => 0,
+		'str_replace' => 2,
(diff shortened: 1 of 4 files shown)


src/Rules/Functions/StringReplaceSubjectParameterRule.php
```diff
@@ -0,0 +1,89 @@
...
+			$constantLiteralCount++;
+		}
+
+		if ($constantLiteralCount !== 2) {
+			return [];
+		}
+
+		return [
+			RuleErrorBuilder::message(sprintf(
+				'Argument #%d in unexpected position in call to function %s.',
```
staabm, line 80: I am not happy with this error message. had no idea yet for a better one.

tests/PHPStan/Rules/Functions/data/string-replace-subject.php
```diff
@@ -0,0 +1,72 @@
...
+	public function wrong(string $path): void
+	{
+		$x = strtr('\\', '/', $path);
+		$x = str_replace($path, '\\', '/');
+		$x = preg_replace($path, '\\', '/');
+		$x = str_replace($path, '\\', '/', $count);
+		$x = preg_replace($path, '\\', '/', $count);
+
+		$x = strtr('\\', $path, '/');
+		$x = str_replace('\\', $path, '/');
```
staabm, line 18: looks like shopware is doing such kind of construction:
- https://github.com/shopware/shopware/blob/1b0e9dc726a70788a1a99bc7c1bd5939ad54dd30/src/Core/Framework/DataAbstractionLayer/Command/CreateHydratorCommand.php#L389
- https://github.com/shopware/shopware/blob/1b0e9dc726a70788a1a99bc7c1bd5939ad54dd30/src/Core/Migration/Test/Migration1643892702AddCaptureRefundTablesTest.php#L79

I feel like shopware should instead better use `sprintf` or something, but it seems we whould not report an error for this situation then...

staabm: same in laravel/framework: 
- https://github.com/laravel/framework/blob/2b3e8d75f10b0ed17416282946355dc026bf326c/src/Illuminate/View/Compilers/BladeCompiler.php#L466

PR #3120, opened 2024-06-03
Fix pathinfo() return-type for union-type flags

staabm:
closes https://github.com/phpstan/phpstan/issues/10760


staabm:
resolved conflicts

ondrejmirtes:
Thank you!

PR #3121, opened 2024-06-03
array_slice() returns non-empty array for existing offsets and positive limit

staabm:
closes https://github.com/phpstan/phpstan/issues/10721

### DIFF
--- a/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
@@ -8,6 +8,7 @@ use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use function count;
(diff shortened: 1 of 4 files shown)


src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
```diff
@@ -51,6 +53,13 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 		}
 
 		if ($valueType->isIterableAtLeastOnce()->yes()) {
+			if ($offsetType !== null
+				&& $valueType->hasOffsetValueType($offsetType)->yes()
+				&& ($limitType === null || IntegerRangeType::fromInterval(1, null)->isSuperTypeOf($limitType)->yes())
+			) {
+				return $valueType;
+			}
```
staabm, line 61: the actual fix. 

all changes above are just refactoring.

tests/PHPStan/Analyser/data/bug-10721.php
```diff
@@ -0,0 +1,61 @@
...
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 2));
+
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, -1, 1));
+		assertType("non-empty-array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 0, 1));
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 1, 1));
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 2, 1));
+
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, -1, 2));
+		assertType("non-empty-array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 0, 2));
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 1, 2));
```
staabm, line 53: I wonder whether we would want to have this a `non-empty-array` ..?

tests/PHPStan/Analyser/data/bug-10721.php
```diff
@@ -0,0 +1,61 @@
...
+			'zib 2',
+		];
+		assertType("array{'zib', 'zib 2'}", $list);
+		shuffle($list);
+		assertType("non-empty-array<0|1, 'zib'|'zib 2'>&list", $list);
+
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, -1));
+		assertType("non-empty-array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 0));
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 1));
+		assertType("array<0|1, 'zib'|'zib 2'>&list", array_slice($list, 2));
```
staabm, line 44: Todo: misses optional offsets

tests/PHPStan/Analyser/data/bug-10721.php
```diff
@@ -0,0 +1,61 @@
...
+
+	public function variants(?int $limit = 20): array
+	{
+		$list = [
+			'zib',
+			'zib 2',
+		];
+		assertType("array{'zib', 'zib 2'}", $list);
+		shuffle($list);
+		assertType("non-empty-array<0|1, 'zib'|'zib 2'>&list", $list);
```
staabm, line 39: I think this PR can't work as is, because the type after `shuffle` is already not precise enough.

at this point we don't know anymore, that all keys will always exist.

staabm: implemented a more generic approach with https://github.com/phpstan/phpstan-src/pull/3124

PR #3124, opened 2024-06-04
array_slice() returns non-empty array for existing offsets and positive limit

staabm:
keep the non-empty-array accessory type when `array_slice` is invoked with a `0` offset and a positive limit.

initially I tried beeing smarter for non-zero offset cases, but this didn't work out because of a [too imprecise type after `shuffle`](https://github.com/phpstan/phpstan-src/pull/3121#discussion_r1625535272).

closes https://github.com/phpstan/phpstan/issues/10721

### DIFF
--- a/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
@@ -5,9 +5,11 @@ namespace PHPStan\Type\Php;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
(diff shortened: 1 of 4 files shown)


src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
```diff
@@ -51,7 +54,17 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-			return TypeCombinator::union($valueType, new ConstantArrayType([], []));
+			$optionalOffsetsType = TypeCombinator::union($valueType, new ConstantArrayType([], []));
+
+			$zero = new ConstantIntegerType(0);
+			if (
+				($offsetType === null || $zero->isSuperTypeOf($offsetType)->yes())
+				&& ($limitType === null || IntegerRangeType::fromInterval(1, null)->isSuperTypeOf($limitType)->yes())
+			) {
+				return TypeCombinator::intersect($optionalOffsetsType, new NonEmptyArrayType());
+			}
```
staabm, line 65: the actual fix. all above is moving code/refactoring

ondrejmirtes:
Thank you!

PR #3125, opened 2024-06-04
Consider a ClosureType() as maybe impure by default when no impurePoints array provided

jf-m:
Following the issue [#11135](https://github.com/phpstan/phpstan/issues/11135) 

- `$impurePoints` is no more promoted in the constructor
- Constructor's argument `$impurePoints` can be null, which will lead in a "Maybe" impure `ClosureType`.
- Making sure `TypeNodeResolver`'s return a Pure Closure when `pure-closure`. 
- Added a testcase & updating the testscases that used to create a pure-closure with `ClosureType()`

### DIFF
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -360,7 +360,7 @@ class TypeNodeResolver
 				return new CallableType(null, null, true, null, null, [], TrinaryLogic::createYes());
 
 			case 'pure-closure':
-				return new ClosureType(null, null, true, null, null, null, [], [], []);
+				return ClosureType::createPure();
 
 			case 'resource':
 				$type = $this->tryResolvePseudoTypeClassType($typeNode, $nameScope);
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -360,7 +360,7 @@ class TypeNodeResolver
 				return new CallableType(null, null, true, null, null, [], TrinaryLogic::createYes());
 
 			case 'pure-closure':
-				return new ClosureType();
+				return new ClosureType(impurePoints: []);
 
 			case 'resource':
 				$type = $this->tryResolvePseudoTypeClassType($typeNode, $nameScope);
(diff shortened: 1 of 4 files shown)


src/PhpDoc/TypeNodeResolver.php
```diff
@@ -360,7 +360,7 @@ private function resolveIdentifierTypeNode(IdentifierTypeNode $typeNode, NameSco
 				return new CallableType(null, null, true, null, null, [], TrinaryLogic::createYes());
 
 			case 'pure-closure':
-				return new ClosureType();
+				return new ClosureType(impurePoints: []);
```
ondrejmirtes: This isn't going to work. Named arguments are not downgraded for PHP 7.2+.

src/Type/ClosureType.php
```diff
@@ -75,12 +75,14 @@ class ClosureType implements TypeWithClassName, CallableParametersAcceptor
 
 	private TemplateTypeVarianceMap $callSiteVarianceMap;
 
+	private array $impurePoints;
```
ondrejmirtes, line 79: Specify what's in the array using PHPDoc.

ondrejmirtes (changes requested):

src/PhpDoc/TypeNodeResolver.php
```diff
@@ -360,7 +360,7 @@ private function resolveIdentifierTypeNode(IdentifierTypeNode $typeNode, NameSco
 				return new CallableType(null, null, true, null, null, [], TrinaryLogic::createYes());
 
 			case 'pure-closure':
-				return new ClosureType();
+				return new ClosureType(null, null, true, null, null, null, [], [], []);
```
ondrejmirtes: I just realized how we could make this nicer - create `public static function createPure(): self`.

ondrejmirtes (changes requested):

src/PhpDoc/TypeNodeResolver.php
jf-m: Good catch!

ondrejmirtes (changes requested):
Some PHPCS violations https://github.com/phpstan/phpstan-src/actions/runs/9368579006/job/25790974993?pr=3125

jf-m:
My bad, I forgot to run `make cs-fix`.

ondrejmirtes:
Thank you!

PR #3126, opened 2024-06-04
vprintf/vsprintf require a non-empty-array argument

staabm:
these functions throw a ValueError when called with an empty array, see
- https://3v4l.org/5Ro2C
- https://3v4l.org/E7Tev

### DIFF
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -303,5 +303,3 @@ services:
 
 	-
 		class: PHPStan\Rules\Functions\PrintfArrayParametersRule
-	-
-		class: PHPStan\Rules\Functions\PrintfHelper
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -26,6 +26,8 @@ conditionalTags:
 		phpstan.rules.rule: %featureToggles.magicConstantOutOfContext%
 	PHPStan\Rules\Functions\UselessFunctionReturnValueRule:
 		phpstan.rules.rule: %featureToggles.uselessReturnValue%
+	PHPStan\Rules\Functions\PrintfArrayParametersRule:
+		phpstan.rules.rule: %featureToggles.printfArrayParameters%
 
 rules:
 	- PHPStan\Rules\Api\ApiInstantiationRule
@@ -79,7 +81,6 @@ rules:
 	- PHPStan\Rules\Functions\InvalidLexicalVariablesInClosureUseRule
 	- PHPStan\Rules\Functions\ParamAttributesRule
 	- PHPStan\Rules\Functions\PrintfParametersRule
-	- PHPStan\Rules\Functions\PrintfArrayParametersRule
 	- PHPStan\Rules\Functions\RedefinedParametersRule
 	- PHPStan\Rules\Functions\ReturnNullsafeByRefRule
 	- PHPStan\Rules\Ignore\IgnoreParseErrorRule
@@ -300,5 +301,7 @@ services:
 	-
 		class: PHPStan\Rules\Functions\UselessFunctionReturnValueRule
 
+	-
+		class: PHPStan\Rules\Functions\PrintfArrayParametersRule
 	-
 		class: PHPStan\Rules\Functions\PrintfHelper
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -79,6 +79,7 @@ rules:
 	- PHPStan\Rules\Functions\InvalidLexicalVariablesInClosureUseRule
 	- PHPStan\Rules\Functions\ParamAttributesRule
 	- PHPStan\Rules\Functions\PrintfParametersRule
+	- PHPStan\Rules\Functions\PrintfArrayParametersRule
 	- PHPStan\Rules\Functions\RedefinedParametersRule
 	- PHPStan\Rules\Functions\ReturnNullsafeByRefRule
 	- PHPStan\Rules\Ignore\IgnoreParseErrorRule
@@ -298,3 +299,6 @@ services:
 
 	-
 		class: PHPStan\Rules\Functions\UselessFunctionReturnValueRule
+
+	-
+		class: PHPStan\Rules\Functions\PrintfHelper
--- a/src/Rules/Functions/PrintfArrayParametersRule.php
+++ b/src/Rules/Functions/PrintfArrayParametersRule.php
@@ -16,7 +16,6 @@ use function count;
 use function in_array;
 use function max;
 use function min;
-use function PHPStan\dumpType;
 use function sprintf;
 
 /**
--- a/src/Rules/Functions/PrintfArrayParametersRule.php
+++ b/src/Rules/Functions/PrintfArrayParametersRule.php
@@ -18,6 +18,14 @@ use function strtolower;
  */
 class PrintfArrayParametersRule implements Rule
 {
+	private const FORMAT_ARGUMENT_POSITIONS = [
+		'vprintf' => 0,
+		'vsprintf' => 0,
+	];
+	private const MINIMUM_NUMBER_OF_ARGUMENTS = [
+		'vprintf' => 1,
+		'vsprintf' => 1,
+	];
 
 	public function __construct(private PrintfHelper $printfHelper)
 	{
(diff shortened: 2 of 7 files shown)


staabm:
should `SprintfFunctionDynamicReturnTypeExtension` return a `ErrorType` in these cases?

ondrejmirtes:
Your example is misleading: https://3v4l.org/Mvl72

It'd make much more sense to create a rule similar to PrintfParametersRule.

src/Rules/Functions/PrintfHelper.php
```diff
@@ -0,0 +1,75 @@
...
+use Nette\Utils\Strings;
+use PHPStan\Php\PhpVersion;
+use function array_filter;
+use function count;
+use function max;
+use function sprintf;
+use function strlen;
+use const PREG_SET_ORDER;
+
+final class PrintfHelper
```
staabm: extracted from `PrintfParametersRule`

src/Rules/Functions/PrintfParametersRule.php
```diff
@@ -2,30 +2,36 @@
...
-use function max;
 use function sprintf;
-use function strlen;
 use function strtolower;
-use const PREG_SET_ORDER;
 
 /**
  * @implements Rule<Node\Expr\FuncCall>
  */
 class PrintfParametersRule implements Rule
```
staabm: no functional changes in this rule

conf/config.level0.neon
```diff
@@ -77,6 +77,7 @@ rules:
 	- PHPStan\Rules\Functions\InvalidLexicalVariablesInClosureUseRule
 	- PHPStan\Rules\Functions\ParamAttributesRule
 	- PHPStan\Rules\Functions\PrintfParametersRule
+	- PHPStan\Rules\Functions\PrintfArrayParametersRule
```
ondrejmirtes: bleeding edge only

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -0,0 +1,143 @@
...
+		return FuncCall::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (!($node->name instanceof Node\Name)) {
+			return [];
+		}
+
+		$name = strtolower((string) $node->name);
```
ondrejmirtes: You need to fetch FunctionReflection from ReflectionProvider and ask about its name.

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -0,0 +1,143 @@
...
+use function sprintf;
+use function strtolower;
+
+/**
+ * @implements Rule<Node\Expr\FuncCall>
+ */
+class PrintfArrayParametersRule implements Rule
+{
+
+	private const FORMAT_ARGUMENT_POSITIONS = [
```
ondrejmirtes: This doesn't need to be in constants if it's always the same.

ondrejmirtes (changes requested):
I don't see tests for general arrays

1) Zero placeholders + array (it's okay)
2) One or more placeholders + array (not okay)
3) One or more placeholders + non-empty-array (okay)

conf/config.level0.neon
```diff
@@ -298,3 +300,8 @@ services:
 
 	-
 		class: PHPStan\Rules\Functions\UselessFunctionReturnValueRule
+
+	-
+		class: PHPStan\Rules\Functions\PrintfArrayParametersRule
+	-
+		class: PHPStan\Rules\Functions\PrintfHelper
```
ondrejmirtes: Helpers are usually added in config.neon.

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -0,0 +1,184 @@
...
+		} else {
+			$formatArgsCount = TypeCombinator::union(...$formatArgsCounts);
+		}
+
+		if (!$this->placeholdersMatchesArgsCount($placeHoldersCount, $formatArgsCount)) {
+
+			if ($placeHoldersCount instanceof IntegerRangeType) {
+				$placeholders = $this->getIntegerRangeAsString($placeHoldersCount);
+				$singlePlaceholder = false;
+			} elseif ($placeHoldersCount instanceof ConstantIntegerType) {
```
ondrejmirtes: This can be just an `else`, right? PHPStan knows that.

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -0,0 +1,184 @@
...
+				$placeholders = $placeHoldersCount->getValue();
+				$singlePlaceholder = $placeholders === 1;
+			} else {
+				throw new ShouldNotHappenException();
+			}
+
+			if ($formatArgsCount instanceof IntegerRangeType) {
+				$values = $this->getIntegerRangeAsString($formatArgsCount);
+				$singleValue = false;
+			} elseif ($formatArgsCount instanceof ConstantIntegerType) {
```
ondrejmirtes: This can be just an `else`, right? PHPStan knows that.

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -0,0 +1,184 @@
...
+				))->identifier(sprintf('argument.%s', $name))->build(),
+			];
+		}
+
+		return [];
+	}
+
+	private function placeholdersMatchesArgsCount(Type $placeHoldersCount, Type $formatArgsCount): bool
+	{
+		if ($placeHoldersCount instanceof ConstantIntegerType && $formatArgsCount instanceof ConstantIntegerType) {
```
ondrejmirtes: This kind of structure makes it hard to see if all possible cases are handled.

1) Please add more specific parameter types, we know these can only be ConstantIntegerType/IntegerRangeType.
2) Do `if ($placeHoldersCount instanceof ConstantIntegerType) { ... }` and then handle all possible conditions inside. Then do `else` and handle all possible conditions again. Do not compound conditions in the method body on the top level with `&&`. It makes it hard to see for example whether `$a` with ConstantIntegerType and `$b` with IntegerRangeType is handled...

src/Rules/Functions/PrintfParametersRule.php
```diff
@@ -40,25 +50,17 @@ public function processNode(Node $node, Scope $scope): array
...
-		$minimumNumberOfArguments = [
-			'printf' => 1,
-			'sprintf' => 1,
-			'sscanf' => 3,
-			'fscanf' => 3,
-		];
-
-		$name = strtolower((string) $node->name);
-		if (!array_key_exists($name, $functionsArgumentPositions)) {
+		if (!$this->reflectionProvider->hasFunction($node->name, $scope)) {
```
ondrejmirtes: Why are you refactoring this so much? Either revert it, or send a PR to merge before this one :)

ondrejmirtes (changes requested):

src/Rules/Functions/PrintfArrayParametersRule.php
staabm: ~~I can't because `IntegerRangeType::fromInterval(..)` can return any Type `NeverType|ConstantIntegerType|IntegerType|IntegerRangeType`~~

I will assert the types instead

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			// Zero placeholders + array
+			if ($placeHoldersCount->getValue() === 0) {
+				return true;
+			}
+		}
+
+		if ($placeHoldersCount instanceof IntegerRangeType
+			&& $formatArgsCount instanceof IntegerRangeType
```
ondrejmirtes: What about handling `$placeHoldersCount` as IntegerRangeType + `$formarArgsCount` as ConstantIntegerType?

ondrejmirtes (changes requested):

src/Rules/Functions/PrintfArrayParametersRule.php
staabm: if I don't oversee anything, this IF statement here will cover the remaining cases.

I simplified it further.

ondrejmirtes:
Thank you.

PR #3127, opened 2024-06-05
Improve image related functions signature

thg2k:
- The parameters involving red, green, blue, and alpha are now required to be in the respective allowed range (0-255 for colors and 0-127 for alpha).
- Return values are specified in their possible ranges (0-max for color indexes, and 0-256 for imagecolorstotal()).
- imagecreate() and imagecreatetruecolor() now returns a benevolent union, it can only return false if width x height exceeds INT_MAX, and the size parameters are required to be in the range 1-max.
- imagesx() and imagesy() return value is in the range 1-max and can no longer return false.
- Return type of imagecolorsforindex() now contains the array signature.


thg2k:
@ondrejmirtes What's the feeling about this? I saw the CI integration errors but not sure how I should handle them. Clearly, restricting the parameters type will break people's SA but it is also true that if those restrictions are not met PHP >= 8.0 throws ValueError exceptions... so I figured that's the whole point of SA.

If you are not happy with this I can submit only the return type restrictions, that should cause no problem.

ondrejmirtes:
Return type changes: can be in functionMap.php
Narrowed parameter types: need to be in functionMap_bleedingEdge.php.

Thanks.

thg2k:
> Return type changes: can be in functionMap.php Narrowed parameter types: need to be in functionMap_bleedingEdge.php.
> 

Seems reasonable! I pushed the rebased change.

ondrejmirtes:
Thank you.

PR #3132, opened 2024-06-09
Feature check array functions which require stringish values

schlndh:
This is an attempt to detect values which are not castable to string in functions which expect an array of values castable to string (e.g. `array_unique([new stdClass(), new stdClass()])`).

I considered using `__stringandstringable` in functionMap, but then I saw this comment: https://github.com/phpstan/phpstan-src/pull/3110#issuecomment-2143473487 . Fortunately, I found that `ImplodeFunctionRule` already does the same thing, so I just generalized it to work with all the relevant functions.


tests/PHPStan/Rules/Functions/data/param-castable-to-string-functions.php
```diff
@@ -0,0 +1,63 @@
...
+	asort($arr1);
+	arsort($arr1);
+	natsort($arr1);
+	natcasesort($arr1);
+	array_count_values($arr1);
+	array_fill_keys($arr1, 5);
+
+
+	array_combine([true, null, 3.14], ['a', 'b', 'c']);
+	array_fill_keys([true, null, 3.14], 'a');
```
schlndh, line 80: Should this be allowed? IMO yes (it could be prohibited by more opinionated rulesets - e.g. strict-rules). However, currently it is a bit inconsistent. See https://phpstan.org/r/266e4f9e-a4c9-4679-ba6f-299684a10f7e

- `array_fill_keys` is allowed.
- `array_combine` is allowed with PHP < 8, and prohibited afterwards. That looks like an oversight, because as far as I can tell the function behaves the same: https://3v4l.org/dvTNC

schlndh:
I checked the integration test errors (https://github.com/phpstan/phpstan-src/actions/runs/9435637126/job/25989364652?pr=3132):

- rector-src: `AttributeGroup` (1st error) and `ComplexType` (2nd error) don't have `__toString()`.
- showare: `PermissionStruct` doesn't have `__toString()`.
- The rest seem obvious.

----

The `AnalyserIntegrationTest::testBug11147` also fails in 1.11.x, so that's unrelated.

ondrejmirtes (changes requested):
For backward compatibility, I'd like this approach:

1) Keep ImplodeFunctionRule
2) Deprecate ImplodeFunctionRule and disable it in bleeding edge (like NoopRule is)
3) Introduce new rule like you did, but enable it only in bleeding edge

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
```diff
@@ -0,0 +1,107 @@
...
+					'arsort',
+					'natcasesort',
+					'natsort',
+					'array_count_values',
+					'array_fill_keys',
+				],
+				true,
+			)
+		) {
+			$argsToCheck = [0 => $args[0]];
```
ondrejmirtes: What's the reasoning behind this condition?

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
```diff
@@ -0,0 +1,107 @@
...
+
+		$functionName = $this->reflectionProvider->resolveFunctionName($node->name, $scope);
+		$args = $node->getArgs();
+		$errorMessage = 'Parameter #%d of function %s expects an array of values castable to string, %s given.';
+		if (in_array($functionName, ['implode', 'join'], true)) {
+			$errorMessage = 'Parameter #%d $array of function %s expects array<string>, %s given.';
+			if (count($args) === 1) {
+				$argsToCheck = [0 => $args[0]];
+			} elseif (count($args) === 2) {
+				$argsToCheck = [1 => $args[1]];
```
ondrejmirtes: I know you probably just copied this from ImplodeFunctionRule, but the assumption is wrong because of named arguments. Someone might call `implode(array: $a, separator: $s)`. and this code would break.

Please write a test for this scenario. The scenario can be fixed by processing `$args` through `ArgumentsNormalizer`.

ondrejmirtes (changes requested):

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
schlndh: Good catch - I didn't think of that, since I just modifed the ImplodeFunctionRule. There was also a similar issue when no arguments were passed etc.

----

However, it lead me down another rabbit hole: when I added support for named arguments, I had to do something with the `Parameter #1 of function ...`, since the number might now be incorrect. So I attempted to make it consistent with how `FunctionCallParametersCheck` names the parameters. But it feels like individual rules shouldn't deal with that at all, since then it will "never" be consistent. If this is already handled somehow, maybe you can point me in the right direction.

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
schlndh: I'm not sure what you're asking about exactly. These are functions which either only have one argument, or which have multiple arguments and only the first one has to be array of values castable to string.

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
```diff
@@ -0,0 +1,170 @@
...
+				return [];
+			}
+		} elseif (in_array($functionName, $checkAllArgsFunctions, true)) {
+			$argsToCheck = $origArgs;
+		} elseif (in_array($functionName, $checkFirstArgFunctions, true)) {
+			$normalizedArgs = $getNormalizedArgs();
+			if ($normalizedArgs === []) {
+				return [];
+			}
+			$argsToCheck = [0 => $normalizedArgs[0]];
```
ondrejmirtes, line 110: What if the index does not exist?

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
```diff
@@ -0,0 +1,170 @@
...
+			return [];
+		}
+
+		$origNamedArgs = [];
+		foreach ($origArgs as $arg) {
+			if ($arg->unpack || $arg->name === null) {
+				continue;
+			}
+
+			$origNamedArgs[$arg->name->toString()] = $arg;
```
ondrejmirtes, line 121: You don't need to do this, on normalized arg they're available in an attribute.

ondrejmirtes (changes requested):

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
ondrejmirtes: Yeah, other rules don't deal with this at all now. We could refactor FunctionCallParametersCheck to be more flexible and to use its internals in other rules, but I don't think it's necessary in this PR.

Another thing this would allow us is to deal with `$arg–>unpack` properly same way FunctionCallParametersCheck does - it actually looks into `$type->getConstantArrays()` and expands them: https://github.com/phpstan/phpstan-src/blob/38f390636cb601b02071d3956c80d64f2f4b362b/src/Rules/FunctionCallParametersCheck.php#L114-L162

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		if (!$this->reflectionProvider->hasFunction($node->name, $scope)) {
+			return [];
+		}
+
+		$functionReflection = $this->reflectionProvider->getFunction($node->name, $scope);
+		$functionName = $functionReflection->getName();
+		$implodeFunctions = ['implode', 'join'];
+		$checkAllArgsFunctions = ['array_intersect', 'array_intersect_assoc', 'array_diff', 'array_diff_assoc'];
```
mvorisek, line 53: `array_intersect_key` should be supported as well for non-first arg.

ondrejmirtes: There's nothing cast to string when you call array_intersect_key.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
schlndh: I didn't notice that. But due to the usage of `$origNamedArgs` for handling implode (I probably added that after you wrote the comment), I still need to have a map of named args.

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
schlndh: I didn't think of that. But it seems that this cannot happen on non-empty `$normalizedArgs`, because the normalizer only fills optional parameter holes, if a required parameter is missing it seems to return null. I guess it doesn't hurt to be extra defensive and check it anyway.

PR #3136, opened 2024-06-10
Added regression test

staabm:
as noted by @thg2k in https://github.com/phpstan/phpstan-src/pull/2812#issuecomment-2156804423

since https://github.com/phpstan/phpstan-src/pull/2812 a expected error is detected, which wasn't reported before.

just adding a regression test.

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/count-array-shift.php
+++ b/tests/PHPStan/Rules/Functions/data/count-array-shift.php
@@ -2,8 +2,6 @@
 
 namespace CountArrayShift;
 
-use function PHPStan\Testing\assertType;
-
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Rules/Functions/data/count-array-shift.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace CountArrayShift;
+
+use function PHPStan\Testing\assertType;
```
ondrejmirtes: No need to use this function here.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3139, opened 2024-06-10
Update/fix ext-amqp signatures

ben-challis:
Resolves https://github.com/phpstan/phpstan/issues/11159.

I've updated the function map to have correct typings aligned with the official stubs - https://github.com/php-amqp/php-amqp/tree/latest/stubs

However, I am not sure this is actually the right approach, could these entries not be entirely removed and the typings be delegated to `jetbrains/phpstorm-stubs` completely? I may just not understand the purpose of this file! Let me know, thanks!

It's worth noting the declarations in the function map are not complete, there are other methods, which I presume are having their signatures pulled through from `jetbrains/phpstorm-stubs`?


ondrejmirtes:
Thank you!

ondrejmirtes:
`functionMap` has many reasons to exist, I've outlined them some time ago here: https://github.com/phpstan/phpstan/issues/9524#issuecomment-1848967050

PR #3140, opened 2024-06-10
Relax test version restriction

thg2k:
There is no reason to restrict the version to 8.0 for this test


staabm:
Added the restriction because of different error messages per php version. 

Should be good enough

thg2k:
> Added the restriction because of different error messages per php version.
> 

Yeah I realized from the CI output, but I'm giving it another try.. today I have no will to do real work 😆

ondrejmirtes:
Thank you.

PR #3141, opened 2024-06-10
Fix param castable to string rule

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/11167, but for some reason the issue is not reported as changed by the Issue Bot action. :thinking:

### DIFF
--- a/src/Rules/Functions/ParameterCastableToStringFunctionRule.php
+++ b/src/Rules/Functions/ParameterCastableToStringFunctionRule.php
@@ -20,6 +20,8 @@ use function array_key_exists;
 use function count;
 use function in_array;
 use function sprintf;
+use const SORT_NUMERIC;
+use const SORT_REGULAR;
 
 /**
  * @implements Rule<Node\Expr\FuncCall>
--- a/src/Rules/Functions/ParameterCastableToStringFunctionRule.php
+++ b/src/Rules/Functions/ParameterCastableToStringFunctionRule.php
@@ -112,13 +112,13 @@ class ParameterCastableToStringFunctionRule implements Rule
 			$argsToCheck = $origArgs;
 		} elseif (in_array($functionName, $checkFirstArgFunctions, true)) {
 			$normalizedArgs = $getNormalizedArgs();
-			if ($normalizedArgs === []) {
+			if (!array_key_exists(0, $normalizedArgs)) {
 				return [];
 			}
 			$argsToCheck = [0 => $normalizedArgs[0]];
 		} elseif (in_array($functionName, $checkSortWithFlagsFunctions, true)) {
 			$normalizedArgs = $getNormalizedArgs();
-			if ($normalizedArgs === []) {
+			if (!array_key_exists(0, $normalizedArgs)) {
 				return [];
 			}
 
--- a/src/Rules/Functions/ParameterCastableToStringFunctionRule.php
+++ b/src/Rules/Functions/ParameterCastableToStringFunctionRule.php
@@ -11,6 +11,7 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
+use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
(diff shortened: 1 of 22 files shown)


src/Rules/Functions/ParameterCastableToStringFunctionRule.php
```diff
@@ -104,10 +114,36 @@ public function processNode(Node $node, Scope $scope): array
...
+			} elseif (array_key_exists(1, $functionParameters)) {
+				$flags = $functionParameters[1]->getDefaultValue();
+			}
+
+			if ($flags === null || $flags->equals(new ConstantIntegerType(SORT_REGULAR))) {
+				return [];
+			}
+
+			foreach (TypeUtils::getConstantIntegers($flags) as $flag) {
+				if ($flag->getValue() !== SORT_NUMERIC) {
```
schlndh: I'm not sure what's the best way to handle flags like this. My thinking is as follows:

- `SORT_REGULAR` does not have any constraints, so I only check for equality. If there is a union (or int), then the stricter check should be maintained.
- `SORT_NUMERIC` seems the strictest (though not exactly, because it does allow arrays, unlike `SORT_STRING`), so if it's a union `SORT_NUMERIC` and something else, I go with `SORT_NUMERIC`.
- Otherwise, I assume `SORT_STRING` (or `SORT_LOCALE_STRING` which is the same for our purposes).

The only edge-case that I can think of right now is e.g. `array_unique([[1], [2]], rand() ? SORT_NUMERIC : SORT_STRING)`. IDK if it's worth it to complicate it further with such edge cases. :sweat_smile:

ondrejmirtes:
An idea: instead of these huge if-else, let's split this rule in multiple ones focused on each family of functions.

schlndh:
I'll try that on Wednesday, I need a bit of a break. :sweat_smile:

src/Rules/Functions/ParameterCastableToStringRule.php
```diff
@@ -67,7 +60,6 @@ public function processNode(Node $node, Scope $scope): array
 		if (
 			!in_array($functionName, $checkAllArgsFunctions, true)
 			&& !in_array($functionName, $checkFirstArgFunctions, true)
```
schlndh: I kept these two groups together, since the only difference is which parameters are checked.

src/Rules/Functions/SortParameterCastableToStringRule.php
```diff
@@ -0,0 +1,149 @@
...
+
+		$argsToCheck = [0 => $normalizedArgs[0]];
+		$flags = null;
+		if (array_key_exists(1, $normalizedArgs)) {
+			$flags = $scope->getType($normalizedArgs[1]->value);
+		} elseif (array_key_exists(1, $functionParameters)) {
+			$flags = $functionParameters[1]->getDefaultValue();
+		}
+
+		if ($flags === null || $flags->equals(new ConstantIntegerType(SORT_REGULAR))) {
```
staabm, line 90: should this check support union-types?

Similar to https://github.com/phpstan/phpstan-src/pull/3120/files#diff-4ce042f332b780306843d684a7f86f0dbec831736743ba6c2fe501d455ac7d3bR58

schlndh: All other valid flags are stricter, so IMO it's not necessary to support union types in this condition.

src/Rules/Functions/ParameterCastableToStringFunctionRule.php
schlndh: Upon further consideration I changed it. My thinking is that it should enforce the strongest constraint which it cannot prove to be unnecessary. So `sort($arr, $int)` and `sort($arr, rand() ? SORT_STRING : SORT_NUMERIC)` will require castability to both string and float. So the previously mentioned edge-case is now prohibited as well.

I found one more edge-case: statically known invalid flag (e.g. `sort($arr, 128)`). Based on my [testing](https://3v4l.org/9crBe) it seems to fall back to `SORT_REGULAR` in that case (I included `array_unique` because it has a different default value - I was thinking that it may fall back to the default value). I allowed this case (it should be prohibited, just not by this rule).

ondrejmirtes:
> for some reason the issue is not reported as changed by the Issue Bot action

Because the issue is not yet tagged as bug / feature request.

src/Rules/Functions/SortParameterCastableToStringRule.php
```diff
@@ -0,0 +1,172 @@
...
+			if ($arg->unpack || $arg->name === null) {
+				continue;
+			}
+
+			$origNamedArgs[$arg->name->toString()] = $arg;
+		}
+
+		$errors = [];
+
+		foreach ($argsToCheck as $argIdx => $arg) {
```
ondrejmirtes, line 126: The duplicated part of the rules should be moved to a common "Something...Check" class.

tests/PHPStan/Rules/Functions/ImplodeParameterCastableToStringRuleTest.php
```diff
@@ -0,0 +1,102 @@
...
+		return new ImplodeParameterCastableToStringRule($broker, new RuleLevelHelper($broker, true, false, true, false, false, true, false));
+	}
+
+	public function testNamedArguments(): void
+	{
+		if (PHP_VERSION_ID < 80000) {
+			$this->markTestSkipped('Test requires PHP 8.0.');
+		}
+
+		$this->analyse([__DIR__ . '/data/param-castable-to-string-functions-named-args.php'], [
```
ondrejmirtes: Sorry to be a PITA about this, but each rule should have separate test files. I'd hate to move some lines around and having to modify tests for 3 separate rules...

ondrejmirtes (changes requested):

ondrejmirtes:
Perfect, thank you!

PR #3142, opened 2024-06-10
Ensure exec group CI is executed in default order

thg2k:
Due to the recent addition of the random execution order in phpunit.xml, the test TraitsCachingIssueIntegrationTest::testCachingIssue was failing due to the expectation to be executed exactly in the data order.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ tests:
 	php vendor/bin/paratest --runner WrapperRunner --no-coverage
 
 tests-integration:
-	php vendor/bin/paratest --runner WrapperRunner --no-coverage --group exec
+	php vendor/bin/paratest --runner WrapperRunner --no-coverage --group exec --order-by default
 
 tests-levels:
 	php vendor/bin/paratest --runner WrapperRunner --no-coverage --group levels


thg2k:
This was broken in #2924 (commit a018f1ebe040d10af3fb71efb949149e10d117fc).

I don't know how else to fix the problem, I digged around phpunit but there seems no annotation for dataProviders to mark the order as default on a single case basis. The current solution seems to fix the problem tho.

ondrejmirtes:
I think we should remove TraitsCachingIssueIntegrationTest and recreate it in e2e-tests.yml. It gave me more headaches recently.

thg2k:
> I think we should remove TraitsCachingIssueIntegrationTest and recreate it in e2e-tests.yml. It gave me more headaches recently.

That I don't know about, but if you agree this patch fixes the current problem I'd suggest to apply it, otherwise this test is bound to fail based on the randomization order. At least you get a green mark :smile:

ondrejmirtes:
I'm sorry, I disagree with this patch. I'd like to pursue the direction I outlined above.

PR #3143, opened 2024-06-10
fix AMQPQueue::unbind routingKey type

ben-challis:
I made a mistake in #3139 and `$routingKey` of `AMQPQueue::unbind` was not correctly updated, it accepts null.

https://github.com/php-amqp/php-amqp/blob/a9cbd81ea401d4ba41b72381008b331f5516823c/stubs/AMQPQueue.php#L373-L384


ondrejmirtes:
Thank you!

PR #3144, opened 2024-06-10
Add parameter definitions to AMQPQueue::consume callback

ben-challis:
```
     * @param callable|null $callback A callback function to which the
     *                                consumed message will be passed. The
     *                                function must accept at a minimum
     *                                one parameter, an AMQPEnvelope object,
     *                                and an optional second parameter
     *                                the AMQPQueue object from which callback
     *                                was invoked. The AMQPQueue::consume() will
     *                                not return the processing thread back to
     *                                the PHP script until the callback
     *                                function returns FALSE.
     *                                If the callback is omitted or null is passed,
     *                                then the messages delivered to this client will
     *                                be made available to the first real callback
     *                                registered. That allows one to have a single
     *                                callback consuming from multiple queues.
```

I was not sure if this could just be defined as `callable(AMQPEnvelope, AMQPQueue=): mixed`, let me know if it can and I'll update.

Note this is an enrichment over `jetbrains/phpstorm-stubs` which only declares it as `callable`.

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -150,7 +150,7 @@ return [
 'AMQPQueue::ack' => ['void', 'deliveryTag'=>'int', 'flags='=>'int|null'],
 'AMQPQueue::bind' => ['void', 'exchangeName'=>'string', 'routingKey='=>'string|null', 'arguments='=>'array'],
 'AMQPQueue::cancel' => ['void', 'consumerTag='=>'string'],
-'AMQPQueue::consume' => ['void', 'callback='=>'callable|null', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
+'AMQPQueue::consume' => ['void', 'callback='=>'callable(AMQPEnvelope): mixed|callable(AMQPEnvelope, AMQPQueue): mixed|null', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
 'AMQPQueue::declareQueue' => ['int'],
 'AMQPQueue::delete' => ['int', 'flags='=>'int|null'],
 'AMQPQueue::get' => ['AMQPEnvelope|null', 'flags='=>'int|null'],
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -150,7 +150,7 @@ return [
 'AMQPQueue::ack' => ['void', 'deliveryTag'=>'int', 'flags='=>'int|null'],
 'AMQPQueue::bind' => ['void', 'exchangeName'=>'string', 'routingKey='=>'string|null', 'arguments='=>'array'],
 'AMQPQueue::cancel' => ['void', 'consumerTag='=>'string'],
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I think that `null|callable(AMQPEnvelope, AMQPQueue): mixed` would work fine. The implementation always calls the callback with two arguments, but the user doesn't have to specify them (if the implementation is not interested in them).

resources/functionMap.php
```diff
@@ -150,7 +150,7 @@
 'AMQPQueue::ack' => ['void', 'deliveryTag'=>'int', 'flags='=>'int|null'],
 'AMQPQueue::bind' => ['void', 'exchangeName'=>'string', 'routingKey='=>'string|null', 'arguments='=>'array'],
 'AMQPQueue::cancel' => ['void', 'consumerTag='=>'string'],
-'AMQPQueue::consume' => ['void', 'callback='=>'callable|null', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
+'AMQPQueue::consume' => ['void', 'callback='=>'callable(AMQPEnvelope): mixed|callable(AMQPEnvelope, AMQPQueue): mixed|null', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
```
ben-challis:
```suggestion
'AMQPQueue::consume' => ['void', 'callback='=>'callable(AMQPEnvelope, AMQPQueue): mixed|null', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
```

resources/functionMap.php
```diff
@@ -150,7 +150,7 @@
 'AMQPQueue::ack' => ['void', 'deliveryTag'=>'int', 'flags='=>'int|null'],
 'AMQPQueue::bind' => ['void', 'exchangeName'=>'string', 'routingKey='=>'string|null', 'arguments='=>'array'],
 'AMQPQueue::cancel' => ['void', 'consumerTag='=>'string'],
-'AMQPQueue::consume' => ['void', 'callback='=>'callable|null', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
+'AMQPQueue::consume' => ['void', 'callback='=>'callable(AMQPEnvelope, AMQPQueue): mixed|null', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
```
ondrejmirtes: This is kind of ambigous and I'm not sure how it's resolved. Please move `null` before the callable, or use parentheses to disambiguate.

ondrejmirtes (changes requested):

resources/functionMap.php
ben-challis:
```suggestion
'AMQPQueue::consume' => ['void', 'callback='=>'null|callable(AMQPEnvelope, AMQPQueue): mixed', 'flags='=>'int|null', 'consumerTag='=>'string|null'],
```

ben-challis: Fair enough, whoops! Near certain it would just be the return type 🤦

ondrejmirtes:
Thank you!

PR #3145, opened 2024-06-10
Implement bit shift operation on integers and unions

thg2k:
This was in my wishlist for a long time.. so I did it :smile:

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -77,6 +77,7 @@ use function count;
 use function dirname;
 use function floor;
 use function in_array;
+use function intval;
 use function is_finite;
 use function is_float;
 use function is_int;
@@ -1255,7 +1256,7 @@ class InitializerExprTypeResolver
 						return new ErrorType();
 					}
 
-					$resultType = $this->getTypeFromValue($leftNumberType->getValue() << $rightNumberType->getValue());
+					$resultType = $this->getTypeFromValue(intval($leftNumberType->getValue()) << intval($rightNumberType->getValue()));
 					if ($generalize) {
 						$resultType = $resultType->generalize(GeneralizePrecision::lessSpecific());
 					}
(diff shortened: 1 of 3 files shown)


thg2k:
@ondrejmirtes I played with this feature for a few days in my codebase and I feel it is ready to go. Can you review this? Is there anythink that I'm overlooking?

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1468,9 +1469,6 @@ private function callOperatorTypeSpecifyingExtensions(Expr\BinaryOp $expr, Type
 		return null;
 	}
 
-	/**
-	 * @param BinaryOp\Plus|BinaryOp\Minus|BinaryOp\Mul|BinaryOp\Div $expr
```
ondrejmirtes, line 1472: Please add ShiftLeft and ShiftRight here, instead of removing the PHPDoc.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1560,7 +1561,6 @@ private function resolveCommonMath(Expr\BinaryOp $expr, Type $leftType, Type $ri
 
 	/**
 	 * @param ConstantIntegerType|IntegerRangeType $range
-	 * @param BinaryOp\Div|BinaryOp\Minus|BinaryOp\Mul|BinaryOp\Plus $node
```
ondrejmirtes, line 1563: Please add ShiftLeft and ShiftRight here, instead of removing the PHPDoc.

ondrejmirtes (changes requested):

src/Reflection/InitializerExprTypeResolver.php
thg2k: Done for both. This should lead to a unreachable statement for the "else" branch, not sure if that is a problem.

thg2k:
Please note that this would conflict with #3152 if it gets merged. That should have the priority as I think it's a bad bug. I'm happy to rebase this (or the other) in case.

src/Reflection/InitializerExprTypeResolver.php
ondrejmirtes: > This should lead to

It doesn't in bleeding edge. PHPStan no longer penalizes for unreachable else branch that throws an exception.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #3146, opened 2024-06-10
Add int ranges to AMQP typings

ben-challis:
I haven't changed the setters as this is more of an annoyance to call site. It will trigger a runtime error if out of bounds. Happy to update setters as well if preferred!


resources/functionMap.php
```diff
@@ -72,10 +72,10 @@
 'AMQPChannel::__construct' => ['void', 'connection'=>'AMQPConnection'],
 'AMQPChannel::basicRecover' => ['void', 'requeue='=>'bool'],
 'AMQPChannel::commitTransaction' => ['void'],
-'AMQPChannel::getChannelId' => ['int'],
+'AMQPChannel::getChannelId' => ['int<1,65535>'],
```
ben-challis: This is defined as a 16bit unsigned int in the spec, but setting to 0 triggers an error always, so cannot be returned from the getter.

resources/functionMap.php
```diff
@@ -72,10 +72,10 @@
...
 'AMQPChannel::getConnection' => ['AMQPConnection'],
-'AMQPChannel::getPrefetchCount' => ['int'],
-'AMQPChannel::getPrefetchSize' => ['int'],
+'AMQPChannel::getPrefetchCount' => ['int<0,65535>'],
```
ben-challis: Similarly a 16bit unsigned int, zero signifies no prefetch.

resources/functionMap.php
```diff
@@ -72,10 +72,10 @@
...
+'AMQPChannel::getPrefetchSize' => ['int<0,max>'],
```
ben-challis: Octets, zero signifying maximum possible.

resources/functionMap.php
```diff
@@ -87,12 +87,12 @@
 'AMQPConnection::disconnect' => ['void'],
 'AMQPConnection::getHost' => ['string'],
 'AMQPConnection::getLogin' => ['string'],
-'AMQPConnection::getMaxChannels' => ['int'],
+'AMQPConnection::getMaxChannels' => ['int<1,65535>'],
```
ben-challis: Similar to above, a channel of 0 is invalid, despite being a 16bit unsigned int.

resources/functionMap.php
```diff
@@ -121,7 +121,7 @@
 'AMQPEnvelope::getHeader' => ['mixed', 'headerName'=>'string'],
 'AMQPEnvelope::getHeaders' => ['array<string,mixed>'],
 'AMQPEnvelope::getMessageId' => ['string|null'],
-'AMQPEnvelope::getPriority' => ['int'],
+'AMQPEnvelope::getPriority' => ['int<0,max>'],
```
ben-challis: AMQP spec defines that a priority MAY be 0-10, where 0 is "default". Above 10 is undefined. RabbitMQ supports up to 255, this value is capped to the maximum supported by the queue, hence leaving it leniant.

thg2k:
@ondrejmirtes Can we have some coding style directives on the function map signatures? I always use `int<0, max>` (with space) as I find it more readable and it matches the official describe of the type, but if you prefer it without space it's also fine. As long as we use the same style in the whole file.

Maybe we add a couple of lines on top of the file saying this? I also would like to see a specification on whether to use `int<1, max>` or `positive-int`. To be onest, I prefer the former as with the latter I always wonder if zero is included or not.

ondrejmirtes:
Thank you.

ondrejmirtes:
> Can we have some coding style directives

I don't care too much about formatting inside these types. Whatever works, use that :)

PR #3148, opened 2024-06-11
Introduce a common base class for all types

thg2k:
I don't have much hope this will be accepted but I'll give it a try 😆 

I wanted to add a `__toString` method to all types, which can be useful while doing quick and dirty debugging with print. But that's not the point.

I believe that having a common class can be useful when adding new methods to the Type interface which would have a default implementation instead of having to copy/paste the same method in 40 or so files.

I had to disable the related phpcs rule to call it AbstractType, but I can also call it in a different way.. AnyType? BaseType?

What do you think?


ondrejmirtes:
Sorry, I don't want types to have a __toString method, and I'm moving away from inheritance in the type system, not towards it.

You can use Type::describe() already.

PR #3149, opened 2024-06-11
Various fixes to the return signature of IntlDateFormatter methods

chosten:
PHP doc: https://www.php.net/manual/en/class.intldateformatter.php


ondrejmirtes:
Thank you!

PR #3150, opened 2024-06-11
Update phpcs config with proper exclude pattern

thg2k:
The pattern is actually matched as a regexp, so the trailing slash is necessary to avoid excluding unwanted files.

Also unified the patterns declared as 'arg' to the proper list.

Detailed list of changes:

1) Unified 'arg' ignores with the 'exclude-pattern' list, sorted and removed duplicates

2) Removed the following obsolete exclusions:
  - src/Reflection/SignatureMap/functionMap.php
  - src/Reflection/SignatureMap/functionMetadata.php

3) Ran phpcbf on the following files that were wrongly excluded:
  - tests/PHPStan/Analyser/TraitStubFilesTest.php
  - tests/PHPStan/Parser/CachedParserTest.php

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -6,7 +6,6 @@
 	<arg name="encoding" value="utf-8"/>
 	<arg name="tab-width" value="4"/>
 	<arg name="cache" value="tmp/cache/phpcs"/>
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Can you please fix the conflict? Thanks.

thg2k:
Sure, I rebased it and also updated the PR's description with more details on the actual changes.

ondrejmirtes:
Thank you!

PR #3151, opened 2024-06-11
Turn TraitsCachingIssueIntegrationTest into a e2e test

staabm:
quoting thg2k:

> Due to the recent addition of the random execution order in phpunit.xml, the test TraitsCachingIssueIntegrationTest::testCachingIssue was failing due to the expectation to be executed exactly in the data order.

turned into e2e test as requested in https://github.com/phpstan/phpstan-src/pull/3142#issuecomment-2158550915


staabm:
//cc @thg2k

thg2k:
Sorry, I am a bit ignorant regarding github CI... What does this PR do actually besides adding more complexity to the system? Now there is yet one more line in that obnoxiously long list of checks...

I mean, I understand it should be moved away from unit tests, as this more e2e since it invokes phpstan itself, but why does it have to be in its own check?

staabm:
e2e tests use separate jobs per design in the test-suite.

Ondrej mentioned he had other problems with the test case recently, but I don't know which.

The test itself is pretty slow and does not need to be run locally on every change when running `make` IMO, so that could be a reason to only run it in CI.

staabm:
I guess ondrej wants to get rid of phpunit for this test-cases but lets see

ondrejmirtes (changes requested):
You misunderstood. I don't want the e2e-tests.yml to run PHPUnit. I want it to do the same steps the test case currently does but in Bash.

thg2k:
> You misunderstood. I don't want the e2e-tests.yml to run PHPUnit. I want it to do the same steps the test case currently does but in Bash.

@ondrejmirtes  I have two concerns with this approach, I'd like to hear your wisdom on both:
1) They are entirely scripted inside the github workflow, so you can't really test it in command line before pushing. Ideally each one should be an indipendent bash script usable from the working copy, and the github workflow should just invoke the script file.
2) Each one creates a different "check", which concurs to this gigantic list to scroll. With the last commit they are 4 more even. Wouldn't it be better if they were all invoked one after the other in a single check? Potentially without stopping if one fails, like with the unit tests in phpunit. Is there some magic github action strategy to do this?

I think the different "checks" in github should more be like "PHP 7.x e2e" and "PHP 8.x e2e", rather than "test this" "test that"..

.github/workflows/e2e-tests.yml
```diff
@@ -184,6 +184,46 @@ jobs:
           - script: |
               cd e2e/env-int-key
               env 1=1 ../../bin/phpstan analyse test.php
+          - script: |
+              cd e2e/trait-caching
+              ../../bin/phpstan clear-result-cache
+              ../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/
+              ../../bin/phpstan clear-result-cache
```
thg2k: Doesn't this second clear-result-cache defeat the purpose of the test?

staabm: the phpunit tests worked the same way, see https://github.com/phpstan/phpstan-src/blob/e45ba45ab0edae1de1facf07543140f19a4e0c3f/tests/PHPStan/Analyser/traitsCachingIssue/TraitsCachingIssueIntegrationTest.php#L117-L140

and I think thats kind of the point of this PR. the top level bash scripts make the overall picture more obvious then the pretty lengthy phpunit test before

thg2k: You are right, it is the same behavior as before, but my comment still stands.. wasn't it broken from the beginning?

staabm: you might be right. I don't know the specifics of the bug which this test is covering.

if you are right, it would mean the test-as it was before this PR did not reproduce the actual bug.

---

edit: the test might have worked initially and got broken with a later re-work in https://github.com/phpstan/phpstan-src/commit/9088b84e1d59154ff8603e76dd6840e9e2db4dec#diff-18b49be48273139575cfe8bdb4075a2f4fe1ac1a26debeba8563b3e477ba916d

.github/workflows/e2e-tests.yml
```diff
@@ -184,6 +184,36 @@ jobs:
...
+              ../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/
+              ../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/
+          - script: |
+              cd e2e/trait-caching
+              ../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/
+              patch -b data/TraitOne.php < TraitOne.patch
+              OUTPUT=$(../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/ || true)
+              echo "$OUTPUT"
+              [ $(echo "$OUTPUT" | wc -l) -eq 1 ]
+              grep 'Method TraitsCachingIssue\\TestClassUsingTrait::doBar() should return stdClass but returns Exception.' <<< "$OUTPUT"
```
ondrejmirtes, line 198: This look like a nice opportunity to start using https://bashunit.typeddevs.com/, what do you think?

ondrejmirtes (changes requested):

ondrejmirtes:
`+56 −186`... yeah, this tells me this is much better than doing it in PHPUnit :)

.github/workflows/e2e-tests.yml
staabm: I am not sure its worth it. 

Would the goal be running the tests in a single job?

We would need to start clearing temp folder etc, so it will get more complicated.

Also not sure the assertions would be more readable.

Could you give me a few more directions what you would expect?

staabm: I will start a new PR to give bashunit a try. this one here should be ready to land.

ondrejmirtes:
Thanks. I want to use "assert_contains" instead of opaque "grep" call I don't understand 😊

PR #3152, opened 2024-06-12
Fix division type result with even integer range

thg2k:
Fixes phpstan/phpstan#11175

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -1760,31 +1760,25 @@ class InitializerExprTypeResolver
 				[$min, $max] = [$max, $min];
 			}
 
-			if ($operand instanceof IntegerRangeType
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

VincentLanglet:
Hi @thg2k, I dunno if a special case should be added for a division by `1` and `-1` ?

https://phpstan.org/r/cda3e12f-8030-4e53-9a96-cf35e889a8f2
https://phpstan.org/r/b1e9fdbc-371b-43e0-aa44-7461168cfc99

thg2k:
> Hi @thg2k, I dunno if a special case should be added for a division by `1` and `-1` ?
> 

We can add the special case of 1 and/or -1, but it seems a bit useless to me. Why would you be dividing for 1 or for a variable that you can prove statically is going to be 1? To flip the sign it's better to multiply for -1.

PR #3154, opened 2024-06-13
include path when counting errors ignored by identifier

pilif:
this fixes phpstan/phpstan#11177


ondrejmirtes:
Thank you very much!

PR #3155, opened 2024-06-13
Refactor NodeScopeResolver unit tests data files with annotations

thg2k:
I'd like to present a three steps plan to refactor the method `NodeScopeResolverTest::dataFileAsserts()` into a better maintainable version, where the data files are automatically included from the data directory without the need to enumerate them.

The steps of this plan are the following:
  1) Remove the inclusions of the data files themselves
  2) Implement a new annotation system specific for data files to declare the conditions
  3) Replace the explicit data files list with a scan of the data directory (in a future PR)

I believe this is a very needed improvement given the amount of test cases.


thg2k:
@ondrejmirtes My only doubt is why some of the data files were included in the instance before being analysed, I really couldn't figure it out, and removing the inclusion does not break tests in my environment. Could you tell me?

ondrejmirtes (changes requested):
Alright, let's try this.

1) Let's omit the `extension_loaded` condition, it might not be neede for the same reason the `require_once` are not needed.
2) Let's use the already existing `lint >=` syntax from PHP-Parallel-Lint for PHP versions. Feel free to copy and credit their code for that.
3) If there's some discrepancy between `PHP_VERSION_ID` condition in NodeScopeResolverTest, and the `lint >=` comment, you can fix that here.

thg2k:
> 1. Let's omit the `extension_loaded` condition, it might not be neede for the same reason the `require_once` are not needed.

With the better reflection phpstan should be able to analyse GD code without having the GD extension installed, right?

> 2. Let's use the already existing `lint >=` syntax from PHP-Parallel-Lint for PHP versions. Feel free to copy and credit their code for that.

I'll take a look at their code, but my mini-parser should do the job, I can easily change it to handle that syntax. My only concern is what would that syntax be?

    // onlyif >= 8.1
    // onlyif php >= 8.1
    // php >= 8.1

I was trying to be flexible in case other conditions should pop up in the future, and also to handle the `PHP_INT_SIZE === 8` condition which appears in a couple of files.

> 3. If there's some discrepancy between `PHP_VERSION_ID` condition in NodeScopeResolverTest, and the `lint >=` comment, you can fix that here.

I suppose the "lint requirement" and the "i-want-to-test-this-file" are rather different? The second should be about the phpVersionId in the config.neon, shouldn't it?

thg2k:
I got stuck in my quest. There are 379 files in the data directory that are **not** in the list of `NodeScopeResolverTest::dataFileAsserts`.

ondrejmirtes:
Don't rush this.

The steps of this plan are the following:

1. Remove the inclusions of the data files themselves
2. Implement a new annotation system specific for data files to declare the conditions
3. Replace the explicit data files list with a scan of the data directory

Step 3. should be a separate PR.

thg2k:
No rush 😄  I'm trying to get this thing fixed.

I admit I did not consider it so complex when I got into this...

Let's see where it goes, I'll give another shot at this tomorrow.

Thanks!

thg2k:
I think i'm getting there, now the tests pass but I cross checked the numbers with the CI run from the 1.11.x branch and there is some discrepancies. I must have messed up some annotations, but strangely they do not fail. I will investigate this further tomorrow.

Here are the numbers:

```
PHP 7.3 ubuntu  BASE  Tests: 18735, Assertions: 90503, Skipped: 444.
PHP 7.3 ubuntu  PR    Tests: 18741, Assertions: 90509, Skipped: 444.
                                +6

PHP 7.3 windows BASE  Tests: 18731, Assertions: 90499, Skipped: 444.
PHP 7.3 windows PR    Tests: 18737, Assertions: 90505, Skipped: 444.
                                +6

PHP 7.4 ubuntu  BASE  Tests: 19103, Assertions: 90943, Skipped: 372.
PHP 7.4 ubuntu  PR    Tests: 19109, Assertions: 90949, Skipped: 372.
                                +6

PHP 7.4 windows BASE  Tests: 19099, Assertions: 90939, Skipped: 372.
PHP 7.4 windows PR    Tests: 19105, Assertions: 90945, Skipped: 372.
                                +6

PHP 8.0 ubuntu  BASE  Tests: 19261, Assertions: 91268, Skipped: 256.
PHP 8.0 ubuntu  PR    Tests: 19261, Assertions: 91268, Skipped: 256.

PHP 8.0 windows BASE  ?
PHP 8.0 windows PR    Tests: 19257, Assertions: 91264, Skipped: 256.

PHP 8.1 ubuntu  BASE  Tests: 19887, Assertions: 92256, Skipped: 73.
PHP 8.1 ubuntu  PR    Tests: 19887, Assertions: 92256, Skipped: 73.

PHP 8.1 windows BASE  Tests: 19883, Assertions: 92252, Skipped: 73.
PHP 8.1 windows PR    Tests: 19883, Assertions: 92252, Skipped: 73.

PHP 8.2 ubuntu  BASE  Tests: 19925, Assertions: 92272, Skipped: 69.
PHP 8.2 ubuntu  PR    Tests: 19925, Assertions: 92272, Skipped: 69.

PHP 8.2 windows BASE  Tests: 19921, Assertions: 92268, Skipped: 69.
PHP 8.2 windows PR    Tests: 19921, Assertions: 92268, Skipped: 69.

PHP 8.3 ubuntu  BASE  Tests: 20024, Assertions: 92362, Skipped: 53.
PHP 8.3 ubuntu  PR    Tests: 20024, Assertions: 92362, Skipped: 53.

PHP 8.3 windows BASE  Tests: 20020, Assertions: 92358, Skipped: 53.
PHP 8.3 windows PR    Tests: 20020, Assertions: 92358, Skipped: 53.
```

src/Testing/TypeInferenceTestCase.php
```diff
@@ -126,12 +133,73 @@ public function assertFileAsserts(
 		}
 	}
 
+	public static function checkOnlyIfAnnotation(string $file): bool
```
staabm: might be helpful to move this logic into a dedicated class and do some unit testing on the parsing

src/Testing/TypeInferenceTestCase.php
```diff
@@ -126,12 +133,73 @@ public function assertFileAsserts(
...
+				foreach ((array) preg_split('/\s*&&\s*/', $ma[1][$i]) as $onlyIfCondition) {
+					$onlyIfConditions[] = (string) $onlyIfCondition;
+				}
+			}
+
+			foreach ($onlyIfConditions as $onlyIfCondition) {
+				if (preg_match('/^(PHP_VERSION_ID|PHP_INT_SIZE) (>|>=|<|<=|==) ([0-9]+)$/', $onlyIfCondition, $m) > 0) {
+					$onlyIfConditionConstant = constant($m[1]);
+					$onlyIfConditionOperator = $m[2];
+					$onlyIfConditionValue = intval($m[3]);
```
staabm, line 156: I think we should validate the type-range of the value. maybe you can re-use the typing PHPStan already knows for these 2 native constants

src/Testing/TypeInferenceTestCase.php
thg2k: It makes sense, even though testing the tests goes into dangerous territory 😆 
Just kidding.. I think it makes sense, I can move it to src/Testing and make a simple test case for it.

thg2k: On a second thought we are already in src/Testing I realized, so maybe we should make it a class only if we ever need it in another class, so we don't have a weird invocation of an unrelated class's method.. but I'm OK either way.

src/Testing/TypeInferenceTestCase.php
thg2k: Sure thing, I actually was always worried to miss a zero in 80000, so making sure it's at least 70000 and no more than 99999?

src/Testing/TypeInferenceTestCase.php
staabm: my comment was misleading. 

I was about extracting the parsing logic of this method, but leave the method here to be available for other users of TypeInferenceTestCase (its a public API class and used in phpstan extension testing outside this repo)

staabm: side note: should this method be private?

src/Testing/TypeInferenceTestCase.php
staabm: yeah something like that.
- `PHP_INT_SIZE` can only be 4 or 8
- 99999 >= `PHP_VERSION_ID` >= 70200 (PHPStan min php-version)

ondrejmirtes (changes requested):
I don't want to add any new syntax to files, we already have `lint >= PHP_VERSION_NUMBER` in these files.

Please simply use this logic to figure out what should be skipped or not: https://github.com/php-parallel-lint/PHP-Parallel-Lint/blob/0c2706086ac36dce31967cb36062ff8915fe03f7/bin/skip-linting.php#L21

ondrejmirtes:
Also please remove all the file moving from this PR. Once we have the fundamentals right and we test them on a few files, I'll merge this and possibly move the files myself. I plan to change the `gatherAssertTypes` implementation temporarily to simply `rename()` the files in question. (Of course not files in `tests/PHPStan/Analyser/data` are for NodeScopeResolverTest so we need to do something like that.)

Also we should figure out automatically where the skipping with `if PHP_VERSION_ID...` and `lint >= X` is not in sync, and fix that.

thg2k:
> I don't want to add any new syntax to files, we already have `lint >= PHP_VERSION_NUMBER` in these files.

That won't work with the range conditions, also why do you want to borrow the "lint" keyword for something different? We should have our own.. also that is quite useless cause most of the data folders are in the Makefile exclusions.

What about the PHP_INT_SIZE check? The syntax I made is extremely basic but mimics the PHP IF so it should be fairly familiar..



> Also please remove all the file moving from this PR.

There are no files moving in this PR, are you referring to commit of the series where I sort all the lines? Please take a look at the single commits, they are organized in a logical manner to simplify review.

> Also we should figure out automatically where the skipping with `if PHP_VERSION_ID...` and `lint >= X` is not in sync, and fix that.

If you scroll the diff you can easily spot those differences as they are all one liners. I would say more than half of the data files are not synced, but again, data files are currently **not** linted due to the Makefile exclusion, so there is that...

src/Testing/TypeInferenceTestCase.php
thg2k: I made it protected, otherwise I cannot test it?

src/Testing/TypeInferenceTestCase.php
thg2k: Added checks on the condition's value and unit tests

ondrejmirtes:
> That won't work with the range conditions

What file is tested based on a range with both lower and upper bound? We don't need to discover all files based on them being in a directory. If there's a complicated condition, we can leave it in, not move the file.

> also why do you want to borrow the "lint" keyword for something different?

I consider this duplicate information. I want to test files on same versions they are linted on.

I see this PR to change more than 1.5k LoC. That's not what I'd expect from this PR. This PR should introduce a new method in TypeInferenceTestCase to discover files in a directory and filter those files based on `lint >=` comments on the first line. And maybe test this method a on a few files.

thg2k:
> What file is tested based on a range with both lower and upper bound? We don't need to discover all files based on them being in a directory. If there's a complicated condition, we can leave it in, not move the file.

These are the data files that won't work with the current lint syntax:

    mb-strlen-php73.php:<?php declare(strict_types=1); // onlyif PHP_VERSION_ID >= 70300 && PHP_VERSION_ID < 80000
    mb-strlen-php8.php:<?php declare(strict_types=1); // onlyif PHP_VERSION_ID >= 80000 && PHP_VERSION_ID < 80200
    mb-strlen-php82.php:<?php declare(strict_types=1); // onlyif PHP_VERSION_ID >= 80200 && PHP_VERSION_ID < 80300
    enum-reflection-php81.php:<?php // onlyif PHP_VERSION_ID >= 80100 && PHP_VERSION_ID < 80200
    data-annotation-1.php:<?php declare(strict_types = 1); // onlyif PHP_VERSION_ID >= 80100 && PHP_VERSION_ID < 80300
    predefined-constants-64bit.php:<?php // onlyif PHP_INT_SIZE == 8
    predefined-constants-32bit.php:<?php // onlyif PHP_INT_SIZE == 4
    random-int.php:<?php // onlyif PHP_INT_SIZE == 8
    bug-5072.php:<?php // onlyif PHP_INT_SIZE == 8

I think I understand where you want to go, you want to use just the lint annotation to decide whether to test it or not. But that's not good design, and it's not intuitive. Imagine explaining that to a new contributor coming into the project. When I read "lint >= 8.1" I figure "well, that's about the syntax, not about the types".

I'm very confident that we need our own syntax. If you don't want to drop the "lint", we can combine them with something like: `<?php // lint >= 8.1 // onlyif PHP_VERSION_ID >= 80000` or `<?php // lint >= 8.1 // onlyif php >= 8.1 && php < 8.2` if you want it more readable than the version id.

But take the mb strlen example, they are all valid syntax under all php versions, so there shouldn't even be a lint annotation! But for our testing purposes, we have very strict version ranges requirements.

The solution you are suggesting, that is splitting the data files into a big directory "stuff that goes away with the lint annotation" and "stuff more complicated", seems like a big anti-pattern. There should be just one `data/NodeScopeResolver/` directory with all the files in it, and a unified annotation system to select them in all possible circumstances.

ondrejmirtes:
I'm sorry, but I don't want to argue about this, it's not worth my time. Feel free to close this PR if you disagree with the direction I want to take.

The current NodeScopeResolverTest design is fine and served us well for many years. I might change it one day but it's not urgent by any means.

thg2k:
> I'm sorry, but I don't want to argue about this, it's not worth my time. Feel free to close this PR if you disagree with the direction I want to take.
> 
> The current NodeScopeResolverTest design is fine and served us well for many years. I might change it one day but it's not urgent by any means.

Oh well, if it's not worth your time then definitely it is not worth mine either. I'll only stick to contributions of features that I actually need (like the shift math).

Anyway, since I invested a lot of time in this, I cleaned up the PR the way I see fit. All tests pass and the number of tests for each version match the previous ones. Feel free to pick it up at a later point if you want.

ondrejmirtes:
Since this topic already occupied my headspace, I decided to do it today. Here's the result: https://github.com/phpstan/phpstan-src/compare/effa1e1ffd60256d61884bcebda65dad3fa5fbd3...231a22780e53af7d711fd42d83537e168d6322b4

* `+317, -1515`
* Didn't change `gatherAssertTypes`, it'd be a BC break
* New method `gatherAssertTypesFromDirectory` that skips files based on `// lint` comment

You can see what files remained used by `gatherAssertTypes` in NodeScopeResolverTest: https://github.com/phpstan/phpstan-src/blob/231a22780e53af7d711fd42d83537e168d6322b4/tests/PHPStan/Analyser/NodeScopeResolverTest.php

* Those analysed by conditions that cannot be expressed by `// lint`
* Those who are not in `data/` but elsewhere

The latter could be moved into `tests/Analyser/nsrt/` while updating their paths in other tests, feel free to contribute that.

Thank you for the idea!

PR #3156, opened 2024-06-13
Improve the return type of array_map() for constant arrays

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/11180

### DIFF
--- a/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
@@ -23,7 +23,6 @@ use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeUtils;
 use function array_slice;
 use function count;
-use function preg_match;
 
 final class ArrayMapFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
 {
@@ -51,18 +50,6 @@ final class ArrayMapFunctionReturnTypeExtension implements DynamicFunctionReturn
 			}
 			$valueType = TypeCombinator::union(...$valueTypes);
 			$callback = $functionCall->getArgs()[0]->value;
-			if ($callback instanceof String_) {
-				/** @var non-falsy-string $callName */
-				$callName = $callback->value;
-				$callback = new Name($callName);
-			} elseif ($callback instanceof FuncCall && $callback->isFirstClassCallable() &&
-				$callback->getAttribute('phpstan_cache_printer') !== null &&
-				preg_match('/\A(?<name>\\\\?[^()]+)\(...\)\z/', $callback->getAttribute('phpstan_cache_printer'), $m) === 1
-			) {
-				/** @var non-falsy-string $callName */
-				$callName = $m['name'];
-				$callback = new Name($callName);
-			}
 		} elseif ($callableIsNull) {
 			$arrayBuilder = ConstantArrayTypeBuilder::createEmpty();
 			foreach (array_slice($functionCall->getArgs(), 1) as $index => $arg) {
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/data/array-map-callable81.php
```diff
@@ -0,0 +1,43 @@
...
+
+	public function doFoo(): void
+	{
+		$a = range(0, 1);
+
+		assertType("array{'0', '1'}", array_map('strval', $a));
+		assertType("array{'0', '1'}", array_map(strval(...), $a));
+		assertType("array{'0', '1'}", array_map(str(...), $a));
+		assertType("array{'0'|'1', '0'|'1'}", array_map(fn ($v) => strval($v), $a));
+		assertType("array{'0'|'1', '0'|'1'}", array_map(fn ($v) => (string)$v, $a));
```
zonuexe: The `ClosureType::$returnType` is bound to the union of the other elements of the array with the type of the callback. I'm looking for a way to get around that.

src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
```diff
@@ -37,13 +42,27 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
 			}
 			$valueType = TypeCombinator::union(...$valueTypes);
+			$callback = $functionCall->getArgs()[0]->value;
+			if ($callback instanceof String_) {
+				/** @var non-falsy-string $callName */
+				$callName = $callback->value;
+				$callback = new Name($callName);
+			} elseif ($callback instanceof FuncCall && $callback->isFirstClassCallable() &&
+				$callback->getAttribute('phpstan_cache_printer') !== null &&
+				preg_match('/\A(?<name>\\\\?[^()]+)\(...\)\z/', $callback->getAttribute('phpstan_cache_printer'), $m) === 1
```
zonuexe: To avoid binding the `ClosureType::$returnType` to a union with the results of other elements, I have replaced it with a direct call to the original function. I don't want to misuse the `phpstan_cache_printer` attribute, but there doesn't seem to be any other suitable source of the original function.

src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
```diff
@@ -37,13 +42,27 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+		$callback = null;
 
 		if ($callableType->isCallable()->yes()) {
 			$valueTypes = [new NeverType()];
 			foreach ($callableType->getCallableParametersAcceptors($scope) as $parametersAcceptor) {
 				$valueTypes[] = $parametersAcceptor->getReturnType();
 			}
 			$valueType = TypeCombinator::union(...$valueTypes);
+			$callback = $functionCall->getArgs()[0]->value;
+			if ($callback instanceof String_) {
```
ondrejmirtes: This entire code block isn't the right solution. You don't need to do any of that. You should be able to achieve the same thing with just:

```php
$callback = $functionCall->getArgs()[0]->value;
$callbackReturnType = $scope->getType(new FuncCall($callback, [
    new Arg(new TypeExpr($constantArray->getValueTypes()[$i])),
]));
```

If that doesn't work for some expressions, then the root cause is elsewhere.

ondrejmirtes (changes requested):

ondrejmirtes:
Please fix the build and rebase the branch to fix conflicts, or close the PR. Thank you.

zonuexe:
It looks better based on #4068. Thanks @VincentLanglet!

PR #258, opened 2020-06-24
Correct return value for PHP internal functions on failure

peter-gribanov:
Correct return value for functions:

* [`bzopen()`](https://www.php.net/manual/en/function.bzopen.php#refsect1-function.bzopen-returnvalues)
* [`dio_open()`](https://www.php.net/manual/en/function.dio-open.php#refsect1-function.dio-open-returnvalues)
* [`expect_popen()`](https://www.php.net/manual/en/function.expect-popen.php#refsect1-function.expect-popen-returnvalues)
* [`fam_open()`](https://www.php.net/manual/en/function.fam-open.php#refsect1-function.fam-open-returnvalues)
* [`fdf_open()`](https://www.php.net/manual/en/function.fdf-open.php#refsect1-function.fdf-open-returnvalues)
* [`ibase_blob_open()`](https://www.php.net/manual/en/function.ibase-blob-open.php#refsect1-function.ibase-blob-open-returnvalues)
* [`mcrypt_module_open()`](https://www.php.net/manual/en/function.mcrypt-module-open.php#refsect1-function.mcrypt-module-open-returnvalues)
* [`openal_device_open()`](https://www.php.net/manual/en/function.openal-device-open.php#refsect1-function.openal-device-open-returnvalues)
* [`pfsockopen()`](https://www.php.net/manual/en/function.pfsockopen)
* [`radius_acct_open()`](https://www.php.net/manual/en/function.radius-acct-open.php#refsect1-function.radius-acct-open-returnvalues)
* [`radius_auth_open()`](https://www.php.net/manual/en/function.radius-auth-open.php#refsect1-function.radius-auth-open-returnvalues)
* [`rpm_open()`](https://www.php.net/manual/en/function.rpm-open.php#refsect1-function.rpm-open-returnvalues)
* [`sqlite_open()`](https://www.php.net/manual/en/function.sqlite-open.php#refsect1-function.sqlite-open-returnvalues)
* [`sqlite_popen()`](https://www.php.net/manual/en/function.sqlite-popen.php#refsect1-function.sqlite-popen-returnvalues)

Return type declared as `resource`, but `false` returned on failure.

PR #3160, opened 2024-06-15
Use bashunit for e2e tests

staabm:


.github/workflows/e2e-tests.yml
```diff
@@ -255,3 +245,36 @@ jobs:
...
+    name: "Result cache E2E tests"
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - script: |
+              cd e2e/result-cache-1
+              bashunit .
```
staabm: do we still want a single isolated github action job per test?

e2e/result-cache-1/test.sh
```diff
@@ -0,0 +1,12 @@
...
+function test_result_cache1() {
+    echo -n > phpstan-baseline.neon
+    ../../bin/phpstan -vvv
+    patch -b src/Bar.php < patch-1.patch
+    cat baseline-1.neon > phpstan-baseline.neon
+    ../../bin/phpstan -vvv
+    mv src/Bar.php.orig src/Bar.php
+    echo -n > phpstan-baseline.neon
+    ../../bin/phpstan -vvv
+} 2>&1
```
staabm: I need to redirect stderr to stdout for all commands of the test because bashunit reports a test-failure as soon as something is written to ~~stdout~~ stderr

ondrejmirtes:
I'd rather install it with a script for a specific version than had it commited https://bashunit.typeddevs.com/installation

e2e/result-cache-1/test.sh
ondrejmirtes: I think I want to ask the authors what's the best way to do it here.

.github/workflows/e2e-tests.yml
ondrejmirtes: Yes, because I'm worried about polluting the state and making the tests worthless.

ondrejmirtes: It'd be great if we could just use the asserts outside of running with bashunit.

staabm: > It'd be great if we could just use the asserts outside of running with bashunit.

do you mean we should leave this script section in the e2e-test.yml file as before but just replacing `grep` with `assert_contains`? should I not add `test.js` like files in the test folders?

ondrejmirtes: Yes that's exactly what I wanted to do, but I don't know if bashunit supports that.

staabm: so something like

```diff
              cd e2e/trait-caching
              ../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/
              patch -b data/TraitOne.php < TraitOne.patch
              OUTPUT=$(../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/ || true)
              echo "$OUTPUT"
              [ $(echo "$OUTPUT" | wc -l) -eq 1 ]
-             grep 'Method TraitsCachingIssue\\TestClassUsingTrait::doBar() should return stdClass but returns Exception.' <<< "$OUTPUT"
+             bashunit assert_contains 'Method TraitsCachingIssue\\TestClassUsingTrait::doBar() should return stdClass but returns Exception.' "$OUTPUT"

```

..?

staabm: lets see https://github.com/TypedDevs/bashunit/issues/257

e2e/assert.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
staabm: this script loads `bashunit` and allows to call [the `assert*` bash-functions of `bashunit`](https://github.com/TypedDevs/bashunit/blob/bf86c013eb7294294237da29317ea907eb8811e5/src/assert.sh).

I made it return 0 on successfull assertion or print the `bashunit` error on failure

grafik

staabm: if you like the approach I could adjust remaining shell assertions in the other test cases

ondrejmirtes:
I have no idea what the assert.sh script does. I'm not a bash expert. I'd like to hear from Bashunit creators on this.

e2e/assert.sh
Chemaclass: This is a pretty cool approach. I was thinking about doing something similar. Maybe, actually, we could have this code somehow inside bashunit itself and allow people using this feature if they would want to as well (inspired by your idea here https://github.com/phpstan/phpstan-src/pull/3160#discussion_r1640951032).

I will experiment a couple of hours in the train today, and I will give you an update later today 🏗️ 

**UPDATE**: Merged https://github.com/TypedDevs/bashunit/pull/258

.github/workflows/e2e-tests.yml
Chemaclass: Would this be what you are looking for? https://github.com/TypedDevs/bashunit/pull/258

Chemaclass:
@ondrejmirtes @staabm you can use **Allow calling assertions outside tests** #258 by installing the beta for now (which uses main branch) -> https://bashunit.typeddevs.com/installation#install-sh and later on use the next official release (which would be out in a week or two). This way, you can start working on this and you aren't block from bashunit anymore :)

.github/workflows/e2e-tests.yml
```diff
@@ -194,26 +194,26 @@ jobs:
...
               cd e2e/trait-caching
               ../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/
               patch -b data/TraitOne.php < TraitOne.patch
               patch -b data/TraitTwo.php < TraitTwo.patch
               OUTPUT=$(../../bin/phpstan analyze --no-progress --level 8 --error-format raw data/ || true)
               echo "$OUTPUT"
-              [ $(echo "$OUTPUT" | wc -l) -eq 2 ]
-              grep 'Method TraitsCachingIssue\\TestClassUsingTrait::doBar() should return stdClass but returns Exception.' <<< "$OUTPUT"
-              grep 'Method class@anonymous/TestClassUsingTrait.php:20::doBar() should return stdClass but returns Exception.' <<< "$OUTPUT"
+              ../bashunit -a line_count 2 "$OUTPUT"
```
staabm, line 214: @Chemaclass thank you.

I have some weird case here.

```
➜  trait-caching git:(bashunit) ✗ echo "$OUTPUT"
/Users/staabm/workspace/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:15:Method TraitsCachingIssue\TestClassUsingTrait::doBar() should return stdClass but returns Exception.
/Users/staabm/workspace/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:29:Method class@anonymous/TestClassUsingTrait.php:20::doBar() should return stdClass but returns Exception.
➜  trait-caching git:(bashunit) ✗ echo "$OUTPUT" | wc -l | tr -d '[:blank:]'
2
➜  trait-caching git:(bashunit) ✗ ../bashunit -a line_count 2 "$OUTPUT"
✗ Failed: /Users/staabm/workspace/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:29:Method class@anonymous/TestClassUsingTrait.php:20::doBar() should return stdClass but returns Exception.
    Expected '/Users/staabm/workspace/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:15:Method TraitsCachingIssue\TestClassUsingTrait::doBar() should return stdClass but returns Exception.'
    to contain number of lines equal to '2'
```

do you have an idea what is wrong here?

Chemaclass: I would like to reproduce it locally, so I cloned your fork (`git@github.com:Chemaclass/staabm-phpstan-src.git`), checkout your `bashunit` branch, and I get this error when running composer install - which I think it shouldn't happen as it is a virgin start (I also set php 8.1), so I guess there is a problem on the deps? Did you face the same issue, @staabm ? Any idea how to find a work around to this?

Screenshot 2024-06-17 at 21 15 14

Chemaclass, line 214: I would like to reproduce it locally, so I cloned your fork (`git@github.com:Chemaclass/staabm-phpstan-src.git`), checkout your bashunit branch, and I get this error when running composer install - which I think it shouldn't happen as it is a virgin start (I also set php 8.1), so I guess there is a problem on the deps? Did you face the same issue, @staabm ? Any idea how to find a work around to this?  340425921-5b9fabbb-1df8-4442-bddc-29454deabaae

ondrejmirtes: See the README.

> If you encounter dependency problem, try using `export COMPOSER_ROOT_VERSION=1.11.x-dev`

Chemaclass: I think having the actual number would help to understand the issue, 
eg `to contain number of lines equal to '2' but found '1'`
because right now we know it's failing and it's not 2 but what is it? 🤔 I will try to add this now
One guess is that maybe it's only considering one line, but I am not sure, I need to debug and I would like to reproduce this bug in my local. Therefore I wrote the other comment below

staabm:
> by installing the beta for now (which uses main branch)

I tried using the beta on my local m1 pro mac, but it didn't work.
the installer reported the "beta" was installed, but the `--help` screen does not include the `-a` option?

```
➜  curl -s https://bashunit.typeddevs.com/install.sh | bash -s ./ beta  
> Downloading non-stable version: 'beta'
> bashunit has been installed in the './' folder

➜  bashunit --help
bashunit [arguments] [options]

Arguments:
  Specifies the directory or file containing the tests to run.
  If a directory is specified, it will execute the tests within files ending with test.sh.
  If you use wildcards, bashunit will run any tests it finds.

Options:
  -f|--filter <filter>
    Filters the tests to run based on the test name.

  -s|simple || -v|verbose
    Enables simplified or verbose output to the console.

  -S|--stop-on-failure
    Force to stop the runner right after encountering one failing test.

  -e|--env <file-path>
    Load a custom env file overriding the .env environment variables.

  --version
    Displays the current version of bashunit.

  --upgrade
    Upgrade to latest version of bashunit.

  --help
    This message.

See more: https://bashunit.typeddevs.com/command-line

➜  bashunit --version
bashunit - 0.12.0

➜  bashunit -a equals "a" "b"
bashunit - 0.12.0

Tests:      0 total
Assertions: 0 total

 No tests found 
➜  bashunit -a equals "a" "a"
bashunit - 0.12.0

Tests:      0 total
Assertions: 0 total

```

.github/workflows/e2e-tests.yml
staabm: I think I can reproduce on the CLI outside of phpstan repo with:

```
OUTPUT="one\ntwo\nthree"
➜  echo $OUTPUT
one
two
three

➜ ./bashunit -a line_count 3 "$OUTPUT"

✗ Failed: Main::exec assert
    Expected 'one\ntwo\nthree'
    to contain number of lines equal to '3'

```

staabm: agree, having the actual value in the message would be helpful.

Chemaclass: @staabm this could help to see the actual value https://github.com/TypedDevs/bashunit/pull/263 and enabling adding extra information to the failed assertions in case it might be needed (optional) - feel free to review/comment/approve it if you want (always! 😄) This can help us debugging the current issue after merging it and you refreshing your beta

Chemaclass: Ok, in such a case the bug seems to be in the `assert_line_count` where it might consider that as 1 line and not as 3. I think that's the direction of the 🐛

staabm: hmm .. the unit test regarding it looked good:
https://github.com/TypedDevs/bashunit/pull/259/files#diff-8c5ca8e706f9d9fb330cda7953d707bb4180b68f55fdff87ec30ad3f8e179ad6R339-R341

I don't know bash good enough, but I had a feeling the problem is related to the assertion via args passing

Chemaclass: The issue is clearly here `actual_line_count=$(echo "$actual" | wc -l | tr -d '[:blank:]')` - still, I don't understand the issue - I mean, I don't know how to solve it yet 🤔 
Here you can see the result of the failing test:
```bash
function test_successful_assert_line_count_multiline_string_in_one_line() {
  assert_empty "$(assert_line_count 4 "one\ntwo\nthree\nfour")"
}
```

Chemaclass: I got a potential fix for this ... creating PR

Chemaclass: I got a potential fix for this: https://github.com/TypedDevs/bashunit/pull/264 (merged)
Can you try again to reproduce it with the latest beta, @staabm ?

staabm: thanks. really appreciate your support

my initial reported case works as expected now. the build here still fails though.

I reduced it to:

```
➜  OUTPUT="/home/runner/work/phpstan-src/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:15:Method TraitsCachingIssue\TestClassUsingTrait::doBar() should return stdClass but returns Exception.
/home/runner/work/phpstan-src/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:29:Method class@anonymous/TestClassUsingTrait.php:20::doBar() should return stdClass but returns Exception."
➜  echo $OUTPUT
/home/runner/work/phpstan-src/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:15:Method TraitsCachingIssue\TestClassUsingTrait::doBar() should return stdClass but returns Exception.
/home/runner/work/phpstan-src/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:29:Method class@anonymous/TestClassUsingTrait.php:20::doBar() should return stdClass but returns Exception.
➜  ./bashunit -a line_count 2 "$OUTPUT"
✗ Failed: /home/runner/work/phpstan-src/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:29:Method class@anonymous/TestClassUsingTrait.php:20::doBar() should return stdClass but returns Exception.
    Expected '/home/runner/work/phpstan-src/phpstan-src/e2e/trait-caching/data/TestClassUsingTrait.php:15:Method TraitsCachingIssue\TestClassUsingTrait::doBar() should return stdClass but returns Exception.'
    to contain number of lines equal to '2'
    but found '1'
```

Chemaclass: I can reproduce it, yes. This is a bashunit issue, I will work on it today 🏗️ 
UPDATE: It took me a couple of hours... but finally I got something out: https://github.com/TypedDevs/bashunit/pull/265

Chemaclass: Alright, https://github.com/TypedDevs/bashunit/pull/265 is merged and we are getting closer to the light!
Can you try again to reproduce it with the latest beta, @staabm ?

Chemaclass: With the latest [error](https://github.com/phpstan/phpstan-src/actions/runs/9567238902/job/26374488151), I think we might have found a major problem wrt multiline strings mmmm I have an idea, but it might take a while... let me see if I can create a draft for this today 🏗️ 
> UPDATE: This might be the fix for the multiline string bug: https://github.com/TypedDevs/bashunit/pull/266

Chemaclass: Same drill: https://github.com/TypedDevs/bashunit/pull/266 is merged, can you try the latest beta again, @staabm? 🤞🏼

staabm: @Chemaclass all e2e are green now. thank you very much for your efforts <3.

staabm:
@ondrejmirtes this one is good to go now :)

Chemaclass (approved):
Awesome 🙌  For now, you can use the beta which contains the new working functionality that you need, and in the coming days, once we launch the new release, then you can pin it to the latest stable instead of beta. 
Anything you need, here we are!

ondrejmirtes:
Awesome!

@staabm Please send an update once a stable release comes out.

ondrejmirtes:
Thank you!

PR #3165, opened 2024-06-18
Fix type of return value for `WeakMap::offsetGet()`

sonnymilton:
Fixes https://github.com/phpstan/phpstan/issues/11194


tests/PHPStan/Analyser/nsrt/weakMap.php
```diff
@@ -0,0 +1,30 @@
+<?php declare(strict_types = 1);
```
ondrejmirtes: First line has to be `<?php // lint >= 8.0`

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3166, opened 2024-06-19
Indicate file on TypeInferenceTestCase validation errors

staabm:
was just hunting down a bug in one of my PRs, but wasn't able to find the offending file, because the error message did not indicate which file the error related to

before this PR:
```
There was 1 error:

1) Error
PHPUnit\Framework\AssertionFailedError: Expected type must be a literal string, string given on line 78.
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:166
/Users/staabm/workspace/phpstan-src/src/Node/ClassStatementsGatherer.php:128
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:650
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:759
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:3938
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:2124
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:758
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:339
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:649
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:339
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:835
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:339
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:803
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:296
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:92
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:145
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:271
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:17
```

after this PR:
```
1) Error
PHPUnit\Framework\AssertionFailedError: Expected type must be a literal string, string given in tests/PHPStan/Analyser/nsrt/param-closure-this.php on line 78.
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:166
/Users/staabm/workspace/phpstan-src/src/Node/ClassStatementsGatherer.php:128
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:650
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:759
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:3938
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:2124
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:758
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:339
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:649
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:339
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:835
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:339
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:803
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:296
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:92
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:145
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:273
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:17
```

note the added filename in the error message

### DIFF
--- a/src/Testing/TypeInferenceTestCase.php
+++ b/src/Testing/TypeInferenceTestCase.php
@@ -12,7 +12,7 @@ use PHPStan\DependencyInjection\Type\DynamicThrowTypeExtensionProvider;
 use PHPStan\DependencyInjection\Type\ParameterClosureTypeExtensionProvider;
 use PHPStan\DependencyInjection\Type\ParameterOutTypeExtensionProvider;
 use PHPStan\File\FileHelper;
-use PHPStan\File\RelativePathHelper;
+use PHPStan\File\SystemAgnosticSimpleRelativePathHelper;
 use PHPStan\Php\PhpVersion;
 use PHPStan\PhpDoc\PhpDocInheritanceResolver;
 use PHPStan\PhpDoc\StubPhpDocProvider;
@@ -146,8 +146,7 @@ abstract class TypeInferenceTestCase extends PHPStanTestCase
 	{
 		$fileHelper = self::getContainer()->getByType(FileHelper::class);
 
-		/** @var RelativePathHelper $relativePathHelper */
-		$relativePathHelper = self::getContainer()->getService('simpleRelativePathHelper');
+		$relativePathHelper = new SystemAgnosticSimpleRelativePathHelper($fileHelper);
 
 		$file = $fileHelper->normalizePath($file);
 
--- a/src/Testing/TypeInferenceTestCase.php
+++ b/src/Testing/TypeInferenceTestCase.php
@@ -12,14 +12,13 @@ use PHPStan\DependencyInjection\Type\DynamicThrowTypeExtensionProvider;
 use PHPStan\DependencyInjection\Type\ParameterClosureTypeExtensionProvider;
 use PHPStan\DependencyInjection\Type\ParameterOutTypeExtensionProvider;
 use PHPStan\File\FileHelper;
-use PHPStan\File\SimpleRelativePathHelper;
+use PHPStan\File\RelativePathHelper;
 use PHPStan\Php\PhpVersion;
 use PHPStan\PhpDoc\PhpDocInheritanceResolver;
 use PHPStan\PhpDoc\StubPhpDocProvider;
 use PHPStan\Reflection\InitializerExprTypeResolver;
 use PHPStan\Reflection\SignatureMap\SignatureMapProvider;
 use PHPStan\Rules\Properties\ReadWritePropertiesExtensionProvider;
-use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\ConstantScalarType;
 use PHPStan\Type\FileTypeMapper;
@@ -32,7 +31,6 @@ use function count;
 use function fclose;
 use function fgets;
 use function fopen;
-use function getcwd;
 use function in_array;
 use function is_dir;
 use function is_string;
@@ -146,11 +144,7 @@ abstract class TypeInferenceTestCase extends PHPStanTestCase
 	 */
 	public static function gatherAssertTypes(string $file): array
 	{
-		$currentWorkingDirectory = getcwd();
-		if ($currentWorkingDirectory === false) {
-			throw new ShouldNotHappenException();
-		}
-		$pathHelper = new SimpleRelativePathHelper($currentWorkingDirectory);
+		$pathHelper = self::getContainer()->getByType(RelativePathHelper::class);
 
 		$asserts = [];
 		self::processFile($file, static function (Node $node, Scope $scope) use (&$asserts, $file, $pathHelper): void {
--- a/src/Testing/TypeInferenceTestCase.php
+++ b/src/Testing/TypeInferenceTestCase.php
@@ -39,6 +39,7 @@ use function sprintf;
 use function stripos;
 use function strpos;
 use function strtolower;
+use function strtr;
 use function version_compare;
 use const PHP_VERSION;
 
--- a/src/Testing/TypeInferenceTestCase.php
+++ b/src/Testing/TypeInferenceTestCase.php
@@ -144,7 +144,8 @@ abstract class TypeInferenceTestCase extends PHPStanTestCase
 	 */
 	public static function gatherAssertTypes(string $file): array
 	{
-		$pathHelper = self::getContainer()->getByType(RelativePathHelper::class);
+		/** @var RelativePathHelper $pathHelper */
+		$pathHelper = self::getContainer()->getService('relativePathHelper');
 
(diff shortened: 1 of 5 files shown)


src/Testing/TypeInferenceTestCase.php
```diff
@@ -152,24 +156,32 @@ public static function gatherAssertTypes(string $file): array
 				return;
 			}
 
+			$projectRoot = realpath(__DIR__ . '/../../');
```
ondrejmirtes: This is useless, we're still inside the PHAR. Current working directory would make more sense.

ondrejmirtes (changes requested):

src/Testing/TypeInferenceTestCase.php
```diff
@@ -141,8 +144,14 @@ public function assertFileAsserts(
 	 */
 	public static function gatherAssertTypes(string $file): array
 	{
+		$currentWorkingDirectory = getcwd();
+		if ($currentWorkingDirectory === false) {
+			throw new ShouldNotHappenException();
+		}
+		$pathHelper = new SimpleRelativePathHelper($currentWorkingDirectory);
```
ondrejmirtes: Isn't this already available in the container?

ondrejmirtes: The CWD parameter for sure, maybe even this exact helper too.

ondrejmirtes (changes requested):

src/Testing/TypeInferenceTestCase.php
staabm: I initially tried to pick it from the container with:
`$pathHelper = self::getContainer()->getByType(SimpleRelativePathHelper::class);`
since this errored hard, I went with hardcoding.

I now had another look and realized that I have to get this type by the interfacename:
`$pathHelper = self::getContainer()->getByType(RelativePathHelper::class);`

ondrejmirtes: I'm worried this might be the fuzzy one? I think you're supposed to get the right one by name (different Container method). See config.neon.

staabm:
I will finalize this PR when back at my windows laptop next week

src/Testing/TypeInferenceTestCase.php
```diff
@@ -157,21 +162,32 @@ public static function gatherAssertTypes(string $file): array
 			$functionName = $nameNode->toString();
 			if (in_array(strtolower($functionName), ['asserttype', 'assertnativetype', 'assertvariablecertainty'], true)) {
 				self::fail(sprintf(
-					'Missing use statement for %s() on line %d.',
+					'Missing use statement for %s() in %s on line %d.',
 					$functionName,
+					$pathHelper->getRelativePath(strtr($file, '\\', '/')),
```
ondrejmirtes: We don't need to change the backslashes here I think. And if you wanted to, we have `FileHelper::normalizePath()` for that. You can run this function on both sides here (implementation & tests).

ondrejmirtes (changes requested):

src/Testing/TypeInferenceTestCase.php
staabm: the currentWorkingDirectory in https://github.com/phpstan/phpstan-src/blob/519982c55e43a20f72b02316e47327dfb1ba702b/src/File/SimpleRelativePathHelper.php#L19 is normalized to forward slashes.

when not doing `strtr($file, '\\', '/')` we pass on windows a path with backslashes and this makes the path not getting returned relative.

staabm: just tried `FileHelper::normalize` instead of `strtr($file, '\\', '/')`: it will not turn the slashes in the correct form to make `$pathHelper->getRelativePath` return a proper relative path

ondrejmirtes: Why wouldn't it? It works everywhere else in PHPStan's codebase.

staabm: I don't think so.

in other places we also do the `\\ -> /`  dance to make it work:
https://github.com/phpstan/phpstan-src/blob/494b7b0de21c08147a795330d5c233bd0166f841/tests/e2e/ResultCacheEndToEndTest.php#L194-L199

but I found a way to make it work.

src/Testing/TypeInferenceTestCase.php
```diff
@@ -143,8 +145,10 @@ public function assertFileAsserts(
 	 */
 	public static function gatherAssertTypes(string $file): array
 	{
+		$pathHelper = new SimpleRelativePathHelper(dirname(__DIR__, 2));
```
ondrejmirtes: This doesn't make sense as it will be somewhere deep in the PHAR, for users that use TypeInferenceTestCase outside of phpstan-src.

ondrejmirtes (changes requested):

src/Testing/TypeInferenceTestCase.php
staabm: last try in the latest git push. if this doesn't work for you I will close this PR.

ondrejmirtes: I committed my vision how it should have been done: https://github.com/phpstan/phpstan-src/pull/3166/commits/0e5311892b490f77e243ae2135d4a4e87126ff7b

ondrejmirtes:
Thank you.

As a next step we could stop using SimpleRelativePathHelper in more places in tests and start using SystemAgnosticSimpleRelativePathHelper instead. It could find more issues like it did here in PathConstantsTest.

PR #3167, opened 2024-06-19
 fix return type of sprintf with single %s format

pilif:
In the return type extension of sprintf(), there was a shortcut to assume the return type to be non-falsy-string when the format-string is non-falsy string.

That's not correct though if the format string is just a single '%s' placeholder (or a single '%x$s' placeholder). In that case, if the requested argument is of stringy-type, then the return type is the type of the stringy argument.

This commit extends the existing support for single-placeholder templates that currently deals with numeric placeholders to also deal with %s

this fixes phpstan/phpstan#11201

### DIFF
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -21,9 +21,11 @@ use function array_key_exists;
 use function array_shift;
 use function count;
 use function in_array;
+use function intval;
 use function is_string;
 use function preg_match;
 use function sprintf;
+use function substr;
 use function vsprintf;
 
 class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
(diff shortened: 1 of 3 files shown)


pilif:
please note the second commit in the PR that updates an existing test ([930266c](https://github.com/phpstan/phpstan-src/pull/3167/commits/930266c63f20dd4e3228671bc6db5530c38091bb))

That test case was invalid to begin with as the line in question would throw an `ArgumentCountError` at runtime.

The other option would be to remove the test-case entirely. I can totally do both.

pilif:
(heads up: I'm about to force-push a much better test that has many more assertions)

staabm:
I am also working on this extension for a few hours with some kind of general overhaul. it will also fix your issue.

we should sync how we want to proceed :)

pilif:
up to you. My PR is, by virtue of me being totally unfamiliar with the code base, the minimal change required to fix the problem.

Feel free to close this one in preference to your overhaul, or merge this one as a stop-gap (the underlying bug has cased real-life issues for me) and rebase your overhaul on top of this one - I'm totally fine with both.

staabm:
just opened a draft of my current state. would it be ok for you, that we concentrate on my PR first (because it moves a lot of moving parts, and we put your one later on top of it?).



https://github.com/phpstan/phpstan-src/pull/3168

pilif:
absolutely. that's good for me too.

I'll have a look whether your PR fixes the underlying bug and if so, all that's needed from me is my test-case (https://github.com/phpstan/phpstan-src/pull/3167/commits/66262bb017f5f8737c326aeb71f8f6834ec1fb59#diff-4d390d0bba719512f2479a712fd2d14e7131ad079d4dd8fcced037744e3f1df8)

ondrejmirtes:
Thank you!

If you have any improvements, you can send a follow-up PR 😊

pilif:
haha - it looks like @ondrejmirtes just caused some additional work for @staabm

Thank you for deciding for us - that saves me personally quite a bit of work

#3168 doesn't fix https://github.com/phpstan/phpstan/issues/11201 on its own I think.

PR #3168, opened 2024-06-19
Improve `sprintf` return type inference

staabm:
closes https://github.com/phpstan/phpstan/issues/11201
closes https://github.com/phpstan/phpstan/issues/10493

### DIFF
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -131,6 +131,11 @@ class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturn
 			$returnType = new IntersectionType($accessories);
 		}
 
+		return $this->getConstantType($argTypes, $returnType, $functionReflection, $scope);
+	}
+
+	private function getConstantType(array $argTypes, Type $fallbackReturnType, FunctionReflection $functionReflection, Scope $scope): Type
+	{
 		$values = [];
 		$combinationsCount = 1;
 		foreach ($argTypes as $argType) {
@@ -145,7 +150,7 @@ class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturn
 			}
 
 			if (count($constantScalarValues) === 0) {
-				return $returnType;
+				return $fallbackReturnType;
 			}
 
 			$values[] = $constantScalarValues;
@@ -153,7 +158,7 @@ class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturn
 		}
 
 		if ($combinationsCount > InitializerExprTypeResolver::CALCULATE_SCALARS_LIMIT) {
-			return $returnType;
+			return $fallbackReturnType;
 		}
 
 		$combinations = CombinationsHelper::combinations($values);
@@ -161,7 +166,7 @@ class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturn
 		foreach ($combinations as $combination) {
 			$format = array_shift($combination);
 			if (!is_string($format)) {
-				return $returnType;
+				return $fallbackReturnType;
 			}
 
 			try {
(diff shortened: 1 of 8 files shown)


tests/PHPStan/Analyser/nsrt/bug-7387.php
```diff
@@ -8,60 +8,61 @@ class HelloWorld
...
-		assertType('numeric-string', sprintf('%2$14F', $mixed, $i));
-		assertType('numeric-string', sprintf('%2$14F', $mixed, $f));
-		assertType('numeric-string', sprintf('%2$14F', $mixed, $s));
+		assertType('non-falsy-string&numeric-string', sprintf('%2$14F', $mixed, $i));
+		assertType('non-falsy-string&numeric-string', sprintf('%2$14F', $mixed, $f));
+		assertType('non-falsy-string&numeric-string', sprintf('%2$14F', $mixed, $s));
 
-		assertType('string', sprintf('%10$14F', $mixed, $s));
+		// XXX should be string because of invalid arguments count
+		assertType('non-falsy-string&numeric-string', sprintf('%10$14F', $mixed, $s));
```
staabm, line 71: after https://github.com/phpstan/phpstan-src/pull/3126 is merged we have the necessary `PrintfHelper` to decide about the invalid number of arguments/placeholders

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -48,70 +47,85 @@ public function getTypeFromFunctionCall(
...
+			if ($functionReflection->getName() === 'vsprintf') {
+				if ($argType->isIterableAtLeastOnce()->yes()) {
+					$isNonEmpty = true;
+				}
+				continue;
+			}
 
-			if ($singlePlaceholderEarlyReturn !== null) {
-				return $singlePlaceholderEarlyReturn;
+			if ($argType->toString()->isNonFalsyString()->yes()) {
```
staabm, line 116: requires https://github.com/phpstan/phpstan-src/pull/3170

staabm:
the composer error is valid, as we previously inferred a non-empty-string return type, even though the sprintf pattern is not known at static analysis time.

other errors are the same ones are unrelated

staabm:
the reported phpstan error is interesstig.. it seems - depending on some unknown factor - reports that we are catching a non-throwable but ignoring this very same error by identifer leads to a "No error with identifier catch.notThrowable is reported on line 200."

see the 2 variants at
- https://github.com/phpstan/phpstan-src/pull/3168/commits/ff9189e5e14e3cbe3dc1b87eba1cd09763cb52da
- https://github.com/phpstan/phpstan-src/pull/3168/commits/fd5e76270f38ebf0e04137fff4426d4db264d7d7

(its unrelated to this PR though)

staabm:
I think its related to this `Consistency::flexEntity` call in the exception classes.. some class-alias stuff is going on there.. this might lead to a race condition depending on the order which classes will be scanned when

ondrejmirtes:
There's an opportunity to grab the code from Hoa packages and make a small single modern library with no magic and simple PSR-4 autoloading and replace Hoa in phpstan-src. I might look into it this summer.

staabm:
> I might look into it this summer.

ok, what can we do to fix the build until then?

ondrejmirtes:
Try to rebase on top of: https://github.com/phpstan/phpstan-src/commit/323315554048a22e6c01d635d6c5bb68f10d5aa1

staabm:
> Try to rebase on top of: [3233155](https://github.com/phpstan/phpstan-src/commit/323315554048a22e6c01d635d6c5bb68f10d5aa1)

looks great, thanks

ondrejmirtes (changes requested):
I'm sorry. I see the diff of `src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php` and my brain just freezes. I'm not able to review this. Can you do this in smaller steps, or maybe not move that many things around?

"Make the change easy, then make the easy change."

PR #3170, opened 2024-06-19
Cast non-zero IntegerRangeType toString() is a non-falsy-string

staabm:
while working on https://github.com/phpstan/phpstan-src/pull/3168 I realized there is some opportunity to make this typing more precise


ondrejmirtes:
Thank you.

PR #3171, opened 2024-06-20
issue-bot: subshell is eating the exit code

staabm:
we want the github action to fail [in case of non zero exit status](https://github.com/phpstan/phpstan-src/blob/c1fc05398165ee36fa984879668371368e040b49/issue-bot/src/Console/EvaluateCommand.php#L163)

repro

```
➜  phpstan-src git:(1.11.x) echo "$(exit 1)" 

➜  phpstan-src git:(1.11.x) echo $?
0
```

### DIFF
--- a/.github/workflows/issue-bot.yml
+++ b/.github/workflows/issue-bot.yml
@@ -163,7 +163,7 @@ jobs:
         if: github.event_name == 'pull_request'
         env:
           GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
-        run: echo "$(./console.php evaluate)" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-bot.yml
```diff
@@ -163,7 +163,7 @@ jobs:
         if: github.event_name == 'pull_request'
         env:
           GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
-        run: echo "$(./console.php evaluate)" >> $GITHUB_STEP_SUMMARY
```
staabm, line 166: I don't know why this was invoked via a subshell - or at least I don't see a reason why

ondrejmirtes:
Thank you!

staabm:
just checked other workflows. could not find the same error somewhere else in this repo

PR #3173, opened 2024-06-20
Fix `conditionalTags` for ParameterOutExtensions

staabm:
locally verified in https://github.com/phpstan/phpstan-src/pull/2589


ondrejmirtes:
The closure parameter type extensions are probably missing too, please take a look.

staabm:
good call.

ondrejmirtes:
Thank you!

PR #3175, opened 2024-06-21
TypeSpecifier - understand preg_match() == 1 same way as === 1

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/2589#issuecomment-2182398610


ondrejmirtes:
Thank you!

PR #3177, opened 2024-06-22
Added ArrayShapeMatcher 3v4l example

staabm:


ondrejmirtes:
Would it be possible to rewrite this logic https://github.com/phpstan/phpstan-src/blob/8f4b3b79249f5af8e3fcfe2e793bbfe014964ab5/src/Type/Php/RegexArrayShapeMatcher.php#L99 so that it doesn't rely on `PREG_UNMATCHED_AS_NULL`?

staabm:
> so that it doesn't rely on `PREG_UNMATCHED_AS_NULL`?

the only way I could think of would be using only the Hoa\Ast, but I didn't find the features which would provide that information yet.

maybe thats something worth looking into, after Hoa\Ast is more intergrated into phpstan-src and we have more control over it

ondrejmirtes:
Can you tell me why? According to documentation:

> This flag tells [preg_match()](https://www.php.net/manual/en/function.preg-match.php) and [preg_match_all()](https://www.php.net/manual/en/function.preg-match-all.php) to include unmatched subpatterns in $matches as [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) values. Without this flag, unmatched subpatterns are reported as empty strings, as if they were empty matches. Setting this flag allows to distinguish between these two cases.

Why we can't achieve the same logic without this flag?

staabm:
> Without this flag, unmatched subpatterns are reported as empty strings, as if they were empty matches. Setting this flag allows to distinguish between these two cases.

the docs are wrong, see https://3v4l.org/6PfcK
(openend a php docs issue at https://github.com/php/doc-en/issues/3483)

as soon as we leave this flag off, the `$matches` array will not contain information about the unmatched matches.
ATM we rely on knowing about all possible matches to get an idea of how the array shape looks like.

ondrejmirtes:
@staabm So if they are not present in the array, we can fill them in ourselves because we know they are unmatched?

staabm:
we don't know atm how many there are. we don't get this information out of the AST atm.

staabm:
maybe we can make the AST regex walking smarter, but I was not yet in this details of Hoa\Ast

ondrejmirtes:
@staabm Alright, is that because we don't know what's in `$captureGroupsRegex` after `Strings::replace($regex, '~.[a-z\s]*$~i', '|(?<phpstanNamedCaptureGroupLast>)$0');`?

I just realized the logic in RegexArrayShapeMatcher is wrong. Our ability to do this doesn't depend on `$this->phpVersion->returnsPregUnmatchedCapturingGroups()` but on `PHP_VERSION_ID >= 70400`.

It doesn't matter how the user configured `phpVersion`. It matters how the runtime behaves.

staabm:
> It doesn't matter how the user configured `phpVersion`. It matters how the runtime behaves.

you are right. as long as we rely on the regex hack, this is the case.

staabm:
I remember another reason why we need the regex hack: there are some regex patterns which Hoa\Regex was not able to build a  AST for (but the PHP runtime provided the necessary information)

staabm:
obsolete with https://github.com/phpstan/phpstan-src/pull/3184

PR #3178, opened 2024-06-22
TypeSpecifier - understand \Composer\Pcre\Preg::match() == 1 same way as === 1

staabm:
required for composer/pcre type narrowing analog
- https://github.com/phpstan/phpstan-src/commit/328fcb91de1605c1a578e45d72f20952b3312501
- https://github.com/phpstan/phpstan-src/commit/d842380a1f07014d045dd42d6ccd4204a85e688a

required for https://github.com/composer/pcre/pull/24#discussion_r1648827531 to [cover the `==1` and `===1` testcases](https://github.com/composer/pcre/pull/24/files#diff-8f259a63a4ab66c0f035859cea17176b75144dfb4c81db9981e04b66168edc4cR32-R50)

### DIFF
--- a/build/composer-dependency-analyser.php
+++ b/build/composer-dependency-analyser.php
@@ -37,4 +37,5 @@ return $config
 	->ignoreUnknownClasses([
 		'JetBrains\PhpStorm\Pure', // not present on composer's classmap
 		'PHPStan\ExtensionInstaller\GeneratedConfig', // generated
+		'Composer\Pcre\Preg', // reverse dependency on composer/pcre
 	]);
--- a/build/composer-dependency-analyser.php
+++ b/build/composer-dependency-analyser.php
@@ -33,9 +33,9 @@ return $config
 	)
 	->ignoreErrorsOnPackage('phpunit/phpunit', [ErrorType::DEV_DEPENDENCY_IN_PROD]) // prepared test tooling
 	->ignoreErrorsOnPackage('jetbrains/phpstorm-stubs', [ErrorType::PROD_DEPENDENCY_ONLY_IN_DEV]) // there is no direct usage, but we need newer version then required by ondrejmirtes/BetterReflection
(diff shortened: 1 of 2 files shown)


build/composer-dependency-analyser.php
```diff
@@ -33,6 +33,7 @@
 	)
 	->ignoreErrorsOnPackage('phpunit/phpunit', [ErrorType::DEV_DEPENDENCY_IN_PROD]) // prepared test tooling
 	->ignoreErrorsOnPackage('jetbrains/phpstorm-stubs', [ErrorType::PROD_DEPENDENCY_ONLY_IN_DEV]) // there is no direct usage, but we need newer version then required by ondrejmirtes/BetterReflection
+	->ignoreErrorsOnPackage('composer/pcre', [ErrorType::SHADOW_DEPENDENCY]) // reverse dependency on composer/pcre to support https://github.com/composer/pcre/pull/24
```
staabm, line 36: @janedbal is this the correct way to handle a optional runtime dependency?

staabm:
Alternativly to this PR we could ask composer/pcre to turn the `0|1` returning api into a real `bool` (bc break).

And a third alternative could be that we handle callables which return `0|1` the same as `bool`

ondrejmirtes:
This isn't right, I don't want to hardcode things about other people's code in PHPStan's core.

The right way to handle this is to let TypeSpecifierContext know about various comparison operators and values on the other side of the comparison. So that extensions can react to this information.

staabm:
So you say we need a new extension type?

ondrejmirtes:
Not new extension type, we'd need to find a way how to pass this information in TypeSpecifierContext.

staabm:
ok, I need to think about that

ondrejmirtes:
My idea: right now we have methods null(), true(), truthy(), falsy(), false().

The new context called for extensions in these cases would return false for all of the current methods.

But there could be a new method called comparison() and it would return information the extension could act on.

staabm:
I will give it a try right now

PR #3183, opened 2024-06-24
More precise `getimagesize*` return type

staabm:
index 0/1 is width/height of the image which cannot be negative


resources/functionMap.php
```diff
@@ -3331,8 +3331,8 @@
 'gethostbyname' => ['string', 'hostname'=>'string'],
 'gethostbynamel' => ['list<string>|false', 'hostname'=>'string'],
 'gethostname' => ['string|false'],
-'getimagesize' => ['array{0:int, 1: int, 2: int, 3: string, mime: string, channels?: int, bits?: int}|false', 'imagefile'=>'string', '&w_info='=>'array'],
-'getimagesizefromstring' => ['array{0:int, 1: int, 2: int, 3: string, mime: string, channels?: int, bits?: int}|false', 'data'=>'string', '&w_info='=>'array'],
+'getimagesize' => ['array{0: positive-int, 1: 0|positive-int, 2: int, 3: string, mime: string, channels?: int, bits?: int}|false', 'imagefile'=>'string', '&w_info='=>'array'],
```
staabm: we already handle `imagesx` and `imagesy` as non-zero return, therefore I turned it eventually into just `positive-int`

ondrejmirtes:
Thank you.

resources/functionMap.php
mfruehauf: The new version of phpstan found this in our code.

https://phpstan.org/r/f0ef9a4e-00d2-4df9-a129-a79857f28273

But the [documentation](https://www.php.net/manual/en/function.getimagesize.php) of PHP says:

> Some formats may contain no image or may contain multiple images. In these cases, getimagesize() might not be able to properly determine the image size. getimagesize() will return zero for width and height in these cases.

staabm: I see thanks. please open a PR which uses `0|positive-int` instead of `positive-int`

mfruehauf: done. https://github.com/phpstan/phpstan-src/pull/3207

pereorga: @mfruehauf thanks!

Also, note that if width is greater than 0, height will necessarily be greater than 0 too. (In your example, there is no need to check for both)

PR #3184, opened 2024-06-24
RegexArrayShapeMatcher: fix nested capturing counting

staabm:
closes https://github.com/phpstan/phpstan/issues/11221


tests/PHPStan/Analyser/nsrt/preg_match_shapes_php74.php
```diff
@@ -232,3 +232,20 @@ function testUnionPattern(string $s): void
...
+}
+
+
+function doFoo2(string $row): void
+{
+	if (preg_match('~^((?<branchCode>\\d{1,6})-)?(?<accountNumber>\\d{1,10})/(?<bankCode>\\d{4})$~', $row, $matches) !== 1) {
+		return;
+	}
+
+	assertType('array{0: string, 1: string, branchCode: string, 2: string, accountNumber?: string, 3?: string, bankCode?: string, 4?: string}', $matches);
```
staabm: looking again at it, I think its better but still wrong :-/

staabm:
Will have another look later

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -85,56 +86,53 @@ public function matchType(Type $patternType, ?Type $flagsType, TrinaryLogic $was
 	 */
 	private function matchRegex(string $regex, ?int $flags, TrinaryLogic $wasMatched): ?Type
 	{
-		// add one capturing group to the end so all capture group keys
-		// are present in the $matches
-		// see https://3v4l.org/sOXbn, https://3v4l.org/3SdDM
-		$captureGroupsRegex = Strings::replace($regex, '~.[a-z\s]*$~i', '|(?<phpstanNamedCaptureGroupLast>)$0');
```
staabm, line 91: reworked the matcher to be AST only -> no more php version differences

tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -172,10 +172,6 @@ public function dataFileAsserts(): iterable
 			yield from $this->gatherAssertTypes(__DIR__ . '/../Rules/Comparison/data/bug-9499.php');
 		}
 
-		if (PHP_VERSION_ID >= 70300 && PHP_VERSION_ID < 70400) {
-			yield from $this->gatherAssertTypes(__DIR__ . '/data/preg_match_shapes_php73.php');
```
staabm, line 176: this test is contained in `preg_match_shapes_php.php` with proper expectations across all php versions. we no longer need separate expectations for 7.2/7.3

staabm:
I did not yet figure out, how/why the result of the last failling test-case can be different on 7.2/7.3

staabm:
> I did not yet figure out, how/why the result of the last failling test-case can be different on 7.2/7.3

I now know what the cause of the problem is:

https://github.com/phpstan/phpstan-src/blob/323315554048a22e6c01d635d6c5bb68f10d5aa1/src/Type/Php/PregMatchParameterOutTypeExtension.php#L26

in PHP 7.2/7.3 the parameter is named "subpatterns" and in 7.4+ its "matches"

tests/PHPStan/Analyser/data/param-out.php
```diff
@@ -485,3 +485,20 @@ function ($s, $t): void {
 	assertType('string', $s);
 	assertType('string', $t);
 };
+
+function fooMatch(string $input): void {
```
staabm, line 489: expectations have been harmonized across php versions, therefore the 2 variants above removed and the tests moved with same expectations in this common file

staabm:
@ondrejmirtes would be great to get this one merged, as it is now php version independent and in turn would also allow to fix the phpstan-nette build.

it fixes all currently known problems (`preg_match bugs`-ticket)

ondrejmirtes:
Vibing cat

Purrrrfect!

ondrejmirtes:
Thank you!

PR #3186, opened 2024-06-25
Forbid _HumbugBox prefixed classes

staabm:
as I stumbled over this problem just today while working on phpstan-src :-)

see https://github.com/box-project/box/blob/bfd4b744e902c063c3eb4fff37c2542170922b72/src/Configuration/Configuration.php#L2664 for the configured prefix in Box


ondrejmirtes:
Thank you.

PR #3188, opened 2024-06-25
Simplify SprintfFunctionDynamicReturnTypeExtension

staabm:
extracted from phpstan/phpstan-src/pull/3168

- adds early returns to reduce complexity
- return early and with default-return in error cases - see https://3v4l.org/u2N69. 
[for `sprintf` we already report an error](https://phpstan.org/r/801c119a-1133-4a75-9d68-05c6edd87c2b) and for `vsprintf` we will emit a error after https://github.com/phpstan/phpstan-src/pull/3126


src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -48,50 +48,64 @@ public function getTypeFromFunctionCall(
 		}
 
 		$formatType = $scope->getType($args[0]->value);
+		if (count($args) === 1) {
+			return null;
+		}
 
-		if (count($formatType->getConstantStrings()) > 0) {
```
staabm, line 53: the loop is the same as before, just the if statement was unwrapped

staabm:
the new composer/composer errors are valid because the typing before this PR was too precise.

we can't know whether `sprintf` will return `non-empty-string` in a scenario where the format-pattern is a non-constant variable.
https://github.com/composer/composer/blob/776ff2ea51bfa83335b14fa9a85fab5fc7d35050/src/Composer/Package/BasePackage.php#L256
https://github.com/composer/composer/blob/776ff2ea51bfa83335b14fa9a85fab5fc7d35050/src/Composer/Package/BasePackage.php#L275

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -48,50 +48,64 @@ public function getTypeFromFunctionCall(
...
-						if ($matches[1] === '0$') {
-							return null;
-						}
-						$checkArg = intval(substr($matches[1], 0, -1));
-					} else {
-						$checkArg = 1;
-					}
+		if (
+			$functionReflection->getName() === 'vsprintf'
+			&& count($args) === 2
```
ondrejmirtes: Please add some tests in case when there is `$arg->unpack === true` involved.

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -48,50 +48,64 @@ public function getTypeFromFunctionCall(
 		}
 
 		$formatType = $scope->getType($args[0]->value);
+		if (count($args) === 1) {
+			return null;
```
ondrejmirtes: We know what would happen so we could be more precise: https://3v4l.org/DeSsG

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/non-falsy-string.php
```diff
@@ -104,8 +104,17 @@ function stringFunctions(string $s, $nonFalsey, $arrayOfNonFalsey, $nonEmptyArra
 
 		assertType('non-falsy-string', preg_quote($nonFalsey));
 
-		assertType('non-falsy-string', sprintf($nonFalsey));
-		assertType('non-falsy-string', vsprintf($nonFalsey, []));
+		assertType('string', sprintf($nonFalsey));
+		assertType("'foo'", sprintf('foo'));
+		assertType("string", sprintf(...$arr));
+		assertType("non-falsy-string", sprintf('%s', ...$arr)); // should be 'string'
```
staabm, line 110: the wrong `non-falsy-string` types will be fixed after we integrated the remaining parts of https://github.com/phpstan/phpstan-src/pull/3168

ondrejmirtes:
Thank you.

ondrejmirtes:
Hey, if I have `sprintf($nonEmptyString, $alsoNonEmptyString)`, why the inferred type isn't `non-empty-string`?

herndlm:
@staabm how do you find the time to work so  much on PHPStan. I'm doing something wrong 😅 recently too busy and I do lack motivation to work on it at night..

PR #3189, opened 2024-06-25
RegexArrayShapeMatcher - when all groups are optional return a more…

staabm:
…precise union, instead of a shape with optional offsets

implements the feature request https://phpstan.org/r/fac92fd0-b78e-4e36-9089-ec8f7ece17c6

refs https://github.com/phpstan/phpstan/issues/11221#issuecomment-2189044796

### DIFF
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -16,6 +16,7 @@ use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\UnionType;
 use function array_reverse;
 use function count;
 use function in_array;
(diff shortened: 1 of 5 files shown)


staabm:
quest for another PR: how to get smarter for alternations in the regex.
atm our typing gets wonkey when the regex patter contains `|` with capturing groups in the alternation.

maybe tomorrow ;)

staabm:
took me a lot of attempts but I think the current design works now.

when a regex contains only a single optional capturing group, we build a more precise constant type.
there is still room for improvement but I think its a good next step.

src/Type/Php/RegexNonCapturingGroup.php
```diff
@@ -0,0 +1,29 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+class RegexNonCapturingGroup
```
staabm, line 5: the optional groups detection, now also takes non-capturing groups into account... these do not capture results itself, but might impact nested capturing groups

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -73,14 +73,88 @@ public function matchType(Type $patternType, ?Type $flagsType, TrinaryLogic $was
...
+		}
+
 		$valueType = $this->getValueType($flags ?? 0);
+		$onlyOptionalTopLevelGroup = $this->getOnlyOptionalTopLevelGroup($groupList);
+		if (
+			$wasMatched->yes()
+			&& $onlyOptionalTopLevelGroup !== null
+		) {
+			// if only one top level capturing optional group exists
+			// we build a more precise constant union of a empty-match and a match with the group
```
staabm, line 97: I am limiting to this specific case for now, because whether $matches contains a offset or not depends on many factors. I did this already for more stuff, but it was easily breaking stuff.

will bring back more sophiscated stuff in future PRs

ondrejmirtes:
Thank you!

PR #3190, opened 2024-06-27
Improved the type of the $mode parameter for the `count()`

kuma3:
From the manual for the `count()`: https://www.php.net/manual/en/function.count.php

the parameter $mode accepts COUNT_NORMAL (0) or COUNT_RECURSIVE (1). I have made the necessary adjustments and added tests accordingly.


ondrejmirtes:
Narrowing parameter types belongs to functionMap_bleedingEdge so that no userland code is broken in a minor release.

kuma3:
OK. I will move the implementation to functionMap_bleedingEdge.

ondrejmirtes:
Thank you.

PR #3191, opened 2024-06-27
Use `NullableTypeNode` for unions in `ConstantArrayType::toPhpDocNode`

ruudk:
Currently, `ConstantArrayType::toPhpDocNode()` produces the following code when you have an array shape with nullable string type for key `foo`:

```
array{foo: (string | null)}
```

It would be great to convert those shapes to:
```
array{foo: ?string}
```

This feels more natural, especially because PHP also has the ?string type.

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -13,6 +13,7 @@ use PHPStan\PhpDocParser\Ast\Type\ArrayShapeItemNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayShapeNode;
 use PHPStan\PhpDocParser\Ast\Type\ConstTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
+use PHPStan\PhpDocParser\Ast\Type\NullableTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\TypeNode;
 use PHPStan\Reflection\Callables\FunctionCallableVariant;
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Hi, I'm sorry, but I disagree with this. You're trying to change the code to adhere to your personal taste, otherwise the resulting PHPDoc node is equivalent before and after.

I have to disagree with this, because someone else might have the exact opposite preferences.

If you're using `Type::toPhpDocNode()` in your own code to achieve something, you can transform the resulting node with NodeTraverser (https://phpstan.github.io/phpdoc-parser/PHPStan.PhpDocParser.Ast.NodeTraverser.html). And not just in ArrayShapeItemNode but everywhere. That gives you the flexibility to make the node look exactly like you want it to.

Thanks for understanding.

ruudk:
Thanks for your reply. I knew this was a long shot so I get it! 😊 

I did not think of traversing the PHPDoc's AST. That should work for the union case. 👍 

I also noticed another thing. When I have a `list` of `string` and call `toPhpDocNode` on it, it produces `array{string}` instead of `list<string>`. Is this a bug? Or should I also try to fix this with traversing?

I think traversing would be hard, unless I'm able to connect all the PHPDocNode's to the original PHPStan types? Is that possible?

ondrejmirtes:
Yeah, that sounds like a bug, although very impropable, please try to reproduce it. There's a specific test case for toPhpDocNode in tests/.

What's in PHPStan\Type\Type that you're missing after conversion to AST?

PR #3192, opened 2024-06-28
More precise sprintf() format-arg based return type

staabm:
extracted from https://github.com/phpstan/phpstan-src/pull/3168

closes https://github.com/phpstan/phpstan/issues/11248
closes https://github.com/phpstan/phpstan/issues/10493

the previous format-type based logic was too optimistic regarding types and false-positively created too precise types in some cases.

### DIFF
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -189,14 +189,14 @@ class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturn
 			return null;
 		}
 
-		try {
-			$dummyValues = array_fill(0, $valuesCount, '');
-			if ($dummyValues === false) { // @phpstan-ignore identical.alwaysFalse PHP7.2 compat
-				return null;
-			}
+		$dummyValues = array_fill(0, $valuesCount, '');
+		if ($dummyValues === false) { // @phpstan-ignore identical.alwaysFalse (PHP7.2 compat)
+			return null;
+		}
 
+		try {
 			$formatted = @vsprintf($format, $dummyValues);
-			if ($formatted === false) { // @phpstan-ignore identical.alwaysFalse PHP7.2 compat
+			if ($formatted === false) { // @phpstan-ignore identical.alwaysFalse (PHP7.2 compat)
 				return null;
 			}
 			return new ConstantStringType($formatted);
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -37,8 +37,6 @@ use function vsprintf;
 class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
 {
 
-	private const MAX_INTERPOLATION_RETRIES = 5;
-
 	public function isFunctionSupported(FunctionReflection $functionReflection): bool
 	{
 		return in_array($functionReflection->getName(), ['sprintf', 'vsprintf'], true);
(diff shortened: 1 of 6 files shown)


src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -118,6 +148,28 @@ public function getTypeFromFunctionCall(
...
 	}
 
+	private function getFormatConstantParts(string $format): ?ConstantStringType
+	{
+		$dummyValues = [];
+		for ($i = 0; $i < self::MAX_INTERPOLATION_RETRIES; $i++) {
+			$dummyValues[] = '';
+
+			try {
+				$formatted = @sprintf($format, ...$dummyValues);
```
staabm: ahh interessting. because of a pre-existing bug, this line results in phpstan errors :).
we are missing `sprintf` can return `false` on php7

https://phpstan.org/r/e0116e22-6029-4e08-9079-4f5128bef6a8

https://3v4l.org/hr2rl

staabm: declined via https://github.com/phpstan/phpstan-src/pull/3193

tests/PHPStan/Analyser/nsrt/non-falsy-string.php
```diff
@@ -107,13 +107,17 @@ function stringFunctions(string $s, $nonFalsey, $arrayOfNonFalsey, $nonEmptyArra
 		assertType('string', sprintf($nonFalsey));
 		assertType("'foo'", sprintf('foo'));
 		assertType("string", sprintf(...$arr));
-		assertType("non-falsy-string", sprintf('%s', ...$arr)); // should be 'string'
```
staabm, line 110: fixes previous too specific types

staabm:
I think the remaining error is a bug: https://github.com/phpstan/phpstan/issues/11249

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -118,6 +152,28 @@ public function getTypeFromFunctionCall(
 		return $this->getConstantType($args, $returnType, $functionReflection, $scope);
 	}
 
+	private function getFormatConstantParts(string $format): ?ConstantStringType
+	{
+		$dummyValues = [];
+		for ($i = 0; $i < self::MAX_INTERPOLATION_RETRIES; $i++) {
```
ondrejmirtes: Errr, what is this for?

ondrejmirtes (changes requested):

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
staabm: the idea is to replace all placeholders in a pattern with empty string, to detect whether the pattern contains any constant parts and which type of string this constant part has

`%sAA%s` -> `AA` -> non-falsy-string
`%s0%s` -> `0` -> non-empty-string
`0%s0%s` -> `00` -> non-falsy-string

so we know a basic result type for the worst case in which we don't know details about parameters passed to the `sprintf` call:

`sprintf('%sAA%s', $s, $s)` -> will always be non-falsy-string, no matter the type of `$s`.

ondrejmirtes: Why is this in a for loop with a number of retries if we know the number of placeholders and we know how many values the function call needs? I feel like this could be done completely statically, with some kind of regular expression perhaps.

staabm: simplified so it no longer loops

staabm:
sorry for all these commits to make it finally build on php 7.2 without errors.

ondrejmirtes:
Thank you!

PR #3193, opened 2024-06-28
(v)sprintf() can return false on PHP7

staabm:
we are missing `sprintf`/`vsprintf` can return `false` on php7

https://3v4l.org/aEqBC


staabm:
hmm I wonder how annoying this change is in practice, as can be seen already at the phpstan own codebase ([example](https://github.com/phpstan/phpstan-src/actions/runs/9709951834/job/26799660753?pr=3193)), that this triggers lots of warnings.

@ondrejmirtes wdyt about turning this into a benevolent `(string|false)` on php7?

ondrejmirtes:
The fact that there's no bug report for this means that this `false` rarely happens in practice and gets rarely handled which means I don't want to add it to the return type on PHP 7.x.

PR #3195, opened 2024-06-28
Avoid using AlwaysRememberedExpr when match expression condition is a `BinaryOperation`

jdecool:
Fix https://github.com/phpstan/phpstan/issues/11246

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3509,6 +3509,9 @@ class MutatingScope implements Scope
 		if ($expr->cond instanceof AlwaysRememberedExpr) {
 			return $this;
 		}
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
I'm sorry, but I fell like this isn't the right solution. But I haven't found the right one yet. I thought that we need to unset `AlwaysRememberedExpr` in `generalizeVariableTypeHolders` and maybe also `mergeVariableHolders` but it didn't work for me.

ondrejmirtes:
This was the right fix: https://github.com/phpstan/phpstan-src/commit/07995276abd80b766407534055f58592ffe1a1a5

Thank for the test.

jdecool:
Thanks @ondrejmirtes

PR #3196, opened 2024-06-30
`Doctrine\ORM\EntityManager` could be overridden in analyzing

simonberger:
I tried out phpstan-src directly to look into some potential changes. While doing so I faced a crash in the doctrine plugin code when analyzing a project which worked fine with the phpstan phar variant.
It turned out to happen because a file of a test `catch-with-unthrown-exception-stubs.php` registers in the autoloader as `Doctrine\ORM\EntityManager` and overrides the real class. The file was added just in March, yet I wondered why it did not affect other developers. Maybe because the doctrine plugin was not used.

I just picked a slightly different name for the EntityManager class. It does not matter for the test to do its job.

### DIFF
--- a/tests/PHPStan/Rules/Exceptions/data/catch-with-unthrown-exception-stubs.php
+++ b/tests/PHPStan/Rules/Exceptions/data/catch-with-unthrown-exception-stubs.php
@@ -1,7 +1,7 @@
 <?php
 
 namespace Doctrine\ORM {
-	class EntityManager
+	class EntityManagerPHPStanAlias
 	{
 
 		public function transactional(callable $cb): void {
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
> I tried out phpstan-src directly to look into some potential changes

phpstan-src is only for PHPStan's development. It's not supported as a way to analyse other projects.

Always use phpstan/phpstan for that.

ondrejmirtes:
This is a WONTFIX. Thanks for understanding.

simonberger:
> phpstan-src is only for PHPStan's development. It's not supported as a way to analyse other projects.
> 
> Always use phpstan/phpstan for that.

I use `phpstan/phpstan`. Yet this project may/must be used on codebase to debug bugs, performance, understanding internals and what not. 
A response like this is discouraging, especially as this change does not hurt anyone.

ondrejmirtes:
In order for this change to be accepted, there would have to be tests for the kind of problem this solves. Otherwise this could be reverted or broken again at any time.

If there were tests it'd signal we would want to support this kind of usage. Which we don't.

Yeah, this change doesn't hurt anyone in the short term but longterm it increases the maintenance costs and steers the direction of the project somewhere I don't want.

If you wanted to use phpstan-src to analyse a project in order to fix a bug in PHPStan, you should reproduce the bug in the test suite here. Then you can fix it the same way.

If you want to make sure your change doesn't break anything, you can open a pull request here which will cause two things to happen:

1) Your change will be tested extensively on many real world projects as part of CI here.
2) You will be able to download phpstan.phar with your change which you can copy to your project and run as you usually run PHPStan.

Hopefully this explanation is sufficient.

PR #3197, opened 2024-06-30
Use early-returns in TypeCombinator

staabm:
just a drive-by cleanup while studying the code. no functional changes.

motivation: improve readability by reducing the indentation

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -765,62 +765,62 @@ class TypeCombinator
 	{
 		$constantArrayValuesCount = self::countConstantArrayValueTypes($types);
 
-		if ($constantArrayValuesCount > ConstantArrayTypeBuilder::ARRAY_COUNT_LIMIT) {
-			$results = [];
-			foreach ($types as $type) {
-				$results[] = TypeTraverser::map($type, static function (Type $type, callable $traverse): Type {
-					if ($type instanceof ConstantArrayType) {
-						if ($type->isIterableAtLeastOnce()->no()) {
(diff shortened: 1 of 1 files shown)


src/Type/TypeCombinator.php
```diff
@@ -765,62 +765,62 @@ private static function optimizeConstantArrays(array $types): array
 	{
 		$constantArrayValuesCount = self::countConstantArrayValueTypes($types);
 
-		if ($constantArrayValuesCount > ConstantArrayTypeBuilder::ARRAY_COUNT_LIMIT) {
```
staabm, line 768: inverted the condition, so we can de-indent the big logic block 1 level

src/Type/TypeCombinator.php
```diff
@@ -765,62 +765,62 @@ private static function optimizeConstantArrays(array $types): array
...
-								}
-
-								return $traverse($type);
-							});
-							$valueTypes[$generalizedValueType->describe(VerbosityLevel::precise())] = $generalizedValueType;
-						}
+		$results = [];
+		foreach ($types as $type) {
+			$results[] = TypeTraverser::map($type, static function (Type $type, callable $traverse): Type {
+				if (!$type instanceof ConstantArrayType) {
```
staabm, line 775: same as above: just added a early return so we can de-indent the code

staabm:
see a shorted diff with https://github.com/phpstan/phpstan-src/pull/3197/files?diff=unified&w=1

ondrejmirtes:
Thank you.

PR #3198, opened 2024-06-30
Added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/9879

was fixed with https://github.com/phpstan/phpstan-src/commit/07995276abd80b766407534055f58592ffe1a1a5


ondrejmirtes:
Thank you!

PR #3201, opened 2024-07-01
TypeCombinator: Don't combine array-shapes of lists with non-lists

staabm:
idea is to not combine lists with non-lists array-shapes. keeping them separate yields a more precise union which later on can lead to better analysis results

motivated by the issue described in https://github.com/phpstan/phpstan/discussions/11256#discussioncomment-9917907

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -1532,6 +1532,10 @@ class ConstantArrayType extends ArrayType implements ConstantType
 			return $keyTypesCount === 0;
 		}
 
+		if (!$this->isList->equals($otherArray->isList)) {
+			return false;
+		}
+
 		$failOnDifferentValueType = $keyTypesCount !== $otherKeyTypesCount || $keyTypesCount < 2;
 
 		$keyTypes = $this->keyTypes;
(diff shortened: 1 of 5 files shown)


staabm:
would the [test from the discussion](https://github.com/phpstan/phpstan/discussions/11256#discussion-6880059) be a good fit for a new `TypeCombinatorTest` ?

staabm:
I think this change makes sense as is in isolation. 

maybe it does not yet address all the problems I need to solve [for the case at hand](https://github.com/phpstan/phpstan/discussions/11256#discussioncomment-9917907) and needs another fix after I can combine it with the open `preg_match` efforts.

src/Type/Constant/ConstantArrayType.php
```diff
@@ -1532,6 +1532,10 @@ public function isKeysSupersetOf(self $otherArray): bool
 			return $keyTypesCount === 0;
 		}
 
+		if (!$this->isList->equals($otherArray->isList)) {
```
ondrejmirtes, line 1535: I'd rather have this line of code say:

```php
if ($this->isList->yes() !== $otherArray->isList->yes()) {
```

But I can't explain why, feel better to me. I'm going to let you come up with test cases that will have different results because of this change :)

ondrejmirtes (changes requested):

staabm:
The initial motivation of this PR was fixed differently

PR #3202, opened 2024-07-01
RegexArrayShapeMatcher - Narrow types in alternation of top level capturing groups

staabm:
fixes the 2 repros of https://github.com/phpstan/phpstan/issues/11221#issuecomment-2198551145

### DIFF
--- a/src/Type/Php/RegexCapturingGroup.php
+++ b/src/Type/Php/RegexCapturingGroup.php
@@ -78,7 +78,7 @@ class RegexCapturingGroup
 			|| $this->parent !== null && $this->parent->isOptional();
 	}
 
-	public function isOptionalAlternation(): bool
+	public function inOptionalAlternation(): bool
 	{
 		if (!$this->inAlternation()) {
 			return false;
--- a/src/Type/Php/RegexCapturingGroup.php
+++ b/src/Type/Php/RegexCapturingGroup.php
@@ -5,40 +5,16 @@ namespace PHPStan\Type\Php;
 class RegexCapturingGroup
 {
 
-	private int $id;
-
 	private bool $forceNonOptional = false;
 
-	private static int $idGenerator = 100;
-
-	private function __construct(
+	public function __construct(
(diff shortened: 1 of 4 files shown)


src/Type/Php/RegexCapturingGroup.php
```diff
@@ -5,49 +5,99 @@
...
 	private function __construct(
 		private ?string $name,
-		private bool $inAlternation,
+		private ?int $alternationId,
 		private bool $inOptionalQuantification,
 		private RegexCapturingGroup|RegexNonCapturingGroup|null $parent,
 	)
 	{
+		$this->id = self::$idGenerator;
+		self::$idGenerator++;
```
ondrejmirtes: I don't like global state like that. This should be passed as a plain number from the outside and counted in the instance that creates objects of this class.

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -306,3 +306,21 @@ function groupsOptional(string $size): void
...
 }
+
+function topLevelAlternation(string $size): void
+{
+	if (preg_match('~^(?:(\\d+)x(\\d+)|(\\d+)|x(\\d+))$~', $size, $matches) !== 1) {
+		throw new InvalidArgumentException(sprintf('Invalid size "%s"', $size));
+	}
+	assertType('array{0: string, 1: string, 2: string, 3?: string, 4?: string}', $matches);
+
+	if (preg_match('~^(?:(\\d+)x(\\d+)|(\\d+)|x(\\d+))?$~', $size, $matches) !== 1) {
```
ondrejmirtes, line 327: Better to divide this in multiple functions, reusing the `$matches` variable like that seems dirty to me.

ondrejmirtes (changes requested):

staabm:
feedback addressed - thank you

ondrejmirtes:
Thank you!

PR #3203, opened 2024-07-02
Faster ObjectType->getEnumCases()

staabm:
blackfire profile of the case in https://github.com/phpstan/phpstan/issues/11263#issuecomment-2202019541

25 %faster in https://github.com/phpstan/phpstan/issues/11263


ondrejmirtes:
Thank you.

PR #3206, opened 2024-07-02
Fix typo in ObjectType

staabm:


src/Type/ObjectType.php
```diff
@@ -1373,9 +1373,11 @@ public function changeSubtractedType(?Type $subtractedType): Type
 			$classReflection = $this->getClassReflection();
 			$allowedSubTypesList = $classReflection !== null ? $classReflection->getAllowedSubTypes() : null;
 			if ($allowedSubTypesList !== null) {
+				$preciseVerbosity = VerbosityLevel::precise();
+
 				$allowedSubTypes = [];
 				foreach ($allowedSubTypesList as $allowedSubType) {
-					$allowedSubTypes[$allowedSubType->describe(VerbosityLevel::precise())] = $allowedSubType;
```
staabm, line 1378: this are a hot loops, so don't create the verbosity over and over

repro of https://github.com/phpstan/phpstan/issues/11263

src/Type/ObjectType.php
```diff
@@ -1230,15 +1230,15 @@ public function getEnumCases(): array
 		$className = $classReflection->getName();
 
 		if ($this->subtractedType !== null) {
-			$subtracedEnumCaseNames = [];
+			$subtractedEnumCaseNames = [];
```
staabm, line 1233: typo-fix

src/Type/ObjectType.php
```diff
@@ -1384,7 +1386,7 @@ public function changeSubtractedType(?Type $subtractedType): Type
 				$subtractedTypesList = TypeUtils::flattenTypes($subtractedType);
 				$subtractedTypes = [];
 				foreach ($subtractedTypesList as $type) {
-					$subtractedTypes[$type->describe(VerbosityLevel::precise())] = $type;
```
staabm, line 1387: I think we don't need the key of this array and building up the descriptions shows up in profiles of phpstan/phpstan#11263

staabm: after getting rid of the key, it seems we can get rid of the whole loop

staabm:
intially started as a typo fix, but after some closer looks I was able to remove some unnecessary code

repro of https://github.com/phpstan/phpstan/issues/11263 before this PR:
```
php bin/phpstan analyze slow.php --debug  
 7.58s user 0.15s system 99% cpu 7.787 total

php bin/phpstan analyze slow.php --debug
 7.60s user 0.14s system 99% cpu 7.811 total
```

after
```
php bin/phpstan analyze slow.php --debug  
 7.13s user 0.14s system 99% cpu 7.286 total

php bin/phpstan analyze slow.php --debug 
 7.12s user 0.15s system 99% cpu 7.292 total

```

ondrejmirtes:
Thank you.

PR #3207, opened 2024-07-02
fix getimagesize: zero is a possible value for height and width

mfruehauf:
https://phpstan.org/r/f0ef9a4e-00d2-4df9-a129-a79857f28273

See [documentation](https://www.php.net/manual/en/function.getimagesize.php):
> Some formats may contain no image or may contain multiple images. In these cases, getimagesize() might not be able to properly determine the image size. getimagesize() will return zero for width and height in these cases.


resources/functionMap.php
```diff
@@ -3331,7 +3331,7 @@
 'gethostbyname' => ['string', 'hostname'=>'string'],
 'gethostbynamel' => ['list<string>|false', 'hostname'=>'string'],
 'gethostname' => ['string|false'],
-'getimagesize' => ['array{0: positive-int, 1: positive-int, 2: int, 3: string, mime: string, channels?: int, bits?: int}|false', 'imagefile'=>'string', '&w_info='=>'array'],
+'getimagesize' => ['array{0: 0|positive-int, 1: 0|positive-int, 2: int, 3: string, mime: string, channels?: int, bits?: int}|false', 'imagefile'=>'string', '&w_info='=>'array'],
 'getimagesizefromstring' => ['array{0: positive-int, 1: positive-int, 2: int, 3: string, mime: string, channels?: int, bits?: int}|false', 'data'=>'string', '&w_info='=>'array'],
```
staabm: IMO we need the same fix in this line

mfruehauf: Yes, I will fix it.

staabm (approved):
Lgtm - thank you

ondrejmirtes:
Thank you.

PR #3208, opened 2024-07-03
Faster EnumCaseObjectType->describe()

staabm:
repro of https://github.com/phpstan/phpstan/issues/11263 before this PR
```
php bin/phpstan analyze slow.php --debug
  7.16s user 0.14s system 99% cpu 7.323 total

php bin/phpstan analyze slow.php --debug
  7.18s user 0.13s system 99% cpu 7.339 total
```

after this PR:
```
php bin/phpstan analyze slow.php --debug
  5.29s user 0.14s system 99% cpu 5.478 total

php bin/phpstan analyze slow.php --debug
  5.25s user 0.14s system 99% cpu 5.403 total
```

### DIFF
--- a/src/Type/Enum/EnumCaseObjectType.php
+++ b/src/Type/Enum/EnumCaseObjectType.php
@@ -20,15 +20,11 @@ use PHPStan\Type\ObjectType;
 use PHPStan\Type\SubtractableType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
-use function array_key_exists;
 use function sprintf;
 
 /** @api */
 class EnumCaseObjectType extends ObjectType
 {
-
-	/** @var array<int, string> */
-	private array $descriptions = [];
 
 	/** @api */
 	public function __construct(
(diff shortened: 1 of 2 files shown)


staabm:
do you think we are fast enough to add a regression test for the bug and call it a day?

src/Type/Enum/EnumCaseObjectType.php
```diff
@@ -43,9 +47,11 @@ public function getEnumCaseName(): string
...
 	public function describe(VerbosityLevel $level): string
 	{
-		$parent = parent::describe($level);
-
-		return sprintf('%s::%s', $parent, $this->enumCaseName);
+		$value = $level->getLevelValue();
+		if (!array_key_exists($value, $this->descriptions)) {
+			$this->descriptions[$value] = sprintf('%s::%s', parent::describe($level), $this->enumCaseName);
+		}
+		return $this->descriptions[$value];
```
ondrejmirtes: Wouldn't it make sense to cache this in the parent instead? In ObjectType

ondrejmirtes (changes requested):

src/Type/Enum/EnumCaseObjectType.php
staabm: it also works. it makes the repro 0,2-0,3 seconds slower though

ondrejmirtes:
FYI I have a local branch where the reproduction takes zero zero nothing seconds, I "just" need to make it fully work. So any further optimizations from you that shave off percents from the time are in the negligible returns territory. Stay tuned! :)

staabm:
Thats great news 💪

src/Type/Enum/EnumCaseObjectType.php
staabm: moved the cache into ObjectType.

feel free to close this PR, in case it no longer serves a purpose.

ondrejmirtes:
@staabm I'm going to test it after my change. The previous PRs made sense on their own, but here we are adding caches (which means more memory consumed) so let's wait if it's still worth it.

PR #3220, opened 2024-07-10
InvalidKeyInArray: report float deprecation since PHP 8.1

janedbal:


janedbal:
Seeing the integration test failures, this is often triggered when numeric operation is performed with mixed. E.g. `$array[$mixed - 1]` as `$mixed - 1` is `float|int`.

staabm:
would it help if `$mixed - 1` would get a benevolent `(int|float)`? I think we did something similar recently

janedbal:
Saying **often** is maybe too abmitious, I checked few :)

----

It would be great if integration tests used `editorUrlTitle` so that you can easily click-to the GitHub source.

staabm:
the error at https://github.com/pmmp/PocketMine-MP/blob/8dedbb747108c45b7534b8ac6f05e8464ee8eb22/src/crash/CrashDump.php#L229-L234 looks like it might be useful handle `IntegerRangeType` or `getFiniteTypes` to get arround the false positive

-----

but maybe its just https://github.com/phpstan/phpstan-src/pull/3220#issuecomment-2220631819


----

alternatively you could consider not reporting the "maybe" error for floats because of the too high false-positive rate.


lets see what ondrey thinks about it

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -54,15 +56,34 @@ public function processNode(Node $node, Scope $scope): array
...
+		$isOk = $isSuperType->yes() || ($isSuperType->maybe() && !$this->reportMaybes);
+		if (!$isOk) {
+			return [
+				RuleErrorBuilder::message(
+					sprintf('%s array key type %s.', $isSuperType->no() ? 'Invalid' : 'Possibly invalid', $dimensionType->describe(VerbosityLevel::typeOnly())),
+				)->identifier('offsetAccess.invalidOffset')->build(),
+			];
+		}
+
+		if ($this->phpVersion->getVersionId() >= 80100) {
```
ondrejmirtes, line 68: Instead of condition like this, we need a new method on PhpVersion that would be used here.

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -54,15 +56,34 @@ public function processNode(Node $node, Scope $scope): array
...
+			$isFloat = $dimensionType->isFloat();
+
+			if ($isFloat->yes()) {
+				return [
+					RuleErrorBuilder::message(
+						'Float used as array key, this emits deprecation notice.',
```
ondrejmirtes, line 74: Message format to be used: `Deprecated in PHP 8.1: Float can no longer be used as array key, %s given.`

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -54,15 +56,34 @@ public function processNode(Node $node, Scope $scope): array
...
+					)->identifier('array.invalidOffset')->build(),
```
ondrejmirtes, line 75: `array.floatKey` might be better?

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -54,15 +56,34 @@ public function processNode(Node $node, Scope $scope): array
...
+				RuleErrorBuilder::message(
+					sprintf('%s array key type %s.', $isSuperType->no() ? 'Invalid' : 'Possibly invalid', $dimensionType->describe(VerbosityLevel::typeOnly())),
+				)->identifier('offsetAccess.invalidOffset')->build(),
+			];
+		}
+
+		if ($this->phpVersion->getVersionId() >= 80100) {
+			$isFloat = $dimensionType->isFloat();
+
+			if ($isFloat->yes()) {
```
ondrejmirtes, line 71: We need to use RuleLevelHelper::findTypeToCheck here, it's going to help us on level 9 and also with benevolent union types.

Case in point: `$mixed - 1` is already benevolent union type so this problem would already be solved: https://phpstan.org/r/5af2741f-b958-42f7-9248-435416d1b584

src/Rules/Arrays/InvalidKeyInArrayItemRule.php
```diff
@@ -47,6 +51,24 @@ public function processNode(Node $node, Scope $scope): array
 			];
 		}
 
+		if ($this->phpVersion->getVersionId() >= 80100) {
```
ondrejmirtes, line 54: All the same points as in the other place :)

ondrejmirtes (changes requested):

ondrejmirtes:
We should also have a test for this in LevelsIntegrationTest. The best way to develop that is to comment out all the existing "topics", and add a new one. Write a few code samples in there and run the test. It will (re)generate the current levels-snapshot JSON.

The behaviour I'd like this to have:

* float always reported since the level where the rule is first registered
* Partially valid type like `float|int` or `float|string` or even `float|null` to be reported on level 7
* Everything still reported on level 8
* `mixed` reported on level 9 (but `mixed~float` should not be reported at all)

The thing about `float|null` on level 7 vs. level 8 is a bit non-standard, because usually it's the other way around. To be honest I'm not sure how to write the `$unionTypeCriteriaCallback` callback here, that's why we need the tests :)

janedbal:
Altough I agree with your points, I'm not sure I'll invest the time into all that. 

Mostly because the core issue is covered in our (even stricter) rule: https://github.com/shipmonk-rnd/phpstan-rules/pull/254

VincentLanglet:
Hi @janedbal, float are now reported on PHP8.1+ with the PR https://github.com/phpstan/phpstan-src/pull/4406 so I think this one can be closed (?)

PR #3221, opened 2024-07-10
Fix new headers functions in php 8.4

Seldaek:
Looks like I was drunk when I sent https://github.com/phpstan/phpstan-src/pull/3093 and everyone just trusted me blindly 😅 

Proof this time https://github.com/php/php-src/blob/29f98e748568ebd66aaae061c0dcefbba92ca058/ext/standard/basic_functions.stub.php#L3036-L3038

### DIFF
--- a/resources/functionMap_php84delta.php
+++ b/resources/functionMap_php84delta.php
@@ -15,8 +15,8 @@
  */
 return [
 	'new' => [
-		'http_get_last_response_header ' => ['list<string>'],
-		'http_clear_last_response_header' => ['void'],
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thanks! I'll start working on PHP 8.4 support after the feature freeze. Some stuff will work and be well understood right away, but new syntax will have to wait until PHPStan 2.0 will PHP-Parser 5 is released later this year.

Seldaek:
That's fine by me.. for new syntax I'll have to wait 10 years or something 😆

PR #3223, opened 2024-07-11
Fix return type of round/ceil/floor in non strict type environment

BackEndTea:
In non declare strict types the round functions are able to handle numeric strings, null and bools.

See also: https://3v4l.org/5io0n

And issue https://github.com/phpstan/phpstan/issues/11319


ondrejmirtes:
Thank you!

BackEndTea:
Thanks for the merge @ondrejmirtes !

PR #3224, opened 2024-07-11
ResultCacheManager: output restore time in debug mode

janedbal:
I was recently dealing with huge result cache which took dozen of seconds to restore.

### DIFF
--- a/src/Analyser/ResultCache/ResultCacheManager.php
+++ b/src/Analyser/ResultCache/ResultCacheManager.php
@@ -33,6 +33,8 @@ use function implode;
 use function is_array;
 use function is_file;
 use function ksort;
+use function microtime;
+use function round;
 use function sha1_file;
 use function sort;
 use function sprintf;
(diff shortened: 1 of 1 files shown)


src/Analyser/ResultCache/ResultCacheManager.php
```diff
@@ -285,8 +288,10 @@ public function restore(array $allAnalysedFiles, bool $debug, bool $onlyFiles, ?
 		$filesToAnalyseCount = count($filesToAnalyse);
 
 		if ($output->isDebug()) {
+			$elapsed = microtime(true) - $startTime;
 			$output->writeLineFormatted(sprintf(
-				'Result cache restored. %d %s will be reanalysed.',
+				'Result cache restored in %f seconds. %d %s will be reanalysed.',
```
ondrejmirtes: Please only output the variant with seconds if it took more than 5 seconds. Otherwise it's distracting, thanks.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3225, opened 2024-07-11
Report useless return values of function calls

staabm:
closes https://github.com/phpstan/phpstan/issues/11320

### DIFF
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -70,6 +70,7 @@ rules:
 	- PHPStan\Rules\Functions\DefineParametersRule
 	- PHPStan\Rules\Functions\ExistingClassesInArrowFunctionTypehintsRule
 	- PHPStan\Rules\Functions\CallToFunctionParametersRule
+	- PHPStan\Rules\Functions\UselessFunctionReturnValueRule
 	- PHPStan\Rules\Functions\ExistingClassesInClosureTypehintsRule
 	- PHPStan\Rules\Functions\ExistingClassesInTypehintsRule
 	- PHPStan\Rules\Functions\FunctionAttributesRule
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -24,6 +24,8 @@ conditionalTags:
 		phpstan.rules.rule: %featureToggles.missingMagicSerializationRule%
 	PHPStan\Rules\Constants\MagicConstantContextRule:
 		phpstan.rules.rule: %featureToggles.magicConstantOutOfContext%
+	PHPStan\Rules\Functions\UselessFunctionReturnValueRule:
+		phpstan.rules.rule: %featureToggles.uselessReturnValue%
 
 rules:
 	- PHPStan\Rules\Api\ApiInstantiationRule
@@ -70,7 +72,6 @@ rules:
 	- PHPStan\Rules\Functions\DefineParametersRule
 	- PHPStan\Rules\Functions\ExistingClassesInArrowFunctionTypehintsRule
 	- PHPStan\Rules\Functions\CallToFunctionParametersRule
-	- PHPStan\Rules\Functions\UselessFunctionReturnValueRule
 	- PHPStan\Rules\Functions\ExistingClassesInClosureTypehintsRule
 	- PHPStan\Rules\Functions\ExistingClassesInTypehintsRule
 	- PHPStan\Rules\Functions\FunctionAttributesRule
@@ -294,3 +295,6 @@ services:
 
 	-
 		class: PHPStan\Rules\Constants\MagicConstantContextRule
+
+	-
+		class: PHPStan\Rules\Functions\UselessFunctionReturnValueRule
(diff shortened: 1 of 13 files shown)


src/Rules/FunctionCallParametersCheck.php
```diff
@@ -244,6 +246,30 @@ public function check(
 				->build();
 		}
 
+		if (
```
staabm: it could also make sense to move this code directly into `src/Rules/Functions/CallToFunctionParametersRule.php` 

----

in case you want this error behind a feature-flag we could also create a separate rule. if so, which level?

ondrejmirtes (changes requested):
Separate rule please. Same level the `void` usage is reported (not sure from the top of my head).

staabm:
> Same level the `void` usage is reported (not sure from the top of my head).

its a level 0 error: https://phpstan.org/r/42b76642-82b2-422a-bd7d-f3678eb6e1b0

src/Rules/Functions/UselessFunctionReturnValueRule.php
```diff
@@ -0,0 +1,74 @@
...
+			$funcCall,
+		);
+
+		if ($reorderedFuncCall === null) {
+			return [];
+		}
+		$reorderedArgs = $reorderedFuncCall->getArgs();
+
+		if (count($reorderedArgs) === 1 || (count($reorderedArgs) >= 2 && $scope->getType($reorderedArgs[1]->value)->isFalse()->yes())) {
+			return [RuleErrorBuilder::message('Return value of call to function ' . $functionReflection->getName() . ' is useless.')
```
ondrejmirtes: I like this, I'd just choose a different error message. Something like:

> `Return value of function print_r() is always true and the result is printed instead of being returned. Pass in true as parameter #%d $%s to return the output instead.`

conf/config.level0.neon
```diff
@@ -70,6 +70,7 @@ rules:
 	- PHPStan\Rules\Functions\DefineParametersRule
 	- PHPStan\Rules\Functions\ExistingClassesInArrowFunctionTypehintsRule
 	- PHPStan\Rules\Functions\CallToFunctionParametersRule
+	- PHPStan\Rules\Functions\UselessFunctionReturnValueRule
```
ondrejmirtes: Needs to be bleeding edge.

ondrejmirtes (changes requested):

staabm:
I have no idea why the PHP7.2 build fails with a "risky error". maybe its related that I ran the levels-tests on a windows machine.

tests/PHPStan/Rules/Functions/UselessFunctionReturnValueRuleTest.php
```diff
@@ -0,0 +1,54 @@
...
+				'Return value of function print_r is always true and the result is printed instead of being returned. Pass in true as parameter #2 $return to return the output instead.',
+				64,
+			],
+		]);
+	}
+
+	public function testUselessReturnValuePhp8(): void
+	{
+		if (PHP_VERSION_ID < 80000) {
+			return;
```
ondrejmirtes: This isn't how to make test for `>= 80000` only. You need to call markTestSkipped. See many examples in other tests. This is why it's marked as risky.

src/Rules/Functions/UselessFunctionReturnValueRule.php
```diff
@@ -0,0 +1,82 @@
...
+
+		if ($reorderedFuncCall === null) {
+			return [];
+		}
+		$reorderedArgs = $reorderedFuncCall->getArgs();
+
+		if (count($reorderedArgs) === 1 || (count($reorderedArgs) >= 2 && $scope->getType($reorderedArgs[1]->value)->isFalse()->yes())) {
+			return [RuleErrorBuilder::message(
+				sprintf(
+					'Return value of function %s is always true and the result is printed instead of being returned. Pass in true as parameter #%d $%s to return the output instead.',
```
ondrejmirtes: Please put `()` after `%s`. So that it's `print_r()` and not just `print_r`.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Functions/UselessFunctionReturnValueRuleTest.php
staabm: ouch. great catch, thanks.

staabm:
in phpstorm stubs I found another function alias worth reporting: [show_source()](https://www.php.net/show_source)

staabm:
just realized the [PrestaShop/PrestaShop shows 5 real world errors](https://github.com/phpstan/phpstan-src/actions/runs/9908035427/job/27373070696?pr=3225) this rule is targetting, simlar to the one of my initial report 💪

ondrejmirtes:
Nice! I really like this addition.

ondrejmirtes:
Thank you.

PR #3227, opened 2024-07-11
Fix statement analysis after early-terminating statements

takaram:
Closes phpstan/phpstan#11179

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -339,6 +339,10 @@ class NodeScopeResolver
 			|| $parentNode instanceof Node\Stmt\ClassMethod
 			|| $parentNode instanceof Expr\Closure;
 		foreach ($stmts as $i => $stmt) {
+			if ($alreadyTerminated && !($stmt instanceof Node\Stmt\Function_ || $stmt instanceof Node\Stmt\ClassLike)) {
+				continue;
+			}
+
 			$isLast = $i === $stmtCount - 1;
 			$statementResult = $this->processStmtNode(
 				$stmt,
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/Debug/DumpTypeRuleTest.php
```diff
@@ -82,4 +82,24 @@ public function testBug10377(): void
 		]);
 	}
 
+	public function testBug11179(): void
+	{
+		$this->analyse([__DIR__ . '/../DeadCode/data/bug-11179.php'], [
```
ondrejmirtes, line 87: Please also add a test into the "Unreachable statement - code above always terminates." rule.

src/Analyser/NodeScopeResolver.php
```diff
@@ -370,16 +378,16 @@ public function processStmtNodes(
...
 
 			$alreadyTerminated = true;
 			$nextStmt = $this->getFirstUnreachableNode(array_slice($stmts, $i + 1), $parentNode instanceof Node\Stmt\Namespace_);
-			if ($nextStmt !== null) {
-				$nodeCallback(new UnreachableStatementNode($nextStmt), $scope);
+			if ($nextStmt === null) {
+				continue;
 			}
-			break;
+			$nodeCallback(new UnreachableStatementNode($nextStmt), $scope);
```
ondrejmirtes, line 390: Sorry, I don't understand these changes. So here we're going to invoke `UnreachableStatementNode` but then we're still going to analyse the next statement if it's not Function_ or ClassLike? I don't get it.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Debug/DumpTypeRuleTest.php
takaram: It's in `UnreachableStatementRuleTest`

src/Analyser/NodeScopeResolver.php
takaram: Given this code:
```php
<?php
namespace Foo;

doSomething();
exit();
foo();
bar();

function doSomething(): void
{
  // ...
}
```

Here `foo();` will be reported as an unreachable statement.
Before this PR, PHPStan stops analyzing statements after `foo()` because analyzing `bar()` is no longer useful.
However, we still need to analyze the body of `doSomething()` actually.

So PHPStan should continue analysis after an unreachable statement, but the target is different;
- Before unreachable statements (`$alreadyTerminated = false`)
  - analyze all statements
- After unreachable statements (`$alreadyTerminated = true`)
  - analyze only functions and class-likes

staabm: what about methods, closures, arrow-functions?

ondrejmirtes: @staabm 1) I don't think you can put `exit` in-between methods in a class
2) I don't think closures after `exit` are executed.

ondrejmirtes: @takaram Thank you, makes total sense!

ondrejmirtes:
Thank you!

PR #3230, opened 2024-07-12
Add hint as to what might be wrong when invalid identifiers are used

Seldaek:
Went on a wild goose hunt to figure out if I needed to register my new identifier somewhere to make it valid before ultimately realizing I used `foo_bar` and that's not legal 🤦🏻‍♂️


ondrejmirtes:
Thank you!

Underscore would probably be harmless but we kind of already settled on camelCaps.camelCaps 😊

Seldaek:
Yeah it's totally fine, just wasn't clear at all from the error.

PR #3231, opened 2024-07-12
Feature check mixed in binary operator

schlndh:
The goal of this PR is to report mixed in binary operators. There's still a few other places where mixed is not properly reported (https://phpstan.org/r/c46a6ace-0135-46c0-a860-dd9d9b77b4ff), but those can be dealt with later.

Fixes: https://github.com/phpstan/phpstan/issues/7538
Fixes: https://github.com/phpstan/phpstan/issues/10440

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -856,6 +856,10 @@ class InitializerExprTypeResolver
 			return $this->getNeverType($leftType, $rightType);
 		}
 
+		if ($leftType->toNumber() instanceof ErrorType || $rightType->toNumber() instanceof ErrorType) {
+			return new ErrorType();
+		}
+
 		$leftTypes = $leftType->getConstantScalarTypes();
 		$rightTypes = $rightType->getConstantScalarTypes();
 		$leftTypesCount = count($leftTypes);
--- a/src/Rules/LazyRegistry.php
+++ b/src/Rules/LazyRegistry.php
@@ -32,7 +32,17 @@ class LazyRegistry implements Registry
 	public function getRules(string $nodeType): array
 	{
 		if (!isset($this->cache[$nodeType])) {
-			$parentNodeTypes = [$nodeType] + class_parents($nodeType) + class_implements($nodeType);
+			$parents = class_parents($nodeType);
+			if ($parents === false) {
+				$parents = [];
+			}
+
+			$implements = class_implements($nodeType);
+			if ($implements === false) {
+				$implements = [];
+			}
+
+			$parentNodeTypes = [$nodeType] + $parents + $implements;
 
 			$rules = [];
 			$rulesFromContainer = $this->getRulesFromContainer();
--- a/src/Rules/LazyRegistry.php
+++ b/src/Rules/LazyRegistry.php
@@ -32,17 +32,7 @@ class LazyRegistry implements Registry
 	public function getRules(string $nodeType): array
 	{
 		if (!isset($this->cache[$nodeType])) {
-			$parents = class_parents($nodeType);
-			if ($parents === false) {
-				$parents = [];
-			}
-
-			$implements = class_implements($nodeType);
-			if ($implements === false) {
-				$implements = [];
-			}
-
-			$parentNodeTypes = [$nodeType] + $parents + $implements;
+			$parentNodeTypes = [$nodeType] + class_parents($nodeType) + class_implements($nodeType);
 
 			$rules = [];
 			$rulesFromContainer = $this->getRulesFromContainer();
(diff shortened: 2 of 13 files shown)


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -856,6 +856,10 @@ public function getModType(Expr $left, Expr $right, callable $getTypeCallback):
 			return $this->getNeverType($leftType, $rightType);
 		}
 
+		if ($leftType->toNumber() instanceof ErrorType || $rightType->toNumber() instanceof ErrorType) {
```
schlndh, line 859: I'm not sure whether this should only be enabled in bleeding edge. On the one hand, I don't see any other case in this class where `return new ErrorType` would be conditioned by bleeding edge. On the other hand, this can definitely result in new errors being reported without bleeding edge. E.g. `5 % []` now reports `Binary operation "%" between 5 and array{} results in an error`.

tests/PHPStan/Rules/Operators/data/invalid-binary-mixed.php
```diff
@@ -0,0 +1,157 @@
+<?php declare(strict_types = 1);
+
+namespace InvalidBinaryOperatorWithMixed;
+
+/**
+ * @template T
+ * @param T $a
+ */
+function genericMixed(mixed $a): void
```
schlndh, line 9: Some operators are missing from this test:

- `instanceof` is not a binary operator in php-parser.
- `&&`, `||`, `and`, `or` and `xor` - everything can be converted to boolean. And they're already covered by strict-rules.
- `??` is OK to use with mixed.
- Comparison operators are handled by `InvalidComparisonOperationRule`.

src/Rules/LazyRegistry.php
```diff
@@ -32,7 +32,17 @@ public function __construct(private Container $container)
 	public function getRules(string $nodeType): array
 	{
 		if (!isset($this->cache[$nodeType])) {
-			$parentNodeTypes = [$nodeType] + class_parents($nodeType) + class_implements($nodeType);
+			$parents = class_parents($nodeType);
```
ondrejmirtes: Why is this piece of code changed? ClassImplementsFunctionReturnTypeExtension makes sure the returned value does not include `false`.

ondrejmirtes (changes requested):

src/Rules/LazyRegistry.php
schlndh: It got reported by `make phpstan`. I'll take another look.

schlndh: There are two errors reported:

```
Binary operation "+" between non-empty-array and (array<string, class-string>|false) results in an error.
Binary operation "+" between array{class-string<TNodeType of PhpParser\Node>} and (array<string, class-string>|false) results in an error.
```

The problem is this (in `InvalidBinaryOperationRule`):
`$callback = static fn (Type $type): bool => !$type->toNumber() instanceof ErrorType;`

`ClassImplementsFunctionReturnTypeExtension` returns benevolent union of array and false. And the callback only keeps false. So the rule checks `array + false`, which is an error.

I'll try to fix it.

src/Rules/Operators/InvalidBinaryOperationRule.php
```diff
@@ -44,81 +45,83 @@ public function processNode(Node $node, Scope $scope): array
...
+			$left = $node->left;
+			$right = $node->right;
+			$newNode->left = $leftVariable;
+			$newNode->right = $rightVariable;
+		}
+
+		if ($node instanceof Node\Expr\AssignOp\Concat || $node instanceof Node\Expr\BinaryOp\Concat) {
+			$callback = static fn (Type $type): bool => !$type->toString() instanceof ErrorType;
+		} elseif ($node instanceof Node\Expr\AssignOp\Plus || $node instanceof Node\Expr\BinaryOp\Plus) {
+			$callback = static fn (Type $type): bool => !$type->toNumber() instanceof ErrorType || $type->isArray()->yes();
```
schlndh, line 77: This is new, otherwise the code in this file should be the same with the exception of the early exit and not requiring `$scope->getType($node) instanceof ErrorType` anymore in bleeding edge.

staabm:
the error in shopsys/shopsys on line

https://github.com/shopsys/shopsys/blob/4a0cce895273dba7021736d0b3984b454f77e70e/packages/framework/src/Component/Cron/CronTimeResolver.php#L40-L41

looks like it could be fixed with a improved `preg_match` inference. I will have a look into simple numeric-string cases

src/Reflection/InitializerExprTypeResolver.php
schlndh: The `$int % $string` use-case is even more compelling example of why it should apply only to bleeding edge.

ondrejmirtes:
Thank you!

PR #3232, opened 2024-07-13
Improved sprintf() inference

staabm:
another small improvement while thinking thru more details of https://github.com/phpstan/phpstan-src/pull/3192

it fixes a bug regressed with https://github.com/phpstan/phpstan-src/pull/3167

### DIFF
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -81,7 +81,10 @@ class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturn
 				// of stringy type, then the return value will be of the same type
 				$checkArgType = $scope->getType($args[$checkArg]->value);
 
-				if ($matches[2] === 's' && $checkArgType->isScalar()->yes()) {
+				if ($matches[2] === 's'
+					&& $checkArgType->isScalar()->yes()
+					&& $checkArgType->getFiniteTypes() === []
+				) {
 					$singlePlaceholderEarlyReturn = $checkArgType->toString();
 				} elseif ($matches[2] !== 's') {
 					$singlePlaceholderEarlyReturn = new IntersectionType([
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/nsrt/bug-11201.php
```diff
@@ -41,3 +45,12 @@ function returnsJustString(): string
...
 assertType('non-falsy-string', $s);
+
+$s = sprintf('%20s', 'abc');
+assertType("'                 abc'", $s);
+
+$s = sprintf('%20s', true);
+assertType("'                   1'", $s);
+
+$s = sprintf('%20s', returnsBool());
+assertType("non-falsy-string", $s);
```
staabm, line 56: regressions from https://github.com/phpstan/phpstan-src/pull/3167 which made `sprintf` 1:1 return constant values from arguments (bypassing sprintf formatting rules)

staabm: //cc @pilif

tests/PHPStan/Analyser/nsrt/bug-7387.php
```diff
@@ -45,9 +45,15 @@ public function specifiers(int $i) {
...
+	/**
+	 * @param positive-int $posInt
+	 * @param negative-int $negInt
+	 */
+	public function positionalArgs($mixed, int $i, float $f, string $s, int $posInt, int $negInt) {
 		// https://3v4l.org/vVL0c
-		assertType('non-falsy-string', sprintf('%2$14s', $mixed, $i));
+		assertType('numeric-string', sprintf('%2$14s', $mixed, $i));
+		assertType('non-falsy-string&numeric-string', sprintf('%2$14s', $mixed, $posInt));
+		assertType('non-falsy-string&numeric-string', sprintf('%2$14s', $mixed, $negInt));
```
staabm, line 58: improvements because of the added `toString()` handling

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -115,19 +119,19 @@ public function getTypeFromFunctionCall(
 			$returnType = new StringType();
 		}
 
-		return $this->getConstantType($args, $returnType, $functionReflection, $scope);
```
staabm, line 118: instead of building the general return type first and pass it as a default-return into `getConstantType()` we now invoke `getConstantType()` first and build the general return type only if we were not able to infer a constant type in the first place

ondrejmirtes:
Thank you!

PR #3233, opened 2024-07-13
RegexArrayShapeMatcher - Support preg_quote()'d patterns

staabm:
closes https://github.com/phpstan/phpstan/issues/11332 - implements the idea described in the issue

since the `preg_quote` detection requires access to the `Expr` of the pattern, I had to add a new public api method, which should be used instead of the previous `matchType` method.

I marked the "old" method as `@deprecated` and suggest the newly added method instead.

### DIFF
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -7,6 +7,9 @@ use Hoa\Compiler\Llk\Parser;
 use Hoa\Compiler\Llk\TreeNode;
 use Hoa\Exception\Exception;
 use Hoa\File\Read;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Name;
+use PHPStan\Analyser\Scope;
 use PHPStan\Php\PhpVersion;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Constant\ConstantArrayType;
@@ -45,7 +48,20 @@ final class RegexArrayShapeMatcher
 	{
 	}
 
+	public function matchExpr(Expr $patternExpr, ?Type $flagsType, TrinaryLogic $wasMatched, Scope $scope): ?Type
+	{
+		return $this->matchPatternType($this->getPatternType($patternExpr, $scope), $flagsType, $wasMatched);
+	}
+
+	/**
+	 * @deprecated use matchExpr() instead for a more precise result
+	 */
 	public function matchType(Type $patternType, ?Type $flagsType, TrinaryLogic $wasMatched): ?Type
+	{
+		return $this->matchPatternType($patternType, $flagsType, $wasMatched);
+	}
+
+	private function matchPatternType(Type $patternType, ?Type $flagsType, TrinaryLogic $wasMatched): ?Type
 	{
 		if ($wasMatched->no()) {
 			return new ConstantArrayType([], []);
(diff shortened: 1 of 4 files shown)


src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -45,7 +48,20 @@ public function __construct(
 	{
 	}
 
+	public function matchExpr(Expr $patternExpr, ?Type $flagsType, TrinaryLogic $wasMatched, Scope $scope): ?Type
```
staabm, line 51: maybe it would even make sense to use `Expr` also for `$flags`, just in case for more flexibility for future enhancements?

staabm:
//cc @Seldaek

src/Type/Php/RegexArrayShapeMatcher.php
staabm: hmm.. having `$flags` as `Expr` would make it harder for composer/pcre to [pass the internal flags](https://github.com/composer/pcre/pull/29/files#diff-f885e19d2e2b46c78de6962b8e1c48ad6c556f056da18e02ecf4cdb9713ff428R33). 

It seems we should leave it as is in this PR

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -484,4 +500,49 @@ private function walkRegexAst(
...
+	}
+
+	private function resolvePatternConcat(Expr\BinaryOp\Concat $concat, Scope $scope): Type
+	{
+		if (
+			$concat->left instanceof Expr\FuncCall
+			&& $concat->left->name instanceof Name
+			&& $concat->left->name->toLowerString() === 'preg_quote'
+		) {
+			$left = new ConstantStringType('');
```
Seldaek, line 526: Is there any way to make this a non empty string if we know the preg quote is called with a non empty string?

Seldaek (commented):
Very cool that this can be done relatively easily :)

src/Type/Php/RegexArrayShapeMatcher.php
staabm: Yeah thats doable. I left it out because atm we don't differentiate different string types. Its a long way to go and I want to look into preg_match_all before doing more precise types

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -484,4 +500,49 @@ private function walkRegexAst(
 		}
 	}
 
+	private function getPatternType(Expr $patternExpr, Scope $scope): Type
+	{
+		if ($patternExpr instanceof Expr\BinaryOp\Concat) {
+			return $this->resolvePatternConcat($patternExpr, $scope);
```
ondrejmirtes, line 506: I don't get it, what's happening here? Shouldn't this be done instead by improving `MutatingScope::resolveType`? By introducing dynamic return type extension for preg_quote or something?

ondrejmirtes (changes requested):

src/Type/Php/RegexArrayShapeMatcher.php
Seldaek: I'll try to explain as I feel this is my fault :D

I think you can't reasonably treat preg quote's return value as a constant string (empty or "x") in any other context than here when trying to get a constant pattern string so that you have a regex to parse.

As such it feels appropriate to me to hack this in here but maybe there are better ways 🤷🏻‍♂️

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -484,4 +500,49 @@ private function walkRegexAst(
...
+		} else {
+			$left = $scope->getType($concat->left);
+		}
+
+		if (
+			$concat->right instanceof Expr\FuncCall
+			&& $concat->right->name instanceof Name
+			&& $concat->right->name->toLowerString() === 'preg_quote'
+		) {
+			$right = new ConstantStringType('');
```
Seldaek, line 538: Another thing.. If you do smth like https://3v4l.org/1J7nFF you see the first call triggers a warning but this code will assume it's fine even tho it is a detectable mistake. If you're evaluating the pattern anyway, you could check the first character and try to run preq_quote on it. If it doesn't come out escaped then any preg quote call that is inline in the concatenated string forming the pattern should have that character passed in as second argument to ensure the regex cannot be broken.

I hope I'm making sense @staabm .. Could be another PR tho for sure as it's smth for a new rule I guess. It sucks a bit that we can't warn from here tho because that would warn also for free in composer/pcre and others integrating with it.

src/Type/Php/RegexArrayShapeMatcher.php
staabm: Its pretty unlikely that we know the args to preg_quote() at static analysis time in a real world use-case. 

Please also look at the added test-cases. These wouldn't work with a return type extension, as the extension couldn't produce a constant-string for these cases. 

So while adding a preg_quote return type extension could make sense in the future, it would not help with some of the examples at hand. The ArrayShapeMatcher just can ignore the preg_quote() parts and still analyse the shape, as long as all other parts of a concatenation are constant. 

This assumption only works for the ArrayShapeMatcher therefore its not implemented for the common case in the Scope

ondrejmirtes: Alright, in that case the code needs some comments. I did not understand at all why you are returning an empty string for a preg_quote call.

staabm: done 👍

ondrejmirtes:
Thank you.

src/Type/Php/RegexArrayShapeMatcher.php
staabm: I agree this sounds useful. please create a new issue with a reproducer

PR #3234, opened 2024-07-14
RegexArrayShapeMatcher - Fix PREG_UNMATCHED_AS_NULL with optional leading groups

staabm:
closes https://github.com/phpstan/phpstan/issues/11331

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-11311.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-11311.php
@@ -29,3 +29,40 @@ function unmatchedAsNullWithOptionalGroup(string $s): void {
 	assertType('array{}|array{string, string|null}', $matches);
 }
 
+function bug11331a(string $url):void {
+	// group a is actually optional as the entire (?:...) around it is optional
+	if (preg_match('{^
+	(?:
+		(?<a>.+)
+	)?
+	(?<b>.+)}mix', $url, $matches, PREG_UNMATCHED_AS_NULL)) {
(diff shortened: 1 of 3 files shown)


staabm:
//cc @Seldaek

tests/PHPStan/Analyser/nsrt/bug-11311.php
```diff
@@ -29,3 +29,40 @@ function unmatchedAsNullWithOptionalGroup(string $s): void {
...
+}
+
+function bug11331c(string $url):void {
+	if (preg_match('{^
+	(?:
+		(?:https?|git)://([^/]+)/ (?# group 1 here can be null if group 2 matches)
+		|                         (?# the alternation making it so that only either should match)
+		git@([^:]+):/?            (?# group 2 here can be null if group 1 matches)
+	)
+	(?# removing what follows makes the two first groups nullable, although it then has group 2 unsettable which looks buggy too as PREG_UNMATCHED_AS_NULL is present)
```
staabm: @Seldaek not sure what this means. could you come up with a concrete code sample and assertions for it?

(sounds unrelated to the other bugs and might be worth a separate report)

Seldaek: I just meant if you delete the lines of regex below this comment, then the type of the groups 1 and 2 become correct.

staabm: I see thanks. I will remove the comment from the test so it not confuses others :)

Seldaek (approved):

ondrejmirtes:
Thank you!

PR #3239, opened 2024-07-15
Print analysis time with -vvv

janedbal:


src/Command/InceptionResult.php
```diff
@@ -84,7 +86,7 @@ public function getGenerateBaselineFile(): ?string
 		return $this->generateBaselineFile;
 	}
 
-	public function handleReturn(int $exitCode, ?int $peakMemoryUsageBytes): int
+	public function handleReturn(int $exitCode, ?int $peakMemoryUsageBytes, ?float $analysisStartTime): int
```
ondrejmirtes: By looking at the diff it seems to me that `null` is never passed here so we can just have `float`, right?

ondrejmirtes (changes requested):

src/Command/InceptionResult.php
janedbal: Yeah, lets get rid of that `?`

ondrejmirtes:
Thank you.

PR #3240, opened 2024-07-15
InternalError is part of `@api`

janedbal:
My custom formatter is outputting internal errors. I'm getting `Calling PHPStan\Analyser\InternalError::getMessage() is not covered by backward compatibility promise.`. I believe this should be part of public api.


ondrejmirtes:
Yes, thank you.

PR #3241, opened 2024-07-15
Support posix character classes in preg_match() inference

staabm:
closes https://github.com/phpstan/phpstan/issues/11323

----

- checkout PR
- run `composer install` - so the patch is applied
- run the test with `vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php` (will test all preg_match() related tests)

----

for a faster debug loop, you might just put the test into a `test.php` file:
```php
<?php

function bug11323(string $s): void {
	if (preg_match('/([*|+?{}()]+)([^*|+[:digit:]?{}()]+)/', $s, $matches)) {
		assertType('array{string, string, string}', $matches);
	}
}
```

and run `php bin/phpstan analyze test.php --debug` on it 
(append an additional `--xdebug` in case you want to step-debug)

### DIFF
--- a/patches/Grammar.patch
+++ b/patches/Grammar.patch
@@ -1,8 +1,17 @@
 diff --git a/Grammar.pp b/Grammar.pp
-index 5157084..8384417 100644
+index 5157084..48959cb 100644
 --- a/Grammar.pp
 +++ b/Grammar.pp
-@@ -73,7 +73,7 @@
+@@ -50,6 +50,8 @@
+ %token  class_                   \[
+ %token _class                    \]
+ %token  range                    \-
++%token class_literal             [^\\\[\]-]
++%token posix_class               \[^?:[a-z]+:\]
+
+ // Internal options.
+ %token  internal_option          \(\?[\-+]?[imsx]\)
+@@ -73,7 +75,7 @@
  %token  co:comment               .*?(?=(?<!\\)\))
 
  // Capturing group.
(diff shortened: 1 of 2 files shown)


Seldaek:
Ok I'll try to debug this, thanks. Might not have time before later this week though.

Seldaek:
Just adding `%token  posix_class              \[^?:[a-z]+:\]` and not the class_literal line because that breaks everything, and then using `    ( <posix_class> | <class_> | range() | literal() | quantifier() | <alternation> | <capturing_> | <_capturing> )+ <range>?` to parse the class contents.. it gets almost to the end:

```
Unexpected token ")" (_capturing) at line 1 and column 34:
/([*|+?{}()]+)([^*|+[:digit:]?{}()]+)/
                                 ↑
```

Unfortunately I don't manage to make it capture the last `)`, no clue why it'd fail there..

And also this is super brittle, and I can break it by adding `[(?:]` for example or any other token to the character class, which then gets matched and results in an:

```
Unexpected token "(?:" (non_capturing_) at line 1 and column 10:
/([*|+?{}(?:)]+)([^*|+[:digit:]?{}()]+)/
         ↑
```

Seldaek:
@mvorisek are you familiar with this parser? Is there a syntax to turn this silly scanning-of-all-tokens off and tell it to not look at other tokens while it is in the middle of parsing a `class` section?

Seldaek:
I guess I'll  need to read up on https://github.com/hoaproject/Compiler - but no time for this anymore today ..

staabm:
here seems to be [something like a documentation](https://github.com/hoaproject/Compiler/blob/master/Documentation/En/Index.xyl)

mvorisek:
> @mvorisek are you familiar with this parser? Is there a syntax to turn this silly scanning-of-all-tokens off and tell it to not look at other tokens while it is in the middle of parsing a `class` section?

Of course `class_literal` needs to go before `class_`, more precisely `class_` needs to be removed - if that is not possible, the only option is to list "all" tokens in https://github.com/phpstan/phpstan/issues/11323#issuecomment-2225214957 and my following comment.

staabm:
as far as I understood the other parallel PRs, current progress happens in https://github.com/phpstan/phpstan-src/pull/3244 and this one here is obsolete. 

therefore closing.

PR #3242, opened 2024-07-15
Nicer analysis time output

janedbal:


staabm:
should these text also be "green" like in the other key-value pair sections above?

grafik

ondrejmirtes:
Thank you.

@staabm Not green because that's not part of the DiagnoseExtension. We can improve the formatting but should be done with overhaul of other outputs that's outside of DiagnoseExtension.

PR #3243, opened 2024-07-15
Ignore non-explicit NEVER in purity check

staabm:
closes https://github.com/phpstan/phpstan/issues/11207

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4623,7 +4623,13 @@ class NodeScopeResolver
 				$hasYield = $hasYield || $exprResult->hasYield();
 
 				if ($exprType->isCallable()->yes()) {
-					$acceptors = $exprType->getCallableParametersAcceptors($scope);
+					$acceptors = [];
+					if (
+						(!$exprType instanceof NeverType || $exprType->isExplicit())
+					) {
+						$acceptors = $exprType->getCallableParametersAcceptors($scope);
+					}
+
 					if (count($acceptors) === 1) {
 						$scope = $this->processImmediatelyCalledCallable($scope, $acceptors[0]->getInvalidateExpressions(), $acceptors[0]->getUsedVariables());
 						if ($callCallbackImmediately) {
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4622,10 +4622,7 @@ class NodeScopeResolver
 				$scope = $exprResult->getScope();
 				$hasYield = $hasYield || $exprResult->hasYield();
 
-				if (
-					$exprType->isCallable()->yes()
-					&& (!$exprType instanceof NeverType || $exprType->isExplicit())
-				) {
+				if ($exprType->isCallable()->yes()) {
 					$acceptors = $exprType->getCallableParametersAcceptors($scope);
 					if (count($acceptors) === 1) {
 						$scope = $this->processImmediatelyCalledCallable($scope, $acceptors[0]->getInvalidateExpressions(), $acceptors[0]->getUsedVariables());
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4330,7 +4330,13 @@ class NodeScopeResolver
 	{
 		$callableParameters = null;
 		if ($args !== null) {
-			$acceptors = $scope->getType($closureExpr)->getCallableParametersAcceptors($scope);
+			$closureType = $scope->getType($closureExpr);
+
+			if ($closureType->isCallable()->no()) {
+				return null;
+			}
+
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Pure/PureMethodRuleTest.php
```diff
@@ -167,4 +177,16 @@ public function testImpureAssignRef(): void
...
+	{
+		$this->treatPhpDocTypesAsCertain = false;
+		$this->analyse([__DIR__ . '/data/bug-11207.php'], []);
+	}
+
+	public function testBug11207b(): void
+	{
+		$this->treatPhpDocTypesAsCertain = true;
+		$this->analyse([__DIR__ . '/data/bug-11207.php'], []);
+	}
```
staabm: I think given example should not produce purity errors no matter `treatPhpDocTypesAsCertain`, as we are dealing with a non-explicit never.

therefore I added both tests, even though they actually have the same assertions. does this make sense?

src/Analyser/NodeScopeResolver.php
```diff
@@ -4622,7 +4622,10 @@ private function processArgs(
 				$scope = $exprResult->getScope();
 				$hasYield = $hasYield || $exprResult->hasYield();
 
-				if ($exprType->isCallable()->yes()) {
+				if (
+					$exprType->isCallable()->yes()
+					&& (!$exprType instanceof NeverType || $exprType->isExplicit())
```
staabm: a different way to fix this would be to make `NeverType` return a different acceptor on `getCallableParametersAcceptors` when non-explicit

ondrejmirtes: Yeah, that would make more sense to me. It's basically never going to be called so a special class instead of TrivialParametersAcceptor would be welcome. It's pure, and probably also returns NeverType?

ondrejmirtes (changes requested):

ondrejmirtes:
Well, or NeverType doesn't have to return yes for isCallable, I'm not sure why it does.

src/Type/NeverType.php
```diff
@@ -334,12 +333,12 @@ public function shuffleArray(): Type
...
 	public function isCallable(): TrinaryLogic
 	{
-		return TrinaryLogic::createYes();
+		return TrinaryLogic::createNo();
 	}
 
 	public function getCallableParametersAcceptors(ClassMemberAccessAnswerer $scope): array
 	{
-		return [new TrivialParametersAcceptor()];
+		return [];
```
ondrejmirtes: This should `throw new ShouldNotHappenException();` instead

tests/PHPStan/Rules/Pure/PureMethodRuleTest.php
```diff
@@ -167,4 +177,16 @@ public function testImpureAssignRef(): void
...
+	public function testBug11207a(): void
+	{
+		$this->treatPhpDocTypesAsCertain = false;
+		$this->analyse([__DIR__ . '/data/bug-11207.php'], []);
+	}
+
+	public function testBug11207b(): void
+	{
+		$this->treatPhpDocTypesAsCertain = true;
+		$this->analyse([__DIR__ . '/data/bug-11207.php'], []);
```
ondrejmirtes: This kind of pattern is usually done via a dataProvider.

ondrejmirtes (changes requested):

src/Type/NeverType.php
staabm: hmm it throws..

```
1) Error
PHPStan\ShouldNotHappenException: Internal error.
/Users/staabm/workspace/phpstan-src/src/Type/NeverType.php:341
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:4361
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:4107
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:4560
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:2320
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:4619
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:2320
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:770
/Users/staabm/workspace/phpstan-src/src/Analyser/NodeScopeResolver.php:304
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:95
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:154
/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:292
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:17

```

ondrejmirtes:
Thank you.

PR #3244, opened 2024-07-15
Fix regular expression parsing

Seldaek:
I moved the grammar file into the repo instead of the hoa regex one with patch, that makes it a lot easier to work with.

This PR fixes parsing of complex character classes which was really not well supported in hoa/regex, along with a bunch of other smaller issues.

### DIFF
--- a/resources/RegexGrammar.pp
+++ b/resources/RegexGrammar.pp
@@ -46,13 +46,23 @@
 %skip   nl                       \n
 
 // Character classes.
-%token  negative_class_          \[\^
-%token  class_                   \[
-%token _class                    \]
-%token  range                    \-
+%token  negative_class_          \[\^               -> class
+%token  class_                   \[                 -> class
+%token class:posix_class         \[:\^?[a-z]+:\]
+%token class:class_              \[
+%token class:_class_literal      (?<=[^\\]\[|[^\\]\[\^)\]
+%token class:_class              \]                 -> default
+%token class:range               \-
+%token class:escaped_end_class   \\\]
+// taken over from literals but class:character has \b support on top (backspace in character classes)
+%token class:character           \\([aefnrtb]|c[\x00-\x7f])
+%token class:dynamic_character   \\([0-7]{3}|x[0-9a-zA-Z]{2}|x{[0-9a-zA-Z]+})
+%token class:character_type      \\([CdDhHNRsSvVwWX]|[pP]{[^}]+})
+%token class:literal             \\.|.
 
 // Internal options.
-%token  internal_option          \(\?[\-+]?[imsx]\)
+// See https://www.regular-expressions.info/refmodifiers.html
+%token  internal_option          \(\?([imsxnJUX^]|xx)?-?([imsxnJUX^]|xx)\)
 
 // Lookahead and lookbehind assertions.
 %token  lookahead_               \(\?=
@@ -77,6 +87,7 @@
 %token  nc:_named_capturing      >                  -> default
 %token  nc:capturing_name        .+?(?=(?<!\\)>)
 %token  non_capturing_           \(\?:
+%token  non_capturing_internal_option \(\?([imsxnJUX^]|xx)?-?([imsxnJUX^]|xx):
 %token  non_capturing_reset_     \(\?\|
 %token  atomic_group_            \(\?>
 %token  capturing_               \(
@@ -168,7 +179,7 @@ quantifier:
         ::negative_class_:: #negativeclass
       | ::class_::
     )
-    ( <class_> | range() | literal() )+
+    ( <range> | <_class_literal> )? ( <posix_class> | <class_> | range() | literal() | <escaped_end_class> )* <range>?
     ::_class::
 
 #range:
@@ -183,15 +194,18 @@ simple:
   | (
         ::named_capturing_:: <capturing_name> ::_named_capturing:: #namedcapturing
       | ::non_capturing_:: #noncapturing
+      | non_capturing_internal_options() #noncapturing
       | ::non_capturing_reset_:: #noncapturingreset
       | ::atomic_group_:: #atomicgroup
       | ::capturing_::
     )
     alternation() ::_capturing::
 
+non_capturing_internal_options:
+    <non_capturing_internal_option>
+
 literal:
     <character>
-  | <range>
   | <dynamic_character>
   | <character_type>
   | <anchor>
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -7,6 +7,7 @@ use Hoa\Compiler\Llk\Parser;
 use Hoa\Compiler\Llk\TreeNode;
 use Hoa\Exception\Exception;
 use Hoa\File\Read;
+use Nette\Utils\RegexpException;
 use Nette\Utils\Strings;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Name;
@@ -31,7 +32,11 @@ use function count;
 use function in_array;
 use function is_int;
 use function is_string;
+use function rtrim;
 use function sscanf;
+use function str_replace;
+use function strlen;
+use function substr;
 use const PREG_OFFSET_CAPTURE;
 use const PREG_UNMATCHED_AS_NULL;
 
@@ -376,6 +381,13 @@ final class RegexArrayShapeMatcher
 		}
 
 		try {
+			Strings::match('', $regex);
+		} catch (RegexpException) {
+			// pattern is invalid, so let the RegularExpressionPatternRule report it
+			return null;
+		}
+
+		try {
 			$ast = self::$parser->parse($regex);
 		} catch (Exception) {
 			return null;
@@ -516,25 +528,37 @@ final class RegexArrayShapeMatcher
 		$lastChild = $node->getChild($node->getChildrenNumber() - 1);
 		$value = $lastChild->getValue();
 
-		if ($value['token'] === 'n_to_m') {
-			if (sscanf($value['value'], '{%d,%d}', $n, $m) !== 2 || !is_int($n) || !is_int($m)) {
+		// normalize away possessive and lazy quantifier-modifiers
+		$token = str_replace(['_possessive', '_lazy'], '', $value['token']);
+		$value = rtrim($value['value'], '+?');
+
+		if ($token === 'n_to_m') {
+			if (sscanf($value, '{%d,%d}', $n, $m) !== 2 || !is_int($n) || !is_int($m)) {
 				throw new ShouldNotHappenException();
 			}
 
 			$min = $n;
 			$max = $m;
-		} elseif ($value['token'] === 'exactly_n') {
-			if (sscanf($value['value'], '{%d}', $n) !== 1 || !is_int($n)) {
+		} elseif ($token === 'n_or_more') {
+			if (sscanf($value, '{%d,}', $n) !== 1 || !is_int($n)) {
+				throw new ShouldNotHappenException();
+			}
+
+			$min = $n;
+		} elseif ($token === 'exactly_n') {
+			if (sscanf($value, '{%d}', $n) !== 1 || !is_int($n)) {
 				throw new ShouldNotHappenException();
 			}
 
 			$min = $n;
 			$max = $n;
-		} elseif ($value['token'] === 'zero_or_one') {
+		} elseif ($token === 'zero_or_one') {
 			$min = 0;
 			$max = 1;
-		} elseif ($value['token'] === 'zero_or_more') {
+		} elseif ($token === 'zero_or_more') {
 			$min = 0;
+		} elseif ($token === 'one_or_more') {
+			$min = 1;
 		}
 
 		return [$min, $max];
@@ -591,20 +615,8 @@ final class RegexArrayShapeMatcher
 			if ($literalValue !== null) {
 				if (Strings::match($literalValue, '/^\d+$/') === null) {
 					$isNumeric = TrinaryLogic::createNo();
-				}
-
-				if (!$inOptionalQuantification) {
-					$isNonEmpty = TrinaryLogic::createYes();
-				}
-			}
-
-			if ($ast->getValueToken() === 'character_type') {
-				if ($ast->getValueValue() === '\d') {
-					if ($isNumeric->maybe()) {
-						$isNumeric = TrinaryLogic::createYes();
-					}
-				} else {
-					$isNumeric = TrinaryLogic::createNo();
+				} elseif ($isNumeric->maybe()) {
+					$isNumeric = TrinaryLogic::createYes();
 				}
 
 				if (!$inOptionalQuantification) {
@@ -624,7 +636,6 @@ final class RegexArrayShapeMatcher
 					}
 
 					if (Strings::match($literalValue, '/^\d+$/') === null) {
-						$allNumeric = false;
 						break;
 					}
 
@@ -653,13 +664,65 @@ final class RegexArrayShapeMatcher
 
 	private function getLiteralValue(TreeNode $node): ?string
 	{
-		if ($node->getId() === 'token' && $node->getValueToken() === 'literal') {
-			return $node->getValueValue();
+		if ($node->getId() !== 'token') {
+			return null;
 		}
 
-		// literal "-" outside of a character class like '~^((\\d{1,6})-)$~'
-		if ($node->getId() === 'token' && $node->getValueToken() === 'range') {
-			return $node->getValueValue();
+		// token is the token name from grammar without the namespace so literal and class:literal are both called literal here
+		$token = $node->getValueToken();
+		$value = $node->getValueValue();
+
+		if (in_array($token, ['literal', 'escaped_end_class'], true)) {
+			if (strlen($node->getValueValue()) > 1 && $value[0] === '\\') {
+				return substr($value, 1);
+			}
+
+			return $value;
+		}
+
+		// literal "-" in front/back of a character class like '[-a-z]' or '[abc-]', not forming a range
+		if ($token === 'range') {
+			return $value;
+		}
+
+		// literal "[" or "]" inside character classes '[[]' or '[]]'
+		if (in_array($token, ['class_', '_class_literal'], true)) {
+			return $value;
+		}
+
+		// character escape sequences, just return a fixed string
+		if (in_array($token, ['character', 'dynamic_character', 'character_type'], true)) {
+			if ($token === 'character_type' && $value === '\d') {
+				return '0';
+			}
+
+			return $value;
+		}
+
+		// [:digit:] and the like, more support coming later
+		if ($token === 'posix_class') {
+			if ($value === '[:digit:]') {
+				return '0';
+			}
+			if (in_array($value, ['[:alpha:]', '[:alnum:]', '[:upper:]', '[:lower:]', '[:word:]', '[:ascii:]', '[:print:]', '[:xdigit:]', '[:graph:]'], true)) {
+				return 'a';
+			}
+			if ($value === '[:blank:]') {
+				return " \t";
+			}
+			if ($value === '[:cntrl:]') {
+				return "\x00\x1F";
+			}
+			if ($value === '[:space:]') {
+				return " \t\r\n\v\f";
+			}
+			if ($value === '[:punct:]') {
+				return '!"#$%&\'()*+,\-./:;<=>?@[\]^_`{|}~';
+			}
+		}
+
+		if ($token === 'anchor' || $token === 'match_point_reset') {
+			return '';
 		}
 
 		return null;
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -625,6 +625,13 @@ final class RegexArrayShapeMatcher
 			}
 		}
 
+		// [^0-9] should not parse as numeric-string, and [^list-everything-but-numbers] is technically
+		// doable but really silly compared to just \d so we can safely assume the string is not numeric
+		// for negative classes
+		if ($ast->getId() === '#negativeclass') {
+			$isNumeric = TrinaryLogic::createNo();
+		}
+
 		foreach ($children as $child) {
 			$this->walkGroupAst(
 				$child,
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -625,33 +625,6 @@ final class RegexArrayShapeMatcher
 			}
 		}
 
-		if ($ast->getId() === '#range' || $ast->getId() === '#class') {
-			if ($isNumeric->maybe()) {
-				$allNumeric = null;
-				foreach ($children as $child) {
-					$literalValue = $this->getLiteralValue($child);
-
-					if ($literalValue === null) {
-						break;
-					}
-
-					if (Strings::match($literalValue, '/^\d+$/') === null) {
-						break;
-					}
-
-					$allNumeric = true;
-				}
-
-				if ($allNumeric === true) {
-					$isNumeric = TrinaryLogic::createYes();
-				}
-			}
-
-			if (!$inOptionalQuantification) {
-				$isNonEmpty = TrinaryLogic::createYes();
-			}
-		}
-
 		foreach ($children as $child) {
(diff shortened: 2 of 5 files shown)


src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -372,12 +372,13 @@ private function parseGroups(string $regex): ?array
 	{
 		if (self::$parser === null) {
 			/** @throws void */
-			self::$parser = Llk::load(new Read('hoa://Library/Regex/Grammar.pp'));
+			self::$parser = Llk::load(new Read(__DIR__ . '/../../../resources/RegexGrammar.pp'));
```
Seldaek, line 380: @ondrejmirtes Are we ok loading the file from here instead? Not sure what this does in terms of licensing, nor if there's a better place to put it. If so I'll delete the patch file and drop the hoa/regex requirement.

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -372,12 +372,13 @@ private function parseGroups(string $regex): ?array
...
 		}
 
 		try {
 			$ast = self::$parser->parse($regex);
 		} catch (Exception) {
+			// TODO log parse errors in vebose mode?
```
Seldaek: How can we achieve this from a matcher is it possible to output diagnostics/errors at least for easier debugging?

resources/RegexGrammar.pp
```diff
@@ -0,0 +1,207 @@
...
+%token class:posix_class         \[\^?:[a-z]+:\]
+%token class:class_              \[
+%token class:_class              \]                 -> default
+%token class:range               \-
+%token class:class_literals      [^\\\[\]-]+
+%token class:escaped_end_class   \\\]
+// taken over from literals
+%token class:character           \\([aefnrt]|c[\x00-\x7f])
+%token class:dynamic_character   \\([0-7]{3}|x[0-9a-zA-Z]{2}|x{[0-9a-zA-Z]+})
+%token class:character_type      \\([CdDhHNRsSvVwWX]|[pP]{[^}]+})
```
Seldaek: @staabm I think something here messes up with your new numeric-string detection (see test failures). I'm not sure what though. If you have a chance to take a look I'd appreciate as that ast walking code is really kinda hell to look at for me as I have no clue about the AST walker and not much clue about phpstan internals :D

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -372,12 +372,13 @@ private function parseGroups(string $regex): ?array
 	{
 		if (self::$parser === null) {
 			/** @throws void */
-			self::$parser = Llk::load(new Read('hoa://Library/Regex/Grammar.pp'));
+			self::$parser = Llk::load(new Read(__DIR__ . '/../../../resources/RegexGrammar.pp'));
 		}
 
 		try {
 			$ast = self::$parser->parse($regex);
```
Seldaek, line 391: @staabm another thing I noticed while working on this, it looks like this parses every regex several times, it might be worth memoizing the result to save cpu cycles. Not sure but maybe worth considering.

src/Type/Php/RegexArrayShapeMatcher.php
staabm: usually I create a `hoa.php`

```
<?php 
// hoa.php

require 'vendor/autoload.php';

$regex = 'ab(c|d){2,4}e?';

// 1. Read the grammar.
$grammar  = new Hoa\File\Read('hoa://Library/Regex/Grammar.pp');

// 2. Load the compiler.
$compiler = Hoa\Compiler\Llk\Llk::load($grammar);

// 3. Lex, parse and produce the AST.
$ast      = $compiler->parse($regex);

// 4. Dump the result.
$dump     = new Hoa\Compiler\Visitor\Dump();
echo $dump->visit($ast);

/**
 * Will output:
 *     >  #expression
 *     >  >  #concatenation
 *     >  >  >  token(literal, a)
 *     >  >  >  token(literal, b)
 *     >  >  >  #quantification
 *     >  >  >  >  #alternation
 *     >  >  >  >  >  token(literal, c)
 *     >  >  >  >  >  token(literal, d)
 *     >  >  >  >  token(n_to_m, {2,4})
 *     >  >  >  #quantification
 *     >  >  >  >  token(literal, e)
 *     >  >  >  >  token(zero_or_one, ?)
 */
```

and invoke it with `php hoa.php`

src/Type/Php/RegexArrayShapeMatcher.php
staabm: I had the same idea today and did some profiles on `tests/PHPStan/Analyser/nsrt/preg_match_shapes.php`. the parsing is consuming quite a bit, but I was not yet able to find a real world project in which this has a impact on overall scan time

src/Type/Php/RegexArrayShapeMatcher.php
Seldaek: Thanks that helps (tho it needs `$grammar  = new Hoa\File\Read('resources/RegexGrammar.pp');` now, but I meant more a way to just output `$e->getMessage()` so we can show the parse error if a regex fails to parse. That might belong more into a rule scanning regexes fed into preg_match for correctness though.

src/Type/Php/RegexArrayShapeMatcher.php
staabm: I am not sure this path will work in the packaged phpstan.phar - while I agree that having the `RegexGrammar.pp` in the phpstan-src repo might ease the contribution experience - as the patching is really cumbersome.

resources/RegexGrammar.pp
staabm: I can have a look

Seldaek: With the hoa.php trick above I think I can work it out..

staabm: before this PR the ast for `'~^(02,([\d.]{10}))$~'` looked like:

```
$ php hoa.php
>  #expression
>  >  #concatenation
>  >  >  token(literal, ~)
>  >  >  token(anchor, ^)
>  >  >  #capturing
>  >  >  >  #concatenation
>  >  >  >  >  token(literal, 0)
>  >  >  >  >  token(literal, 2)
>  >  >  >  >  token(literal, ,)
>  >  >  >  >  #capturing
>  >  >  >  >  >  #quantification
>  >  >  >  >  >  >  #class
>  >  >  >  >  >  >  >  token(character_type, \d)
>  >  >  >  >  >  >  >  token(literal, .)
>  >  >  >  >  >  >  token(exactly_n, {10})
>  >  >  token(anchor, $)
```

after this PR it looks like

```
$ php hoa.php
>  #expression
>  >  #concatenation
>  >  >  token(literal, ~)
>  >  >  token(anchor, ^)
>  >  >  #capturing
>  >  >  >  #concatenation
>  >  >  >  >  token(literal, 0)
>  >  >  >  >  token(literal, 2)
>  >  >  >  >  token(literal, ,)
>  >  >  >  >  #capturing
>  >  >  >  >  >  #quantification
>  >  >  >  >  >  >  #class
>  >  >  >  >  >  >  >  token(class:character_type, \d)
>  >  >  >  >  >  >  >  token(class:class_literals, .)
>  >  >  >  >  >  >  token(exactly_n, {10})
>  >  >  token(anchor, $)
>  >  >  token(literal, ~)
```

e.g. `character_type` vs. `class:character_type` (more differences in there)

the change in token-names need to be reflected in the `RegexArrayShapeMatcher`. I guess you just need to search for the hardcoded string values.

but I wonder where this `class:` prefix is coming from

staabm: don't waste too much time thinking thru the ast walking. in case you are happy with the grammar I can look into the failling stuff

src/Type/Php/RegexArrayShapeMatcher.php
mvorisek: I completely agree moving from the patching as the patching is very hard to understand and the changes needed are more and more complex.

resources/RegexGrammar.pp
Seldaek: Ok that helped a LOT, thanks. It would have saved me a lot of guesswork earlier 😅  But anyway glad I got it now.

src/Type/Php/RegexArrayShapeMatcher.php
Seldaek: Ok I got rid of it, we'll see what @ondrejmirtes says :)

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -653,15 +656,22 @@ private function walkGroupAst(TreeNode $ast, TrinaryLogic &$isNonEmpty, TrinaryL
...
+		if ($node->getId() === 'token' && $node->getValueToken() === 'class:range') {
 			return $node->getValueValue();
 		}
 
+		// [:digit:] and the like, more support coming later
+		if ($node->getId() === 'token' && $node->getValueToken() === 'class:posix_class') {
+			if ($node->getValueValue() === '[:digit:]') {
+				return implode('', range(0, 9));
+			}
+		}
```
Seldaek: @staabm I think it's mergeable as is now, but just for future reference, there is a lot more we can support here:

In terms of token types, you can use this to get a full list of tokens that can appear within character classes:

```
<?php
require 'vendor/autoload.php';

$regex = '{[-\p{L}[\]*|\x03\a\b+?{}(?:)-]+[^[:digit:]?{}a-z0-9k-#]+a-z}';
$grammar  = new Hoa\File\Read('resources/RegexGrammar.pp');
$compiler = Hoa\Compiler\Llk\Llk::load($grammar);
$ast      = $compiler->parse($regex);
$dump     = new Hoa\Compiler\Visitor\Dump();
echo $dump->visit($ast);
```

Output with some notes:

```
>  >  >  #quantification
>  >  >  >  #class
>  >  >  >  >  token(class:range, -)
>  >  >  >  >  token(class:character_type, \p{L})
>  >  >  >  >  token(class:class_, [)
>  >  >  >  >  token(class:escaped_end_class, \]) // this is matching a literal `]`, the backslash has to be dropped
>  >  >  >  >  token(class:literal, *)
>  >  >  >  >  token(class:literal, |)
>  >  >  >  >  token(class:dynamic_character, \x03)
>  >  >  >  >  token(class:character, \a)
>  >  >  >  >  token(class:character, \b)
>  >  >  >  >  token(class:literal, +)
>  >  >  >  >  token(class:literal, ?)
>  >  >  >  >  token(class:literal, {)
>  >  >  >  >  token(class:literal, })
>  >  >  >  >  token(class:literal, ()
>  >  >  >  >  token(class:literal, ?)
>  >  >  >  >  token(class:literal, :)
>  >  >  >  >  token(class:literal, ))
>  >  >  >  >  token(class:range, -)
>  >  >  >  token(one_or_more, +)
>  >  >  #quantification
>  >  >  >  #negativeclass
>  >  >  >  >  token(class:posix_class, [:digit:]) // I added tentative support for digit, I can add more later but first want to discuss how to best do this
>  >  >  >  >  token(class:literal, ?)
>  >  >  >  >  token(class:literal, {)
>  >  >  >  >  token(class:literal, })
>  >  >  >  >  #range
>  >  >  >  >  >  token(class:literal, a)
>  >  >  >  >  >  token(class:literal, z)
>  >  >  >  >  #range
>  >  >  >  >  >  token(class:literal, 0)
>  >  >  >  >  >  token(class:literal, 9)
>  >  >  >  >  #range
>  >  >  >  >  >  token(class:literal, k)
>  >  >  >  >  >  token(class:literal, #) // gotta watch out ranges can be from/to any character, this matches any of kjihgfedcba`_^]\[ZYXWVUTSRQPONMLKJIHGFEDCBA@?>=<;:9876543210/.-,+*)('&%$#
>  >  >  >  token(one_or_more, +)
>  >  >  token(literal, a)
>  >  >  token(literal, -)
>  >  >  token(literal, z)
```

staabm: Thanks for that. If you have more concrete examples of regex patterns which go wrong I am all ears. 

I don't need/want to support any super edge case format, but we should at least try hard to fix any failure in type-inference (e.g where the current logic decides its a numeric-string but its actually just string)

(At best in new issues, so we can get this PR merged)

Seldaek: Yes I think this could be merged already. Once merged I'll check again how it looks on the Composer repo (I somehow cannot run phpstan from source on the composer project), but I believe this will fix all regex parse errors I encountered. I am not sure though as the parse errors are swallowed.. that's why I wanted a way to surface them to users so they know that phpstan couldn't parse a given regex, and we could get reports and fix those. Right now you only realize it couldn't parse it when the type of $matches is array<string> and that causes further problems in the code.

And yeah I agree supporting everything including constant strings etc may not make sense, but right now I think you probably can end up with patterns where it infers numeric-string wrongly. I'll give that a closer look some other time :)

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -372,15 +375,22 @@ private function parseGroups(string $regex): ?array
...
 		}
 
 		try {
-			$ast = self::$parser->parse($regex);
-		} catch (Exception) {
+			Strings::match('', $regex);
+		} catch (RegexpException $e) {
+			// pattern is invalid, so let the RegularExpressionPatternRule report it
 			return null;
 		}
```
Seldaek, line 388: I just added a first validation pass here @staabm, and a ShouldNotHappenException below so that if the grammar fails we know.. But it's a bit brutal to fail the whole process due to that. Is there not another softer exception than ShouldNotHappenException to surface an internal error to users without aborting the entire run?

staabm: I think we should have a separate rule which depends on a feature flag .. maybe bleeding edge.. maybe some separate flag.

maybe it can get part of `RegularExpressionPatternRule` but maybe completly separate rule is also useful.
or maybe we can enable exception throwing via a new $flags constant - similar to how we did it with `RegexArrayShapeMatcher::PREG_UNMATCHED_AS_NULL_ON_72_73`

Seldaek: Ok I got rid of the throw now.

Seldaek:
@staabm please review the last ~4~5 commits and especially test changes.

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -472,3 +470,21 @@ function (string $s): void {
 		assertType("array{string, non-empty-string}", $matches);
 	}
 };
+
+function bug11323(string $s): void {
+	if (preg_match('/([*|+?{}()]+)([^*|+[:digit:]?{}()]+)/', $s, $matches)) {
+		assertType('array{string, non-empty-string, non-empty-string}', $matches);
+	}
+	if (preg_match('/\p{L}[[\]]+([-*|+?{}(?:)]+)([^*|+[:digit:]?{a-z}(\p{L})\a-]+)/', $s, $matches, PREG_UNMATCHED_AS_NULL)) {
```
staabm: I usually put tests which use `PREG_UNMATCHED_AS_NULL` into `tests/PHPStan/Analyser/nsrt/bug-11311.php` for assertions which only work with php74+

additionally it would make sense to put the same test with different assertions (for php < 7.4) into `tests/PHPStan/Analyser/nsrt/bug-11311-php72.php`

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -472,3 +470,21 @@ function (string $s): void {
 		assertType("array{string, non-empty-string}", $matches);
 	}
 };
+
+function bug11323(string $s): void {
+	if (preg_match('/([*|+?{}()]+)([^*|+[:digit:]?{}()]+)/', $s, $matches)) {
```
staabm, line 467: I would expect a test with a capturing group of `([:digit:])` which resolves to a numeric-string.

or was this intentionally not implemented with this PR?

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
staabm: I would also expect a test with just `(\p{L})` capturing, to verify we have the proper type for this regex character class inferred

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -129,26 +129,24 @@ function doUnknownFlags(string $s, int $flags): void {
...
 	}
-	assertType('array<string>', $matches);
+	assertType('array{}|array{0: string, Foo: numeric-string|non-empty-string, 1: non-empty-string, 2?: numeric-string}', $matches);
 }
 
 function doMultipleConsecutiveCaptureGroupsWithSameNameWithModifier(string $s): void {
 	if (preg_match('/(?J)(?<Foo>[a-z]+)|(?<Foo>[0-9]+)/', $s, $matches)) {
-		// could be assertType('array{0: string, Foo: string, 1: string}', $matches);
-		assertType('array<string>', $matches);
+		assertType('array{0: string, Foo: numeric-string|non-empty-string, 1: non-empty-string, 2?: numeric-string}', $matches);
```
staabm, line 139: this union `Foo: numeric-string|non-empty-string` is pretty awesome.

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -8003,7 +8003,7 @@ public function dataPassedByReference(): array
 				'$arr',
 			],
 			[
-				'array{0?: string}',
+				'array<string>',
```
staabm, line 8006: ~~did we loose precision in this test?~~

edit: oh I just see there is a [invalid pattern in the `preg_match`](https://github.com/phpstan/phpstan-src/blob/f257541ac081bec65018b05c795c62513d830201/tests/PHPStan/Analyser/data/passed-by-reference.php#L13) and I guess the new grammar is catching that?

Seldaek: Yes well the new sanity check I added before parsing catches it and it bails out. The grammar doesn't actually check for delimiters so it sees 'a' as a literal a and is happy with that.

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
Seldaek: I just removed the PREG_UNMATCHED_AS_NULL as it's completely irrelevant, just landed there through copy paste. Will add a few new tests according to your feedback 👍🏻

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -591,20 +617,8 @@ private function walkGroupAst(TreeNode $ast, TrinaryLogic &$isNonEmpty, TrinaryL
...
-
-			if ($ast->getValueToken() === 'character_type') {
-				if ($ast->getValueValue() === '\d') {
-					if ($isNumeric->maybe()) {
-						$isNumeric = TrinaryLogic::createYes();
-					}
-				} else {
-					$isNumeric = TrinaryLogic::createNo();
+				} elseif ($isNumeric->maybe()) {
+					$isNumeric = TrinaryLogic::createYes();
```
staabm, line 619: I don't think this else-if can work.

a single `\d` does not turn a expression which exists of more tokens into a numeric-string.

we can assign `$isNumeric = TrinaryLogic::createYes();` only in cases in which we know the whole expression.

staabm: hmm I tried to find examples which break this logic, but couldn't find any. 

maybe its actually correct ;)

staabm: hmm I think we need a test for `(.\d)` which asserts this gets a `non-empty-string`

Seldaek: Ah I see, I missed the by-ref, I guess that's also why $allNumeric = false before the break wasn't strictly needed (but I still think it's safer to include). Fixed in https://github.com/phpstan/phpstan-src/pull/3244/commits/4034c59ea8ca9785cdb28b740f28125861006e98

Seldaek: I could also not find a way to trigger a problem the other way, so I just removed it.

Seldaek: Ah sorry I'm confusing myself here without full context.. I removed the code below handling \d in character_type as getLiteralValue now handles \d and returns `'0'`, that's why I added the isNumeric here.

I think it works because it can still be set to false/createNo later if some other literal is not a digit.

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -620,6 +632,7 @@ private function walkGroupAst(TreeNode $ast, TrinaryLogic &$isNonEmpty, TrinaryL
 					$literalValue = $this->getLiteralValue($child);
 
 					if ($literalValue === null) {
+						$allNumeric = false;
```
staabm: at this point, I think we can just drop the 2 `$allNumeric = false;` occurences

staabm (approved):
this looks really good for me.

only open question (which Ondrej has to decide on): what todo with the `RegexGrammar.pp`.

src/Type/Php/RegexArrayShapeMatcher.php
Seldaek: Ok done in 292e21147 let's see if the build agrees :D

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -624,7 +636,6 @@ private function walkGroupAst(TreeNode $ast, TrinaryLogic &$isNonEmpty, TrinaryL
 					}
 
 					if (Strings::match($literalValue, '/^\d+$/') === null) {
-						$allNumeric = false;
```
staabm, line 627: I played a bit more with it. 

I have a feeling this whole `if ($ast->getId() === '#range' || $ast->getId() === '#class') {...}` can be removed, because we achieve the same thing already via recursive `walkGroupAst()` and handling of the `if ($ast->getId() === 'token') {` above (ranges and classes can only contain tokens, right?)

at least no test failed on my machine as I deleted this whole if-block :D

Seldaek: Yeah that is possible, but can we maybe just let this sit until @ondrejmirtes can review and then bring further cleanups later? :D

resources/RegexGrammar.pp
```diff
@@ -0,0 +1,213 @@
...
+// parser. More informations at http://pcre.org/pcre.txt, sections pcrepattern &
+// pcresyntax.
+//
+// @copyright  Copyright © 2007-2017 Hoa community.
+// @license    New BSD License
+//
+
+
+// Skip.
+%skip   nl                       \n
```
mvorisek, line 46: `\n` can be in pcre class as well

src/Type/Php/RegexArrayShapeMatcher.php
mvorisek: I propose to add the whole and unpatched `RegexGrammar.pp` as first/separate commit, then the patched/actual grammar and then the changes from this PR. This will hugely improve the review and git history.

resources/RegexGrammar.pp
Seldaek: Yes I am unsure why this is here, I haven't experimented much, but it is rare to have actual newlines in regexes as they're usually escaped as \n, while this here matches a real newline.

I'll create a new issue for this to look at it later, together with proper support for the x modifier, because it's very much related. Edit: https://github.com/phpstan/phpstan/issues/11360

src/Type/Php/RegexArrayShapeMatcher.php
Seldaek: Good idea, done 👍🏻

src/Type/Php/RegexArrayShapeMatcher.php
Seldaek: Ok I've done it after all in https://github.com/phpstan/phpstan-src/pull/3244/commits/64901f62c36a17e4e9a54243b795e2affbcadd86 as I was modifying the PR again anyway.. Looks good :)

ondrejmirtes:
I cherry-picked the grammar in the repo directly to 1.11.x, rebased this PR, and merged it. Thanks! :)

PR #3247, opened 2024-07-16
Refactor PrintfParametersRule

staabm:
extracted from https://github.com/phpstan/phpstan-src/pull/3126

factor out `PrintfHelper` for later re-use in a to be created `PrintfArrayParametersRule`

### DIFF
--- a/src/Rules/Functions/PrintfHelper.php
+++ b/src/Rules/Functions/PrintfHelper.php
@@ -1,0 +1,75 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Functions;
+
+use Nette\Utils\Strings;
+use PHPStan\Php\PhpVersion;
+use function array_filter;
+use function count;
+use function max;
+use function sprintf;
+use function strlen;
+use const PREG_SET_ORDER;
+
+final class PrintfHelper
+{
+
+	public function __construct(private PhpVersion $phpVersion)
+	{
+	}
+
+	public function getPrintfPlaceholdersCount(string $format): int
+	{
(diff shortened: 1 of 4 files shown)


src/Rules/Functions/PrintfHelper.php
```diff
@@ -0,0 +1,75 @@
...
+
+	public function getPrintfPlaceholdersCount(string $format): int
+	{
+		return $this->getPlaceholdersCount('(?:[bs%s]|l?[cdeEgfFGouxX])', $format);
+	}
+
+	public function getScanfPlaceholdersCount(string $format): int
+	{
+		return $this->getPlaceholdersCount('(?:[cdDeEfinosuxX%s]|\[[^\]]+\])', $format);
+	}
```
staabm, line 29: I need this use-cases separated because I only need the `getPrintfPlaceholdersCount` in `PrintfArrayParametersRule`

src/Rules/Functions/PrintfParametersRule.php
```diff
@@ -40,25 +50,17 @@ public function processNode(Node $node, Scope $scope): array
...
-		];
-
-		$name = strtolower((string) $node->name);
-		if (!array_key_exists($name, $functionsArgumentPositions)) {
+		if (!$this->reflectionProvider->hasFunction($node->name, $scope)) {
 			return [];
 		}
 
-		$formatArgumentPosition = $functionsArgumentPositions[$name];
+		$functionReflection = $this->reflectionProvider->getFunction($node->name, $scope);
```
staabm, line 57: this was one of the few rules which did not validate the `$node->name` via `ReflectionProvider`. 

I guess its not strictly necessary because we are handling php-src native functions - I added it for consistency.

ondrejmirtes:
Thank you!

PR #3249, opened 2024-07-16
RegexArrayShapeMatcher - More precise non-empty-string and numeric-string

staabm:
based on the regex ast we try to detect more precise string types, like non-empty-string and numeric-string.

I intentionally left out constant-string types for now to reduce complexity and learn from real world use-cases in the wild.

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-11311-php72.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-11311-php72.php
@@ -8,23 +8,23 @@ use function PHPStan\Testing\assertType;
 
 function doFoo(string $s) {
 	if (1 === preg_match('/(?<major>\d+)\.(?<minor>\d+)(?:\.(?<patch>\d+))?/', $s, $matches, PREG_UNMATCHED_AS_NULL)) {
-		assertType('array{0: string, major: string, 1: string, minor: string, 2: string, patch?: string, 3?: string}', $matches);
+		assertType('array{0: string, major: numeric-string, 1: numeric-string, minor: numeric-string, 2: numeric-string, patch?: numeric-string, 3?: numeric-string}', $matches);
 	}
 }
 
 function doUnmatchedAsNull(string $s): void {
 	if (preg_match('/(foo)?(bar)?(baz)?/', $s, $matches, PREG_UNMATCHED_AS_NULL)) {
-		assertType('array{0: string, 1?: string, 2?: string, 3?: string}', $matches);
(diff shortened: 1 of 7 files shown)


staabm:
//cc @Seldaek would be great to have another pair of eyeballs on the test-expectations

ondrejmirtes:
Thank you.

ondrejmirtes:
Thank you.

ondrejmirtes:
Please send update to phpstan-nette :)

tests/PHPStan/Analyser/nsrt/bug-11311-php72.php
```diff
@@ -8,23 +8,23 @@
...
 	}
-	assertType('array{}|array{0: string, 1?: string, 2?: string, 3?: string}', $matches);
+	assertType('array{}|array{0: string, 1?: non-empty-string, 2?: non-empty-string, 3?: non-empty-string}', $matches);
 }
 
 // see https://3v4l.org/VeDob#veol
 function unmatchedAsNullWithOptionalGroup(string $s): void {
 	if (preg_match('/Price: (£|€)?\d+/', $s, $matches, PREG_UNMATCHED_AS_NULL)) {
-		assertType('array{0: string, 1?: string}', $matches);
+		assertType('array{0: string, 1?: non-empty-string}', $matches);
```
Seldaek, line 25: Given unmatched as null this should be `1: non-empty-string|null` IMO.

`preg_match('/Price: (£|€)?\d+/', 'Price: 3', $matches, PREG_UNMATCHED_AS_NULL)` for ex yields:

```
array (size=2)
  0 => string 'Price: 3' (length=8)
  1 => null
```

Seldaek:
Ok this is amazing :D I can't wait to try

tests/PHPStan/Analyser/nsrt/bug-11311-php72.php
staabm: this file is the php72 variant, which does not "correctly support" nulls in the `PREG_UNMATCHED_AS_NULL` case (because php-src behaves differently).

see https://3v4l.org/No3Tf#v7.3.26

there is the identical test but for php7.4+ with the expectations you just shared:
https://github.com/phpstan/phpstan-src/pull/3249/files#diff-587efc95b2825fa3f959191a2c1dae92edda5ad901653ae174e58bc723152fa9R25

Seldaek: Oh sorry I missed that 🤦🏻‍♂️

PR #3251, opened 2024-07-18
Simplify RegularExpressionPatternRule

staabm:
Only refactoring to improve readability.

moving the IFs outside the loops instead of loop + continue.


ondrejmirtes:
Thank you.

PR #3252, opened 2024-07-18
Implement RegularExpressionQuotingRule

staabm:
implements the idea described in https://github.com/phpstan/phpstan/issues/11338#issuecomment-2235985564

closes https://github.com/phpstan/phpstan/issues/11338

### DIFF
--- a/src/Rules/Regexp/RegularExpressionQuotingRule.php
+++ b/src/Rules/Regexp/RegularExpressionQuotingRule.php
@@ -14,11 +14,15 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Constant\ConstantStringType;
+use function array_filter;
 use function array_merge;
+use function array_values;
 use function count;
 use function in_array;
 use function sprintf;
+use function strlen;
 use function substr;
 
 /**
@@ -132,6 +136,11 @@ class RegularExpressionQuotingRule implements Rule
 			return null;
 		}
 
+		$patternDelimiters = $this->removeDefaultEscapedDelimitiers($patternDelimiters);
+		if ($patternDelimiters === []) {
+			return null;
+		}
+
 		if (count($args) === 1) {
 			if (count($patternDelimiters) === 1) {
 				return RuleErrorBuilder::message(sprintf('Call to preg_quote() is missing delimiter %s to be effective.', $patternDelimiters[0]))
@@ -147,16 +156,29 @@ class RegularExpressionQuotingRule implements Rule
 		}
 
 		if (count($args) >= 2) {
+
 			foreach ($scope->getType($args[1]->value)->getConstantStrings() as $quoteDelimiterType) {
-				if (!in_array($quoteDelimiterType->getValue(), $patternDelimiters, true)) {
+				$quoteDelimiter = $quoteDelimiterType->getValue();
+
+				$quoteDelimiters = $this->removeDefaultEscapedDelimitiers([$quoteDelimiter]);
+				if ($quoteDelimiters === []) {
+					continue;
+				}
+
+				if (count($quoteDelimiters) !== 1) {
+					throw new ShouldNotHappenException();
+				}
+				$quoteDelimiter = $quoteDelimiters[0];
+
+				if (!in_array($quoteDelimiter, $patternDelimiters, true)) {
 					if (count($patternDelimiters) === 1) {
-						return RuleErrorBuilder::message(sprintf('Call to preg_quote() uses invalid delimiter %s while pattern uses %s.', $quoteDelimiterType->getValue(), $patternDelimiters[0]))
+						return RuleErrorBuilder::message(sprintf('Call to preg_quote() uses invalid delimiter %s while pattern uses %s.', $quoteDelimiter, $patternDelimiters[0]))
 							->line($pregQuote->getStartLine())
 							->identifier('argument.invalidPregQuote')
 							->build();
 					}
 
-					return RuleErrorBuilder::message(sprintf('Call to preg_quote() uses invalid delimiter %s.', $quoteDelimiterType->getValue()))
+					return RuleErrorBuilder::message(sprintf('Call to preg_quote() uses invalid delimiter %s.', $quoteDelimiter))
 						->line($pregQuote->getStartLine())
 						->identifier('argument.invalidPregQuote')
 						->build();
@@ -198,19 +220,31 @@ class RegularExpressionQuotingRule implements Rule
 			return null;
 		}
 
-		$firstChar = substr($string->getValue(), 0, 1);
+		return substr($string->getValue(), 0, 1);
+	}
 
-		// check for delimiters which get properly escaped by default
-		if (in_array(
-			$firstChar,
-			['.', '\\',  '+', '*', '?', '[', '^', ']', '$', '(', ')', '{', '}', '=', '!', '<', '>', '|', ':', '-', '#'],
-			true,
-		)
-		) {
-			return null;
+	/**
+	 * @param string[] $patternDelimiters
+	 *
+	 * @return list<string>
+	 */
+	private function removeDefaultEscapedDelimitiers(array $delimiters): array
+	{
+		return array_values(array_filter($delimiters, fn (string $delimiter): bool => !$this->isDefaultEscaped($delimiter)));
+	}
+
+	private function isDefaultEscaped(string $delimiter): bool
+	{
+		if (strlen($delimiter) !== 1) {
+			return false;
 		}
 
-		return $firstChar;
+		return in_array(
+			$delimiter,
+			// these delimiters are escaped, not matter what preg_quote() 2nd arg looks like
+			['.', '\\',  '+', '*', '?', '[', '^', ']', '$', '(', ')', '{', '}', '=', '!', '<', '>', '|', ':', '-', '#'],
+			true,
+		);
 	}
 
 	private function getNormalizedArgs(FuncCall $functionCall, Scope $scope, FunctionReflection $functionReflection): ?array
--- a/src/Rules/Regexp/RegularExpressionQuotingRule.php
+++ b/src/Rules/Regexp/RegularExpressionQuotingRule.php
@@ -6,7 +6,11 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\BinaryOp\Concat;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Name;
+use PHPStan\Analyser\ArgumentsNormalizer;
 use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
@@ -15,7 +19,6 @@ use function array_merge;
 use function count;
 use function in_array;
 use function sprintf;
-use function strtolower;
 use function substr;
 
 /**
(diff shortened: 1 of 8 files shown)


staabm:
//cc @Seldaek

staabm:
as pointed out by Jordi in https://github.com/phpstan/phpstan/issues/11338#issuecomment-2236236844 we could also think about introspecting every string-concatenation which contains a `preg_quote` function invocation.

that way this rule would also work for composer/pcre and nette/strings automatically.

this could report false-positives though when someone is building a regex out of several independent variables which get concatenated together and also brings the need to check _every_ string concat in the code-base.

----

alternatively we could invoke the rule on all `CallLike` (or `FuncCall|MethodCall`) instead of just `FuncCall`.
that way we would at least cover calls to e.g. composer/pcre and nette/utils which use the `preg_quote` call directly in the call, like in 
```php
if (Preg::match('{fo'. preg_quote($s) .'+}', $string, $matches))
{ 
  // ...
} 
```

----

one last option could be to create a helper service which can be re-used from rules (similar to `RegexArrayShapeMatcher` - and implement separate new rules which wire the same underlying service.

----

Ondrej, what do you think about this use-case/implications?

Seldaek:
> this could report false-positives though when someone is building a regex out of several independent variables which get concatenated together and also brings the need to check _every_ string concat in the code-base.

> one last option could be to create a helper service which can be re-used from rules (similar to `RegexArrayShapeMatcher` - and implement separate new rules which wire the same underlying service.

On further consideration, because this is such a rare problem in real life (unless you stick to `/` for regex delimiters, but that's just a bad idea IMO additionally because we're in a web environment full of slashes that have to be escaped if they become special) I think it's maybe not worth too much complexity/effort/CPU cycles.

Just my 2c though :)

staabm:
but wrong escaping - at best with user provided input - would be a security vulnerability, right?
(in case the end-user can influence the regex pattern, because of wrong escaping)

Seldaek:
You could trigger a crash or in really bad cases maybe exploit something further down if you can break a regex validation or so but that's unsure how.. It's not like the good old days anymore where you had the e/execute modifier to call arbitrary functions anymore. That's gone as of php 7.0

ondrejmirtes:
Devil's advocate here: is this really a common mistake we should spend our time on?

Also: it'd be much more powerful to express this in the typesystem (but also a bit more complex). If we were able to carry "this is a preg_quoted string with such and such delimiter" in a subclass of ConstantStringType, we'd be able to check more code - that for example first saves the result of `preg_quote` in a variable and only then uses it in Concat in an argument to preg_match.

But other than that, I'm not sure even if this PR is worth it. I'd much rather have other existing functions like `preg_match_all` (shouldn't be too hard) or `preg_replace_callback` be aware of the awesome array shape inference from regexes :)

staabm:
I don't have a strong opinion on this one. it was easy to implement and sounded useful.

I am ok with closing if you guys don't see value in it

staabm:
> But other than that, I'm not sure even if this PR is worth it. I'd much rather have other existing functions like `preg_match_all` (shouldn't be too hard) or `preg_replace_callback` be aware of the awesome array shape inference from regexes :)

its not forgotten - I promise ;-D

staabm:
one last data-point, then I am silent

> (unless you stick to `/` for regex delimiters, but that's just a bad idea IMO additionally because we're in a web environment full of slashes that have to be escaped if they become special)

at least on github.com we have ~650k files which use `/` as delimiter
https://github.com/search?q=%22preg_match%28%27%2F%22+language%3APHP&type=code&l=PHP

Seldaek:
Yeah I know it's extremely common because some languages like JS have native regex types with `/` delimiters.. But just a terrible idea IMO

ondrejmirtes:
@Seldaek Sorry, I'm too slow. How is JS syntax relevant to quoting strings inside regexes?

Seldaek:
I mean in JS you have to define regexes as `/foo/`, not as strings with any choice of delimiter. Also sed and probably other old tools propagated the use of `/` as a default delimiter.

That means `/` is a very common delimiter, but it is not escaped by default by preg_quote, as it is not a metacharacter. That makes it quite common IMO that people do preg_quote and forget to pass `'/'` as second argument. And if they do that they might experience crashes/problems at runtime if the user content contains `/`.

staabm:
Should I rebase, or will it be closed?

ondrejmirtes:
Alright, I get it, so `preg_quote($s)` is always wrong because it doesn't even fit the `/`-delimited regex.

The problem most often isn't using the wrong delimiter `'#' . preg_quote($s, '/') . '#'` but using no delimiter in the call.

In that case this is definitely a useful addition!

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,182 @@
...
+		return FuncCall::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (!$node->name instanceof Node\Name) {
+			return [];
+		}
+
+		$functionName = strtolower((string) $node->name);
```
ondrejmirtes: You've forgot to do the usual hasFunction/getFunction dance because of functions in namespaces.

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,182 @@
...
+				'preg_replace',
+				'preg_replace_callback',
+				'preg_split',
+			], true)
+		) {
+			return [];
+		}
+
+		$args = $node->getArgs();
+		if (!isset($args[0])) {
```
ondrejmirtes: ArgumentsNormalizer is needed for named arguments.

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,182 @@
...
+
+	/**
+	 * @param string[] $patternDelimiters
+	 */
+	private function validatePregQuote(FuncCall $pregQuote, Scope $scope, array $patternDelimiters): ?IdentifierRuleError
+	{
+		$args = $pregQuote->getArgs();
+
+		if (
+			count($args) === 1
```
ondrejmirtes: No need for multiline here

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,182 @@
...
+		) {
+			return RuleErrorBuilder::message('Call to preg_quote() is missing delimiter parameter to be effective.')
```
ondrejmirtes, line 156: Say which delimiter should be used here. Also: ArgumentsNormalizer and named arguments.

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,182 @@
...
+				->line($pregQuote->getStartLine())
+				->identifier('argument.invalidPregQuote')
+				->build();
+		}
+
+		if (count($args) >= 2) {
+			foreach ($scope->getType($args[1]->value)->getConstantStrings() as $quoteDelimiterType) {
+				if (!in_array($quoteDelimiterType->getValue(), $patternDelimiters, true)) {
+					return RuleErrorBuilder::message(sprintf('Call to preg_quote() uses invalid delimiter %s.', $quoteDelimiterType->getValue()))
```
ondrejmirtes: Say which delimiter should be used.

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,182 @@
...
+	{
+		if ($string->getValue() === '') {
+			return null;
+		}
+
+		$firstChar = substr($string->getValue(), 0, 1);
+
+		if (in_array(
+			$firstChar,
+			['.', '\\',  '+', '*', '?', '[', '^', ']', '$', '(', ')', '{', '}', '=', '!', '<', '>', '|', ':', '-', '#'],
```
ondrejmirtes: What's this list? This would be nice to have a comment.

ondrejmirtes (changes requested):

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,217 @@
...
+		return null;
+	}
+
+	private function getDelimiterFromString(ConstantStringType $string): ?string
+	{
+		if ($string->getValue() === '') {
+			return null;
+		}
+
+		$firstChar = substr($string->getValue(), 0, 1);
```
Seldaek: If you want to reduce false positives from partial pattern strings, you could look at the end of the string if you find the end delimiter followed by known modifiers. If you don't find that most likely it's not a complete pattern and you shouldn't warn. Special care has to be taken tho with ({[ that the end delimiter isn't the same character but)}]

If you think its too much then skip this imo and lets see first feedback

staabm: do you say we shouldn't error on 
`preg_match('{' . preg_quote($s, '&') . 'pattern}', $s);`
?

or could you give a concrete example? I think - since we only analyze a Concat directly given in the func call - there cannot be a partial regex (or only a invalid regex)

staabm:
btw: the new rule reports 7 real world bugs in the drupal CI :-)

src/Rules/Regexp/RegularExpressionQuotingRule.php
Seldaek: Ah right you check only the calls. Never mind then. As for your example above I'd say if delimiter is { and preg quote is called with smth else you don't really need to warn as preg quote will still escape it AFAIK. It's an unnecessary arg then but not dangerous.

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,217 @@
...
+	private function getDelimiterFromConcat(Concat $concat, Scope $scope): ?string
+	{
+		if ($concat->left instanceof Concat) {
+			return $this->getDelimiterFromConcat($concat->left, $scope);
+		}
+
+		$left = $scope->getType($concat->left);
+
+		$constantStrings = $left->getConstantStrings();
+		// since we analyze a Expr directly given to a call, it can only ever resolve to a single constant string
```
ondrejmirtes: You bet it can be multiple strings. It should be handled here.

What about `$fooOrBar . preg_quote($s, '#') . $barOrBaz`.

Where `$fooOrBar` is `'foo'|'bar`.

src/Rules/Regexp/RegularExpressionQuotingRule.php
```diff
@@ -0,0 +1,217 @@
...
+		}
+		$functionReflection = $this->reflectionProvider->getFunction($pregQuote->name, $scope);
+
+		$normalizedArgs = $this->getNormalizedArgs($pregQuote, $scope, $functionReflection);
+		if ($normalizedArgs === null) {
+			return null;
+		}
+
+		if (!isset($normalizedArgs[1])) {
+			return RuleErrorBuilder::message(sprintf('Call to preg_quote() is missing delimiter %s to be effective.', $patternDelimiter))
```
ondrejmirtes: This doesn't sound really true in context of `check for delimiters which get properly escaped by default`.

If the delimiter of the regex is one of those characters in the list, then a single-arg call to `preg_quote` is OK?

tests/PHPStan/Rules/Regexp/RegularExpressionQuotingRuleTest.php
```diff
@@ -0,0 +1,85 @@
...
+					'Call to preg_quote() uses invalid delimiter / while pattern uses &.',
+					24,
+				],
+			],
+		);
+	}
+
+	public function testRulePhp8(): void
+	{
+		$this->analyse(
```
ondrejmirtes, line 79: Skip before PHP 8

ondrejmirtes (changes requested):

src/Rules/Regexp/RegularExpressionQuotingRule.php
staabm: this code was not reached for the case mentioned.

I refactored the code to make it more reasonable.

ondrejmirtes:
Thank you!

PR #3253, opened 2024-07-19
Bleeding edge - check mixed in unary operator

schlndh:
Fixes https://phpstan.org/r/36c7d1ec-c6df-4118-b288-72062b3fa963

### DIFF
--- a/tests/PHPStan/Rules/Operators/data/invalid-unary.php
+++ b/tests/PHPStan/Rules/Operators/data/invalid-unary.php
@@ -1,5 +1,5 @@
 <?php
-
+namespace InvalidUnary;
 function (
 	int $i,
 	string $str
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Rules/Operators/data/invalid-unary.php
```diff
@@ -1,5 +1,5 @@
 <?php
-
+namespace InvalidUnary;
```
schlndh, line 2: I was getting some weird unrelated test failures (probably due to adding `foo` to the global namespace), so I added the namespace here.

ondrejmirtes: Yes 👍

ondrejmirtes:
Thank you!

PR #3254, opened 2024-07-20
Fix PHP_VERSION_ID comparison in test

staabm:
problem identified via https://github.com/phpstan/phpstan-src/pull/2968/


ondrejmirtes:
Thank you.

PR #3255, opened 2024-07-20
Bleeding edge: stricter ++/-- operator check

schlndh:
The goal of this PR was to check mixed in `++`/`--`, but while I was at it I also fixed several other cases which were not previously reported (pretty much the same as the binary/unary operator PRs).

Current behavior: https://phpstan.org/r/974fc0d6-90ba-4b72-8d83-edff03ac3482

### DIFF
--- a/tests/PHPStan/Rules/Operators/data/invalid-inc-dec.php
+++ b/tests/PHPStan/Rules/Operators/data/invalid-inc-dec.php
@@ -2,7 +2,7 @@
 
 namespace InvalidIncDec;
 
-function ($a, int $i, ?float $j, string $str, \stdClass $std) {
+function ($a, int $i, ?float $j, string $str, \stdClass $std, \SimpleXMLElement $simpleXMLElement) {
 	$a++;
 
 	$b = [1];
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Rules/Operators/data/invalid-inc-dec.php
```diff
@@ -15,4 +15,41 @@ function ($a, int $i, ?float $j, string $str, \stdClass $std) {
...
+	$bool = false;
+	--$bool;
+	$null = null;
+	$null++;
+	$null = null;
+	--$null;
+	$a = $simpleXMLElement;
+	$a++;
+	$a = $simpleXMLElement;
+	--$a;
```
schlndh, line 46: This was allowed by the previous version of the rule as well (due to `ObjectType::toNumber`). And it does seem to work in PHP 8.3, but not in earlier versions. However, I don't see this change mentioned in the documentation. I'm not entirely sure how best to handle it, so I kept it allowed.

https://3v4l.org/DVmtc

tests/PHPStan/Rules/Operators/data/invalid-inc-dec.php
```diff
@@ -15,4 +15,41 @@ function ($a, int $i, ?float $j, string $str, \stdClass $std) {
 	$j++;
 	$str++;
 	$std++;
+	$classWithToString = new ClassWithToString();
+	$classWithToString++;
+	$classWithToString = new ClassWithToString();
+	--$classWithToString;
```
schlndh, line 21: ~~Bug:~~ This does work in PHP 7: https://3v4l.org/oXvs8

schlndh: It actually works even without `__toString()`, but I'm not sure whether it can be used for anything useful with custom classes: https://3v4l.org/uXLrP I'd prefer to keep it disabled even for PHP 7 unless there is a use-case for it.

ondrejmirtes:
Thank you!

PR #3256, opened 2024-07-21
Implement PregMatchAllParameterOutTypeExtension

staabm:
Starting in small steps, as I think `preg_match_all` has some edge-cases each worth a separate PR ;)


staabm:
for a green build this PR requires https://github.com/phpstan/phpstan-src/pull/3257

Seldaek:
Nice! Doesn't look like too much code so far, which is also what I expected. The return type is different because it's an array of arrays of matches, but otherwise the big chunk of the pattern extraction etc is the same really.

ondrejmirtes (changes requested):
1) All the flags described here https://www.php.net/manual/en/function.preg-match-all.php need implementation and tests.
2) We also need TypeSpecifyingExtension for code like:

* `if (preg_match_all($x, $y, $z))`
* `if (preg_match_all($x, $y, $z) > 0)`
* `if (preg_match_all($x, $y, $z) != false)` (loose `!=` on purpose)
* `if (preg_match_all($x, $y, $z) == true)` (two `==` on purpose)
* `if (preg_match_all($x, $y, $z) === 1)` (any number higher than 0)
* Falsy contexts of all of the above

ondrejmirtes:
I'm also looking forward to https://www.php.net/manual/en/function.preg-replace-callback.php because I believe that's going to take advantage of the new-ish FunctionParameterClosureTypeExtension :)

staabm:
I think I am at a point where I don't see any mistakes anymore, because I am so deep into it.
lets see what you gyus think about it

ondrejmirtes:
Thank you.

PR #3257, opened 2024-07-21
RegexArrayShapeMatcher - `(\d*)` can be empty-string

staabm:


staabm:
//cc @Seldaek

tests/PHPStan/Analyser/nsrt/bug-11311.php
```diff
@@ -187,3 +187,13 @@ function (string $s): void {
...
 };
+
+function (string $s): void {
+	preg_match('/%a(\d*)/', $s, $matches, PREG_UNMATCHED_AS_NULL);
+	assertType("array{0?: string, 1?: ''|numeric-string|null}", $matches); // could be array{0?: string, 1?: ''|numeric-string}
+};
+
+function (string $s): void {
+	preg_match('/%a(\d*)?/', $s, $matches, PREG_UNMATCHED_AS_NULL);
+	assertType("array{0?: string, 1?: ''|numeric-string|null}", $matches); // could be array{0?: string, 1?: ''|numeric-string}
```
Seldaek, line 199: Good catch! What's the matter here tho?

staabm: in this case php-src does not return `NULL`  for the unmatched group even if `PREG_UNMATCHED_AS_NULL`.

that would be another special case we need to hardcode - therefore I went with leaving it less precise and have this unnecessary `null` in here.

I have a feeling that the ArrayShapeMatcher will get enough special cases with [`preg_match_all` support](https://github.com/phpstan/phpstan-src/pull/3256) ;)

ondrejmirtes:
Thank you!

tests/PHPStan/Analyser/nsrt/bug-11311.php
Seldaek: Yes ok, it should be `''|numeric-string`, or unset if outside the true-match-arm.

https://3v4l.org/Tgk4m

PREG_UNMATCHED_AS_NULL has no impact here because `(\d*)` matches an empty string too, so it always matches. It's technically not optional despite the `?`  that follows, as long as it is not wrapped in something that contains some content.. e.g. the third example here https://3v4l.org/92geX

So I guess one rule could be that if the inside of a match group resolves to MAYBE empty string, then the match group should never be nullable as long as its parents are also not nullable.

PR #3258, opened 2024-07-22
Add extended reasons tips to param out errors

thg2k:
This adds the very useful tips on complex types also to param-out.

Please carefully review the correctness, I assume that isSuperTypeOf and acceptsWithReason are equivalent.

Also not sure how to set the strictness parameter, as it is a param-out it should not be affected by the strict types declare.

Before:
```
 ------ ----------------------------------------------------------------------------------------------------------------------------
  Line   case2.php
 ------ ----------------------------------------------------------------------------------------------------------------------------
  8      Parameter &$foo by-ref type of function foo() expects array{a: int, b: int, c: string}, array{a: int, b: int, c: 5} given.
           parameterByRef.type
          You can change the parameter out type with @param-out PHPDoc tag.
 ------ ----------------------------------------------------------------------------------------------------------------------------
```

After:
```
 ------ ----------------------------------------------------------------------------------------------------------------------------
  Line   case2.php
 ------ ----------------------------------------------------------------------------------------------------------------------------
  8      Parameter &$foo by-ref type of function foo() expects array{a: int, b: int, c: string}, array{a: int, b: int, c: 5} given.
           parameterByRef.type
          Offset 'c' (string) does not accept type int.
          You can change the parameter out type with @param-out PHPDoc tag.
 ------ ----------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/src/Rules/Variables/ParameterOutAssignedTypeRule.php
+++ b/src/Rules/Variables/ParameterOutAssignedTypeRule.php
@@ -108,10 +108,15 @@ class ParameterOutAssignedTypeRule implements Rule
 			$functionDescription,
 			$outType->describe($verbosityLevel),
 			$assignedExprType->describe($verbosityLevel),
-		))->identifier(sprintf('%s.type', $isParamOutType ? 'paramOut' : 'parameterByRef'));
+		))->identifier($isParamOutType ? 'paramOut.type' : 'parameterByRef.type');
+
+		$accepts = $outType->acceptsWithReason($assignedExprType, true);
+		if (!$accepts->result->yes()) {
+			$errorBuilder->acceptsReasonsTip($accepts->reasons);
+		}
 
 		if (!$isParamOutType) {
-			$errorBuilder->tip('You can change the parameter out type with @param-out PHPDoc tag.');
+			$errorBuilder->addTip('You can change the parameter out type with @param-out PHPDoc tag.');
 		}
 
 		return [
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
`accepts` and `isSuperTypeOf` are not equivalent. The former is about "Does this type accept that type?" and the latter one is "Is this type part of that type?"

Some practical differences:

* Float accepts any int, but float isn't an int
* Without `strict_types=1`, object of class with `__toString` (Stringable) is accepted in a `string`.
* Generics acceptance rules are stricter based on TemplateTypeStrategy

What we could do is implement `isSuperTypeOfWithReason`. We could use that in all the rules where `isSuperTypeOf` is used.

Another example besides `@param-out` where this would be useful is this issue: https://github.com/phpstan/phpstan/issues/10421

thg2k:
To be very honest I did not fully understand your answer 😄

Is it correct to assume that isSuperTypeOf()->yes() fully implies acceptance? Because in this case I can do it very dirty and attempt to pull the reasons from acceptsWithReason only if isSuperTypeOf check fails.

This way the semantic should be preserved and we still give the user some insights when available, worst case the message stays the same. Please check the updated PR.

ondrejmirtes:
One piece I forgot to mention: we're using isSuperTypeOf because no native type casting is involved, only PHPDocs. So the type must match otherwise you might end up with inconvenient surprise down the road.

I don't want a hack like that, that's not sustainable. It's worth investing in proper solutions.

thg2k:
It is a hack indeed, but the isSuperTypeWithReason implementation is too big of an endeavor for me, so feel free to close the PR.

ondrejmirtes:
I worry that the hack would show non-sensical tips we wouldn't have under control. A proper solution would also have other benefits.

thg2k:
Note to self:
See also https://github.com/phpstan/phpstan/discussions/11460

PR #3259, opened 2024-07-22
Apply the stricter signatures for php 8.0 after the basic ones

thg2k:
This change is required to make the stricter map effective.


staabm:
do you have a failling test-case about the bug you try to fix?

looking at e.g. https://github.com/thg2k/phpstan-src/commit/f46e4700de024cb4fc13d3ab0ba3c0b56ffb9899 looks like the `resources/functionMap_php80delta_bleedingEdge.php` works and overrides signatures from `resources/functionMap_bleedingEdge.php`

thg2k:
> do you have a failling test-case about the bug you try to fix?
> 
> looking at e.g. [thg2k@f46e470](https://github.com/thg2k/phpstan-src/commit/f46e4700de024cb4fc13d3ab0ba3c0b56ffb9899) looks like the `resources/functionMap_php80delta_bleedingEdge.php` works and overrides signatures from `resources/functionMap_bleedingEdge.php`

That's because error_log does not appear in the previous deltas, otherwise first the signature is loaded from the functionMap_php80delta_bleedingEdge.php file and then overwritten by the functionMap_php80delta.php. In this case, error_log does not appear there.

I noticed because imagecreatetruecolor is overridden on both files, resulting in the wrong signatures with bleeding edge enabled.

ondrejmirtes:
Thank you!

PR #3262, opened 2024-07-23
Refactor function signature map files loading

thg2k:
I apologize for the not-so-great PR, I understand that this class might need a more serious refactoring, but this is just to reduce the copy-paste and make it a bit more readable.

### DIFF
--- a/src/Reflection/SignatureMap/FunctionSignatureMapProvider.php
+++ b/src/Reflection/SignatureMap/FunctionSignatureMapProvider.php
@@ -184,35 +184,35 @@ class FunctionSignatureMapProvider implements SignatureMapProvider
 		$signatureMap = array_change_key_case($signatureMap, CASE_LOWER);
 
 		if ($this->stricterFunctionMap) {
-			$signatureMap = $this->computeSignatureMapFile($signatureMap, 'functionMap_bleedingEdge.php');
+			$signatureMap = $this->computeSignatureMapFile($signatureMap, __DIR__ . '/../../../resources/functionMap_bleedingEdge.php');
 		}
 
 		if ($this->phpVersion->getVersionId() >= 70400) {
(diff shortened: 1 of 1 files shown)


src/Reflection/SignatureMap/FunctionSignatureMapProvider.php
```diff
@@ -184,80 +184,54 @@ public function getSignatureMap(): array
 		$signatureMap = array_change_key_case($signatureMap, CASE_LOWER);
 
 		if ($this->stricterFunctionMap) {
-			$stricterFunctionMap = require __DIR__ . '/../../../resources/functionMap_bleedingEdge.php';
-			if (!is_array($stricterFunctionMap)) {
-				throw new ShouldNotHappenException('Signature map could not be loaded.');
-			}
-
-			$signatureMap = $this->computeSignatureMap($signatureMap, $stricterFunctionMap);
+			$signatureMap = $this->computeSignatureMapFile($signatureMap, 'functionMap_bleedingEdge.php');
```
ondrejmirtes: This makes the file not-clickable in the IDE which is worse experience.

ondrejmirtes (changes requested):

src/Reflection/SignatureMap/FunctionSignatureMapProvider.php
thg2k: I didn't think about that. This can be easily fixed, see follow up

thg2k:
To solve the IDE clickability issue, does it also need the "require" keyword together with the file or the full path is enough?

Also notice I changed require to include, that is to get the more meaningful exception in case the file is missing for whatever reason, instead of just a php error.

ondrejmirtes:
Thank you!

PR #3263, opened 2024-07-24
Improve `abs()` return type

julienfalque:
Fixes https://github.com/phpstan/phpstan/issues/11367.

### DIFF
--- a/src/Type/Accessory/AccessoryLiteralStringType.php
+++ b/src/Type/Accessory/AccessoryLiteralStringType.php
@@ -171,6 +171,11 @@ class AccessoryLiteralStringType implements CompoundType, AccessoryType
 		return new ErrorType();
 	}
 
+	public function toAbsoluteNumber(): Type
+	{
+		return new ErrorType();
+	}
+
 	public function toInteger(): Type
 	{
 		return new IntegerType();
--- a/src/Type/Accessory/AccessoryNonEmptyStringType.php
+++ b/src/Type/Accessory/AccessoryNonEmptyStringType.php
@@ -177,6 +177,11 @@ class AccessoryNonEmptyStringType implements CompoundType, AccessoryType
 		return new ErrorType();
 	}
 
+	public function toAbsoluteNumber(): Type
+	{
+		return new ErrorType();
+	}
+
 	public function toInteger(): Type
 	{
 		return new IntegerType();
--- a/src/Type/Accessory/AccessoryNonFalsyStringType.php
+++ b/src/Type/Accessory/AccessoryNonFalsyStringType.php
@@ -173,6 +173,11 @@ class AccessoryNonFalsyStringType implements CompoundType, AccessoryType
 		return new ErrorType();
 	}
 
+	public function toAbsoluteNumber(): Type
+	{
+		return new ErrorType();
+	}
+
 	public function toInteger(): Type
 	{
 		return new IntegerType();
--- a/src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php
@@ -38,49 +38,53 @@ class AbsFunctionDynamicReturnTypeExtension implements DynamicFunctionReturnType
 		$type = $scope->getType($args[0]->value);
 
 		if ($type instanceof UnionType) {
-			$ranges = [];
+			$absUnionTypes = [];
 
 			foreach ($type->getTypes() as $unionType) {
-				if (
-					!$unionType instanceof ConstantIntegerType
-					&& !$unionType instanceof IntegerRangeType
-					&& !$unionType instanceof ConstantFloatType
-				) {
+				$absUnionType = $this->tryAbsType($unionType);
+
+				if ($absUnionType === null) {
 					return null;
 				}
 
-				$absRange = $this->absType($unionType);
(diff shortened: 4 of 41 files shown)


src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		$type = $scope->getType($args[0]->value);
+
+		if ($type instanceof UnionType) {
+			$ranges = [];
+
+			foreach ($type->getTypes() as $unionType) {
+				if (
+					!$unionType instanceof ConstantIntegerType
```
staabm: I think we could also support constant numeric strings:

https://3v4l.org/9eXpi

src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,115 @@
...
+	{
+		$args = $functionCall->getArgs();
+
+		if (!isset($args[0])) {
+			return null;
+		}
+
+		$type = $scope->getType($args[0]->value);
+
+		if ($type instanceof UnionType) {
```
staabm: this if/else construction feels like something which can be simplified with `TypeTraverser::map`

src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php
julienfalque: Done.

src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php
julienfalque: I'm not sure this will simplify things for union types:
* there's a case where the loop returns `null` to avoid overriding the return type of `abs()`. We would need to emulate this from the callback passed to `TypeTraverser::map` which cannot return `null`.
* the loop tries to merge ranges when possible, e.g. `int<-15, -5>|int<10, 20>` → `int<5, 20>`. This does not seem to be trivial to do with `TypeTraverser::map`

ondrejmirtes (changes requested):
Many `instanceof` questions about types. This will no doubt be imperfect. The need for such code usually means there should a method on `Type` instead which will force us to ask the right question in the right places.

Something like `Type::getAbsoluteNumericValue(): Type` is the way to go.

src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php
julienfalque: Solved with last refactoring.

julienfalque:
Added a `Type::toAbsoluteNumber()` method, this definitely improve things. Let me know if the naming is OK. Also, I'm not certain all implementations are correct.

src/Type/Accessory/AccessoryLiteralStringType.php
```diff
@@ -171,6 +171,11 @@ public function toNumber(): Type
 		return new ErrorType();
 	}
 
+	public function toAbsoluteNumber(): Type
+	{
+		return new ErrorType();
```
staabm, line 176: you should add a test for `abs('123')` and `abs('hello')` and fix this type

Also add a test for 
```
/** @var literal-string $s */
abs($s);
```

src/Type/Accessory/AccessoryNonEmptyStringType.php
```diff
@@ -177,6 +177,11 @@ public function toNumber(): Type
 		return new ErrorType();
 	}
 
+	public function toAbsoluteNumber(): Type
+	{
+		return new ErrorType();
```
staabm, line 182: a non-empty string could be a string with numbers, so its not an error

src/Type/Accessory/AccessoryNonFalsyStringType.php
```diff
@@ -173,6 +173,11 @@ public function toNumber(): Type
 		return new ErrorType();
 	}
 
+	public function toAbsoluteNumber(): Type
+	{
+		return new ErrorType();
```
staabm, line 178: simlar to `AccessoryNonEmptyStringType.php`

src/Type/Accessory/AccessoryLiteralStringType.php
julienfalque: I added a few tests but they passed without any code change.

src/Type/Accessory/AccessoryNonEmptyStringType.php
julienfalque: I understand that but then why does `toNumber()` return an `ErrorType`? Shouldn't this be changed as well?

julienfalque:
By the way, `ErrorType` inherits from `MixedType` so:

* `toNumber()` returns `int|float`;
* `toAbsoluteNumber()` returns `int<0, max>|float`.

Shouldn't both methods return `ErrorType`?

src/Type/Accessory/AccessoryLiteralStringType.php
staabm: I mixed it up. thanks for adding the tests. it seems returning the ErrorType here is correct (and in the other comments)

ondrejmirtes:
I love it, thanks!

PR #3265, opened 2024-07-25
RegexArrayShapeMatcher: Support resolving of constants in patterns

staabm:
Refactors RegexArrayShapeMatcher to re-use InitializerExprTypeResolver

closes https://github.com/phpstan/phpstan/issues/11384

### DIFF
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -13,6 +13,7 @@ use PhpParser\Node\Expr;
 use PhpParser\Node\Name;
 use PHPStan\Analyser\Scope;
 use PHPStan\Php\PhpVersion;
+use PHPStan\Reflection\InitializerExprTypeResolver;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
@@ -55,6 +56,7 @@ final class RegexArrayShapeMatcher
 
 	public function __construct(
 		private PhpVersion $phpVersion,
+		private InitializerExprTypeResolver $initializerExprTypeResolver,
 	)
 	{
 	}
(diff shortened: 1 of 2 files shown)


staabm:
I think I now have a better understand how the `InitializerExprTypeResolver` works and I will have a look whether we can do a similar thing within `RegularExpressionQuotingRule`

Seldaek:
Cool, less code and less bugs :D

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -724,38 +726,33 @@ private function getPatternType(Expr $patternExpr, Scope $scope): Type
...
-			&& $concat->right->name instanceof Name
-			&& $concat->right->name->toLowerString() === 'preg_quote'
-		) {
-			$right = new ConstantStringType('');
-		} elseif ($concat->right instanceof Expr\BinaryOp\Concat) {
-			$right = $this->resolvePatternConcat($concat->right, $scope);
-		} else {
-			$right = $scope->getType($concat->right);
-		}
+		$resolveConcat = static function (Expr $expr) use (&$resolveConcat, $scope) {
```
ondrejmirtes: These are usually memory leaks. It'd be nicer to have this as a private method.

ondrejmirtes (changes requested):

src/Type/Php/RegexArrayShapeMatcher.php
staabm: I went with a inline class, hopefully that works for you.
didn't have a better idea on how to get the lexical $scope variable into the callable

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -724,38 +726,42 @@ private function getPatternType(Expr $patternExpr, Scope $scope): Type
...
+				}
+
+				return $this->scope->getType($expr);
 			}
-		}
 
-		return TypeCombinator::union(...$strings);
+		};
+
+		return $this->initializerExprTypeResolver->getConcatType($concat->left, $concat->right, static fn (Expr $expr): Type => $resolver->resolve($expr));
```
Seldaek, line 764: Not sure if this is supported by the php8.1=>7.2 transpile but hopefully :)

```suggestion
		return $this->initializerExprTypeResolver->getConcatType($concat->left, $concat->right, $resolver->resolve(...));
```

ondrejmirtes: It's not :)

ondrejmirtes:
Thank you.

PR #3266, opened 2024-07-25
UselessFunctionReturnValueRule: More precise error message

staabm:
closes https://github.com/phpstan/phpstan/issues/11377

### DIFF
--- a/src/Rules/Functions/UselessFunctionReturnValueRule.php
+++ b/src/Rules/Functions/UselessFunctionReturnValueRule.php
@@ -10,8 +10,8 @@ use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use function array_key_exists;
 use function count;
-use function in_array;
 use function sprintf;
 
 /**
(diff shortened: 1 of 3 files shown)


takaram:
`highlight_string` never returns false while its return type is declared as `string|bool`
PHP 8.4 is going to change it to `string|true`.
https://github.com/php/php-src/blob/php-8.4.0alpha2/NEWS#L26-L27

`highlight_file` and `show_source` may return both true and false, but it means that using the return value is not useless, doesn't it?

staabm:
> `highlight_file` and `show_source` may return both true and false, but it means that using the return value is not useless, doesn't it?

I am not sure. this might mean we should drop these 2 from the rule

ondrejmirtes (changes requested):
Yes, highlight_file and show_source should be dropped from this rule. It's more complicated with them.

Instead, these functions could have a stub file with a conditional return type.

Then, someone is free to write a rule that disallows `bool` and potentially other types to be used in a string context, like concat and others. (But it's a bad fit for PHPStan core because it's not a bug.)

staabm:
> Yes, highlight_file and show_source should be dropped from this rule. It's more complicated with them.

done

> Instead, these functions could have a stub file with a conditional return type.

we already have [these stubs](https://github.com/phpstan/phpstan-src/blob/bd2cec118592f7c66dff5a7ae28882654daf6468/stubs/core.stub#L26-L36)

ondrejmirtes:
Thank you.

PR #3268, opened 2024-07-25
Regex parsing fixes for newline characters and marker verbs

Seldaek:
- Fixes handling of actual newlines in regexes, refs phpstan/phpstan#11360
- Fixes parsing for marker verbs, fixes phpstan/phpstan#8948

cc @staabm @mvorisek

### DIFF
--- a/resources/RegexGrammar.pp
+++ b/resources/RegexGrammar.pp
@@ -41,10 +41,6 @@
 // @license    New BSD License
 //
 
-
-// Skip.
-%skip   nl                       \n
-
 // Character classes.
 %token  negative_class_          \[\^               -> class
 %token  class_                   \[                 -> class
@@ -58,7 +54,7 @@
 %token class:character           \\([aefnrtb]|c[\x00-\x7f])
 %token class:dynamic_character   \\([0-7]{3}|x[0-9a-zA-Z]{2}|x{[0-9a-zA-Z]+})
 %token class:character_type      \\([CdDhHNRsSvVwWX]|[pP]{[^}]+})
-%token class:literal             \\.|.
+%token class:literal             \\.|.|\n
 
 // Internal options.
 // See https://www.regular-expressions.info/refmodifiers.html
@@ -122,7 +118,7 @@
 %token character_type            \\([CdDhHNRsSvVwWX]|[pP]{[^}]+})
 %token anchor                    \\([bBAZzG])|\^|\$
 %token match_point_reset         \\K
-%token literal                   \\.|.
+%token literal                   \\.|.|\n
 
 
 // Rules.
--- a/resources/RegexGrammar.pp
+++ b/resources/RegexGrammar.pp
@@ -78,6 +78,11 @@
 %token  co:_comment              \)                 -> default
 %token  co:comment               .*?(?=(?<!\\)\))
 
+// Marker verbs
+%token  marker_                  \(\*:              -> mark
+%token  mark:name                [^)]+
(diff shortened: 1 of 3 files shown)


src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -325,6 +329,14 @@ private function buildArrayType(
 			$i++;
 		}
 
+		if ($hasMark) {
+			$builder->setOffsetValueType(
+				$this->getKeyType('MARK'),
+				new IntersectionType([new StringType(), new AccessoryNonEmptyStringType()]),
```
mvorisek: why is mark capture non-empty string?

Seldaek: Because it captures the name of the mark which can't be empty, it's just whatever follows until)

https://3v4l.org/BLiiT

mvorisek: Let's union the parsed mark constant strings instead.

resources/RegexGrammar.pp
```diff
@@ -122,7 +123,7 @@
 %token character_type            \\([CdDhHNRsSvVwWX]|[pP]{[^}]+})
 %token anchor                    \\([bBAZzG])|\^|\$
 %token match_point_reset         \\K
-%token literal                   \\.|.
+%token literal                   \\.|.|\n
```
staabm, line 126: Do we need \r here, too?

\r alone on the other hand might be rare enough that its not worth parsing..

Seldaek: `\r` is very much worth parsing because even in `\r\n` if you can't parse `\r` then you fail, but it is included in `.` so all good.

src/Type/Php/RegexArrayShapeMatcher.php
Seldaek: Done in https://github.com/phpstan/phpstan-src/pull/3268/commits/ddd34d1222614e14bc9ad08b485e417095b96862

mvorisek: Great ❤!

staabm (approved):
lgtm - thanks

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -518,6 +518,12 @@ function bug11323(string $s): void {
 	if (preg_match('{([^1-4])}', $s, $matches)) {
 		assertType('array{string, non-empty-string}', $matches);
 	}
+	if (preg_match("{([\r\n]+)(\n)([\n])}", $s, $matches)) {
+		assertType('array{string, non-empty-string, non-empty-string, non-empty-string}', $matches);
+	}
+	if (preg_match('/foo(*:first)|bar(*:second)([x])/', $s, $matches)) {
+		assertType("array{0: string, 1?: non-empty-string, MARK?: 'first'|'second'}", $matches);
```
staabm, line 525: should MARK be non-optional since we are sure here about the successfull match?
(in other words: should it be optional/non-optional depending on `$wasMatched` (or other alternation rules)?

Seldaek: I left it as such because it depends where the mark is https://3v4l.org/vJofm - and it seems to me not really worth the effort to ensure whether it's in a match group that is optional or not, given how niche this feature is.

ondrejmirtes:
Thank you!

PR #3270, opened 2024-07-26
RegularExpressionPatternRule: validate preg_quote'd patterns

staabm:
similar to how we added support for preg_quote'd patterns in ArrayShapeMatcher, we now re-use the logic in RegularExpressionPatternRule

Review at best commit by commit
commit 1: refactoring of ArrayShapeMatcher for re-use
commit 2: adjust RegularExpressionPatternRule to make use of the refactoring


ondrejmirtes:
Thank you.

PR #3271, opened 2024-07-27
RegexArrayShapeMatcher - Support 'n' modifier

staabm:
review commit by commit

### DIFF
--- a/src/Type/Php/RegexExpressionHelper.php
+++ b/src/Type/Php/RegexExpressionHelper.php
@@ -3,12 +3,14 @@
 namespace PHPStan\Type\Php;
 
 use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\BinaryOp\Concat;
 use PhpParser\Node\Name;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\InitializerExprTypeResolver;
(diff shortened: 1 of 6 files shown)


staabm:
@Seldaek with this PR we can now take regex pattern modifiers into account.
this should help with implementing https://github.com/phpstan/phpstan/issues/11360

ondrejmirtes:
Thank you!

Please next time instead of separate commits, send separate PRs.

src/Type/Php/RegexExpressionHelper.php
```diff
@@ -66,4 +69,59 @@ public function resolve(Expr $expr): Type
...
 	}
 
+	public function getPatternModifiers(string $pattern): ?string
+	{
+		$delimiter = $this->getDelimiterFromString(new ConstantStringType($pattern));
+		if ($delimiter === null) {
+			return null;
+		}
+
+		if ($delimiter === '{') {
```
Seldaek, line 79: () and [] are also valid reciprocal delimiters.
<> too but that's a shitty one to use really.. Still probably should support for completeness

https://www.php.net/manual/en/regexp.reference.delimiters.php

Seldaek (commented):
Looks good to me aside from some missing delimiter support

PR #3273, opened 2024-07-27
RegexExpressionHelper - Support all bracket style delimiters

staabm:
refs https://github.com/phpstan/phpstan-src/pull/3271#discussion_r1694000252

I was not aware these variants exist

### DIFF
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -406,8 +406,9 @@ final class RegexArrayShapeMatcher
 			return null;
 		}
 
+		$rawRegex = $this->regexExpressionHelper->removeDelimitersAndModifiers($regex);
 		try {
-			$ast = self::$parser->parse($regex);
+			$ast = self::$parser->parse($rawRegex);
 		} catch (Exception) {
 			return null;
 		}
--- a/src/Type/Php/RegexExpressionHelper.php
+++ b/src/Type/Php/RegexExpressionHelper.php
@@ -72,9 +72,31 @@ final class RegexExpressionHelper
 
 	public function getPatternModifiers(string $pattern): ?string
 	{
-		$delimiter = $this->getDelimiterFromString(new ConstantStringType($pattern));
-		if ($delimiter === null) {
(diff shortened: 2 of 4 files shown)


staabm:
//cc @Seldaek 

I was not able to add a test for `[(\d+)(?P<num>\d+)]n`

```
function (string $s): void {
	if (preg_match('[(\d+)(?P<num>\d+)]n', $s, $matches)) {
		assertType('array{0: string, num: numeric-string, 1: numeric-string}', $matches);
	}
};
```

 because the hoa ast ist different for it in comparison to the other bracket style delims

```
<?php

require __DIR__.'/vendor/autoload.php';

// 1. Read the grammar.
$grammar  = new Hoa\File\Read(__DIR__.'/resources/RegexGrammar.pp');

// 2. Load the compiler.
$compiler = Hoa\Compiler\Llk\Llk::load($grammar);

// 3. Lex, parse and produce the AST.
$ast      = $compiler->parse('[(\d+)(?P<num>\d+)]n');

// 4. Dump the result.
$dump     = new Hoa\Compiler\Visitor\Dump();
echo $dump->visit($ast);

```

Seldaek:
Yeah that's because the parser has no concept of delimiters.. And supporting those delimiters would probably be a pain tbh. So I guess let's leave it as is for now.

Seldaek:
On second thought I guess actually what you should do is probably parse delimiters first, remove delimiters/modifiers, then pass the actual pattern to the parser. That'd probably be the more accurate way to use it. Because otherwise when using `()` as delimiters, aren't you ending up with one match group too many?

staabm:
> On second thought I guess actually what you should do is probably parse delimiters first, remove delimiters/modifiers, then pass the actual pattern to the parser.

nice catch. looking at [hoa-regex README](https://github.com/staabm/HoaRegex) the example actually also doesn't use modifiers/delimiters

ondrejmirtes:
Thank you!

BTW: Could you please look into `\Q...\E`? https://stackoverflow.com/questions/20518758/how-the-preg-match-handles-the-delimiter-when-q-e-used

Can't find the issue right now but someone tried to use it in `ignoreErrors` too, and the validator rejected it.

PR #3275, opened 2024-07-28
RegularExpressionPatternRule: fix false positive in preg_quote() handling

staabm:
closes https://github.com/phpstan/phpstan/issues/11403

### DIFF
--- a/src/Type/Php/RegexExpressionHelper.php
+++ b/src/Type/Php/RegexExpressionHelper.php
@@ -44,7 +44,7 @@ final class RegexExpressionHelper
 					&& $expr->name instanceof Name
 					&& $expr->name->toLowerString() === 'preg_quote'
 				) {
-					return new ConstantStringType('');
+					return new ConstantStringType('.*');
 				}
 
 				if ($expr instanceof Concat) {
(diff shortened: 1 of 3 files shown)


staabm:
//cc @Seldaek

src/Type/Php/RegexExpressionHelper.php
```diff
@@ -44,7 +44,7 @@ public function resolve(Expr $expr): Type
 					&& $expr->name instanceof Name
 					&& $expr->name->toLowerString() === 'preg_quote'
 				) {
-					return new ConstantStringType('');
+					return new ConstantStringType('.*');
```
staabm, line 47: instead of dummy-replace `preg_quote($x)` with `''` we use `.*` as a placeholder, so we don't produce invalid patterns

ondrejmirtes:
Thank you.

src/Type/Php/RegexExpressionHelper.php
mvorisek: Can you please explain the constant string meaning? AFAICT such assumption here can produce misleading assumptions/phpstan errors.

staabm: We replace a preg_quote with some constant value so we can finally resolve a string concatenation to a pattern at analysis time. 

The preg_quoted value cannot return something which creates capturing groups or syntax errors, therefore we simplify it to improve coverage

Seldaek: It'd probably help to have some more comments explaining these edge cases / hacks to future archeologists 😅 but PR looks good thanks

PR #3280, opened 2024-08-02
RegularExpressionPatternRule: fix false positive in preg_quote() handling

staabm:
closes https://github.com/phpstan/phpstan/issues/11432

### DIFF
--- a/src/Type/Php/RegexExpressionHelper.php
+++ b/src/Type/Php/RegexExpressionHelper.php
@@ -40,12 +40,16 @@ final class RegexExpressionHelper
 
 			public function resolve(Expr $expr): Type
 			{
+				// assume preg_quote() cannot create capturing groups or contain meta characters.
+				// replace it with a pattern which matches anything, does not affect $matches results
+				// and does not produce regex errors when followed by a quantifier.
+				// this allows us to turn string concatenations with preg_quote() into static analyzable strings.
 				if (
 					$expr instanceof Expr\FuncCall
 					&& $expr->name instanceof Name
 					&& $expr->name->toLowerString() === 'preg_quote'
 				) {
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you.

ondrejmirtes:
Some new error appeared based on this, can you look if it's valid or not? https://github.com/phpstan/phpstan/actions/runs/10226167599/job/28296101528

staabm:
> Some new error appeared based on this, can you look if it's valid or not? [phpstan/phpstan/actions/runs/10226167599/job/28296101528](https://github.com/phpstan/phpstan/actions/runs/10226167599/job/28296101528)

thats a interessting case: https://github.com/PrestaShop/PrestaShop/blob/ca9d81aae0bb17f3e767bb5d835348ed144a3ab5/src/Adapter/Assets/AssetUrlGeneratorTrait.php#L57

@Seldaek do you agree that for proper escaping the PrestaShop code would need a tripe backslash like in https://3v4l.org/8fsWi ?

another example - which IMO also proves the regex in PrestShop is missing a 3rd backslash - is https://3v4l.org/Td9vj

Seldaek:
Yeah that's a case of needing 4 backslashes imo.. You really want one in the pattern so you have to escape it as 2 and then each of those again escaped for the php string. Technically 3 works the same i suppose as long as it's not double quoted strings and depending what follows the 3rd.

ondrejmirtes:
FYI this is currently blocking me from release, this bug is still present and generating errors in Prestashop.

staabm:
As I understood it, its a bug in prestashop because they wrongly escape.

ondrejmirtes:
OK, thanks!

PR #3281, opened 2024-08-02
Bleeding edge - Precise array shape for preg_replace_callback() $matches

staabm:
will do `preg_replace_callback_array` in a separate PR

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -295,6 +295,8 @@ conditionalTags:
 		phpstan.typeSpecifier.functionTypeSpecifyingExtension: %featureToggles.narrowPregMatches%
 	PHPStan\Type\Php\PregMatchParameterOutTypeExtension:
 		phpstan.functionParameterOutTypeExtension: %featureToggles.narrowPregMatches%
+	PHPStan\Type\Php\PregMatchCallbackClosureTypeExtension:
+		phpstan.functionParameterClosureTypeExtension: %featureToggles.narrowPregMatches%
 
 services:
 	-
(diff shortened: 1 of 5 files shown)


staabm:
//cc @Seldaek @mvorisek

ondrejmirtes:
Thank you.

ondrejmirtes:
Could/should this be improved? https://phpstan.org/r/de0085e1-29c0-4346-849a-79948ec1c385

It's code from phpdoc-parser and copied from PHP-Parser I think, of course I don't understand it at all.

Seldaek:
Sounds feasible in theory to me, as you have a branch of the match group 1 which is "u<dynamicstring>", but that'd require a way to have partially static strings where you can look up index 0 and if index0 is known then we return it if not then it's just string or ''. That part you have to say if feasible.

staabm:
I don't understand the regex either but I hope I got you right:

we would need dependent types, because `$matches[2]` only always exists after `if ($str[0] === 'u')` - right?
I think this kind of dependent type is not even possible in the plain PHPStan type-system (a type depending on the offset of a string)?

ondrejmirtes:
Not dependent types, but something that is called tagged unions. Those are supported: https://phpstan.org/r/90cb0d94-b149-4b37-8c20-e5f7d984a160

Do you think the type here can be expressed like that?

staabm:
oh I see. I think we do similar things already in 

https://github.com/phpstan/phpstan-src/blob/4755b374f706b766aa4b724344a9dc5408dac055/src/Type/Php/RegexArrayShapeMatcher.php#L130-L206

I will have a look

PR #3284, opened 2024-08-04
Support literal strings in RegexGroupParser

staabm:
by detecting static string literals in the regex pattern, we make constant-array result-types of the [patterns more different](https://github.com/phpstan/phpstan/issues/11443#issuecomment-2266709936).
this helps the TypeCombinator to produce tagged unions more often, which can result in more precise overall types.

first step into the direction described in https://github.com/phpstan/phpstan/issues/11443#issuecomment-2266700800

### DIFF
--- a/src/Type/Php/RegexGroupParser.php
+++ b/src/Type/Php/RegexGroupParser.php
@@ -113,12 +113,12 @@ final class RegexGroupParser
 		array &$groupCombinations,
 		array &$markVerbs,
 		bool $captureOnlyNamed,
-		bool $repeatedMoreThenOnce,
+		bool $repeatedMoreThanOnce,
 	): void
 	{
 		$group = null;
 		if ($ast->getId() === '#capturing') {
-			$maybeConstant = !$repeatedMoreThenOnce;
+			$maybeConstant = !$repeatedMoreThanOnce;
 			if ($parentGroup !== null && $parentGroup->resetsGroupCounter()) {
 				$maybeConstant = false;
 			}
@@ -133,7 +133,7 @@ final class RegexGroupParser
 			);
 			$parentGroup = $group;
 		} elseif ($ast->getId() === '#namedcapturing') {
-			$maybeConstant = !$repeatedMoreThenOnce;
+			$maybeConstant = !$repeatedMoreThanOnce;
 			if ($parentGroup !== null && $parentGroup->resetsGroupCounter()) {
 				$maybeConstant = false;
 			}
@@ -175,7 +175,7 @@ final class RegexGroupParser
 			}
 
 			if ($max === null || $max > 1) {
-				$repeatedMoreThenOnce = true;
+				$repeatedMoreThanOnce = true;
 			}
 		}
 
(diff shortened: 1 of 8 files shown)


staabm:
phpstan-nette needs some expectation adjustments after merge. I will work on it after this one landed.

src/Type/Php/RegexGroupParser.php
```diff
@@ -111,28 +113,39 @@ private function walkRegexAst(
 		array &$groupCombinations,
 		array &$markVerbs,
 		bool $captureOnlyNamed,
+		bool $repeatedMoreThenOnce,
```
ondrejmirtes: Than

ondrejmirtes (changes requested):

ondrejmirtes:
This is becoming ridiculously good! Thank you.

Seldaek:
Woah this is awesome! Makes me kinda want to handle foo? as 'fo'|'foo' for example.. And even character classes with very few possible outcomes could be done that way like [abc] can just be one of those 3.. Not sure if it'd bring enough value but as a regex nerd it just sounds cool. This may need to be forked as a regex static analysis lib tho 😅 I'm only half kidding this is getting so advanced it feels sad that it's not reusable by others.

staabm:
@Seldaek character classes have already been implemented with https://github.com/phpstan/phpstan-src/pull/3285

> this is getting so advanced it feels sad that it's not reusable by others.

The current impl is tightly coupled with the PHPStan type-system. Its not that easy to be reused. 

I think we need a concrete use-case to see whether re-using can make sense.

Seldaek:
Damn you've been on fire these last few days. I'm on vacation so haven't kept track too closely.

src/Type/Php/RegexGroupParser.php
```diff
@@ -21,6 +21,7 @@
 use PHPStan\Type\TypeCombinator;
 use function array_key_exists;
 use function count;
+use function implode;
```
mvorisek, line 24: there is some issue - https://phpstan.org/r/78e9a062-ad8b-4a3d-a944-e671a4f17a9d - index `1` is optional, ie. the type is missing `|array{string}`

wider repro - https://phpstan.org/r/89288a2e-96e9-4c48-8419-3adc807bea74 - the last type is wrong too, the `1` and `2` indexes are definitely not present when index `3` is

staabm: Index 1/2 is present but empty when 3 exists: https://3v4l.org/VgCUJ

I agree there is a constant type problem though

mvorisek: > the last type is wrong too, the `1` and `2` indexes are definitely not present when index `3` is

sorry, I meant empty string of course

the 2nd repro would be best fixed by "conditional types", ie. like if the inner capturing group type were dependent on the outer capturing group type

staabm: please open issues with reproducing snippets for the known problems. there are too many open todos in already-merged PRs and I can't keep track of all of them

thank you

PR #3285, opened 2024-08-04
Support classes of literal strings in RegexGroupParser

staabm:
support literals within classes for constant type inference.

I do intentionally not support constant types for classes which define "from-to" ranges like `a-z` or `0-9`, as the semantics can get complex easily and the list of constant chars explodes pretty fast.

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
+++ b/tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
@@ -579,7 +579,7 @@ function (string $s): void {
 
 function (string $s): void {
 	if (preg_match('/Price: ([0-9])/i', $s, $matches)) {
-		assertType("array{string, non-empty-string}", $matches);
+		assertType("array{string, numeric-string}", $matches);
 	}
 };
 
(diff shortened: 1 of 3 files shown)


staabm:
btw: I think the issue-bot is confused, as the result seem to be reporting wrong assertion errors

ondrejmirtes:
That's because the issue-bot already posted comments from 1.12.x where the preg array shapes do not require bleeding edge. Here we're still running on 1.11.x.

ondrejmirtes:
Thank you!

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -570,3 +570,33 @@ function (string $s): void {
...
+
+function (string $s): void {
+	if (preg_match('/Price: ([xXa])/i', $s, $matches)) {
+		assertType("array{string, 'a'|'X'|'x'}", $matches);
+	}
+};
+
+function (string $s): void {
+	if (preg_match('/Price: (ba[rz])/i', $s, $matches)) {
+		assertType("array{string, 'bar'|'baz'}", $matches);
```
mvorisek, line 594: `i` modifier must be honored, ie. `'baR'` must be present in the expected type

ondrejmirtes: At that point I'd fall back to not have literal strings at all. Too many combinations -> performance problems

staabm: Yes, had the same thing in mind. Will fix. 

Thanks

staabm: fix in https://github.com/phpstan/phpstan-src/pull/3288

PR #3288, opened 2024-08-05
Don't infer constant types for "i"-modfied patterns

staabm:
because it likely yields too many combinations as discussed in https://github.com/phpstan/phpstan-src/pull/3285#discussion_r1703321209

### DIFF
--- a/src/Type/Php/RegexGroupParser.php
+++ b/src/Type/Php/RegexGroupParser.php
@@ -67,9 +67,9 @@ final class RegexGroupParser
 		}
 
 		$captureOnlyNamed = false;
+		$modifiers = $this->regexExpressionHelper->getPatternModifiers($regex) ?? '';
 		if ($this->phpVersion->supportsPregCaptureOnlyNamedGroups()) {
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

src/Type/Php/RegexGroupParser.php
```diff
@@ -227,6 +220,29 @@ private function walkRegexAst(
 		}
 	}
 
+	private function allowConstantTypes(
+		string $patternModifiers,
+		bool $repeatedMoreThanOnce,
+		RegexCapturingGroup|RegexNonCapturingGroup|null $parentGroup,
+	): bool
+	{
+		if (str_contains($patternModifiers, 'i')) {
```
mvorisek, line 229: see https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.61.1/src/DocBlock/TypeExpression.php#L58 regex, modifiers does not have to be global

staabm: most helpful would be a issue report with a reproducing snippet.

but I am already working on this one, so thanks for the heads up.

PR #3289, opened 2024-08-05
Support non-falsy-string in RegexGroupParser

staabm:
we could infer `non-falsy-string` in more situtations but this one is a easy start, which make constant-array result-types of the [patterns more different](https://github.com/phpstan/phpstan/issues/11443#issuecomment-2266709936).


ondrejmirtes:
Thank you!

PR #3291, opened 2024-08-05
RegexGroupParser: Ignore whitespaces on "x"-modified patterns

staabm:
support "x" pattern modifier

refs https://github.com/phpstan/phpstan/issues/11360

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
+++ b/tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
@@ -638,6 +638,12 @@ function (string $s): void {
 };
 
 function (string $s): void {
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -328,7 +328,7 @@ function (string $size): void {
 function bug11277a(string $value): void
 {
 	if (preg_match('/^\[(.+,?)*\]$/', $value, $matches)) {
-		assertType('array{0: string, 1?: non-falsy-string}', $matches);
```
staabm, line 331: these were too optimistic before `(.+,?)` can be `0`

ondrejmirtes:
Thank you!

PR #3293, opened 2024-08-05
Simplify RegexArrayShapeMatcher

staabm:
simpler fix for https://github.com/phpstan/phpstan/issues/11457 and few more tests


ondrejmirtes:
Thank you!

PR #3294, opened 2024-08-05
Rule to check if required file exists

Bellangelo:
Closes https://github.com/phpstan/phpstan/issues/3397

Originally created at: https://github.com/Bellangelo/phpstan-require-file-exists


src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,130 @@
...
+
+	public function getNodeType(): string
+	{
+		return Include_::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if ($this->shouldProcessNode($node)) {
+			$filePath = $this->resolveFilePath($node->expr, $scope);
```
ondrejmirtes: Instead of handling AST nodes manually, this should just be `$scope->getType($node->expr)` and then asking for `$type->getConstantStrings()`.

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,130 @@
...
+			if ($filePath !== null && !file_exists($filePath)) {
```
ondrejmirtes: is_file would be better

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,130 @@
...
+				return [
+					RuleErrorBuilder::message(
+						sprintf(
+							'Required file "%s" does not exist.',
```
ondrejmirtes: Add `require`/`require_once`/`include`/`include_once` into the message.

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,130 @@
...
+				];
+			}
+		}
+
+		return [];
+	}
+
+	private function shouldProcessNode(Node $node): bool
+	{
+		return $node instanceof Include_ && (
```
ondrejmirtes: This is always true afaik

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,130 @@
...
+use function defined;
+use function dirname;
+use function file_exists;
+use function is_string;
+use function sprintf;
+
+/**
+ * @implements Rule<Include_>
+ */
+final class RequireFileExistsRule implements Rule
```
ondrejmirtes, line 24: The rule needs to be registered in the config. Register it for bleeding edge only (here's an inspiration how https://github.com/phpstan/phpstan-src/commit/c4c02698414934def38e0e1a4a002a2c139d9242).

ondrejmirtes (changes requested):

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/2df8b4c8082ba0eb25f14a3832cd663a8b097ce1

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Did you mean something like this https://github.com/phpstan/phpstan-src/pull/3294/commits/c277d532f8bac79a8d5a19c472fef1408bcad9d4? I haven't added `include` and `include_once` as I have made the rule to skip these as per the description in the issue.
Do you think that we should be handling these as well?

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Wow.. what a great suggestion! It removed so much code and it now handles variables as well.
Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/8c055dddb731784184d68724e568d8f66adbe355

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Thanks for the example! Rule registered here https://github.com/phpstan/phpstan-src/pull/3294/commits/35ad0ee8fec66e1f03b76b478b69f6de3476b34f and here https://github.com/phpstan/phpstan-src/pull/3294/commits/a7212ef7525d5b047f1825c116d20b93801484a8

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: But there isn't any validation that the `$node` would be a `Include_` except for trusting the PHPStan itself, right? Otherwise, phpstan would be also whining about it.

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Here I converted the syntax to <=PHP7.4 as well https://github.com/phpstan/phpstan-src/pull/3294/commits/9be5f52309dfe0fb574c1ca073ed404e744ba734

conf/config.level0.neon
```diff
@@ -115,6 +117,7 @@ rules:
 	- PHPStan\Rules\Types\InvalidTypesInUnionRule
 	- PHPStan\Rules\Variables\UnsetRule
 	- PHPStan\Rules\Whitespace\FileWhitespaceRule
+	- PHPStan\Rules\Keywords\RequireFileExistsRule
```
ondrejmirtes: This can't be in the rules section, it'd always be registered, regardless of the conditionalTags.

Needs to be in services.

phpstan-baseline.neon
```diff
@@ -566,6 +566,11 @@ parameters:
 			count: 1
 			path: src/Rules/Generics/TemplateTypeCheck.php
 
+		-
+			message: "#^Method PHPStan\\\\Rules\\\\Keywords\\\\RequireFileExistsRule\\:\\:processNode\\(\\) should return list\\<PHPStan\\\\Rules\\\\IdentifierRuleError\\> but returns array\\{PHPStan\\\\Rules\\\\RuleError\\}\\.$#"
+			count: 1
+			path: src/Rules/Keywords/RequireFileExistsRule.php
```
ondrejmirtes: This can't be baselined. The error needs an identifier.

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,78 @@
...
+			),
+		)->build();
+	}
+
+	private function resolveFilePath(Include_ $node, Scope $scope): ?string
+	{
+		$type = $scope->getType($node->expr);
+		$paths = $type->getConstantStrings();
+
+		return isset($paths[0]) ? $paths[0]->getValue() : null;
```
ondrejmirtes: We want the error reported even for multiple constant strings, not only one.

src/Rules/Keywords/RequireFileExistsRule.php
ondrejmirtes: We can trust PHPStan. Generics take care of that.

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		return [];
+	}
+
+	private function shouldProcessNode(Node $node): bool
+	{
+		return $node instanceof Include_ && (
+				$node->type === Include_::TYPE_REQUIRE
+				|| $node->type === Include_::TYPE_REQUIRE_ONCE
```
ondrejmirtes: What about include()? I want to report that too.

ondrejmirtes (changes requested):

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/10c9ff79ec679a343d292231744bb6f7a905e974

conf/config.level0.neon
Bellangelo: Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/62d47af944d4db90f4c56f1bc7dfe0e04067c843

phpstan-baseline.neon
Bellangelo: Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/5d4563383f16601d47175086e8b154a4f43efc91

conf/config.level0.neon
```diff
@@ -309,3 +311,7 @@ services:
 
 	-
 		class: PHPStan\Rules\Regexp\RegularExpressionQuotingRule
+	-
+		class: PHPStan\Rules\Keywords\RequireFileExistsRule
+		tags:
+			- phpstan.rules.rule
```
ondrejmirtes: Remove the tag, this makes it always registered

ondrejmirtes (changes requested):

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: When does it contain more than one constant string? I tried multiple variations such as `__DIR__ . '/folder' . '/file.txt'` and it always concatenates them into a single string value.

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: I just noticed actually that it doesn't read variables but just ignores them since it cannot read them. The same exists for the `__DIR__` constants. Am I doing something wrong or it does indeed need some special handling?

conf/config.level0.neon
Bellangelo: Removed in https://github.com/phpstan/phpstan-src/pull/3294/commits/9b60d6fec4cbfbd7bdb1289c2739407cabe4c7af

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: After a few hours I found that the reason it wasn't working was the `usePathConstantsAsConstantString` which was set to `false`.

Bellangelo: Tests updated here: https://github.com/phpstan/phpstan-src/pull/3294/commits/ab173a336d443975921cc97ecd830787cd897538

src/Rules/Keywords/RequireFileExistsRule.php
staabm:
```php
$path = 'myPath';
if (rand(0,1)) {
  $path = 'anotherPath';
}
include $path;
```

Bellangelo: Aha, so `getConstantStrings()` returns all possible cases, got it. Thanks a lot @staabm for the example, here is the fix https://github.com/phpstan/phpstan-src/pull/3294/commits/23c409ef0407b4c0880eaf739391cf844ee787f0

ondrejmirtes (changes requested):
I just realized one big thing we need to implement here.

See the documentation: https://www.php.net/manual/en/function.include.php

> Files are included based on the file path given or, if none is given, the [include_path](https://www.php.net/manual/en/ini.core.php#ini.include-path) specified. If the file isn't found in the [include_path](https://www.php.net/manual/en/ini.core.php#ini.include-path), include will finally check in the calling script's own directory and the current working directory before failing. The include construct will emit an [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) if it cannot find a file; this is different behavior from [require](https://www.php.net/manual/en/function.require.php), which will emit an [E_ERROR](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-error).

So - for absolute paths we can just check `is_file` before throwing an error, but for relative paths, we need to implement the logic described in the PHP documentation.

Also - we don't need these many tests, the tests you added simply test the logic that is already present in `$scope->getType()`, not the code that's in the rule itself.

I'd be fine if just tested a few cases where general `string` is passed into `require_once` (no error reported) and a union of a couple of constant strings (which are checked against existing files).

Instead of these tests, we should thoroughly test the logic I quoted above.

Thanks.

Bellangelo:
> I just realized one big thing we need to implement here.
> 
> See the documentation: https://www.php.net/manual/en/function.include.php
> 
> > Files are included based on the file path given or, if none is given, the [include_path](https://www.php.net/manual/en/ini.core.php#ini.include-path) specified. If the file isn't found in the [include_path](https://www.php.net/manual/en/ini.core.php#ini.include-path), include will finally check in the calling script's own directory and the current working directory before failing. The include construct will emit an [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) if it cannot find a file; this is different behavior from [require](https://www.php.net/manual/en/function.require.php), which will emit an [E_ERROR](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-error).
> 
> So - for absolute paths we can just check `is_file` before throwing an error, but for relative paths, we need to implement the logic described in the PHP documentation.
> 

Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/e1469de9a119718f494d69df892533b4be7e56df. 

> Also - we don't need these many tests, the tests you added simply test the logic that is already present in `$scope->getType()`, not the code that's in the rule itself.
> 
> I'd be fine if just tested a few cases where general `string` is passed into `require_once` (no error reported) and a union of a couple of constant strings (which are checked against existing files).
> 
> Instead of these tests, we should thoroughly test the logic I quoted above.
> 
> Thanks.

Tests decluttered here https://github.com/phpstan/phpstan-src/pull/3294/commits/4ae5df524fe2e8bb6ae4a397ebb102b2c55549e6 and then a new test was added for the include paths https://github.com/phpstan/phpstan-src/pull/3294/commits/bec70b89d3d6681590b96e7831724c6cc2ca264f

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,91 @@
...
+		return Include_::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$errors = [];
+		$paths = $this->resolveFilePaths($node, $scope);
+
+		foreach ($paths as $path) {
+			if (stream_resolve_include_path($path) !== false) {
```
ondrejmirtes: This is a clever shortcut but I worry it doesn't work.

Example:

* We're analysing file /var/www/A.php
* We're including 'B.php'
* This file is next to A.php
* include() works but stream_resolve_include_path called from inside PHPStan rule will fail because its PHP code is elsewhere on the disk.
* The result is a false positive.

Can you verify this assumption here? Thanks.

ondrejmirtes (changes requested):
I appreciate your effort and your patience here! I bet you didn't expect this to be so much work but the bar of what can go into PHPStan is really high.

I'm pointing out all the issues I can think of, because from my experience it's really important. The users will report even more issues, issues we didn't even think of.

So please bear with me so we can bring this over the finish line. Thank you.

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: You are kind of right. This is actually the reason why I didn't touch the relative paths at all at the beginning and I was showing through the tests that they do not work. The `stream_resolve_include_path` should work exactly like the `include` but the problem is this line from the php documentation: 
> include will finally check in the calling script's own directory and the current working directory before failing.

Unfortunately, we cannot know the actual working directory of the application. The problem becomes even worse if you think that different files may have different working directories. 
In your case, it would work only if you run PHPStan from the `var/www` directory.
One solution is for users to call phpstan from the working directory that they want to test against. Another thing that we could do is to make the rule accept an array of working directories that we would check in addition of the `stream_resolve_include_path`.

Since there are a few cases that produce both false positives and false negatives we can disallow the usage of relative paths. A simple `stream_resolve_include_path($path) === $path` should show if the path was relative or not.

Bellangelo: You can test that it doesn't matter where the PHPStan's code is but instead where what the working directory is by adding in the `require-file-relative-path.php` this line `include 'preload.php'`. This file is in the root directory and depending from which directory you run the phpunit you will have different results.

Bellangelo: @ondrejmirtes Kind reminder.

ondrejmirtes: I'd be fine with getting the same CWD PHPStan has, it's reasonable to assume it's going to be the same for the application.

But your code is missing one use case - "calling script's own directory". That's why you can't use `stream_resolve_include_path` but instead have to check all the same places PHP checks with own code.

Bellangelo: > But your code is missing one use case - "calling script's own directory".

But can we know the called script? To make it more specific, if my application is being called through the `public/index.php` can the PHPStan understand this?
Or what you are saying is that `stream_resolve_include_path` would also check the calling script's own directory and since in our case the calling script either PHPStan or PHPUnit it might generate false results?

ondrejmirtes: I just tested that. It's not a directory with the application entrypoint like index.php, but it's the same directory where the script with the `require` call is. Which means that when we're analysing a `require` AST node, we need to look at `$scope–>getFile()`, get its directory and look for the file in that directory.

Example:

- a.php
- sub/b.php
- sub/c.php
- c.php

a.php includes `sub/b.php` (with absolute path). `sub/b.php` includes `c.php` with a relative path.

`sub/c.php` says `echo "in subdirectory"`.

When we execute `a.php`, the question is: will `c.php` or `sub/c.php` be executed? The answer is `sub/c.php`. Because PHP prints "in subdirectory".

Bellangelo: You are absolutely correct. At first, I got different results than yours but only because my working directory was in the same level as `a.php` and `c.php`. This made me recheck the priority that file has and here are the results:
1. Working directory has the highest priority `getcwd()`.
2. Include paths are checked second `get_include_path()`.
3. Last but not least, is what you observed. Basically, the folder of the current file is checked.

Bellangelo: Aaaand i think we are finally close to the end. Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/0a53397f3bb9d708f62a25a33814bf84fc1e3080

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,132 @@
...
+	 * This method simulates the behavior of `stream_resolve_include_path` but for the given scope.
+	 * The priority order is the following:
+	 * 	1. The current working directory.
+	 * 	2. The include path.
+	 *  3. The path of the script that is being executed.
+	 */
+	private function doesFileExist(string $path, Scope $scope): bool
+	{
+		$directories = [
+			getcwd(),
```
ondrejmirtes: Inject `$currentWorkingDirectory` into the constructor. Look for `%currentWorkingDirectory%` in `conf/` to see how it's done.

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,132 @@
...
+
+		$identifier = sprintf('%s.fileNotFound', str_replace('_once', 'Once', $type));
+
+		return RuleErrorBuilder::message(
+			sprintf(
+				$message,
+				$type,
+				$filePath,
+			),
+		)->nonIgnorable()->identifier($identifier)->build();
```
ondrejmirtes: Let it be ignorable

ondrejmirtes (changes requested):

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/67ee4d818e82eff2adf5b509654502a1e918721d

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/4e049ea2df543d7554901f2eb64fd453f491ea2c

VincentLanglet (commented):
Shouldn't a test be added for the case
```

$path = __DIR__ . '/include-me-to-prove-you-work.txt';

if (file_exists('a-file-that-does-not-exist.php')) {
	$path = 'a-file-that-does-not-exist.php';
}

include $path;
include_once $path;
require $path;
require_once $path;
```
and
```

$path = __DIR__ . '/include-me-to-prove-you-work.txt';

if (rand(0,1)) {
	$path = 'a-file-that-does-not-exist.php';
}

if (file_exists($path)) {
    include $path;
    include_once $path;
    require $path;
    require_once $path;
}
```

because this shouldn't report any error ?

Bellangelo:
Hi @VincentLanglet, I think what you are proposing is outside of the scope of the PR. It seems that PHPStan doesn't evaluate correctly if `file_exists` always returns true/false and as such it doesn't work.

Here is an example: https://phpstan.org/r/3e79db4e-40ab-46d7-9e6f-9dde7a850fe3 where you would expect the variable to always exist.

VincentLanglet:
> Hi @VincentLanglet, I think what you are proposing is outside of the scope of the PR. It seems that PHPStan doesn't evaluate correctly if `file_exists` always returns true/false and as such it doesn't work.
> 
> Here is an example: [phpstan.org/r/3e79db4e-40ab-46d7-9e6f-9dde7a850fe3](https://phpstan.org/r/3e79db4e-40ab-46d7-9e6f-9dde7a850fe3) where you would expect the variable to always exist.

In both of my example, the variable `$path` is always defined. That's very different from your example.

What you're trying to introduce is a rule which add a PHPStan error when include/require is used on the path if the path doesn't exist, regardless the path has been checked as existing !

Phpstan does the difference between 
https://phpstan.org/r/a082ea0f-619e-4294-b909-eefa6ff7c43c
and 
https://phpstan.org/r/16b7bed5-1a95-4534-a43b-3a2c02da11c3

it must the be same for the existence of files.

```
$path = __DIR__ . '/include-me-to-prove-you-work.txt';

if (file_exists('a-file-that-does-not-exist.php')) {
	$path = 'a-file-that-does-not-exist.php';
}

include $path; // Here I assume you have a phpstan error reported ; it shouldn't because path can only be `'a-file-that-does-not-exist.php'` if the file exists.
```

```
$path = __DIR__ . '/include-me-to-prove-you-work.txt';

if (rand(0,1)) {
	$path = 'a-file-that-does-not-exist.php';
}

if (file_exists($path)) {
    include $path; // Here I assume you have a PHPStan error reported, you shouldn't since `file_exists` check is done before.
}
```

Bellangelo:
> In both of my example, the variable $path is always defined. That's very different from your example.

My example showcases that PHPStan currently does not understand that `file_exists(__FILE__)` is always true and its variations. So, how can we know the type of the variable if PHPStan itself doesn't report it back to us correctly? Here is another [example](https://phpstan.org/r/34b4dc62-976b-4567-bbe6-ad16e875efb4) where PHPStan identifies correctly that `if(1===1)` is always true and as such it identifies the variable's type.
I found another [example](https://phpstan.org/r/e1b9906b-83f3-40a3-a0ed-06c3b8961ab0) without the usage of `file_exists` which shows that it cannot identify correctly the conditions. It could be that we need to enable another option but I could be missing more than this.

VincentLanglet:
I dunno why you quoting totally out of debate example with always true condition. (Especially the "isAlwaysTrue" method which is a misconception of how PHPStan works since it rely on the signature and not the implementation of the method).

Currently PHPStan has no issue with the include/require method since there is no rule about it. You are trying to include one ; all I am saying is the fact that you rule introduces false-positive. If you want to prove me wrong, just add the example I gave you in your tests.

You are throwing a PHPStan error when a constantString fileName does not exists ; but I can use some code on different environment (and the file might exist sometimes, and might not exist the other time). If I check for file_exist before include/require the file, PHPStan should not warn me with a false positive error.

Bellangelo:
> I dunno why you quoting totally out of debate example with always true condition. (Especially the "isAlwaysTrue" method which is a misconception of how PHPStan works since it rely on the signature and not the implementation of the method).
> 
> Currently PHPStan has no issue with the include/require method since there is no rule about it. You are trying to include one ; all I am saying is the fact that you rule introduces false-positive. If you want to prove me wrong, just add the example I gave you in your tests.
> 
> You are throwing a PHPStan error when a constantString fileName does not exists ; but I can use some code on different environment (and the file might exist sometimes, and might not exist the other time). If I check for file_exist before include/require the file, PHPStan should not warn me with a false positive error.

@VincentLanglet The examples are not out of debate. They show that the problem you are mentioning pre-exists of my new rule. The rule does not parse by itself the variables used in the `include`, it just uses PHPStan to get their values. This is what I have been trying to show you with my examples. My examples reproduce the problem you are mentioning but without the usage of the new rule. 
And what is the problem? The problem is that PHPStan in the current setup does not handles correctly the variable types when used inside a `if(callToFunction())`.
You can check again the [first example](https://phpstan.org/r/3e79db4e-40ab-46d7-9e6f-9dde7a850fe3) that I send you which uses the `file_exists`. As you can see, even if in my example the `file_exists` returns true, it still doesn't recognize that the variable exists. Here is also an [updated version](https://phpstan.org/r/c85f8921-9ba6-4097-904e-2927d76c0014) of it. As you can see, its dumped type is mixed and not string.  All of these prove that the behaviour you are trying to test is in the PHPStan itself and not in the rule per se.
Should my rule do something differently to retrieve the real value of the variable?

VincentLanglet:
> @VincentLanglet The examples are not out of debate. They show that the problem you are mentioning pre-exists of my new rule. The rule does not parse by itself the variables used in the `include`, it just uses PHPStan to get their values. This is what I have been trying to show you with my examples. My examples reproduce the problem you are mentioning but without the usage of the new rule.

We definitely don't understand each others.

Can you please focus use the example I gave you ? I'm not concerned about the fact that some `file_exist($path)` condition are always true or always false (like you said, it's already the case). I'm concerned about the fact that your new rule about `require/include` introduce false positive.
Sure the current analyse from PHPStan about `require`, `include`, `file_exist` method (and certainly more) is not perfect, but it's better to have no error reported rather than introducing false positive. (Especially if this rule is enabled by default in bleeding edge / PHPStan 2.0)

As I understand your rule
```
require 'path-which-exists';
```
is OK and
```
require 'path-which-not-exists';
```
is reported as an error.

According to your test `testFileDoesNotExistConditionally ` https://github.com/phpstan/phpstan-src/pull/3294/files#diff-c86e2fa4121a5853c7f1d13452c7599ef1e871bd07ddc9ccd93ee4e21a7617fdR67, 
```
require 'path-which-exists'|'path-which-not-exists';
```
is supposed to be reported as an error, which would be https://phpstan.org/r/9a199041-399e-4254-8642-a92bdf12ca65

But with your rule, https://phpstan.org/r/c4083b8a-a572-4dc2-b48e-4e95e247aab7 will be thrown as an error which is wrong. It's better to have no rule about `include/require` rather than file positive because PHPStan doesn't understand file_exist file. Or it require to improve some TypeSpecifier/StringType (with maybe an accessory filename-string) things.

Bellangelo:
@VincentLanglet Ok, I think we pretty much saying the same thing but we disagree on what should happen next. Your last comment makes it clear that you don't want this rule to be merged since it doesn't catch all cases. On the other hand, all I am saying is that this problem exists in all rules with similar cases, since it isn't the rule's fault, and as such you can create new rules with similar issues as this "bug" is expected.

Bellangelo:
Hi @ondrejmirtes, how shall we proceed with this one? Is what [VincentLanglet](https://github.com/VincentLanglet) saying a valid concern or we can live without it? Should we schedule perhaps some next steps? For example, add support for `file_exists`?

ondrejmirtes:
@Bellangelo I say we continue. Real-world user feedback after releasing this will tell if this is a real problem or not.

The way to address this problem would be to introduce a new string accessory type like `file-string`. A new `AccessoryFileStringType` could be attached to `StringType` or `ConstantStringType` in an `IntersectionType`. After asking for `is_file()` we could narrow the type of the argument, and skip reporting that in the rule.

src/Rules/Keywords/RequireFileExistsRule.php
```diff
@@ -0,0 +1,135 @@
...
+		return is_file($absolutePath);
+	}
+
+	private function getErrorMessage(Include_ $node, string $filePath): IdentifierRuleError
+	{
+		$message = 'Path in %s() "%s" is not a file or it does not exist.';
+
+		switch ($node->type) {
+			case Include_::TYPE_REQUIRE:
+				$type = 'require';
```
ondrejmirtes, line 92: One last simple change: I worry the identifier cannot be statically inferred here because of `str_replace`. We need that for https://github.com/phpstan/phpstan/actions/workflows/extract-identifiers.yml which updates https://phpstan.org/error-identifiers.

Please assign `$identifier = ...` in the switch in each case, and use that in RuleErrorBuilder.

Then I'll merge this. Thank you for the effort!

ondrejmirtes (changes requested):

src/Rules/Keywords/RequireFileExistsRule.php
Bellangelo: Fixed in https://github.com/phpstan/phpstan-src/pull/3294/commits/e861017a0793a3d8f6dc5b7192c6d4d7c59d217f.

Shouldn't this also run as a quality check before merging a PR?

ondrejmirtes: Yes, feel free to contribute this rule 😊

ondrejmirtes:
Thank you!

ondrejmirtes:
More about the `file-string` type. I discussed it with @jbafford and he thinks that this concept can't be expressed with the typesystem. Because the fact about file existence can change between the calls.

```php
if (is_file($path)) {
    require_once $path; // can fail, file might have been deleted
}
```

So reporting an error in such cases is actually a feature, not a bug.

PR #3296, opened 2024-08-06
Refactor RegexGroupParser - extract RegexAlternation class

staabm:
Goal: ease further bugfixes based on properties of alternations


ondrejmirtes:
An idea: Everything that's not `@api` around this, move to `PHPStan\Type\Regex` namespace.

ondrejmirtes:
Thank you.

PR #3297, opened 2024-08-06
ShouldNotHappenException extends LogicException

staabm:
before this PR PHPStorm false-postively reporting missing handling of ShouldNotHappenExceptions

grafik

----

LogicException

> Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code. 

see https://www.php.net/LogicException

### DIFF
--- a/src/ShouldNotHappenException.php
+++ b/src/ShouldNotHappenException.php
@@ -2,9 +2,9 @@
 
 namespace PHPStan;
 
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Errr, no. I always turn off these PhpStorm inspections. PHPStan's exception checking is much more flexible.

PR #3298, opened 2024-08-06
is_resource: adjust type specification for truthy context only

janedbal:
Should fix https://github.com/phpstan/phpstan/issues/9687


thg2k:
I don't think you need a dedicated class for this, just replace the stub assertion with:

```
@phpstan-assert-if-true =resource $value
```

The (undocumented?) equal sign indicates that if false the opposite should not be asserted.

janedbal:
Oh god, the whole psalm-originated `@phpstan-assert-if-true` is so un-intuitive! 

Thank you!

tests/PHPStan/Analyser/data/specifiedTypesUsingIsFunctions.php
```diff
@@ -41,7 +41,8 @@ function () {
 		throw new \Exception();
 	}
 
-	if (!is_resource($resource)) {
+	$resource = fopen('php://memory', 'r');
+	if ($resource === false) {
```
ondrejmirtes: Not sure why the change is here. Usually changing existing (and especially this old) tests is a red flag. Did we break something?

ondrejmirtes (changes requested):
Since we're trying to fix "Call to function is_resource() with resource will always evaluate to true.", we need a test for ImpossibleCheckTypeFunctionCallRule. Thanks.

tests/PHPStan/Analyser/data/specifiedTypesUsingIsFunctions.php
janedbal: Reverted, not needed.

ondrejmirtes:
Thank you!

PR #3299, opened 2024-08-06
RegexArrayShapeMatcher - Fix shape of single top level alternations

staabm:
closes https://github.com/phpstan/phpstan/issues/11462

additionally in more cases we get more precise tagged unions 🥳

### DIFF
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -14,6 +14,7 @@ use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
+use PHPStan\Type\NullType;
 use PHPStan\Type\Regex\RegexAlternation;
 use PHPStan\Type\Regex\RegexCapturingGroup;
 use PHPStan\Type\Regex\RegexExpressionHelper;
@@ -140,7 +141,7 @@ final class RegexArrayShapeMatcher
 			&& $onlyOptionalTopLevelGroup !== null
 		) {
 			// if only one top level capturing optional group exists
-			// we build a more precise constant union of a empty-match and a match with the group
+			// we build a more precise tagged union of a empty-match and a match with the group
 
 			$onlyOptionalTopLevelGroup->forceNonOptional();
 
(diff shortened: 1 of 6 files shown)


staabm:
@mvorisek please double check the correctness of the test expectations

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -648,3 +648,15 @@ function (string $s): void {
 		assertType('array{string, non-empty-string}', $matches);
 	}
 };
+
+function (string $s): void {
+	if (preg_match('~a|(\d)|(\s)~', $s, $matches) === 1) {
+		assertType("array{0: string, 1?: numeric-string}|array{string, '', non-empty-string}", $matches);
```
mvorisek, line 682: LGTM, thank you for fixing this!

tests/PHPStan/Analyser/nsrt/bug-11311.php
```diff
@@ -198,3 +198,29 @@ function (string $s): void {
...
+
+function (string $s): void {
+	if (preg_match('~a|(\d)|(\s)~', $s, $matches, PREG_UNMATCHED_AS_NULL|PREG_OFFSET_CAPTURE) === 1) {
+		assertType("array{array{string|null, int<-1, max>}, array{numeric-string|null, int<-1, max>}, array{non-empty-string|null, int<-1, max>}}", $matches);
+	}
+};
+
+function (string $s): void {
+	if (preg_match('~a|((u)x)|((v)y)~', $s, $matches, PREG_UNMATCHED_AS_NULL) === 1) {
+		assertType("array{string, 'ux'|null, 'u'|null, 'vy'|null, 'v'|null}", $matches);
```
staabm, line 219: I really like this tagged unions

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -154,18 +155,24 @@ private function matchRegex(string $regex, ?int $flags, TrinaryLogic $wasMatched
...
 
 			if (!$this->containsUnmatchedAsNull($flags, $matchesAll)) {
+				// positive match has a subject but not any capturing group
 				$combiType = TypeCombinator::union(
 					new ConstantArrayType([new ConstantIntegerType(0)], [$this->createSubjectValueType($flags, $matchesAll)], [0], [], true),
 					$combiType,
 				);
 			}
 
+			$onlyOptionalTopLevelGroup->clearOverrides();
```
ondrejmirtes, line 165: This smells of too much statefulness and mutability. Is this pattern in more places in Regex\? I'd much prefer immutable stateless value objects.

ondrejmirtes (changes requested):

src/Type/Php/RegexArrayShapeMatcher.php
staabm: totally agree. 

I tried doing it in a immutable way, but my attempts don't work because the objects beeing changed here, are referenced by other objects via `parent` property and we need the objects referenced by `$onlyOptionalTopLevelGroup` to change, no matter which object in the `$groupList` directly or nested references it.

staabm: > Is this pattern in more places in Regex?

we use this pattern for the 2 tagged union special cases in this method here. nowhere else.

ondrejmirtes:
Conflict here

ondrejmirtes:
Thank you!

PR #3301, opened 2024-08-07
Fix loose comparison between '1' and '+1'

thg2k:
Here is a bug fix on @ondrejmirtes's favorite PHP operator: `==`

Basically I just removed the use of the identical operator, falling back on the Type::looseCompare() implementation which is already correct.


ondrejmirtes:
Thank you.

PR #3303, opened 2024-08-08
Update composer-dependency-analyser to 1.7.0

janedbal:
- https://github.com/shipmonk-rnd/composer-dependency-analyser/releases/tag/1.7.0


ondrejmirtes:
Thank you!

PR #47, opened 2019-12-11
Handle bad reflection parameters more gracefully

fabacino:
The method `\ReflectionParameter::getName()` is supposed to always return a string. However, it is also possible that `null` gets returned. This happened with the method `get` of the PHP extension
`Memcache` on version 3.0.8+.

This PR throws a new `BadParameterFromReflectionException` in case a bad reflection paramer is encountered. A future user then at least knows the name of the class and the method at fault without extensive debugging and can go on from there.

Relates to https://github.com/phpstan/phpstan/issues/2697

PR #3304, opened 2024-08-08
Add basic type narrowing in the loose equality with the empty string

thg2k:
As promised a [long time ago](https://github.com/phpstan/phpstan/issues/7668#issuecomment-1191379214), I submit my first real attempt to narrow the type after `if ($a != "") ...`

This is by no means a complete analysis of the loose equality operator, but it should be sound. The discrepancy between php 7.x and 8.x are handled by preserving the two offending types (integer and float zero) in the type for both the true and false context.

I have the feeling that the definitive solution would be to replace `Type::looseCompare()` with `Type::looseIntersectWith()`, which should solve all the loose comparison problems. Do you think it would be something worth pursuing?

If in the meanwhile you think this PR is acceptable, it would solve a lot of problems in my codebase.

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1611,7 +1611,7 @@ final class TypeSpecifier
 	}
 
 	/**
-	 * @return array{Expr, ConstantScalarType}|null
+	 * @return array{Expr, ConstantScalarType, Type}|null
 	 */
 	private function findTypeExpressionsFromBinaryOperation(Scope $scope, Node\Expr\BinaryOp $binaryOperation): ?array
 	{
@@ -1633,13 +1633,13 @@ final class TypeSpecifier
 			&& !$rightExpr instanceof ConstFetch
 			&& !$rightExpr instanceof ClassConstFetch
 		) {
-			return [$binaryOperation->right, $leftType];
+			return [$binaryOperation->right, $leftType, $rightType];
 		} elseif (
 			$rightType instanceof ConstantScalarType
 			&& !$leftExpr instanceof ConstFetch
 			&& !$leftExpr instanceof ClassConstFetch
 		) {
-			return [$binaryOperation->left, $rightType];
+			return [$binaryOperation->left, $rightType, $leftType];
 		}
 
 		return null;
@@ -1950,6 +1950,7 @@ final class TypeSpecifier
 		if ($expressions !== null) {
 			$exprNode = $expressions[0];
 			$constantType = $expressions[1];
+			$otherType = $expressions[2];
 
 			if (!$context->null() && $constantType->getValue() === null) {
 				$trueTypes = [
@@ -1979,6 +1980,30 @@ final class TypeSpecifier
 					$context->true() ? TypeSpecifierContext::createTruthy() : TypeSpecifierContext::createTruthy()->negate(),
 					$rootExpr,
 				);
+			}
+
+			if (!$context->null() && $constantType->getValue() === 0 && !$otherType->isInteger()->yes()) {
+				/* There is a difference between php 7.x and 8.x on the equality
+				 * behavior between zero and the empty string, so to be conservative
+				 * we leave it untouched regardless of the language version */
+				if ($context->true()) {
+					$trueTypes = [
+						new NullType(),
+						new ConstantBooleanType(false),
+						new ConstantIntegerType(0),
+						new ConstantFloatType(0.0),
+						new StringType(),
+					];
+				} else {
+					$trueTypes = [
+						new NullType(),
+						new ConstantBooleanType(false),
+						new ConstantIntegerType(0),
+						new ConstantFloatType(0.0),
+						new ConstantStringType('0'),
+					];
+				}
+				return $this->create($exprNode, new UnionType($trueTypes), $context, false, $scope, $rootExpr);
 			}
 
 			if (!$context->null() && $constantType->getValue() === '') {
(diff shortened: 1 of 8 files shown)


ondrejmirtes:
We need to solve two things in regard to `==` and `!=`.

1) Whether the result of the comparison is `true` or `false`. This is currently achieved with `Type::looseCompare()`.
2) How types are narrowed in `==` and `!=` conditions.

I'm in favour of adding a new method (or methods) on `Type` that will solve both of these needs at once.

I'd like the implementations to easily follow the loose comparison table from https://www.php.net/manual/en/types.comparisons.php.

The new method on Type could return a union of all types it loosely compares to with `true`. If we take `ConstantStringType` with `'-1'`, the method should return a union type of:

* true
* -1
* `-1` (`$this`)

To resolve the result of `==` and `!=`, we could run `$leftType->isSuperTypeOf($rightType->looseComparison())`.

For type narrowing, we can run `TypeCombinator::intersect($leftType->looseComparison(), $rightType)`.

I'm not saying this will work for all cases, maybe we'll have to adjust, but it's a start.

thg2k:
I'm a bit lost on this approach, but I'd be happy to investigate it. Should I open a dedicated issue for it? This PR is about getting one subset of the problem handled.

ondrejmirtes:
I'm sorry, I don't want a narrow partial solution if we can do a proper one.

thg2k:
I still don't understand the approach of returning the types, are you sure it shouldn't be an looseIntersectWith($type, $version)? How would you tackle things that don't involve basic constants, like `'15foo' == 15 (true < 8.0, false >= 8.0)`?

Anyway I managed to convert it to a plugin by leveraging the nette DI, I had to override also the TypeSpecifierFactory, as there is an explicit reference to the class name there.

thg2k:
Another example that comes to mind is: if ($a == 15) ... 

```
/** @var mixed $a */
if ($a == 15) { assertType('15|numeric-string', $a); }
```

In this case `$a` can be '15', '+15', '015', '0015' and so on.

So a proper solution is really not easy. I think the best solution is to follow the approach in this PR and just handle all the cases in the table there, I basically handlded the empty string case but it can easily be duplicated for the other cases. Would you accept the PR if I complete the table in the docs?

thg2k:
I'm working on an updated PR to handle more cases. This also fixes a problem I found, see this:

https://phpstan.org/r/62751f1c-596f-4d7c-8a82-ff027b0515d5

Note that the evaluation of the expression actually gets it right, it's the type narrowing that gets it wrong.. is there no possibility to get this done in one method without duplicating the logic?
saying that '$a == $b' is true/false/bool should be a subset of the problem of narrowing types for $a and $b, but I get lost in the thought process...

ondrejmirtes (changes requested):
1) Please fix the tests.
2) Please express the different behaviour between PHP 7.x and 8.x with a new method in PhpVersion and doing the if in TypeSpecifier.

thg2k:
> 1. Please fix the tests.

I just pushed one round of fixes, I'll check the CI and in case fix it again.

SIDE NOTE: It took me a moment to find the failing case in the TypeSpecifierTest::testCondition data series. Would you mind if I send a separate PR after this one is merged to add labels to the data cases? This way phpunit shows the label instead of a unhelpful `#47`.

> 2. Please express the different behaviour between PHP 7.x and 8.x with a new method in PhpVersion and doing the if in TypeSpecifier.

I'll tackle this tomorrow. Are you sure it's the right thing to do? When I made this PR I gave it some thought and figured it was better to implement the php7 behavior, but I do not remember the reasoning.. Anyway this should be easy to fix.

ondrejmirtes:
> add labels to the data cases

Yeah, sure, if you can describe them in a readable and descriptive way. What I usually do is I search for the asserted type and find it pretty fast.

> Are you sure it's the right thing to do?

Sure, otherwise it's either going to be wrong for 7.x or for 8.x fols.

thg2k:
I finally managed to get tests to pass, please review the tests that I patched,
especially 'bug-11674.php': I can't come up with any value that is false as `if ((string) $var)` but true as `if ($var)`.

> > Are you sure it's the right thing to do?
> 
> Sure, otherwise it's either going to be wrong for 7.x or for 8.x fols.

About this... Actually, it wouldn't be "wrong". I now remember the my reasoning from a few months ago, please indulge me for a moment: Here we are talking about type narrowing, so this is not about implementing either php7 or php8 behavior, but.. BOTH at the same time. When it comes to type narrowing, not narrowing a type is not wrong, so my idea is, as many people develop applications that works for both php7 and php8, to simply not narrow the type at all when it comes to the two cases (string vs zero) that have a different behavior in the two major versions. This puts us on the safest side of things, and still have a correct analysis anyway.

ondrejmirtes:
Thank you! If you have any follow-up feel free to send another PR :)

> I now remember the my reasoning from a few months ago

Yeah, it might be okay, we'll see from the feedback.

thg2k:
Oh that was a fast merge! I left some random comments while browsing through the code that didn't belong to this PR.. I was going to clean them up but.. too late 😆  I actually think they are helpful for newcomers, the code can be.. a bit confusing on a first approach.

thg2k:
Hmm I think the 2.0.x merge is not correct, I'm happy to fix it, should I send a PR?

ondrejmirtes:
It's not late, you can send a follow-up PR with whatever :)

I'm already fixing 2.0.x.

thg2k:
closes phpstan/phpstan#10338

PR #3306, opened 2024-08-08
RegexArrayShapeMatcher - optional non-last groups can be empty-string

staabm:
closes https://github.com/phpstan/phpstan/issues/11479


ondrejmirtes:
Thank you!

PR #3308, opened 2024-08-09
Fix bug in array-shape subtraction

staabm:
closes https://github.com/phpstan/phpstan/issues/11488

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -366,15 +366,6 @@ class ConstantArrayType extends ArrayType implements ConstantType
 				return $type->isIterableAtLeastOnce()->negate();
 			}
 
-			if (
-				$this->isList->yes()
-				&& $type->isList->yes()
-				&& count($this->keyTypes) !== count($type->keyTypes)
-				&& count($type->optionalKeys) === 0
-			) {
-				return TrinaryLogic::createNo();
-			}
-
 			$results = [];
 			foreach ($this->keyTypes as $i => $keyType) {
 				$hasOffset = $type->hasOffsetValueType($keyType);
@@ -394,6 +385,25 @@ class ConstantArrayType extends ArrayType implements ConstantType
 					return TrinaryLogic::createNo();
 				}
 				$results[] = $isValueSuperType;
+			}
+
+			if (
+				$this->isList->yes()
+				&& $type->isList->yes()
+				&& count($this->keyTypes) !== count($type->keyTypes)
+				&& count($type->optionalKeys) === 0
+			) {
+				$keepSeparate = true;
+				foreach ($this->valueTypes as $valueType) {
+					if ($valueType->isConstantValue()->yes()) {
+						$keepSeparate = false;
+						break;
+					}
+				}
+
+				if ($keepSeparate) {
+					return TrinaryLogic::createNo();
+				}
 			}
 
 			return TrinaryLogic::createYes()->and(...$results);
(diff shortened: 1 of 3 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -388,6 +388,25 @@ public function isSuperTypeOf(Type $type): TrinaryLogic
...
+				&& count($this->keyTypes) !== count($type->keyTypes)
+				&& count($type->optionalKeys) === 0
+			) {
+				$keepSeparate = true;
+				foreach ($this->valueTypes as $valueType) {
+					if ($valueType->isConstantValue()->yes()) {
+						$keepSeparate = false;
+						break;
+					}
+				}
```
staabm, line 402: we somehow need to identify which constant arrays we can/want to  keep separte (in our case tagged unions), without running into "is always false" comparison false positives like 

> list<array{string, 1|2|3|4|5|6|7}> and list<array{string, 1|2|3|4|5|6|7, int}> will always evaluate to false

I think my approach is wrong, but I have no better idea atm

tests/PHPStan/Analyser/nsrt/bug11488.php
```diff
@@ -0,0 +1,64 @@
...
+			assertType('array{0: mixed, 1?: string|null}', $row);
+		} else {
+			assertType('array{0: mixed, 1?: string|null}', $row);
+		}
+
+		if (count($row) !== 2) {
+			assertType('array{0: mixed, 1?: string|null}', $row);
+		} else {
+			// should be array{0: mixed, 1: string|null}
+			assertType('array{mixed, mixed}', $row);
```
staabm, line 30: I think thats another bug

staabm: fixed via https://github.com/phpstan/phpstan-src/pull/3309

src/Type/Constant/ConstantArrayType.php
ondrejmirtes: Any change to isSuperTypeOf logic needs tests in TypeCombinatorTest::dataUnion and dataIntersect. So please express there what you're trying to achieve, only when reviewing that I can decide if it makes sense or not.

ondrejmirtes (changes requested):

src/Type/Constant/ConstantArrayType.php
staabm: I could not figure out a dataUnion and dataIntersect test.

but what made sense to me was a test like
subtract `array{mixed}` from `array{mixed, string|null, mixed}|array{mixed}` should lead to `array{mixed, string|null, mixed}`

mvorisek:
Can this PR be finished please, many of our CI pipelines are still failing because of this. Thank you in advance!

staabm:
@mvorisek will continue after vacation. Please make sure to sponsor my efforts. Thanks in advance

ondrejmirtes:
@staabm FYI my patience with @mvorisek has run out and I blocked him for 1 week.

mvorisek:
@ondrejmirtes, Ondřej, I do not want you to talk to me like this. If some of my communication ruin your patience, please always share the reasons with me to be able to understand your attitude better. I hope we can talk about these feelings openly. Thank you.

VincentLanglet:
Hi @staabm do you remember what is missing in this PR ? Do you want me to try to finish it ?

staabm:
@VincentLanglet I can't remember. feel free to continue it

PR #3309, opened 2024-08-09
Fix item-type in list to constant-array conversion with count()

staabm:
before this PR we were loosing precision in this process as item-types were kind of generalized before.

I guess the original implementation was done before `Type->getOffsetValueType()` was a thing.

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -991,7 +991,7 @@ class TypeSpecifier
 		return null;
 	}
 
-	private function turnListIntoConstantArray(FuncCall $countFuncCall, Type $listOrArray, Type $sizeType, Scope $scope): ?Type
+	private function turnListIntoConstantArray(FuncCall $countFuncCall, Type $type, Type $sizeType, Scope $scope): ?Type
 	{
 		$argType = $scope->getType($countFuncCall->getArgs()[0]->value);
 
@@ -1004,15 +1004,15 @@ class TypeSpecifier
 
 		if (
 			$isNormalCount->yes()
+			&& $type->isList()->yes()
 			&& $sizeType instanceof ConstantIntegerType
-			&& $listOrArray->isList()->yes()
 			&& $sizeType->getValue() < ConstantArrayTypeBuilder::ARRAY_COUNT_LIMIT
 		) {
 			// turn optional offsets non-optional
 			$valueTypesBuilder = ConstantArrayTypeBuilder::createEmpty();
 			for ($i = 0; $i < $sizeType->getValue(); $i++) {
 				$offsetType = new ConstantIntegerType($i);
-				$valueTypesBuilder->setOffsetValueType($offsetType, $listOrArray->getOffsetValueType($offsetType));
+				$valueTypesBuilder->setOffsetValueType($offsetType, $type->getOffsetValueType($offsetType));
 			}
 			return $valueTypesBuilder->getArray();
 		}
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -971,14 +971,9 @@ class TypeSpecifier
 						continue;
 					}
 
-					if ($sizeType instanceof ConstantIntegerType && $innerType->isList()->yes() && $sizeType->getValue() < ConstantArrayTypeBuilder::ARRAY_COUNT_LIMIT) {
-						// turn optional offsets non-optional
-						$valueTypesBuilder = ConstantArrayTypeBuilder::createEmpty();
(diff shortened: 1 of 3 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -1012,6 +1012,16 @@ private function specifyTypesForConstantBinaryExpression(
 						}
 					}
 
+					if ($context->truthy() && $isSize->maybe()) {
+						// turn optional offsets non-optional
```
ondrejmirtes: I'm not sure about this piece of code.

1) it only makes sense for lists
2) Why only for `$isSize->maybe()`?

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
staabm: good point. 

> Why only for $isSize->maybe()

we can do it for anything, but I think only in a maybe-sized array we can find optional offsets.
but I think you are right thats its kind of argumenting backwards.

fixed and added test

src/Analyser/TypeSpecifier.php
```diff
@@ -1005,6 +1005,16 @@ private function specifyTypesForConstantBinaryExpression(
 						if ($isSize->no()) {
 							continue;
 						}
+
+						if ($innerType->isList()->yes()) {
+							// turn optional offsets non-optional
+							$valueTypesBuilder = ConstantArrayTypeBuilder::createEmpty();
+							for ($i = 0; $i < $constantType->getValue(); $i++) {
+								$offsetType = new ConstantIntegerType($i);
+								$valueTypesBuilder->setOffsetValueType($offsetType, $innerType->getOffsetValueType($offsetType));
```
ondrejmirtes: What about doing intersect with HasOffsetType instead of the for loop?

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
staabm: I tried using `HasOffsetValueType` but it still requires the loop

```diff
-                                               $valueTypesBuilder = ConstantArrayTypeBuilder::createEmpty();
+                                               $hasOffsetTypes = [];
                                                for ($i = 0; $i < $constantType->getValue(); $i++) {
-                                                       $offsetType = new ConstantIntegerType($i);
-                                                       $valueTypesBuilder->setOffsetValueType($offsetType, $argType->getOffsetValueType($offsetType));
+                                                       $hasOffsetTypes[] = new HasOffsetValueType(new ConstantIntegerType($i), $argType->getOffsetValueType(new ConstantIntegerType(
$i)));
                                                }
-                                               $valueTypes = $this->create($exprNode->getArgs()[0]->value, $valueTypesBuilder->getArray(), $context, false, $scope, $rootExpr);
+                                               $valueTypes = $this->create($exprNode->getArgs()[0]->value, TypeCombinator::intersect($argType, ...$hasOffsetTypes), $context, false,
 $scope, $rootExpr);
```

and it leads to less precise types, I think (at least less readable)

```
--- Expected
+++ Actual
@@ @@
-'array{ListCount\A, ListCount\A, ListCount\A}'
+'list<ListCount\A>&hasOffsetValue(0, ListCount\A)&hasOffsetValue(1, ListCount\A)&hasOffsetValue(2, ListCount\A)'
```

staabm: to drop the loop we could maybe use `new HasOffsetValueType(IntegerRangeType::fromInterval(0, $constantType->getValue(), $offsetValueType)` but this doesn't work because $offsetValueType is different per offset

ondrejmirtes:
Conflict :)

src/Analyser/TypeSpecifier.php
```diff
@@ -986,6 +991,35 @@ private function narrowUnionByArraySize(FuncCall $countFuncCall, UnionType $argT
...
+			$isNormalCount = TrinaryLogic::createYes();
+		} else {
+			$mode = $scope->getType($countFuncCall->getArgs()[1]->value);
+			$isNormalCount = (new ConstantIntegerType(COUNT_NORMAL))->isSuperTypeOf($mode)->or($argType->getIterableValueType()->isArray()->negate());
+		}
+
+		if (
+			$isNormalCount->yes()
+			&& $type->isList()->yes()
+			&& $sizeType instanceof ConstantIntegerType
```
staabm, line 1008: Note to future me: Misses int-range support?

ondrejmirtes:
Thank you!

src/Analyser/TypeSpecifier.php
staabm: here we go: https://github.com/phpstan/phpstan-src/pull/3335

PR #3310, opened 2024-08-09
Improve sprintf support

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11491

### DIFF
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -125,32 +125,45 @@ class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunctionReturn
 			return $singlePlaceholderEarlyReturn;
 		}
 
-		$isNonEmpty = $allPatternsNonEmpty;
-		if (
-			count($formatStrings) === 0
-			&& $functionReflection->getName() === 'sprintf'
-			&& count($args) === 2
-			&& $formatType->isNonEmptyString()->yes()
-			&& $scope->getType($args[1]->value)->isNonEmptyString()->yes()
-		) {
-			$isNonEmpty = true;
-		}
-
 		if ($allPatternsNonFalsy) {
-			$returnType = new IntersectionType([
+			return new IntersectionType([
 				new StringType(),
 				new AccessoryNonFalsyStringType(),
 			]);
-		} elseif ($isNonEmpty) {
-			$returnType = new IntersectionType([
+		}
+
+		$isNonEmpty = $allPatternsNonEmpty;
+		if (
+			!$isNonEmpty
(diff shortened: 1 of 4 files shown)


src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -125,32 +125,45 @@ public function getTypeFromFunctionCall(
 			return $singlePlaceholderEarlyReturn;
 		}
 
+		if ($allPatternsNonFalsy) {
```
VincentLanglet, line 128: Early return to avoid the following computation.

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -125,32 +125,45 @@ public function getTypeFromFunctionCall(
...
+			return new IntersectionType([
+				new StringType(),
+				new AccessoryNonFalsyStringType(),
+			]);
+		}
+
 		$isNonEmpty = $allPatternsNonEmpty;
 		if (
-			count($formatStrings) === 0
+			!$isNonEmpty
```
VincentLanglet, line 137: If it's already known as non empty, we don't have to compute this.

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -125,32 +125,45 @@ public function getTypeFromFunctionCall(
...
 			&& $functionReflection->getName() === 'sprintf'
-			&& count($args) === 2
+			&& count($args) >= 2
```
VincentLanglet, line 139: There was a special case for sprintf($nonEmpty, $nonEmpty, $nonEmpty) but this can be extended to `sprintf($nonEmpty, $nonEmpty, $nonEmpty)` and so on.

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -125,32 +125,45 @@ public function getTypeFromFunctionCall(
...
-			&& $scope->getType($args[1]->value)->isNonEmptyString()->yes()
 		) {
-			$isNonEmpty = true;
+			$allArgsNonEmpty = true;
+			foreach ($args as $key => $arg) {
+				if ($key === 0) {
+					continue;
+				}
+
+				if (!$scope->getType($arg->value)->toString()->isNonEmptyString()->yes()) {
```
VincentLanglet, line 148: The toString was added for things like IntegerRange which need to be casted before the isNonEmptyString check.

VincentLanglet (commented):
Since you made https://github.com/phpstan/phpstan-src/pull/3192 you might want to review @staabm

ondrejmirtes:
Thank you!

PR #3312, opened 2024-08-09
add callback types for array_uintersect etc v2

schlndh:
This is a more advanced alternative to https://github.com/phpstan/phpstan-src/pull/3282
Fixes https://github.com/phpstan/phpstan/issues/7707

I didn't go the route you suggested here: https://github.com/phpstan/phpstan-src/pull/3282#issuecomment-2266653829 . Instead, I bypassed php8-stubs for these functions so that they don't conflict with the new stubs. And I added support for in-the-middle variadic parameter to `FunctionCallParametersCheck`. IMO handling it directly in the reflection should cause less inconsistency.


src/Reflection/SignatureMap/Php8SignatureMapProvider.php
```diff
@@ -185,7 +186,10 @@ public function getMethodSignatures(string $className, string $methodName, ?Refl
 	public function getFunctionSignatures(string $functionName, ?string $className, ReflectionFunctionAbstract|null $reflectionFunction): array
 	{
 		$lowerName = strtolower($functionName);
-		if (!array_key_exists($lowerName, $this->map->functions)) {
+		if (
+			!array_key_exists($lowerName, $this->map->functions)
+			|| ($reflectionFunction !== null && $reflectionFunction->getDocComment() !== false && str_contains($reflectionFunction->getDocComment(), '@phpstan-skip-php8-stubs'))
```
schlndh, line 191: I didn't want to complicate it with the php8-stubs. The issue is that the stubs look like this:

```php
/** @param array|callable $rest */
function array_intersect_uassoc(array $array, ...$rest): array
{
}
```

which then overrides the newly introduced stub. So I needed to skip them somehow. I suppose an alternative would be to skip php8-stubs where the last parameter is `...$rest`, but I prefer this more explicit way.

src/Type/TypehintHelper.php
```diff
@@ -201,8 +201,11 @@ public static function decideType(
 			}
 
 			if (
-				(!$phpDocType instanceof NeverType || ($type instanceof MixedType && !$type->isExplicitMixed()))
-				&& $type->isSuperTypeOf(TemplateTypeHelper::resolveToBounds($phpDocType))->yes()
+				($type->isCallable()->yes() && $phpDocType->isCallable()->yes())
```
schlndh, line 204: Alternatively, I could broaden the type of the callbacks in function map to `callable` and then I wouldn't need this. I'm not sure what's better.

src/Reflection/SignatureMap/Php8SignatureMapProvider.php
```diff
@@ -185,7 +186,29 @@ public function getMethodSignatures(string $className, string $methodName, ?Refl
...
+						'array_udiff',
+						'array_udiff_assoc',
+						'array_udiff_uassoc',
+						'array_uintersect',
+						'array_uintersect_assoc',
+						'array_uintersect_uassoc',
+					],
+					true,
+				)
+			)
```
schlndh, line 210: I didn't want to complicate it with the php8-stubs. The issue is that the stubs look like this:

```php
/** @param array|callable $rest */
function array_intersect_uassoc(array $array, ...$rest): array
{
}
```

which then overrides the newly introduced stub. So I needed to skip them somehow. I suppose an alternative would be to skip php8-stubs where the last parameter is `...$rest`, but I prefer this more explicit way.

ondrejmirtes:
Hi, I appreciate your effort but I'm not really sure about this approach. This isn't really valid PHP:

```php
function array_udiff(
    array $array,
    array ...$arrays,
    callable $value_compare_func,
): array {}
```

So stubs shouldn't be written like this and info with this structure should not appear in reflection. The fact you need to handle this situation in src/Rules/FunctionCallParametersCheck.php (in the rules layer) is a sign we might break a bunch of code that would forget to handle it.

To modify the signature on the fly based on the actual call args with my suggestion from here https://github.com/phpstan/phpstan-src/pull/3282#issuecomment-2266653829 would be much more palatable for me.

PR #3313, opened 2024-08-09
Handle condition with always true/false sub-condition

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11276

### DIFF
--- a/tests/PHPStan/Analyser/TypeSpecifierTest.php
+++ b/tests/PHPStan/Analyser/TypeSpecifierTest.php
@@ -70,7 +70,9 @@ class TypeSpecifierTest extends PHPStanTestCase
 		$this->scope = $this->scope->assignVariable('barOrFalse', new UnionType([new ObjectType('Bar'), new ConstantBooleanType(false)]), new UnionType([new ObjectType('Bar'), new ConstantBooleanType(false)]));
 		$this->scope = $this->scope->assignVariable('stringOrFalse', new UnionType([new StringType(), new ConstantBooleanType(false)]), new UnionType([new StringType(), new ConstantBooleanType(false)]));
 		$this->scope = $this->scope->assignVariable('array', new ArrayType(new MixedType(), new MixedType()), new ArrayType(new MixedType(), new MixedType()));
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/TypeSpecifierTest.php
```diff
@@ -349,9 +351,17 @@ public function dataCondition(): iterable
 					$this->createFunctionCall('is_int', 'foo'),
 					$this->createFunctionCall('is_string', 'bar'),
 				),
-				[],
+				['$foo' => 'int'],
```
VincentLanglet, line 354: Since `bar` is an object
```
$this->scope = $this->scope->assignVariable('bar', new ObjectType('Bar'), new ObjectType('Bar'));
```
we can be sure foo is an int.

I added the same test with `is_int($foo) || is_string($mixed)` to reproduce the old behavior.

tests/PHPStan/Analyser/TypeSpecifierTest.php
```diff
@@ -648,19 +658,37 @@ public function dataCondition(): iterable
 			[
 				new Expr\Empty_(new Variable('array')),
 				[
-					'$array' => 'array{}|null',
```
VincentLanglet, line 651: Since `array` is defined as an non nullable array
```
$this->scope = $this->scope->assignVariable('array', new ArrayType(new MixedType(), new MixedType()), new ArrayType(new MixedType(), new MixedType()));
```
we can be sure it's an array.

staabm:
@VincentLanglet do you remember why you closed this PR?

I think this fix would be useful to [fix a few of the ~1200 errors](https://github.com/phpstan/phpstan/issues/13714#issuecomment-3436032994)

VincentLanglet:
> @VincentLanglet do you remember why you closed this PR?
> 
> I think this fix would be useful to [fix a few of the ~1200 errors](https://github.com/phpstan/phpstan/issues/13714#issuecomment-3436032994)

No I don't... Maybe it wasn't working ?

You can try to reopen it on 2.1.x to see the result of the build

PR #3314, opened 2024-08-09
RegexArrayShapeMatcher - Fix matching literal "-" in character classes

staabm:
closes https://github.com/phpstan/phpstan/issues/11490


ondrejmirtes:
Thank you!

PR #3317, opened 2024-08-10
Fix array_intersect_key with non empty array

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10561

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/array-intersect-key.php
+++ b/tests/PHPStan/Analyser/nsrt/array-intersect-key.php
@@ -15,9 +15,9 @@ class Foo
 	public function nonEmpty(array $arr, array $arr2): void
 	{
 		assertType('non-empty-array<int|string, string>', array_intersect_key($arr));
-		assertType('non-empty-array<int|string, string>', array_intersect_key($arr, $arr));
+		assertType('array<int|string, string>', array_intersect_key($arr, $arr));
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/nsrt/array-intersect-key.php
```diff
@@ -15,9 +15,9 @@ class Foo
 	public function nonEmpty(array $arr, array $arr2): void
 	{
 		assertType('non-empty-array<int|string, string>', array_intersect_key($arr));
-		assertType('non-empty-array<int|string, string>', array_intersect_key($arr, $arr));
+		assertType('array<int|string, string>', array_intersect_key($arr, $arr));
```
VincentLanglet, line 18: I'm not sure it's worth adding complexity to the ArrayIntersectKeyFunctionReturnTypeExtension in order to check if all the variable are the same. This array_intersect_key  call is useless...

staabm: The test is only about the types. No need for checking identical vars passed

VincentLanglet: Then 
```
array_intersect_key($nonEmptyArray1, $nonEmptyArray2)
```
is now correctly understood as an array that may be empty.

ondrejmirtes:
Thank you!

PR #3319, opened 2024-08-11
Improve non strict in_array

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9436

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/in_array_loose.php
+++ b/tests/PHPStan/Analyser/nsrt/in_array_loose.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace InArrayLoose;
+
+use function PHPStan\Testing\assertType;
+
+class Foo
+{
+	public function looseComparison(
+		string $string,
+		int $int,
+		float $float,
+		bool $bool,
+		string|int $stringOrInt,
+		string|null $stringOrNull,
+	): void {
+		if (in_array($string, ['1', 'a'])) {
+			assertType("'1'|'a'", $string);
+		}
+		if (in_array($string, [1, 'a'])) {
+			assertType("string", $string); // could be '1'|'a'
+		}
+		if (in_array($int, [1, 2])) {
+			assertType('1|2', $int);
+		}
+		if (in_array($int, ['1', 2])) {
+			assertType('int', $int); // could be 1|2
+		}
+		if (in_array($bool, [true])) {
+			assertType('true', $bool);
+		}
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/nsrt/in_array_loose.php
```diff
@@ -0,0 +1,48 @@
...
+		float $float,
+		bool $bool,
+		string|int $stringOrInt,
+		string|null $stringOrNull,
+	): void {
+		if (in_array($string, ['1', 'a'])) {
+			assertType("'1'|'a'", $string);
+		}
+		if (in_array($string, [1, 'a'])) {
+			assertType("string", $string);
```
VincentLanglet: Some assertion will be improved later for example when https://phpstan.org/r/49e37d0b-b17a-4a90-9fff-0618a5c7fc41 will be improved.

VincentLanglet: Some assertion will be improved later for example when https://phpstan.org/r/49e37d0b-b17a-4a90-9fff-0618a5c7fc41 will be improved.

staabm: Usually I would drop a comment in the test code like `could be xyz` to make people aware that a more precise result is expected.

The review comment itself is usually not that helpful

ondrejmirtes:
Fixes https://github.com/phpstan/phpstan/issues/9752 too, please add a regression test.

VincentLanglet:
> Fixes [phpstan/phpstan#9752](https://github.com/phpstan/phpstan/issues/9752) too, please add a regression test.

Hi, 

It fixes this comment https://github.com/phpstan/phpstan/issues/9752#issuecomment-1697288324
with the snippet https://phpstan.org/r/4cbdb7cd-ab24-4a52-a816-b1e378cc11b4

but it doesn't fix the original issue https://github.com/phpstan/phpstan/issues/9752#issue-1847075031 with the snippet https://phpstan.org/r/341e08ac-54fe-4f45-9586-703c1707060c if I understand correctly the issue bot.

So it doesn't close https://github.com/phpstan/phpstan/issues/9752.
Do you still want a non-regression test for the snippet it fixes ?

ondrejmirtes:
Sorry I didn't notice it, forget what I said, we don't need a test for that :)

ondrejmirtes:
Thank you!

PR #3320, opened 2024-08-11
Handle BenevolentUnionType of array|float|int PLUS mixed

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/3759


ondrejmirtes:
Thank you!

PR #3324, opened 2024-08-13
Narrow arrays in union based on count() with smaller/greater operator

staabm:
closes https://github.com/phpstan/phpstan/issues/11480

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -253,7 +253,7 @@ class TypeSpecifier
 						$sizeType = IntegerRangeType::createAllGreaterThan($leftType->getValue());
 					}
 
-					$narrowed = $this->narrowUnionBySize($expr->right, $argType, $sizeType, $context, $scope, $rootExpr);
+					$narrowed = $this->narrowUnionByArraySize($expr->right, $argType, $sizeType, $context, $scope, $rootExpr);
 					if ($narrowed !== null) {
 						return $narrowed;
 					}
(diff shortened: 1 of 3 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -244,12 +244,51 @@ public function specifyTypesInCondition(
...
+								continue;
+							}
+						}
+						if ($context->falsey()) {
+							if (!$isSize->yes()) {
+								continue;
+							}
+						}
+
+						$result[] = $innerType;
```
ondrejmirtes: I'd like some DRY around this, we already have the same code elsewhere. We could have a method that just accepts the narrowed array size as Type, and return SpecifiedTypes.

src/Analyser/TypeSpecifier.php
```diff
@@ -244,12 +244,51 @@ public function specifyTypesInCondition(
...
+
+				if (
+					$isNormalCount->yes()
+					&& $argType instanceof UnionType
+					&& $leftType instanceof ConstantIntegerType
+				) {
+					if ($orEqual) {
+						$constantType = IntegerRangeType::createAllGreaterThanOrEqualTo($leftType->getValue());
+					} else {
+						$constantType = IntegerRangeType::createAllGreaterThan($leftType->getValue());
```
ondrejmirtes: We already have this elsewhere:

```php
					$context->true() && (IntegerRangeType::createAllGreaterThanOrEqualTo(1 - $offset)->isSuperTypeOf($leftType)->yes())
					|| ($context->false() && (new ConstantIntegerType(1 - $offset))->isSuperTypeOf($leftType)->yes())
```

Would be nice to reuse that.

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
staabm: I don't think its the same thing. in the first block we work on the concrete constant integer, used in the comparison while in the second block we just check for the comparison value beeing >0 or >1.

src/Analyser/TypeSpecifier.php
staabm: done

ondrejmirtes:
Awesome, thank you!

PR #3325, opened 2024-08-13
Improve IntDivThrowTypeExtension to support IntRange

VincentLanglet:


src/Type/Php/IntdivThrowTypeExtension.php
```diff
@@ -29,39 +28,19 @@ public function getThrowTypeFromFunctionCall(FunctionReflection $functionReflect
...
-			}
-
-			$valueType = TypeCombinator::remove($valueType, $constantScalarType);
-		}
-
-		if (!$valueType instanceof NeverType) {
-			$containsMin = true;
-		}
+		$valueType = $scope->getType($funcCall->getArgs()[0]->value)->toInteger();
+		$containsMin = $valueType->acceptsWithReason(new ConstantIntegerType(PHP_INT_MIN), true);
```
ondrejmirtes: isSuperTypeOf() should be used instead of accepts()

ondrejmirtes (changes requested):

src/Type/Php/IntdivThrowTypeExtension.php
VincentLanglet: Updated.

Out of curiosity, what's the diff. Seems like both works.

ondrejmirtes: https://github.com/phpstan/phpstan/discussions/11460

VincentLanglet: Thanks

ondrejmirtes:
Thank you.

PR #3326, opened 2024-08-15
Fix NumericString to array key

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/4163
Fixes https://github.com/phpstan/phpstan/issues/4671 (already closed with wrong result)
Closes https://github.com/phpstan/phpstan/issues/8592
Closes https://github.com/phpstan/phpstan/issues/11390
Closes https://github.com/phpstan/phpstan/issues/12413

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -609,11 +609,17 @@ class ConstantArrayType extends ArrayType implements ConstantType
 
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
-		$offsetType = $offsetType->toArrayKey();
+		$offsetArrayKeyType = $offsetType->toArrayKey();
+
+		return $this->recursiveHasOffsetValueType($offsetArrayKeyType);
+	}
+
+	private function recursiveHasOffsetValueType(Type $offsetType): TrinaryLogic
+	{
 		if ($offsetType instanceof UnionType) {
 			$results = [];
 			foreach ($offsetType->getTypes() as $innerType) {
-				$results[] = $this->hasOffsetValueType($innerType);
+				$results[] = $this->recursiveHasOffsetValueType($innerType);
 			}
 
 			return TrinaryLogic::extremeIdentity(...$results);
@@ -623,7 +629,7 @@ class ConstantArrayType extends ArrayType implements ConstantType
 			if ($finiteTypes !== []) {
 				$results = [];
 				foreach ($finiteTypes as $innerType) {
-					$results[] = $this->hasOffsetValueType($innerType);
+					$results[] = $this->recursiveHasOffsetValueType($innerType);
 				}
 
 				return TrinaryLogic::extremeIdentity(...$results);
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -1029,7 +1029,10 @@ class IntersectionType implements CompoundType
 	public function toArrayKey(): Type
 	{
 		if ($this->isNumericString()->yes()) {
-			return new IntegerType();
+			return new UnionType([
+				new IntegerType(),
+				$this,
+			]);
 		}
 
 		if ($this->isString()->yes()) {
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -1029,10 +1029,10 @@ class IntersectionType implements CompoundType
 	public function toArrayKey(): Type
 	{
 		if ($this->isNumericString()->yes()) {
-			return new UnionType([
+			return TypeCombinator::union(
 				new IntegerType(),
 				$this,
-			]);
+			);
 		}
 
 		if ($this->isString()->yes()) {
--- a/tests/PHPStan/Rules/Variables/IssetRuleTest.php
+++ b/tests/PHPStan/Rules/Variables/IssetRuleTest.php
@@ -233,10 +233,7 @@ class IssetRuleTest extends RuleTestCase
 	{
 		$this->treatPhpDocTypesAsCertain = true;
 		$this->strictUnnecessaryNullsafePropertyFetch = false;
-		$this->analyse([__DIR__ . '/data/bug-4671.php'], [[
-			'Offset numeric-string on array<string, string> in isset() does not exist.',
-			13,
-		]]);
+		$this->analyse([__DIR__ . '/data/bug-4671.php'], []);
 	}
 
 	public function testVariableCertaintyInIsset(): void
--- a/src/Type/Accessory/AccessoryNumericStringType.php
+++ b/src/Type/Accessory/AccessoryNumericStringType.php
@@ -229,7 +229,13 @@ class AccessoryNumericStringType implements CompoundType, AccessoryType
 
 	public function toArrayKey(): Type
 	{
-		return new IntegerType();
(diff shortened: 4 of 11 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -602,17 +602,20 @@ public function findTypeAndMethodNames(): array
...
-		$offsetType = $offsetType->toArrayKey();
-		if ($offsetType instanceof UnionType) {
+		$offsetArrayKeyType = $offsetType->toArrayKey();
+
+		if ($offsetArrayKeyType instanceof UnionType) {
 			$results = [];
-			foreach ($offsetType->getTypes() as $innerType) {
-				$results[] = $this->hasOffsetValueType($innerType);
+			foreach ($offsetArrayKeyType->getTypes() as $innerType) {
+				if (!$innerType->isSuperTypeOf($offsetType)->yes()) {
```
VincentLanglet: Since numeric-string->toArrayKey returns int|numeric-string, we have a infinite loop without this condition.

Not sure if this is the best fix thought...

tests/PHPStan/Rules/Variables/IssetRuleTest.php
```diff
@@ -233,10 +233,7 @@ public function testBug4671(): void
 	{
 		$this->treatPhpDocTypesAsCertain = true;
 		$this->strictUnnecessaryNullsafePropertyFetch = false;
-		$this->analyse([__DIR__ . '/data/bug-4671.php'], [[
-			'Offset numeric-string on array<string, string> in isset() does not exist.',
```
VincentLanglet, line 237: This was wrong

src/Type/Constant/ConstantArrayType.php
```diff
@@ -602,11 +602,17 @@ public function findTypeAndMethodNames(): array
...
+		return $this->recursiveHasOffsetValueType($offsetArrayKeyType);
+	}
+
+	private function recursiveHasOffsetValueType(Type $offsetType): TrinaryLogic
+	{
 		if ($offsetType instanceof UnionType) {
 			$results = [];
 			foreach ($offsetType->getTypes() as $innerType) {
-				$results[] = $this->hasOffsetValueType($innerType);
+				$results[] = $this->recursiveHasOffsetValueType($innerType);
```
VincentLanglet, line 622: With the previous logic, toArrayKey was again called, creating a new UnionType for numeric string and an infinite loop.

ondrejmirtes:
Hey, this area is really tricky. This isn't the right fix, I'd tackle this completely differently. Your change would break too many scenarios.

My idea is: Introduce `non-int-string` type, and with an optional flag being false by default, I'd interpret `array<string, X>` as `array<non-int-string, X>`. Which would make the analysis stricter and correct. That would be a step in the right direction of fixing arrays-with-string-keys issues.

We can't fix `numeric-string`. Asking `is_numeric()` does not tell us anything about what's going to happen inside array keys. That's currently broken design. Because only "integer strings" are cast to integers in arrays, but numeric strings also include "float strings"

ondrejmirtes:
I'll give these changes a bit more thought, stay tuned and thanks, I appreciate it 😊

VincentLanglet:
> Hey, this area is really tricky.

I agree

> This isn't the right fix, I'd tackle this completely differently. Your change would break too many scenarios.

Which one ? Looking at tests and issues, it seems to only solve issues without breaking tests.

I was worried too that it would open a rabbit hole, but according to the CI it seems not.

I agree it would be the case if I wanted to change the behavior for `string` but here it's only numeric-string.
And it seems weird that currently the intersection of `string->toArrayKey` and `numeric-string->toArrayKey` is NEVER.
 
> My idea is: Introduce `non-int-string` type, and with an optional flag being false by default, I'd interpret `array<string, X>` as `array<non-int-string, X>`. Which would make the analysis stricter and correct. That would be a step in the right direction of fixing arrays-with-string-keys issues.

I definitely agree on this `non-int-string` type, but here I think it's different since we're on numeric-string. 

Currently string->toArrayKey is considered as string but numeric-string->toArrayKey is considered as int.
It seems better to me to consider that numeric-string->toArrayKey might still be a string,
I especially didn't touch to the string->toArrayKey behavior to avoid the tricky part.

> We can't fix `numeric-string`. Asking `is_numeric()` does not tell us anything about what's going to happen inside array keys. That's currently broken design. Because only "integer strings" are cast to integers in arrays, but numeric strings also include "float strings"

That especially because numeric-string can include float string that the arrayKey shouldn't be considered as only int.

Imho, 'numeric-string' should be considered as `int|numeric-string` as array key,
while `int-string` and `non-int-string` will be considered respectively as `int` and `string` as array key the day a native php `is_int_string` will be introduced (I just tried an RFC on php internals about such functions to know if it could be introduced in next php version).

> I'll give these changes a bit more thought, stay tuned and thanks, I appreciate it 😊

Sure, I'm not in hurry.

VincentLanglet:
> I'll give these changes a bit more thought, stay tuned and thanks, I appreciate it 😊

I rebased the PR just in case.

VincentLanglet:
> I'll give these changes a bit more thought, stay tuned and thanks, I appreciate it 😊

Hi @ondrejmirtes, you planned to think again about these change. 
Since it's been 6 month, what do you want me to do with this PR ?

Is it worth rebasing it for a new review ?
Do you prefer me to target 2.1.x ?

VincentLanglet:
Hi, I rebased the branch and really would like to move forward this PR if possible @ondrejmirtes. Can we talk again about it ?

I do agree that one day we will need `non-int-string`. But I feel like it’s not the same topic. 
Currently PHPStan consider that numeric-string is the same than `int-string`, but it’s not. Numeric-string should be considered as `int-string|float-string` and while `int-string` are casted to `int` when used as array key, `float-string` are still `float-string` when used as array-key.  

So NumericString->toArrayKey should be Int|Numeric-string, and not just Int.

At first sight, with all the existing non-regression tests, this change seems to fix a lot of existing issues without introducing a regresion.
Do you prefer this PR to target 2.1.x ?

src/Type/IntersectionType.php
```diff
@@ -1029,7 +1029,10 @@ public function toArray(): Type
 	public function toArrayKey(): Type
 	{
 		if ($this->isNumericString()->yes()) {
-			return new IntegerType();
+			return new UnionType([
```
ondrejmirtes: This should be `TypeCombinator::union()` instead of `new UnionType` to ensure type normalization.

ondrejmirtes (changes requested):
I'm willing to try this again and see what the impact is on real-world projects.

ondrejmirtes:
Thank you.

VincentLanglet:
You're welcome, please feel free to ping me after your tests if you need extra work on this and I can help.

ondrejmirtes:
This is causing infinite recursion on one of the test projects. I'm investigating details.

VincentLanglet:
> This is causing infinite recursion on one of the test projects. I'm investigating details.

I had one infinite loop with PHPStan codebase/tests which I fixed this way https://github.com/phpstan/phpstan-src/pull/3326/files#r1718409129 if it can help for the other one you found.

ondrejmirtes:
I'm sorry, it wasn't fault of this PR but a different one.

PR #3328, opened 2024-08-17
Support multiple anonymous class definitions on the same line

tscni:
Fixes https://github.com/phpstan/phpstan/issues/5597, https://github.com/phpstan/phpstan/issues/11511

### DIFF
--- a/src/Broker/AnonymousClassNameHelper.php
+++ b/src/Broker/AnonymousClassNameHelper.php
@@ -34,7 +34,7 @@ class AnonymousClassNameHelper
 
 		return sprintf(
 			'AnonymousClass%s',
-			md5(sprintf('%s:%s', $filename, $classNode->getStartLine())),
+			md5(sprintf('%s:%s', $filename, $classNode->getStartFilePos())),
 		);
 	}
 
--- a/src/Broker/AnonymousClassNameHelper.php
+++ b/src/Broker/AnonymousClassNameHelper.php
@@ -5,6 +5,7 @@ namespace PHPStan\Broker;
 use PhpParser\Node;
 use PHPStan\File\FileHelper;
 use PHPStan\File\RelativePathHelper;
+use PHPStan\Parser\AnonymousClassVisitor;
 use PHPStan\ShouldNotHappenException;
 use function md5;
 use function sprintf;
@@ -32,9 +33,17 @@ class AnonymousClassNameHelper
 			$this->fileHelper->normalizePath($filename, '/'),
 		);
 
+		/** @var int|null $lineIndex */
+		$lineIndex = $classNode->getAttribute(AnonymousClassVisitor::ATTRIBUTE_LINE_INDEX);
+		if ($lineIndex === null) {
+			$hash = md5(sprintf('%s:%s', $filename, $classNode->getStartLine()));
+		} else {
+			$hash = md5(sprintf('%s:%s:%d', $filename, $classNode->getStartLine(), $lineIndex));
+		}
+
 		return sprintf(
 			'AnonymousClass%s',
-			md5(sprintf('%s:%s', $filename, $classNode->getStartFilePos())),
+			$hash,
 		);
 	}
 
--- a/src/Parser/AnonymousClassVisitor.php
+++ b/src/Parser/AnonymousClassVisitor.php
@@ -1,0 +1,49 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Parser;
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+use function count;
+
+class AnonymousClassVisitor extends NodeVisitorAbstract
+{
+
+	public const ATTRIBUTE_LINE_INDEX = 'anonymousClassLineIndex';
+
+	/** @var array<int, non-empty-list<Node\Stmt\Class_>> */
+	private array $nodesPerLine = [];
+
+	public function beforeTraverse(array $nodes): ?array
+	{
+		$this->nodesPerLine = [];
+		return null;
+	}
+
+	public function enterNode(Node $node): ?Node
+	{
+		if (!$node instanceof Node\Stmt\Class_ || !$node->isAnonymous()) {
+			return null;
+		}
+
+		$this->nodesPerLine[$node->getStartLine()][] = $node;
+
+		return null;
+	}
+
+	public function afterTraverse(array $nodes): ?array
+	{
+		foreach ($this->nodesPerLine as $nodesOnLine) {
+			if (count($nodesOnLine) === 1) {
+				continue;
+			}
+			for ($i = 0; $i < count($nodesOnLine); $i++) {
+				$nodesOnLine[$i]->setAttribute(self::ATTRIBUTE_LINE_INDEX, $i + 1);
+			}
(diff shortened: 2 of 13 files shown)


tscni:
It would probably be suitable to adjust this as well: https://github.com/phpstan/phpstan-src/blob/6277fb4cab46eec41ce603f3b53190faa5998c55/src/Reflection/BetterReflection/BetterReflectionProvider.php#L230

But adding the file position there isn't very helpful to users. 
PHP itself uses a 0-indexed increment suffix after the line (`:1$0`, `:1$1`, `:1$2`). Reusing that approach would seem the most ideal. I'd just have to figure out how to collect & carry that state to the respective usages.

Maybe a stateful node visitor might be a good approach for that.

ondrejmirtes:
The problem here is that the `ClassReflection::getCacheKey` consists of the display name. Which would be the same for two anonymous classes on the same line. Thus mangling the cache.

My idea for the fix: a node visitor that would add some kind of attribute on `Class_` if there are two anonymous classes on the same line.

If there was just a single class on the same line, the name generated by AnonymousClassNameHelper would not change. If there were two, the name would be `line:order`, so `24:1` and `24:2` for example.

The same would apply to the display name in ClassReflection.

src/Reflection/BetterReflection/BetterReflectionProvider.php
```diff
@@ -214,6 +214,14 @@ public function getAnonymousClassReflection(Node\Stmt\Class_ $classNode, Scope $
 			null,
 		);
 
+		/** @var int|null $classLineIndex */
+		$classLineIndex = $classNode->getAttribute(AnonymousClassVisitor::ATTRIBUTE_LINE_INDEX);
+		if ($classLineIndex === null) {
+			$displayName = sprintf('class@anonymous/%s:%s', $filename, $classNode->getStartLine());
+		} else {
+			$displayName = sprintf('class@anonymous/%s:%s:%d', $filename, $classNode->getStartLine(), $classLineIndex);
+		}
```
tscni, line 223: > If there was just a single class on the same line, the name generated by AnonymousClassNameHelper would not change. If there were two, the name would be line:order, so 24:1 and 24:2 for example.
>
> The same would apply to the display name in ClassReflection.

I've done this now and am personally happy with it.

But a small notes on this:
E.g. `path/to/file.php:12:34` would usually indicate that something is located in file `path/to/file.php` at line `12` at column / character `34`
In that regard this seems slightly unusual and might potentially lead to some confusion. If we do care to solve this (I don't mind it), maybe `$` as a separator, like the one PHP uses, might be a good alternative.
We could alternatively also use the character position of course, but that seems a bit more cumbersome to add in contexts without the reflection logic.

src/Reflection/BetterReflection/BetterReflectionProvider.php
```diff
@@ -201,7 +202,6 @@ public function getAnonymousClassReflection(Node\Stmt\Class_ $classNode, Scope $
 			$scopeFile,
 		);
 		$classNode->name = new Node\Identifier($className);
-		$classNode->setAttribute('anonymousClass', true);
```
tscni, line 204: Moving this into `AnonymousClassVisitor` felt like a reasonable step to take now that it exists.

In theory the node visitor would also offer the possibility to fix `Node\Stmt\Class_::isAnonymous()` by implementing an `AnonymousClass extends Node\Stmt\Class_` and replacing the original node with that.
But maybe that's more complexity than desired?

src/Parser/AnonymousClassVisitor.php
```diff
@@ -0,0 +1,51 @@
...
+use function count;
+
+class AnonymousClassVisitor extends NodeVisitorAbstract
+{
+
+	public const ATTRIBUTE_ANONYMOUS_CLASS = 'anonymousClass';
+	public const ATTRIBUTE_LINE_INDEX = 'anonymousClassLineIndex';
+
+	/** @var array<int, non-empty-list<Node\Stmt\Class_>> */
+	private array $nodesPerLine = [];
```
tscni, line 16: I'm slightly concerned about keeping the node reference, especially considering that nodes could be replaced.
It seems unlikely that a node visitor might replace class nodes though (except that I propose exactly that in another comment.. :P)

Ultimately it's a performance optimization to avoid having to traverse through all nodes in `afterTraverse()`.

src/Broker/AnonymousClassNameHelper.php
```diff
@@ -32,9 +33,17 @@ public function getAnonymousClassName(
...
+		$lineIndex = $classNode->getAttribute(AnonymousClassVisitor::ATTRIBUTE_LINE_INDEX);
+		if ($lineIndex === null) {
+			$hash = md5(sprintf('%s:%s', $filename, $classNode->getStartLine()));
+		} else {
+			$hash = md5(sprintf('%s:%s:%d', $filename, $classNode->getStartLine(), $lineIndex));
+		}
+
 		return sprintf(
 			'AnonymousClass%s',
-			md5(sprintf('%s:%s', $filename, $classNode->getStartLine())),
```
tscni, line 37: What's not clear to me, why differ between the class name (`AnonymousClass<hash>`) and the display name (`class@anonymous<location>`) at all?
Is it a memory optimization, avoiding problematic characters in some context, or just an artifact? Or in other words, would it be a problem to align them?

tscni: In the same turn I was thinking about setting the name already in the node visitor, but then stumbled upon `PHPStan\Parser\Parser::parseString()`.
In theory that could be amended with an optional file name argument.

That could then lead to some minor simplifications in the handling of anonymous class names.

tscni (commented):
This should work now. I left some notes / questions around possible improvements.

src/Reflection/BetterReflection/BetterReflectionProvider.php
ondrejmirtes: That's a good idea for a follow up PR! Although rules should better rely on `InClassNode` and `ClassReflection` so that they don't have to make any differences between traditional and anonymous classes, in theory someone could ask about `Class_::isAnonymous()` and it's always better to give the right answer :) Please take a note and send a follow up PR :)

src/Reflection/BetterReflection/BetterReflectionProvider.php
ondrejmirtes: I'm aware of this, but it's fine, the number should be low enough not to cause confusion with columns.

ondrejmirtes (changes requested):
This is almost perfect. I admire that as your first contribution, you really found your way around this complex codebase so easily. You didn't need any hand-holding and figured everything by yourself. Thanks!

I have one more request: there isn't really any test proving the anonymous class display name with ATTRIBUTE_LINE_INDEX. We just see the resulting hash.

A solution would be to create a dummy rule just for test purposes, like the one in VirtualNullsafeMethodCallTest. Its goal would be to report anonymous class display name, and to show that a standalone class has just `:line` and more classes with `:line:index`.

Thank you once more!

tscni:
Thanks for the kind words :)

I've now added the test.

ondrejmirtes:
Awesome, thank you!

PR #3329, opened 2024-08-17
`array_filter` with a first-class callable can narrow its return type 

takaram:
Closes phpstan/phpstan#11337

### DIFF
--- a/src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
@@ -6,22 +6,24 @@ use PhpParser\Node\Arg;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\ArrowFunction;
 use PhpParser\Node\Expr\Closure;
-use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\Error;
 use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Name;
-use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\Stmt\Return_;
 use PHPStan\Analyser\MutatingScope;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\MixedType;
@@ -35,11 +37,19 @@ use function array_map;
 use function count;
 use function in_array;
 use function is_string;
-use function strtolower;
+use function sprintf;
 use function substr;
 
-final class ArrayFilterFunctionReturnTypeReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+final class ArrayFilterFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
 {
+
+	private const USE_BOTH = 1;
+	private const USE_KEY = 2;
+	private const USE_ITEM = 3;
+
+	public function __construct(private ReflectionProvider $reflectionProvider)
+	{
+	}
 
 	public function isFunctionSupported(FunctionReflection $functionReflection): bool
 	{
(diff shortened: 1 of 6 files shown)


src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
```diff
@@ -78,64 +84,59 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-				if ($funcName === null) {
-					return new ErrorType();
-				}
-
-				$itemVar = new Variable('item');
-				$expr = new FuncCall($funcName, [new Arg($itemVar)]);
-				return $this->filterByTruthyValue($scope, $itemVar, $arrayArgType, null, $expr);
-			}
+			$mode = self::USE_ITEM;
+		} elseif ($flagArg instanceof ConstFetch && $flagArg->name->getParts()[0] === 'ARRAY_FILTER_USE_KEY') {
```
staabm: Could use `scope->getType()` instead of raw-Ast, or even https://github.com/phpstan/phpstan-src/blob/1.12.x/src/Type/BitwiseFlagHelper.php

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
```diff
@@ -78,64 +84,59 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+				$itemVar = $callbackArg->params[0]->var;
+			} elseif ($mode === self::USE_KEY) {
+				$keyVar = $callbackArg->params[0]->var;
+				$itemVar = null;
+			} elseif ($mode === self::USE_BOTH) {
+				$keyVar = $callbackArg->params[1]->var ?? null;
+				$itemVar = $callbackArg->params[0]->var;
+			}
+			return $this->filterByTruthyValue($scope, $itemVar, $arrayArgType, $keyVar, $callbackArg->expr);
+		} elseif ($callbackArg instanceof String_) {
```
staabm: Could use `Type->getConstantStrings()` instead of raw Ast nodes

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
```diff
@@ -277,4 +278,26 @@ private static function createFunctionName(string $funcName): ?Name
 		return new Name($funcName);
 	}
 
+	/**
+	 * @param self::USE_* $mode
+	 * @return array{list<Arg>, ?Variable, ?Variable}
+	 */
+	public function createDummyArgs(int $mode): array
```
staabm: private

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
```diff
@@ -78,64 +84,59 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			$funcName = self::createFunctionName($callbackArg->value);
+			if ($funcName === null) {
+				return new ErrorType();
 			}
+
+			[$args, $itemVar, $keyVar] = $this->createDummyArgs($mode);
+			$expr = new FuncCall($funcName, $args);
+			return $this->filterByTruthyValue($scope, $itemVar, $arrayArgType, $keyVar, $expr);
+		} elseif (
+			($callbackArg instanceof FuncCall || $callbackArg instanceof MethodCall || $callbackArg instanceof StaticCall)
```
staabm, line 124: Does it miss support for `NullsafeMethodCall`?

takaram: `NullsafeMethodCall` is not supported on purpose.

Quoted from https://www.php.net/manual/en/functions.first_class_callable_syntax.php:
> Note:
> The first-class callable syntax cannot be combined with the nullsafe operator. Both of the following result in a compile-time error:
>```php
><?php
>$obj?->method(...);
>$obj?->prop->method(...);
>?>
>```

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
takaram: Fixed.
I borrowed `getConstant` method from `FilterFunctionReturnTypeHelper` and `PathinfoFunctionDynamicReturnTypeExtension`.
Probably it is better to move them into a helper class, but it can be done in another PR.

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
takaram: Fixed

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
takaram: Fixed.

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
```diff
@@ -71,69 +81,62 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			&& $callbackArg->isFirstClassCallable()
+		) {
+			[$args, $itemVar, $keyVar] = $this->createDummyArgs($mode);
+			$expr = clone $callbackArg;
+			$expr->args = $args;
+			return $this->filterByTruthyValue($scope, $itemVar, $arrayArgType, $keyVar, $expr);
+		} else {
+			$constantStrings = $scope->getType($callbackArg)->getConstantStrings();
+			if (count($constantStrings) === 1) {
```
staabm: Can't you filter by all constant strings (and union the results?)

Testcase:

$fun = 'is_even';
if (rand(0,1)) $fun = 'is_int';

array_filter($array, $fun)

staabm:
Just a random idea maybe worth investigating.

would it be possible to rewrite a first-class-callable at the AST level to e.g. a regular callable (or `closure::fromCallable`), so we don't need special cases für callable and first-class-callables in every phpstan extension…?

otherwise we would need a similar fix we have in this PR in other extensions, e.g. in array_map for https://github.com/phpstan/phpstan/issues/9224

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
takaram: Done.

ondrejmirtes:
@staabm You're on the right track. I've recently started adding these things to CallableParametersAcceptor. If we added the right method to that interface, like `public function getAsserts(): Assertions`, we could simplify the extension by a lot + it'd help other places to be analysed correctly as well.

These assertions would be propagated to ClosureType when creating it from function or method in `MutatingScope:.createFirstClassCallable`.

This PR fixes array_filter extension, but a proper long-term solution is what I described here.

PR #3330, opened 2024-08-17
Allow reinitialization of a readonly property in `__clone` since PHP8.3

takaram:
Closes phpstan/phpstan#11495

### DIFF
--- a/tests/PHPStan/Rules/Properties/data/bug-11495.php
+++ b/tests/PHPStan/Rules/Properties/data/bug-11495.php
@@ -1,0 +1,24 @@
+<?php // lint >= 8.1
+declare(strict_types = 1);
+
+namespace Bug11495;
+
+class HelloWorld
+{
+	private readonly string $foo;
+
+	public function __construct()
+	{
+		$this->foo = 'bar';
+	}
+
+	public function __clone()
+	{
+		$this->foo = 'baz';
+	}
+
+	public function getFoo(): string
+	{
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Properties/data/bug-11495.php
```diff
@@ -0,0 +1,24 @@
...
+	private readonly string $foo;
+
+	public function __construct()
+	{
+		$this->foo = 'bar';
+	}
+
+	public function __clone()
+	{
+		$this->foo = 'baz';
```
ondrejmirtes, line 17: I'd like to have more tests here:

1) To verify double assignment is still detected as error (https://3v4l.org/oaLm9)
2) To verify assignment to something else than `$this` is detected as error: 

```php
	public function __clone()
	{
		$s = new self();
		$s->foo = 'baz';
	}
```

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Properties/data/bug-11495.php
takaram: I added a test for 2.

For 1., no error is reported on double assignment. This looks not so easy to implement.
I'll take a deeper look...

takaram:
Closing since addressing [this comment](https://github.com/phpstan/phpstan-src/pull/3330#discussion_r1721254159) seems hard.

PR #3333, opened 2024-08-19
ContainerFactory::clearOldContainers check only ctime, not atime

janedbal:
- When cache is pulled from some storage (e.g. S3), it gets recent atimes and this does not work. That causes unwanted cache swelling.
- Ref: https://x.com/janedbal/status/1825450878708462024

### DIFF
--- a/src/DependencyInjection/ContainerFactory.php
+++ b/src/DependencyInjection/ContainerFactory.php
@@ -218,9 +218,6 @@ class ContainerFactory
 			if ($path === false) {
 				continue;
 			}
-			if ($containerFile->getATime() > $twoDaysAgo) {
-				continue;
-			}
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I don't think this is a great fix. Other systems might have problems with different timestamp metadata.

PHPStan should keep some kind of "journal" with real access times for each generated container, and delete them based on that.

janedbal:
> PHPStan should keep some kind of "journal" with real access times for each generated container, and delete them based on that.

I like that. Closing

ondrejmirtes:
Here's a quick attempt https://github.com/phpstan/phpstan-src/commit/57c65888e6372a4056afbbacc8207d411ea8559a

janedbal:
Glad you are looking into this! This was pretty major issue in my eyes :) Thanks!

PR #3337, opened 2024-08-22
Narrow to non-falsy-string from strlen() on integer-range

staabm:


staabm:
realized we also could do better in == and ===

logged as https://github.com/phpstan/phpstan/issues/11548 as I don't want todo it right now

src/Analyser/TypeSpecifier.php
```diff
@@ -323,7 +323,8 @@ public function specifyTypesInCondition(
 					$argType = $scope->getType($expr->right->getArgs()[0]->value);
 					if ($argType->isString()->yes()) {
 						$accessory = new AccessoryNonEmptyStringType();
-						if ($leftType instanceof ConstantIntegerType && $leftType->getValue() >= 2) {
```
staabm, line 326: the old code missed to handle `$offset` which turns into off-by-one errors on smaller/greater comparisons

ondrejmirtes:
Thank you!

PR #3338, opened 2024-08-22
Remove null-return on preg_replace() with valid patterns

staabm:
~Remove null-return on preg_replace() with valid patterns~
~Return benevolent union from preg_replace()~

closes https://github.com/phpstan/phpstan/issues/11547

requires https://github.com/phpstan/phpstan-src/pull/3339

### DIFF
--- a/src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
@@ -8,6 +8,7 @@ use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\MixedType;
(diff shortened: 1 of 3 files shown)


staabm:
//cc @thg2k

thg2k:
See the original issue. I was proven wrong: a perfectly valid regexp can still fail based on the input. So maybe if the regexp looks correct we return a benevolent union with null?

staabm:
failing build in phpdoc-parser repo is related to a baselined error which no longer happens after the PR (because benevolent)

tests/PHPStan/Analyser/nsrt/non-empty-string-replace-functions.php
```diff
@@ -18,7 +18,7 @@ public function replace(string $search, string $replacement, string $subject){
         assertType('non-empty-string', str_replace($search, $replacement, $subject));
         assertType('non-empty-string', str_ireplace($search, $replacement, $subject));
 
-        assertType('non-empty-string|null', preg_replace($search, $replacement, $subject));
+        assertType('(non-empty-string|null)', preg_replace($search, $replacement, $subject));
```
thg2k: Seems a preexisting problem, but it's wrong: `preg_replace('/^a(b?)/', '\\1', 'a');`

staabm: Please open a new issue

ondrejmirtes:
Do we want this? Why it would be benevolent? Do we not want the user to always check for errors?

staabm:
I think it was one of the requests of the initial issue. dropped benevolence now and reduced it to the bugfix only

ondrejmirtes:
Thank you!

PR #3341, opened 2024-08-22
Fix return type of `array_reverse()` with `$preserve_keys = true` for lists

herndlm:


src/Type/Php/ArrayReverseFunctionReturnTypeExtension.php
```diff
@@ -43,6 +45,16 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
 		}
 
+		if ($preserveKeys) {
+			// Remove list
+			$newArrayType = new ArrayType($type->getIterableKeyType(), $type->getIterableValueType());
+			if ($type->isIterableAtLeastOnce()->yes()) {
+				$newArrayType = TypeCombinator::intersect($newArrayType, new NonEmptyArrayType());
+			}
+
+			return $newArrayType;
```
herndlm, line 55: An almost identical construct exists already somewhere. And I might add it another time somewhere else. I was wondering if it makes sense to re-use this somehow somewhere..

herndlm: the cleaner alternative of course would be a refactor that moves `reverse()` to `Type` (arrayReverse()?). The more I think about it, the more I like it 😊

herndlm: yeah, moving this to type fixes like at least 3 other edge cases as well....

herndlm:
Superseded by https://github.com/phpstan/phpstan-src/pull/3344

PR #3342, opened 2024-08-22
Narrow string on strlen() == and === comparison with integer-range

staabm:
closes https://github.com/phpstan/phpstan/issues/11548

----

after implementing I realized this also makes `count($arr) == integer-range` narrow to `non-empty-array`

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1056,13 +1056,19 @@ final class TypeSpecifier
 
 	private function specifyTypesForConstantBinaryExpression(
 		Expr $exprNode,
-		ConstantScalarType $constantType,
+		Type $constantType,
 		TypeSpecifierContext $context,
 		Scope $scope,
 		?Expr $rootExpr,
 	): ?SpecifiedTypes
 	{
-		if (!$context->null() && $constantType->getValue() === false) {
+		$scalarValues = $constantType->getConstantScalarValues();
+		if (count($scalarValues) !== 1) {
+			return null;
+		}
+		$constValue = $scalarValues[0];
+
+		if (!$context->null() && $constValue === false) {
 			$types = $this->create($exprNode, $constantType, $context, false, $scope, $rootExpr);
 			if ($exprNode instanceof Expr\NullsafeMethodCall || $exprNode instanceof Expr\NullsafePropertyFetch) {
 				return $types;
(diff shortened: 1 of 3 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -2172,6 +2172,29 @@ public function resolveIdentical(Expr\BinaryOp\Identical $expr, Scope $scope, Ty
 			}
 		}
 
+		if ($rightType instanceof IntegerRangeType) {
```
staabm: inspired by the previous loop with constant-strings, I have added a similar case for integer ranges

src/Analyser/TypeSpecifier.php
```diff
@@ -1090,10 +1096,6 @@ private function specifyTypesForConstantBinaryExpression(
 			));
 		}
 
-		if ($constantType->getValue() === null) {
-			return $this->create($exprNode, $constantType, $context, false, $scope, $rootExpr);
-		}
-
```
staabm, line 1096: looked like dead code

src/Analyser/TypeSpecifier.php
```diff
@@ -2149,10 +2157,19 @@ public function resolveIdentical(Expr\BinaryOp\Identical $expr, Scope $scope, Ty
 			}
 		}
 
-		if (count($rightType->getConstantStrings()) > 0) {
+		if ($rightType->isInteger()->yes() || $rightType->isString()->yes()) {
 			$types = null;
-			foreach ($rightType->getConstantStrings() as $constantString) {
-				$specifiedType = $this->specifyTypesForConstantStringBinaryExpression($unwrappedLeftExpr, $constantString, $context, $scope, $rootExpr);
+			foreach ($rightType->getFiniteTypes() as $finiteType) {
```
staabm, line 2170: I need to check whether this leads to dead code in IntegerRange handling somewhere

src/Analyser/TypeSpecifier.php
```diff
@@ -1102,6 +1104,10 @@ private function specifyTypesForConstantBinaryExpression(
 			&& in_array(strtolower((string) $exprNode->name), ['count', 'sizeof'], true)
 			&& $constantType instanceof ConstantIntegerType
```
staabm, line 1105: in a followup PR I would factor out the 2 `$constantType instanceof ConstantIntegerType` if-blocks of this method into the parent caller and do a proper integer range check instead, which will fix https://github.com/phpstan/phpstan/issues/11558

I don't do it now to keep this PR simpler and the next one will also be easier to review

ondrejmirtes:
Thank you!

PR #3343, opened 2024-08-22
Repair PhpParser\Node\Stmt\Class_::isAnonymous()

tscni:
As proposed here: https://github.com/phpstan/phpstan-src/pull/3328#discussion_r1722052102

### DIFF
--- a/src/Parser/AnonymousClassVisitor.php
+++ b/src/Parser/AnonymousClassVisitor.php
@@ -27,17 +27,7 @@ class AnonymousClassVisitor extends NodeVisitorAbstract
 			return null;
 		}
 
-		$node = new AnonymousClassNode(
-			$node->name,
-			[
-				'flags' => $node->flags,
-				'extends' => $node->extends,
-				'implements' => $node->implements,
-				'stmts' => $node->stmts,
-				'attrGroups' => $node->attrGroups,
-			],
-			$node->getAttributes(),
-		);
+		$node = AnonymousClassNode::createFromClassNode($node);
 		$this->nodesPerLine[$node->getStartLine()][] = $node;
 
 		return $node;
--- a/src/Parser/AnonymousClassVisitor.php
+++ b/src/Parser/AnonymousClassVisitor.php
@@ -4,15 +4,15 @@ namespace PHPStan\Parser;
 
 use PhpParser\Node;
 use PhpParser\NodeVisitorAbstract;
+use PHPStan\Node\AnonymousClassNode;
 use function count;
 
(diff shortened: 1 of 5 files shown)


src/Parser/AnonymousClassVisitor.php
```diff
@@ -27,10 +27,20 @@ public function enterNode(Node $node): ?Node
...
-		$node->setAttribute(self::ATTRIBUTE_ANONYMOUS_CLASS, true);
+		$node = new AnonymousClassNode(
+			$node->name,
+			[
+				'flags' => $node->flags,
+				'extends' => $node->extends,
+				'implements' => $node->implements,
+				'stmts' => $node->stmts,
+				'attrGroups' => $node->attrGroups,
+			],
```
tscni, line 38: This feels a bit brittle.. if this array is ever extended we'll have a "broken" class node. This might e.g. happen if PHP adds support for generics.

Maybe this could serve as a case to demonstrate to https://github.com/nikic/PHP-Parser that an alternative implementation with e.g. a separate flag might be more optimal.
It's a bit unclear what the backward compatibility promise is there, so I'll just ask.

src/Parser/AnonymousClassVisitor.php
```diff
@@ -27,10 +27,20 @@ public function enterNode(Node $node): ?Node
 			return null;
 		}
 
-		$node->setAttribute(self::ATTRIBUTE_ANONYMOUS_CLASS, true);
```
tscni, line 30: In theory this isn't covered by the backward compatibility promise, right?
Even so, the removal of this would cause some breakages (see e.g. https://github.com/mglaman/phpstan-drupal/pull/195)

tscni (commented):
This should work, but I'm slightly concerned regarding backward compatibility of node attributes.

src/Parser/AnonymousClassVisitor.php
ondrejmirtes: Let's keep the attribute but don't use it ourselves anywhere anymore.

ondrejmirtes (changes requested):

src/Parser/AnonymousClassVisitor.php
tscni: Done. I opted against keeping the constant since it's not meant to be reused.

ondrejmirtes:
Thank you! Btw I wouldn't bother PHP-Parser with this, it's a niche edge case we've already solved, no need to change the underlying library.

PR #3344, opened 2024-08-23
Add `Type::reverseArray()`

herndlm:
Improves return type specification of lists and various edge cases with `array_reverse()`.

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -837,6 +837,23 @@ class ConstantArrayType extends ArrayType implements ConstantType
 		return $this->removeLastElements(1);
 	}
 
+	private function reverseConstantArray(TrinaryLogic $preserveKeys): self
+	{
+		$keyTypesReversed = array_reverse($this->keyTypes, true);
+		$keyTypes = array_values($keyTypesReversed);
+		$keyTypesReversedKeys = array_keys($keyTypesReversed);
+		$optionalKeys = array_map(static fn (int $optionalKey): int => $keyTypesReversedKeys[$optionalKey], $this->optionalKeys);
+
+		$reversed = new self($keyTypes, array_reverse($this->valueTypes), $this->nextAutoIndexes, $optionalKeys, TrinaryLogic::createNo());
+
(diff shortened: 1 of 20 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -1118,12 +1130,21 @@ public function slice(int $offset, ?int $limit, bool $preserveKeys = false): sel
...
-			return $this->reverse(true)
-				->slice($reversedOffset, $reversedLimit, $preserveKeys)
-				->reverse(true);
+			$sliced = $reversed->slice($reversedOffset, $reversedLimit, $preserveKeys)
+				->reverseArray(TrinaryLogic::createYes());
+			if (!$sliced instanceof self) {
+				throw new ShouldNotHappenException();
+			}
+
+			return $sliced;
```
herndlm: this is a bit ugly, but didn't find a good solution for the type problems yet..

we'd need a private `reverseConstantArray()` helper to have `self` as return type most likely.

src/Type/Constant/ConstantArrayType.php
```diff
@@ -1162,16 +1179,10 @@ public function slice(int $offset, ?int $limit, bool $preserveKeys = false): sel
 		return $preserveKeys ? $slice : $slice->reindex();
 	}
 
+	/** @deprecated Use reverseArray() instead */
```
staabm, line 1182: Does it make sense to deprecate the method only for the ConstantArrayType?

herndlm: Sorry, I didn't fully get that. The method only exists there and nowhere else.

staabm: Ohh I see ;-)

herndlm:
this one is simple (I think) and should be ready :) sorry for the impatience, I want to work on a follow-up that should improve array_reverse for constant arrays :)

ondrejmirtes:
Love it, thanks!

herndlm:
thank you! felt good finally contributing something again, even if it's a simple thing like that :)

PR #3345, opened 2024-08-23
Move result cache output from debug to very verbose mode

janedbal:
Since there are more and more `DiagnoseExtensions` with spammy output being added, we are switching from `-vvv` to `-vv` mode. But I think result cache info is still pretty important even when debug mode is not enabled, so I changed it.

### DIFF
--- a/src/Analyser/ResultCache/ResultCacheManager.php
+++ b/src/Analyser/ResultCache/ResultCacheManager.php
@@ -287,7 +287,7 @@ final class ResultCacheManager
 		$filesToAnalyse = array_unique($filesToAnalyse);
 		$filesToAnalyseCount = count($filesToAnalyse);
 
-		if ($output->isVerbose()) {
+		if ($output->isVeryVerbose()) {
 			$elapsed = microtime(true) - $startTime;
 			$elapsedString = $elapsed > 5
 				? sprintf(' in %f seconds', round($elapsed, 1))
--- a/src/Analyser/ResultCache/ResultCacheManager.php
+++ b/src/Analyser/ResultCache/ResultCacheManager.php
@@ -412,20 +412,20 @@ final class ResultCacheManager
 		}
 		$doSave = function (array $errorsByFile, $locallyIgnoredErrorsByFile, $linesToIgnore, $unmatchedLineIgnores, $collectedDataByFile, ?array $dependencies, array $exportedNodes, array $projectExtensionFiles) use ($internalErrors, $resultCache, $output, $onlyFiles, $meta): bool {
 			if ($onlyFiles) {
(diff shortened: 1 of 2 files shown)


janedbal:
Any remarks? We switched to `-vv`, but this important info is missing now...

src/Analyser/ResultCache/ResultCacheManager.php
```diff
@@ -92,21 +92,21 @@ public function restore(array $allAnalysedFiles, bool $debug, bool $onlyFiles, ?
 	{
 		$startTime = microtime(true);
 		if ($debug) {
-			if ($output->isDebug()) {
+			if ($output->isVerbose()) {
```
ondrejmirtes: I'd switch to `isVeryVerbose` (it's part of the interface already)

ondrejmirtes (changes requested):

src/Analyser/ResultCache/ResultCacheManager.php
janedbal: Adjusted

src/Analyser/ResultCache/ResultCacheManager.php
```diff
@@ -159,7 +159,7 @@ public function restore(array $allAnalysedFiles, bool $debug, bool $onlyFiles, ?
 				continue;
 			}
 			if (!is_file($extensionFile)) {
-				if ($output->isDebug()) {
+				if ($output->isVerbose()) {
```
ondrejmirtes: Should be `isVeryVerbose()` in places where you still left `isVerbose()`.

ondrejmirtes (changes requested):

src/Analyser/ResultCache/ResultCacheManager.php
janedbal: Missed this one, thx

janedbal: Fixed. Double checked, last one in verbose mode is the "Result cache restored", would you keep it?

src/Analyser/ResultCache/ResultCacheManager.php
```diff
@@ -287,7 +287,7 @@ public function restore(array $allAnalysedFiles, bool $debug, bool $onlyFiles, ?
 		$filesToAnalyse = array_unique($filesToAnalyse);
 		$filesToAnalyseCount = count($filesToAnalyse);
 
-		if ($output->isDebug()) {
+		if ($output->isVerbose()) {
```
ondrejmirtes:
```suggestion
		if ($output->isVeryVerbose()) {
```

ondrejmirtes:
Thank you.

PR #3346, opened 2024-08-23
Improve curl_init() return type analysis

tscni:
I got annoyed by the failing e2e/php8 test (https://github.com/phpstan/phpstan/blob/1.12.x/e2e/php8/test.php#L11, https://phpstan.org/r/32a5c25f-c266-4c04-9bf1-2ec0562397d8), so I wanted to resolve it by restoring the expected issue there. (But only now do I notice that this doesn't help that at all - I completely misunderstood the issue there :P)

Following, https://github.com/phpstan/phpstan/issues/1274, https://github.com/phpstan/phpstan/issues/1465, https://github.com/phpstan/phpstan/pull/1484 I looked into `curl_init()` a bit more.

Here's my analysis: (with verified behavior across all latest PHP minor versions and libcurl 7.10.5 + 8.9.1)
One important point to start with, if we presume that `curl_init()` without arguments never returns `false`, we generally ignore initialization errors or the case where curl fails to allocate more memory.
- `curl_init()` with a string argument will only return false if it fails to set `CURLOPT_URL` with it
  https://github.com/php/php-src/blob/php-8.4.0beta3/ext/curl/interface.c#L1154-L1159
- On PHP >= 8.0 Setting `CURLOPT_URL` will fail with a value error on the PHP side if the URL contains a null byte
  https://github.com/php/php-src/blob/php-8.4.0beta3/ext/curl/interface.c#L139
  https://github.com/php/php-src/blob/php-8.0.30/ext/curl/interface.c#L104-L107
- On PHP < 8.0 Setting `CURLOPT_URL` will fail with a warning (`false` return) on the PHP side if the URL contains a null byte
  https://github.com/php/php-src/blob/php-7.4.33/ext/curl/interface.c#L112-L115
  But, if open_basedir is enabled, it will also fail when the URL is invalid or uses the `file` scheme
  https://github.com/php/php-src/blob/php-7.4.33/ext/curl/interface.c#L146-L155
- On the side of libcurl, setting the URL option will only fail if the input exceeds a relatively high limit, but only since version 7.65.0 (https://github.com/curl/curl/commit/5fc28510a4664f46459d9a40187d81cc08571e60)
  https://github.com/curl/curl/blob/curl-8_9_1/lib/setopt.c#L3256
  https://github.com/curl/curl/blob/curl-8_9_1/lib/setopt.c#L1470
  https://github.com/curl/curl/blob/curl-8_9_1/lib/setopt.c#L70-L71

### DIFF
--- a/src/Type/Php/CurlInitReturnTypeExtension.php
+++ b/src/Type/Php/CurlInitReturnTypeExtension.php
@@ -11,8 +11,12 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\NeverType;
+use PHPStan\Type\NullType;
+use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\UnionType;
+use function array_map;
 use function count;
 use function is_string;
 use function parse_url;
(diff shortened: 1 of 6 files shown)


src/Type/Php/CurlInitReturnTypeExtension.php
```diff
@@ -26,10 +41,49 @@ public function getTypeFromFunctionCall(
...
+		$urlArgType = $scope->getType($args[0]->value);
+		if ($urlArgType->isNull()->yes()) {
+			return $notFalseReturnType;
+		}
+		if ($urlArgType->isString()->yes()) {
+			$urlArgValue = $urlArgType->getConstantScalarValues()[0] ?? null;
+			if ($urlArgValue !== null) {
+				if (!is_string($urlArgValue)) {
+					throw new ShouldNotHappenException();
+				}
```
tscni: This feels a bit cumbersome to do, but I couldn't find a simpler way to get the constant value of a string.
Maybe a `Type::getConstantStringValue(): ?string` would be useful. But then that might lead to `Type` exploding with helper functions eventually.

staabm: Does `Type::getConstantStrings` work for you?
(And you then check all strings, not just the first)

Testcase:

```
$url = 'example.org';
If (rand(0,1)) $url = 'example.com';

curl_init($url);
```

tscni: Ah I missed that method. But for multiple constant values it turned out to not be helpful. I overlooked that possibility completely.
Thanks a lot!

src/Type/Php/CurlInitReturnTypeExtension.php
```diff
@@ -26,13 +45,58 @@ public function getTypeFromFunctionCall(
...
+		}
+		if (str_contains($urlArgValue, "\0")) {
+			if (!$this->phpVersion->throwsValueErrorForInternalFunctions()) {
+				// https://github.com/php/php-src/blob/php-7.4.33/ext/curl/interface.c#L112-L115
+				return new ConstantBooleanType(false);
+			}
+			// https://github.com/php/php-src/blob/php-8.0.0/ext/curl/interface.c#L104-L107
+			return new NeverType();
+		}
+		if ($this->phpVersion->getVersionId() < 80000) {
```
staabm: Should be a new PHPVersion method

tscni: Done (`isCurloptUrlCheckingFileSchemeWithOpenBasedir()`)

ondrejmirtes:
Hi, thank you very much! This is way too much research effort for such a simple function, but I'm grateful for it. Typically I'd solve this with a benevolent union type (decreases number of false positives, does not increase the number of reported errors). (More info here: https://phpstan.org/config-reference#checkbenevolentuniontypes)

PR #3347, opened 2024-08-25
apply parameter closure type extension in FunctionCallParametersCheck

schlndh:
Fixes https://phpstan.org/r/8ff9bc8f-4162-4ccf-bf5c-8b3d2061e68d
Partially fixes: https://github.com/phpstan/phpstan/issues/10396
Related to https://github.com/phpstan/phpstan-src/pull/3282#issuecomment-2306883685

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -7,6 +7,8 @@ use PhpParser\Node\Expr;
 use PHPStan\Analyser\MutatingScope;
 use PHPStan\Analyser\Scope;
 use PHPStan\Php\PhpVersion;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParameterReflection;
 use PHPStan\Reflection\ParameterReflectionWithPhpDocs;
 use PHPStan\Reflection\ParametersAcceptor;
@@ -20,6 +22,7 @@ use PHPStan\Type\ErrorType;
 use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\NeverType;
+use PHPStan\Type\ParameterClosureTypeHelper;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeTraverser;
@@ -42,6 +45,7 @@ final class FunctionCallParametersCheck
 		private PhpVersion $phpVersion,
 		private UnresolvableTypeHelper $unresolvableTypeHelper,
 		private PropertyReflectionFinder $propertyReflectionFinder,
+		private ParameterClosureTypeHelper $parameterClosureTypeHelper,
 		private bool $checkArgumentTypes,
 		private bool $checkArgumentsPassedByReference,
 		private bool $checkExtraArguments,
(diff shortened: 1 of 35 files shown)


ondrejmirtes:
I thought about this. I thought I wanted a different solution using virtual nodes but it'd need a lot of refactoring. And maybe it'd still be very hard.

For similar reasons I already added $scope->pushInFunctionCall() in the same Check class.

So feel free to pursue the same direction you started 😊

src/Rules/FunctionCallParametersCheck.php
```diff
@@ -52,12 +56,14 @@ public function __construct(
...
 
 	/**
+	 * @param MethodReflection|FunctionReflection|null $callReflection
 	 * @param Node\Expr\FuncCall|Node\Expr\MethodCall|Node\Expr\StaticCall|Node\Expr\New_ $funcCall
 	 * @param array{0: string, 1: string, 2: string, 3: string, 4: string, 5: string, 6: string, 7: string, 8: string, 9: string, 10: string, 11: string, 12: string, 13?: string} $messages
 	 * @param 'attribute'|'callable'|'method'|'staticMethod'|'function'|'new' $nodeType
 	 * @return list<IdentifierRuleError>
 	 */
 	public function check(
+		$callReflection,
```
schlndh: This breaks larastan: 

```
TypeError: PHPStan\Rules\FunctionCallParametersCheck::check(): Argument #2 ($parametersAcceptor) must be of type PHPStan\Reflection\ParametersAcceptor, PHPStan\Analyser\MutatingScope given, called in /home/runner/work/phpstan-src/phpstan-src/e2e/integration/repo/src/Rules/CheckDispatchArgumentTypesCompatibleWithClassConstructorRule.php on line 133
```

ondrejmirtes: Yeah please add the parameters here as last and optional

tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php
```diff
@@ -34,9 +47,69 @@ protected function getRule(): Rule
...
+
+						public function getMethodParameterClosureTypeExtensions(): array
+						{
+							return [
+								new class implements MethodParameterClosureTypeExtension
+								{
+
+									public function isMethodSupported(MethodReflection $methodReflection, ParameterReflection $parameter): bool
+									{
+										return $methodReflection->getName() === 'foo' && $methodReflection->getDeclaringClass()->getName() === 'MethodParameterClosureTypeExtension\\Foo';
```
schlndh, line 76: When I run `make phpstan` locally it complains about `$methodReflection->getDeclaringClass()->getName() === 'MethodParameterClosureTypeExtension\\Foo'` being always false. Perhaps PHPStan is overly strict in this case? The code is fine IMO. It's especially annoying that it doesn't happen in Github Actions, so I can't even ignore it by adding it to the baseline.

src/Rules/FunctionCallParametersCheck.php
schlndh: I was just going to ignore it, since it's not `@api`. But OK. 😄

tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php
ondrejmirtes: Try running `composer dump` first and then `make phpstan` again. What happens?

schlndh: I haven't tried this exactly, but I rebased it on top of 1.12.x, ran `composer install` and now it doesn't complain anymore. So you were probably right.

ondrejmirtes:
I've looked at this and here are my thoughts: I don't like that `FunctionCallParametersCheck::check()` gains full reflection as another parameter. This isn't how I would design this. There are enough things already passed **from reflection** that this change really feels like it's bolted on to achieve a single goal.

Since the goal is to basically change `$parameter` from already passed `$parametersAcceptor`, I feel like we should modify `$parametersAcceptor` before it's passed to the `check()` method.

So the rules that call `FunctionCallParametersCheck` should modify `ParametersAcceptor` before it's passed there.

ondrejmirtes:
Alright, I see this is just an alternative to https://github.com/phpstan/phpstan-src/pull/3282. I will merge that instead. Thank you.

PR #3349, opened 2024-08-26
RoundFunctionReturnTypeExtension supports types with ConstantScalar

ShioPy0101:
Returns a constant when `$num` parameter and option of the round function are constants.

In this case , the conventional type is returned.
- `$num` parameter is a numeric string.
- `$num` parameter union contains a numeric string.
- `round()` option is not a constant.

### DIFF
--- a/src/Type/Php/RoundFunctionReturnTypeExtension.php
+++ b/src/Type/Php/RoundFunctionReturnTypeExtension.php
@@ -3,6 +3,7 @@
 namespace PHPStan\Type\Php;
 
 use Closure;
+use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Php\PhpVersion;
@@ -66,58 +67,44 @@ final class RoundFunctionReturnTypeExtension implements DynamicFunctionReturnTyp
 
 		$args = $functionCall->getArgs();
 
-		//引数長さ0ならNeverType/NullTypeを返す
 		if (count($args) < 1) {
 			return $noArgsReturnType;
 		}
 
 		$argType = $scope->getType($args[0]->value);
-
 		$functionName = $functionReflection->getName();
-
 		$proc = $this->getProc($functionName, $args, $scope);
 
 		if ($proc !== null) {
 			$constantScalarValues = $argType->getConstantScalarValues();
-			$rv = [];
+			$returnValueTypes = [];
 
 			foreach ($constantScalarValues as $constantScalarValue) {
-
 				if (!is_int($constantScalarValue) && !is_float($constantScalarValue)) {
-					$rv = [];
+					$returnValueTypes = [];
 					break;
 				}
 
-				$value = $proc($constantScalarValue);
-
-				$rv[] = new ConstantFloatType($value);
+				$returnValueTypes[] = new ConstantFloatType($proc($constantScalarValue));
 			}
 
-			if (count($rv) >= 1) {
-
-				$rvUnion = TypeCombinator::union(...array_map(static fn ($l) => $l, $rv));
-				return $rvUnion;
+			if (count($returnValueTypes) >= 1) {
+				return TypeCombinator::union(...array_map(static fn ($l) => $l, $returnValueTypes));
 			}
 		}
 
-		//最初の引数のTypeを取得
 		$firstArgType = $scope->getType($functionCall->getArgs()[0]->value);
 
-		//$firstArgType が MixedTypeなら $defaultReturnTypeを返す
 		if ($firstArgType instanceof MixedType) {
 			return $defaultReturnType;
 		}
 
-		//この条件分岐はバージョン情報
 		if ($this->phpVersion->hasStricterRoundFunctions()) {
-
-			//PHP言語仕様として、引数として指定されてるIntegerTypeとFloatTypeを指定
 			$allowed = TypeCombinator::union(
 				new IntegerType(),
 				new FloatType(),
 			);
 
-			//厳密な型を宣言しないならとういう条件分岐
 			if (!$scope->isDeclareStrictTypes()) {
 				$allowed = TypeCombinator::union(
 					$allowed,
@@ -130,7 +117,6 @@ final class RoundFunctionReturnTypeExtension implements DynamicFunctionReturnTyp
 				);
 			}
 
-			//スーパータイプではないなら、NeverTypeを返す
 			if ($allowed->isSuperTypeOf($firstArgType)->no()) {
 				// PHP 8 fatals if the parameter is not an integer or float.
 				return new NeverType(true);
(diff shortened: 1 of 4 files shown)


staabm:
which problem does this change solve? whats your real world use-case?

src/Type/Php/RoundFunctionReturnTypeExtension.php
```diff
@@ -96,4 +116,67 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			if (!is_int($constantScalarValue) && !is_float($constantScalarValue)) {
+				$returnValueTypes = [];
+				break;
+			}
+
+			$returnValueTypes[] = new ConstantFloatType($proc($constantScalarValue));
+		}
+
+		if (count($returnValueTypes) >= 1) {
+			return TypeCombinator::union(...array_map(static fn ($l) => $l, $returnValueTypes));
```
zonuexe:
```suggestion
			return TypeCombinator::union(...$returnValueTypes);
```

zonuexe:
@staabm 

I'm sure there are several use cases for this, but one I quickly spotted in my private code is the following pattern, which uses `$last_page` for some logic other than just printing it.

Ideally it would be even more useful if a float range type was implemented, but I feel like this is worthwhile enough on its own.

```php
const MAX = 1000;
const PER_PAGE = 30;
$last_page = (int)floor(MAX / PER_PAGE);
```

https://phpstan.org/r/b583dc30-3b4e-4b1f-af60-713159f43309

VincentLanglet:
Hi @ShioPy0101, are you still interested by this PR ? Could you rebase and target 2.3.x branch ?

ShioPy0101:
Hi @VincentLanglet, I’ve rebased the PR onto 2.3.x and updated the target branch. Could you take another look when you have a chance?

Thank you for bringing this PR back to my attention.

src/Type/Php/RoundFunctionReturnTypeExtension.php
```diff
@@ -99,4 +118,68 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			return null;
+		}
+
+		$constantScalarValues = $argType->getConstantScalarValues();
+		$returnValueTypes = [];
+
+		foreach ($constantScalarValues as $constantScalarValue) {
+			if (!is_int($constantScalarValue) && !is_float($constantScalarValue)) {
+				$returnValueTypes = [];
+				break;
```
VincentLanglet: You could return null directly no ?

src/Type/Php/RoundFunctionReturnTypeExtension.php
```diff
@@ -99,4 +118,68 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+
+		if ($proc === null) {
+			return null;
+		}
+
+		$constantScalarValues = $argType->getConstantScalarValues();
+		$returnValueTypes = [];
+
+		foreach ($constantScalarValues as $constantScalarValue) {
+			if (!is_int($constantScalarValue) && !is_float($constantScalarValue)) {
```
VincentLanglet: Why this check ?

I saw in test you have the assertion
```
assertType('float', round($IntOrNumStr));
```
but can't we also understand `'2.5'` ?

src/Type/Php/RoundFunctionReturnTypeExtension.php
```diff
@@ -99,4 +118,68 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+				}
+
+				if (!isset($args[2]->value)) {
+					$proc = static fn ($name) => round($name, $precision);
+				} else {
+					$modeArg = $args[2]->value;
+					$modeType = $scope->getType($modeArg);
+					$mode = $modeType->getConstantScalarValues();
+
+					if (count($mode) === 1 && in_array($mode[0], [PHP_ROUND_HALF_UP, PHP_ROUND_HALF_DOWN, PHP_ROUND_HALF_EVEN, PHP_ROUND_HALF_ODD], true)) {
```
VincentLanglet: There is now new RoundingMode
https://www.php.net/manual/en/enum.roundingmode.php

src/Type/Php/RoundFunctionReturnTypeExtension.php
ShioPy0101: Yes, you're right. I'll simplify this to return null directly.

src/Type/Php/RoundFunctionReturnTypeExtension.php
ShioPy0101: Thanks for pointing it out — I’ll update the implementation for the current PHP versions.

src/Type/Php/RoundFunctionReturnTypeExtension.php
ShioPy0101: Thank you for pointing this out. I agree that constant numeric strings can also be handled in non-strict mode. I’ll update the implementation and tests while preserving the strict-types behavior.

src/Type/Php/RoundFunctionReturnTypeExtension.php
```diff
@@ -99,4 +120,116 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			return $mode[0];
+		}
+
+		$enumCase = $modeType->getEnumCaseObject();
+
+		if ($enumCase === null || $enumCase->getClassName() !== 'RoundingMode') {
+			return null;
+		}
+
+		switch ($enumCase->getEnumCaseName()) {
```
VincentLanglet, line 149: Why this mapping ?

This loose some case.
```
assertType('float', round(9.5, 0, \RoundingMode::TowardsZero));
	assertType('float', round(9.5, 0, \RoundingMode::AwayFromZero));
	assertType('float', round(9.5, 0, \RoundingMode::NegativeInfinity));
	assertType('float', round(9.5, 0, \RoundingMode::PositiveInfinity));
```

You could keep the original value when the Enum exists no ?

VincentLanglet (commented):
Seems good otherwise

src/Type/Php/RoundFunctionReturnTypeExtension.php
ShioPy0101: I intentionally limited the mapping to the four RoundingMode cases that have direct equivalents in the legacy PHP_ROUND_HALF_* constants. The intent was to avoid making constant evaluation depend on the runtime availability of the RoundingMode enum.

For the remaining cases, I currently fall back to the existing return-type logic because they do not have a direct legacy constant mapping.

That said, I agree that this is an implementation limitation rather than a limitation of `RoundingMode` itself. Preserving the enum case internally would allow all cases to be evaluated more precisely, so I think that would be a better direction.

ShioPy0101: I am still considering the best approach here. The README states that PHPStan's source code is developed on PHP 8.2, while the distributed package and PHAR are transformed to run on PHP 7.2 and higher. Given that, I would prefer to avoid making constant evaluation depend on the runtime availability of RoundingMode.

VincentLanglet: Any suggestion @staabm ?

PR #3350, opened 2024-08-26
PHP 8.4 runtime support

ondrejmirtes:


thg2k:
What do you think of the implicitly nullable parameters deprecation in 8.4?

ondrejmirtes:
@thg2k I don't understand the question. Can you be more specific about what you're asking?

thg2k:
> @thg2k I don't understand the question. Can you be more specific about what you're asking?

I mean that those deprecation notices are a pain. I'm thinking about submitting an RFC to revert that

ondrejmirtes:
@thg2k It's too late for that. And it was accepted unanimously so it's unlikely to be reverted.

thg2k:
I know, but it's still an idiotic change, and reverting a deprecation should still be doable even in beta period, as it's not a new feature..

ondrejmirtes:
This is the wrong place to discuss that.

thg2k:
> This is the wrong place to discuss that.

I know, I was just asking your opinion en passant, as I value it.

PR #3361, opened 2024-08-27
Fix late static binding calls

mvorisek:
fix https://github.com/phpstan/phpstan/issues/10469

based on https://github.com/phpstan/phpstan/issues/10469#issuecomment-1906065516 analysis

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2083,6 +2083,18 @@ final class MutatingScope implements Scope
 				$typeCallback = function () use ($node): Type {
 					if ($node->class instanceof Name) {
 						$staticMethodCalledOnType = $this->resolveTypeByName($node->class);
+						if (
+							$staticMethodCalledOnType instanceof StaticType
+							&& !in_array($node->class->toLowerString(), ['self', 'static', 'parent'], true)
+						) {
+							$methodReflectionCandidate = $this->getMethodReflection(
+								$staticMethodCalledOnType,
+								$node->name->name,
+							);
+							if ($methodReflectionCandidate !== null && $methodReflectionCandidate->isStatic()) {
+								$staticMethodCalledOnType = $staticMethodCalledOnType->getStaticObjectType();
+							}
+						}
 					} else {
 						$staticMethodCalledOnType = $this->getNativeType($node->class);
 					}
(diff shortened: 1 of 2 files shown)


mvorisek:
@ondrejmirtes might I ask you to review this fix

src/Analyser/MutatingScope.php
```diff
@@ -5637,6 +5637,21 @@ private function methodCallReturnType(Type $typeWithMethod, string $methodName,
 		}
 
 		$methodReflection = $typeWithMethod->getMethod($methodName, $this);
+
+		// https://www.php.net/manual/en/language.oop5.late-static-bindings.php
+		if (
+			$methodCall instanceof Expr\StaticCall
+			&& $typeWithMethod instanceof StaticType
```
ondrejmirtes: This isn't a proper fix. The place where the StaticType transformation gets decided is in `ObjectType::getUnresolvedMethodPrototype()` and `StaticType::getUnresolvedMethodPrototype()`. One of them transforms `static` to the class it's called on, and the other one preserves `static`. Some piece of logic should be written there.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
mvorisek: But how - in `ObjectType::getUnresolvedMethodPrototype()` there is no info about how the call was made - ie. `$methodCall` and `$typeWithMethod` is not passed. Maybe you can look into the fix with the tests below?

ondrejmirtes: What kind of call do you want to transform and what kind of call do you not want to transform?

mvorisek: See condition a few lines below - `['self', 'static', 'parent']` - even if `self` resolves to the same class where declared, it does not behave the same due php "late static binding".

ondrejmirtes: I've asked for examples of method calls.

mvorisek: see the tests and/or https://github.com/phpstan/phpstan/issues/10469

ondrejmirtes: Yes, this comment explains it as I need it: https://github.com/phpstan/phpstan/issues/10469#issuecomment-1906046819

mvorisek: Thanks - now the question is how to fix it.

ondrejmirtes:
This works for me. Thank you.

PR #3363, opened 2024-08-27
Test chaining with `static`/`self` and `$this` more

mvorisek:
add more tests for #4902 for https://github.com/phpstan/phpstan/issues/5946

`self::` method calls and properties are resolved using late-static binding, but `Foo::` are not. For methods, this works perfectly already, for properties, expected type is commented in edge cases.

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/static-with-this-chained-bug-5946.php
+++ b/tests/PHPStan/Analyser/nsrt/static-with-this-chained-bug-5946.php
@@ -1,0 +1,108 @@
+<?php
+
+namespace StaticWithThisChained;
+
+use function PHPStan\Testing\assertType;
+
+class Foo
+{
+    /** @var $this */
+    public $propThis;
+
+    /** @var static */
+    public $propStatic;
+
+    /** @var static */
+    public static $propStaticStatic;
+
+    /**
+     * @return $this
+     */
+    public function getThis()
+    {
+        return $this;
+    }
+
+    /**
+     * @return static
+     */
+    public function newStatic()
+    {
+        return new static();
+    }
+
+    /**
+     * @return static
+     */
+    public static function newStaticStatic()
+    {
+        return new static();
+    }
+
+    public function testMethodChained(): void
+    {
+        assertType('$this(StaticWithThisChained\Foo)', $this->getThis());
+        assertType('static(StaticWithThisChained\Foo)', $this->newStatic());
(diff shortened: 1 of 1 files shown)


tests/PHPStan/Analyser/nsrt/static-with-this-chained.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function testMethodChained(): void
+    {
+        assertType('$this(StaticWithThisChained\Foo)', $this->getThis());
+        assertType('static(StaticWithThisChained\Foo)', $this->newStatic());
+        assertType('static(StaticWithThisChained\Foo)', static::newStaticStatic());
+        assertType('StaticWithThisChained\Foo', self::newStaticStatic());
+        assertType('StaticWithThisChained\Foo', Foo::newStaticStatic());
+
+        assertType('static(StaticWithThisChained\Foo)', $this->newStatic()->getThis());
```
mvorisek, line 50: Help needed - how can I mutate the scope to be static when `static` return type is analysed?

mvorisek: @ondrejmirtes please see https://phpstan.org/r/29fcfdb5-f19e-4df2-9177-61ad88490c64 repro - do you know where the issue comes from and how it can be fixed? (`static` must never be converted to `$this` )

mvorisek: @staabm might I ask you for a help? I have looked into the fix several times, but I am not familiar with mutating the scope logic and I do not know how to fix this myself.

mvorisek:
PR is passing, please review.

mvorisek:
@ondrejmirtes can this be reviewed please?

VincentLanglet (changes requested):
HI,

Rather than providing tests with "non-expected" result related to a solved and closed issues. I recommend either
- Providing a PR with extra fixes
- Or an issue with dedicated exemple about what is needed to be improved to let the bot working on it.

I currently don't see the value adding assertion with `// should be ` comments.

mvorisek:
Hi @VincentLanglet. This is comprehensive test to a fixed issue, but when landed, these tests were not added, so they have value to prevent breaking them later.

The test is covering all chaining scenarios and I prefer to keep also the cases which are still broken. I prefer to keep the broken cases to keep them logically together.

PR #3364, opened 2024-08-27
Fix infer new templated type from initial assign into static property

mvorisek:
fix https://github.com/phpstan/phpstan/issues/5551


mvorisek:
@ondrejmirtes can I please have your review here

ondrejmirtes:
Thank you.

PR #3365, opened 2024-08-27
Error on offset assignment to specialized strings

staabm:
closes https://github.com/phpstan/phpstan/issues/11572


ondrejmirtes:
Thank you!

PR #3366, opened 2024-08-27
Fix array_filter with callback optional persistance

robotomarvin:
Attempt to fix https://github.com/phpstan/phpstan/issues/11570


ondrejmirtes:
Perfect, thank you!

PR #3369, opened 2024-08-28
RegexArrayShapeMatcher - infer constant string types in alternations

staabm:
closes https://github.com/phpstan/phpstan/issues/11222


staabm:
phpstan-nette errors are expected. needs [expectation updates in the extension](https://github.com/phpstan/phpstan-nette/pull/161).

ondrejmirtes:
Thank you!

PR #3370, opened 2024-08-28
Detect function variadic-ness within top-level function_exists()

staabm:
closes https://github.com/phpstan/phpstan/issues/11559
closes https://github.com/phpstan/phpstan/issues/4753

### DIFF
--- a/src/Reflection/Php/PhpFunctionReflection.php
+++ b/src/Reflection/Php/PhpFunctionReflection.php
@@ -152,7 +152,7 @@ final class PhpFunctionReflection implements FunctionReflection
 				$cachedResult = $this->cache->load($key, $variableCacheKey);
 				if ($cachedResult === null) {
 					$nodes = $this->parser->parseFile($fileName);
-					$result = !$this->callsFuncGetArgs($nodes)->no();
+					$result = !$this->containsVariadicFunction($nodes)->no();
 					$this->cache->save($key, $variableCacheKey, $result);
 					return $result;
 				}
@@ -167,7 +167,7 @@ final class PhpFunctionReflection implements FunctionReflection
 	/**
 	 * @param Node[]|scalar[]|Node $node
 	 */
-	private function callsFuncGetArgs(array|Node $node): TrinaryLogic
+	private function containsVariadicFunction(array|Node $node): TrinaryLogic
 	{
 		$result = TrinaryLogic::createMaybe();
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
I feel like this code is too complicated. We could simply traverse through everything similarly to how we do here: https://github.com/phpstan/phpstan-src/blob/ad3286151a8edd63ccd7b51d2396f691ab0e0a5a/src/Type/FileTypeMapper.php#L758-L772

What's more important to also solve here is to answer the question: What if we find the same function multiple times? Is it enough if one of them is variadic, or all of them need to be variadic?

Also we might find multiple functions and some of them might be variadic natively (`...$params`) and some only through `func_get_args()`. The code needs to account for that.

And last but not least, do not forget to bump the version in the cache key here :)

staabm:
> I feel like this code is too complicated

from the pre-existing code I got the impression its on purpose filtering only the smallest possible AST, therefore I went with this ugly if-else analysis. adjusted it now to analyze the whole AST.

> Also we might find multiple functions and some of them might be variadic natively (`...$params`) and some only through `func_get_args()`. The code needs to account for that.

I added tests for the cases, I think it works as expected now.

> And last but not least, do not forget to bump the version in the cache key here :)

great catch

src/Reflection/Php/PhpFunctionReflection.php
```diff
@@ -167,41 +165,40 @@ private function isVariadic(): bool
...
-		foreach ($nodes as $node) {
+		$result = TrinaryLogic::createMaybe();
+
+		if ($node instanceof Node) {
 			if ($node instanceof Function_) {
 				$functionName = (string) $node->namespacedName;
 
 				if ($functionName === $this->reflection->getName()) {
-					return $this->functionCallStatementFinder->findFunctionCallInStatements(ParametersAcceptor::VARIADIC_FUNCTIONS, $node->getStmts()) !== null;
+					return TrinaryLogic::createFromBoolean($this->functionCallStatementFinder->findFunctionCallInStatements(ParametersAcceptor::VARIADIC_FUNCTIONS, $node->getStmts()) !== null);
```
ondrejmirtes: I think I'm missing looking at `Param::$variadic` here. Now it looks like it treats `function foo($bar) { }` and `function foo(...$bar) { }` the same and passes the function as variadic in either case. I think it should be more strict.

ondrejmirtes (changes requested):

src/Reflection/Php/PhpFunctionReflection.php
staabm: native variadic is already [checked in the outer scope](https://github.com/phpstan/phpstan-src/pull/3370/files#diff-3ac4e5d4a520618a490963b70e3fcd428d56aea61683641020393167fff2ca5dR140-R141).

for native variadic we don't call into `callsFuncGetArgs` at all

ondrejmirtes: We can have a situation when `$isNativelyVariadic = $this->reflection->isVariadic();` returns false and still one of the function declarations in the file might have `function foo(...$bar) `.

ondrejmirtes: Please reproduce this situation and fix it, thanks.

staabm: I was not able to reproduce it initially because of the caching involved. 

After temporarily using different cache keys locally the case was showing up :-)

ondrejmirtes (approved):

ondrejmirtes:
Alright, this works for me :) Thank you!

PR #3372, opened 2024-08-30
ConstantArrayType: fix returned ConstantArrayTypeAndMethod

janedbal:
With this, we can recognize those cases:

```php

class Parnt
{
	public function __construct()
	{
		array_map([self::class, 'method1'], [1]); // calls Parnt::method1
		array_map([$this, 'method1'], [1]); // calls Child::method1
	}

	public function method1(): void {
	    echo 'parent';
	}
	
}

class Child extends Parnt
{
	public function __construct()
	{
		parent::__construct();
	}

	public function method1(): void {
	    echo 'child';
	}
	
}

new Child();
```

### DIFF
--- a/tests/PHPStan/Type/Constant/ConstantArrayTypeTest.php
+++ b/tests/PHPStan/Type/Constant/ConstantArrayTypeTest.php
@@ -1051,4 +1051,34 @@ class ConstantArrayTypeTest extends PHPStanTestCase
 		$this->assertSame($expectedType->getNextAutoIndexes(), $actualType->getNextAutoIndexes());
 	}
 
+	public function testFindTypeAndMethodNames(): void
+	{
+		$classStringArray = new ConstantArrayType([
+			new ConstantIntegerType(0),
+			new ConstantIntegerType(1),
+		], [
+			new ConstantStringType(Closure::class, true),
+			new ConstantStringType('bind'),
+		]);
+		$objectArray = new ConstantArrayType([
+			new ConstantIntegerType(0),
+			new ConstantIntegerType(1),
+		], [
+			new ObjectType(Closure::class, null, $this->createReflectionProvider()->getClass(Closure::class)),
+			new ConstantStringType('bind'),
+		]);
+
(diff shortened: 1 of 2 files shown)


staabm:
Is missing a test

janedbal:
I'm not sure if there is any native functionality to test against. I need this function for [proper dead-code analysis](https://github.com/shipmonk-rnd/dead-code-detector/pull/48/files#diff-98f320f8c7e46825ecec63dc14e12ad45264cb78c54b41c75b15fe30a0b3d6cfR185).

ondrejmirtes:
If nothing else is possible then we need a unit test in ConstantArrayTypeTest. Make sure to initialize `$this->createReflectionProvider()` so that the Type works properly.

tests/PHPStan/Type/Constant/ConstantArrayTypeTest.php
```diff
@@ -1051,4 +1051,34 @@ public function testValuesArray(ConstantArrayType $type, ConstantArrayType $expe
...
+			new ObjectType(Closure::class, null, $this->createReflectionProvider()->getClass(Closure::class)),
+			new ConstantStringType('bind'),
+		]);
+
+		$classStringResult = $classStringArray->findTypeAndMethodNames();
+		$objectResult = $objectArray->findTypeAndMethodNames();
+
+		$this->assertCount(1, $classStringResult);
+		$this->assertCount(1, $objectResult);
+		$this->assertInstanceOf(ConstantStringType::class, $classStringResult[0]->getType());
```
janedbal, line 1076: Without the fix, this is still just `ObjectType` here.

janedbal:
Added a test and rebased on top of 2.0.x

ondrejmirtes:
Thank you.

ondrejmirtes:
I'm sorry, had to revert this. There are regressions (could be seen in the issue bot):

* https://github.com/phpstan/phpstan/issues/11859
* https://github.com/phpstan/phpstan/issues/8071

PR #3373, opened 2024-08-30
Narrow array on count(), string on strlen() with positive-int

staabm:
closes https://github.com/phpstan/phpstan/issues/11558

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1049,7 +1049,11 @@ final class TypeSpecifier
 				}
 
 			}
-			return $valueTypesBuilder->getArray();
+
+			$arrayType = $valueTypesBuilder->getArray();
+			if ($arrayType->isIterableAtLeastOnce()->yes()) {
+				return $arrayType;
+			}
 		}
 
 		return null;
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -2093,15 +2093,25 @@ final class TypeSpecifier
 				return $this->create($unwrappedLeftExpr->getArgs()[0]->value, new NeverType(), $context, false, $scope, $rootExpr);
 			}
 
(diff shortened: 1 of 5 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -1096,87 +1101,6 @@ private function specifyTypesForConstantBinaryExpression(
 			));
 		}
 
-		if (
```
staabm, line 1099: moved this 2 if-blocks out of `specifyTypesForConstantBinaryExpression` as we no longer check for a constant type operand, but do a more general `isInteger` based handling, which includes unbound `IntegerRangeType`.

src/Analyser/TypeSpecifier.php
```diff
@@ -1037,7 +1038,7 @@ private function turnListIntoConstantArray(FuncCall $countFuncCall, Type $type,
 					$offsetType = new ConstantIntegerType($i);
 					$valueTypesBuilder->setOffsetValueType($offsetType, $type->getOffsetValueType($offsetType), true);
 				}
-			} else {
+			} elseif ($type->isConstantArray()->yes()) {
```
staabm, line 1041: compensated the now passed in maybe unbound `IntegerRangeType`s

tests/PHPStan/Analyser/nsrt/count-type.php
```diff
@@ -44,12 +44,12 @@ public function doFooBar(
 		if (count($arr) == $maybeZero) {
 			assertType('array', $arr);
 		} else {
-			assertType('non-empty-array', $arr);
+			assertType('array', $arr);
```
staabm, line 47: I think I did a mistake here in previous PR, as the array can still be empty in the else branch

https://3v4l.org/409Jp

ondrejmirtes (changes requested):
Too many changes. I can't read that diff. Please send this in smaller chunks.

PR #3374, opened 2024-08-30
Fix `get_debug_type` produces wrong type for anonymous classes with parent

patrickkusebauch:
Closes: https://github.com/phpstan/phpstan/issues/11562

Fix poorly documented case of `get_debug_type` where it provides more specific output in cases where the anonymous class extends another class or implements an interface.


patrickkusebauch:
None of the failing integration/extension tests seems to be caused by this PR by my estimation.

ondrejmirtes:
Thanks a lot!

PR #3375, opened 2024-08-30
RegexArrayShapeMatcher - improve type inference in alternations

staabm:
builds a more precise type for a capturing group, in case the capturing group only contains a alternation at the top level


ondrejmirtes:
Thank you!

PR #3377, opened 2024-08-31
Update PhpStorm stubs

VincentLanglet:
Hi @ondrejmirtes, 

I tried to work on the update https://github.com/phpstan/phpstan-src/pull/3334

First I thought some DynamicMethodThrowTypeExtension was needed but it seems like it is correctly implemented since your PR https://github.com/phpstan/phpstan-src/commit/aa7989f5839ce3196cd5b874284f6efe49c62e24

But the issue is the fact that currently CallToMethodStatementWithoutSideEffectsRule
Does not seems to use those dynamicExtension but rely on `$method->getThrowType()` instead 
which seems to only rely on the phpdoc of the method.

An example can be found with https://phpstan.org/r/b071a436-7779-47de-816e-d520f4d70bdd, even if the dynamicThrowTypeExtension exists (and say there is no exception with 2 params https://github.com/phpstan/phpstan-src/blob/1.12.x/src/Type/Php/DsMapDynamicMethodThrowTypeExtension.php).

I see 3 solutions:
- This is done by design, so I ignore the "error" like in this PR
- There is an available method to use or something easy to implement instead of `$method->getThrowType()`, and I'll be happy to update the PR with it. (but which one ?)
- It's more complex and we ignore the error until then

I'll be happy to try working on this, but so far I have trouble to see if I need to play with the ThrowPoint things or just duplicate the use of DynamicMethodThrowTypeExtension....

### DIFF
--- a/src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
+++ b/src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
@@ -10,7 +10,6 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\Type\ErrorType;
-use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
 use function sprintf;
 
--- a/src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
+++ b/src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
@@ -5,6 +5,7 @@ namespace PHPStan\Rules\Methods;
 use PhpParser\Node;
 use PHPStan\Analyser\NullsafeOperatorHelper;
 use PHPStan\Analyser\Scope;
+use PHPStan\Node\NoopExpressionNode;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
@@ -14,7 +15,7 @@ use PHPStan\Type\Type;
 use function sprintf;
 
 /**
- * @implements Rule<Node\Stmt\Expression>
+ * @implements Rule<NoopExpressionNode>
  */
 final class CallToMethodStatementWithoutSideEffectsRule implements Rule
 {
@@ -25,18 +26,18 @@ final class CallToMethodStatementWithoutSideEffectsRule implements Rule
 
 	public function getNodeType(): string
 	{
-		return Node\Stmt\Expression::class;
+		return NoopExpressionNode::class;
 	}
 
 	public function processNode(Node $node, Scope $scope): array
 	{
-		if ($node->expr instanceof Node\Expr\NullsafeMethodCall) {
-			$scope = $scope->filterByTruthyValue(new Node\Expr\BinaryOp\NotIdentical($node->expr->var, new Node\Expr\ConstFetch(new Node\Name('null'))));
-		} elseif (!$node->expr instanceof Node\Expr\MethodCall) {
+		$methodCall = $node->getOriginalExpr();
+		if ($methodCall instanceof Node\Expr\NullsafeMethodCall) {
+			$scope = $scope->filterByTruthyValue(new Node\Expr\BinaryOp\NotIdentical($methodCall->var, new Node\Expr\ConstFetch(new Node\Name('null'))));
+		} elseif (!$methodCall instanceof Node\Expr\MethodCall) {
 			return [];
 		}
 
-		$methodCall = $node->expr;
 		if (!$methodCall->name instanceof Node\Identifier) {
 			return [];
 		}
@@ -61,8 +62,8 @@ final class CallToMethodStatementWithoutSideEffectsRule implements Rule
 		}
 
 		$method = $calledOnType->getMethod($methodName, $scope);
-		if ($method->hasSideEffects()->no() || $node->expr->isFirstClassCallable()) {
-			if (!$node->expr->isFirstClassCallable()) {
+		if ($method->hasSideEffects()->no() || $methodCall->isFirstClassCallable()) {
+			if (!$methodCall->isFirstClassCallable()) {
 				$throwsType = $method->getThrowType();
 				if ($throwsType !== null && !$throwsType->isVoid()->yes()) {
 					return [];
--- a/src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
+++ b/src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
@@ -10,6 +10,7 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
 use function sprintf;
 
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
The right solution is to take dynamic throw type extensions into account in `*WithoutSideEffectsRule` rules.

But we need to introduce a new virtual node that would wrap `Node\Stmt\Expression` node and also provide things the StatementResult provides. We could then refactor these rules to use this virtual node.

ondrejmirtes:
Oh, we already have this virtual node: NoopExpressionNode

Try rewriting the rules by using it. The main rule that uses it is BetterNoopRule and it actually contains skips for situations that are handled by `*WithoutSideEffectsRule` rules. So it should be possible to rewrite `*WithoutSideEffectsRule` rules to use NoopExpressionNode too.

VincentLanglet:
> Oh, we already have this virtual node: NoopExpressionNode
> 
> Try rewriting the rules by using it. The main rule that uses it is BetterNoopRule and it actually contains skips for situations that are handled by `*WithoutSideEffectsRule` rules. So it should be possible to rewrite `*WithoutSideEffectsRule` rules to use NoopExpressionNode too.

Am I wrong or currently NoopExpressionNode doesn't have access to things the StatementResult provides ?

ondrejmirtes:
Doesn't matter, this node is invoked only for expressions with zero throw points.

src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
```diff
@@ -61,8 +62,8 @@ public function processNode(Node $node, Scope $scope): array
 		}
 
 		$method = $calledOnType->getMethod($methodName, $scope);
-		if ($method->hasSideEffects()->no() || $node->expr->isFirstClassCallable()) {
-			if (!$node->expr->isFirstClassCallable()) {
+		if ($method->hasSideEffects()->no() || $methodCall->isFirstClassCallable()) {
```
ondrejmirtes: You don't need to check `$method->hasSideEffects()`, that's already implied by NoopExpressionNode.

ondrejmirtes: And you don't need to check `instanceof NeverType` on line 74 either, that's also implied.

ondrejmirtes (changes requested):
Please also rewrite the 3 other rules as well (static call, function, constructor).

VincentLanglet:
> Please also rewrite the 3 other rules as well (static call, function, constructor).

Sure I will, but since tests are still failing, it means that PHPStan still believe there is throwPoints

src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
```diff
@@ -61,8 +62,8 @@ public function processNode(Node $node, Scope $scope): array
...
+			if (!$methodCall->isFirstClassCallable()) {
 				$throwsType = $method->getThrowType();
```
ondrejmirtes: Also remove this part about throw type.

ondrejmirtes (changes requested):

src/Rules/Methods/CallToMethodStatementWithoutSideEffectsRule.php
ondrejmirtes: IMHO this is why the test still fails.

VincentLanglet: Oh this might be the reason it was still failing. Thanks

VincentLanglet:
I had to add back the check on NeverType ([ff5c552](https://github.com/phpstan/phpstan-src/pull/3377/commits/ff5c55222919dfe21e00b708977b3ab6ce962eb1))  to solve two tests.

The last test was failing because PHPStan consider there is a ImpurePoint (methodCall: call to unknown method) in
```
$dt::createFromFormat('Y-m-d', '2019-07-24');
```

Which seems true, I can produce an exception with
```
class Foo extends \DateTimeImmutable {
	public static function createFromFormat($format, $time, ?DateTimeZone $object = NULL): DateTimeImmutable|false {
		throw new \Exception('Fail');
	}
}

$foo = new Foo();

var_dump($foo::createFromFormat('Y-m-d', '2019-07-24'));
```

VincentLanglet:
Thanks a lot for your help @ondrejmirtes, it's now ready for a new review

ondrejmirtes:
Awesome, thank you!

VincentLanglet:
https://github.com/phpstan/phpstan/issues/11503#issuecomment-2322888087 is showing there is a (small ?) impact of the current refacto.

A reproducer would be https://phpstan.org/r/37abe2c8-b4bb-4ba6-bb7d-0eedcf703d0c

There is an impure point `instantiation of class DateTimeZone`, which comes from 
https://github.com/phpstan/phpstan-src/blob/2e2757395b0c3bd9a845e265f62803008b751bbe/src/Analyser/NodeScopeResolver.php#L3270-L3281

There is a DateTimeZoneConstructorThrowTypeExtension https://github.com/phpstan/phpstan-src/blob/2e2757395b0c3bd9a845e265f62803008b751bbe/src/Type/Php/DateTimeZoneConstructorThrowTypeExtension.php#L17
it seems used for DeadCatch detection but not NoSideEffect detection, is there something missing or is this behavior wanted ?

Edit: I don't see `DateTime::__construct()`, `DateTimeImmutable::__construct()`, `DateTimeZone::__construct()`, and so on in the functionMetadata, I feel like it's related but dunno if it's done on purpose.

ondrejmirtes:
Does `new DateTimeImmutable('xyz')` have the same problem or not? If not, how is that achieved? We could do the same thing for new DateTimeZone.

If it has the same problem then we should fix both.

VincentLanglet:
> Does `new DateTimeImmutable('xyz')` have the same problem or not? 

Not sure to understand the question but I'll do my best to answer it:

```
new \DateTimeZone('UTC')
```
has 
- 1 impure point `instantiation of class DateTimeZone`
- 0 throw point

```
new \DateTimeImmutable()
```
has 
- 1 impure point `instantiation of class DateTimeImmutable`
- 0 throw point

```
new \DateTimeImmutable('xyz')
```
has 
- 1 impure point `instantiation of class DateTimeImmutable`
- 1 throw point => this is expected

Which gives https://phpstan.org/r/dfe3ff19-92e9-4d6b-be05-4673be64ffec

I expect 0 impure point everywhere

ondrejmirtes:
I say we're fine here, nothing to fix.

VincentLanglet:
> I say we're fine here, nothing to fix.

Would you mind develop a bit ?
Why a single line with
```
new \DateTimeImmutable();
```
without assigning the variable doesn't require to be reported ? (and is considering has having side effects)

I was planning on adding
```
'DateInterval::__construct' => ['hasSideEffects' => false],
'DatePeriod::__construct' => ['hasSideEffects' => false],
'DateTime::__construct' => ['hasSideEffects' => false],
'DateTimeImmutable::__construct' => ['hasSideEffects' => false],
'DateTimeZone::__construct' => ['hasSideEffects' => false],
```
in the functionMetadata.php file. (which seems to solve the issue)

ondrejmirtes:
`new \DateTimeImmutable();` is not pure, it depends on system clock and always creates a different time object.

ondrejmirtes:
So the purity depends on how the arguments look like. And we don't have an extension point for that now.

VincentLanglet:
Oh, I see.

I thought PHPStan had a level between purity and side effect, in a way that
- Pure imply no side effects
- But no side effects could be done in a impure function

Like https://phpstan.org/r/60731ffd-1e61-4cc0-bfdd-ec8529197b52

But seems like it's not.

So indeed, we're "fine" until such feature would be implemented.

PR #3378, opened 2024-08-31
Add DateTimeSubMethodThrowTypeExtension

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11503

### DIFF
--- a/src/Type/Php/DateTimeSubMethodThrowTypeExtension.php
+++ b/src/Type/Php/DateTimeSubMethodThrowTypeExtension.php
@@ -1,0 +1,43 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use DateTime;
+use DateTimeImmutable;
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Php\PhpVersion;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\DynamicMethodThrowTypeExtension;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+use function count;
+use function in_array;
+
+final class DateTimeSubMethodThrowTypeExtension implements DynamicMethodThrowTypeExtension
+{
+
+	public function __construct(private PhpVersion $phpVersion)
+	{
+	}
+
+	public function isMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === 'sub'
+			&& in_array($methodReflection->getDeclaringClass()->getName(), [DateTime::class, DateTimeImmutable::class], true);
+	}
+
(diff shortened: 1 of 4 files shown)


src/Type/Php/DateTimeSubMethodThrowTypeExtension.php
```diff
@@ -0,0 +1,43 @@
...
+	{
+		if (count($methodCall->getArgs()) === 0) {
+			return null;
+		}
+
+		if (!$this->phpVersion->hasDateTimeExceptions()) {
+			return null;
+		}
+
+		return new ObjectType('DateMalformedStringException');
```
ondrejmirtes: We need to check the constant strings in the argument type and only throw this exception if some of them are invalid. Or if there are zero constant strings.

ondrejmirtes (changes requested):
Also: what about `add` method?

src/Type/Php/DateTimeSubMethodThrowTypeExtension.php
VincentLanglet: `DateTime::sub` does not accepts constant strings, the signature is
```
public DateTime::sub(DateInterval $interval): DateTime
```

It only throw an exception if the DateInterval is relative like
```
$dt->sub(new DateInterval('next weekday'))
```
but I didnt see a way to have something like `getDateIntervalConstantStrings()`. I wasn't sure
- if it would require to introduce a DateIntervalType like the QueryType (https://github.com/phpstan/phpstan-doctrine/blob/38db3bad8f1567d7bf64806738d724261f8a2b5c/src/Type/Doctrine/Query/QueryType.php#L11) 
- if I was able to reproduce the implementation...

So as a first step, it improve the situation for all the PHP version lower than 8.3.

VincentLanglet (commented):
> Also: what about add method?

```
public DateTime::add(DateInterval $interval): DateTime
```
does not throw exceptions.

ondrejmirtes:
I apologize, I didn't check the method signature before.

But I don't see the same behaviour you do. Can you show me when the `sub` method throws an exception and not the DateInterval constructor? https://3v4l.org/pfAXh

VincentLanglet:
I made a copy-paste error, it should have been a `DateInvalidOperationException` in the `DateTimeSubMethodThrowTypeExtension`

> But I don't see the same behaviour you do. Can you show me when the `sub` method throws an exception and not the DateInterval constructor? [3v4l.org/pfAXh](https://3v4l.org/pfAXh)

See https://3v4l.org/T91Mv

It seems like the DateInterval constructor does not accept relative interval.

So 
```
$datetime->sub(new \DateInterval(...))
```
won't throw an exception while
```
$datetime->sub(DateInterval::createFromDateFormat(...))
```
could.

ondrejmirtes:
Thank you.

PR #3354, opened 2024-08-26
Allow dot-prefixed config files

sreichel:
- closes https://github.com/phpstan/phpstan/issues/8567 (?)

PR #3379, opened 2024-08-31
Respect dist priority for dot or not dot files

VincentLanglet:
Cf https://github.com/phpstan/phpstan-src/pull/3354#discussion_r1739760051


ondrejmirtes:
Thank you!

PR #3380, opened 2024-09-01
TypeSpecifier: Narrow `(bool) $expr` like `$expr == true`

staabm:
closes https://github.com/phpstan/phpstan/issues/8881
closes https://github.com/phpstan/phpstan/issues/7685
closes https://github.com/phpstan/phpstan/issues/6006
closes https://github.com/phpstan/phpstan/issues/10528


tests/PHPStan/Analyser/nsrt/narrow-bool-cast.php
```diff
@@ -0,0 +1,64 @@
...
+	}
+	assertType('array{}|array{string}', $matches);
+}
+
+
+interface Reader {
+	public function getFilePath(): string|false;
+}
+
+function bug7685(Reader $reader): void {
```
ondrejmirtes: Extremely cool! But please follow the usual convention for regression tests - one file per issue, with `bug-xxx.php` filename and namespace. I'm used to look for them in this format.

ondrejmirtes (changes requested):
Also implementing other casts would be cool as hell in a follow up PR :)

ondrejmirtes:
Thank you!

PR #3381, opened 2024-09-01
Update ConstantArrayType::accepts() for lists

VincentLanglet:
First fix of https://github.com/phpstan/phpstan/issues/11600
but does not close the issue since the second part is not fixed (and is a hard one...).

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -303,6 +303,10 @@ class ConstantArrayType extends ArrayType implements ConstantType
 			return AcceptsResult::createFromBoolean(count($type->keyTypes) === 0);
 		}
 
+		if ($this->isList()->yes() && $type->isList()->no()) {
+			return AcceptsResult::createNo();
+		}
+
 		$result = AcceptsResult::createYes();
 		foreach ($this->keyTypes as $i => $keyType) {
 			$valueType = $this->valueTypes[$i];
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -307,7 +307,12 @@ class ConstantArrayType extends ArrayType implements ConstantType
 			return AcceptsResult::createNo();
 		}
 
-		$result = AcceptsResult::createYes();
+		if ($this->isList()->yes() && $type->isList()->maybe()) {
+			$result = AcceptsResult::createMaybe();
+		} else {
+			$result = AcceptsResult::createYes();
+		}
+
 		foreach ($this->keyTypes as $i => $keyType) {
 			$valueType = $this->valueTypes[$i];
 			$hasOffsetValueType = $type->hasOffsetValueType($keyType);
--- a/tests/PHPStan/Rules/Generators/YieldFromTypeRuleTest.php
+++ b/tests/PHPStan/Rules/Generators/YieldFromTypeRuleTest.php
@@ -39,7 +39,6 @@ class YieldFromTypeRuleTest extends RuleTestCase
 			[
 				'Generator expects value type array{DateTime, DateTime, stdClass, DateTimeImmutable}, array{0: DateTime, 1: DateTime, 2: stdClass, 4: DateTimeImmutable} given.',
 				74,
-				'Array does not have offset 3.',
 			],
 			[
 				'Result of yield from (void) is used.',
--- a/tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
+++ b/tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
@@ -1017,7 +1017,7 @@ class ImpossibleCheckTypeFunctionCallRuleTest extends RuleTestCase
 					unset($i[2]);
 				}
 
-				return $i;
+				return array_values($i);
 			},
 			self::getLooseComparisonAgainsEnumsIssues(),
 		);
--- a/tests/PHPStan/Rules/Methods/CallStaticMethodsRuleTest.php
+++ b/tests/PHPStan/Rules/Methods/CallStaticMethodsRuleTest.php
@@ -841,4 +841,15 @@ class CallStaticMethodsRuleTest extends RuleTestCase
 		]);
 	}
 
+	public function testBug11600(): void
+	{
+		$this->checkThisOnly = false;
+		$this->analyse([__DIR__ . '/data/bug-11600.php'], [
+			[
+				'Parameter #1 $array of static method Bug10600\HelloWorld::sayHello() expects array{1, 2}, array{1: 2, 0: 1} given.',
+				18,
+			],
+		]);
+	}
+
 }
--- a/tests/PHPStan/Rules/Methods/data/bug-11600.php
+++ b/tests/PHPStan/Rules/Methods/data/bug-11600.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Bug10600;
+
+class HelloWorld
+{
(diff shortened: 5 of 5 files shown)


tests/PHPStan/Rules/Generators/YieldFromTypeRuleTest.php
```diff
@@ -39,7 +39,6 @@ public function testRule(): void
 			[
 				'Generator expects value type array{DateTime, DateTime, stdClass, DateTimeImmutable}, array{0: DateTime, 1: DateTime, 2: stdClass, 4: DateTimeImmutable} given.',
 				74,
-				'Array does not have offset 3.',
```
VincentLanglet, line 42: Since it's decided by the early return about list we don't have the information anymore.
- Not sure if it's an ok loss or we need to keep this (and how)
- Not sure if we need an help "This must be a list"

Your call.

src/Type/Constant/ConstantArrayTypeBuilder.php
```diff
@@ -257,6 +255,8 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $opt
 					return;
 				}
 			}
+
+			$this->isList = TrinaryLogic::createNo();
```
VincentLanglet: This fix is needed or https://github.com/phpstan/phpstan-src/blob/2e2757395b0c3bd9a845e265f62803008b751bbe/tests/PHPStan/Rules/Properties/data/bug-3339.php#L4 will fail.

PHPStan was considering that
```
$this->tuple = [true, true, true];

		for ($i = 0; $i < 3; ++$i)
		{
			$this->tuple[$i] = false;
		}

// Here, `$this->tuple` is array{bool, bool, bool} but is not considered as a list
```

I added a non regression test [array-is-list-offset.php](https://github.com/phpstan/phpstan-src/pull/3381/files#diff-27c8d6ba604b96b0c4c1daab0a1520d5dee644d212dd58c16e327ca8b807b2de) which fails without this fix.

The idea is to not touch the `list` property if we're in the case
```
if ($match) {
					$this->valueTypes = $valueTypes;
					return;
				}
```
were all the constantScalar keys are already matching.

VincentLanglet:
This assertion is false
https://github.com/phpstan/phpstan-src/blob/2e2757395b0c3bd9a845e265f62803008b751bbe/tests/PHPStan/Analyser/nsrt/array-is-list-unset.php#L22

cf https://onlinephp.io/c/28b76

src/Type/Constant/ConstantArrayType.php
```diff
@@ -734,7 +738,9 @@ public function unsetOffset(Type $offsetType): Type
 					$k++;
 				}
 
-				return new self($newKeyTypes, $newValueTypes, $this->nextAutoIndexes, $newOptionalKeys, TrinaryLogic::createNo());
+				$isList = $this->isList()->and(TrinaryLogic::createFromBoolean($i === count($this->keyTypes) - 1));
```
ondrejmirtes: A list stops being a list even if the last index is removed. One thing that makes list a list is what happens when you append a new item to it.

ondrejmirtes (changes requested):
Please separate the changes outside of `ConstantArrayType::accepts()` to another PR and add proper tests there.

I'd expect new test methods added to ConstantArrayTypeBuilderTest.

src/Type/Constant/ConstantArrayType.php
VincentLanglet: This is untrue according to https://onlinephp.io/c/28b76
`array_is_list` still returns True after we unset a value.

Without this fix, we get an error https://github.com/phpstan/phpstan-src/actions/runs/10653522924/job/29528721910?pr=3381

```
Parameter #2 $expectedErrors of method
PHPStan\Testing\RuleTestCase<PHPStan\Rules\Comparison\ImpossibleCheckTypeFunctionCallRule>::analyse()  
expects list<array{0: string, 1: int, 2?: string|null}>,
list<array{0: string, 1: int, 2?: string}> given.        
```

VincentLanglet:
With all the existing fix merged, this PR is now minimal.

This commit does the fix with a test https://github.com/phpstan/phpstan-src/pull/3381/commits/13b6747fee78e5208e08885c2afccd0572fa0563

But as shown by https://github.com/phpstan/phpstan-src/actions/runs/10943488747/job/30383079314?pr=3381 we're getting an error
```
Parameter #2 $expectedErrors of method                                                                 
         PHPStan\Testing\RuleTestCase<PHPStan\Rules\Comparison\ImpossibleCheckTypeFunctionCallRule>::analyse()  
         expects list<array{0: string, 1: int, 2?: string|null}>,                                               
         list<array{0: string, 1: int, 2?: string}> given.         
```
because of the code
```
static function (array $i): array {
    if (($i[2] ?? null) === 'BUG') {
         unset($i[2]);
     }

     return $i;
},
```
This is because of the unset call which means the array may not be a list in the PHPStan definition.

It's easily fixed with an array_values call but 
- The error message is not helping
- Technically the PHPdoc ask for `array{0: string, 1: int, 2?: string|null}` but never asked for a "PHPStan-list" in the way that the next offset might be something else than 3.

I didn't see an easy way to solve this situation unless PHPStan stop to consider that
```
list{string, string}
```
and
```
array{0: string, 1: string}
```
are the same thing when it comes from the phpdoc. `list` need to be explicit.

That would require to change
https://github.com/phpstan/phpstan-src/blob/e8027e7a59ba4e70925c89df51081902aa577e03/src/PhpDoc/TypeNodeResolver.php#L973
to something like
```
if ($typeNode->kind !== ArrayShapeNode::KIND_LIST) {
     $builder->degradeListCertainty();
}
```
which would do `$this->list = $this->list->and(TrinaryLogic::createMaybe())`.

I can try create another PR with such change but I'd like your opinion on this @ondrejmirtes.
It's an "important" change in the way PHPStan understand array vs list...

src/Type/Constant/ConstantArrayType.php
```diff
@@ -303,6 +303,10 @@ public function acceptsWithReason(Type $type, bool $strictTypes): AcceptsResult
 			return AcceptsResult::createFromBoolean(count($type->keyTypes) === 0);
 		}
 
+		if ($this->isList()->yes() && $type->isList()->no()) {
+			return AcceptsResult::createNo();
```
ondrejmirtes, line 307: What about `$type->isList()->maybe()`? I think that at that point the `$result` below should be assigned to `createMaybe()` and not `createYes()`

ondrejmirtes (changes requested):

src/Type/Constant/ConstantArrayType.php
VincentLanglet: I added https://github.com/phpstan/phpstan-src/pull/3381/commits/102b53b239123b4f560282c37fdcea44e144ce56

but what about the following issue https://github.com/phpstan/phpstan-src/pull/3381#issuecomment-2361360419 ?

ondrejmirtes: I don't really know what to do about that. Maybe this whole change is a hack and we need to approach it differently.

VincentLanglet: I think your right and this things is a hack, or at least not the right fix because it's not really related to list.

PHPStan sometimes consider an array is ordered and sometimes not.

An easy reproducer is https://phpstan.org/r/1ffe9cb5-65b1-4a7c-9010-3aa24560b462 vs real world example https://3v4l.org/TqTBn

Maybe `ConstantArray::getKeysOrValuesArray` is wrong... or a `ordered-array` (or `unordered-array`) need to be introduced...

At least this PR helped to find some others bugs.

PR #3382, opened 2024-09-01
Fix setting offset to a List

VincentLanglet:
Extracted from https://github.com/phpstan/phpstan-src/pull/3381

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/array-is-list-offset.php
+++ b/tests/PHPStan/Analyser/nsrt/array-is-list-offset.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace ArrayIsListUnset;
+
+use function PHPStan\Testing\assertType;
+
+class Foo {
+	/**
+	 * @param array{bool, bool, bool} $array
+	 * @param int<0, 1> $key
+	 */
+	public function test(array $array, int $key) {
+		assertType('int<0, 1>', $key);
+		assertType('true', array_is_list($array));
+
+		$array[$key] = false;
+		assertType('true', array_is_list($array));
+	}
+}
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/nsrt/array-is-list-offset.php
```diff
@@ -0,0 +1,19 @@
...
+	/**
+	 * @param array{bool, bool, bool} $array
+	 * @param int<0, 1> $key
+	 */
+	public function test(array $array, int $key) {
+		assertType('int<0, 1>', $key);
+		assertType('true', array_is_list($array));
+
+		$array[$key] = false;
+		assertType('true', array_is_list($array));
```
VincentLanglet, line 17: Prior to this PR, it was considered as false.

ondrejmirtes (changes requested):
Hey, I specifically asked for ConstantArrayTypeBuilderTest reproduction so please add that.

VincentLanglet:
> Hey, I specifically asked for ConstantArrayTypeBuilderTest reproduction so please add that.

Sorry for the misunderstanding, it should be better now (?)

tests/PHPStan/Type/Constant/ConstantArrayTypeBuilderTest.php
```diff
@@ -128,4 +129,29 @@ public function testIsList(): void
...
+		$builder->setOffsetValueType(new ConstantIntegerType(0), new NullType());
+		$this->assertTrue($builder->isList());
+
+		$builder->setOffsetValueType(new ConstantIntegerType(1), new NullType());
+		$this->assertTrue($builder->isList());
+
+		$builder->setOffsetValueType(new ConstantIntegerType(2), new NullType());
+		$this->assertTrue($builder->isList());
+
+		$oneOrZero = TypeCombinator::union(
```
ondrejmirtes, line 148: Please also write some counter-example test as well. Like when we're setting a union offset that does not exist yet, that will make the array be no longer a list. Thanks.

ondrejmirtes (changes requested):
Otherwise I couldn't find any problem with this edit 👍

ondrejmirtes:
Thank you.

PR #3384, opened 2024-09-01
TypeSpecifier: Narrow `(int) $expr` like `$expr != 0`

staabm:
refs https://github.com/phpstan/phpstan-src/pull/3380#discussion_r1740085169

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -196,15 +196,17 @@ final class TypeSpecifier
 				$context,
 				$rootExpr,
 			);
-		} elseif ($expr instanceof Expr\Cast\String_) {
+		} elseif (
+			$expr instanceof Expr\Cast\String_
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/nsrt/narrow-bool-cast.php
```diff
@@ -27,3 +27,13 @@ function doFoo(string $x, array $arr): void {
 	}
 	assertType('array{}|array{string}', $matches);
 }
+
+
+/** @param int<-5, 5> $x */
+function castInt($x) {
```
staabm: I couldn't think of a better test-sample. 

I think testing type-inference will get more meaningful examples, after we understand loose comparisons better

ondrejmirtes:
Conflict here

staabm:
> Conflict here

resolved

ondrejmirtes:
Thank you!

PR #3387, opened 2024-09-02
TypeSpecifier: Narrow `(string) $expr` like `$expr != ''`

staabm:
refs https://github.com/phpstan/phpstan-src/pull/3380#discussion_r1740085169

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -196,6 +196,13 @@ final class TypeSpecifier
 				$context,
 				$rootExpr,
 			);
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
I think you can narrow all of them as `!= false`.

Because here you missed `'0'` which is falsy.

staabm:
> I think you can narrow all of them as `!= false`.

great suggestion. since we have already a better loose comparison of constant bools, this also leads to more precise results already :)

staabm:
(I will de-duplicate the existing logic on the next PR, when currently open ones are merged)

ondrejmirtes:
Thank you.

PR #3389, opened 2024-09-02
Narrow array on count() with positive-int

staabm:
With this PR we also narrow in case of unbound integer ranges 

separated from https://github.com/phpstan/phpstan-src/pull/3373

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -39,6 +39,7 @@ use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\ConditionalTypeForParameter;
+use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantIntegerType;
@@ -1049,7 +1050,7 @@ final class TypeSpecifier
 					$offsetType = new ConstantIntegerType($i);
 					$valueTypesBuilder->setOffsetValueType($offsetType, $type->getOffsetValueType($offsetType), true);
 				}
-			} else {
+			} elseif ($type->isConstantArray()->yes()) {
 				for ($i = $sizeType->getMin();; $i++) {
 					$offsetType = new ConstantIntegerType($i);
 					$hasOffset = $type->hasOffsetValueType($offsetType);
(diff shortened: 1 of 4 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -1057,7 +1058,11 @@ private function turnListIntoConstantArray(FuncCall $countFuncCall, Type $type,
 				}
 
 			}
-			return $valueTypesBuilder->getArray();
+
+			$arrayType = $valueTypesBuilder->getArray();
```
staabm, line 1062: compensated the now passed in maybe unbound `IntegerRangeType`s

tests/PHPStan/Analyser/nsrt/count-type.php
```diff
@@ -44,12 +44,12 @@ public function doFooBar(
 		if (count($arr) == $maybeZero) {
 			assertType('array', $arr);
 		} else {
-			assertType('non-empty-array', $arr);
+			assertType('array', $arr);
 		}
 		if (count($arr) === $maybeZero) {
 			assertType('array', $arr);
 		} else {
-			assertType('non-empty-array', $arr);
```
staabm, line 52: I think I did a mistake here in previous PR, as the array can still be empty in the else branch

https://3v4l.org/409Jp

src/Analyser/TypeSpecifier.php
```diff
@@ -1046,7 +1047,7 @@ private function turnListIntoConstantArray(FuncCall $countFuncCall, Type $type,
 					$offsetType = new ConstantIntegerType($i);
 					$valueTypesBuilder->setOffsetValueType($offsetType, $type->getOffsetValueType($offsetType), true);
 				}
-			} else {
+			} elseif ($type->isConstantArray()->yes()) {
```
staabm, line 1053: compensated the now passed in maybe unbound `IntegerRangeType`s

src/Analyser/TypeSpecifier.php
```diff
@@ -2140,6 +2097,70 @@ public function resolveIdentical(Expr\BinaryOp\Identical $expr, Scope $scope, Ty
 		}
 		$rightType = $scope->getType($rightExpr);
 
+		if (
+			!$context->null()
+			&& $unwrappedLeftExpr instanceof FuncCall
+			&& count($unwrappedLeftExpr->getArgs()) >= 1
+			&& $unwrappedLeftExpr->name instanceof Name
+			&& in_array(strtolower((string) $unwrappedLeftExpr->name), ['count', 'sizeof'], true)
+			&& $rightType->isInteger()->yes()
```
staabm, line 2103: uses now a proper `Type` method instead of `&& $constantType instanceof ConstantIntegerType`

ondrejmirtes:
Thank you!

PR #3390, opened 2024-09-02
Improve type narrowing after string functions

staabm:
Before this PR we only narrowed in comparisons with constant strings. 

After this PR it works for any type of string

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1201,41 +1201,6 @@ final class TypeSpecifier
 		$constantStringValue = $scalarValues[0];
 
 		if (
-			$context->truthy()
-			&& $exprNode instanceof FuncCall
-			&& $exprNode->name instanceof Name
-			&& in_array(strtolower($exprNode->name->toString()), [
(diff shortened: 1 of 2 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -2174,6 +2139,41 @@ public function resolveIdentical(Expr\BinaryOp\Identical $expr, Scope $scope, Ty
...
+				'mb_substr', 'mb_strstr', 'mb_stristr', 'mb_strchr', 'mb_strrchr', 'mb_strtolower', 'mb_strtoupper', 'mb_ucfirst', 'mb_lcfirst',
+				'ucwords', 'mb_convert_case', 'mb_convert_kana',
+			], true)
+			&& isset($unwrappedLeftExpr->getArgs()[0])
+			&& $rightType->isNonEmptyString()->yes()
+		) {
+			$argType = $scope->getType($unwrappedLeftExpr->getArgs()[0]->value);
+
+			if ($argType->isString()->yes()) {
+				if ($rightType->isNonFalsyString()->yes()) {
```
staabm, line 2154: instead of `if ($constantStringValue !== '0') {` we now rely on the type-system

ondrejmirtes:
Thank you!

PR #3394, opened 2024-09-03
ArrayShapeMatcher - Fix alternations containing a $-only case

staabm:
closes https://github.com/phpstan/phpstan/issues/11622


ondrejmirtes:
Thank you!

PR #3395, opened 2024-09-03
ArrayShapeMatcher - Don't optimize alternations with optional-groups for tagged-unions

staabm:
closes https://github.com/phpstan/phpstan/issues/11604


ondrejmirtes:
Thank you!

PR #3396, opened 2024-09-03
Add non regression test for bug 6642

VincentLanglet:
I saw in https://github.com/phpstan/phpstan/discussions/11623 that the issue https://github.com/phpstan/phpstan/issues/6642 is referenced but not closed.

I assume you'd like a non regression test so I added one.

Closes https://github.com/phpstan/phpstan/issues/6642


ondrejmirtes:
Thank you.

PR #3399, opened 2024-09-04
version_compare() operator-arg can be null

staabm:
refs https://github.com/phpstan/phpstan/issues/4457

tested locally with the snippet of the issue


ondrejmirtes:
Thank you.

VincentLanglet:
Is a non regression test wanted to close https://github.com/phpstan/phpstan/issues/4457 ?

ondrejmirtes:
The whole issue was not addressed, only part of it.

PR #3404, opened 2024-09-05
Fix ImpossibleCheckTypeFunctionCallRule for `is_subclass_of` and `is_a`

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/3979
which can be simplified to https://phpstan.org/r/ef72a996-8fbe-4a4e-831f-4d78c542ad19

I'm not sure there is a better (easy ?) fix for this issue. `is_subclass_of($string, $bar)` specify the type of `$string` to class-string when true, but it doesn't mean that the call is useless if the param is already a class-string.

I discovered there was an exclusion of `is_a` in the `ImpossibleCheckTypeFunctionCallRule` (maybe for the same reason)
so I think it could be the same for `is_subclass_of` since the behavior is similar.

### DIFF
--- a/tests/PHPStan/Analyser/TypeSpecifierTest.php
+++ b/tests/PHPStan/Analyser/TypeSpecifierTest.php
@@ -1133,9 +1133,7 @@ class TypeSpecifierTest extends PHPStanTestCase
 					new Arg(new Variable('stringOrNull')),
 					new Arg(new Expr\ConstFetch(new Name('false'))),
 				]),
-				[
-					'$object' => 'object',
-				],
+				[],
 				[],
 			],
 			[
--- a/src/Rules/Comparison/ImpossibleCheckTypeFunctionCallRule.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeFunctionCallRule.php
@@ -38,9 +38,6 @@ final class ImpossibleCheckTypeFunctionCallRule implements Rule
 		}
 
 		$functionName = (string) $node->name;
-		if (in_array(strtolower($functionName), ['is_a', 'is_subclass_of'], true)) {
-			return [];
-		}
 		$isAlways = $this->impossibleCheckTypeHelper->findSpecifiedType($scope, $node);
 		if ($isAlways === null) {
 			return [];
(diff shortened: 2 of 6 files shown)


ondrejmirtes:
Changing the rule is not going to help you against "Result of && is always false.".

I'd much rather change the logic in IsSubclassOfFunctionTypeSpecifyingExtension.

tests/PHPStan/Analyser/TypeSpecifierTest.php
```diff
@@ -1133,9 +1133,7 @@ public function dataCondition(): iterable
 					new Arg(new Variable('stringOrNull')),
 					new Arg(new Expr\ConstFetch(new Name('false'))),
 				]),
-				[
-					'$object' => 'object',
```
VincentLanglet, line 1137:
```
if (is_subclass_of($object, $stringOrNull, false))
```
specified `$object` as `object` but this was already known so it's not a big loss.

VincentLanglet:
> I'd much rather change the logic in IsSubclassOfFunctionTypeSpecifyingExtension.

Indeed, it seems possible with
```
$superType = $allowString
			? TypeCombinator::union(new ObjectWithoutClassType(), new ClassStringType())
			: new ObjectWithoutClassType();
		// prevent false-positives in IsAFunctionTypeSpecifyingHelper
		if ($superType->isSuperTypeOf($objectOrClassType)->yes() && !$classType->isClassStringType()->yes()) {
			return new SpecifiedTypes([], []);
		}
```

I added the same fix to `IsAFunctionTypeSpecifyingExtension` and remove the special exclusion in the `ImpossibleCheckTypeFunctionCallRule`.

VincentLanglet:
> I'd much rather change the logic in IsSubclassOfFunctionTypeSpecifyingExtension.

I think it was possible indeed, I get something working.

VincentLanglet:
Friendly ping @ondrejmirtes ; this PR is ready to be re-reviewed :)

ondrejmirtes:
Thank you.

ondrejmirtes:
Unfortunately I have to revert this. it leads to false positives like: https://phpstan.org/r/5941cd95-26d7-49cf-a2e0-ffc21c4a7697

VincentLanglet:
> Unfortunately I have to revert this. it leads to false positives like: [phpstan.org/r/5941cd95-26d7-49cf-a2e0-ffc21c4a7697](https://phpstan.org/r/5941cd95-26d7-49cf-a2e0-ffc21c4a7697)

No problem, I'll try to improve the idea in https://github.com/phpstan/phpstan-src/pull/3788
Did you had another false positive I have to add as non-regression test ?

PR #3405, opened 2024-09-05
Prevent resolving conditional types in callable param/return types

rvanvelzen:
Unresolvable conditional types should not be forcibly resolved in callable param/return types because they generally depend on calling that callable.

Fix phpstan/phpstan#11472.

### DIFF
--- a/src/Type/TypeUtils.php
+++ b/src/Type/TypeUtils.php
@@ -411,9 +411,20 @@ class TypeUtils
 
 	public static function resolveLateResolvableTypes(Type $type, bool $resolveUnresolvableTypes = true): Type
 	{
-		return TypeTraverser::map($type, static function (Type $type, callable $traverse) use ($resolveUnresolvableTypes): Type {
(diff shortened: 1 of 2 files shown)


src/Type/TypeUtils.php
```diff
@@ -411,11 +411,21 @@ public static function containsTemplateType(Type $type): bool
 
 	public static function resolveLateResolvableTypes(Type $type, bool $resolveUnresolvableTypes = true): Type
 	{
-		return TypeTraverser::map($type, static function (Type $type, callable $traverse) use ($resolveUnresolvableTypes): Type {
-			while ($type instanceof LateResolvableType && ($resolveUnresolvableTypes || $type->isResolvable())) {
+		$ignoreResolveUnresolvableTypesLevel = 0;
```
ondrejmirtes, line 415: Just do `@var int` here :)

ondrejmirtes (changes requested):
I'm sorry about the Composer build failures here. I need to look into them. Probably has something to do with publishing the 2.0.x branch.

ondrejmirtes:
Thank you!

PR #3406, opened 2024-09-05
Fix return type of `array_reverse()` with optional keys

herndlm:
Fixes https://github.com/phpstan/phpstan/issues/11549

We cannot simply reverse indices and values for arrays with optional keys and preserve_keys=false. In that case the array has to be re-built to account for the optional keys properly.

fyi @tscni and thanks for the nice example / test case :)

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -50,7 +50,6 @@ use function array_map;
 use function array_merge;
 use function array_pop;
 use function array_push;
-use function array_reverse;
 use function array_slice;
 use function array_unique;
 use function array_values;
@@ -886,14 +885,16 @@ class ConstantArrayType extends ArrayType implements ConstantType
 
 	public function reverseArray(TrinaryLogic $preserveKeys): Type
 	{
-		$keyTypesReversed = array_reverse($this->keyTypes, true);
-		$keyTypes = array_values($keyTypesReversed);
-		$keyTypesReversedKeys = array_keys($keyTypesReversed);
-		$optionalKeys = array_map(static fn (int $optionalKey): int => $keyTypesReversedKeys[$optionalKey], $this->optionalKeys);
+		$builder = ConstantArrayTypeBuilder::createEmpty();
 
-		$reversed = new self($keyTypes, array_reverse($this->valueTypes), $this->nextAutoIndexes, $optionalKeys, TrinaryLogic::createNo());
+		for ($i = count($this->keyTypes) - 1; $i >= 0; $i--) {
+			$offsetType = $preserveKeys->yes() || $this->keyTypes[$i]->isInteger()->no()
+				? $this->keyTypes[$i]
+				: null;
+			$builder->setOffsetValueType($offsetType, $this->valueTypes[$i], $this->isOptionalKey($i));
+		}
 
-		return $preserveKeys->yes() ? $reversed : $reversed->reindex();
+		return $builder->getArray();
 	}
 
 	public function searchArray(Type $needleType): Type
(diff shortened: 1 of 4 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -839,14 +839,29 @@ public function popArray(): Type
...
+			$builder->setOffsetValueType($offsetType, $this->valueTypes[$i], $this->isOptionalKey($i));
+		}
 
-		return $preserveKeys->yes() ? $reversed : $reversed->reindex();
+		$array = $builder->getArray();
+		if (!$array instanceof self) {
+			throw new ShouldNotHappenException();
+		}
+
+		return $array;
```
herndlm: it would be possible to use the same implementation with the builder also  if $preserveKeys->yes(), but the existing code which reverses indices and values might be slightly more optimised I suppose? not sure. we can have either both or only the new one :)

src/Type/Constant/ConstantArrayType.php
```diff
@@ -839,14 +839,29 @@ public function popArray(): Type
...
+			$offsetType = $this->keyTypes[$i]->isInteger()->no()
+				? $this->keyTypes[$i]
+				: null;
+			$builder->setOffsetValueType($offsetType, $this->valueTypes[$i], $this->isOptionalKey($i));
+		}
 
-		return $preserveKeys->yes() ? $reversed : $reversed->reindex();
+		$array = $builder->getArray();
+		if (!$array instanceof self) {
+			throw new ShouldNotHappenException();
```
ondrejmirtes: I think we should remove this condition and just change the return type of this method to `Type`. It's private so no big deal.

ondrejmirtes (changes requested):

src/Type/Constant/ConstantArrayType.php
herndlm: had a similar thought in the previous MR, but the whole `reverseConstantArray()` basically only exists to ensure the return type is `self` so that https://github.com/phpstan/phpstan-src/blob/1.12.2/src/Type/Constant/ConstantArrayType.php#L1141-L1143 is less annoying to deal with (where the return type is also self and recursion happens too)...

it bothers me a bit, most likely `Type` should have been the return type everywhere all along.

herndlm: might be able to improve this and deprecate old methods with narrow return types by creating `Type::sliceArray()` and `Type::chunkArray()`, but not sure if that's needed / worth it.

src/Type/Constant/ConstantArrayType.php
herndlm: I remove the old logic, it feels like it makes things unnecessarily more complex and tbh I don't even know how much more optimised it is

src/Type/Constant/ConstantArrayType.php
ondrejmirtes: I'd be rather more safe and I propose these steps:

1) Remove the ShouldNotHappenException
2) At the place where reverseConstantArray is called, check if the returned type is ConstantArrayType and only after that call `slice`...

ondrejmirtes: But otherwise Type::sliceArray and Type::chunkArray would be nice :)

herndlm: > I'd be rather more safe and I propose these steps:
> 
>     1. Remove the ShouldNotHappenException
> 
>     2. At the place where reverseConstantArray is called, check if the returned type is ConstantArrayType and only after that call `slice`...

you mean without throwing exceptions, and e.g. returning the array right away if it's not a constant array anymore, right? would mean more instanceof I suppose, but I think I can clean those up later when adding new type methods

herndlm: but it feels a bit weird to return the unsliced array after the first reverse for example. it should really not happen I suppose xD

herndlm: going to do the refactor first..

herndlm: all good now :)

herndlm:
after merging this up into 2.0.x a small but nice `instanceof self` cleanup/removal in `ConstantArrayType` in 2.0.x will be possible which I'll open a PR for :)

ondrejmirtes:
Thank you.

PR #3407, opened 2024-09-06
Narrow string on *strlen() with positive-int

staabm:
closes https://github.com/phpstan/phpstan/issues/11558

separated from https://github.com/phpstan/phpstan-src/pull/3373

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1124,39 +1124,6 @@ final class TypeSpecifier
 			));
 		}
 
-		if (
-			!$context->null()
-			&& $exprNode instanceof FuncCall
-			&& count($exprNode->getArgs()) === 1
-			&& $exprNode->name instanceof Name
-			&& in_array(strtolower((string) $exprNode->name), ['strlen', 'mb_strlen'], true)
-			&& $constantType instanceof ConstantIntegerType
-		) {
-			if ($constantType->getValue() < 0) {
-				return $this->create($exprNode->getArgs()[0]->value, new NeverType(), $context, false, $scope, $rootExpr);
-			}
-
-			if ($context->truthy() || $constantType->getValue() === 0) {
-				$newContext = $context;
-				if ($constantType->getValue() === 0) {
-					$newContext = $newContext->negate();
-				}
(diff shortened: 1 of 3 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -1107,39 +1107,6 @@ private function specifyTypesForConstantBinaryExpression(
 			));
 		}
 
-		if (
-			!$context->null()
-			&& $exprNode instanceof FuncCall
-			&& count($exprNode->getArgs()) === 1
-			&& $exprNode->name instanceof Name
-			&& in_array(strtolower((string) $exprNode->name), ['strlen', 'mb_strlen'], true)
-			&& $constantType instanceof ConstantIntegerType
```
staabm, line 1133: instead of comparison with `ConstantIntegerType` we now support `IntegerRangeType`

src/Analyser/TypeSpecifier.php
```diff
@@ -2192,14 +2195,11 @@ public function resolveIdentical(Expr\BinaryOp\Identical $expr, Scope $scope, Ty
 			}
 		}
 
-		if ($rightType->isInteger()->yes() || $rightType->isString()->yes()) {
+		if ($rightType->isString()->yes()) {
```
staabm, line 2215: as we no longer do `ConstantIntegerType` related checks in `specifyTypesForConstantBinaryExpression` we can simplify this loop again to what it looked like a few weeks ago.

tests/PHPStan/Analyser/TypeSpecifierTest.php
```diff
@@ -1283,7 +1283,7 @@ public function dataCondition(): iterable
 					),
 				),
 				[
-					'$foo' => 'non-empty-string',
+					'$foo' => "string & ~''",
```
VincentLanglet, line 1286: Isn't this change weird ?

Shouldn't `string & ~''` be considered as non-empty-string ?

staabm: I think its the same thing, but just a difference how the type is printed

ondrejmirtes:
Thank you!

PR #3408, opened 2024-09-06
Add `Type::chunkArray()`

herndlm:
just cleaning up to simplify things and hopefully be able to cleanup some type checks in ConstantArrayType later..

### DIFF
--- a/src/Type/Accessory/HasOffsetType.php
+++ b/src/Type/Accessory/HasOffsetType.php
@@ -175,6 +175,11 @@ class HasOffsetType implements CompoundType, AccessoryType
 		return $this;
 	}
 
+	public function chunkArray(Type $lengthType, TrinaryLogic $preserveKeys): Type
+	{
+		return new NonEmptyArrayType();
+	}
+
 	public function fillKeysArray(Type $valueType): Type
 	{
 		return new NonEmptyArrayType();
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -73,6 +73,7 @@ class ConstantArrayType extends ArrayType implements ConstantType
 {
 
 	private const DESCRIBE_LIMIT = 8;
+	private const CHUNK_FINITE_TYPES_LIMIT = 5;
 
 	private TrinaryLogic $isList;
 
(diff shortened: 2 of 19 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -780,6 +781,36 @@ public function unsetOffset(Type $offsetType): Type
 		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, $isList);
 	}
 
+	public function chunkArray(Type $lengthType, TrinaryLogic $preserveKeys): Type
+	{
+		$biggerOne = IntegerRangeType::fromInterval(1, null);
+		$finiteTypes = $lengthType->getFiniteTypes();
+		if ($biggerOne->isSuperTypeOf($lengthType)->yes() && count($finiteTypes) < self::CHUNK_FINITE_TYPES_LIMIT) {
```
staabm, line 788: maybe?

```suggestion
		$limit = IntegerRangeType::fromInterval(1, self::CHUNK_FINITE_TYPES_LIMIT);
		if ($limit->isSuperTypeOf($lengthType)->yes()) {
		  $finiteTypes = $lengthType->getFiniteTypes();
```

herndlm: that seems to break something. but I just copied it over from the extension, I think you added it originally ;). we could still improve later, I don't have the time and patience to get into this now tbh :)

staabm: :+1:

herndlm: fyi it was just a off-by-one thing. Limit needs to be int<1, 4> because before there was a `<` check instead of `<=` :)

these range super type checks + finite types are interesting I have to say hmm

herndlm: oh, weird. it breaks another test now, because `int<1, 4>` is considered to be a super type of `60` 🤔

herndlm: ah right, no, not that easy, because e.g. `int<1, 4>` is not a super type of constant int 60, but we still want it to do the high precision chunking. I'll leave it for now :)

herndlm:
This one should be good. Waiting for it to get merged to open the next one for sliceArray btw 😊

src/Type/Accessory/HasOffsetType.php
```diff
@@ -175,6 +175,11 @@ public function unsetOffset(Type $offsetType): Type
 		return $this;
 	}
 
+	public function chunkArray(Type $lengthType, TrinaryLogic $preserveKeys): Type
+	{
+		return new NonEmptyArrayType();
```
ondrejmirtes, line 180: I think this once and the one in HasOffsetValueType is wrong. You should test what happens when you pass `string&hasOffset` to `array_chunk`.

A solution might be to return `$this`, or maybe handle this in `IntersectionType::chunkArray` (which knows what this is intersected with).

ondrejmirtes (changes requested):

src/Type/Accessory/HasOffsetType.php
herndlm: I see what you mean. It's not wrong IMO, but it could be better. I pushed tests to show the current behaviour.

Within `HasOffset*Type` we cannot return `$this`, that would be wrong because the offset ends up being nested after `array_chunk`. `NonEmptyArray()` is correct IMO but we could do better. See also https://3v4l.org/dCOoR which is the same as the type in the new tests (`non-empty-list<non-empty-list<mixed>>` or `non-empty-list<non-empty-array>`.

Using my new tests to check the current behaviour also shows that this PR would make it less precise, so that's definitely not good :/ https://phpstan.org/r/c4dfc504-4d6f-41fc-bad2-7bb9c90272bc

And yes, you're correct, looks like making it better means adapting `IntersectionType` most likely :) I'll try to push the fix soon.

herndlm: or is there a way to return this nested structure in the accessory type? so far I'm then getting a `*NEVER*` as result if I do that :/ guess I need to debug the types that it operates on a bit more..

herndlm: never mind, I think this is easier than I thought 😊

herndlm: sorry for the noise :P

I managed to return whatever type I want, but realized that the current "live" behavior might be wrong already. e.g. how would we type the return type of https://3v4l.org/TfHjW assuming it's not a constant array but there are those 3 known offsets and there could be more we don't know about since its a generic array.

At first I thought it would be ` non-empty-list<hasOffsetValue('a','foo')&hasOffsetValue('b','bar')&hasOffsetValue('c','baz')&non-empty-array>` but that's not correct since we have 3 arrays inside and the generic representation is just `non-empty-list<non-empty-array>`, right?

Then I thought we could use the offset and value info to make something like
`non-empty-list<non-empty-array<'a'|'b'|'c','foo'|'bar'|'baz'>>` but that's also wrong since, again, there could be other keys and values.

I seem to be coming back to the current behavior / result or am I missing something? As long as it's not a constant array we don't know in which "chunk" that offset (+ value) might end up in, right?

ondrejmirtes:
Alright, let's try it like this :) Thank you.

PR #3409, opened 2024-09-06
Don't prevent checking for curl_init() false returns

tscni:
Fixes https://github.com/phpstan/phpstan/issues/11640


conf/config.neon
```diff
@@ -1360,11 +1360,6 @@ services:
 		tags:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
 
-	-
-		class: PHPStan\Type\Php\CurlInitReturnTypeExtension
-		tags:
-			- phpstan.broker.dynamicFunctionReturnTypeExtension
```
tscni, line 1366: When we no longer ignore `curl_init()` memory allocation & initialization issues, this extension becomes mostly useless.
It would really only handle the following
- If the input contains `\0` it will always return false / never
- If the input is really huge it will likely return false
- If the input is an invalid URL or uses the `file://` scheme it might return false on PHP 7

These are all very unlikely edge-cases, thus it seemed more reasonable to move the return type definition to the function map.

In short: `curl_init()` will now always return `(resource|false)` or `(CurlHandle|false)`.

ondrejmirtes:
Thank you!

PR #3411, opened 2024-09-06
Handle conditional return type in ArrayMap

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10715


VincentLanglet:
Failure on PHP 7.4 seems unrelated and are present on the base branch https://github.com/phpstan/phpstan-src/actions/runs/10734305611/job/29769226428

ondrejmirtes:
I've used this instead https://github.com/phpstan/phpstan-src/pull/3425 but thank you.

PR #3415, opened 2024-09-06
Add non regression test for 10499

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10499


ondrejmirtes:
Thank you!

PR #3417, opened 2024-09-07
Implement InvalidDivisionOperationRule

staabm:
closes https://github.com/phpstan/phpstan/issues/11090

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -63,5 +63,6 @@ parameters:
 		explicitThrow: true
 		absentTypeChecks: true
 		requireFileExists: true
+		divisionByZero: true
 	stubFiles:
 		- ../stubs/bleedingEdge/Rule.stub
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
There is already InvalidBinaryOperationRule so we don't need a new rule: https://phpstan.org/r/42eed56b-0bb3-42b9-8f97-8c53752c4d35

VincentLanglet:
Also, exception checks about `DivisionByZeroError` already covering this case.
It require
- Either to catch the DivisionByZeroError
- Either to add it to the phpdoc

https://phpstan.org/r/06bc1121-7847-4c30-9d19-59c34c9369a0

Adding a new rule will create a duplicate and false positive since
```
	/**
	 * @throws DivisionByZeroError
	 */
	public function sayHello2(int $i, int $j): void
	{
		echo $i / $j; // Not error it's documented
	}
```
is fully valid code.

staabm:
thanks for pointing this out.

> You have to enable "missingCheckedExceptionInThrows", "DivisionByZeroError"

I don't think coverage for a division-by-zero error should be a opt-in thing and "hidden" in the config file.

> There is already InvalidBinaryOperationRule

@ondrejmirtes wdyt about adding the logic of this PR into InvalidBinaryOperationRule (or the type system) while checking the config whether DivisionByZeroError was configured as missingCheckedExceptionInThrows to prevent duplicate errors?

staabm:
submitted a alternative impl as a base for discussion: https://github.com/phpstan/phpstan-src/pull/3419

PR #3419, opened 2024-09-08
Report maybe division by zero in InvalidBinaryOperationRule

staabm:
alternative implementation to https://github.com/phpstan/phpstan-src/pull/3417

closes https://github.com/phpstan/phpstan/issues/11090

### DIFF
--- a/src/Rules/Operators/InvalidBinaryOperationRule.php
+++ b/src/Rules/Operators/InvalidBinaryOperationRule.php
@@ -2,17 +2,24 @@
 
 namespace PHPStan\Rules\Operators;
 
+use DivisionByZeroError;
 use PhpParser\Node;
 use PHPStan\Analyser\MutatingScope;
 use PHPStan\Analyser\Scope;
 use PHPStan\Node\Printer\ExprPrinter;
+use PHPStan\Parser\TryCatchTypeVisitor;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\BenevolentUnionType;
+use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\VerbosityLevel;
+use function array_map;
 use function sprintf;
 use function strlen;
 use function substr;
@@ -27,6 +34,7 @@ final class InvalidBinaryOperationRule implements Rule
 		private ExprPrinter $exprPrinter,
 		private RuleLevelHelper $ruleLevelHelper,
 		private bool $bleedingEdge,
+		private bool $reportMaybes,
 	)
 	{
 	}
@@ -108,6 +116,39 @@ final class InvalidBinaryOperationRule implements Rule
 			->assignVariable($rightName, $rightType, $rightType);
 
 		if (!$scope->getType($newNode) instanceof ErrorType) {
+			if (
+				$this->reportMaybes
+				&& (
+					$node instanceof Node\Expr\AssignOp\Div
+					|| $node instanceof Node\Expr\AssignOp\Mod
+					|| $node instanceof Node\Expr\BinaryOp\Div
+					|| $node instanceof Node\Expr\BinaryOp\Mod
+				)
+				&& !$this->isDivisionByZeroCaught($node)
+				&& !$this->hasDivisionByZeroThrowsTag($scope)
+			) {
+				$rightType = $scope->getType($right);
+				// as long as we don't support float-ranges, we prevent false positives for maybe floats
+				if ($rightType instanceof BenevolentUnionType && $rightType->isFloat()->maybe()) {
+					return [];
+				}
+
+				$zeroType = new ConstantIntegerType(0);
+				if ($zeroType->isSuperTypeOf($rightType)->maybe()) {
+					return [
+						RuleErrorBuilder::message(sprintf(
+							'Binary operation "%s" between %s and %s might result in an error.',
+							substr(substr($this->exprPrinter->printExpr($newNode), strlen($leftName) + 2), 0, -(strlen($rightName) + 2)),
+							$scope->getType($left)->describe(VerbosityLevel::value()),
+							$scope->getType($right)->describe(VerbosityLevel::value()),
+						))
+							->line($left->getStartLine())
+							->identifier(sprintf('%s.invalid', $identifier))
+							->build(),
+					];
+				}
+			}
+
 			return [];
 		}
 
@@ -122,6 +163,33 @@ final class InvalidBinaryOperationRule implements Rule
 				->identifier(sprintf('%s.invalid', $identifier))
 				->build(),
 		];
+	}
+
+	private function isDivisionByZeroCaught(Node $node): bool
+	{
+		$tryCatchTypes = $node->getAttribute(TryCatchTypeVisitor::ATTRIBUTE_NAME);
+		if ($tryCatchTypes === null) {
+			return false;
+		}
+
+		$tryCatchType = TypeCombinator::union(...array_map(static fn (string $class) => new ObjectType($class), $tryCatchTypes));
+
+		return $tryCatchType->isSuperTypeOf(new ObjectType(DivisionByZeroError::class))->yes();
+	}
+
+	private function hasDivisionByZeroThrowsTag(Scope $scope): bool
+	{
+		$function = $scope->getFunction();
+		if ($function === null) {
+			return false;
+		}
+
+		$throwsType = $function->getThrowType();
+		if ($throwsType === null) {
+			return false;
+		}
+
+		return $throwsType->isSuperTypeOf(new ObjectType(DivisionByZeroError::class))->yes();
 	}
 
 }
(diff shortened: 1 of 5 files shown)


staabm:
(submitted, so we can discuss the 2 different approaches)

src/Rules/Operators/InvalidBinaryOperationRule.php
```diff
@@ -108,6 +111,28 @@ public function processNode(Node $node, Scope $scope): array
...
 
 		if (!$scope->getType($newNode) instanceof ErrorType) {
+			if (
+				$node instanceof Node\Expr\AssignOp\Div
+				|| $node instanceof Node\Expr\AssignOp\Mod
+				|| $node instanceof Node\Expr\BinaryOp\Div
+				|| $node instanceof Node\Expr\BinaryOp\Mod
+			) {
+				$zeroType = new UnionType([new ConstantIntegerType(0), new ConstantFloatType(0.0)]);
+				if ($zeroType->isSuperTypeOf($rightType)->maybe()) {
```
ondrejmirtes, line 137: This logic should be implemented as a `$callback` for the RuleLevelHelper above instead. It will make sure this is correctly reported on level 2 (when it's always zero) and on level 7 (when it might be a zero).

ondrejmirtes (changes requested):
We should do it only for ConstantIntegerType. FloatType might always be a zero because we don't have float ranges yet.

src/Rules/Operators/InvalidBinaryOperationRule.php
staabm: the callback passed to `RuleLevelHelper` is only invoked for union types and intersection types.
in our case we pass in a expr which is resolved into a IntegerRangeType, which means the callback does not have any effect in this context.

ondrejmirtes: I get it. But still I'm missing at least asking for `$this->reportMaybes` so that maybe-zero is reported only on level 7+.

VincentLanglet:
> submitted a alternative impl as a base for discussion: #3419

So I assume you prefer to discuss about it here.

> thanks for pointing this out.
> 
> > You have to enable "missingCheckedExceptionInThrows", "DivisionByZeroError"
> 
> I don't think coverage for a division-by-zero error should be a opt-in thing and "hidden" in the config file.

I agree it could make sens to have such checks (at level 7 for int ranges).

Also, if you go in that path @staabm, currently you're implementing the rule for the `/` and `%`.
cf the throw point: https://github.com/phpstan/phpstan-src/blob/893b4cf9324981bd84703bb2ae12b219c02cc746/src/Analyser/NodeScopeResolver.php#L2316-L2321

But there is also:
- IntDiv: https://github.com/phpstan/phpstan-src/blob/893b4cf9324981bd84703bb2ae12b219c02cc746/src/Type/Php/IntdivThrowTypeExtension.php#L42-L45
- BcDiv: https://phpstan.org/r/b7160c2f-7040-43fc-ae29-22a74dbc9f93 (and currently there is no ThrowTypeExtension for BC methods https://phpstan.org/r/0e9c14a9-e502-4b4f-b7e4-8aa7dbc41b65)

We should be consistent about the PHPStan checks.

> > There is already InvalidBinaryOperationRule
> 
> @ondrejmirtes wdyt about adding the logic of this PR into InvalidBinaryOperationRule (or the type system) while checking the config whether DivisionByZeroError was configured as missingCheckedExceptionInThrows to prevent duplicate errors?

For the "What to do about the DivisionByZeroError", I'm not sure looking for `DivisionByZeroError` in `missingCheckedExceptionInThrows` might be the right thing to do:
- If I add `DivisionByZeroError` in the list of uncheckedException, it's maybe because I don't want to check for 0, so I won't want this rule too.
- If I add `Error` in the list of uncheckedException, we dunno if I was thinking about `DivisionByZeroError`.

If we look at something similar, like the TypeError, we can see in https://phpstan.org/r/06610d2b-47e5-48c5-8e25-fe4eabc331c9 that
- PHPStan reports an error about the type passed to the method
- PHPStan consider the catch is useless (but it is https://3v4l.org/46LqD)

That's a (good ?) strategy, and we could do the same for DivisionByZeroError by considering that since there is a rule checking this error, we dunno care about the error anymore. But I'm not sure it's the wanted path or Ondrej wouldn't have open the following issue https://github.com/phpstan/phpstan/issues/9146 (with the code https://phpstan.org/r/9fbbf072-4da5-4b79-9fda-4559405cb0f5).

So I dunno what's the best to avoid duplicated errors...

src/Rules/Operators/InvalidBinaryOperationRule.php
```diff
@@ -124,4 +157,16 @@ public function processNode(Node $node, Scope $scope): array
 		];
 	}
 
+	private function isDivisionByZeroCaught(Node $node): bool
```
staabm, line 168: inspired by https://github.com/phpstan/phpstan-src/blob/64d4a3b61c81aced05fa690bb7a418d34bfb4213/src/Rules/Comparison/MatchExpressionRule.php#L168-L181

VincentLanglet: Nice.

The only remaining question is to decide if
```
/** @throws DivisionByZeroError */
function foo() {
        $i / $x;
}
```
should be reported or not.

staabm: I have this already in the pipeline :).

it might make sense to extract the "isXXExceptionCaught" and "hasXXExceptionThrowsTag" into a service to ease re-use

staabm:
> * currently there is no ThrowTypeExtension

I have patch prepared, which will be submitted after other currently open PRs are merged

staabm:
unrelated CI error will be fixed in https://github.com/phpstan/phpstan-src/pull/3424

src/Rules/Operators/InvalidBinaryOperationRule.php
ondrejmirtes: Hey, it'd still be useful if the RuleLeveLHelper eliminated some situations. I can see that in the tests you're currently asserting an error that shouldn't be asserted because it's on a BenevolentUnionType.

tests/PHPStan/Rules/Operators/InvalidBinaryOperationRuleTest.php
```diff
@@ -659,6 +665,14 @@ public function testBenevolentUnion(): void
 				'Binary operation "/" between (array<string, string>|bool|int|object|resource) and BinaryOpBenevolentUnion\\Foo results in an error.',
 				105,
 			],
+			[
+				'Binary operation "/" between (array<string, string>|bool|int|object|resource) and (array<string, string>|bool|int|object|resource) might result in an error.',
```
ondrejmirtes, line 671: This one. It might be `int / int` so this shouldn't be reported.

ondrejmirtes: Oh I get it, but `int` can still include zero so this should be reported.

But: We can still create an artifical example for example of `(float|int)` that shouldn't be reported because we don't report `X / float`.

src/Rules/Operators/InvalidBinaryOperationRule.php
staabm: I tried testing the benevolent case, but for some reason the benevolence is lost somewhere?

https://phpstan.org/r/c0d12375-468d-46a2-8778-fcedc0d5e26f

ondrejmirtes: That's not how you type a benevolent union type in PHPDoc.

It's not officially supported, there's just a semi-secret pseudotype `__benevolent<...>`. Check the functionMap.

src/Rules/Operators/InvalidBinaryOperationRule.php
```diff
@@ -127,7 +128,10 @@ public function processNode(Node $node, Scope $scope): array
 				&& !$this->hasDivisionByZeroThrowsTag($scope)
 			) {
 				$zeroType = new ConstantIntegerType(0);
-				if ($zeroType->isSuperTypeOf($rightType)->maybe()) {
+				if (
+					$zeroType->isSuperTypeOf($rightType)->maybe()
+					&& !$rightType instanceof BenevolentUnionType
```
VincentLanglet: This wouldn't go against `checkBenevolentUnionTypes` ?
https://phpstan.org/config-reference#checkbenevolentuniontypes

Shouldn't it be
```
&& ($this->checkBenevolentUnionTypes || !$rightType instanceof BenevolentUnionType)
```

staabm: I think the flag is working, because we are using the RuleLevelHelper, see https://github.com/phpstan/phpstan-src/pull/3419/files#diff-cb8642718212e42a4ffea566a1b6d116db66af1685c26e4effdba3ce72539302R883-R886

src/Rules/Operators/InvalidBinaryOperationRule.php
staabm: ok, hopefully I got it right now. my understanding about the expected results about the divisor are
- we should not error on benevolent unions containing float (no matter the `$checkBenevolentUnionTypes` flag)
- we should not error on float
- we should error on unions with int|float
- we should error on benevolent unions with non-float but number-castable types respecting `$checkBenevolentUnionTypes` (e.g. `(int|true)`)

I tried endless variants of how to represent this condition via RuleLevelHelper callback, but couldn't find one which works for all this conditions

VincentLanglet:
https://github.com/phpstan/phpstan/issues/11090 was closed

PR #3420, opened 2024-09-08
More precise mixed-type subtraction

staabm:
realized this nit while looking more into division related types

### DIFF
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -506,7 +506,7 @@ class MixedType implements CompoundType, SubtractableType
 			new ConstantIntegerType(0),
 			new ConstantArrayType([], []),
 			new StringType(),
-			new FloatType(),
+			new FloatType(), // every 0.x float casts to int(0)
 		]);
 		if (
 			$this->subtractedType !== null
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -27,7 +27,9 @@ use PHPStan\Type\Accessory\AccessoryNumericStringType;
 use PHPStan\Type\Accessory\OversizedArrayType;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantFloatType;
 use PHPStan\Type\Constant\ConstantIntegerType;
+use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\Generic\TemplateMixedType;
 use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\Traits\NonGeneralizableTypeTrait;
@@ -526,6 +528,32 @@ class MixedType implements CompoundType, SubtractableType
 
 	public function toString(): Type
 	{
+		if ($this->subtractedType !== null) {
+			$castsToEmptyString = new UnionType([
+				new NullType(),
+				new ConstantBooleanType(false),
+				new ConstantStringType(''),
+			]);
+			if ($this->subtractedType->isSuperTypeOf($castsToEmptyString)->yes()) {
+				$accessories = [
+					new StringType(),
+					new AccessoryNonEmptyStringType(),
+				];
+
+				$castsToZeroString = new UnionType([
+					new ConstantFloatType(0.0),
+					new ConstantStringType('0'),
+					new ConstantIntegerType(0),
+				]);
+				if ($this->subtractedType->isSuperTypeOf($castsToZeroString)->yes()) {
+					$accessories[] = new AccessoryNonFalsyStringType();
+				}
+				return new IntersectionType(
+					$accessories,
+				);
+			}
+		}
+
 		return new StringType();
 	}
 
--- a/tests/PHPStan/Analyser/nsrt/non-empty-string.php
+++ b/tests/PHPStan/Analyser/nsrt/non-empty-string.php
@@ -408,4 +408,40 @@ class MoreNonEmptyStringFunctions
 		assertType('non-empty-string', sprintf($nonEmpty, $s));
 		assertType('non-falsy-string', sprintf($nonFalsy, $s));
 	}
+
+	function subtract($m) {
+		if ($m) {
+			assertType("mixed~(0|0.0|''|'0'|array{}|false|null)", $m);
+			assertType('non-falsy-string', (string) $m);
+		}
+		if ($m != '') {
+			assertType("mixed", $m);
+			assertType('string', (string) $m);
+		}
+		if ($m !== '') {
+			assertType("mixed~''", $m);
+			assertType('string', (string) $m);
+		}
+		if (!is_string($m)) {
+			assertType("mixed~string", $m);
+			assertType('string', (string) $m);
+		}
+
+		if ($m !== true) {
+			assertType("mixed~true", $m);
+			assertType('string', (string) $m);
+		}
+		if ($m !== false) {
+			assertType("mixed~false", $m);
+			assertType('string', (string) $m);
+		}
+		if ($m !== false && $m !== '' && $m !== null) {
+			assertType("mixed~(''|false|null)", $m);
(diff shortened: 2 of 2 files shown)


src/Type/MixedType.php
```diff
@@ -472,8 +474,8 @@ public function toBoolean(): BooleanType
 	public function toNumber(): Type
 	{
 		return new UnionType([
-			$this->toInteger(),
-			$this->toFloat(),
+			new IntegerType(),
+			new FloatType(),
```
staabm: required change to prevent

```
     Internal error: Cannot create PHPStan\Type\UnionType with: int<min, -1>|int<1, max>, float while analysing file                                                            
     /Users/staabm/workspace/phpstan-src/src/PhpDoc/TypeNodeResolver.php                                                                                                        
     Post the following stack trace to https://github.com/phpstan/phpstan/issues/new?template=Bug_report.yaml:                                                                  
     ## /Users/staabm/workspace/phpstan-src/src/Type/UnionType.php(60)                                                                                                          
     #0 /Users/staabm/workspace/phpstan-src/src/Type/UnionType.php(77): PHPStan\Type\UnionType::PHPStan\Type\{closure}()                                                        
     #1 /Users/staabm/workspace/phpstan-src/src/Type/MixedType.php(481): PHPStan\Type\UnionType->__construct(Array)                                                             
     #2 /Users/staabm/workspace/phpstan-src/src/Reflection/InitializerExprTypeResolver.php(698): PHPStan\Type\MixedType->toNumber()                                             
     #3 /Users/staabm/workspace/phpstan-src/src/Analyser/MutatingScope.php(1041):        
```

src/Type/MixedType.php
```diff
@@ -484,6 +486,15 @@ public function toAbsoluteNumber(): Type
 
 	public function toInteger(): Type
 	{
+		if ($this->subtractedType !== null) {
+			if ($this->subtractedType->isSuperTypeOf(new ConstantIntegerType(0))->yes()) {
```
staabm: I went with this special case for `0`, because a non-zero int can later on easily be turned in more precise types (like a non-falsy-string etc.)

additionally I think its pretty common to subtract 0 from a given int.

src/Type/MixedType.php
ondrejmirtes: More correct would be to utilize `TypeCombinator::union()` and keep the original `->to*` calls.

src/Type/MixedType.php
ondrejmirtes: Wrong logic. For example: if you don't subtract strings, `toInteger` could still be `0` because strings include `'0'`.

src/Type/MixedType.php
```diff
@@ -494,6 +505,21 @@ public function toFloat(): Type
 
 	public function toString(): Type
 	{
+		if ($this->subtractedType !== null) {
+			if ($this->subtractedType->isSuperTypeOf(new ConstantStringType(''))->yes()) {
```
ondrejmirtes: Wrong logic still.

ondrejmirtes (changes requested):

staabm:
after a break and thinking about your feedback, I came to a new implementation idea :).

tests/PHPStan/Analyser/nsrt/integer-range-types.php
```diff
@@ -446,3 +446,36 @@ public function zeroIssues($positive, $negative)
...
+		assertType('0', (int) $m);
+	}
+	if ($m !== true) {
+		assertType("mixed~true", $m);
+		assertType('int', (int) $m);
+	}
+
+	if ($m != false) {
+		assertType("mixed~0|0.0|''|'0'|array{}|false|null", $m);
+		assertType('int<min, -1>|int<1, max>', (int) $m);
```
staabm: running [my new tool](https://x.com/markusstaab/status/1834140685059977219) thru this test, told me that `(int) $m` can still be zero at this point:

https://3v4l.org/AiaKX

we already knew that [because ondrey told us](https://github.com/phpstan/phpstan-src/pull/3420#discussion_r1749890043) - but as of now I can find this errors myself without wasting his time

staabm:
just reduced this PR to `mixed->toString()` and will to the `toInteger` part in a separate PR

src/Type/MixedType.php
```diff
@@ -494,6 +497,32 @@ public function toFloat(): Type
...
+				$castsToZeroString = new UnionType([
+					new ConstantFloatType(0.0),
+					new ConstantStringType('0'),
+					new ConstantIntegerType(0),
+				]);
+				if ($this->subtractedType->isSuperTypeOf($castsToZeroString)->yes()) {
+					$accessories[] = new AccessoryNonFalsyStringType();
+				}
+				return new IntersectionType(
+					$accessories,
```
ondrejmirtes, line 552: I think I'm missing floats here. See https://3v4l.org/7kgpL

ondrejmirtes (changes requested):

src/Type/MixedType.php
staabm: maybe I missed your point, but float `0.0` is substracted in line 544

ondrejmirtes: I was caught off guard a bit because this method (https://github.com/phpstan/phpstan-src/pull/3434/files#diff-23a89368c86f55bda583e138270384edec06835b889677bb17fc84e581e36dfcR494) mentions whole `FloatType` and this one mentions just constant `0.0`, is there a difference on purpose or does it even matter?

staabm: the difference is, that there is a bunch of float numbers which cast to int `0`, but only a single one which casts to string `'0'`

https://3v4l.org/Q1cva

ondrejmirtes: Oh right :)

ondrejmirtes:
Thank you!

src/Type/MixedType.php
```diff
@@ -526,6 +528,32 @@ public function toFloat(): Type
 
 	public function toString(): Type
 	{
+		if ($this->subtractedType !== null) {
+			$castsToEmptyString = new UnionType([
```
VincentLanglet, line 532: In the same way there `AllowedArrayKeysTypes::getType` I dunno if it would be interesting to introduce a class/static methods for
- castsToZero
- castsToEmptyString
- castsToZeroString
- ...

staabm: do you see a use-case where we need the same implementation somewhere else?

VincentLanglet: No indeed, there is all the following union which is part similar but always different so far

NullType::getSmallerOrEqualType
```
		return new UnionType([
			new NullType(),
			new ConstantBooleanType(false),
			new ConstantIntegerType(0),
			new ConstantFloatType(0.0),
			new ConstantStringType(''),
			new ConstantArrayType([], []),
		]);
```

NullType::getGreaterType
```
return new MixedType(false, new UnionType([
			new NullType(),
			new ConstantBooleanType(false),
			new ConstantIntegerType(0),
			new ConstantFloatType(0.0),
			new ConstantStringType(''),
			new ConstantArrayType([], []),
		]));
```

```
$castsToZero = new UnionType([
			new NullType(),
			new ConstantBooleanType(false),
			new ConstantIntegerType(0),
			new ConstantArrayType([], []),
			new StringType(),
			new FloatType(), // every 0.x float casts to int(0)
		]);
```

```
$castsToEmptyString = new UnionType([
				new NullType(),
				new ConstantBooleanType(false),
				new ConstantStringType(''),
			]);
```

```
$castsToZeroString = new UnionType([
					new ConstantFloatType(0.0),
					new ConstantStringType('0'),
					new ConstantIntegerType(0),
				]);
```

StaticTypeFactory::falsey
```
$falsey = new UnionType([
				new NullType(),
				new ConstantBooleanType(false),
				new ConstantIntegerType(0),
				new ConstantFloatType(0.0),
				new ConstantStringType(''),
				new ConstantStringType('0'),
				new ConstantArrayType([], []),
			]);
```

PR #3421, opened 2024-09-08
Fix const bool subtraction from mixed

staabm:


src/Type/MixedType.php
```diff
@@ -462,8 +462,13 @@ function () use ($level): string {
 
 	public function toBoolean(): BooleanType
 	{
-		if ($this->subtractedType !== null && StaticTypeFactory::falsey()->equals($this->subtractedType)) {
```
staabm: [this never worked](https://phpstan.org/r/0667aa34-946e-42e4-b088-c0abc2927092) because `mixed->equals($x)` always returns false given `$x` is non-mixed

see https://github.com/phpstan/phpstan-src/blob/2c8ab048d91378718116bd42a5e84bf68ed20dac/src/Type/MixedType.php#L285-L291

src/Type/MixedType.php
```diff
@@ -462,8 +462,13 @@ function () use ($level): string {
...
-			return new ConstantBooleanType(true);
+		if ($this->subtractedType !== null) {
+			if ($this->subtractedType->isSuperTypeOf(new ConstantBooleanType(false))->yes()) {
```
ondrejmirtes: This logic is wrong. For example if subtracted type is `bool`, this condition will pass and `toBoolean` will return `true`.

Also many things cast to true or false, making this also wrong.

Here you need something more like `if ($this->subtractedType->toBoolean()->isFalse()->yes())`.

ondrejmirtes (changes requested):

src/Type/MixedType.php
staabm: great catch!

staabm:
suprisingly its more complicated then initially expected. I think its in better shape now.

src/Type/MixedType.php
```diff
@@ -462,8 +462,16 @@ function () use ($level): string {
...
-			return new ConstantBooleanType(true);
+		if ($this->subtractedType !== null) {
+			if (StaticTypeFactory::falsey()->equals($this->subtractedType)) {
+				return new ConstantBooleanType(true);
+			}
+			if ((new ConstantBooleanType(false))->isSuperTypeOf($this->subtractedType)->yes()) {
+				return new ConstantBooleanType(true);
+			}
+			if ((new ConstantBooleanType(true))->isSuperTypeOf($this->subtractedType)->yes()) {
+				return new ConstantBooleanType(false);
```
ondrejmirtes: I still feel this logic is wrong.

1) Let's say we subtract more than what's just falsy. Falsy is `0|0.0|''|'0'|array{}|false|null`. But if we subtract more than that, like `int|0.0|''|'0'|array{}|false|null` or `0|0.0|''|'0'|array{}|bool|null`, the result should still be the same.
2) `(new ConstantBooleanType(false))->isSuperTypeOf($this->subtractedType)->yes()` is also wrong. This represents `mixed~false`. But something from `mixed` could still be falsy.

ondrejmirtes (changes requested):

src/Type/MixedType.php
staabm: after more iterations, I think it turns out I am chasing ghosts and there is no bug.

the tests I came up with now don't fail any longer before this PR.
you may decide whether the tests are worth merging or we just close here, I guess?

tests/PHPStan/Analyser/nsrt/mixed-subtract.php
```diff
@@ -0,0 +1,56 @@
...
+		}
+	}
+
+	if (!$m || is_int($m)) {
+		assertType("0.0|''|'0'|array{}|int|false|null", $m);
+		assertType('bool', (bool) $m);
+	}
+
+	if ($m !== $moreThenFalsy) {
+		assertType('mixed', $m);
```
staabm, line 53: seems the substraction logic atm cannot deal with "subtract more then falsy", as the type stays `mixed`.
(I think its a different unrelated bug)

ondrejmirtes:
Thank you!

PR #3424, opened 2024-09-09
Prevent warning in range() on php 7.x

staabm:
given invalid parameters the `range()` function can emit warnings, as I recently experienced in https://github.com/phpstan/phpstan-src/pull/3419

see https://3v4l.org/nYbAo

there are a few cases which only emit warnings
see https://www.php.net/manual/en/function.range.php

grafik


ondrejmirtes:
Thank you!

PR #3425, opened 2024-09-09
Fix conditional types in array_map() return value

rvanvelzen:
Fix phpstan/phpstan#10715. Alternative to #3411 to handle conditional types correctly.

### DIFF
--- a/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
@@ -2,9 +2,11 @@
 
 namespace PHPStan\Type\Php;
 
+use PhpParser\Node;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you!

VincentLanglet:
Hi, this fix is definitely better since it seems to fix more issues than mine.

I feel like 
- https://github.com/phpstan/phpstan/issues/11056 is solved too.
- maybe https://github.com/phpstan/phpstan/issues/10685 (at least https://phpstan.org/r/870d442a-918b-4ea8-8181-afe38d8faae2)
- part of https://github.com/phpstan/phpstan/issues/9224 (it's solved with `array_map('abs', $arr)` but not `array_map(abs(...), $arr)`)

Am I wrong ?

Do you plan to write non-regression test about these or do you want me to do it @rvanvelzen ?

rvanvelzen:
@VincentLanglet those issues seem to be solved as well indeed. Feel free to write tests for them 👍

VincentLanglet:
> @VincentLanglet those issues seem to be solved as well indeed. Feel free to write tests for them 👍

Sure I'm on it.

To close https://github.com/phpstan/phpstan/issues/9224 there is still an issue to solve.
Maybe you'll know why we have a different behavior (and how to improve your fix) in https://phpstan.org/r/e283bf82-214a-4115-a3b7-2c34b1d8b2cf

rvanvelzen:
@VincentLanglet the final issue there was in fact something completely different. I created #3433 to resolve it.

PR #3427, opened 2024-09-10
Add non regression test for 11056

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11056

Solved by https://github.com/phpstan/phpstan-src/pull/3425


ondrejmirtes:
Thank you.

PR #3431, opened 2024-09-10
Fix false positive when type casting in IF statement

staabm:
closes https://github.com/phpstan/phpstan/issues/11674

see https://3v4l.org/NeugB

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/narrow-cast.php
+++ b/tests/PHPStan/Analyser/nsrt/narrow-cast.php
@@ -31,9 +31,9 @@ function doFoo(string $x, array $arr): void {
 /** @param int<-5, 5> $x */
 function castString($x, string $s, bool $b) {
 	if ((string) $x) {
-		assertType('int<-5, -1>|int<1, 5>', $x);
+		assertType('int<-5, 5>', $x);
 	} else {
-		assertType('0', $x);
+		assertType('int<-5, 5>', $x);
 	}
 
 	if ((string) $b) {
@@ -70,7 +70,7 @@ function castInt($x, string $s, bool $b) {
 	}
 
     if ((int) strpos($s, 'xy')) {
-        assertType('non-falsy-string', $s);
+        assertType('string', $s);
     } else {
         assertType('string', $s);
     }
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/nsrt/narrow-cast.php
```diff
@@ -63,6 +63,12 @@ function castInt($x, string $s, bool $b) {
         assertType('false', $b);
     }
 
+	if ((int) $s) {
+		assertType('string', $s); // could be *NEVER*
```
mvorisek: reachable for `1` or `01`, `numeric-string&non-falsy-string`?

staabm: thank you

staabm: we reach ELSE for `'0.1'` which is a numeric string - https://3v4l.org/NeugB

mvorisek: and not for `1.5` - https://3v4l.org/VBPn2 - so `string` is best or something like `string~(''|'0')`?

also notice `1x` - https://3v4l.org/PGs9f

staabm: `string~(''|'0')` is `non-falsy-string`. but I think we are in crazy territory land where `string` should be good enough ;)

tests/PHPStan/Rules/Comparison/ElseIfConstantConditionRuleTest.php
```diff
@@ -134,10 +134,6 @@ public function testBug11674(): void
 				'Elseif condition is always false.',
 				28,
 			],
-			[
-				'Elseif condition is always false.',
-				36,
```
ondrejmirtes: Took me a while but I found the bad boy for which the elseif can be false: https://3v4l.org/p1sON

ondrejmirtes:
I think this logic in TypeSpecifier makes more sense. Before the issue we didn't know that we can't compare all the casts to `!= false`. So it makes sense to separate the conditions like this.

We lost precision in a few cases but now it's up to `==` / `!=` resolution logic in TypeSpecifier to make them more precise again. We don't need to do that now of course.

ondrejmirtes:
Thank you.

staabm:
thank you!

PR #3434, opened 2024-09-12
More precise mixed-type subtraction

staabm:
`Mixed->toInteger()` part of https://github.com/phpstan/phpstan-src/pull/3420

### DIFF
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -487,10 +487,19 @@ class MixedType implements CompoundType, SubtractableType
 
 	public function toInteger(): Type
 	{
+		$castsToZero =  new UnionType([
+			new NullType(),
+			new ConstantBooleanType(false),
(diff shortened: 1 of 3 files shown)


src/Type/MixedType.php
```diff
@@ -484,6 +487,24 @@ public function toAbsoluteNumber(): Type
 
 	public function toInteger(): Type
 	{
+		$castsToZero =  new UnionType([
+			new NullType(),
+			new ConstantBooleanType(false),
+			new ConstantIntegerType(0),
+			new ConstantArrayType([], []),
+			new StringType(),
+			new FloatType()
```
staabm: `0.1` also casts to `0` (found with the new tool)

https://3v4l.org/IcWTK

ondrejmirtes:
Yeah, this is correct :) Thanks!

PR #3436, opened 2024-09-13
Expose Output::isDecorated and Output::isVeryVerbose

ruudk:
This can be used by an ErrorFormatter to determine if `--ansi` or `--no-ansi` was used.

A use case can be to toggle rendering of colors or hyperlinks.

When ansi is not enabled, one might want to render a link differently.

Needed for https://github.com/TicketSwap/phpstan-error-formatter

### DIFF
--- a/src/Command/Output.php
+++ b/src/Command/Output.php
@@ -16,6 +16,8 @@ interface Output
 
 	public function isVerbose(): bool;
 
+	public function isVeryVerbose(): bool;
+
 	public function isDebug(): bool;
 
 	public function isDecorated(): bool;
--- a/src/Command/Output.php
+++ b/src/Command/Output.php
@@ -18,4 +18,6 @@ interface Output
 
 	public function isDebug(): bool;
 
+	public function isDecorated(): bool;
+
 }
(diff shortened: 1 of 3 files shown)


ruudk:
@ondrejmirtes This might be a breaking change for 1.12. Maybe we can add the method to the SymfonyOutput in 1.12 . And add it to the Output interface in 2.x? 

Then in my code, I can do a check if `isDecorated` exists, and use it.

ondrejmirtes:
> This might be a breaking change for 1.12.

I don't think so. No one has a reason to implement a custom Output. Please add the interface to BcUncoveredInterface (a list of interfaces that users can use but not implement).

src/Command/Output.php
```diff
@@ -18,4 +18,6 @@ public function isVerbose(): bool;
 
 	public function isDebug(): bool;
 
+	public function isDecorated(): bool;
```
ondrejmirtes, line 23: Please also add `isVeryVerbose()`, I realized it's needed a few days ago. Thanks :)

ondrejmirtes (changes requested):

src/Command/Output.php
ruudk: Done!

ruudk:
Updated the PR

ondrejmirtes:
Thank you!

PR #3437, opened 2024-09-13
Do not truncate offset key

ruudk:
When you have an array with long key names, it becomes hard / impossible to see which one errors:

> 💡 Offset 'lockedPersonalizati…' (F\Q\C\N) does not accept type F\Q\C\N|null.

I think it would be better to be precise here.

@ondrejmirtes curious if you agree 😊


voku:
Maybe it would be an idea to make this decision available in the configuration, because every project is different and it might be useful sometimes and sometimes not? 🤔

ondrejmirtes:
This is just abou an error tip so it should always be precise and informative. Thank you!

ruudk:
> Maybe it would be an idea to make this decision available in the configuration, because every project is different and it might be useful sometimes and sometimes not? 🤔

@voku Created an feature request for this: https://github.com/phpstan/phpstan/issues/11760

PR #3438, opened 2024-09-13
Introduce lowercase-string

VincentLanglet:


src/Type/Accessory/AccessoryLowercaseStringType.php
```diff
@@ -0,0 +1,378 @@
...
+		return $type->isLiteralString();
+	}
+
+	public function isSubTypeOf(Type $otherType): TrinaryLogic
+	{
+		if ($otherType instanceof UnionType || $otherType instanceof IntersectionType) {
+			return $otherType->isSuperTypeOf($this);
+		}
+
+		return $otherType->isLiteralString()
```
staabm:
```suggestion
		return $otherType->isLowercaseString()
```

src/Type/Php/SubstrDynamicReturnTypeExtension.php
```diff
@@ -37,72 +38,73 @@ public function getTypeFromFunctionCall(
 	): ?Type
 	{
 		$args = $functionCall->getArgs();
-		if (count($args) === 0) {
+		if (count($args) < 2) {
 			return null;
 		}
 
-		if (count($args) >= 2) {
```
staabm: this cleanup regarding count-of-args and the indent change over the whole block could be a separate PR to make this diff here more readable

staabm:
edge-case which came to mind while looking at the PR:

offset-lookup in a lower-case string yields a lower-case string:
https://3v4l.org/1eWRH

src/Type/Php/SubstrDynamicReturnTypeExtension.php
VincentLanglet: I opened https://github.com/phpstan/phpstan-src/pull/3439

VincentLanglet:
There is at least
```
src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
src/Type/Php/StrPadFunctionReturnTypeExtension.php
src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
src/Type/Php/StrTokFunctionReturnTypeExtension.php
```
wich can be improved too, but maybe it's better in another PR.

thg2k:
What's the use case? Is there an issue linked explaining why we need this? Why not also uppercase string?

VincentLanglet:
> Why not also uppercase string?

`lowercase-string` is the only one supported by Psalm so far and codebase more often use lowercase-string than uppercase string. https://psalm.dev/r/c4d35bbe26

You can see it's used a lot in their codebase https://github.com/search?q=repo%3Avimeo%2Fpsalm%20lowercase-string&type=code
But it's also used in some others codebase like https://github.com/search?q=repo%3ARoave%2FBetterReflection%20lowercase-string&type=code (and certainly more).

It's useful
- For case-insensitive mapping/behavior
- When working with case insensitive data, like URL.

I personally work a lot with URL recently, and it's useful to unsure/checks that I always work with lowercase string.

VincentLanglet:
@staabm Any suggestion about the assertion
```
assertType('non-empty-list<lowercase-string>', explode($s, 'foo'));
		assertType('non-empty-list<string>', explode($s, 'FOO'));
```
which is untrue on php 7.3/7.4 ? Is there a way to have condition based on php version ?

staabm:
Use separate test files in the nsrt/ folder and use a different `// lint` comment condition in the first line and `PhpVersion` method for the version check in the impl

VincentLanglet:
I think it will be easier and better to have
```
src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
src/Type/Php/StrPadFunctionReturnTypeExtension.php
src/Type/Php/StrRepeatFunctionReturnTypeExtension.php
src/Type/Php/StrTokFunctionReturnTypeExtension.php
```
in another PR.

staabm:
I think this should include str-concatenation of 2 lowercase strings, because its a super low level thing

VincentLanglet:
> I think this should include str-concatenation of 2 lowercase strings, because its a super low level thing

Indeed, it's supported now.

staabm:
I think we are missing a rule test for e.g. 

```php
/** @param lowercase-string $s */
function doFoo($s) {
  if ($s === 'AB') {}
  if ('AB' === $s) {}
  if ('AB' !== $s) {}
  if ($s === 'ab') {}
  if ($s !== 'ab') {}
  if ($s === 'aBc') {}
  if ($s !== 'aBc') {}
  if ($s === '01') {}
  if ($s === '1e2') {}
  if ($s === MyClass::myConst) {}
  if ($s === A_GLOBAL_CONST) {}
  if (doFoo('hi') === 'AB') {}
}
```

which asserts rule error(s) when comparing constant with lowercase-string.

----

as a followup it also can be useful to test comparisons for string-functions


```php
/** @param lowercase-string $s */
function doFoo($s, int $x, int $y) {
  $part = substr($s, $x, $y);
  if ($part === 'AB) {} // always false
}
```

src/PhpDoc/TypeNodeResolver.php
```diff
@@ -287,10 +290,16 @@ private function resolveIdentifierTypeNode(IdentifierTypeNode $typeNode, NameSco
 				]);
 
 			case 'non-empty-string':
+				return new IntersectionType([
+					new StringType(),
+					new AccessoryNonEmptyStringType(),
+				]);
+
 			case 'non-empty-lowercase-string':
```
staabm, line 298: do we have a test somewhere for this type?
(I think I see it the very first time today)

tests/PHPStan/Type/TypeToPhpDocNodeTest.php
```diff
@@ -217,6 +218,11 @@ public function dataToPhpDocNode(): iterable
 			'literal-string',
 		];
 
+		yield [
```
staabm, line 221: in this test-file we might need a test for `non-empty-lowercase-string`

VincentLanglet: This is typeToPhpDoc and not PhpDocToType.

So you'll get `non-empty-string & lowercase-string`

src/PhpDoc/TypeNodeResolver.php
VincentLanglet: Added in https://github.com/phpstan/phpstan-src/pull/3438/commits/9636e83175427d8050c98e23d3a6807118457951

VincentLanglet:
After this PR, we may want to introduce check for always false result
StrPosRule/StrStartsWithRule/StrContainRule/StrEndswithRule
cf https://phpstan.org/r/49d9f64f-9488-44ca-81e4-8d73767c4084

src/Type/Accessory/AccessoryLowercaseStringType.php
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	public function accepts(Type $type, bool $strictTypes): TrinaryLogic
+	{
+		return $this->acceptsWithReason($type, $strictTypes)->result;
+	}
+
+	public function acceptsWithReason(Type $type, bool $strictTypes): AcceptsResult
+	{
+		if ($type instanceof MixedType) {
```
ondrejmirtes: You copied this from accessory literal string type, but it does not belong here.

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -1129,19 +1129,19 @@ public function dataArrayDestructuring(): array
 				'$fourthStringArrayForeachList',
 			],
 			[
-				'string',
+				'lowercase-string',
```
ondrejmirtes, line 1132: I don't get this change. IMHO it should remain a string.

ondrejmirtes: Or maybe it should be numeric-string.

ondrejmirtes (changes requested):
1) There are no tests about accepting `lowercase-string`. Add a new test method to CallMethodsRuleTest and try passing different things like a constant string, another lowercase-string and general string to a parameter that accepts lowecase-string.
2) This change is quite disruptive as you can see from the tests that needed changing. We should do this only for bleeding edge, similarly to the list type. This means that all the places that currently do `new IntersectionType` and add the new accesory type will need to do `AccessoryLowercaseStringType::intersectWith()` instead. Also, TypeNodeResolver will interpret `lowercase-string` type the new way only with bleeding edge.

src/Type/Accessory/AccessoryNumericStringType.php
```diff
@@ -296,6 +296,11 @@ public function isLiteralString(): TrinaryLogic
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isLowercaseString(): TrinaryLogic
+	{
+		return TrinaryLogic::createMaybe();
```
ondrejmirtes, line 301: I'd say yes.

ondrejmirtes (changes requested):
Also: We need excessive tests in TypeCombinatorTest::dataUnion and dataIntersect about the interaction with other string types.

src/Type/Accessory/AccessoryNumericStringType.php
thg2k: "10E4" is a numeric string not lowercase

VincentLanglet: Yes that is the reason I kept maybe

ondrejmirtes:
Maybe we don't need to do this only for bleeding edgw, just do a better job in VerbosityLevel::getRecommendedLevelByType

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
VincentLanglet: `'2018-12-19'` is technically a lowercase-string.

And `explode($lowercaseString)` is `array<lowercase-string>`.

So every element of the array is precised from `string` to `lowercase-string`.

VincentLanglet:
> Maybe we don't need to do this only for bleeding edgw, just do a better job in VerbosityLevel::getRecommendedLevelByType

Do you have a preferred solution between bleeding edge and changing getRecommendedLevelByType ?

I unsure to understand what is needed to change in `getRecommendedLevelByType`.

ondrejmirtes:
I'd like for `lowercase-string` to be in an error message about not-accepted type only if there's `lowercase-string` on the accepting side.

Please start by showing in a test how `IntersectionType::describe()` behaves with different VerbosityLevel when there's AccessoryLowercaseStringType involved.

I think we could bump it from `value` to `precise`.

And then in `getRecommendedLevelByType` we should look at `$acceptedType` and bump the returned level from `value` to `precise` only if `$acceptingType->isLowecaseString()->yes()`.

After you do that, the impact on existing rules tests should be minimal. And I could merge it for everyone, not just on bleeding edge.

VincentLanglet:
> I'd like for `lowercase-string` to be in an error message about not-accepted type only if there's `lowercase-string` on the accepting side. 
> 
> Please start by showing in a test how `IntersectionType::describe()` behaves with different VerbosityLevel when there's AccessoryLowercaseStringType involved.
> 
> I think we could bump it from `value` to `precise`.
> 
> And then in `getRecommendedLevelByType` we should look at `$acceptedType` and bump the returned level from `value` to `precise` only if `$acceptingType->isLowecaseString()->yes()`.

Ok, so If I understood correctly, it should be something like
https://github.com/phpstan/phpstan-src/pull/3438/commits/1c20bb1e29f62e2da1abfce4a6112140168eb1ea

I need to update some tests, but still I have some issue with error message like
```
Strict comparison using === between string and 'AB' will always evaluate to false.
```
because not every Rules use `VerbosityLevel::getRecommendedLevelByType`, as an example
https://github.com/phpstan/phpstan-src/blob/a42be43ee436bc14fbcdcd39d51d5e583810cf2a/src/Rules/Comparison/StrictComparisonOfDifferentTypesRule.php#L62-L67

So I assume such thing need to be changed first...

VincentLanglet:
For `testParsingDesiredTypeDescription` and `testDesiredTypeDescription` the `VerbosityLevel::value()` is used.

Should I bump it to precise @ondrejmirtes ?

Also, do you want me to extract https://github.com/phpstan/phpstan-src/pull/3438/commits/053e6cd0357f0cd853fb3d3be26af422924ed29d in another PR first ? I dunno if more rule will need the same change.

@staabm You're the one who suggested tests for this rule (https://github.com/phpstan/phpstan-src/pull/3438#issuecomment-2350876930) do you have more rule with lowercase-string in mind ?

src/Rules/Comparison/StrictComparisonOfDifferentTypesRule.php
```diff
@@ -62,8 +62,8 @@ public function processNode(Node $node, Scope $scope): array
 				$addTip(RuleErrorBuilder::message(sprintf(
 					'Strict comparison using %s between %s and %s will always evaluate to false.',
 					$node->getOperatorSigil(),
-					$leftType->describe(VerbosityLevel::value()),
-					$rightType->describe(VerbosityLevel::value()),
+					$leftType->describe(VerbosityLevel::getRecommendedLevelByType($leftType, $rightType)),
```
ondrejmirtes: I feel like this is a breaking change. The difference: When you have `1` and `'foo'`, before it used these exact scalar values. Now it's just going to say `int` and `string`.

ondrejmirtes: Just revert the change in this rule and we'll see what fails.

src/Type/VerbosityLevel.php
```diff
@@ -156,6 +168,8 @@ public static function getRecommendedLevelByType(Type $acceptingType, ?Type $acc
 
 		/** @var bool $moreVerbose */
 		$moreVerbose = false;
+		/** @var bool $veryVerbose */
+		$veryVerbose = false;
 		TypeTraverser::map($acceptedType, $moreVerboseCallback);
 
 		return $moreVerbose ? self::value() : self::typeOnly();
```
ondrejmirtes, line 179: This line should also check for veryVerbose and return precise based on that

ondrejmirtes (changes requested):
This is in a very good shape and close to merging. I feel like we don't need to update `testParsingDesiredTypeDescription` and `testDesiredTypeDescription`. Instead, add a test to `TypeToPhpDocNodeTest`.

VincentLanglet:
> This is in a very good shape and close to merging. 

Thanks for the kind works. I wasn't sure about it, it motivates a lot.

> I feel like we don't need to update `testParsingDesiredTypeDescription` and `testDesiredTypeDescription`. Instead, add a test to `TypeToPhpDocNodeTest`.

I removed the test in `testParsingDesiredTypeDescription` and `testDesiredTypeDescription`.
I already added one in `TypeToPhpDocNodeTest`
```
yield [
			new IntersectionType([new StringType(), new AccessoryLowercaseStringType()]),
			'lowercase-string',
		];
```

> Just revert the change in this rule and we'll see what fails.

I reverted https://github.com/phpstan/phpstan-src/pull/3438#discussion_r1776502146, you can now see the failure (cf https://github.com/phpstan/phpstan-src/actions/runs/11049021257/job/30693708332?pr=3438) ; since `StrictComparisonOfDifferentTypesRule` use `VerbosityLevel::value()`, the error message are like
```
Strict comparison using === between string and 'AB' will always evaluate to false.
```
which are not helping, instead of
```
Strict comparison using === between lowercase-string and 'AB' will always evaluate to false.
```

ondrejmirtes:
See https://github.com/phpstan/phpstan-src/pull/3438/commits/7b1ade2632172a423550f357761904bdb5b8d75d

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

ondrejmirtes:
Now you can tackle https://github.com/phpstan/phpstan/issues/7703 with `Type::isNominalString(string $name): TrinaryLogic` :) That would be awesome.

staabm:
Excellent!

PR #3439, opened 2024-09-13
Simplify SubstrDynamicReturnTypeExtension

VincentLanglet:
Extracted from https://github.com/phpstan/phpstan-src/pull/3438

### DIFF
--- a/src/Type/Php/SubstrDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SubstrDynamicReturnTypeExtension.php
@@ -37,71 +37,69 @@ final class SubstrDynamicReturnTypeExtension implements DynamicFunctionReturnTyp
 	): ?Type
 	{
 		$args = $functionCall->getArgs();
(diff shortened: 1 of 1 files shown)


src/Type/Php/SubstrDynamicReturnTypeExtension.php
```diff
@@ -37,74 +37,70 @@ public function getTypeFromFunctionCall(
...
 				]);
+
 			}
+			return new IntersectionType([
+				new StringType(),
+				new AccessoryNonEmptyStringType(),
+			]);
 		}
-
-		return null;
```
staabm, line 107: the missing `return null` breaks the build

VincentLanglet: Fixed

staabm (approved):

ondrejmirtes:
Thank you.

PR #3441, opened 2024-09-15
Fix string types sorting

VincentLanglet:
Needed for https://github.com/phpstan/phpstan-src/pull/3438#pullrequestreview-2305028004
or testUnion and testUnionReversed doesn't give the same result.

As seen in
https://phpstan.org/r/f9b06e12-01e6-45c2-92dc-2fcd514a5dbb
PHPStan is not consistent about union of strings.

I'm not familiar with the UnionTypeHelper method, so I'm not sure if 
```
if ($a->isString()->yes() && $b->isString()->yes()) {
				return self::compareStrings($a->describe(VerbosityLevel::value()), $b->describe(VerbosityLevel::value()));
			}
```
is the best fix or if something more generic like
```
instanceof IntersectionType
```
should be preferred.

### DIFF
--- a/src/Type/UnionTypeHelper.php
+++ b/src/Type/UnionTypeHelper.php
@@ -116,6 +116,10 @@ final class UnionTypeHelper
 				return self::compareStrings($a->describe(VerbosityLevel::value()), $b->describe(VerbosityLevel::value()));
 			}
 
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you!

PR #3444, opened 2024-09-16
RegexArrayShapeMatcher - Fix matching literal dot character

staabm:
closes https://github.com/phpstan/phpstan/issues/11699

PR #3445, opened 2024-09-16
Added regression test

staabm:
added another test in the context of https://github.com/phpstan/phpstan-src/pull/3444


ondrejmirtes:
Thank you!

PR #3447, opened 2024-09-16
`range()` with float step should return an array of floats

staabm:
closes https://github.com/phpstan/phpstan/issues/11692

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-11692.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-11692.php
@@ -4,12 +4,16 @@ namespace Bug11692;
 
 use function PHPStan\Testing\assertType;
 
-function doFoo(int $i, float $f): void {
-	assertType('non-empty-list<float|int>', range(1, 9, .01));
+/** @param int|float $floatOrInt */
+function doFoo(int $i, float $f, $floatOrInt): void {
+	assertType('non-empty-list<float>', range(1, 9, .01));
 	assertType('array{1, 4, 7}', range(1, 9, 3));
 
-	assertType('non-empty-list<float|int>', range(1, 9999, .01));
+	assertType('non-empty-list<float>', range(1, 9999, .01));
 	assertType('non-empty-list<int<1, 9999>>', range(1, 9999, 3));
+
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/nsrt/bug-11692.php
```diff
@@ -0,0 +1,20 @@
...
+	assertType('array{1, 4, 7}', range(1, 9, 3));
+
+	assertType('non-empty-list<float|int>', range(1, 9999, .01));
+	assertType('non-empty-list<int<1, 9999>>', range(1, 9999, 3));
+
+	assertType('list<int>', range(1, 3, $i));
+	assertType('list<float|int>', range(1, 3, $f));
+
+	assertType('list<int>', range(1, 9999, $i));
+	assertType('list<float|int>', range(1, 9999, $f));
```
mvorisek, line 22: The type depends on PHP version - https://3v4l.org/vIjsa

if step could be intepreted as int, int is used

and if either start or stop (or step after the transformation to int described above) is float, the type is always `list<float>`

staabm: while I agree with your sentiment I think the extension is already too complicated because it tries to be too smart for the value it produces (and I am pretty sure the already existing logic contains more bugs then the one this PR fixes)

tests/PHPStan/Analyser/nsrt/bug-11692.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Bug11692;
+
+use function PHPStan\Testing\assertType;
+
+function doFoo(int $i, float $f): void {
+	assertType('non-empty-list<float|int>', range(1, 9, .01));
```
ondrejmirtes: This should be `non-empty-list<float>`.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/bug-11692.php
staabm: fixed

ondrejmirtes:
Thank you!

PR #3450, opened 2024-09-17
Fix FileExcluder on windows

staabm:
fixes 1 of the test errors. one remaining but I am not sure yet whether its a problem in the test or the impl.

underlying problem was

```php
<?php

var_dump(fnmatch(
	'C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\File\*',
	'C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\File/data/excluded-file.php',
	FNM_NOESCAPE | FNM_CASEFOLD
));

var_dump(fnmatch(
	'C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\File\*',
	'C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\File\data\excluded-file.php',
	FNM_NOESCAPE | FNM_CASEFOLD
));
```

see https://3v4l.org/WAAkK

### DIFF
--- a/src/File/FileExcluder.php
+++ b/src/File/FileExcluder.php
@@ -50,7 +50,7 @@ final class FileExcluder
 	 * @param string[] $analyseExcludes
 	 */
 	public function __construct(
-		FileHelper $fileHelper,
+		private FileHelper $fileHelper,
 		array $analyseExcludes,
 	)
 	{
@@ -89,6 +89,8 @@ final class FileExcluder
 
 	public function isExcludedFromAnalysing(string $file): bool
 	{
+		$file = $this->fileHelper->normalizePath($file);
+
 		foreach ($this->literalAnalyseExcludes as $exclude) {
 			if (str_starts_with($file, $exclude)) {
 				return true;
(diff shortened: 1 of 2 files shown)


staabm:
the 1 remaing error is related to

```php
<?php

var_dump(fnmatch(
	'tests\PHPStan\File\data\*',
	'C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\File\data\parse-error.php',
	FNM_NOESCAPE | FNM_CASEFOLD
));
```

for this test-case: https://github.com/phpstan/phpstan-src/blob/f0b3d52184aa22e81c53ba6762c073f78be8374c/tests/PHPStan/File/FileExcluderTest.php#L65-L69

ondrejmirtes:
The remaining error should be solved by changing the excluded path to `*\tests\PHPStan\File\data\*`. That's generally how it should be configured.

ondrejmirtes:
Awesome, thank you for your help!

PR #3451, opened 2024-09-17
Fix duplicate paths in composerAutoloaderProjectPaths

staabm:
closes https://github.com/phpstan/phpstan/issues/11706

### DIFF
--- a/bin/phpstan
+++ b/bin/phpstan
@@ -159,7 +159,9 @@ use Symfony\Component\Console\Helper\ProgressBar;
 	$application->setDefaultCommand('analyse');
 	ProgressBar::setFormatDefinition('file_download', ' [%bar%] %percent:3s%% %fileSize%');
 
-	$composerAutoloaderProjectPaths = array_map('realpath', $composerAutoloaderProjectPaths);
+	$composerAutoloaderProjectPaths = array_map(function(string $s): string {
+		return str_replace(DIRECTORY_SEPARATOR, '/', $s);
+	}, $composerAutoloaderProjectPaths);
 	$reversedComposerAutoloaderProjectPaths = array_values(array_unique(array_reverse($composerAutoloaderProjectPaths)));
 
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I'd rather not call realpath, too much logic, can fail. You can normalize the paths with DIRECTORY_SEPARATOR.

ondrejmirtes:
Thank you. This should lead to performance improvements as well.

staabm:
I measured the before/after stats of this patch. 

I can see a ~5-7% faster end-to-end analysis time after the patch with a primed cache.
with a empty cache I can see a 3-4% improvement.

staabm:
@ondrejmirtes looking at the result cache, I can see `\PHPStan\Analyser\Error` beeing serialized with a few duplicated properties:

```
    \PHPStan\Analyser\Error::__set_state(array(
       'message' => 'Call to method isValidPost() of deprecated class CsrfToken:
use DI injected CsrfTokenAbstract instead',
       'file' => 'C:\\dvl\\Workspace\\kunzmann\\app\\portal\\controllers\\ServiceController.php',
       'line' => 145,
       'canBeIgnored' => true,
       'filePath' => 'C:\\dvl\\Workspace\\kunzmann\\app\\portal\\controllers\\ServiceController.php',
       'traitFilePath' => NULL,
       'tip' => NULL,
       'nodeLine' => 145,
       'nodeType' => 'PhpParser\\Node\\Expr\\StaticCall',
       'identifier' => 'staticMethod.deprecatedClass',
       'metadata' => 
      array (
      ),
```

would you accept a patch, which e.g. would prevent serializing `'file'` and `'filePath` in case they contain 1:1 the identical string?
(to reduce the amount of data in the result cache file)

staabm:
Found out that in our projects one big "offender" of result-cache bloat was `rectory/type-perfect` which gets fixed with https://github.com/rectorphp/type-perfect/pull/44

ondrejmirtes:
> would you accept a patch

No, not worth it.

PR #3453, opened 2024-09-19
Truthy `isset($arr[$k])` should narrow `$k`

staabm:
closes https://github.com/phpstan/phpstan/issues/11716
closes https://github.com/phpstan/phpstan/issues/8559

see https://www.php.net/manual/en/language.types.array.php

grafik

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -820,14 +820,13 @@ final class TypeSpecifier
 						);
 					} elseif ($var->dim instanceof Expr\Variable) {
 						$varType = $scope->getType($var->var);
-						if (!$varType->isIterableAtLeastOnce()->no()) {
+						if ($varType->isArray()->yes() && !$varType->isIterableAtLeastOnce()->no()) {
 							$varIterableKeyType = $varType->getIterableKeyType();
 
 							if ($varIterableKeyType->isConstantScalarValue()->yes()) {
-								$varIterableKeyType = TypeCombinator::union(
+								$narrowedKey = TypeCombinator::union(
 									$varIterableKeyType,
 									TypeCombinator::remove($varIterableKeyType->toString(), new ConstantStringType('')),
-									$varIterableKeyType->toFloat(),
 								);
 
 								$zero = new UnionType([
@@ -835,9 +834,9 @@ final class TypeSpecifier
 									new ConstantIntegerType(0),
 								]);
 								if (!$varIterableKeyType->isSuperTypeOf($zero)->no()) {
-									$varIterableKeyType = TypeCombinator::union(
-										$varIterableKeyType,
-										new ConstantBooleanType(false)
+									$narrowedKey = TypeCombinator::union(
+										$narrowedKey,
+										new ConstantBooleanType(false),
 									);
 								}
 
@@ -846,20 +845,24 @@ final class TypeSpecifier
 									new ConstantIntegerType(1),
 								]);
 								if (!$varIterableKeyType->isSuperTypeOf($one)->no()) {
-									$varIterableKeyType = TypeCombinator::union(
-										$varIterableKeyType,
-										new ConstantBooleanType(true)
+									$narrowedKey = TypeCombinator::union(
+										$narrowedKey,
+										new ConstantBooleanType(true),
 									);
 								}
 
 								if (!$varIterableKeyType->isSuperTypeOf(new ConstantStringType(''))->no()) {
-									$varIterableKeyType = TypeCombinator::addNull($varIterableKeyType);
+									$narrowedKey = TypeCombinator::addNull($narrowedKey);
+								}
+
+								if (!$varIterableKeyType->isNumericString()->no() || !$varIterableKeyType->isInteger()->no()) {
+									$narrowedKey = TypeCombinator::union($narrowedKey, new FloatType());
 								}
 
 								$types = $types->unionWith(
 									$this->create(
 										$var->dim,
-										$varIterableKeyType,
+										$narrowedKey,
 										$context,
 										false,
 										$scope,
@@ -875,7 +878,7 @@ final class TypeSpecifier
 											new UnionType([
 												new ArrayType(new MixedType(), new MixedType()),
 												new ObjectWithoutClassType(),
-												new ResourceType()
+												new ResourceType(),
 											]),
 										),
 										$context,
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -818,7 +818,7 @@ final class TypeSpecifier
 								$rootExpr,
 							),
 						);
-					} elseif ($var->dim instanceof Expr\Variable) {
+					} else {
 						$varType = $scope->getType($var->var);
 						if ($varType->isArray()->yes() && !$varType->isIterableAtLeastOnce()->no()) {
 							$varIterableKeyType = $varType->getIterableKeyType();
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -37,7 +37,6 @@ use PHPStan\Type\Accessory\HasOffsetType;
 use PHPStan\Type\Accessory\HasPropertyType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
-use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\ConditionalTypeForParameter;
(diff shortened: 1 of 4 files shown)


staabm:
I am puzzled by the test and its expectations:

https://github.com/phpstan/phpstan-src/blob/cb8f9103f4191f176d1b52cc45f661c3326f194f/tests/PHPStan/Rules/Variables/data/bug-10577.php#L9-L27

I feel the assertion in this test is wrong, see https://3v4l.org/WD7OG
after this PR we now get a `*ERROR*` which is not better either though ;)

tests/PHPStan/Analyser/nsrt/bug-11716.php
```diff
@@ -0,0 +1,112 @@
...
+
+	if (isset($arr[$i])) {
+		assertType('int', $i);
+	} else {
+		assertType('int', $i);
+	}
+	assertType('int', $i);
+
+	if (isset($arr[$s])) {
+		assertType('*NEVER*', $s);
```
staabm: just realized we need to care for php implicit casting rules for array-keys - so this is wrong atm

tests/PHPStan/Analyser/nsrt/bug-11716.php
```diff
@@ -40,7 +40,7 @@ public function parse(string $glue): string
  */
 function narrowKey($mixed, string $s, int $i, array $generalArr, array $arr, array $numericKeyArray): void {
 	if (isset($generalArr[$mixed])) {
-		assertType('(int|string)', $mixed); // array-key cast semantics
+		assertType('int|string', $mixed); // array-key cast semantics
```
mvorisek: what if `$mixed` was/is float like `5.0`?

src/Analyser/TypeSpecifier.php
```diff
@@ -818,6 +818,22 @@ public function specifyTypesInCondition(
...
+								$this->create(
+									$var->dim,
+									$varIterableKeyType,
+									$context,
+									false,
+									$scope,
+									$rootExpr,
+								),
+							);
+						}
```
staabm: the more iterations I do on this thing, the more I feel I need a new `Type` method, e.g. `Type->toIssetOffset()`

staabm: memo to me: maybe take inspiration from `array_key_exists`:

https://github.com/phpstan/phpstan-src/blob/e8027e7a59ba4e70925c89df51081902aa577e03/src/Type/Php/ArrayKeyExistsFunctionTypeSpecifyingExtension.php#L61-L93

staabm: `array_key_exists` also needs improvements https://github.com/phpstan/phpstan/issues/11724 :)

tests/PHPStan/Analyser/nsrt/bug-11716.php
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+/**
+ * @param array<int, string> $arr
+ * @param array<1|'2', string> $numericKeyArray
+ */
+function narrowKey($mixed, string $s, int $i, array $generalArr, array $arr, array $numericKeyArray): void {
+	if (isset($generalArr[$mixed])) {
+		assertType('mixed~array|object', $mixed);
```
mvorisek: also `~resource`?

tests/PHPStan/Analyser/nsrt/bug-11716.php
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+/**
+ * @param array<int, string> $arr
+ * @param array<1|'2', string> $numericKeyArray
+ */
+function narrowKey($mixed, string $s, int $i, array $generalArr, array $arr, array $numericKeyArray): void {
+	if (isset($generalArr[$mixed])) {
+		assertType('mixed~array|object|resource', $mixed);
```
mvorisek: Not directly related, but I wonder if substraction should be further and always generalized if possible, `mixed~type` can be always generalized, ie. into `string|int|float|bool|null` here.

staabm: I don't think it would have a functional difference. or do you have something in mind which would improve?

mvorisek: Pure readability of the type.

tests/PHPStan/Analyser/nsrt/bug-11716.php
```diff
@@ -136,3 +136,16 @@ function emptyArrr($mixed, array $arr)
...
+
+function emptyString($mixed)
+{
+    // see https://3v4l.org/XHZdr
+    $arr = ['' => 1, 'a' => 2];
+
+    if (isset($arr[$mixed])) {
+        assertType("''|'a'|null", $mixed);
+    } else {
+        assertType('mixed', $mixed);
```
mvorisek, line 211: `mixed~(''|'a'|null)`

staabm: I guess this comment went on the wrong line (or I have puhsed a change at the same time you did the comment) and therefore it does not make sense atm?

mvorisek: I see it on correct line, it is for falsy context:
```
    $arr = [0 => 0, 1 => 1, 2 => 2];
    if (isset($arr[$mixed])) {
    } else {
        // falsy context
    }
```

if `$arr[$k]` is set (and not `null` because of `isset`), then the consts types in truthy context can be substracted in the falsy context

staabm: This is sealed vs unsealed array territory. I would not do this until we have a clear separation of the 2 types

tests/PHPStan/Analyser/nsrt/bug-11716.php
```diff
@@ -141,9 +133,46 @@ function emptyString($mixed)
...
+    }
+    assertType('mixed', $mixed);
+}
+
+function numericString($mixed)
+{
+    $arr = ['1' => 1, '2' => 2];
+    if (isset($arr[$mixed])) {
+        assertType("1|1.0|2|2.0|'1'|'2'|true", $mixed);
```
mvorisek: https://3v4l.org/Rml9s - we need https://github.com/phpstan/phpstan/issues/6963 or sadly `float` here

tests/PHPStan/Analyser/nsrt/bug-11716.php
mvorisek: Really? If array specifies some non-optional keys, they must be present (in sealed and unsealed array) and can be substracted in the else branch, can't they?

staabm: looking at it again, I think you are right. 

I still don't want to implement that because the PR is already big enough, `isset()` handling already is pretty complicated, and I think its pretty unlikely this falsy-context narrowing would be useful in practice.

src/Analyser/TypeSpecifier.php
```diff
@@ -818,6 +818,60 @@ public function specifyTypesInCondition(
 								$rootExpr,
 							),
 						);
+					} elseif ($var->dim instanceof Expr\Variable) {
```
ondrejmirtes: Doesn't have to be just a variable. Getting rid of `instanceof` like these leads to bugfixes:

* https://github.com/phpstan/phpstan-src/commit/d3a2a92fcd612bf42bbfd19cd3a5625481ff7522?w=1
* https://github.com/phpstan/phpstan-src/commit/00d2caf39514380610899b0a305413f60c1c5830?w=1 (regression tests in a follow-up commit)

ondrejmirtes (changes requested):

staabm:
in a followup I will look into implications for $k in `array_key_exists`

tests/PHPStan/Analyser/nsrt/bug-11716.php
mvorisek: ```!isset($arr[$k])``` is quite common code pattern, can it be impl. at least for constant types excl. float?

ondrejmirtes (changes requested):
I think this is solid enough to merge this :) One more thing: regression test for https://github.com/phpstan/phpstan/issues/8559 please.

tests/PHPStan/Analyser/nsrt/bug-11716.php
staabm: Please open a new issue with test expectations

src/Analyser/TypeSpecifier.php
mvorisek: https://github.com/phpstan/phpstan/issues/5734 should be fixed then too

ondrejmirtes:
Thank you!

PR #3455, opened 2024-09-20
Fix unassigning the last offset in a ConstantArray that is a list

herndlm:
Those keep being lists.

Fixes https://github.com/phpstan/phpstan/issues/11718

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -734,7 +734,9 @@ class ConstantArrayType extends ArrayType implements ConstantType
 					$k++;
 				}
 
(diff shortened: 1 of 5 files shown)


herndlm:
nope, I was overlooking https://3v4l.org/n5E7F

so the array is a list until another element is added again. ouch. might need changes to `setOffsetValueType()` too..

UPDATE: adapted

ondrejmirtes:
Closing because of: https://github.com/phpstan/phpstan/issues/11718#issuecomment-2363791774

herndlm:
ok. the changes here made the "array with last item unset"-is-still-a-list part work while also removing the list when more is assigned. it indeed feels slightly weird and a non-well-defined state though

ondrejmirtes:
We can't do that because if you passed it somewhere else as a list, and the other place appended an element, the assumption would be broken, and PHPStan would think it has a list but in reality it wouldn't be a list.

PR #3456, opened 2024-09-20
add generic types for array_values

schlndh:
Fixes: https://phpstan.org/r/0865ef07-67b3-47a7-9747-2b34554c57f3

[Phpstorm-stubs](https://github.com/JetBrains/phpstorm-stubs/blob/63db92ae3d59bb746cda2ff6bc6f2949473a5ca3/standard/standard_9.php#L101) don't have a generic type for `array_values` (and they wouldn't have the conditional return type anyway).

### DIFF
--- a/stubs/arrayFunctions.stub
+++ b/stubs/arrayFunctions.stub
@@ -17,6 +17,14 @@ function array_reduce(
 ) {}
 
 /**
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #3459, opened 2024-09-21
ArrayDimFetch on class constant array can return null

staabm:
A offset array fetch on a constant array can return `null` if the offset does not exist, see https://3v4l.org/us7UK

triggered by https://github.com/phpstan/phpstan-src/pull/3453#issuecomment-2360493603

closes https://github.com/phpstan/phpstan/issues/10847


staabm:
we now see errors popping up in cases, where the implementation did not verify before whether the offset actually exists before fetching it (at least not in the method local code):

https://github.com/phpstan/phpstan-src/blob/06ab32096388b46b51f586e484823ea148ff536f/src/Type/Php/StrContainingTypeSpecifyingExtension.php#L61-L70

I am not sure this was done on purpose to prevent annoying PHPStan errors?
(or its just an oversight in the implemetations of the 2 extensions and its missing a offset-assuring narrowing?)

staabm:
> I am not sure this was done on purpose to prevent annoying PHPStan errors?
> (or its just an oversight in the implemetations of the 2 extensions and its missing a offset-assuring narrowing?)

just fixed the 2 extensions. as this change also fixes https://github.com/phpstan/phpstan/issues/10847 I think its the right thing todo

staabm:
there is another ClassConstFetch bug for scalar constants:
https://phpstan.org/r/25b31c30-04d0-4b02-bc41-f5c8d47af2dd

fix it here or in a new PR?

VincentLanglet:
> A offset array fetch on a constant array can return `null` if the offset does not exist, see [3v4l.org/us7UK](https://3v4l.org/us7UK)

There is a warning `Warning: Undefined array key "something"`, so it could be considered as an error by PHPStan instead of a valid case which can return NULL.

I feel like this PR doesn't go in the right direction since it basically enforce `reportPossiblyNonexistentConstantArrayOffset: true` on every project https://phpstan.org/config-reference#reportpossiblynonexistentconstantarrayoffset.

According to the doc with `reportPossiblyNonexistentConstantArrayOffset: false`, the code
```
public function doFoo(string $s): int
{
    $a = ['foo' => 1];
    return $a[$s];
}
```
is valid; but it won't with your PR, since it will report doFoo reports int|null.

staabm:
I think this PR is related but different.

I am doing this logic on ClassConstFetch only, which means its about hardcoded constant arrays only. I think it won't affect your example - I will test in a few hours when in the train.

VincentLanglet:
> I think this PR is related but different.
> 
> I am doing this logic on ClassConstFetch only, which means its about hardcoded constant arrays only. I think it won't affect your example - I will test in a few hours when in the train.

I gave an example based on your example https://3v4l.org/us7UK and the PHPStan doc.

But I can give another. https://phpstan.org/r/cf71d857-d4a7-4d98-ac07-4eabf80e1623
IMHO this snippet:
- Should be valid with `reportPossiblyNonexistentConstantArrayOffset: false`
- Should report an error on `reportPossiblyNonexistentConstantArrayOffset: true`

And I assume with your PR, it will report an error because it may returns `int|null` (therefor reportPossiblyNonexistentConstantArrayOffset will become kinda useless)

staabm:
less invasive fix in https://github.com/phpstan/phpstan-src/pull/3460

thanks for your feedback

PR #3460, opened 2024-09-21
Fix ErrorType after ArrayDimFetch

staabm:
another try at fixing `ErrorType`s which showed up in https://github.com/phpstan/phpstan-src/pull/3453#issuecomment-2360493603

before this PR: https://phpstan.org/r/1eacbfc8-49e2-4b6e-ba01-ef56aa2b8f56

I got to this point with inspiration from https://github.com/phpstan/phpstan-src/commit/2fb66328fdd7119922ce4579f8951ee776320ef6 which added support for maybe-has-offset for constant arrays.
I took the same condition to make sure `getOffsetValueType()` will not return `ErrorType` for this path when `hasOffsetValueType` returns `Maybe`.

----

Goal is, to get a green build in https://github.com/phpstan/phpstan-src/pull/3453

---- 

1st commit is the actual fix
2nd commit is a refactor to de-duplicate logic

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -651,6 +651,10 @@ class ConstantArrayType extends ArrayType implements ConstantType
 
 	public function getOffsetValueType(Type $offsetType): Type
 	{
+		if (count($this->keyTypes) === 0) {
+			return new ErrorType();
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #3461, opened 2024-09-21
Overflowed array has count int<0, max>

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11703


ondrejmirtes:
Thank you!

PR #3467, opened 2024-09-22
Makefile: Disable xdebug in dev tools

staabm:
when working locally often-times xdebug is loaded which slows down some `make` tasks.

lets prepend `XDEBUG_MODE=off` to disable xdebug for commands often used and for the tools that do not already auto-disabling restarts on their own.

`make cs-fix` with this PR (xdebug loaded):
`Time: 32.12 secs; Memory: 156.5MB`

`make cs-fix` before this PR (xdebug loaded):
`Time: 1 mins, 9.14 secs; Memory: 156.5MB`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 build: cs tests phpstan
 
 tests:
-	php vendor/bin/paratest --runner WrapperRunner --no-coverage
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

PR #3471, opened 2024-09-23
More precise `MixedType::toBoolean()` with subtracted type

staabm:
I now understood why my previous test was not failling and was able to apply the suggested fix

followup to https://github.com/phpstan/phpstan-src/pull/3421


tests/PHPStan/Analyser/nsrt/mixed-subtract.php
```diff
@@ -53,4 +53,9 @@ function subtract(mixed $m, $moreThenFalsy) {
 		assertType('mixed', $m);
 		assertType('bool', (bool) $m); // could be true
 	}
+
+	if ($m != 0 && !is_array($m) && $m != null && !is_object($m)) { // subtract more types then falsy
```
staabm, line 57: the test in the above `if` (line 52) is still not able to narrow the type, but I think its related to that we compare 2 variables.

when comparing against literals - as we do in the new test - we at least were able to trigger [the bug](https://phpstan.org/r/2a069d7f-de83-4365-b39b-d2ae8226d55a) in the previous `StaticTypeFactory::falsey()->equals($this->subtractedType)` logic

ondrejmirtes:
Thank you!

PR #3472, opened 2024-09-23
`isset()` narrows string-key in int-keyed-array to numeric-string

staabm:
initial steps to make `isset()` implementation even more similar to `array_key_exists` so we can in the end hopefully share the implementation

commit 1: plain 1:1 refactoring, no implementation changes
commit 2: the fix as the PR topic suggests

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -28,6 +28,7 @@ use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ParametersAcceptorWithPhpDocs;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Reflection\ResolvedFunctionVariant;
+use PHPStan\Rules\Arrays\AllowedArrayKeysTypes;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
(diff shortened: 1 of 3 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -820,47 +821,8 @@ public function specifyTypesInCondition(
...
-									false,
-									new UnionType([
-										new ArrayType(new MixedType(), new MixedType()),
-										new ObjectWithoutClassType(),
-										new ResourceType(),
-									]),
-								);
-							}
-
+						$narrowedKey = AllowedArrayKeysTypes::narrowOffsetKeyType($varType, $dimType);
```
staabm, line 824: as a first step I think its easier to have a static method for this.

after we get more similar with the `array_key_exists` impl we will see whether it makes sense to move this method into the `ConstantScalarType`-interface, as we discussed in dresden

tests/PHPStan/Analyser/nsrt/bug-11716.php
```diff
@@ -59,21 +60,42 @@ function narrowKey($mixed, string $s, int $i, array $generalArr, array $arr): vo
...
+	if (isset($intKeyedArr[$i])) {
 		assertType('int', $i);
 	} else {
 		assertType('int', $i);
 	}
 	assertType('int', $i);
 
-	if (isset($arr[$s])) {
+	if (isset($intKeyedArr[$s])) {
+		assertType('numeric-string', $s);
```
staabm: this assertion is the interessting one, which got more precise with this PR

staabm:
~~I am not sure whether this CI errors are related to my PR, or whether these are triggered by recent BC breaks on the 2.x branch.~~ 

-> it seems to be caused by some github infrastructure thing -> works again

ondrejmirtes:
Thank you!

PR #3473, opened 2024-09-23
Added regression test

staabm:
the error I have expected in https://github.com/phpstan/phpstan/issues/7863 was already there. we did not see it though, because it's hidden behind `checkImplicitMixed`.

closes https://github.com/phpstan/phpstan/issues/7863


ondrejmirtes:
Thank you!

PR #3474, opened 2024-09-23
Fix sprintf() inference for constant values with format-width in pattern

staabm:
we should not infer constant strings 1:1 when sprintf expects a format-width which can add padding characters and what-not.

---

fixes wrong constant value inference
https://phpstan.org/r/ea3080f4-f032-4f2f-8f3b-c5c0f0062da0
https://3v4l.org/HI0pE

We could make this more precise but I fixed only what is blocking https://github.com/phpstan/phpstan-src/pull/3475

### DIFF
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -84,13 +84,13 @@ final class SprintfFunctionDynamicReturnTypeExtension implements DynamicFunction
 			}
 
 			// The printf format is %[argnum$][flags][width][.precision]specifier.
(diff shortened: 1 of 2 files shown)


src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -103,11 +103,13 @@ public function getTypeFromFunctionCall(
 				// if the format string is just a placeholder and specified an argument
 				// of stringy type, then the return value will be of the same type
 				$checkArgType = $scope->getType($args[$checkArg]->value);
-				if ($matches[2] === 's'
+				if (
+					$matches['specifier'] === 's'
+					&& ($checkArgType->isConstantValue()->no() || $matches['width'] === '')
```
staabm, line 108: the actual fix is here. everything else is refactoring/making code easier to read

ondrejmirtes:
Thank you!

PR #3478, opened 2024-09-24
Add PhpVersion param on isSmallerThanOrEqual and isGreaterThanOrEqual

VincentLanglet:
As recommended in https://github.com/phpstan/phpstan/issues/11732#issuecomment-2371803227

I'll do the implementation/fix in another PR, to be sure to not delay the signature update.

### DIFF
--- a/src/Type/CompoundType.php
+++ b/src/Type/CompoundType.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type;
 
+use PHPStan\Php\PhpVersion;
 use PHPStan\TrinaryLogic;
 
 /** @api */
@@ -16,6 +17,6 @@ interface CompoundType extends Type
 
 	public function isGreaterThan(Type $otherType): TrinaryLogic;
 
-	public function isGreaterThanOrEqual(Type $otherType): TrinaryLogic;
+	public function isGreaterThanOrEqual(Type $otherType, PhpVersion $phpVersion): TrinaryLogic;
 
 }
--- a/src/Type/CompoundType.php
+++ b/src/Type/CompoundType.php
@@ -15,7 +15,7 @@ interface CompoundType extends Type
 
 	public function isAcceptedWithReasonBy(Type $acceptingType, bool $strictTypes): AcceptsResult;
 
-	public function isGreaterThan(Type $otherType): TrinaryLogic;
+	public function isGreaterThan(Type $otherType, PhpVersion $phpVersion): TrinaryLogic;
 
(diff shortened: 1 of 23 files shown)


src/Type/CompoundType.php
```diff
@@ -16,6 +17,6 @@ public function isAcceptedWithReasonBy(Type $acceptingType, bool $strictTypes):
 
 	public function isGreaterThan(Type $otherType): TrinaryLogic;
```
ondrejmirtes: What about here?

src/Type/Type.php
```diff
@@ -217,7 +217,7 @@ public function toArrayKey(): Type;
 
 	public function isSmallerThan(Type $otherType): TrinaryLogic;
```
ondrejmirtes: What about here?

ondrejmirtes (changes requested):
Also other potential methods to add it to:

* getSmallerType
* getSmallerOrEqualType
* getGreaterType
* getGreaterOrEqualType

src/Type/CompoundType.php
VincentLanglet: Indeed, `var_dump('' < 0);` is true on PHP 8 but false on PHP 7.4

ondrejmirtes (changes requested):
Mention it in UPGRADING.md in minor breaks section at the bottom. Thanks!

staabm:
I think we need a similar PR for methods which currently use `PhpVersionStaticAccessor::getInstance()` instead of using a `PhpVersion` parameter explicitly?

ondrejmirtes:
That would be awesome!

VincentLanglet:
> I think we need a similar PR for methods which currently use `PhpVersionStaticAccessor::getInstance()` instead of using a `PhpVersion` parameter explicitly?

It's used by 
- StubValidator => seems easy to change
- `ConstantArrayType::findTypeAndMethodNames` => Need to pass `phpVersion` to `isCallable` and `getCallableParametersAcceptors`
- `ConstantStringType::isCallable`

Changing the signature of `isCallable` seems to have a bigger impact since it's used in
- `Type::inferTemplateTypes`
- `Type::isSubTypeOf`
- `Type::isSuperTypeOf`
- `Type::acceptsWithReason`
- `Type::traverseSimultaneously`

And changing those signature lead to too many changes...

ondrejmirtes:
Yeah, I wouldn't want to do this for `Type::isCallable()` / `ConstantArrayType::findTypeAndMethodNames`. It's just covering a quirk/edge-case of pre-PHP 8.0 behaviour.

ondrejmirtes:
Thank you.

PR #3479, opened 2024-09-25
Refactor RegexGroupParser for more immutability and less pass-by-ref

staabm:
less magic, less error prone

### DIFF
--- a/src/Type/Regex/RegexGroupParser.php
+++ b/src/Type/Regex/RegexGroupParser.php
@@ -304,35 +304,25 @@ final class RegexGroupParser
 			return TypeCombinator::union(...$types);
 		}
 
-		$isNonEmpty = TrinaryLogic::createMaybe();
-		$isNonFalsy = TrinaryLogic::createMaybe();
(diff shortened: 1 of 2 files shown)


src/Type/Regex/RegexGroupParser.php
```diff
@@ -376,20 +366,13 @@ private function getRootAlternation(TreeNode $group): ?TreeNode
...
-	 * @param array<string>|null $onlyLiterals
-	 */
 	private function walkGroupAst(
 		TreeNode $ast,
 		bool $inAlternation,
-		TrinaryLogic &$isNonEmpty,
-		TrinaryLogic &$isNonFalsy,
-		TrinaryLogic &$isNumeric,
-		bool &$inOptionalQuantification,
-		?array &$onlyLiterals,
```
staabm, line 389: essentially I moved these by-ref parameters into a separate class without anymore by-ref semantics

ondrejmirtes:
Thank you!

PR #3480, opened 2024-09-25
Optimize memory hungry hot code path

staabm:
Trying a few things

### DIFF
--- a/src/Type/FileTypeMapper.php
+++ b/src/Type/FileTypeMapper.php
@@ -23,6 +23,7 @@ use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\Generic\TemplateTypeVariance;
 use PHPStan\Type\Generic\TemplateTypeVarianceMap;
 use function array_key_exists;
+use function array_key_first;
 use function array_keys;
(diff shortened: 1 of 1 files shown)


src/Type/FileTypeMapper.php
```diff
@@ -130,13 +131,7 @@ private function createResolvedPhpDocBlock(string $phpDocKey, NameScope $nameSco
...
 		$phpDocNode = $this->phpDocStringResolver->resolve($phpDocString);
 		if ($this->resolvedPhpDocBlockCacheCount >= 2048) {
-			$this->resolvedPhpDocBlockCache = array_slice(
-				$this->resolvedPhpDocBlockCache,
-				1,
-				null,
-				true,
-			);
-
+			unset($this->resolvedPhpDocBlockCache[array_key_first($this->resolvedPhpDocBlockCache)]);
```
mvorisek, line 134: As arrays are ordered, removing the first key is quite CPU expensive. The array should be probably reduced by 10-50% at once.

Also I am not exactly sure how only 2048 temporary copied keys can cause dramatic memory peak increase.

PR #3481, opened 2024-09-25
Discover phpstan.php automatically

ruudk:
This makes it a bit easier to use PHP config (which is superior to Neon if you ask me 😊)

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -186,6 +186,7 @@ final class CommandHelper
 				'phpstan.neon.dist',
 				'.phpstan.dist.neon',
 				'phpstan.dist.neon',
+				'phpstan.php',
 			];
 			foreach ($discoverableConfigNames as $discoverableConfigName) {
 				$discoverableConfigFile = $currentWorkingDirectory . DIRECTORY_SEPARATOR . $discoverableConfigName;


ondrejmirtes:
PHP config is not officially supported. I don't want people to use it, I don't like it as much. You're writing the same array as you're writing in .neon.

I see two valid usages for the PHP config:

1) PHP baseline, for performance reasons - documented.
2) Conditional includes like here: https://github.com/phpstan/phpstan-src/blob/2.0.x/build/ignore-by-php-version.neon.php (not documented, but I recommended it to a couple of users)

In the future I'd want to get rid of the .neon format and introduce PHP-based config, but it's going to look differently than it currently does.

ruudk:
There is another real benefit: FQCN references become visible for PHPStorm as usage. Especially handy when working with PHPStan-disallowed extension. And besides that, everyone knows PHP arrays, so it's easier than NEON.

But I get it. Would be great to have a proper PHP config with objects.

ondrejmirtes:
Try the PhpStorm Neon plugin: https://plugins.jetbrains.com/plugin/18387-neon-nette-support

There's also a free one that might be sufficient https://plugins.jetbrains.com/plugin/7060-neon-support

PR #3482, opened 2024-09-25
Implement @pure-unless-callable-is-impure

zonuexe:
Implements the `@pure-unless-callable-is-impure` PHPDoc tag (parser support: phpstan/phpdoc-parser#253, merged): a parameter-level annotation declaring that the function/method is pure **unless** the callable passed to that parameter is impure. Rebuilt on top of `2.2.x`.

## How it works

- The tag (and its `@phpstan-` prefixed alias) is resolved into per-parameter flags, threaded through the reflection layer (`ExtendedParameterReflection::isPureUnlessCallableIsImpureParameter()`), and populated for builtins via `functionMetadata` (`array_filter`, `array_map`, `array_reduce`, the `array_*u*` comparator family, `call_user_func(_array)`, `forward_static_call(_array)`, `preg_replace_callback(_array)`, …) and for `array_reduce` via its stub.
- `SimpleImpurePoint::createFromVariant()` evaluates the purity of arguments passed to flagged parameters:
  - provably **pure** callback (or omitted / `null`) → **no impure point** — the call is treated as pure;
  - provably **impure** callback → the impure point becomes **certain**;
  - unknown callable (opaque `callable` type) → unchanged **possibly-impure** behavior.
- Since the no-effect detection (`NoopExpressionNode`) is driven by "the statement has zero impure points", standalone `array_map($pureCb, $arr);` statements are now reported by `CallToFunctionStatementWithoutSideEffectsRule` with no rule changes.

```php
/** @phpstan-pure */
function doubled(array $arr): array
{
    // no longer false positive "Possibly impure"
    return array_map(static fn (int $x): int => $x * 2, $arr);
}

// now: "Call to function array_map() on a separate line has no effect."
array_map('is_string', $arr);
// still allowed as a statement (impure callback)
array_map(static function ($x) { echo $x; }, $arr);
```

- [x] Add function map
- [x] Add tests (fail-first: `PureFunctionRule`/`PureMethodRule` regressions, `CallToFunctionStatementWithoutSideEffectsRule` cases incl. impure/opaque/by-ref non-regressions)
- [x] Add rules (no new rule needed — the impure-point suppression feeds both the pure-context check and the existing no-effect rule)

refs https://github.com/phpstan/phpstan-src/pull/3106, https://github.com/phpstan/phpstan-src/pull/5912

Closes https://github.com/phpstan/phpstan/issues/11101
Closes https://github.com/phpstan/phpstan/issues/11100


src/PhpDoc/PhpDocNodeResolver.php
```diff
@@ -413,6 +414,24 @@ public function resolveParamImmediatelyInvokedCallable(PhpDocNode $phpDocNode):
 		return $parameters;
 	}
 
+	/**
+	 * @return array<string, bool>
+	 */
+	public function resolveParamPureUnlessCallableIsImpure(PhpDocNode $phpDocNode): array
+	{
+		$parameters = [];
+		// TODO: implement phpstan/phpdoc-parser
```
zonuexe: Will be replaced by https://github.com/phpstan/phpdoc-parser/pull/253

src/Reflection/Native/NativeParameterWithPhpDocsReflection.php
```diff
@@ -22,6 +22,7 @@ public function __construct(
 		private ?Type $outType,
 		private TrinaryLogic $immediatelyInvokedCallable,
 		private ?Type $closureThisType,
+		private bool $pureUnlessCallableIsImpureParameter = false,
```
zonuexe: WIP. `= false` will be removed.

ondrejmirtes:
Just updated phpdoc-parser here so you can use that.

staabm:
@zonuexe any plans to move this forward?

zonuexe:
@staabm Hi, I know this branch is important but it was blocked due to my motivation. I'm going to rebase it on 2.0.x and do the necessary work. If I miss it tonight, I won't be able to start the rest of the work until next week, so it would be quicker for you to take over then.

staabm:
I don't want to put pressure on you. It was not clear to me whether you stopped working on it or not.

I am totally fine if this takes a bit more time - thank you

zonuexe:
No problem, this is a very interesting subject for me too, so I'm happy to work on it because I realized that there are other people who are interested in it. More importantly, I'm always helped by your hard work.

VincentLanglet:
Friendly ping @zonuexe, still interested in this PR ? Do you have time to finish it ?
Do you look for help ? What is missing ?

zonuexe:
@VincentLanglet 
Thank you for the suggestion, sorry for making you wait so long.
I've been working on PHPStan 2.1.18 a few times over the last few weeks, and just finished a PHPStan workshop at a local PHP conference, so I'll get back to that tomorrow.

VincentLanglet:
Friendly ping @zonuexe 

I don't want to put pressure on you. It's just not clear to me whether you stopped working on it or not.
What is missing ?

It will unblock the https://github.com/phpstan/phpdoc-parser/pull/259 pr :)

VincentLanglet:
Maybe I'm wrong, but I start to think this annotation is useless
https://phpstan.org/r/d3af976b-17ef-4ca6-87cb-4a1325622cee

PHPStan already handle cb with immediately/later invoked callable, even for a pure function/method.

So we just need to declare array_map and so as pure.

devbanana:
i'm happy to help contribute towards a solution if it's decided which direction is best.

VincentLanglet:
Hi @devbanana feel free to work on this.

Laybe @zonuexe could explain what he had in mind ; or @ondrejmirtes will have some guidelines

zonuexe:
Sorry for keeping you all waiting so long. I have rebased onto the latest 2.2.x version and added the wiring that was missing from the previous changes.

resources/functionMetadata.php
```diff
@@ -733,43 +733,47 @@
 	'array_diff' => ['hasSideEffects' => false],
 	'array_diff_assoc' => ['hasSideEffects' => false],
 	'array_diff_key' => ['hasSideEffects' => false],
-	'array_diff_uassoc' => ['hasSideEffects' => false],
-	'array_diff_ukey' => ['hasSideEffects' => false],
+	'array_diff_uassoc' => ['pureUnlessCallableIsImpureParameters' => ['key_compare_func' => true]],
```
staabm, line 746: since this file format got more complicated, I think it would make sense to describe at the file-level comment how the shape is.
we could even add a inline `@var` array shape definition right before the huge array.

(I did not verify whether we analyse this big file though)

src/Reflection/Php/PhpClassReflectionExtension.php
```diff
@@ -642,7 +642,7 @@ private function createMethod(
 
 			$isPure = null;
 			if ($this->signatureMapProvider->hasMethodMetadata($declaringClassName, $methodReflection->getName())) {
-				$isPure = !$this->signatureMapProvider->getMethodMetadata($declaringClassName, $methodReflection->getName())['hasSideEffects'];
+				$isPure = !($this->signatureMapProvider->getMethodMetadata($declaringClassName, $methodReflection->getName())['hasSideEffects'] ?? true);
```
staabm: I feel this would be more readable when written in multiple lines

```suggestion
				$hasSideEffects = $this->signatureMapProvider->getMethodMetadata($declaringClassName, $methodReflection->getName())['hasSideEffects'] ?? true;
				$isPure = !$hasSideEffects;
```

otherwise its not 100% clear what this line defaults to

---

alternatively maybe even do it like a few lines below

```
$metadata = $this->signatureMapProvider->getMethodMetadata($declaringClassName, $methodReflection->getName());
$hasSideEffects = ['hasSideEffects'] ?? true;
$isPure = !$hasSideEffects;
 ```

File: src/Reflection/Php/PhpFunctionFromParserNodeReflection.php [hunk 5]
```diff
@@ -177,6 +179,12 @@ public function getParameters(): array
 				$closureThisType = null;
 			}
 
+			if (isset($this->pureUnlessCallableIsImpureParameters[$parameter->var->name])) {
+				$pureUnlessCallableIsImpureParameter = $this->pureUnlessCallableIsImpureParameters[$parameter->var->name];
+			} else {
+				$pureUnlessCallableIsImpureParameter = false;
+			}
```
staabm: `??` instead?

src/Reflection/SignatureMap/NativeFunctionReflectionProvider.php
```diff
@@ -110,18 +110,30 @@ public function findFunctionReflection(string $functionName): ?NativeFunctionRef
...
 					null,
-					array_map(static function (ParameterSignature $parameterSignature) use ($phpDoc, $lowerCasedFunctionName, $allowedConstantsMapProvider): ExtendedNativeParameterReflection {
+					array_map(static function (ParameterSignature $parameterSignature) use ($phpDoc, $lowerCasedFunctionName, $allowedConstantsMapProvider, $pureUnlessCallableIsImpureParameters): ExtendedNativeParameterReflection {
+						$name = $parameterSignature->getName();
 						$type = $parameterSignature->getType();
 
 						$phpDocType = null;
 						$immediatelyInvokedCallable = TrinaryLogic::createMaybe();
 						$closureThisType = null;
+						$pureUnlessCallableIsImpureParameter = isset($pureUnlessCallableIsImpureParameters[$name]) && $pureUnlessCallableIsImpureParameters[$name];
```
staabm: `??`

src/Reflection/SignatureMap/NativeFunctionReflectionProvider.php
```diff
@@ -158,8 +174,8 @@ public function findFunctionReflection(string $functionName): ?NativeFunctionRef
 			}
 		}
 
-		if ($this->signatureMapProvider->hasFunctionMetadata($lowerCasedFunctionName)) {
-			$hasSideEffects = TrinaryLogic::createFromBoolean($this->signatureMapProvider->getFunctionMetadata($lowerCasedFunctionName)['hasSideEffects']);
+		if (isset($functionMetadata['hasSideEffects'])) {
```
staabm: `??`

stubs/arrayFunctions.stub
```diff
@@ -9,6 +9,8 @@
  * @param TReturn $three
  *
  * @return TReturn
+ *
+ * @pure-unless-callable-is-impure $two
```
staabm, line 13: did you check more stubs whether these need `@pure-unless-callable-is-impure`?

staabm:
please also double check issue-bot results.

I love it already. thank you :)

src/Reflection/SignatureMap/NativeFunctionReflectionProvider.php
```diff
@@ -158,11 +174,9 @@ public function findFunctionReflection(string $functionName): ?NativeFunctionRef
...
 		}
 
-		if ($this->signatureMapProvider->hasFunctionMetadata($lowerCasedFunctionName)) {
-			$hasSideEffects = TrinaryLogic::createFromBoolean($this->signatureMapProvider->getFunctionMetadata($lowerCasedFunctionName)['hasSideEffects']);
-		} else {
-			$hasSideEffects = TrinaryLogic::createMaybe();
-		}
+		$hasSideEffects = isset($functionMetadata['hasSideEffects'])
+			? TrinaryLogic::createFromBoolean($functionMetadata['hasSideEffects'])
+			: TrinaryLogic::createMaybe();
```
zonuexe, line 179: @staabm re your `??` suggestion at https://github.com/phpstan/phpstan-src/pull/3482#discussion_r3517629103

For the other `isset(...) && ...` spots I switched to `?? false` as you suggested. I kept an explicit ternary here, though:

```php
$hasSideEffects = isset($functionMetadata['hasSideEffects'])
    ? TrinaryLogic::createFromBoolean($functionMetadata['hasSideEffects'])
    : TrinaryLogic::createMaybe();
```

A plain `??` doesn't fit this one: the absent branch has to become `createMaybe()`, not `createFromBoolean(false)`. `createFromBoolean($functionMetadata['hasSideEffects'] ?? false)` would turn "metadata exists but has no `hasSideEffects` key" — e.g. `array_map()`, which only carries `pureUnlessCallableIsImpureParameters` — into `No` (definitely pure) instead of `Maybe`, silently dropping the impure-callback detection.

zonuexe:
Thanks for the review! Addressed the rest:

- Readability/doc nits (`?? false`, split reads, metadata shape comment + `@var`) — done, in one commit. One exception on the `hasSideEffects` read; left a note inline on why a plain `??` doesn't fit there.
- Re: "did you check more stubs" — audited the callback-accepting builtins, which surfaced a real gap: generic stubs like `array_find()` are reflected via BetterReflection (`PhpFunctionReflection`), whose path only read the `@pure-unless-callable-is-impure` PHPDoc tag and **ignored `functionMetadata.php`** — so `array_find()`'s existing entry was silently a no-op. Fixed that path to honor the metadata too (same source of truth as the native path), which makes `array_find()` actually work and lets me add `array_find_key()`, `array_all()`, `array_any()` (PHP 8.4).
- Non-candidates left as-is: `array_walk*` / `usort` family mutate by-ref; `set_*_handler` / `register_shutdown_function` / `spl_autoload_register` are already `@param-later-invoked-callable`. `preg_replace_callback_array` is pure-modulo-callback but its callable is an `array<string, callable>` the verdict logic doesn't unwrap yet — follow-up rather than scope creep here.

VincentLanglet (approved):

staabm (approved):

src/Reflection/Dummy/ChangedTypeMethodReflection.php
```diff
@@ -168,6 +168,11 @@ public function isPure(): TrinaryLogic
 		return $this->reflection->isPure();
 	}
 
+	public function getPureUnlessCallableIsImpureParameters(): array
+	{
+		return [];
```
staabm:
```suggestion
		return $this->reflection->getPureUnlessCallableIsImpureParameters();
```

(seems we miss a test)

src/Reflection/Php/ClosureCallMethodReflection.php
```diff
@@ -199,6 +199,11 @@ public function isPure(): TrinaryLogic
 		return $this->nativeMethodReflection->isPure();
 	}
 
+	public function getPureUnlessCallableIsImpureParameters(): array
+	{
+		return [];
```
staabm:
```suggestion
		return $this->nativeMethodReflection->getPureUnlessCallableIsImpureParameters();
```

src/Reflection/Php/PhpFunctionFromParserNodeReflection.php
```diff
@@ -334,6 +339,11 @@ public function isPure(): TrinaryLogic
 		return TrinaryLogic::createFromBoolean($this->isPure);
 	}
 
+	public function getPureUnlessCallableIsImpureParameters(): array
+	{
+		return [];
```
staabm: hmm does this miss the impl?

src/Reflection/Type/IntersectionTypeMethodReflection.php
```diff
@@ -214,6 +214,11 @@ public function isPure(): TrinaryLogic
 		return TrinaryLogic::lazyMaxMin($this->methods, static fn (ExtendedMethodReflection $method): TrinaryLogic => $method->isPure());
 	}
 
+	public function getPureUnlessCallableIsImpureParameters(): array
+	{
+		return [];
```
staabm: what about this?

src/Reflection/Type/UnionTypeMethodReflection.php
```diff
@@ -171,6 +171,11 @@ public function isPure(): TrinaryLogic
 		return TrinaryLogic::lazyExtremeIdentity($this->methods, static fn (ExtendedMethodReflection $method): TrinaryLogic => $method->isPure());
 	}
 
+	public function getPureUnlessCallableIsImpureParameters(): array
+	{
+		return [];
```
staabm: and here?

src/Reflection/Dummy/DummyConstructorReflection.php
```diff
@@ -153,6 +153,11 @@ public function isPure(): TrinaryLogic
 		return TrinaryLogic::createYes();
 	}
 
+	public function getPureUnlessCallableIsImpureParameters(): array
```
staabm, line 156: also needs tests with constructor

zonuexe: Good catch. You were right that method support was entirely non-functional, not only those stubs. Two gaps caused it.

- `PhpClassReflectionExtension` read the pure-unless map only from built-in method metadata, never from a userland method's own `@pure-unless-callable-is-impure` tag.
- `PhpMethodReflection::getParameters()` did not pass the flag on to `PhpParameterReflection`, while the function path already did.

I fixed both and made the wrappers you pointed at delegate. I also added a method regression test with pure, impure and opaque callbacks, including a union-typed receiver.

zonuexe:
Follow-up on method support.

- Method-level `@pure-unless-callable-is-impure` now works end to end, at parity with functions. Verified for direct, union-typed and intersection-typed receivers.
- I audited every parameter-construction site for the same flag-dropping pattern. The others are correct, because the `instanceof ExtendedParameterReflection ? ... : new ExtendedDummyParameter(...)` wraps only ever see non-extended parameters, which cannot carry the flag.
- Two paths stay un-threaded on purpose and now carry a comment: `createNativeMethodVariant()`, since no built-in method carries the metadata, and `Closure::call()`'s own parameters, since a closure cannot carry the tag.

staabm:
I think with this PR we should not have a error on line 32 anymore
https://phpstan.org/r/88f8eb07-b4ab-44c2-ad75-1362d55898e6

wdyt?

zonuexe:
@staabm 
You were right, it was not honored for `new`. Turned your playground example into a test and fixed `NewHandler`, which now applies the same pure-unless verdict as function and method calls:

- `new Bar($pureClosure)` in a pure function: no error.
- `new Bar($impureClosure)`: `Impure instantiation of class Bar`.
- `new Bar($opaqueCallable)`: stays `Possibly impure instantiation`.

One open point: mutation testing. Two `TrinaryLogicMutator` mutants on the new verdict branch survive in CI. I could not get them killed there, even though the test does kill them locally (verified by hand-mutation) and the coverage report maps the mutated lines to it. The identical verdict branch in `SimpleImpurePoint::createFromVariant()` is covered fine, so this looks specific to the `new` path. Any idea what I am missing?

staabm:
thank you, this looks really good already.

one more case I just came over while testing:
https://phpstan.org/r/f72a097c-0f82-4710-a3ac-a177fa22fea1

when the callable is "maybe pure/impure", I think `@pure-unless-callable-is-impure` should be pessimistic and report itself as impure. In other words, I guess we miss a error on line 18.

wdyt?

zonuexe:
@staabm 

Turned your playground into a regression test. With `@param pure-callable $pureCallable`, pure-unless resolves as you would want:

 - `new Bar($pureCallable)`: no error, the callback is known pure.
 - `new Bar(rand(0, 1) ? $pureCallable : $callable)`: stays `Possibly impure instantiation`, since the union `pure-callable|callable` might be the impure branch.

`doFoo()` itself is not `@phpstan-pure` here, so nothing inside it is checked for purity anyway; the only reported error is the unrelated missing return type. Added both cases as tests in `pure-unless-callable-is-impure.php`.

staabm (approved):

VincentLanglet (approved):

src/Reflection/ExtendedParameterReflection.php
```diff
@@ -36,4 +36,6 @@ public function getAllowedConstants(): ?ParameterAllowedConstants;
 	 */
 	public function checkAllowedConstants(array $constants): AllowedConstantsResult;
 
+	public function isPureUnlessCallableIsImpureParameter(): bool;
```
staabm: should return trinary.

quoting ondrey:

> I feel like on ExtendedParametersAcceptor, the method isPureUnlessCallableIsImpureParameter should return TrinaryLogic.
> Because you need to correctly set it on unions and intersection methods

tests/PHPStan/Rules/Pure/data/pure-unless-callable-is-impure.php
```diff
@@ -0,0 +1,241 @@
...
+function pureWithOpaqueCallback(array $arr, callable $cb): array
+{
+	return array_map($cb, $arr);
+}
+
+/**
+ * @param callable(int): int $f
+ * @param array<int> $arr
+ * @return array<int>
+ * @pure-unless-callable-is-impure $f
```
staabm, line 43: we need a test which shows that a function which is declared `@pure-unless-callable-is-impure $f` errors about impure side-effects when the impure statement/expression is outside the callable $f

tests/PHPStan/Rules/Pure/data/pure-unless-callable-is-impure.php
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return $result;
+}
+
+/**
+ * @param callable(int): int $f
+ * @param array<int> $arr
+ * @return array<int>
+ * @phpstan-pure-unless-callable-is-impure $f
```
staabm, line 59: If `@pure-unless-callable-is-impure`  is paired with `pure-callable` or `pure-Closure` , that should be a new kind of error (that the PHPdoc tag could be replaced with `@phpstan-pure` essentially

src/Reflection/ExtendedMethodReflection.php
```diff
@@ -69,6 +69,11 @@ public function isBuiltin(): TrinaryLogic|bool;
 	 */
 	public function isPure(): TrinaryLogic;
 
+	/**
+	 * @return array<string, bool>
```
ondrejmirtes: Should still be TrinaryLogic here. When you have UnionMethodReflection or IntersectionMethodReflection (and the methods in question have different PHPDoc tags involved, you need TrinaryLogic to properly model the behaviour). Also add some tests.

ondrejmirtes (changes requested):
One more thing - in MethodSignatureRule, we need to enforce covariance for these methods. Check around `$this->reportMethodPurityOverride`.

If a method has `@phpstan-pure` or `@phpstan-pure-unless-callable-is-impure`, the child method cannot be `@phpstan-impure`.

Also are there same tests around `@phpstan-pure-unless-callable-is-impure` being inherited implicitly? (The child method does not need to mention the parameter but if parent method has it, it should also apply to child. Even when the parameter is renamed.)

zonuexe:
@ondrejmirtes @staabm Both review points are addressed.

**1. TrinaryLogic on the map getter.** `getPureUnlessCallableIsImpureParameters()` now returns `array<string, TrinaryLogic>` across the interfaces and all implementers. `UnionTypeMethodReflection` / `IntersectionTypeMethodReflection` merge their members per parameter (new `MergedPureUnlessCallableIsImpureParameters` helper) so a parameter tagged in some members but not others becomes `Maybe`, and `combineAcceptors()` merges the parameter-level flag with equals-or-Maybe instead of `->or()`, so a union receiver yields a real `Maybe` at the call site. Added union and implicit-inheritance tests (child does not re-declare the tag), including the renamed-parameter case.

**2. Purity covariance.** `MethodSignatureRule` now reports a `@phpstan-impure` child overriding a parent that carries `@pure-unless-callable-is-impure` (`method.impureOverridePureUnlessCallable`), gated by `reportMethodPurityOverride`.

Note: the Mutation Testing job is still red on a few mutants around the verdict/override branches. The tests seem to kill them locally, so I will leave it to you two to decide how to proceed.

ondrejmirtes:
Might be valuable to fix the escaped mutant. It means that either the code logic needs to change, or a test has to be added to fail when the code is changed this way.

zonuexe:
@ondrejmirtes @staabm 
On the Mutation Testing job: I built a local Infection setup with the same custom TrinaryLogicMutator and ran it (scoped to the affected files/suites, single-threaded) against the exact same mutation points. All of the mutants currently shown as escaped in CI come back as killed locally, the only mutant that actually escaped in my run was on an unrelated, pre-existing line. So this looks like something specific to the full CI run rather than these lines being under-tested.

ondrejmirtes:
@zonuexe I don't think you need to set up Infection locally to verify this mutant. The question is: if you modify the file:

```
-			if ($parameter->isPureUnlessCallableIsImpureParameter()->no()) {
+			if (!$parameter->isPureUnlessCallableIsImpureParameter()->yes()) {
```

Do you get a failing test locally or not? If no tests fail the mutant is valid and then it's easy to fix the code/add the test.

zonuexe:
I ran the exact custom `TrinaryLogicMutator` locally to check this properly.

Using the same mutator with `--git-diff-lines` over the changed files, and coverage generated from the `Rules/Pure`, `Rules/Methods` and `Reflection/Callables` suites, every mutant on the changed lines is killed by the test framework: **8 mutants generated, 8 killed, Covered Code MSI 100%**, nothing escaped.

In CI the escaped mutant is a moving target: across the last three runs it was `SimpleImpurePoint:151`, then `NewHandler:263/266`, then `NewHandler:262` - each killed by `PureFunctionRuleTest` locally (verified by hand-applying the `->yes()` <-> `!->no()` mutation). So the tests do cover and kill these mutants; the remaining CI failure looks like coverage attribution in the parallel Infection run rather than a missing test.

I briefly tried a direct unit test plus factoring the verdict branching into one helper, but since that does not change the CI attribution I reverted it to keep the PR focused on the reviewed changes.

ondrejmirtes (approved):

staabm:
Thank you @zonuexe

zonuexe:
I am always helped by the great work of the PHPStan team!

ondrejmirtes:
Yeah, thank you!

And it's documented already. I love these times!

* https://github.com/phpstan/phpstan/pull/14934
* https://github.com/phpstan/phpstan/pull/14933

staabm:
As a followup it would be great to look into phpstan 1st party extension packages and also often used packages in the php ecosystem, which could utilize the new phpdoc

PR #3483, opened 2024-09-25
Change iptcparse return type

ManuelHu:
the inner arrays are always lists, see https://github.com/php/php-src/blob/e035a957237a7c17aa3dc2810d86333c93654c11/ext/standard/iptc.c#L365


ondrejmirtes:
Thank you.

PR #3488, opened 2024-09-26
Fix a few regex class parsing usecases

mvorisek:


src/Type/Regex/RegexExpressionHelper.php
```diff
@@ -88,13 +91,24 @@ public function getPatternModifiers(string $pattern): ?string
 
 	public function removeDelimitersAndModifiers(string $pattern): string
 	{
+		$pattern = ltrim($pattern);
```
mvorisek, line 94: @staabm in `getPatternDelimiter` we do `ltrim` but here it was missing, but tests are not failing - do you know why? However, I belive `ltrim` should be here as we strip the start delimiter using `substr($v, 1, ...`.

src/Type/Regex/RegexExpressionHelper.php
```diff
@@ -88,13 +91,24 @@ public function getPatternModifiers(string $pattern): ?string
...
 
-		return substr($pattern, 1, $endDelimiterPos - 1);
+		$raw = substr($pattern, 1, $endDelimiterPos - 1);
+
+		$startDelimiter = substr($pattern, 0, 1);
+		$endDelimiter = $this->getPatternDelimiter($pattern);
+		assert($endDelimiter !== null);
+
+		$raw = preg_replace('~(?<!\\\)(\\\\\\\)*\K\\\(?=[' . preg_quote($startDelimiter . $endDelimiter, '~') . '])~', '', $raw);
+		assert($raw !== null);
```
mvorisek: @staabm can you please review?

staabm: why is this ugly regex needed? what is it supposed todo?

mvorisek: The regex is beautiful :), it strips escaping `\` before delimiter if not preceded by another (escaped) `\`.

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -768,3 +768,55 @@ function bug11604b (string $string): void {
...
 }
+
+function testLtrimDelimiter (string $string): void {
+	if (preg_match('/(x)/', $string, $matches)) {
+		assertType("array{string, 'x'}", $matches);
+	}
+}
+
+function testUnescapeBackslash (string $string): void {
+	if (preg_match(<<<'EOD'
```
staabm, line 786: is this nowdoc usage important for the tests or can it be simplified?

src/Type/Regex/RegexExpressionHelper.php
staabm: @Seldaek do you have an idea whether this can be achieved in a more readable way?

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
mvorisek: Of course not, but it simplied the human understanding of many backslashes which needs to be otherwise escaped in php strings...

src/Type/Regex/RegexExpressionHelper.php
mvorisek: The regex is not longer needed, regexes are parsed differently.

resources/RegexGrammar.pp
```diff
@@ -42,11 +42,13 @@
 //
 
 // Character classes.
+%token  negative_class_fc_       \[\^(?=\])         -> class_fc
```
staabm, line 46: what does "fc" stand for?

resources/RegexGrammar.pp
```diff
@@ -42,11 +42,13 @@
...
+%token  class_fc_                \[(?=\])           -> class_fc
```
staabm, line 47: Is the indent intentional different in a few of this lines?

resources/RegexGrammar.pp
mvorisek: first character - I know it is hard to understand, but should full/long name be used which will break the indentation?

if yes, feel free to propose changes

resources/RegexGrammar.pp
mvorisek: it looks weird, but I tried to follow the original indentation

resources/RegexGrammar.pp
staabm: maybe at least add a comment at the end of the line

resources/RegexGrammar.pp
```diff
@@ -177,10 +179,14 @@
 
 #class:
     (
-        ::negative_class_:: #negativeclass
+        ::negative_class_fc_:: #negativeclass
```
staabm, line 183:
```suggestion
        ::negative_class_fc_:: #negative class first-char 
```

mvorisek: why? - `#negativeclass` should be there 2x the same as we want 2x the same AST

Seldaek: Yeah this is not a comment, this is a marker so it appears as `negativeclass` in the AST

Seldaek: But adding a comment with // on the %token above could make sense

staabm: ohh I see. I had the impression its a comment.

great to see you here @Seldaek . what be awesome you could review the grammar changes.

resources/RegexGrammar.pp
```diff
@@ -57,8 +59,8 @@
 %token class:literal             \\.|.|\n
 
 // Internal options.
-// See https://www.regular-expressions.info/refmodifiers.html
-%token  internal_option          \(\?([imsxnJUX^]|xx)?-?([imsxnJUX^]|xx)\)
+// See https://www.php.net/manual/en/regexp.reference.internal-options.php
```
Seldaek: I find the other document better as it contains for example `(?^)` which is not in the php docs. Please keep both links at least.

resources/RegexGrammar.pp
mvorisek: added

resources/RegexGrammar.pp
```diff
@@ -88,7 +90,7 @@
 %token  nc:_named_capturing      >                  -> default
 %token  nc:capturing_name        .+?(?=(?<!\\)>)
 %token  non_capturing_           \(\?:
-%token  non_capturing_internal_option \(\?([imsxnJUX^]|xx)?-?([imsxnJUX^]|xx):
+%token  non_capturing_internal_option \(\?[imsxnJUX]*-?[imsxnJUX]+:
```
staabm: why did these change?

resources/RegexGrammar.pp
```diff
@@ -57,8 +59,8 @@
 %token class:literal             \\.|.|\n
 
 // Internal options.
-// See https://www.regular-expressions.info/refmodifiers.html
-%token  internal_option          \(\?([imsxnJUX^]|xx)?-?([imsxnJUX^]|xx)\)
+// See https://www.php.net/manual/en/regexp.reference.internal-options.php
+%token  internal_option          \(\?[imsxnJUX]*-?[imsxnJUX]+\)
```
Seldaek: You removed `(?^)` support, but it is supported by PHP as per https://3v4l.org/BPeFg and regular-expressions.info

Aside from that, I am not sure anymore why I only made it one modifier max and no `+`. I think there was a reason but I cannot find it anymore and it seems to be supported so I guess I am 👍🏻 for the change overall.

Note that when ^ is used, you cannot remove modes as it resets them (see how the first one fails to compile here https://3v4l.org/5Snv2) so probably this should be `\(\?(?:[imsxnJUX]*-?|\^)[imsxnJUX]+\)` to be strict. Or just `\(\?[imsxnJUX]*-?[imsxnJUX^]+\)` to keep it simpler. I don't mind too much as this is a super fringe feature, and regex compilation would fail before anyway if it's invalid so we probably don't need to worry too much about invalid regexes here.

resources/RegexGrammar.pp
mvorisek: see added testcase - `xx` is not PCRE syntax, in theory we still support is as newly multiple options are allowed (per valid PCRE grammar). `^` - I am not sure what it meant, I think it is not supported by PCRE.

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -768,3 +771,125 @@ function bug11604b (string $string): void {
 		assertType("array{0: string, 1?: ''|'XX', 2?: ''|'YY', 3?: 'ZZ'}", $matches);
 	}
 }
+
+function testLtrimDelimiter (string $string): void {
```
staabm: most of the newly added tests did not fail before this PR:
https://phpstan.org/r/403f8d7e-8f7c-491b-800d-c7384d62a13e

but I think its still fine to have such edge case stuff

resources/RegexGrammar.pp
mvorisek: > But adding a comment with // on the %token above could make sense

already done in https://github.com/phpstan/phpstan-src/pull/3488#discussion_r1778117255

resources/RegexGrammar.pp
Seldaek: Needs the same changes as above in `internal_option` for `^` 

And btw `xx` is also described on https://www.regular-expressions.info/refmodifiers.html - it acts like x but also supports spaces inside character classes. It only works as an inline modifier though, not in main regex modifiers https://3v4l.org/ssAQn

resources/RegexGrammar.pp
```diff
@@ -42,11 +42,13 @@
 //
 
 // Character classes.
+%token  negative_class_fc_       \[\^(?=\])         -> class_fc // same as "negative_class_" but followed by "class:_class"
+%token  class_fc_                \[(?=\])           -> class_fc // same as "class_"          but followed by "class:_class"
+%token class_fc:_class           \]                 -> class
```
Seldaek: Is this really needed? Can you give me an example of what it fixes? I was under the impression it worked fine with `class:_class_literal` before, but maybe I misunderstand something.

resources/RegexGrammar.pp
Seldaek: But btw no need to support `xx` specifically in the regex here anymore, as you now allow more than one modifier xx is anyway valid now just like `mmmmmmmm` will be valid too. This is only relevant information when handling modifiers in the PHP code later, where you want to make sure that if there is an internal option with `xx` then the character class handling should handle spaces differently in what follows.

resources/RegexGrammar.pp
mvorisek: yes - https://phpstan.org/r/7ed7c54b-ec78-4165-a8d2-857d3dbbb14e

Seldaek: Ah I see. Makes sense thanks 👍🏻

resources/RegexGrammar.pp
```diff
@@ -42,11 +42,13 @@
 //
 
 // Character classes.
+%token  negative_class_fc_       \[\^(?=\])         -> class_fc // same as "negative_class_" but followed by "class:_class"
+%token  class_fc_                \[(?=\])           -> class_fc // same as "class_"          but followed by "class:_class"
```
Seldaek: Maybe this would be more human readable? :)

```suggestion
%token  negative_class_fc_       \[\^(?=\])         -> class_fc // same as "negative_class_" but for when the class starts with "]"
%token  class_fc_                \[(?=\])           -> class_fc // same as "class_" but for when the class starts with "]"
```

resources/RegexGrammar.pp
mvorisek: done and thank you

resources/RegexGrammar.pp
mvorisek: done

resources/RegexGrammar.pp
mvorisek: done - https://github.com/phpstan/phpstan-src/pull/3488#discussion_r1778185558

resources/RegexGrammar.pp
mvorisek: please check the latest changes, comments are not allowed here

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -768,3 +771,125 @@ function bug11604b (string $string): void {
...
+	if (preg_match(<<<'EOD'
+		~([a[])~
+		EOD, $string, $matches)) {
+		assertType("array{string, '['|'a'}", $matches);
+	}
+
+	if (preg_match(<<<'EOD'
+		~([a\]b])~
+		EOD, $string, $matches)) {
+		assertType("array{string, 'a'|'b'}", $matches); // should be array{string, ']'|'a'|'b'}
```
Seldaek: I feel like these should be fixed as it's clearly a bug and not just a not-as-specific-as-it-could-be result.

Seldaek (approved):
Aside from the two tests that should be changed and bug fixed (but maybe @staabm can take a look in a follow-up PR as he's probably more familiar with that code), the changes here look good to me now 👍🏻

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
mvorisek: done - https://github.com/phpstan/phpstan-src/pull/3488/commits/6eed070cef2f04d7ca90557dd85f4fbb93f97968

Seldaek:
Cool, thanks for the last couple fixes 👍🏻

staabm (approved):
lgtm, thanks

mvorisek:
@ondrejmirtes can this PR be merged?

ondrejmirtes:
@mvorisek It's been only 4 days since you opened it. Don't be needy.

ondrejmirtes:
Thank you.

PR #3493, opened 2024-09-26
Add support for lowercase string on parse_url

VincentLanglet:
Since Type results are cached, I had to duplicate them two handle the case if the string is or not a lowercase-string.

Since some part of an url are not case sensitive, it might be useful to keep the `lowercase-string` accessory on it.
I have multiple example in my work codebase where we're doing
```
$url = strtolower($url); // lowercase-string

// Some manipulation on the URL which keeps the `lowercase-string` status (like trimming)

$host = parse_url($url, PHP_URL_HOST); // I expect to keep the lowercase-string

$this->manipulateHost($host); // `manipulateHost` require a lowercase-string
```


ondrejmirtes:
Thank you!

VincentLanglet:
@ondrejmirtes Do you see more functions where the DynamicReturnExtension need to be written/modified ?

staabm:
@VincentLanglet I would search the codebase for places where other string accessory types are handled

Likely sprintf/printf is worth testing

PR #3494, opened 2024-09-26
Fix test

VincentLanglet:
In https://github.com/phpstan/phpstan-src/pull/3492 tests were failing ; I fix it here


ondrejmirtes:
Thank you.

PR #3495, opened 2024-09-26
Fix missing ltrim in regex parse

mvorisek:


tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
```diff
@@ -768,3 +768,13 @@ function bug11604b (string $string): void {
 		assertType("array{0: string, 1?: ''|'XX', 2?: ''|'YY', 3?: 'ZZ'}", $matches);
 	}
 }
+
+function testLtrimDelimiter (string $string): void {
+	if (preg_match(' /(x)/', $string, $matches)) {
```
staabm, line 773: Instead use one of the regex in above tests with spaces prepended should work?

mvorisek: yes - but I prefer to keep such testcase separately

I would even prefer to use `' /x/'` but whole regex is currently not parsed into constant type.

staabm: Yeah, so the testcase does not Prevent regressions for the fix

mvorisek: see PR description why it is currently impossible to test

but I have tested what is otherwise passed to regex parse - it is/was a hidden bug

staabm:
Needs cs fix

mvorisek:
what cs fix?

staabm:
grafik

https://github.com/phpstan/phpstan-src/actions/runs/11057274125/job/30720684200?pr=3495

mvorisek:
tell me how, seems unrelated to me:
image

ondrejmirtes (changes requested):
The test does no fail even if I revert changes from src/

mvorisek:
@ondrejmirtes as explained above, it cannot be currently tested

ondrejmirtes:
I don't get it what's the purpose of this PR then.

tests/PHPStan/Analyser/nsrt/preg_match_shapes.php
mvorisek: test added in https://github.com/phpstan/phpstan-src/pull/3495/commits/dd8e9634d8dad3b21988d575b6b4eab947b49f52

mvorisek:
@ondrejmirtes test added

ondrejmirtes:
Now we have a failing unit test but I still don't understand what kind of issue this PR fixes.

mvorisek:
I understand that it does not currently fix a functional usecase, but it fixes internal hidden bug. It fixes a method behaviour that is needed for future implementations of regex shape narrowing as the initial delimiter needs to be properly removed per PCRE grammar.

ondrejmirtes:
Alright, I get it.

ondrejmirtes:
Thank you.

PR #3496, opened 2024-09-26
Improve Vsprintf inference

VincentLanglet:
The same logic done for `sprintf` could be done for `vsprintf`.
Instead of a foreach on every values, I just need to check the iterableValueType. 

I refacto to a method `allValuesSatisfies`, because I have in mind to re-use this method to check if every values are lowercase-string.

cc @staabm if you wanna review, since you seemed to work multiple times on this dynamicReturnTypeExtension.


staabm (approved):

ondrejmirtes:
Thank you.

PR #3498, opened 2024-09-26
Handle lowercase string in sprintf

VincentLanglet:


src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -147,7 +148,22 @@ public function getTypeFromFunctionCall(
 			return $singlePlaceholderEarlyReturn;
```
staabm: Does this case also need lowercase stuff?

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -130,10 +138,18 @@ public function getTypeFromFunctionCall(
...
+							new StringType(),
+							new AccessoryLowercaseStringType(),
+							new AccessoryNumericStringType(),
+						]);
+					} else {
+						$singlePlaceholderEarlyReturn = new IntersectionType([
+							new StringType(),
+							new AccessoryNumericStringType(),
+						]);
+					}
```
staabm: maybe it would make sense to extract this "if lowercase, add lowercase-type" into a separate method, so you don't need to repeat it over and over

VincentLanglet: Tried https://github.com/phpstan/phpstan-src/pull/3498/commits/b22c45b18b2b14316a2dbcf040d276c1c0e5bb60

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -130,10 +139,10 @@ public function getTypeFromFunctionCall(
 
 					$singlePlaceholderEarlyReturn = $checkArgType->toString();
```
staabm, line 140: I wonder this line is also missing lowercase handling

maybe instead of fiddling with this complex loop and all its paths, just add it once here:
```
		if ($singlePlaceholderEarlyReturn !== null) {
			return $singlePlaceholderEarlyReturn;
		}
```

VincentLanglet: I shouldn't add lowercase handling here and just rely on `toString`.

I'll open a PR to update `IntegerType::toString` to fix https://phpstan.org/r/79f89e3e-f19f-4a44-bd58-63f3c77db487

ondrejmirtes:
Thank you!

PR #3502, opened 2024-09-27
Offset access on lowercase-string is lowercase-string

staabm:
just adding a test, as it is implemented but not yet tested


ondrejmirtes:
Thank you.

PR #3503, opened 2024-09-27
Allow asserting array offset certainty

ruudk:
This makes it possible to do this:

```php
/**
 * @param array{firstName: string, lastName?: string} $context
 */
public function arrayOffset(array $context) : void
{
    assertVariableCertainty(TrinaryLogic::createYes(), $context['firstName']);
    assertVariableCertainty(TrinaryLogic::createMaybe(), $context['lastName']);
    assertVariableCertainty(TrinaryLogic::createNo(), $context['email']);
}
```

### DIFF
--- a/src/Testing/TypeInferenceTestCase.php
+++ b/src/Testing/TypeInferenceTestCase.php
@@ -135,7 +135,7 @@ abstract class TypeInferenceTestCase extends PHPStanTestCase
 			$variableName = $args[2];
 			$this->assertTrue(
 				$expectedCertainty->equals($actualCertainty),
-				sprintf('Expected %s, actual certainty of variable $%s is %s in %s on line %d.', $expectedCertainty->describe(), $variableName, $actualCertainty->describe(), $file, $args[3]),
+				sprintf('Expected %s, actual certainty of %s is %s in %s on line %d.', $expectedCertainty->describe(), $variableName, $actualCertainty->describe(), $file, $args[3]),
 			);
 		}
 	}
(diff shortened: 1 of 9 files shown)


src/Testing/TypeInferenceTestCase.php
```diff
@@ -135,7 +135,7 @@ public function assertFileAsserts(
 			$variableName = $args[2];
 			$this->assertTrue(
 				$expectedCertainty->equals($actualCertainty),
-				sprintf('Expected %s, actual certainty of variable $%s is %s in %s on line %d.', $expectedCertainty->describe(), $variableName, $actualCertainty->describe(), $file, $args[3]),
+				sprintf('Expected %s, actual certainty of variable %s is %s in %s on line %d.', $expectedCertainty->describe(), $variableName, $actualCertainty->describe(), $file, $args[3]),
```
ondrejmirtes:
```suggestion
				sprintf('Expected %s, actual certainty of %s is %s in %s on line %d.', $expectedCertainty->describe(), $variableDescription, $actualCertainty->describe(), $file, $args[3]),
```

And then in `$variableDescription` either have `variable $%s`, or `offset %s`...

Also would be nice to somehow test the assertion message here.

ondrejmirtes (changes requested):

src/Testing/TypeInferenceTestCase.php
ruudk: I applied the feedback. I added an assertion on the error in FileAssertRuleTest. I think it matters there. As the other errors are currently not reporting (which is good) and are only when the test fails?

ruudk: But if you have an idea where to add the extra test, I make the change 😊

staabm:
nice one!

ruudk:
@ondrejmirtes I pushed another commit to add assertions on the errors. Is this what you had in mind?

ruudk:
@ondrejmirtes Would be great to have this part of the next release 🙏. Does this need more testing?

tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -101,138 +101,148 @@ private static function findTestFiles(): iterable
...
+			yield $testName => [$file, $expectedFailures];
 		}
 	}
 
 	/**
+	 * @param list<string> $expectedFailures
 	 * @dataProvider dataFile
 	 */
-	public function testFile(string $file): void
+	public function testFile(string $file, array $expectedFailures = []): void
```
ondrejmirtes: Don't change this signature. Don't change the many `yield` lines. It doesn't make sense to expect failures outside of testing the actual implementation.

There's already TypeInferenceTestCaseTest. Test it there.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/NodeScopeResolverTest.php
ruudk: It seems the last part of your comment was not stored...

ondrejmirtes: I meant to delete that :)

ruudk: @ondrejmirtes Done, I hope this is what you had in mind.

tests/PHPStan/Testing/TypeInferenceTestCaseTest.php
```diff
@@ -90,4 +91,14 @@ public function testFileAssertionFailedErrors(string $filePath, string $errorMes
...
 	}
 
+	public function testVariableOrOffsetDescription(): void
+	{
+		$filePath = __DIR__ . '/data/assert-certainty-variable-or-offset.php';
+
+		[$variableAssert, $offsetAssert] = array_values($this->gatherAssertTypes($filePath));
+
+		$this->assertSame('variable $context', $variableAssert[4]);
+		$this->assertSame("offset 'email'", $offsetAssert[4]);
```
ruudk, line 101: I did it this way, because the rest in this test did not actually ran the assertions, it only waited for thrown acceptors. See line 91 above. 

So if you think we should actually call the asserts, please tell me how, and I'll make the change. 

Then we can assert on the full error message.

ruudk:
@ondrejmirtes I think this one is now ready to go. The failing tests seem not related?

ondrejmirtes:
Thank you!

PR #3504, opened 2024-09-27
Report precise offsets in errors

ruudk:
Fixes phpstan/phpstan#11760


ondrejmirtes (changes requested):
Some CS violations here https://github.com/phpstan/phpstan-src/actions/runs/11069234135/job/30756357522

ruudk:
Fixed! 

💡 TIL: `make cs + make cs-fix`

ruudk:
I believe the 71 failing tests are unrelated to my change. Is that correct?

ruudk:
I'm wondering. Shouldn't we also port this to 1.12 under bleeding edge? You say that 1.12 + bleeding edge == 2.0, right?

ondrejmirtes:
2.0 is close enough :)

ondrejmirtes:
Thank you.

PR #3505, opened 2024-09-27
Report unparseable regexp when compilable

mvorisek:
needs https://github.com/phpstan/phpstan-src/pull/3488 to be merged first

Figuring out regexes that are not parsed by PHPStan grammar is hard. Let's help developers and they can help us.

With good reports PHPStan grammar can be improved quickly so all PCRE patterns are supported by PHPStan parser.

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -63,5 +63,6 @@ parameters:
 		explicitThrow: true
 		absentTypeChecks: true
 		requireFileExists: true
+		reportUnparsedRegexpWhenCompilable: true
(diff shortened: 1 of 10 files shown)


Seldaek:
I'm scared to see what shows up, but that'd definitely be valuable.

ondrejmirtes:
Errors reported by PHPStan should be bugs in analysed code fixable by the developer.

Errors reported by PHPStan should not be PHPStan bugs and give more work to PHPStan users.

Feel free to use this rule in your own projects, and fix those bugs in PHPStan based on that.

PR #3506, opened 2024-09-27
Fix preg_match_all with PREG_SET_ORDER does not see capture group as optional

staabm:
closes https://github.com/phpstan/phpstan/issues/11661


ondrejmirtes:
Thank you.

PR #3507, opened 2024-09-27
Fix empty regex and empty alternation parse

mvorisek:
fix https://github.com/phpstan/phpstan/issues/11762

### DIFF
--- a/resources/RegexGrammar.pp
+++ b/resources/RegexGrammar.pp
@@ -135,7 +135,7 @@
     alternation()
 
 alternation:
-    concatenation() ( ::alternation:: concatenation() #alternation )*
+    concatenation()? ( <alternation> concatenation()? #alternation )*
 
 concatenation:
     (   internal_options() | assertion() | quantification() | condition() )
@@ -154,8 +154,8 @@ concatenation:
         <index>
       | ::assertion_reference_:: alternation() #assertioncondition
     )
-    ::_capturing:: concatenation()?
-    ( ::alternation:: concatenation()? )?
+    ::_capturing::
+    alternation()
     ::_capturing::
 
 assertion:
@@ -165,7 +165,8 @@ assertion:
       | ::lookbehind_::          #lookbehind
       | ::negative_lookbehind_:: #negativelookbehind
     )
-    alternation() ::_capturing::
+    alternation()
+    ::_capturing::
 
 quantification:
     ( class() | simple() ) ( quantifier() #quantification )?
@@ -208,7 +209,8 @@ simple:
       | ::atomic_group_:: #atomicgroup
       | ::capturing_::
     )
-    alternation() ::_capturing::
+    alternation()
+    ::_capturing::
 
 non_capturing_internal_options:
     <non_capturing_internal_option>
--- a/tests/PHPStan/Command/IgnoredRegexValidatorTest.php
+++ b/tests/PHPStan/Command/IgnoredRegexValidatorTest.php
@@ -101,10 +101,46 @@ class IgnoredRegexValidatorTest extends PHPStanTestCase
 				false,
 			],
 			[
+				'~(a\()~',
+				[],
+				false,
+				false,
+			],
+			[
+				'~b\\\()~',
+				[],
+				false,
+				true,
+			],
+			[
+				'~(c\\\\\()~',
+				[],
+				false,
+				false,
+			],
+			[
 				'~Result of || is always true.~',
 				[],
 				false,
 				true,
+			],
+			[
+				'~a\||~',
+				[],
+				false,
+				false,
+			],
+			[
+				'~b\\\||~',
+				[],
+				false,
+				true,
+			],
+			[
+				'~c\\\\\||~',
+				[],
+				false,
+				false,
 			],
 			[
 				'#Method PragmaRX\Notified\Data\Repositories\Notified::firstOrCreateByEvent() should return PragmaRX\Notified\Data\Models\Notified but returns Illuminate\Database\Eloquent\Model|null#',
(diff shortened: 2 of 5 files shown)


resources/RegexGrammar.pp
```diff
@@ -132,7 +135,7 @@
     alternation()
 
 alternation:
-    concatenation() ( ::alternation:: concatenation() #alternation )*
+    concatenation()? ( <alternation> concatenation()? #alternation )*
```
mvorisek, line 138: @staabm would you sending a PR into this PR branch for this change?

I looked into https://github.com/phpstan/phpstan-src/blob/1.12.5/src/Type/Regex/RegexGroupParser.php#L180 and I do not know the internals, why count is passed etc.

The effective change of this line is `|` tokens are newly captured inside the AST and if there are empty tokens before/after `|`, then the alternation expression matches `''`.

staabm, line 138: sorry, I don't get what you ask me for...?

we use the count of alternations to build a more precise tagged union for some special cases

mvorisek: Previously "PCRE alternation" captured only the tokens between `|`, newly also `|` tokens are captured and the parsing code needs to be adjusted.

mvorisek: Well, I coped with tha by modifying the parsed AST to produce the same AST as before this PR but with empty tokens added.

staabm:
Do empty alternations or empty capturing groups have real world relevance?

I don't think I have ever seen a regex using it..?

mvorisek:
They definitely do, `~@(|phpstan-)param~` is a small example.

tests/PHPStan/Command/IgnoredRegexValidatorTest.php
```diff
@@ -104,13 +104,15 @@ public function dataValidate(): array
 				'~Result of || is always true.~',
 				[],
 				false,
-				true,
+				false,
```
ondrejmirtes, line 131: This is a blocker, I can't merge this. This is the main reason why IgnoredRegexValidator exists at all.

ondrejmirtes (changes requested):

tests/PHPStan/Command/IgnoredRegexValidatorTest.php
mvorisek: I do not understand what the test does. Please tell me what to do or feel free to push directly.

mvorisek: @ondrejmirtes if I understand the test correctly it was rejecting `...||...` regex but the regex is valid, `|` is an alternation regex token and thus `||` means alternation with empty string which is fixed by this PR, would you mind amending the test and pushing to by branch directly?

ondrejmirtes: IgnoredRegexValidator makes sure that if people have this in their configuration:

```
parameters:
	ignoreErrors:
		- '~Result of || is always true.~'

```

It makes PHPStan fail with the following error:

```
Invalid entry in ignoreErrors:
Ignored error ~Result of || is always true.~ has an unescaped '||' which leads to ignoring all errors. Use '\|\|' instead.
```

mvorisek: Sure, but it MUST NOT fail, as it is valid regex :)

So please tell me how to make everyone happy.

ondrejmirtes: It's not a valid regex for `ignoreErrors`.

mvorisek: Why?

(I know it is probably a mistate as unescaped, but in general, all regexes are supported)

mvorisek: @ondrejmirtes I have updated the PR so these "ignoreErrors errors" are reported as before.

mvorisek:
@staabm can I have your review here?

tests/PHPStan/Command/IgnoredRegexValidatorTest.php
mvorisek: @ondrejmirtes these "obviously unescaped `ignoreErrors` cases" are not handled. Can you check?

mvorisek: @ondrejmirtes can you please check and let me know if there is anything to address?

mvorisek: @ondrejmirtes would you mind sharing your review?

ondrejmirtes: Maybe in December, after I release 2.0. Don't expect it sooner.

mvorisek: This is a bugfix into 1.12.x, might I understand why so late?

ondrejmirtes: Because I have different priorities and this is not a severe bug affecting a lot of users.

mvorisek: @ondrejmirtes is now a better time?

About the usecase, see https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.65.0/src/DocBlock/TypeExpression.php#L62 for example.

`(a|b|)` is shorter syntax of `((?:a|b|)?)`, the first one is currently now parsed by PHPStan and this PR fixes it.

mvorisek: @ondrejmirtes the https://github.com/phpstan/phpstan-src/pull/3507#discussion_r1789111901 review note has been addressed

mvorisek: @ondrejmirtes might I have this PR reviewed, please.

mvorisek: Hi @ondrejmirtes, please review, the feedback was addressed.

ondrejmirtes:
Looks fine, thank you.

mvorisek:
Thank you ❤

PR #3508, opened 2024-09-27
Refactor RegexGroupParser for more immutability and less pass-by-ref

staabm:
less magic, less error prone

Part 2

### DIFF
--- a/src/Type/Regex/RegexAstWalkResult.php
+++ b/src/Type/Regex/RegexAstWalkResult.php
@@ -23,6 +23,7 @@ final class RegexAstWalkResult
 	{
 		return new self(
 			-1,
+			// use different start-index for groups to make it easier to distinguish groupids from other ids
 			100,
 			[],
 			[],
--- a/src/Type/Regex/RegexAstWalkResult.php
+++ b/src/Type/Regex/RegexAstWalkResult.php
@@ -1,0 +1,104 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Regex;
+
+/** @immutable */
+final class RegexAstWalkResult
+{
+
+	/**
+	 * @param array<int, RegexCapturingGroup> $capturingGroups
+	 * @param list<string> $markVerbs
+	 */
+	public function __construct(
+		private int $alternationId,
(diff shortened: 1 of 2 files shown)


staabm:
//cc @mvorisek lets wait until a few of the open Regex-inference PRs are merged so we don't step on each others toes

src/Type/Regex/RegexAstWalkResult.php
```diff
@@ -0,0 +1,104 @@
...
+		private array $markVerbs,
+	)
+	{
+	}
+
+	public static function createEmpty(): self
+	{
+		return new self(
+			-1,
+			100,
```
mvorisek, line 27: I know you moved the code only, but why 100?

staabm: in the initial implementation there we so many arrays involved, that I used a different start-index to trigger php warnings in case I mixed up the keys between different arrays ;)

mvorisek: please add a comment into the code

src/Type/Regex/RegexAstWalkResult.php
```diff
@@ -0,0 +1,104 @@
...
+			[],
+			[],
+		);
+	}
+
+	public function nextAlternationId(): self
```
mvorisek, line 33: name it better, it does not return `int`

staabm: it returns a "copy" of the current object, which contains the next alternation id.

I think the name fits.

src/Type/Regex/RegexAstWalkResult.php
```diff
@@ -0,0 +1,104 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Regex;
```
mvorisek, line 3:
```suggestion
namespace PHPStan\Type\Regex\AST;
```

the Hoa AST should be parsed into PHPStan AST with clean object design

walking such AST will be much easier and cleaner then

ondrejmirtes:
Thank you.

PR #3509, opened 2024-09-27
Fix false-positive with preg_match(): Strict comparison using === between '' and non-falsy-string…ween '' and non-falsy-string

staabm:
closes https://github.com/phpstan/phpstan/issues/11744


ondrejmirtes:
Conflict here.

ondrejmirtes:
Thank you.

PR #3510, opened 2024-09-27
IntegerType::toString is lowercase

VincentLanglet:
Which is useful to keep 
```
$lowercaseString . $integer
```
as a lowercase string.


staabm:
I recently learned about https://3v4l.org/nvKP7

~~not sure thats something anyone is doing though~~

Update: ahh its a float, doesn't matter then: https://3v4l.org/sC13o

VincentLanglet:
Yes,

Int to string is a lowercase string but not float to string.

ondrejmirtes:
Thank you!

ondrejmirtes:
Hi, this broke tests on phpstan-doctrine for some PHP versions and I don't understand why only for some. I'm reverting this. Feel free to investigate and fix it there. Thanks.

See https://github.com/phpstan/phpstan/actions/runs/11202848434

VincentLanglet:
> Hi, this broke tests on phpstan-doctrine for some PHP versions and I don't understand why only for some. I'm reverting this. Feel free to investigate and fix it there. Thanks.
> 
> See [phpstan/phpstan/actions/runs/11202848434](https://github.com/phpstan/phpstan/actions/runs/11202848434)

Hi @ondrejmirtes this is "expected", the behavior seems to be different in PHP < 8.1
cf this code
https://github.com/phpstan/phpstan-doctrine/blob/14a59bb5b719d3740224b6ba8aec80c0294142f9/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L1633-L1643

I assume adding the LowerCase accessory to https://github.com/phpstan/phpstan-doctrine/blob/14a59bb5b719d3740224b6ba8aec80c0294142f9/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L1590 might solve all the issues.

So to me, it's all good, we could revert the revert and just fix the tests.

ondrejmirtes:
Please try to send a PR to phpstan-doctrine on 1.5.x branch and pin phpstan/phpstan to the non-reverted version (1.12.6).

VincentLanglet:
> Please try to send a PR to phpstan-doctrine on 1.5.x branch and pin phpstan/phpstan to the non-reverted version (1.12.6).

Done https://github.com/phpstan/phpstan-doctrine/pull/614 @ondrejmirtes

PR #3511, opened 2024-09-27
Add phpdoc to Exception properties

VincentLanglet:
Not sure if it was the expected way to solve https://github.com/phpstan/phpstan/issues/11761

Closes https://github.com/phpstan/phpstan/issues/11761

### DIFF
--- a/stubs/Exception.stub
+++ b/stubs/Exception.stub
@@ -52,6 +52,17 @@ interface Throwable
 
 class Exception implements Throwable
 {
+	/** @var string */
+	protected $message;
+
+	/** @var mixed */
+	protected $code;
+
+	/** @var string */
+	protected $file;
+
+	/** @var int */
+	protected $line;
 
 	/**
 	 * @return string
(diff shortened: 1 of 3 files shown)


stubs/Exception.stub
```diff
@@ -52,6 +52,17 @@ interface Throwable
 
 class Exception implements Throwable
 {
+	/** @var string */
+	protected $message;
+
+	/** @var mixed */
```
staabm, line 58: The only exception type I am aware of which doesnt has a `int` code is PDOException

VincentLanglet: Yes me too.

But getCode is typed as returning mixed... so I think I have to match this

I'll wait for ondrej decision

VincentLanglet:
Hi, do you have any request change on this PR @ondrejmirtes.

Thanks.

ondrejmirtes:
> Not sure if it was the expected way to solve

I'm also not sure if this is the expected way. I will have to think about this more deeply when I get the time.

PR #3512, opened 2024-09-28
Fix extract signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11759


VincentLanglet:
Params names are coming from https://www.php.net/manual/en/function.extract.php
They are also here:
https://github.com/php/php-src/blob/7e45e57d8f85f8d1fa2521028f394da30268187d/ext/standard/basic_functions.stub.php#L1643-L1644

I added some tests to check `&` does not break the `extract` inference.

Is something more needed for this PR ? @ondrejmirtes

thg2k:
The param names should be the ones from php7. Reflection takes care of the correct names for 8.x.

VincentLanglet:
> The param names should be the ones from php7. Reflection takes care of the correct names for 8.x.

Was the param names different in php 7 ?

claudepache:
> Was the param names different in php 7 ?

See https://3v4l.org/kMrVd . The current parameter names aren’t exactly those of PHP 7 either…

VincentLanglet:
Since named parameter is only a PHP 8 feature, isn't it better to use them in stubs ?
I'm not sure what's the best for PHPStan

tests/PHPStan/Analyser/nsrt/extract.php
```diff
@@ -66,3 +66,17 @@ function doTyped3(array $vars): void
...
 }
+
+function doTyped4(): void
+{
+	extract(['foo' => 42]);
+
+	assertVariableCertainty(TrinaryLogic::createYes(), $foo);
+	assertType('42', $foo);
+
+	$foo = ['foo' => 42];
```
ondrejmirtes, line 81: Would make more sense to test this separately in a different function.

ondrejmirtes (changes requested):

ondrejmirtes:
Also: please search for `@prefer-ref` in vendor/phpstan/php-8-stubs and update all the functions you find too.

VincentLanglet:
> Would make more sense to test this separately in a different function.

Done in https://github.com/phpstan/phpstan-src/pull/3512/commits/f9fa213fc39e2b9b4f9817dba28e030fa51cfb5c

> Also: please search for `@prefer-ref` in vendor/phpstan/php-8-stubs and update all the functions you find too.

Done in https://github.com/phpstan/phpstan-src/pull/3512/commits/b5e5de50331a0e6eb8f7af1440d04bba619d9232 if I understood correctly

resources/functionMap.php
```diff
@@ -286,7 +286,7 @@
 'array_map' => ['array', 'callback'=>'?callable', 'array'=>'array', '...args='=>'array'],
 'array_merge' => ['array', 'arr1'=>'array', '...args='=>'array'],
 'array_merge_recursive' => ['array', 'arr1'=>'array', '...args='=>'array'],
-'array_multisort' => ['bool', '&rw_array1'=>'array', 'array1_sort_order='=>'array|int', 'array1_sort_flags='=>'array|int', '...args='=>'array|int'],
+'array_multisort' => ['bool', 'rw_array1'=>'array', 'array1_sort_order='=>'array|int', 'array1_sort_flags='=>'array|int', '...args='=>'array|int'],
```
ondrejmirtes: The name of the parameter is `array1`, not `rw_array1`.

ondrejmirtes (changes requested):

resources/functionMap.php
VincentLanglet: I saw many other param with a wrong name, like
- array_pop => rw_stack
- array_push => rw_stack
- array_shift => rw_stack
- array_splice => rw_input
- array_unsplice => rw_stack
- ...

There is at least 63 occurences of `&rw_` param which might need to be checked...

Do you want me to open another PR with such param renaming ?

ondrejmirtes: These aren't part of the name when there's `&` in front of them. There's an explanation at the top of functionMap.php

ondrejmirtes:
Thank you.

PR #3514, opened 2024-09-29
Add `Type::sliceArray()`

herndlm:
Improves type specification edge cases slightly, but there is more potential which I definitely don't want to get into here (e.g. currently `$offsetType` and `$lengthType` are only used for constant arrays if they are of type `ConstantIntegerType`).

The test change is correct I believe, those array slices might not be lists, see [3v4l.org/hAv10](https://3v4l.org/hAv10)

I dislike the 2 new `instanceof self` checks, but they are only for BC and can be removed / will not be there in 2.0

### DIFF
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -247,6 +247,19 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 		return $this;
 	}
 
+	public function sliceArray(Type $offsetType, Type $lengthType, TrinaryLogic $preserveKeys): Type
+	{
+		if ($preserveKeys->no()) {
+			return $this;
+		}
+
+		if ((new ConstantIntegerType(0))->isSuperTypeOf($offsetType)->yes()) {
+			return $this;
+		}
+
+		return new MixedType();
+	}
+
 	public function isIterable(): TrinaryLogic
 	{
 		return TrinaryLogic::createYes();
--- a/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
@@ -4,18 +4,21 @@ namespace PHPStan\Type\Php;
 
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
+use PHPStan\Php\PhpVersion;
 use PHPStan\Reflection\FunctionReflection;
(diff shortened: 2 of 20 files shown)


src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -247,6 +247,15 @@ public function shuffleArray(): Type
 		return $this;
 	}
 
+	public function sliceArray(Type $offsetType, Type $lengthType, TrinaryLogic $preserveKeys): Type
+	{
+		if ($offsetType instanceof ConstantIntegerType && $offsetType->getValue() === 0) {
```
herndlm: Note to self: should be a super type check and support for preserveKeys no can be added

src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -247,6 +247,15 @@ public function shuffleArray(): Type
...
+			return $this;
+		}
+
+		return $this;
```
VincentLanglet: It returns this in the two situations, so the if check is useless, isn't it ?

herndlm: thx, yeah I messed this implementation somehow up with reverting parts and finally missing things.. should be good now :)

src/Type/Php/ArraySliceFunctionReturnTypeExtension.php
```diff
@@ -25,49 +28,22 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
-			) {
-				return TypeCombinator::intersect($optionalOffsetsType, new NonEmptyArrayType());
-			}
+		$offsetType = $scope->getType($args[1]->value);
+		$lengthType = isset($args[2]) ? $scope->getType($args[2]->value) : new NullType();
 
-			return $optionalOffsetsType;
-		}
+		$preserveKeysType = isset($args[3]) ? $scope->getType($args[3]->value) : new ConstantBooleanType(false);
+		$preserveKeys = (new ConstantBooleanType(true))->isSuperTypeOf($preserveKeysType);
```
staabm: maybe

```suggestion
		$preserveKeys = $preserveKeysType->isTrue()->yes();
```

herndlm: thx, good idea! slightly different but nice cleanup 2d50299d28611c61e3864b6f7863c5a6bdc35d29

(I'll adapt this same pattern I started sometime in the other places too in a dedicated MR :))

ondrejmirtes:
Thank you!

PR #3516, opened 2024-09-30
Simplify preserveKeys TrinaryLogic creation in 2 extensions

herndlm:
Small refactor improving readability, based on what @staabm suggested in https://github.com/phpstan/phpstan-src/pull/3514#discussion_r1780612312


ondrejmirtes:
Thank you!

PR #3518, opened 2024-09-30
Add ArrayFindFunctionReturnTypeExtension

zonuexe:
Supports [PHP RFC: array_find](https://wiki.php.net/rfc/array_find) for PHP 8.4.

### DIFF
--- a/src/Reflection/ParametersAcceptorSelector.php
+++ b/src/Reflection/ParametersAcceptorSelector.php
@@ -9,6 +9,7 @@ use PHPStan\Analyser\MutatingScope;
 use PHPStan\Analyser\Scope;
 use PHPStan\Node\Expr\ParameterVariableOriginalValueExpr;
 use PHPStan\Parser\ArrayFilterArgVisitor;
+use PHPStan\Parser\ArrayFindArgVisitor;
 use PHPStan\Parser\ArrayMapArgVisitor;
 use PHPStan\Parser\ArrayWalkArgVisitor;
 use PHPStan\Parser\ClosureBindArgVisitor;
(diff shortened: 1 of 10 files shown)


tests/PHPStan/Analyser/nsrt/array-find.php
```diff
@@ -0,0 +1,41 @@
...
+		assertType('int|null', array_find($array, 'is_int'));
+	}
+
+	/**
+	 * @param array{1, 'foo', \DateTime} $array
+	 */
+	function testConstant(array $array, callable $callback): void
+	{
+		assertType("1|'foo'|DateTime|null", array_find($array, $callback));
+		assertType("1|null", array_find($array, 'is_int'));
```
VincentLanglet: This cannot be Null.

tests/PHPStan/Analyser/nsrt/array-find.php
```diff
@@ -0,0 +1,41 @@
...
+
+	use function PHPStan\Testing\assertType;
+
+	/**
+	 * @param array<mixed> $array
+	 */
+	function testMixed(array $array, callable $callback): void
+	{
+		assertType('mixed', array_find($array, $callback));
+		assertType('int|null', array_find($array, 'is_int'));
```
VincentLanglet, line 38: You should also test it on non empty array, and the result should be an int

tests/PHPStan/Analyser/nsrt/array-find-key.php
```diff
@@ -0,0 +1,41 @@
...
+		assertType('int|string|null', array_find_key($array, 'is_int'));
+	}
+
+	/**
+	 * @param array{1, 'foo', \DateTime} $array
+	 */
+	function testConstant(array $array, callable $callback): void
+	{
+		assertType("0|1|2|null", array_find_key($array, $callback));
+		assertType("0|1|2|null", array_find_key($array, 'is_int'));
```
VincentLanglet, line 47: It should be 0

tests/PHPStan/Analyser/nsrt/array-find-key.php
```diff
@@ -0,0 +1,41 @@
...
+
+	use function PHPStan\Testing\assertType;
+
+	/**
+	 * @param array<mixed> $array
+	 */
+	function testMixed(array $array, callable $callback): void
+	{
+		assertType('int|string|null', array_find_key($array, $callback));
+		assertType('int|string|null', array_find_key($array, 'is_int'));
```
VincentLanglet, line 37: You should test on non-empty-array<int> and it should gives int|string

tests/PHPStan/Analyser/nsrt/array-find-key.php
zonuexe: Ideally, that's correct. This PR does not implement the logic to detect this.

VincentLanglet: What's the current behavior ? I assume the type is considered as mixed.

This means that merging the PR as if will introduce a lot of false positif PHPstan errors on level 7 (unions) when it's currently on level 9 (mixed).
That's why IMHO if a dynamic return type extension is introduced, the implementation need to handle such case.

staabm: > What's the current behavior ? I assume the type is considered as mixed.

just copy the test over into the playground:
https://phpstan.org/r/ece55e5c-33bf-47cb-b538-35708d025dfa

tests/PHPStan/Analyser/nsrt/array-find.php
zonuexe: Fixed https://github.com/phpstan/phpstan-src/pull/3518/commits/f0d9c2578802baeb0e6b8ba8cdf8663c54487df3

tests/PHPStan/Analyser/nsrt/array-find-key.php
zonuexe: It seems like it can be improved at the same time as `array_filter()`.

https://phpstan.org/r/ab62bff1-d485-46ae-9dfa-1931ce320a25

tests/PHPStan/Analyser/nsrt/array-find.php
zonuexe: It seems like it can be improved at the same time as `array_filter()`.

https://phpstan.org/r/ab62bff1-d485-46ae-9dfa-1931ce320a25

tests/PHPStan/Analyser/nsrt/array-find-key.php
zonuexe: We can add a test case, but if you pass a `non-empty-array<mixed>` the function may return any `int|string|null`.

https://3v4l.org/Biok8

VincentLanglet: Indeed, my bad.

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
```diff
@@ -62,290 +27,7 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-		}
-
-		return new Name($funcName);
-	}
-
-	/**
-	 * @param self::USE_* $mode
-	 * @return array{list<Arg>, ?Variable, ?Variable}
-	 */
-	private function createDummyArgs(int $mode): array
```
ondrejmirtes: Do this refactoring in a separate PR first please.

ondrejmirtes (changes requested):

src/Type/Php/ArrayFilterFunctionReturnTypeExtension.php
zonuexe: Ok, create a pull request.

zonuexe: Cherry-picked https://github.com/phpstan/phpstan-src/pull/3606/

zonuexe:
I just cherry-picked them, so there should be no conflicts when https://github.com/phpstan/phpstan-src/pull/3606 is merged.

ondrejmirtes:
Fix CS please.

zonuexe:
@ondrejmirtes I fixed and rebased.

src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -257,6 +258,70 @@ public static function selectFromArgs(
...
+						$acceptor->getResolvedTemplateTypeMap(),
+						$parameters,
+						$acceptor->isVariadic(),
+						$acceptor->getReturnType(),
+						$acceptor instanceof ParametersAcceptorWithPhpDocs ? $acceptor->getCallSiteVarianceMap() : TemplateTypeVarianceMap::createEmpty(),
+					),
+				];
+			}
+
+			if (isset($args[0]) && (bool) $args[0]->getAttribute(ArrayWalkArgVisitor::ATTRIBUTE_NAME)) {
```
ondrejmirtes: Why is there a section about array_walk and it has no tests? It should be a separate PR in any case.

ondrejmirtes (changes requested):

src/Reflection/ParametersAcceptorSelector.php
zonuexe: Oops, I copied too much existing code.
I'm sure you're right, I'll split out the changes to ParametersAcceptorSelector into a separate PR.

ondrejmirtes (changes requested):
Now you’ve removed too much 😀 It’s useful to declare the callable parameters for two reasons:

1) To infer the parameter type inside the callable based on array (even when the parameter does not have a native type).
2) To report wrong callable parameter tpe when it’s different to the passed array.

Please add tests for these.

zonuexe:
Ok, I'll get to it right away.

ondrejmirtes:
Perfect, thank you!

PR #3520, opened 2024-10-01
Update function map for ext-mongodb 1.20

alcaeus:
This updates the function map to include methods introduced in ext-mongodb 1.20.


resources/functionMap.php
```diff
@@ -6860,7 +6864,7 @@
 'MongoDB\Driver\ClientEncryption::rewrapManyDataKey' => ['object', 'filter'=>'object|array', 'options='=>'?array'],
 'MongoDB\Driver\Command::__construct' => ['void', 'document'=>'object|array', 'commandOptions='=>'?array'],
 'MongoDB\Driver\Cursor::current' => ['object|array|null'],
-'MongoDB\Driver\Cursor::getId' => ['MongoDB\Driver\CursorId'],
+'MongoDB\Driver\Cursor::getId' => ['MongoDB\Driver\CursorId|MongoDB\BSON\Int64', 'asInt64='=>'bool'],
```
staabm, line 6867: this reads like the return type would depend on the given parameter `$asInt64`.
if this is the case, this method would benefit from a stub, similar to e.g.

https://github.com/phpstan/phpstan-src/blob/f38addda2b151b6e41a746a37659c0bbe9e2293b/stubs/core.stub#L3-L9

that way phpstan can know when `Cursor::getId` is invoked with `true`, it will return `MongoDB\BSON\Int64`

GromNaN: In ext-mongodb v2.0, the return type will change to be always `Int64`. This parameter provides a migration path from v1.x to v2.0.

alcaeus: @staabm to elaborate on what @GromNaN said, here are the signatures in various versions of `ext-mongodb`:
* Before 1.20: `public function getId(): CursorId`
* 1.20 and newer: `public function getId(bool $asInt64): Int64|CursorId` with the conditional return type based on the value of `$asInt64`
* 2.0: `public function getId(): Int64`

What's the preferred way for handling these signature differences in PHPStan? There are different function maps for different versions of PHP, but I don't see a mechanism similar to that. It's a problem that already exists, as when we add functionality to a new extension version, PHPStan will apply those maps even to older versions of the extension. It usually isn't a big issue with minor releases, but since 2.0 will drop some methods from the map, this would cause issues for users that continue to use the 1.x releases.

alcaeus:
No longer relevant after #3891

PR #3521, opened 2024-10-01
Introduce Scope::getMaybeDefinedVariables

ruudk:
Fixes phpstan/phpstan#11772

Where would be the best way to create a test for this?

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -574,6 +574,27 @@ final class MutatingScope implements Scope
 		return $variables;
 	}
 
+	/**
+	 * @api
+	 * @return array<int, string>
+	 */
+	public function getMaybeDefinedVariables(): array
+	{
+		$variables = [];
+		foreach ($this->expressionTypes as $exprString => $holder) {
+			if (!$holder->getExpr() instanceof Variable) {
+				continue;
+			}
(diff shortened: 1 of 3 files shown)


ruudk:
Every time 🙈

staabm:
I think a test could fit in

https://github.com/phpstan/phpstan-src/blob/1.12.x/tests/PHPStan/Analyser/ScopeTest.php

ruudk:
@staabm Thanks. Somehow I could not find it when using `Go to test` in PHPStorm. I fixed that now.

The tests are added.

@ondrejmirtes This is ready for review!

ondrejmirtes:
Thank you!

src/Analyser/MutatingScope.php
```diff
@@ -574,6 +574,27 @@ public function getDefinedVariables(): array
 		return $variables;
 	}
 
+	/**
+	 * @api
+	 * @return array<int, string>
```
janedbal, line 579: why not `list<string>` ?

ruudk: Copied from getDefinedVars

janedbal: Fixed: https://github.com/phpstan/phpstan-src/pull/3530/files

ruudk: Great 🎉

PR #3522, opened 2024-10-01
Set normalized in BenevolentUnionType

herndlm:
Stumbled over this in UnionType and thought it should be set for BenevolentUnionType as well. I also noticed that it is not used yet. Tried early exiting in `TypeCombinator::union()` if only a normilzed union was passed, but it doesn't seem to affect performance much. Felt like a useless micro-optimization.

So this doesn't change anything really :)


ondrejmirtes:
Thank you.

PR #3523, opened 2024-10-02
Support `@readonly` PHPDoc on the class as alternative to `@immutable`

herndlm:
Closes https://github.com/phpstan/phpstan/issues/11775
Docs PR: https://github.com/phpstan/phpstan/pull/11783

I decided to directly adapt `ClassReflection::isImmutable()` because there is already `ClassReflection::isReadonly()` for native readonly and that would clash. An alternative would be to add a new `ClassReflection::isReadonlyPhpDoc()` or so, but I dislike that isImmutable is also PHPDoc only and it then feels somewhat inconsistent / even more complicated :)

the rules are still, as before, only enabled with the `readOnlyByPhpDoc` feature toggle.

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -1238,7 +1238,7 @@ class ClassReflection
 	{
 		if ($this->isImmutable === null) {
 			$resolvedPhpDoc = $this->getResolvedPhpDoc();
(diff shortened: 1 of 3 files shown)


herndlm:
I hope I didn't break GH actions somehow by, as always, choosing the wrong branch and then quickly changing and force-pushing :D something's weird right now. or just slow

keradus:
love it!

ondrejmirtes:
Thank you.

PR #3527, opened 2024-10-03
curl_multi_getcontent() can return null

staabm:
fixes

PHP 7.x 
> Call to function is_string() with string will always evaluate to true.

https://phpstan.org/r/bef8ccfb-a36f-4bc6-9884-f5627583859a
https://www.php.net/curl_multi_getcontent


ondrejmirtes:
Thank you.

PR #3529, opened 2024-10-03
nikic/php-parser v4 BC layer for renamed classes

ondrejmirtes:


compiler/src/Console/PrepareCommand.php
```diff
@@ -203,6 +205,8 @@ private function buildPreloadScript(): void
 			$output .= 'require_once __DIR__ . ' . var_export($path, true) . ';' . "\n";
 		}
 
+		$output .= 'class_alias(Node\\ArrayItem::class, Node\\Expr\\ArrayItem::class);' . "\n";
```
staabm, line 208: can we somehow make uses of the old class be reported as deprecation errors?

ondrejmirtes:
I realized the BC layer cannot work and instead, I'll write a new rule which makes these classes unknown in PHPStan-related code.

PR #3531, opened 2024-10-03
Add error reporting for unresolvable types in @phpstan-assert tags

zonuexe:
Fix https://github.com/phpstan/phpstan/issues/11780


ondrejmirtes:
Better fit would be to add it to AssertRuleHelper. We need to check for:

* Nonexistent classes and traits, like here: https://github.com/phpstan/phpstan-src/blob/b009a445bd3d96013d02746109d6ebd777275da4/src/Rules/FunctionDefinitionCheck.php#L157-L179
* Unresolvable types, like here: https://github.com/phpstan/phpstan-src/blob/b009a445bd3d96013d02746109d6ebd777275da4/src/Rules/PhpDoc/IncompatiblePhpDocTypeRule.php#L84-L91
* Missing types, like here: https://github.com/phpstan/phpstan-src/blob/b009a445bd3d96013d02746109d6ebd777275da4/src/Rules/Classes/MixinCheck.php#L71-L101
* Generics, like here: https://github.com/phpstan/phpstan-src/blob/b009a445bd3d96013d02746109d6ebd777275da4/src/Rules/Classes/MixinCheck.php#L137-L145

src/Rules/PhpDoc/AssertRuleHelper.php
```diff
@@ -55,6 +56,22 @@ public function check(ExtendedMethodReflection|FunctionReflection $reflection, P
 				continue;
 			}
 
+			if ($assert->getType() instanceof ErrorType) {
```
ondrejmirtes: Call UnresolvableTypeHelper instead please. You should follow the code samples I sent you :)

ondrejmirtes (changes requested):

src/Rules/PhpDoc/AssertRuleHelper.php
ondrejmirtes: Also please add all the other things that should be checked too.

ondrejmirtes:
I'm sorry, I'm in a bit of a hurry, finishing this myself.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/commit/ca0a7e9955397eef453b38c94ac67ba6faf7356b

PR #3532, opened 2024-10-03
Add stub for openssl_sign() function

piotrekkr:
Adding openssl_sign() stub so PHPStan will correctly deduce `$signature` variable type which is now shown as `mixed` instead of `string`. 

Fixes https://github.com/phpstan/phpstan/issues/11791


piotrekkr:
Will rebase properly in a minute

ondrejmirtes (changes requested):
You have to register this stub file in config/config.neon.

piotrekkr:
@ondrejmirtes Added it to conf/config.neon

piotrekkr:
Do I need to type all arguments or just this `@param-out` and the rest will be taken from signature? Asking because tests fail https://github.com/phpstan/phpstan-src/actions/runs/11162040542/job/31025901918 and they mention `$private_key` which is typed in function but seems like it is not enough.

ondrejmirtes (changes requested):
Please fix proc_open() from https://github.com/phpstan/phpstan/issues/11777 too.

ondrejmirtes:
> Do I need to type all arguments

Copy the PHPdoc from jetbrains/phpstorm-stubs and leave out unimportant stuff like description.

piotrekkr:
@ondrejmirtes I've updated phpdoc. Let me know if I need to do something more. Thanks

piotrekkr:
@ondrejmirtes I think there is some bug in PHP manual of this `openssl_sign()` function. They mention about private key that can be one of those
```
#[\SensitiveParameter] OpenSSLAsymmetricKey|OpenSSLCertificate|array|string $private_key,
```
In description of input params, there is no mention of `array`, so my guess is that array is not allowed. I removed it from stub but not sure if it is good way to go. Lent me know what you think.

piotrekkr:
Made bug report for php docs https://github.com/php/doc-en/issues/3829

piotrekkr:
Also signature changed in PHP 8 so I should probably create two stub files right? Something like `openssl.stub` and `openssl_php8.stub`. Then do it similar to [this](https://github.com/phpstan/phpstan-src/blob/8f55339fa525386da6e4182156d342951e14b990/src/PhpDoc/SocketSelectStubFilesExtension.php#L17) to have both new and older php handled. Is it correct?

piotrekkr:
Okay [asked about this `array` in `private_key`](https://github.com/php/php-src/issues/16275) and it seems it is correct but docs are not updated properly so will need to wait for it to be updated and will then update this PR

piotrekkr:
I'll close this since there is no progress here.

PR #3533, opened 2024-10-03
Improve e2e tests with bashunit 0.17

Chemaclass:
On the latest bashunit 0.17 version we have a better way to handle the stdout of an executed script, so you dont need to append an `|| true` syntax, while validating the exit code and you are even able to redirect the stderr if needed.

Blog post notes: 
- https://bashunit.typeddevs.com/blog/2024-10-01-release-0-17#standalone-exit-code-output-nicer-336


ondrejmirtes:
Thank you!

PR #3534, opened 2024-10-03
Remove inefficient caching from PhpMethodReflection

staabm:
as described in https://github.com/phpstan/phpstan/discussions/11748#discussioncomment-10753288


staabm:
as of now this PR only implements the `PhpMethodReflection` part. we would need the same thing for `PhpFunctionReflection`. Will do the 2nd part after a first feedback round.

ondrejmirtes:
PhpFunctionReflection too please.

src/Reflection/Php/PhpFunctionReflection.php
```diff
@@ -45,16 +41,14 @@ public function __construct(
...
-		private Cache $cache,
 		private TemplateTypeMap $templateTypeMap,
 		private array $phpDocParameterTypes,
 		private ?Type $phpDocReturnType,
 		private ?Type $phpDocThrowType,
 		private ?string $deprecatedDescription,
 		private bool $isDeprecated,
 		private bool $isInternal,
-		private bool $isFinal,
+		private bool $isFinal, // @phpstan-ignore property.onlyWritten
```
staabm: interessting that this PHPStan error shows up with this PR, because the propery was already only-written before this PR

staabm:
> PhpFunctionReflection too please.

here we go

conf/config.neon
```diff
@@ -302,6 +302,16 @@ services:
 		tags:
 			- phpstan.parser.richParserNodeVisitor
 
+	-
+		class: PHPStan\Parser\VariadicMethodsVisitor
+		tags:
+			- phpstan.parser.richParserNodeVisitor
```
ondrejmirtes, line 308: This isn't sufficient just for the RichParser, it has to be done in `SimpleParser` too. Because it's relevant for sources that aren't analysed too.

src/Parser/VariadicFunctionsVisitor.php
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		if ($node instanceof Node\Stmt\Function_) {
+			$functionName = $this->inNamespace !== null ? $this->inNamespace . '\\' . $node->name->name : $node->name->name;
+
+			if (!array_key_exists($functionName, $this->variadicFunctions)) {
+				$this->variadicFunctions[$functionName] = TrinaryLogic::createMaybe();
+			}
+
+			$isVariadic = $this->functionCallStatementFinder->findFunctionCallInStatements(ParametersAcceptor::VARIADIC_FUNCTIONS, $node->getStmts()) !== null;
```
ondrejmirtes: This causes to go through the AST multiple times. Instead, you could check `FuncCall`s when they are entered here in `enterNode`. After the refactoring `FunctionCallStatementFinder` can be deleted.

ondrejmirtes (changes requested):
Also, having SimpleParserTest unit test to check for the attributes existence would also be nice.

conf/config.neon
staabm: I am not sure I understand what you mean. 

I though doing https://github.com/phpstan/phpstan-src/pull/3534/files#diff-04a831854bbfa49602b74bbce1164b0f12d847200d418504d7dd11bf34702cc9R53-R54 is what you suggested initially.

do I need some neon config? I don't see any other config for a tag containing "simpleparser" or something

ondrejmirtes: You need to add the visitors in SimpleParser same way NameResolver is added there.

And yeah, it also needs to be in RichParser so the unit test should probably run for both.

Right now on 1.12.x variadic functions are properly detected in vendor code, but your PR breaks that.

staabm: > You need to add the visitors in SimpleParser same way NameResolver is added there.

I think I already did. Added tests which prove it. Hopefully I got you right ;).

src/Reflection/Php/PhpFunctionReflection.php
staabm: reproduced in https://github.com/phpstan/phpstan/issues/11802

src/Parser/VariadicMethodsVisitor.php
```diff
@@ -0,0 +1,108 @@
...
+		if ($this->topNode === null) {
+			$this->topNode = $node;
+		}
+
+		if ($node instanceof Node\Stmt\Namespace_ && $node->name !== null) {
+			$this->inNamespace = $node->name->toString();
+		}
+
+		if ($node instanceof Node\Stmt\ClassLike && $node->name instanceof Node\Identifier) {
+			$this->inClassLike = $this->inNamespace !== null ? $this->inNamespace . '\\' . $node->name->name : $node->name->name;
```
ondrejmirtes: This will break on anonymous classes. Please add some tests that would get thrown off by this logic, and then fix it by having an array stack instead of a single property value.

ondrejmirtes (changes requested):

src/Parser/VariadicMethodsVisitor.php
staabm: hmm. 

to properly support anonymous classes I would have to create a name key which I can later on reproduce in `PhpMethodReflection` so I can find the correct array-offsets again in https://github.com/phpstan/phpstan-src/pull/3534/files#diff-aa8ffb29754c75aa88928c72d3877f0f030a934d9cff263dfab527aaf4037172R256-R260

the name returned by `ClassReflection->getName()` [depends on the filename](https://github.com/phpstan/phpstan-src/blob/e2654b7dc87951e1481bc46d8f1f7ff587e3f484/src/Broker/AnonymousClassNameHelper.php#L23-L48). In this Visitor I don't have access to the file-name though.

when building a different key e.g. based on the AST-Node doesn't work either because in `PhpMethodReflection` I don't have access to the AST.

ondrejmirtes: It's fine if we don't find variadic methods inside anonymous classes (I don't think we do now either), it's sufficient if the code doesn't break on classes containing anonymous classes.

staabm: is this test good enough?

https://github.com/phpstan/phpstan-src/pull/3534/files#diff-c2d7f467d27d2e9f8c672d41471a89079412f952f7e686ef9d3fc7a431c8e782R15-R21

staabm: ohh I see the problem now

ondrejmirtes: We need a test that breaks the visitor.

Use-case: There's anonymous class in the middle of a class. After the visitor leaves the anonymous class, the class name is set to null. Which means that if there's a class method with `func_get_args()` call, the visitor is not going to correctly set the attribute for it.

staabm:
btw: On my mac with this PR we are ~1-2 seconds slower when running `time make phpstan` in comparison to before this PR - I will have another look why this is the case

src/Parser/VariadicMethodsVisitor.php
```diff
@@ -0,0 +1,133 @@
...
+			$this->topNode = $node;
+		}
+
+		if ($node instanceof Node\Stmt\Namespace_ && $node->name !== null) {
+			$this->inNamespace = $node->name->toString();
+		}
+
+		if (
+			$node instanceof Node\Stmt\Class_
+			|| $node instanceof Node\Stmt\ClassLike
```
ondrejmirtes: `$node instanceof Node\Stmt\ClassLike` would be sufficient

src/Parser/VariadicMethodsVisitor.php
```diff
@@ -0,0 +1,133 @@
...
+		) {
+			if (!$node->name instanceof Node\Identifier) {
+				$className = 'class@anonymous:' . $node->getStartLine();
+			} else {
+				$className = $node->name->name;
+			}
+
+			$this->classStack[] = $className;
+			$this->inClassLike = $this->inNamespace !== null ? $this->inNamespace . '\\' . implode('\\', $this->classStack) : implode('\\', $this->classStack);
```
ondrejmirtes: We don't need this property. We should always read the last entry in `classStack`

src/Parser/VariadicMethodsVisitor.php
```diff
@@ -0,0 +1,133 @@
...
+			$node instanceof ClassMethod
+			&& $this->inClassLike !== null
+		) {
+			$this->variadicMethods[$this->inClassLike][$node->name->name] ??= TrinaryLogic::createNo();
+			$this->inMethod = null;
+		}
+
+		if (
+			$node instanceof Node\Stmt\Class_
+			|| $node instanceof Node\Stmt\ClassLike
```
ondrejmirtes: dtto

tests/PHPStan/Parser/ParserTest.php
```diff
@@ -0,0 +1,98 @@
...
+
+	/**
+	 * @dataProvider dataVariadicCallLikes
+	 * @param array<string, TrinaryLogic>|array<string, array<string, TrinaryLogic>> $expectedVariadics
+	 * @throws ParserErrorsException
+	 */
+	public function testSimpleParserVariadicCallLikes(string $file, string $attributeName, array $expectedVariadics): void
+	{
+		/** @var RichParser $parser */
+		$parser = self::getContainer()->getService('currentPhpVersionSimpleParser');
```
ondrejmirtes, line 69: This `@var` is wrong

src/Parser/VariadicMethodsVisitor.php
```diff
@@ -0,0 +1,133 @@
...
+		if ($node instanceof Node\Stmt\Namespace_ && $node->name !== null) {
+			$this->inNamespace = $node->name->toString();
+		}
+
+		if (
+			$node instanceof Node\Stmt\Class_
+			|| $node instanceof Node\Stmt\ClassLike
+		) {
+			if (!$node->name instanceof Node\Identifier) {
+				$className = 'class@anonymous:' . $node->getStartLine();
```
ondrejmirtes: This isn't sufficiently unique. This was recently fixed elsewhere by https://github.com/phpstan/phpstan-src/pull/3328.

ondrejmirtes (changes requested):

src/Parser/VariadicMethodsVisitor.php
staabm: `AnonymousClassNode` is not contained in `ClassLike`

ondrejmirtes: Err, what? `final class AnonymousClassNode extends Class_`, `class Class_ extends ClassLike`.

src/Reflection/Php/PhpMethodReflection.php
```diff
@@ -253,81 +245,27 @@ private function isVariadic(): bool
...
+			$nodes = $this->parser->parseFile($filename);
+			if (count($nodes) > 0) {
+				$variadicMethods = $nodes[0]->getAttribute(VariadicMethodsVisitor::ATTRIBUTE_NAME);
 
-				$methodName = $node->name->name;
-				if ($methodName === $this->reflection->getName()) {
-					return $this->functionCallStatementFinder->findFunctionCallInStatements(ParametersAcceptor::VARIADIC_FUNCTIONS, $node->getStmts()) !== null;
+				if (
+					is_array($variadicMethods)
+					&& array_key_exists($declaringClass->getName(), $variadicMethods)
```
ondrejmirtes: This should do the `class@anonymous` thing that's done by the visitor too. I don't think we'll be able to replicate the `anonymous class index` here and we should do more something like a column instead.

ondrejmirtes (changes requested):

src/Parser/VariadicMethodsVisitor.php
```diff
@@ -0,0 +1,130 @@
...
+
+		if ($node instanceof Node\Stmt\Namespace_ && $node->name !== null) {
+			$this->inNamespace = $node->name->toString();
+		}
+
+		if ($node instanceof Node\Stmt\ClassLike) {
+			if (!$node->name instanceof Node\Identifier) {
+				$className = sprintf('class@anonymous:%s:%s', $node->getStartLine(), $node->getEndLine());
+				$this->classStack[] = $className;
+				$this->inClassLike = $className; // anonymous classes are in global namespace
```
staabm: because of this special case I had to re-introduce the `inClassLike` property to make sure the keys of a anonymous class don't contain a namespace. thats required so I am able to build a matching array-offset in `PhpMethodReflection`.

staabm:
with the latest push we are now able to properly handle `func_get_args` based variadics in anonymous classes, which we did not got right before this PR:

https://phpstan.org/r/5c9c464b-5bef-40c3-b902-c9891308a2f7

src/Reflection/Php/PhpMethodReflection.php
staabm: I did not find column information in the reflection api.
instead I used start/end-of line combination of the class definition for now.

ondrejmirtes:
Hey, I still saw too many things I'd do differently so I rolled up my sleeves and did them: https://github.com/phpstan/phpstan-src/pull/3534/commits/d4d5c1a79fc497dc17729fb790ce4a4f6c461776

No tests are failing for me. Feel free to write more tests if you think your code covered a scenario that mine doesn't. Thanks!

staabm:
thank you. thats way simpler - awesome.

src/Reflection/Php/PhpFunctionReflection.php
```diff
@@ -296,19 +252,4 @@ public function acceptsNamedArguments(): TrinaryLogic
 		return TrinaryLogic::createFromBoolean($this->acceptsNamedArguments);
 	}
 
-	private function isFunctionNodeVariadic(Function_ $node): bool
```
staabm, line 299: I wonder why this `$node->params` logic is no longer reflected in the patch. it seems it was dead code before

ondrejmirtes: It's really old code, maybe there just wasn't a test for it. Since we're in 2.0.x, we can risk it, and maybe introduce it back if it breaks someone's scenario.

staabm: I think it was unnecessary logic I implemented with https://github.com/phpstan/phpstan-src/commit/0ee67d58c66bf5c603cb4ba60baa91f3c5030f33#diff-3ac4e5d4a520618a490963b70e3fcd428d56aea61683641020393167fff2ca5d

deleting the lines on 2.0.x before this PR, also doesn't break tests :).

all good.

ondrejmirtes:
It's a bit slower but maybe we'll find opportunities elsewhere :) Thank you.

staabm:
thank you!

ondrejmirtes:
Oh yeah, I did it: https://github.com/phpstan/phpstan-src/commit/19cd999f252ce582061c26b126819103425c75be

Previously `VariadicMethodsVisitor::$cache` was not really useful, it only stored `true` values so the files were parsed over and over again.

ondrejmirtes:
In my measurements it's now fast as before without writing anything to disk.

PR #3535, opened 2024-10-04
Fix issue with template of union

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11663

Cf https://github.com/phpstan/phpstan/issues/11663#issuecomment-2380558907 I dunno if there is a already-known better way to handle such situation.

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -106,7 +106,6 @@ use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\Generic\TemplateTypeHelper;
 use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\Generic\TemplateTypeVarianceMap;
-use PHPStan\Type\Generic\TemplateUnionType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
 use PHPStan\Type\IntersectionType;
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -106,6 +106,7 @@ use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\Generic\TemplateTypeHelper;
 use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\Generic\TemplateTypeVarianceMap;
+use PHPStan\Type\Generic\TemplateUnionType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
 use PHPStan\Type\IntersectionType;
--- a/src/Type/UnionType.php
+++ b/src/Type/UnionType.php
@@ -92,12 +92,18 @@ class UnionType implements CompoundType
 	public function filterTypes(callable $filterCb): Type
 	{
 		$newTypes = [];
+		$changed = false;
 		foreach ($this->getTypes() as $innerType) {
 			if (!$filterCb($innerType)) {
+				$changed = true;
 				continue;
 			}
 
 			$newTypes[] = $innerType;
+		}
+
+		if (!$changed) {
+			return $this;
 		}
 
 		return TypeCombinator::union(...$newTypes);
--- a/src/Type/UnionType.php
+++ b/src/Type/UnionType.php
@@ -22,6 +22,7 @@ use PHPStan\TrinaryLogic;
 use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\Generic\TemplateMixedType;
 use PHPStan\Type\Generic\TemplateType;
+use PHPStan\Type\Generic\TemplateTypeFactory;
 use PHPStan\Type\Generic\TemplateTypeMap;
 use PHPStan\Type\Generic\TemplateTypeVariance;
 use PHPStan\Type\Generic\TemplateUnionType;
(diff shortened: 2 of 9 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -5602,8 +5603,10 @@ private function filterTypeWithMethod(Type $typeWithMethod, string $methodName):
 	{
 		if ($typeWithMethod instanceof UnionType) {
 			$newTypes = [];
+			$oneTypeHasBeenRemoved = false;
```
staabm: I think whether this is a problem which should be solved with a `Type` method.

maybe we can use a `Type` method in which you pass in a callable and it returns a preserved union type (maybe such method even exists...?)

VincentLanglet: The whole method `filterTypeWithMethod` could be introduce as something like `Type::filter` if wanted but this would "only" be a refacto. (Which may be in another PR)

There is UnionType::traverse
```
public function traverse(callable $cb): Type
	{
		$types = [];
		$changed = false;

			$newType = $cb($type);
			if ($type !== $newType) {
				$changed = true;
			}
			$types[] = $newType;
		}

		if ($changed) {
			return TypeCombinator::union(...$types);
		}

		return $this;
	}
```
but it doesn't do the exact same thing ; we can still notice a `changed` check is done the same way.

herndlm: don't want to disturb you guys too much, but another idea to throw in - does `TypeTraverser::map()` help in any way here maybe? e.g. the types you don't want you could replace with a `NeverType` potentially? that should leave the outer `TemplateUnionType` untouched at least. that could be a replacement for the foreach, AFAIK `TypeCombinator::union()` should keep the TemplateUnionType then and remove the never I suppose.

but not sure if that makes it even more complicated potentially 😅

VincentLanglet:
Hi, what do you think about this PR @ondrejmirtes ?

Thanks

ondrejmirtes:
I have no idea how `$oneTypeHasBeenRemoved` solves a problem so I don't want to merge it right now.

VincentLanglet:
> I have no idea how `$oneTypeHasBeenRemoved` solves a problem so I don't want to merge it right now.

The idea was kinda similar to `UnionType::traverse`
```
public function traverse(callable $cb): Type
	{
		$types = [];
		$changed = false;

			$newType = $cb($type);
			if ($type !== $newType) {
				$changed = true;
			}
			$types[] = $newType;
		}

		if ($changed) {
			return TypeCombinator::union(...$types);
		}

		return $this;
	}
```
the goal is to keep the same object when nothing change (here when no type is removed).


Because the existing code
```
if ($typeWithMethod instanceof UnionType) {
			$newTypes = [];
            foreach ($typeWithMethod->getTypes() as $innerType) {
				if (!$innerType->hasMethod($methodName)->yes()) {
					continue;
				}

				$newTypes[] = $innerType;
			}
			if (count($newTypes) === 0) {
				return null;
			}
			$typeWithMethod = TypeCombinator::union(...$newTypes);
```
always change the instance of `typeWithMethod` when it's an UnionType, which means:
- BenevolentUnion become an UnionType
- TemplateUnion become an UnionType

But when `oneTypeHasBeenRemoved = false` we could keep the initial instance of UnionType since we keep all the subTypes. This way, for the existing issue, we keep the TemplateUnion we all the needed informations.

src/Analyser/MutatingScope.php
ondrejmirtes: I get it, but the variable name should be different. Something like `$unionTypeHasChanged`.

Also it's a lot of repeated code. I'd welcome some refactoring first, and then the change should be done in a single place.

ondrejmirtes (changes requested):

VincentLanglet:
PR is simplified now

ondrejmirtes:
We could improve this further, right? Even if something changes in UnionType subtype, in this method we could make sure the correct class is created.

The subtypes from the top of my head are:

* BenevolentUnionType
* TemplateUnionType
* TemplateBenevolentUnionType

VincentLanglet:
> We could improve this further, right? Even if something changes in UnionType subtype, in this method we could make sure the correct class is created.
> 
> The subtypes from the top of my head are:
> 
> * BenevolentUnionType
> * TemplateUnionType
> * TemplateBenevolentUnionType

I understand the use case for BenevolentUnion (__benevolent(A|B|C) become __benevolent(A|B)), but it's complicated for me to see the use case for TemplateUnionType and TemplateBenevolentUnionType.
If we have `T of A|B|C` it feel wrong to change this to `T of A|B` isn't it ?

Do you have an example to add to the tests ?

ondrejmirtes:
Examples:

* TemplateUnionType: https://phpstan.org/r/f7585d6f-e7e8-4a4f-ad70-52c8d5a28b2c (it should not report "Function foo() should return T of A|B but returns A.")
* TemplateBenevolentUnionType: https://phpstan.org/r/c9a5c400-94e9-4c67-9636-2fc35b219156

src/Type/UnionType.php
```diff
@@ -92,15 +93,36 @@ public function getTypes(): array
...
-		return TypeCombinator::union(...$newTypes);
+		if (!$changed) {
+			return $this;
+		}
+
+		$result = TypeCombinator::union(...$newTypes);
+		if ($this instanceof BenevolentUnionType) {
+			$result = TypeUtils::toBenevolentUnion($result);
+		}
+		if ($this instanceof TemplateType) {
```
ondrejmirtes: I'd rather have this logic in the subtypes in overridden methods.

ondrejmirtes (changes requested):

src/Type/UnionType.php
VincentLanglet: Done

ondrejmirtes:
Thank you.

PR #3536, opened 2024-10-04
Use precise Verbosity level for getArgumentsDescription

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11799

At first I discover the issue for `lowercase-string` and feel like the message could be improved for other situation like
```
Call to function is_string() with mixed will always evaluate to false.
```
which could explain that string is not in the mixed type, and become 
```
Call to function is_string() with mixed~string will always evaluate to false.
```
but maybe it's voluntary...

So currently I dunno:
- If we consider there is no issue, and `Call to function in_array() with arguments string, array{'publishDate', 'approvedAt', 'allowedValues'} and true will always evaluate to false.` is an OK message. Then I can close both the PR and the issue
- If this PR is OK to be merged
- If this PR should be done on 2.0 since it's kinda a BC break
- If we consider the behavior should only be fixed for lowercase-string (how ? I assume this fix is harder)
- If the fix should be somewhere else... (Maybe in 2.0 the Verbosity of lowercase-string should become value)

Your call @ondrejmirtes

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -343,7 +343,7 @@ final class ImpossibleCheckTypeHelper
 			return '';
 		}
 
-		$descriptions = array_map(fn (Arg $arg): string => ($this->treatPhpDocTypesAsCertain ? $scope->getType($arg->value) : $scope->getNativeType($arg->value))->describe(VerbosityLevel::value()), $args);
+		$descriptions = array_map(fn (Arg $arg): string => ($this->treatPhpDocTypesAsCertain ? $scope->getType($arg->value) : $scope->getNativeType($arg->value))->describe(VerbosityLevel::precise()), $args);
 
 		if (count($descriptions) < 3) {
 			return sprintf(' with %s', implode(' and ', $descriptions));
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
> I assume this fix is harder

You're right, the fix is harder. It's hard for a type-specifying extension like `in_array`, because we'd have to somehow track it from ImpossibleCheckTypeHelper to ImpossibleCheckTypeFunctionCallRule and tell that the reason is `Type::isLowercaseString()`.

Relevant: 1.12.x has `Type::acceptsWithReason()`. It'd be beneficial to also have `Type::isSuperTypeOfWithReason()`.

In 2.0.x the original `Type::accepts()` is deleted and the new `Type::acceptsWithReason()` gets renamed to `Type::accepts()`.

VincentLanglet:
> Relevant: 1.12.x has `Type::acceptsWithReason()`. It'd be beneficial to also have `Type::isSuperTypeOfWithReason()`.
> 
> In 2.0.x the original `Type::accepts()` is deleted and the new `Type::acceptsWithReason()` gets renamed to `Type::accepts()`.

I tried to introduce this method then https://github.com/phpstan/phpstan-src/pull/3538

PR #3538, opened 2024-10-04
Introduce isSuperTypeOfWithReason

VincentLanglet:
As suggested by https://github.com/phpstan/phpstan-src/pull/3536#issuecomment-2392985397

### DIFF
--- a/src/Type/IntegerRangeType.php
+++ b/src/Type/IntegerRangeType.php
@@ -15,6 +15,7 @@ use PHPStan\Type\Accessory\AccessoryNumericStringType;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use function array_filter;
+use function array_map;
 use function assert;
 use function ceil;
 use function count;
@@ -209,7 +210,7 @@ class IntegerRangeType extends IntegerType implements CompoundType
 	public function acceptsWithReason(Type $type, bool $strictTypes): AcceptsResult
 	{
 		if ($type instanceof parent) {
-			return new AcceptsResult($this->isSuperTypeOf($type), []);
+			return $this->isSuperTypeOfWithReason($type)->toAcceptsResult();
 		}
 
 		if ($type instanceof CompoundType) {
@@ -221,6 +222,11 @@ class IntegerRangeType extends IntegerType implements CompoundType
 
 	public function isSuperTypeOf(Type $type): TrinaryLogic
 	{
+		return $this->isSuperTypeOfWithReason($type)->result;
+	}
+
+	public function isSuperTypeOfWithReason(Type $type): IsSuperTypeOfResult
+	{
 		if ($type instanceof self || $type instanceof ConstantIntegerType) {
 			if ($type instanceof self) {
 				$typeMin = $type->min;
@@ -231,48 +237,53 @@ class IntegerRangeType extends IntegerType implements CompoundType
 			}
 
 			if (self::isDisjoint($this->min, $this->max, $typeMin, $typeMax)) {
-				return TrinaryLogic::createNo();
+				return IsSuperTypeOfResult::createNo();
 			}
 
 			if (
 				($this->min === null || $typeMin !== null && $this->min <= $typeMin)
 				&& ($this->max === null || $typeMax !== null && $this->max >= $typeMax)
 			) {
-				return TrinaryLogic::createYes();
+				return IsSuperTypeOfResult::createYes();
 			}
 
-			return TrinaryLogic::createMaybe();
+			return IsSuperTypeOfResult::createMaybe();
 		}
 
 		if ($type instanceof parent) {
-			return TrinaryLogic::createMaybe();
+			return IsSuperTypeOfResult::createMaybe();
 		}
 
 		if ($type instanceof CompoundType) {
-			return $type->isSubTypeOf($this);
+			return $type->isSubTypeOfWithReason($this);
 		}
 
-		return TrinaryLogic::createNo();
+		return IsSuperTypeOfResult::createNo();
 	}
 
 	public function isSubTypeOf(Type $otherType): TrinaryLogic
 	{
+		return $this->isSubTypeOfWithReason($otherType)->result;
+	}
+
+	public function isSubTypeOfWithReason(Type $otherType): IsSuperTypeOfResult
+	{
 		if ($otherType instanceof parent) {
-			return $otherType->isSuperTypeOf($this);
+			return $otherType->isSuperTypeOfWithReason($this);
 		}
 
 		if ($otherType instanceof UnionType) {
-			return $this->isSubTypeOfUnion($otherType);
+			return $this->isSubTypeOfUnionWithReason($otherType);
 		}
 
 		if ($otherType instanceof IntersectionType) {
-			return $otherType->isSuperTypeOf($this);
+			return $otherType->isSuperTypeOfWithReason($this);
 		}
 
-		return TrinaryLogic::createNo();
+		return IsSuperTypeOfResult::createNo();
 	}
 
-	private function isSubTypeOfUnion(UnionType $otherType): TrinaryLogic
+	private function isSubTypeOfUnionWithReason(UnionType $otherType): IsSuperTypeOfResult
 	{
 		if ($this->min !== null && $this->max !== null) {
 			$matchingConstantIntegers = array_filter(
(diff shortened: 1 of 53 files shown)


src/Type/AcceptsResult.php
```diff
@@ -118,14 +126,30 @@ public static function maxMin(self ...$operands): self
...
 
 		$result = TrinaryLogic::maxMin(...array_map(static fn (self $result) => $result->result, $operands));
+
+		return new self($result, self::mergeReasons($operands));
+	}
+
+	public function negate(): self
+	{
+		return new self($this->result->negate(), $this->reasons);
+	}
```
VincentLanglet: I dunno if it makes sens to have a negate method (Should the reasons be lost ?)

I created one since I had multiple Trinary::negate calls to replace.

src/Type/IntegerRangeType.php
```diff
@@ -281,11 +291,11 @@ private function isSubTypeOfUnion(UnionType $otherType): TrinaryLogic
 			);
 
 			if (count($matchingConstantIntegers) === ($this->max - $this->min + 1)) {
-				return TrinaryLogic::createYes();
+				return AcceptsResult::createYes();
 			}
 		}
 
-		return TrinaryLogic::createNo()->lazyOr($otherType->getTypes(), fn (Type $innerType) => $this->isSubTypeOf($innerType));
+		return AcceptsResult::createNo()->or(...array_map(fn (Type $innerType) => $this->isSubTypeOfWithReason($innerType), $otherType->getTypes()));
```
VincentLanglet: Here we're loosing a `lazyOr` into a `or`.

I dunno if the lazy could be kept on AcceptsResults but I saw similar lazy lost when `isAcceptedWithReasonBy` was introduced

src/Type/IntersectionType.php
```diff
@@ -221,28 +221,38 @@ public function acceptsWithReason(Type $otherType, bool $strictTypes): AcceptsRe
...
+			return AcceptsResult::createYes();
 		}
 
 		if ($otherType instanceof NeverType) {
-			return TrinaryLogic::createYes();
+			return AcceptsResult::createYes();
 		}
 
-		return TrinaryLogic::createYes()->lazyAnd($this->getTypes(), static fn (Type $innerType) => $innerType->isSuperTypeOf($otherType));
+		return AcceptsResult::createYes()->and(...array_map(static fn (Type $innerType) => $innerType->isSuperTypeOfWithReason($otherType), $this->types));
```
VincentLanglet: Here we're loosing a `lazyAnd` into a `and`.

I dunno if the lazy could be kept on AcceptsResults but I saw similar lazy lost when `isAcceptedWithReasonBy` was introduced

src/Type/IntersectionType.php
```diff
@@ -221,28 +221,38 @@ public function acceptsWithReason(Type $otherType, bool $strictTypes): AcceptsRe
...
+
+	public function isSubTypeOfWithReason(Type $otherType): AcceptsResult
 	{
 		if (($otherType instanceof self || $otherType instanceof UnionType) && !$otherType instanceof TemplateType) {
-			return $otherType->isSuperTypeOf($this);
+			return $otherType->isSuperTypeOfWithReason($this);
 		}
 
-		$result = TrinaryLogic::lazyMaxMin($this->getTypes(), static fn (Type $innerType) => $otherType->isSuperTypeOf($innerType));
+		$result = AcceptsResult::maxMin(...array_map(static fn (Type $innerType) => $otherType->isSuperTypeOfWithReason($innerType), $this->types));
```
VincentLanglet: Here we're loosing a `lazyMaxMin ` into a `maxMin`.

I dunno if the lazy could be kept on AcceptsResults but I saw similar lazy lost when `isAcceptedWithReasonBy` was introduced (In this same class for instance).

src/Type/UnionType.php
```diff
@@ -214,24 +219,29 @@ public function isSuperTypeOf(Type $otherType): TrinaryLogic
 			|| $otherType instanceof ConditionalTypeForParameter
 			|| $otherType instanceof IntegerRangeType
 		) {
-			return $otherType->isSubTypeOf($this);
+			return $otherType->isSubTypeOfWithReason($this);
 		}
 
-		$result = TrinaryLogic::createNo()->lazyOr($this->getTypes(), static fn (Type $innerType) => $innerType->isSuperTypeOf($otherType));
```
VincentLanglet, line 220: Same, lazyOr into or.

src/Type/UnionType.php
```diff
@@ -214,24 +219,29 @@ public function isSuperTypeOf(Type $otherType): TrinaryLogic
...
-			return $result->or($otherType->isSubTypeOf($this));
+			return $result->or($otherType->isSubTypeOfWithReason($this));
 		}
 
 		return $result;
 	}
 
 	public function isSubTypeOf(Type $otherType): TrinaryLogic
 	{
-		return TrinaryLogic::lazyExtremeIdentity($this->getTypes(), static fn (Type $innerType) => $otherType->isSuperTypeOf($innerType));
```
VincentLanglet, line 234: Same `lazyExtremeIdentity` into `extremeIdentity`, like it's done for isAcceptedWithReasonBy

ondrejmirtes:
This looks promising. Could you do a quick proof of concept where this might be useful?

In StrictComparisonOfDifferentTypesRule we could read the reasons and output them as a tip (similar to what places that call `Type::acceptsWithReason()` do). A great opportunity for that is `MixedType::isSuperTypeOfWithReason()` when there's a subtracted type. So instead of `mixed~string === string is always false`, the error message would stay the same, but additionally there'd be a tip about "Type string has been eliminated from mixed already."

ondrejmirtes:
I understand it might be a bit challenging because we might have to propagate that all the way from `initializerExprTypeResolver->resolveIdenticalType`.

ondrejmirtes:
Maybe better to leave it for later, so this can be merged. So nevermind for now, I'll review this PR.

VincentLanglet:
> Maybe better to leave it for later, so this can be merged. So nevermind for now, I'll review this PR.

Yes I had in mind to work on it step by step ^^'

ondrejmirtes (changes requested):
Hey, I was thinking about this a bit more and it'd help me if:

1) We didn't reuse AcceptsResult for the `isSuperTypeOfWithReason` return type.
2) This PR did not do any changes to AcceptsResult.
3) Instead, it introduced new class called `IsSuperTypeOfResult` by copying and modifying AcceptsResult to suit the method's needs.

After that I'm very likely to merge this.

VincentLanglet:
Since I created isSubtypeOfWithReason too, does it mean we need a IsSubtypeOfResult too @ondrejmirtes ?

This may complicate things since i need to trasnform often subtypeOfResult into superTypeOfResult.

ondrejmirtes:
@VincentLanglet It's fine to use IsSuperTypeOfResult for isSubtypeOf too.

VincentLanglet:
> 1. We didn't reuse AcceptsResult for the `isSuperTypeOfWithReason` return type.
> 2. This PR did not do any changes to AcceptsResult.
> 3. Instead, it introduced new class called `IsSuperTypeOfResult` by copying and modifying AcceptsResult to suit the method's needs.

I'll end with something very similar for AcceptsResult and IsSuperTypeOfResult.
They will have the same and, or, ... methods.

I was wondering what's the idea between duplicating such code ; wouldn't it better to introduce something like a Result class which works for everything ?

It would make sens because there is code like 
```
public function acceptsWithReason(Type $type, bool $strictTypes): AcceptsResult
	{
		if ($type instanceof CompoundType && !$type instanceof self) {
			return $type->isAcceptedWithReasonBy($this, $strictTypes);
		}

		return $this->isSuperTypeOfInternal($type, true);
	}
	
public function isSuperTypeOfWithReason(Type $type): IsSuperTypeOfResult
	{
		if ($type instanceof CompoundType && !$type instanceof self) {
			return $type->isSubTypeOfWithReason($this);
		}

		return $this->isSuperTypeOfInternal($type, false);
	}
```

So both method rely on isSuperTypeOfInternal and could use the same value object as return type.

Another "weird" situation is the fact that isParametersAcceptorSuperTypeOf currently returns an AcceptsResult when the method talk about SuperTypeOf...
https://github.com/phpstan/phpstan-src/blob/c77064a09e95cf8171e38fccb189ae53cf6523e5/src/Type/CallableTypeHelper.php#L15-L19

So I'm not sure how to evolve this PR with the use of a IsSuperTypeOf without using a lot of method like
`IsSuperTypeOfResult::toAcceptsResult` and `AcceptsResult::ToIsSuperTypeOfResult`.

ondrejmirtes:
> I was wondering what's the idea between duplicating such code ; wouldn't it better to introduce something like a Result class which works for everything ?

This is accidental duplication. The fact these are the same right now doesn't mean they're going to be the same in the future. I tend to develop a hunch for these things.

ondrejmirtes:
> Another "weird" situation is the fact that isParametersAcceptorSuperTypeOf currently returns an AcceptsResult when the method talk about SuperTypeOf...

This method should return IsSuperTypeOfResult in your PR, because it's used in `isSuperTypeOfInternal` methods.

ondrejmirtes:
Hey, I need this really soon so I can announce an RC period for 2.0. I'm changing this locally right now.

VincentLanglet:
> Hey, I need this really soon so I can announce an RC period for 2.0. I'm changing this locally right now.

I started the changes, do you want me to stop and letting u doing all the changes then ? (Or I push the current changes ?)

ondrejmirtes:
Thank you.

VincentLanglet:
Thanks you too !

PR #3539, opened 2024-10-04
Fix unused private property is not sometimes detected

staabm:
closes https://github.com/phpstan/phpstan/issues/11802


staabm:
Just realized we have the same bug in `UnusedPrivateMethodRule` - hold my beer

ondrejmirtes:
Thank you!

PR #3540, opened 2024-10-06
`token_name()` returns non-empty-string

staabm:
closes https://github.com/phpstan/phpstan/issues/11808

tested locally with the snippet of the issue


ondrejmirtes:
Thank you!

PR #3541, opened 2024-10-06
More precise md5/sha1 return type

staabm:
utilizing lowercase-string for this old school hashing functions

//cc @VincentLanglet


VincentLanglet:
If I understand correctly

You are using numeric-string in case only numbers are used ?
But numbers like 1234 is considered as a numeric string lowercased. (Only things like 10E2 are not but I dont think there is uppercase letter in sha or md5)

So it's non-falsy&numeric&lowercase|non-falsy&lowercase
Which can be simplified to non-falsy&lowercase.

I dont think there is a benefit to specify that sometimes it's a fully numeric string

staabm:
great feedback. I like the simplified version a lot more.

I am not yet used to thinking about numbers to be lowercase strings ;).

this also means we can do similar stuff in the `hash()` dynamic return type extension when `$algo` is a constant string known

resources/functionMap.php
```diff
@@ -6403,8 +6403,8 @@
 'mcrypt_module_open' => ['resource|false', 'cipher'=>'string', 'cipher_directory'=>'string', 'mode'=>'string', 'mode_directory'=>'string'],
 'mcrypt_module_self_test' => ['bool', 'algorithm'=>'string', 'lib_dir='=>'string'],
 'mcrypt_ofb' => ['string', 'cipher'=>'string', 'key'=>'string', 'data'=>'string', 'mode'=>'int', 'iv='=>'string'],
-'md5' => ['non-falsy-string', 'str'=>'string', 'raw_output='=>'bool'],
-'md5_file' => ['non-falsy-string|false', 'filename'=>'string', 'raw_output='=>'bool'],
+'md5' => ['(non-falsy-string&lowercase-string)', 'str'=>'string', 'raw_output='=>'bool'],
```
ondrejmirtes: No need for the parentheses here.

ondrejmirtes (changes requested):

VincentLanglet (approved):

ondrejmirtes:
Thank you!

VincentLanglet:
@staabm I'm not familiar with sha256 ; should the return type be updated too ?

staabm:
I am not familiar either. 

Tested a few inputs on https://emn178.github.io/online-tools/sha256.html and it seems you are right

VincentLanglet:
> I am not familiar either.
> 
> Tested a few inputs on [emn178.github.io/online-tools/sha256.html](https://emn178.github.io/online-tools/sha256.html) and it seems you are right

I'll do the PR as soon as I have confirmation this function really exists (cf my interrogation https://github.com/phpstan/phpstan/issues/12048#issuecomment-2496156102)

PR #3542, opened 2024-10-06
Get rid of unnecessary `instanceof self` in `ConstantArrayType`

herndlm:
Gets rid of now unnecessary `instanceof self` in `ConstantArrayType`  by slightly improving `sliceArray()` which made it possible to remove `reindex()` completely :)

Contains changes of https://github.com/phpstan/phpstan-src/pull/3406 and should be merged after that one came into 2.0.x. Should apply cleanly, if not I can update it of course.


ondrejmirtes:
Thank you.

ondrejmirtes:
> Contains changes of https://github.com/phpstan/phpstan-src/pull/3406 and should be merged after that one came into 2.0.x. Should apply cleanly, if not I can update it of course.

Don't do this please next time. I smashed the green button without thinking. You should have sent this PR after I merged #3406.

herndlm:
yes, will refrain from doing this next time. ah I saw you reverted it again, should I re-create the core change?

PR #3545, opened 2024-10-07
Remove $isFinal dead-code in PhpFunctionReflection

staabm:
its only dead code in 2.0.x.

extracted from https://github.com/phpstan/phpstan-src/pull/3534


ondrejmirtes:
Thank you.

PR #3548, opened 2024-10-07
Reflection `getAttributes` returns `list`

ManuelHu:
constructed attributes returned by `ReflectionX->getAttributes()` are always added with the next array indices:
* https://github.com/php/php-src/blob/85cf4a39d744ab4a30a0815ac61b347b8d155f94/ext/reflection/php_reflection.c#L1193-L1194
* https://github.com/php/php-src/blob/85cf4a39d744ab4a30a0815ac61b347b8d155f94/ext/reflection/php_reflection.c#L1225-L1226

So the return type should be `list<...>` instead of `array<mixed,,...>`.


ondrejmirtes:
Thank you!

PR #3549, opened 2024-10-07
Add ArrayChangeKeyCaseFunctionReturnTypeExtension

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11795
Closes https://github.com/phpstan/phpstan/issues/10960

cc @staabm

### DIFF
--- a/src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
@@ -106,21 +106,20 @@ final class ArrayChangeKeyCaseFunctionReturnTypeExtension implements DynamicFunc
 				}
 
 				if ($type->isString()->yes()) {
-					if ($case === CASE_LOWER) {
-						return TypeCombinator::intersect($type, new AccessoryLowercaseStringType());
-					} elseif ($type->isLowercaseString()->yes()) {
-						$types = [new StringType()];
-						if ($type->isNonFalsyString()->yes()) {
-							$types[] = new AccessoryNonFalsyStringType();
-						} elseif ($type->isNonEmptyString()->yes()) {
-							$types[] = new AccessoryNonEmptyStringType();
-						}
-						if ($type->isNumericString()->yes()) {
-							$types[] = new AccessoryNumericStringType();
-						}
-
-						return TypeCombinator::intersect(...$types);
+					$types = [new StringType()];
+					if ($type->isNonFalsyString()->yes()) {
+						$types[] = new AccessoryNonFalsyStringType();
+					} elseif ($type->isNonEmptyString()->yes()) {
+						$types[] = new AccessoryNonEmptyStringType();
 					}
+					if ($type->isNumericString()->yes()) {
+						$types[] = new AccessoryNumericStringType();
+					}
+					if ($case === CASE_LOWER) {
+						$types[] = new AccessoryLowercaseStringType();
+					}
+
+					return TypeCombinator::intersect(...$types);
 				}
 
 				return $type;
--- a/src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
@@ -7,6 +7,9 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\Accessory\AccessoryLowercaseStringType;
+use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
+use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
+use PHPStan\Type\Accessory\AccessoryNumericStringType;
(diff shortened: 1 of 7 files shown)


src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,126 @@
...
+					if ($case === CASE_LOWER) {
+						return TypeCombinator::intersect($type, new AccessoryLowercaseStringType());
+					} elseif ($type->isLowercaseString()->yes()) {
+						return TypeCombinator::intersect(
+							new StringType(),
+							...array_filter(
+								TypeUtils::getAccessoryTypes($type),
+								static fn (Type $accessory): bool => !$accessory instanceof AccessoryLowercaseStringType,
+							),
+						);
```
VincentLanglet: I dunno if there is a better way to remove an accessory ?

staabm: AFAIR we usually build the type from zero instead of removing existing accessories

staabm: AFAIR we usually build the type from zero instead of removing existing accessories

VincentLanglet: Indeed, it's better since I don't keep the literal accessory too.

src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,156 @@
...
+						...array_map(
+							fn (ConstantStringType $type): Type => $this->mapConstantString($type, $case),
+							$constantStrings,
+						),
+					);
+				}
+
+				if ($type->isString()->yes()) {
+					if ($case === CASE_LOWER) {
+						return TypeCombinator::intersect($type, new AccessoryLowercaseStringType());
```
staabm: Lower can still be combined with non-empty/non-false ?

VincentLanglet: It was already the case because I intersect.
But I rework this because I also need to lose the literal Accessory.

staabm (approved):
lgtm - thanks

VincentLanglet:
Friendly ping @ondrejmirtes if you have time.

This might be interesting to be merged to be updated in the uppercase-string PR https://github.com/phpstan/phpstan-src/pull/3613

VincentLanglet:
I updated the PR to also support `uppercase-string` @ondrejmirtes

src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,158 @@
...
+			}
+		}
+
+		$constantArrays = $arrayType->getConstantArrays();
+		if (count($constantArrays) > 0) {
+			$arrayTypes = [];
+			foreach ($constantArrays as $constantArray) {
+				$newConstantArrayBuilder = ConstantArrayTypeBuilder::createEmpty();
+				foreach ($constantArray->getKeyTypes() as $i => $keyType) {
+					$valueType = $constantArray->getOffsetValueType($keyType);
```
ondrejmirtes: Calling getOffsetValueType is too expensive. You can `$valueTypes = $constantArray->getValueTypes()` and then access `$i` to get it.

ondrejmirtes (changes requested):

src/Type/Php/ArrayChangeKeyCaseFunctionReturnTypeExtension.php
VincentLanglet: Changed

ondrejmirtes:
Thank you.

ondrejmirtes:
Would be nice to add a similar extension for `mb_convert_case`, would you look into it? Context: https://github.com/phpstan/phpstan/discussions/12452

Thanks!

VincentLanglet:
> Would be nice to add a similar extension for `mb_convert_case`, would you look into it? Context: [phpstan/phpstan#12452](https://github.com/phpstan/phpstan/discussions/12452)
> 
> Thanks!

Do you have a reproducer ?

Because, for me, the mb_convert_case extension is handled by https://github.com/phpstan/phpstan-src/blob/e2b5a60fa52460ba3addd519b07e2244809878ee/src/Type/Php/StrCaseFunctionsReturnTypeExtension.php#L45

And we can see in the history that it was handled for lowercase-string and upper-case string
https://github.com/phpstan/phpstan-src/commits/e2b5a60fa52460ba3addd519b07e2244809878ee/src/Type/Php/StrCaseFunctionsReturnTypeExtension.php

ondrejmirtes:
I'm sorry, I haven't checked personally and trusted the user. PHPStan already handles it well https://phpstan.org/r/fad685d0-5331-4af2-8bb9-d0d1a0f81d00.

VincentLanglet:
Yes, it works fine https://phpstan.org/r/b269ff18-e0b8-42d4-9586-87edaee9fccf

PR #3552, opened 2024-10-08
Get rid of unnecessary `instanceof self` in `ConstantArrayType`

herndlm:
Another try of https://github.com/phpstan/phpstan-src/pull/3542 which was reverted. Since the precondition  https://github.com/phpstan/phpstan-src/pull/3406 was merged we can't see the array_reverse changes anymore in the diff.

Sorry for having that in there in the first place, I was impatient.

This removes `ConstantArrayType::reindex()` which was a bad design decision from my side from the beginning.


ondrejmirtes:
Thank you.

PR #3553, opened 2024-10-08
Support CallLike with variable method name

staabm:
closes https://github.com/phpstan/phpstan/issues/2920

### DIFF
--- a/src/Rules/Methods/CallMethodsRule.php
+++ b/src/Rules/Methods/CallMethodsRule.php
@@ -8,7 +8,10 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Internal\SprintfHelper;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Rules\FunctionCallParametersCheck;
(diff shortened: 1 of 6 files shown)


staabm:
just now saw the feedback in https://github.com/phpstan/phpstan/issues/2920#issuecomment-2399602921 - will consider

src/Rules/Methods/CallMethodsRule.php
```diff
@@ -31,12 +34,33 @@ public function getNodeType(): string
 
 	public function processNode(Node $node, Scope $scope): array
 	{
-		if (!$node->name instanceof Node\Identifier) {
-			return [];
+		if ($node->name instanceof Node\Identifier) {
+			$methodNames = [$node->name->name];
+		} else {
```
staabm, line 39: I see there are a lot of rules which early exit on :

```
if (!$node->name instanceof Node\Identifier) {
			return [];
}
```

maybe these are easy to enhance with a similar check I did here.
(just supporting the `getConstantStrings()` case)

staabm:
the issue targeted with this PR was resolved in the meantime

PR #1628, opened 2022-08-16
Use argument types as parameter types for inline closures (#7798)

leongersen:
Implements [#7798](https://github.com/phpstan/phpstan/issues/7798). Thank you for your advise, it was very helpful.

> The relevant code for this in NodeScopeResolver is this but you won't have to touch that for this feature request, it works in a general sense: https://github.com/phpstan/phpstan-src/blob/4b6d9647784966438c742b367066656cf269a8d4/src/Analyser/NodeScopeResolver.php#L2964-L2975

This unfortunately is not used for this case; `$passedToType` is only passed to `processClosureNode` when handling a closure as a function argument:
https://github.com/phpstan/phpstan-src/blob/4b6d9647784966438c742b367066656cf269a8d4/src/Analyser/NodeScopeResolver.php#L3223-L3226

https://github.com/phpstan/phpstan-src/blob/4b6d9647784966438c742b367066656cf269a8d4/src/Analyser/NodeScopeResolver.php#L2220-L2221

Because the closure is never the argument (it is the `name` of the `FuncCall`), this case is slightly different than for `array_map`, as the `Expr\Closure` is never handled by `ParametersAcceptorSelector::selectFromArgs` (only its `ClosureType` is).

Please let me know if I misunderstood.

PR #3554, opened 2024-10-08
Use argument types as parameter types for inline closures when assigned

ruudk:
The functionality was introduced in #1628. It works. But as soon as you use an inline assign expression it breaks. Let's support this case too.

Sometimes, you want to call something inline and also use the callback later.


ruudk:
@ondrejmirtes This one is ready for review. 

I don't understand the failing tests related to the issues. It complains that it cannot find level10.neon.

ondrejmirtes:
@ruudk We can't do much about that. Those issues were created with 2.0.x playground on level 10 but here we are in 1.12.x

ondrejmirtes:
Thank you!

ondrejmirtes:
Solved https://github.com/phpstan/phpstan-src/pull/3568

PR #3555, opened 2024-10-08
Make dumpType() more consistent with PHPDoc types

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/11823


src/Type/Constant/ConstantStringType.php
```diff
@@ -139,12 +139,7 @@ function (): string {
 
 	private function export(string $value): string
 	{
-		$escapedValue = addcslashes($value, "\0..\37");
-		if ($escapedValue !== $value) {
-			return '"' . addcslashes($value, "\0..\37\\\"") . '"';
```
zonuexe, line 144: The escape of DEL (`\177`) was missing.

ondrejmirtes:
Hi, thank you for this effort, but I'm closing it because of: https://github.com/phpstan/phpstan/issues/11823#issuecomment-2401526639

In this hypothetical `\PHPStan\dumpPhpDocType()` function that you can contribute, you don't need to reimplement all the escaping and so on, you can just call `Type::toPhpDocNode()` and print it with `Printer` from phpdoc-parser package.

PR #3558, opened 2024-10-09
Improve return type of token_name() and PhpToken::getTokenName()

zonuexe:
The return value of [`token_name()`](https://www.php.net/token_name) and [`PhpToken::getTokenName()`](https://www.php.net/manual/phptoken.gettokenname.php) is neither `''` nor `'0'`, so they can be typed more precisely.

refs https://github.com/phpstan/phpstan/issues/11808, https://github.com/phpstan/phpstan-src/pull/3540


ondrejmirtes:
Thank you!

PR #3559, opened 2024-10-09
Add PHPStan\dumpPhpDocType()

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/11823, refs https://github.com/phpstan/phpstan-src/pull/3555

### DIFF
--- a/src/Rules/Debug/DumpPhpDocTypeRule.php
+++ b/src/Rules/Debug/DumpPhpDocTypeRule.php
@@ -1,0 +1,58 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Debug;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function count;
+use function sprintf;
+use function strtolower;
+
+/**
+ * @implements Rule<Node\Expr\FuncCall>
+ */
+final class DumpPhpDocTypeRule implements Rule
+{
+
+	public function __construct(private ReflectionProvider $reflectionProvider)
+	{
+	}
+
+	public function getNodeType(): string
+	{
+		return Node\Expr\FuncCall::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (!$node->name instanceof Node\Name) {
+			return [];
+		}
+
+		$functionName = $this->reflectionProvider->resolveFunctionName($node->name, $scope);
+		if ($functionName === null) {
+			return [];
+		}
+
+		if (strtolower($functionName) !== 'phpstan\dumpphpdoctype') {
+			return [];
+		}
+
+		if (count($node->getArgs()) === 0) {
+			return [];
+		}
+
+		return [
+			RuleErrorBuilder::message(
+				sprintf(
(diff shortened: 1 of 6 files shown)


src/Rules/Debug/DumpPhpDocTypeRule.php
```diff
@@ -0,0 +1,58 @@
...
+
+		if (count($node->getArgs()) === 0) {
+			return [];
+		}
+
+		return [
+			RuleErrorBuilder::message(
+				sprintf(
+					'Dumped type: %s',
+					$scope->getType($node->getArgs()[0]->value)->toPhpDocNode(),
```
ondrejmirtes: Use the Printer here (https://phpstan.github.io/phpdoc-parser/2.0.x/PHPStan.PhpDocParser.Printer.Printer.html#_print), not `Node::__toString()`. It's more precise, uses less parentheses, only where needed.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Debug/DumpPhpDocTypeRuleTest.php
```diff
@@ -0,0 +1,89 @@
...
+
+	public function testRuleSymbols(): void
+	{
+		$this->analyse([__DIR__ . '/data/dump-phpdoc-type.php'], [
+			[
+				"Dumped type: array{'': ''}",
+				5,
+			],
+			[
+				"Dumped type: array{'\0': 'NUL', NUL: '\0'}",
```
ondrejmirtes, line 28: Can you also try some of the special multi-byte characters you did in the previous PR? (`\x??`)

ondrejmirtes (changes requested):

src/dumpType.php
```diff
@@ -13,3 +13,15 @@ function dumpType($value) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
...
+/**
+ * @phpstan-pure
+ * @param mixed $value
+ * @return mixed
+ *
+ * @throws void
+ */
+function dumpPhpDocType($value) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
+{
+	return null;
```
ruudk, line 26: I know this is copied from `dumpType`, but I always wondered: Why does this even return something? 

PHPStorm always complains that the result is not used.

Can't we make this return `void`? Then do the same for the other helpers?

/cc @ondrejmirtes

ondrejmirtes: We can’t make void functions pure, that’s why.

Personally I turn this off in PhpStorm, as PHPStan will tell me the same thing.

zonuexe: I think this is because if declared as void, `dumpType()` would affect the purity of the embedded function.

ondrejmirtes:
Please fix CS https://github.com/phpstan/phpstan-src/actions/runs/11252140969/job/31284635295#step:7:1

ondrejmirtes:
Thank you!

staabm:
@zonuexe please see PHP7.4 only test errors

grafik


https://github.com/phpstan/phpstan-src/actions/runs/11254262628/job/31291308442?pr=3561
https://github.com/phpstan/phpstan-src/actions/runs/11254262628/job/31291309269?pr=3561

interesstingly these appear in [a followup PR](https://github.com/phpstan/phpstan-src/pull/3561), and were not detected in this PR here.

maybe its a PHPStan 2.x only thing

PR #3562, opened 2024-10-11
Add null to array_map(null, $a, $b)

schlndh:
Fixes: https://3v4l.org/QKHrt (https://phpstan.org/r/357b09e6-80b7-4c50-97db-35995ff4cb67)

Unfortunately, since it depends on values of multiple variables at the same time (which may be correlated), there will be false positives like this:

```php
if (rand()) {
	$a = [1];
	$b = [2];
} else {
	$a = [1, 2];
	$b = [3, 4];
}

array_map(null, $a, $b);
```

But I guess it's better to be safe than sorry. I at least added support for `array_map(null, $a, $a)`.

### DIFF
--- a/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
@@ -19,6 +19,7 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeUtils;
 use function array_map;
+use function array_reduce;
 use function array_slice;
 use function count;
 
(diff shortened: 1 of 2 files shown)


src/Type/Php/ArrayMapFunctionReturnTypeExtension.php
```diff
@@ -32,7 +33,8 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
 
 	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
 	{
-		if (count($functionCall->getArgs()) < 2) {
+		$numArgs = count($functionCall->getArgs());
```
ddevsr, line 36: By the way, i see multiple call function `$functionCall->getArgs()`. Can refactor?

ondrejmirtes:
Thank you.

PR #3567, opened 2024-10-12
Improve lowercase string verbosity level

VincentLanglet:


tests/PHPStan/Rules/Methods/ReturnTypeRuleTest.php
```diff
@@ -1054,4 +1054,19 @@ public function testBug10715(): void
 		$this->analyse([__DIR__ . '/data/bug-10715.php'], []);
 	}
 
+	public function testBug11835(): void
+	{
```
ondrejmirtes: VerbosityLevel unit test would make more sense than this because this error will inevitably go away.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Methods/ReturnTypeRuleTest.php
VincentLanglet: I updated the PR with another test. I hope it's what you wanted :)

ondrejmirtes:
Thank you!

PR #3573, opened 2024-10-14
Support `for` endless loops

herndlm:
Closes https://github.com/phpstan/phpstan/issues/6807
Closes https://github.com/phpstan/phpstan/issues/8463
Closes https://github.com/phpstan/phpstan/issues/9374

analogue to `while`.  contains also a slight simplification in the setting of `$isAlwaysTerminating` for `while` loops to keep this more in sync.

I had to switch the 2 test blocks in `tests/PHPStan/Rules/Comparison/data/strict-comparison.php` because the first one would be never exiting endless loop after this change, which would make the second one dead code and not testing anything useful any more.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1194,8 +1194,10 @@ final class NodeScopeResolver
 
 			if ($alwaysIterates) {
 				$isAlwaysTerminating = count($finalScopeResult->getExitPointsByType(Break_::class)) === 0;
+			} elseif ($isIterableAtLeastOnce) {
+				$isAlwaysTerminating = $finalScopeResult->isAlwaysTerminating();
 			} else {
-				$isAlwaysTerminating = $isIterableAtLeastOnce && $finalScopeResult->isAlwaysTerminating();
+				$isAlwaysTerminating = false;
 			}
 			$condScope = $condResult->getFalseyScope();
 			if (!$isIterableAtLeastOnce) {
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1414,8 +1414,10 @@ final class NodeScopeResolver
 
 			if ($alwaysIterates) {
 				$isAlwaysTerminating = count($finalScopeResult->getExitPointsByType(Break_::class)) === 0;
+			} elseif ($isIterableAtLeastOnce->yes()) {
+				$isAlwaysTerminating = $finalScopeResult->isAlwaysTerminating();
 			} else {
-				$isAlwaysTerminating = false; // $finalScopeResult->isAlwaysTerminating() && $isAlwaysIterable
+				$isAlwaysTerminating = false;
 			}
 
 			return new StatementResult(
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1314,7 +1314,6 @@ final class NodeScopeResolver
 			}
 
 			$bodyScope = $initScope;
-			$alwaysIterates = $stmt->cond === [] && $context->isTopLevel();
 			$isIterableAtLeastOnce = TrinaryLogic::createYes();
 			$lastCondExpr = $stmt->cond[count($stmt->cond) - 1] ?? null;
 			foreach ($stmt->cond as $condExpr) {
@@ -1385,7 +1384,10 @@ final class NodeScopeResolver
 				$loopScope = $this->processExprNode($stmt, $loopExpr, $loopScope, $nodeCallback, ExpressionContext::createTopLevel())->getScope();
 			}
 			$finalScope = $finalScope->generalizeWith($loopScope);
+
+			$alwaysIterates = TrinaryLogic::createFromBoolean($context->isTopLevel());
 			if ($lastCondExpr !== null) {
+				$alwaysIterates = $alwaysIterates->and($finalScope->getType($lastCondExpr)->toBoolean()->isTrue());
 				$finalScope = $finalScope->filterByFalseyValue($lastCondExpr);
 			}
 
@@ -1412,7 +1414,7 @@ final class NodeScopeResolver
 				}
 			}
 
-			if ($alwaysIterates) {
+			if ($alwaysIterates->yes()) {
 				$isAlwaysTerminating = count($finalScopeResult->getExitPointsByType(Break_::class)) === 0;
 			} elseif ($isIterableAtLeastOnce->yes()) {
 				$isAlwaysTerminating = $finalScopeResult->isAlwaysTerminating();
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1194,10 +1194,8 @@ final class NodeScopeResolver
 
 			if ($alwaysIterates) {
 				$isAlwaysTerminating = count($finalScopeResult->getExitPointsByType(Break_::class)) === 0;
-			} elseif ($isIterableAtLeastOnce) {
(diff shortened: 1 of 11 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -1410,10 +1409,16 @@ private function processStmtNode(
 				}
 			}
 
+			if ($alwaysIterates) {
+				$isAlwaysTerminating = count($finalScopeResult->getExitPointsByType(Break_::class)) === 0;
+			} else {
+				$isAlwaysTerminating = false; // $finalScopeResult->isAlwaysTerminating() && $isAlwaysIterable
```
herndlm: not a 100% sure what to do with this comment. it looks outdated too and should most likely be `$isIterableAtLeastOnce->yes() && $finalScopeResult->isAlwaysTerminating()`. just worried that deleting it might lead to loosing _some_  old information

herndlm: actually doing this instead of setting it just to `false` leads to one failing tests:
```
1) PHPStan\Analyser\StatementResultTest::testIsAlwaysTerminating with data set #56 ('for ($i = 0; $i < 10; $i++) {...urn; }', false)
Failed asserting that true is identical to false.
```

https://github.com/phpstan/phpstan-src/blob/eb6a95a9230d9c40e738c92edfef4f89e56678d2/tests/PHPStan/Analyser/StatementResultTest.php#L232-L235

but messing with this is out of scope of this MR anyway I'd say.

ondrejmirtes (changes requested):
Please do not change anything about `while`. Change only `for`. After that is merged, you can send the `while` change as another PR and then we can think about it.

herndlm:
> Please do not change anything about `while`. Change only `for`. After that is merged, you can send the `while` change as another PR and then we can think about it.

It shouldn't change anything, just merges the elseif/else , so only refactor. But I might be overlooking something and can revert it again 😊

herndlm:
adapted. the failing test cases seem to be OK, since in https://github.com/nikic/PHP-Parser/blob/v3.1.5/lib/PhpParser/ParserAbstract.php#L175-L344 there are 2 endless loops and the outer one does not seem to have a break statement. Replacing them with a `while (true)` makes it fail the same way.

src/Analyser/NodeScopeResolver.php
ondrejmirtes: I'd say this should be true so updating the test should be correct.

src/Analyser/NodeScopeResolver.php
```diff
@@ -1410,10 +1411,16 @@ private function processStmtNode(
...
+			}
+
 			return new StatementResult(
 				$finalScope,
 				$finalScopeResult->hasYield() || $hasYield,
-				false/* $finalScopeResult->isAlwaysTerminating() && $isAlwaysIterable*/,
+				$isAlwaysTerminating,
```
ondrejmirtes, line 1428: I feel like I'm missing `$finalScopeResult->isAlwaysTerminating()` here. It's related to the failing test we talk above.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Also please update the E2E test so it's not failing.

herndlm: adapted I think. not sure if I did it the right way, but I manually checked-out PHP-Parser and manually run PHPStan like in the test and let it update the baseline there.

herndlm:
Saw this fixes some more issues, will add regression tests..
UPDATE: added, those were basically duplicates

tests/PHPStan/Analyser/StatementResultTest.php
```diff
@@ -173,6 +173,18 @@ public function dataIsAlwaysTerminating(): array
...
+			[
+				'for (;;) { }',
+				true,
+			],
+			[
+				'for (;;) { return; }',
+				true,
+			],
+			[
+				'for (;;) { break; }',
```
staabm, line 229: this seems to be the only test you are actually using `break;`. 
would a more functional test like the ones in the `bug*` files make sense with a `break`, or maybe even a `break 2` and nested loops?

staabm: I think another useful test could be
`'for (;;) { exit; }'`

herndlm: done, those tests with nested loops and break 2 show that breaking the endless loops is not correctly working for both `while` and `for` at the moment, as long as I didn't mess them up at least. Maybe this can be fixed / improved in a follow-up? 🤔

herndlm: but I think I know how to fix them. I could open this as follow-up for both `while` and `for` right after I believe.

staabm: I am totally fine with doing follow-ups. I just tried to think about cases maybe relevant but maybe not yet covered.

thanks for looking into it

edit: another thing (maybe also followup) would be to cover `match` in `StatementResultTest`

herndlm: oh man, I was looking too long at this and none of the tests made sense any more. I think they actually are correct and everything is looking good. I need a break.

herndlm:
realized that I can make this better..

To do so I'd need https://github.com/phpstan/phpstan-src/pull/3578 first though :)

herndlm:
Thx to the pre-condition PR that was merged, I could make the endless loop detection more generic and better 🎉

Had to adapt a bit of test code in `DefinedVariableRuleTest` since it correctly figured out now that after those endless loops variables might not be defined.

I like the following one in particular where the conditions were switched and by mistake this resulted in an endless loop basically:
```diff
-for ($forI = 0; $forI < 10, $anotherVariableFromForCond = 1; $forI++, $forJ = $forI) {
+for ($forI = 0; $anotherVariableFromForCond = 1, $forI < 10; $forI++, $forJ = $forI) {
```

src/Analyser/NodeScopeResolver.php
```diff
@@ -1384,7 +1384,10 @@ private function processStmtNode(
 				$loopScope = $this->processExprNode($stmt, $loopExpr, $loopScope, $nodeCallback, ExpressionContext::createTopLevel())->getScope();
 			}
 			$finalScope = $finalScope->generalizeWith($loopScope);
+
+			$alwaysIterates = TrinaryLogic::createFromBoolean($context->isTopLevel());
```
ondrejmirtes, line 1388: Would something fail if you said `$alwaysIterates = TrinaryLogic::createYes()` here?

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
herndlm: no, apparently not. this comes basically from the other loops which have code like the following
```php
$alwaysIterates = $condBooleanType->isTrue()->yes() && $context->isTopLevel();
```

but tbh, I didn't fully understand what the level here changes or does

ondrejmirtes: OK, makes sense to be consistent.

This condition asks about whether it's the "analysis iteration", or one of the "let's find out if types stabilize in the loop before we generalize them" that's done for all loops in a do-while like this: https://github.com/phpstan/phpstan-src/blob/c3bb0557b3b34770e9a68cf7f5f5cc3c1f6291fd/src/Analyser/NodeScopeResolver.php#L1330-L1360

ondrejmirtes:
Thank you.

PR #3578, opened 2024-10-18
Only use last for condition to filter scope

herndlm:
Improves https://github.com/phpstan/phpstan/issues/11872 but does not exactly what I hope would be possible. But with the way things currently work (going through conditions and filtering the scope) this might just not be achievable or worth it. So less precise type is better than more precise type that is wrong for me tbh :)

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1314,13 +1314,18 @@ final class NodeScopeResolver
 
 			$bodyScope = $initScope;
 			$isIterableAtLeastOnce = TrinaryLogic::createYes();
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Can you explain the PHP behaviour here like I'm five? :) I don't get what's going on.

herndlm:
I think the main thing is that (see https://www.php.net/manual/en/control-structures.for.php) all `cond` (expr2) expressions are executed, but only the last one is used to determine if the loop should continue. Based on that it doesn't make sense (to me at least ;)) to filter the scope by any other than the last expression to narrow types.

UPDATE: having said that and while that makes sense to me, I'm not a 100% sure anymore if I adapted the right places here..

ondrejmirtes:
Alright, thank you!

PR #3579, opened 2024-10-20
Decorate ImpossibleCheckTypeHelper with reason

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11799

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeFunctionCallRule.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeFunctionCallRule.php
@@ -7,6 +7,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Parser\LastConditionVisitor;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use function count;
 use function sprintf;
 
 /**
@@ -36,17 +37,21 @@ final class ImpossibleCheckTypeFunctionCallRule implements Rule
 		}
 
 		$functionName = (string) $node->name;
-		$isAlways = $this->impossibleCheckTypeHelper->findSpecifiedType($scope, $node);
+		[$isAlways, $reasons] = $this->impossibleCheckTypeHelper->findSpecifiedType($scope, $node);
 		if ($isAlways === null) {
 			return [];
 		}
 
-		$addTip = function (RuleErrorBuilder $ruleErrorBuilder) use ($scope, $node): RuleErrorBuilder {
+		$addTip = function (RuleErrorBuilder $ruleErrorBuilder) use ($scope, $node, $reasons): RuleErrorBuilder {
+			if (count($reasons) > 0) {
+				$ruleErrorBuilder->acceptsReasonsTip($reasons);
+			}
+
 			if (!$this->treatPhpDocTypesAsCertain) {
 				return $ruleErrorBuilder;
 			}
 
-			$isAlways = $this->impossibleCheckTypeHelper->doNotTreatPhpDocTypesAsCertain()->findSpecifiedType($scope, $node);
+			$isAlways = $this->impossibleCheckTypeHelper->doNotTreatPhpDocTypesAsCertain()->findSpecifiedType($scope, $node)[0];
 			if ($isAlways !== null) {
 				return $ruleErrorBuilder;
 			}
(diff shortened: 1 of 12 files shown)


src/Rules/Comparison/ImpossibleCheckTypeFunctionCallRule.php
```diff
@@ -41,17 +42,21 @@ public function processNode(Node $node, Scope $scope): array
...
-		$isAlways = $this->impossibleCheckTypeHelper->findSpecifiedType($scope, $node);
+		[$isAlways, $reasons] = $this->impossibleCheckTypeHelper->findSpecifiedType($scope, $node);
 		if ($isAlways === null) {
 			return [];
 		}
 
-		$addTip = function (RuleErrorBuilder $ruleErrorBuilder) use ($scope, $node): RuleErrorBuilder {
+		$addTip = function (RuleErrorBuilder $ruleErrorBuilder) use ($scope, $node, $reasons): RuleErrorBuilder {
+			if (count($reasons) > 0) {
+				return $ruleErrorBuilder->acceptsReasonsTip($reasons);
```
VincentLanglet: I copied https://github.com/phpstan/phpstan-src/compare/6ac62d3b00f64e72c02b7aa477e425e90f46fa65...dc5d8f4d3eef18b1d80b8ffb3a1adfe8de6d7268#diff-6f5d047ab5b60f37c6dca68882f80e1f7ba2d946e0ffb91290d39589cf55acc2R58-R60

but looking at the failing test, I feel like we want to keep the `treatPhpDocTypesAsCertainTip` @ondrejmirtes , no ?
(So maybe the StrictComparisonOfDifferentTypesRule too ?)

Currently I dunno if I need to
- Update the tests.
- Or update this line of code.

Your call.

tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
```diff
@@ -1102,4 +1102,19 @@ public function testAlwaysTruePregMatch(): void
...
+	{
+		$this->checkAlwaysTrueCheckTypeFunctionCall = true;
+		$this->treatPhpDocTypesAsCertain = true;
+		$this->analyse([__DIR__ . '/data/bug-11799.php'], [
+			[
+				"Call to function in_array() with arguments string, array{'publishDate', 'approvedAt', 'allowedValues'} and true will always evaluate to false.",
+				11,
+				"• 'publishDate' is not lowercase.
+• 'approvedAt' is not lowercase.
+• 'allowedValues' is not lowercase.",
```
VincentLanglet, line 989: Since the comparison is between `'publishDate'|'approvedAt'|'allowedValues'` and `lowercase string`, and because of how `UnionType::isSuperTypeOfWithReason` works
```
$result = IsSuperTypeOfResult::createNo()->or(...array_map(static fn (Type $innerType) => $innerType->isSuperTypeOfWithReason($otherType), $this->types));
		if ($result->yes()) {
			return $result;
		}
```
we have a list with three reasons.

Is it ok for you @ondrejmirtes ?

VincentLanglet:
Not sure it's a satisfying solution

ondrejmirtes:
There are conflicts now.

ondrejmirtes:
Also - maybe base this on 2.1.x now.

VincentLanglet:
> Also - maybe base this on 2.1.x now.

It stopped working on 2.1.x when I rebased for the case 
```
in_array($sortBy, ['publishDate', 'approvedAt', 'allowedValues'], true)
```
because:
- In 1.12.x, `$specifiedTypes->getRootExpr()` returns NULL
- In 2.1.x, it now returns `PhpParser\Node\Expr\BinaryOp\Identical`. So the code used is
```
$rootExprType = ($this->treatPhpDocTypesAsCertain ? $scope->getType($rootExpr) : $scope->getNativeType($rootExpr));
if ($rootExprType instanceof ConstantBooleanType) {
     return [$rootExprType->getValue(), []];
}
```

It seems like `in_array($sortBy, ['publishDate', 'approvedAt', 'allowedValues']` is now understood as `$sortBy === 'publishDate' || $sortBy === 'approvedAt' || $sortBy === 'allowedValues'`.

The issue is that when using `$rootExprType` we don't have the reason.

I dunno if you see a fix/solution and/or if you're still interested in a way by this PR since it might have some inconsistency between when a reason is given and when it's not given @ondrejmirtes ?

PR #3581, opened 2024-10-24
Dead code detection

janedbal:
- I just used phpstan-src for testing rewritten dead code analysis (call-graph based) within our [shipmonk/dead-code-detector](https://github.com/shipmonk-rnd/dead-code-detector). 
- With just few ignores and one provider, it already reports some truely dead code.
- To make this usable in CI, few minor tweaks would be required, but I think you can check the violations even now and remove the dead stuff.


janedbal:
Superseded by #3996

PR #3582, opened 2024-10-24
Use a benevolent union type in `openssl_cipher_iv_length`

stof:
This function returns `false` (and triggers a warning) only when the argument is an unknown algorithm, which is not something worth checking when using this function with a known algorithm.


ondrejmirtes (changes requested):
Nope, this should be handled more like `mb_` functions are handled with `mb_list_encodings`. See MbFunctionsReturnTypeExtensionTrait.

stof:
I submitted another PR with the proper approach (and also applying it to `openssl_cipher_key_length`)

PR #3583, opened 2024-10-26
Allow mixed type dynamic constants

ssigwart:
Addresses https://github.com/phpstan/phpstan/issues/7520

### DIFF
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -412,8 +412,8 @@ final class ConstantResolver
 		if ($constantType->isConstantValue()->yes()) {
 			if (array_key_exists($constantName, $this->dynamicConstantNames)) {
 				$phpdocTypes = $this->dynamicConstantNames[$constantName];
-				$typeStringResolver = $this->container?->getByType(TypeStringResolver::class);
-				if ($typeStringResolver !== null) {
+				if ($this->container !== null) {
+					$typeStringResolver = $this->container->getByType(TypeStringResolver::class);
 					return $typeStringResolver->resolve($phpdocTypes, new NameScope(null, [], null));
 				}
 			}
@@ -435,8 +435,8 @@ final class ConstantResolver
 
 			if ($constantType->isConstantValue()->yes()) {
 				$phpdocTypes = $this->dynamicConstantNames[$lookupConstantName];
-				$typeStringResolver = $this->container?->getByType(TypeStringResolver::class);
-				if ($typeStringResolver !== null) {
+				if ($this->container !== null) {
+					$typeStringResolver = $this->container->getByType(TypeStringResolver::class);
 					return $typeStringResolver->resolve($phpdocTypes, new NameScope(null, [], $className));
 				}
 			}
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -428,10 +428,6 @@ final class ConstantResolver
 	{
 		$lookupConstantName = sprintf('%s::%s', $className, $constantName);
 		if (array_key_exists($lookupConstantName, $this->dynamicConstantNames)) {
-			if ($nativeType !== null) {
-				return $nativeType;
-			}
-
 			if ($constantType->isConstantValue()->yes()) {
 				$phpdocTypes = $this->dynamicConstantNames[$lookupConstantName];
 				if ($this->container !== null) {
@@ -439,6 +435,12 @@ final class ConstantResolver
 					return $typeStringResolver->resolve($phpdocTypes, new NameScope(null, [], $className));
 				}
 			}
+
+			if ($nativeType !== null) {
+				return $nativeType;
+			}
+			
+			return $constantType;
 		}
 
 		if (in_array($lookupConstantName, $this->dynamicConstantNames, true)) {
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -52,8 +52,6 @@ final class ConstantResolver
 		private array $dynamicConstantNames,
 		private int|array|null $phpVersion,
 		private ComposerPhpVersionFactory $composerPhpVersionFactory,
-		private ?PhpVersion $composerMinPhpVersion,
-		private ?PhpVersion $composerMaxPhpVersion,
 		private ?Container $container,
 	)
 	{
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -414,6 +414,7 @@ final class ConstantResolver
 					$typeStringResolver = $this->container->getByType(TypeStringResolver::class);
 					return $typeStringResolver->resolve($phpdocTypes, new NameScope(null, [], null));
 				}
+				return $constantType;
 			}
 			if (in_array($constantName, $this->dynamicConstantNames, true)) {
 				return $constantType->generalize(GeneralizePrecision::lessSpecific());
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -4,8 +4,10 @@ namespace PHPStan\Analyser;
 
 use PhpParser\Node\Name;
 use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\DependencyInjection\Container;
 use PHPStan\Php\ComposerPhpVersionFactory;
 use PHPStan\Php\PhpVersion;
+use PHPStan\PhpDoc\TypeStringResolver;
 use PHPStan\Reflection\NamespaceAnswerer;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Reflection\ReflectionProvider\ReflectionProviderProvider;
@@ -50,6 +52,9 @@ final class ConstantResolver
 		private array $dynamicConstantNames,
 		private int|array|null $phpVersion,
 		private ComposerPhpVersionFactory $composerPhpVersionFactory,
+		private ?PhpVersion $composerMinPhpVersion,
+		private ?PhpVersion $composerMaxPhpVersion,
+		private ?Container $container,
 	)
 	{
 	}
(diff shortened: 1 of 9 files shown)


tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -8535,6 +8535,102 @@ public function testDynamicConstantsWithNativeTypes(
...
+			[
+				"'abc123def'",
+				'MixedTypeConstants\MixedTypeConstantClass::PURE_CONSTANT_IN_CLASS',
+			],
+			[
+				"'xyz'",
+				'MixedTypeConstants\NoMixedTypeConstantClass::MIXED_TYPE_CONSTANT_IN_CLASS',
+			],
+			[
+				'false',
```
ssigwart: This doesn't seem right to me, but this test is copied and modified from `testDynamicConstants` where it also has `false`. That changed from `bool` to `false` in https://github.com/phpstan/phpstan-src/commit/285b49f2072e18ffe9ff72f221151512b8a92fe6#diff-52d2cce81a3daff27b7d6d8ba619b045420ab9a9acffc57c1c379af4eba0b127R8384.

I tried tracking down how I might be able to fix it. I found that the constant gets defined here:
https://github.com/phpstan/phpstan-src/blob/d73acef87f9d3ebd04bb6671d6b83755a8e62e8e/src/Type/Php/DefineConstantTypeSpecifyingExtension.php#L45-L60

The issue is that `$scope->getType($node->getArgs()[1]->value)` is just going based off the value, so it never reaches where I think it would determine that the constant is dynamic:
https://github.com/phpstan/phpstan-src/blob/d73acef87f9d3ebd04bb6671d6b83755a8e62e8e/src/Analyser/MutatingScope.php#L1943

ondrejmirtes:
It'd make sense to me to allow this syntax to force a specific constant type:

```neon
parameters:
	dynamicConstantNames:
		- DATABASE_ENGINE
		- Foo::BAR_CONSTANT
		ES_ENGINE: string|null
		Foo::BAZ_CONSTANT: mixed
```

ssigwart:
> It'd make sense to me to allow this syntax to force a specific constant type:
> 
> ```neon
> parameters:
> 	dynamicConstantNames:
> 		- DATABASE_ENGINE
> 		- Foo::BAR_CONSTANT
> 		ES_ENGINE: string|null
> 		Foo::BAZ_CONSTANT: mixed
> ```

That makes sense. I can get it as far as detecting that it's in `dynamicConstantNames` and getting the string value (e.g. `string|null` for `ES_ENGINE`), but haven't figured out how `ConstantResolver` can take that string and convert it to a `PHPStan\Type\Type`. Any suggestions?

staabm:
> That makes sense. I can get it as far as detecting that it's in `dynamicConstantNames` and getting the string value (e.g. `string|null` for `ES_ENGINE`), but haven't figured out how `ConstantResolver` can take that string and convert it to a `PHPStan\Type\Type`. Any suggestions?

I think you can take inspiration from 

https://github.com/phpstan/phpstan-src/blob/830b6d5a4e68512784da127305394bbd13604e75/src/Reflection/SignatureMap/SignatureMapParser.php#L49-L56

ssigwart:
Thanks, @staabm! That got me on the right track. I got this working. I even fixed https://github.com/phpstan/phpstan-src/pull/3583#discussion_r1817665299. I'm not 100% sure if the new interfaces I added are desired, but it gets the job done. Please let me know if you want me to make any changes.

I could use some help with the 1 failing test though:
https://github.com/phpstan/phpstan-src/actions/runs/11542351640/job/32125482914?pr=3583

src/Analyser/ConstantResolver.php
```diff
@@ -302,8 +302,14 @@ public function resolvePredefinedConstant(string $resolvedConstantName): ?Type
 
 	public function resolveConstantType(string $constantName, Type $constantType): Type
 	{
-		if ($constantType->isConstantValue()->yes() && in_array($constantName, $this->dynamicConstantNames, true)) {
-			return $constantType->generalize(GeneralizePrecision::lessSpecific());
+		if ($constantType->isConstantValue()->yes()) {
+			if (array_key_exists($constantName, $this->dynamicConstantNames)) {
+				$phpdocTypes = $this->dynamicConstantNames[$constantName];
+				return $this->reflectionProviderProvider->getReflectionProvider()->getSignatureMapProvider()->getTypeFromString($phpdocTypes, null);
```
ondrejmirtes: You should use TypeStringResolver for this, not SignatureMapProvider.

src/Analyser/ConstantResolver.php
```diff
@@ -312,6 +318,17 @@ public function resolveConstantType(string $constantName, Type $constantType): T
...
 	{
 		$lookupConstantName = sprintf('%s::%s', $className, $constantName);
+		if (array_key_exists($lookupConstantName, $this->dynamicConstantNames)) {
+			if ($nativeType !== null) {
+				return $nativeType;
+			}
+
+			if ($constantType->isConstantValue()->yes()) {
+				$phpdocTypes = $this->dynamicConstantNames[$lookupConstantName];
+				return $this->getReflectionProvider()->getSignatureMapProvider()->getTypeFromString($phpdocTypes, $className);
```
ondrejmirtes: dtto

src/Reflection/ReflectionProvider.php
```diff
@@ -36,4 +37,6 @@ public function getConstant(Node\Name $nameNode, ?NamespaceAnswerer $namespaceAn
 
 	public function resolveConstantName(Node\Name $nameNode, ?NamespaceAnswerer $namespaceAnswerer): ?string;
 
+	public function getSignatureMapProvider(): SignatureMapProvider;
```
ondrejmirtes: Don't add these methods please.

ondrejmirtes (changes requested):

src/Analyser/ConstantResolver.php
ssigwart: How would I get an instance of `TypeStringResolver` from here?

ssigwart:
Thanks, @ondrejmirtes. I made updates. I'm not sure if passing around the container is the right thing to do, but I got it working.

src/Analyser/ConstantResolver.php
```diff
@@ -41,6 +43,7 @@ public function __construct(
 		private array $dynamicConstantNames,
 		private ?PhpVersion $composerMinPhpVersion,
 		private ?PhpVersion $composerMaxPhpVersion,
+		private ?Container $container,
```
staabm, line 55: make this new param the TypeStringResolver itself. 

It will be injected by the DIC automatically

ssigwart: I initially thought that might be the way to do it. However, when I tried it, I wound up getting an error about circular dependencies. So that's why I switched to the entire container. Do you have any ideas on how to get around the circular dependency error? It's also possible I just did it wrong. I assumed I needed to update the places that called the constructor, so I was using the container there to get the TypeStringResolver.

staabm: ah I can see it now.

`TypeStringResolver` has a `TypeNodeResolver` contruction dependency which in turn has a `ConstantResolver` dependency.

so when you add a `TypeStringResolver` dependency on `ConstantResolver`  you build a cycle.

I will think about it and report back

staabm: I guess one way we could break the cycle is to remove the `ConstantResolver` construction-time dependency of `TypeNodeResolver` and turn it into a setter injection (similar how we do it in `TypeSpecifierFactory` for the `TypeSpecifier`).

in my local testing I did not yet succeed in doing so :)

staabm: or maybe better instead remove `TypeNodeResolver` from `TypeStringResolver` construction-time deps and move it to setter-injection

ssigwart: Sorry, @staabm. I don't have experience with DI in PHP, so I'm not entirely sure how to do this. I assume you meant to add a `setTypeNodeResolver` function to `TypeStringResolver`. I'm not sure where I would call `setTypeNodeResolver` though. I thought maybe I needed to add a `TypeStringResolverFactory`. I'm guessing that maybe I need to update `conf/config.neon`, so I changed it to this:
```neon
-
	class: PHPStan\PhpDoc\TypeStringResolver
typeStringResolverFactory:
	class: PHPStan\PhpDoc\TypeStringResolverFactory
```

In a new `TypeStringResolverFactory` class, I added this:
```php
public function create(): TypeStringResolver
{
	$typeStringResolver = new TypeStringResolver(
		$this->typeLexer,
		$this->typeParser
	);
	$typeStringResolver->setTypeNodeResolver($this->container->getByType(TypeNodeResolver::class));

	return $typeStringResolver;
}
```

I still wound up with this when running the test though:
```
In Container.php line 327:

  Circular reference detected for: reflectionProvider, betterReflectionProvider, 093,
  058, 042, 040.
```

Am I on the right track? Do you have any suggestions? Thanks.

ssigwart:
@staabm, should I rebase this to resolve the merge conflicts?

staabm:
Yes, please resolve the conflicts

ssigwart:
@staabm, I rebased it.

ssigwart:
I rebased this to fix the merge conflict.

src/Analyser/ConstantResolver.php
```diff
@@ -404,8 +407,17 @@ private function getMaxPhpVersion(): ?PhpVersion
...
 	{
-		if ($constantType->isConstantValue()->yes() && in_array($constantName, $this->dynamicConstantNames, true)) {
-			return $constantType->generalize(GeneralizePrecision::lessSpecific());
+		if ($constantType->isConstantValue()->yes()) {
+			if (array_key_exists($constantName, $this->dynamicConstantNames)) {
+				$phpdocTypes = $this->dynamicConstantNames[$constantName];
+				if ($this->container !== null) {
+					$typeStringResolver = $this->container->getByType(TypeStringResolver::class);
+					return $typeStringResolver->resolve($phpdocTypes, new NameScope(null, [], null));
+				}
```
ondrejmirtes, line 416:
```suggestion
				}
				return $constantType;
```

src/Analyser/ConstantResolver.php
```diff
@@ -414,6 +426,20 @@ public function resolveConstantType(string $constantName, Type $constantType): T
...
+				return $nativeType;
+			}
+
+			if ($constantType->isConstantValue()->yes()) {
+				$phpdocTypes = $this->dynamicConstantNames[$lookupConstantName];
+				if ($this->container !== null) {
+					$typeStringResolver = $this->container->getByType(TypeStringResolver::class);
+					return $typeStringResolver->resolve($phpdocTypes, new NameScope(null, [], $className));
+				}
+			}
```
ondrejmirtes:
```suggestion
			if ($constantType->isConstantValue()->yes()) {
				$phpdocTypes = $this->dynamicConstantNames[$lookupConstantName];
				if ($this->container !== null) {
					$typeStringResolver = $this->container->getByType(TypeStringResolver::class);
					return $typeStringResolver->resolve($phpdocTypes, new NameScope(null, [], $className));
				}
			}

			if ($nativeType !== null) {
				return $nativeType;
			}
			
			return $constantType;
```

ondrejmirtes:
Thank you.

ssigwart:
Thanks for cleaning it up and getting it merged.

PR #3584, opened 2024-10-26
Utilize composer min-php version to narrow PHP_VERSION_ID

staabm:
idea is to catch errors like https://github.com/Roave/BetterReflection/pull/1406#discussion_r1523065225 in which bogus conditions like `if (PHP_VERSION_ID >= 80100) {}` have been used in tests. 

it's a bug in the linked case, because the project itself only supports `"php": "~8.2.0 || ~8.3.2",`.
Since CI only is running against in composer.json declared supported versions this error was not spotted before.

----

rebased version of https://github.com/phpstan/phpstan-src/pull/2968 for 2.x


staabm:
PHPStan 2.x seems the perfect point in time to re-vive this PR, as it found some dead tests in phpstan-extension repos :) (because of the raised min-php version)

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,125 @@
...
+	private function getComposerRequireVersion(): ?string
+	{
+		$composerPhpVersion = null;
+		if (count($this->composerAutoloaderProjectPaths) > 0) {
+			$composerJsonPath = end($this->composerAutoloaderProjectPaths) . '/composer.json';
+			if (is_file($composerJsonPath)) {
+				try {
+					$composerJsonContents = FileReader::read($composerJsonPath);
+					$composer = Json::decode($composerJsonContents, Json::FORCE_ARRAY);
+					$requiredVersion = $composer['require']['php'] ?? null;
```
mvorisek, line 123: `php-64bit` should be supported as well - the minimal version should be used of these two if both are present

conf/bleedingEdge.neon
```diff
@@ -3,3 +3,5 @@ parameters:
 		bleedingEdge: true
 		skipCheckGenericClasses!: []
 		stricterFunctionMap: true
+		narrowPhpVersionFromComposerJson: true
```
ondrejmirtes: You have based this PR on 2.0.x. You shouldn't be adding a new flag to bleedingEdge.

ondrejmirtes (changes requested):

conf/bleedingEdge.neon
staabm: my thinking was, that this feature was never part of any stable 1.x release and therefore could need some time to stabilize within bleeding edge.

dropped the toggle now

e2e/composer-min-open-end-version/test.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+\PHPStan\Testing\assertType('int<80100, max>', PHP_VERSION_ID);
```
ondrejmirtes: This could be until 80499 which is the max version allowed in the config. It'd help us prevent more typos and be consistent.

e2e/composer-min-version/test.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+\PHPStan\Testing\assertType('int<80100, 90000>', PHP_VERSION_ID);
```
ondrejmirtes: 80499 as max

src/Php/PhpVersion.php
```diff
@@ -41,11 +41,26 @@ public function getVersionId(): int
 		return $this->versionId;
 	}
 
+	public function getMajor(): int
```
ondrejmirtes: should say "getMajorVersionId"

src/Php/PhpVersion.php
```diff
@@ -41,11 +41,26 @@ public function getVersionId(): int
...
+	{
+		return (int) floor($this->versionId / 10000);
+	}
+
+	public function getMinor(): int
+	{
+		return (int) floor(($this->versionId % 10000) / 100);
```
ondrejmirtes, line 51: dtto

src/Php/PhpVersion.php
```diff
@@ -41,11 +41,26 @@ public function getVersionId(): int
...
+	}
+
+	public function getPatch(): int
+	{
+		return (int) floor($this->versionId % 100);
```
ondrejmirtes, line 56: dtto

src/Php/ComposerPhpVersionFactory.php
```diff
@@ -0,0 +1,120 @@
...
+
+			if ($minVersion !== null) {
+				$this->minVersion = new PhpVersion($minVersion->getVersionId());
+			}
+		}
+		if ($constraint->getUpperBound()->isPositiveInfinity()) {
+			return;
+		}
+
+		$this->maxVersion = $this->buildVersion($constraint->getUpperBound()->getVersion());
```
ondrejmirtes: Don't do this work in the construtor, but lazily in the getters. (Of course it should still be cached in the properties.)

ondrejmirtes (changes requested):
Now I can finally review it :) Hopefully it makes 2.0, otherwise we'll have to add the bleeding edge toggles again 😂

staabm:
not sure what we wanne do with this test:
https://github.com/phpstan/phpstan-src/blob/2.0.x/tests/PHPStan/Analyser/data/bug-5639.php

e2e/composer-min-open-end-version/test.php
staabm: it now takes the max into account.

I have also added the min-version

ondrejmirtes: Do not change anything about the min version. It might seem inconsistent to you, but PHPStan can support analysing PHP 5 codebases without major issues and this would be too annoying.

Please also add E2E tests what would PHPStan think about the `PHP_` constants with:

* something like `^5.6` in composer.json. The result should be as precise as possible  -meaning 50600-50699 (there was no 5.7)
* something like `^7.0` in composer.json. The result should be as precise as possible  -meaning 70000-70499 (there was no 7.5)

ondrejmirtes: I want to limit the `min` key in phpstan.neon still to 70100 (because PHPStan doesn't offer any specific logic to lower versions), but I don't want to flag asking for PHP 5 as an error when there's nothing configured about a PHP version in phpstan.neon or composer.json. (Or it's configured in a way that allows for PHP 5, like `<=8.3` or `^5.6`)

ondrejmirtes: Hope it's understandable.

ondrejmirtes: Coming to thing of it, there's a 3rd E2E test proposal:

* No version in composer.json, no version in phpstan.neon - assert the `PHP_` constant values.

staabm: I think I adressed everything

ondrejmirtes:
Thank you. I'd like a follow up PR - update PHPStanDiagnoseExtension and add information:

* When the `phpVersion` in phpstan.neon is a min-max range
* When there's `php` in `require` in `composer.json` - add it as a new row

PR #3585, opened 2024-10-26
Fix too early inference of Encapsed string

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11852

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -1129,9 +1129,6 @@ final class MutatingScope implements Scope
 				}
 
 				$resultType = $this->initializerExprTypeResolver->resolveConcatType($resultType, $partType);
-				if (count($resultType->getConstantStrings()) === 0) {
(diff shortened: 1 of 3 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -1129,9 +1129,6 @@ private function resolveType(string $exprString, Expr $node): Type
 				}
 
 				$resultType = $this->initializerExprTypeResolver->resolveConcatType($resultType, $partType);
-				if (count($resultType->getConstantStrings()) === 0) {
```
VincentLanglet, line 1132: We will still need to check every part to see if it's a lowercase string.

I'm unsure I should change this to
```
if (count($resultType->getConstantStrings()) === 0 && $resultType->isLowercaseString()->no()) {
```
to avoid a similar bug to come back if we introduce another string Accessory.

ondrejmirtes:
Thank you.

PR #3588, opened 2024-10-28
Imagick::writeImage(s)File supporting `format` parameter

kingjia90:
Encountered here https://github.com/pimcore/pimcore/actions/runs/11551407073/job/32148347130?pr=17767 
The Codeception tests works fine

See also https://www.php.net/manual/en/imagick.writeimagefile.php

https://pecl.php.net/package-info.php?package=imagick&version=3.2.0b1#:~:text=Added%20additional%20parameter%20to%20writeImageFile%20to%20allow%20setting%20format


resources/functionMap.php
```diff
@@ -5016,7 +5016,7 @@
 'Imagick::waveImage' => ['bool', 'amplitude'=>'float', 'length'=>'float'],
 'Imagick::whiteThresholdImage' => ['bool', 'threshold'=>'mixed'],
 'Imagick::writeImage' => ['bool', 'filename='=>'string'],
-'Imagick::writeImageFile' => ['bool', 'filehandle'=>'resource'],
+'Imagick::writeImageFile' => ['bool', 'filehandle'=>'resource', 'format'=>'string'],
```
kingjia90: since it's 
image
should it be 
```suggestion
'Imagick::writeImageFile' => ['bool', 'filehandle'=>'resource', 'format'=>'?string'],
```
?

blankse: @kingjia90 Yes, the parameter is nullable, and its default value is `null`:
https://github.com/Imagick/imagick/blob/28f27044e435a2b203e32675e942eb8de620ee58/Imagick.stub.php#L222

Additionally, the `Imagick::writeImagesFile` method has a `format` parameter, which we should also add there.

resources/functionMap.php
```diff
@@ -5016,9 +5016,9 @@
 'Imagick::waveImage' => ['bool', 'amplitude'=>'float', 'length'=>'float'],
 'Imagick::whiteThresholdImage' => ['bool', 'threshold'=>'mixed'],
 'Imagick::writeImage' => ['bool', 'filename='=>'string'],
-'Imagick::writeImageFile' => ['bool', 'filehandle'=>'resource'],
+'Imagick::writeImageFile' => ['bool', 'filehandle'=>'resource', 'format'=>'?string'],
 'Imagick::writeImages' => ['bool', 'filename'=>'string', 'adjoin'=>'bool'],
-'Imagick::writeImagesFile' => ['bool', 'filehandle'=>'resource'],
+'Imagick::writeImagesFile' => ['bool', 'filehandle'=>'resource', 'format'=>'?string'],
```
ondrejmirtes: According to documentation this parameter is optional. Needs to be `format=`.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3590, opened 2024-10-28
More precise `array_keys` return type

staabm:
closes https://github.com/phpstan/phpstan/issues/11928

### DIFF
--- a/src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
@@ -7,10 +7,14 @@ use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Php\PhpVersion;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Accessory\AccessoryArrayListType;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\NullType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
 use function count;
 use function strtolower;
 
@@ -29,7 +33,7 @@ final class ArrayKeysFunctionDynamicReturnTypeExtension implements DynamicFuncti
 
 	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
 	{
-		if (count($functionCall->getArgs()) !== 1) {
+		if (count($functionCall->getArgs()) < 1) {
 			return null;
 		}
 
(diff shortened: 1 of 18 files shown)


src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
```diff
@@ -36,7 +40,22 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			return $keysArray;
+		}
+
+		$newArrayType = $keysArray;
+		if (!$keysArray->isConstantArray()->no()) {
+			$newArrayType = new ArrayType(
+				$keysArray->getIterableKeyType()->generalize(GeneralizePrecision::lessSpecific()),
+				$keysArray->getIterableValueType()->generalize(GeneralizePrecision::lessSpecific()),
+			);
+		}
```
staabm: we could make this more sophisticated and support more filtering on constant array, but I think this is good enough for now.

until a few days ago I didn't even know `array_keys` can filter by values - it feels pretty rare

herndlm:
Just wondering if any of this could also be done inside `Type::getKeysArray()` as well? to e.g. avoid having to deal with lists outside of it again.

See also https://github.com/phpstan/phpstan-src/pull/2117 where I played around with adding filter support 😊

herndlm:
Nice, sorry, noticed too late and should have mentioned it earlier. Later on, e.g. in https://github.com/phpstan/phpstan-src/pull/3514 I used `TrinaryLogic` instead of `bool` in the new Type function. Do you think you can use that too?

staabm:
I used $strict as bool, because other methods in Type also use it. 

Do you see a use-case where Trinary would come in handy?

herndlm:
Not necessarily. If others use that too, then consistency makes sense IMO.

src/Type/Type.php
```diff
@@ -134,7 +134,7 @@ public function setExistingOffsetValueType(Type $offsetType, Type $valueType): T
 
 	public function unsetOffset(Type $offsetType): Type;
 
-	public function getKeysArray(): Type;
+	public function getKeysArray(?Type $filterValueType = null, bool $strict = false): Type;
```
ondrejmirtes: 1) This a BC break
2) I don't like optional parameters at all.

The solution is to add another method that you'll use in the places where you're passing in the parameters now.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/bug-11928.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Bug11928;
```
staabm, line 3: see https://phpstan.org/r/9a33d7fd-a8de-4eda-b4b5-0356e75da5a2

src/Type/Php/ArrayKeysFunctionDynamicReturnTypeExtension.php
herndlm: I think the diff makes sense, but don't get me wrong, just curious, currently this PR does effectively not change much of the behaviour, right? at least it does not use the new filter or strict arg, or am I missing something? It could be added in a follow-up, sure, but if that's also not the plan, than this only adds some list type to the result, or am I wrong?

src/Type/Type.php
staabm: fixed

ondrejmirtes:
Please fix the conflict.

staabm:
> Please fix the conflict.

resolved. 

build errors are unrelated and fixed separately in https://github.com/phpstan/phpstan-src/pull/4104

ondrejmirtes:
Thank you!

PR #3592, opened 2024-10-30

staabm:
closes https://github.com/phpstan/phpstan/issues/11561

### DIFF
--- a/src/Parser/ImmediatelyInvokedClosureVisitor.php
+++ b/src/Parser/ImmediatelyInvokedClosureVisitor.php
@@ -1,0 +1,46 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Parser;
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+
+final class ImmediatelyInvokedClosureVisitor extends NodeVisitorAbstract
+{
+
+	public const ATTRIBUTE_NAME = 'isImmediatelyInvokedClosure';
+
+	private bool $inFuncCall = false;
+
+	public function beforeTraverse(array $nodes): ?array
+	{
+		$this->inFuncCall = false;
+		return null;
+	}
+
+	public function enterNode(Node $node): ?Node
+	{
+		if ($node instanceof Node\Expr\FuncCall) {
+			$this->inFuncCall = true;
+		}
+
+		if (
+			$this->inFuncCall
+			&& $node instanceof Node\Expr\Closure
+		) {
+			$node->setAttribute(self::ATTRIBUTE_NAME, true);
+		}
+
+		return null;
+	}
+
+	public function leaveNode(Node $node): ?Node
+	{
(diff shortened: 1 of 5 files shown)


ruudk:
Thanks so much @staabm. I just tested this on TwigStan and it solves the problem. 🎉  

This was a major blocker, as proper analysis of templates after such immediately involved callable became impossible. 

It would be really great if this can make it in a last 1.12 release 🙏

src/Parser/ImmediatelyInvokedClosureVisitor.php
```diff
@@ -0,0 +1,46 @@
...
+	{
+		if ($node instanceof Node\Expr\FuncCall) {
+			$this->inFuncCall = true;
+		}
+
+		if (
+			$this->inFuncCall
+			&& $node instanceof Node\Expr\Closure
+		) {
+			$node->setAttribute(self::ATTRIBUTE_NAME, true);
```
ondrejmirtes, line 31: This would lead to a lot of false positives. You only want to tag this if in `$node->name`, but right now the visitor is tagging args too.

Just check `$node->name` inside `$node instanceof Node\Expr\FuncCall`.

ondrejmirtes (changes requested):

staabm:
retargeted for 1.12 in https://github.com/phpstan/phpstan-src/pull/3593

tests/PHPStan/Analyser/nsrt/bug-11561.php
```diff
@@ -0,0 +1,40 @@
...
+use DateTime;
+
+/** @param array{date: DateTime} $c */
+function main(mixed $c): void{
+	assertType('array{date: DateTime}', $c);
+	$c['id']=1;
+	assertType('array{date: DateTime, id: 1}', $c);
+
+	$x = (function() use (&$c) {
+		assertType("array{date: DateTime, id: 1, name?: 'ruud'}", $c);
```
ruudk, line 15: I believe this is wrong, see https://github.com/phpstan/phpstan/issues/11945

tests/PHPStan/Analyser/nsrt/bug-11561.php
```diff
@@ -0,0 +1,40 @@
...
+		$c['name'] = 'ruud';
+		assertType("array{date: DateTime, id: 1, name: 'ruud'}", $c);
+		return 'x';
+	})();
+
+	assertType("array{date: DateTime, id: 1, name?: 'ruud'}", $c);
```
ruudk, line 21: I believe this is wrong, see https://github.com/phpstan/phpstan/issues/11945

tests/PHPStan/Analyser/nsrt/bug-11561.php
ruudk: My reasoning is, that since this function is invoked immediately, it should be treated as the same code as normally, without the function.

PR #3593, opened 2024-10-30

staabm:
closes https://github.com/phpstan/phpstan/issues/11561

- rebased for 1.12
- addressed feedback from https://github.com/phpstan/phpstan-src/pull/3592#discussion_r1823143065


ondrejmirtes:
Thank you!

PR #3594, opened 2024-10-31
Report invalid string offsets

staabm:
closes https://github.com/phpstan/phpstan/issues/11946

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -5,3 +5,4 @@ parameters:
 		skipCheckGenericClasses!: []
 		stricterFunctionMap: true
 		reportPreciseLineForUnusedFunctionParameter: true
+		reportPossiblyNonexistentStringOffset: true
--- a/tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
+++ b/tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
@@ -803,6 +803,26 @@ class NonexistentOffsetInArrayDimFetchRuleTest extends RuleTestCase
 				'Offset -1 does not exist on string.',
 				25,
 			],
+			[
+				'Offset int<-5, 5> might not exist on string.',
+				45
+			],
+			[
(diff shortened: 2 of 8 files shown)


tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -754,4 +754,30 @@ public function testBug2634(): void
...
+			[
+				'Offset -1 does not exist on non-empty-string.',
+				23,
+			],
+			[
+				'Offset -1 does not exist on non-falsy-string.',
+				24,
+			],
+			[
+				'Offset -1 does not exist on string.',
```
staabm, line 807: @VincentLanglet I wonder whether we intentionally don't get a `lowercase-string` in the error?

VincentLanglet: Yes, lowercase string is not mentioned when it's not needed to not make message overcomplicated.

conf/bleedingEdge.neon
```diff
@@ -3,3 +3,5 @@ parameters:
 		bleedingEdge: true
 		skipCheckGenericClasses!: []
 		stricterFunctionMap: true
+
+	reportPossiblyNonexistentStringOffset: true
```
staabm: Move into feature toggles

conf/bleedingEdge.neon
```diff
@@ -3,3 +3,4 @@ parameters:
 		bleedingEdge: true
 		skipCheckGenericClasses!: []
 		stricterFunctionMap: true
+		reportPossiblyNonexistentStringOffset: true
```
staabm, line 8: this might be enabled by default instead of using bleeding edge on 2.x, depending on how we decide here:
https://github.com/phpstan/phpstan-src/pull/3584#discussion_r1829330437

staabm:
leaving a note for myself: double check how this PR and https://github.com/phpstan/phpstan/issues/12122 make sense in combination

PR #3596, opened 2024-11-01
Support returning an array or a string in `count_chars()`

u01jmg3:
- https://www.php.net/manual/en/function.count-chars.php

### DIFF
--- a/src/Type/Php/CountCharsFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/CountCharsFunctionDynamicReturnTypeExtension.php
@@ -1,0 +1,59 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Php\PhpVersion;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\IntegerRangeType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeUtils;
+use PHPStan\Type\UnionType;
+use function count;
+
+final class CountCharsFunctionDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+{
+
+	public function __construct(private PhpVersion $phpVersion)
+	{
+	}
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return $functionReflection->getName() === 'count_chars';
+	}
+
+	public function getTypeFromFunctionCall(
+		FunctionReflection $functionReflection,
+		FuncCall $functionCall,
+		Scope $scope,
+	): ?Type
+	{
+		if (count($functionCall->getArgs()) < 1) {
+			return null;
+		}
+
+		$modeType = $scope->getType($functionCall->getArgs()[1]->value);
+
+		if (IntegerRangeType::fromInterval(0, 2)->isSuperTypeOf($modeType)->yes()) {
+			$arrayType = new ArrayType(new IntegerType(), new IntegerType());
+
+			return $this->phpVersion->throwsValueErrorForInternalFunctions()
+				? $arrayType
+				: TypeUtils::toBenevolentUnion(new UnionType([$arrayType, new ConstantBooleanType(false)]));
+		}
+
+		$stringType = new StringType();
+
+		return $this->phpVersion->throwsValueErrorForInternalFunctions()
+			? $stringType
+			: TypeUtils::toBenevolentUnion(new UnionType([$stringType, new ConstantBooleanType(false)]));
+	}
+
+}
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -1262,10 +1262,15 @@ services:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
 
 	-
-	    class: PHPStan\Type\Php\ConstantHelper
+		class: PHPStan\Type\Php\ConstantHelper
 
 	-
 		class: PHPStan\Type\Php\CountFunctionReturnTypeExtension
+		tags:
+			- phpstan.broker.dynamicFunctionReturnTypeExtension
+
+	-
+		class: PHPStan\Type\Php\CountCharsFunctionDynamicReturnTypeExtension
 		tags:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
 
--- a/tests/PHPStan/Analyser/nsrt/count-chars-7.4.php
+++ b/tests/PHPStan/Analyser/nsrt/count-chars-7.4.php
@@ -1,0 +1,21 @@
+<?php // lint <= 7.4
+
+namespace CountChars;
+
+use function PHPStan\Testing\assertType;
+
+class X {
+	const ABC = 'abcdef';
+
+	function doFoo(): void {
+		assertType('array<int, int>|false', count_chars(self::ABC, 0));
+		assertType('array<int, int>|false', count_chars(self::ABC, 1));
+		assertType('array<int, int>|false', count_chars(self::ABC, 2));
+
+		assertType('string|false', count_chars(self::ABC, 3));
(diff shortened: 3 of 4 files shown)


stubs/core.stub
```diff
@@ -328,3 +328,8 @@ function ltrim(string $string, string $characters = " \n\r\t\v\x00"): string {}
  * @return ($string is lowercase-string ? lowercase-string : string)
  */
 function rtrim(string $string, string $characters = " \n\r\t\v\x00"): string {}
+
+/**
+ * @return ($mode is 3 ? string : ($mode is 4 ? string : array<int, int>))
```
mvorisek:
```suggestion
 * @return ($mode is 3|4 ? string : array<int, int>)
```

u01jmg3:
Still need to support the method returning `false` in PHP <8

{2C798760-C136-4A8C-B96B-ED392D31C926}

tests/PHPStan/Analyser/nsrt/count-chars.php
```diff
@@ -0,0 +1,17 @@
...
+
+class X {
+	const ABC = 'abcdef';
+
+	function doFoo(): void {
+		assertType('array', count_chars(self::ABC, 0));
+		assertType('array', count_chars(self::ABC, 1));
+		assertType('array', count_chars(self::ABC, 2));
+		assertType('string', count_chars(self::ABC, 3));
+		assertType('string', count_chars(self::ABC, 4));
```
staabm: Please also test with non constant integer.

tests/PHPStan/Analyser/nsrt/count-chars.php
```diff
@@ -0,0 +1,17 @@
+<?php
```
staabm: Files in nsrt/ folder should have a `// lint` comment. 

That way you also can have 2 different files for testing different php versions

stubs/core.stub
```diff
@@ -328,3 +328,8 @@ function ltrim(string $string, string $characters = " \n\r\t\v\x00"): string {}
...
+ */
+function count_chars(string $string, int $mode = 0): array|string {}
```
staabm: please remove the native return type.

u01jmg3: Which part do you mean?

staabm: the native union`: array|string`. see the PHPStan error

> Function count_chars() uses native union types but they're supported only on PHP 8.0 and later.

tests/PHPStan/Analyser/nsrt/conditional-types.php
```diff
@@ -68,7 +68,7 @@ public function testMicrotime(): void
 	}
 
 	/**
-	 * @return ($version is 8 ? true : ($version is 10 ? true : false))
+	 * @return ($version is 8|10 ? true : false)
```
staabm: please don't change unrelated tests

u01jmg3: Sure, I can undo. I changed it because it's the example I copied from and then @mvorisek kindly suggested a worthy edit so going forward I didn't want others to make the same mistake I made.

staabm: yeah thats true for stub declarations. but not for tests in which all kind of scenarios want to be tested

staabm:
> Still need to support the method returning `false` in PHP <8

for a php version dependent return-type a [`DynamicFunctionReturnTypeExtension`](https://phpstan.org/developing-extensions/dynamic-return-type-extensions) would be a better fit then a stub.
sorry, I just saw this comment right now.

we have a few of them in the codebase already for inspiration.

u01jmg3:
Yikes - as this codebase is so new to me, I'm struggling to know where to start. It looks like this is the first time a string function has made use of the [`DynamicReturnTypeExtension`](https://phpstan.org/developing-extensions/dynamic-return-type-extensions).

[`NativeReflectionEnumReturnDynamicReturnTypeExtension`](https://github.com/phpstan/phpstan-src/blob/2.0.x/src/Reflection/PHPStan/NativeReflectionEnumReturnDynamicReturnTypeExtension.php) looks to be one of the simplest but I don't know what [`count_chars()`](https://www.php.net/manual/en/function.count-chars.php) needs for `getClass()|isMethodSupported()|getTypeFromMethodCall()` and all of this just to capture and also return `false` on PHP <8. 😮‍💨

- Is there much value when PHPStan 2 requires `^8.1`?
- Is this PR no longer valid as is?

staabm:
> t looks like this is the first time a string function has made use of the [`DynamicReturnTypeExtension`](https://phpstan.org/developing-extensions/dynamic-return-type-extensions).

see e.g. https://github.com/phpstan/phpstan-src/blob/2.0.x/src/Type/Php/AbsFunctionDynamicReturnTypeExtension.php which is similar to what we need.

> * Is there much value when PHPStan 2 requires `^8.1`?

PHPStan 2.x will have min php version of 7.4.
8.1 is only used for development in the code-base but is downgraded for 7.4 compat on release.

> * Is this PR no longer valid as is?

the test-cases are still useful

src/Type/Php/CountCharsFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,50 @@
...
+		return $functionReflection->getName() === 'count_chars';
+	}
+
+	public function getTypeFromFunctionCall(
+		FunctionReflection $functionReflection,
+		FuncCall $functionCall,
+		Scope $scope,
+	): Type
+	{
+		$args = $functionCall->getArgs();
```
staabm: you need to check args count and `return null` early in case not enough args are provided.

src/Type/Php/CountCharsFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,50 @@
...
+
+		if ($scope->getType($args[1]->value)->getValue() <= 2) {
+			$arrayType = new ArrayType(new IntegerType(), new IntegerType());
+
+			return $this->phpVersion->getVersionId() >= 80000
```
staabm: use `phpVersion->throwsValueErrorForInternalFunctions()` instead

staabm (commented):
this looks pretty good already

src/Type/Php/CountCharsFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,57 @@
...
+	): ?Type
+	{
+		if (count($functionCall->getArgs()) < 1) {
+			return null;
+		}
+
+		$secondArg = $functionCall->getArgs()[1]->value;
+		$modeParamType = $scope->getType($secondArg);
+
+		if ($modeParamType instanceof ConstantIntegerType && $modeParamType->getValue() <= 2) {
```
staabm: compare with `IntegerRangeType` like

```
$modeType = $scope->getType($args[1]->value);
if (IntegerRangeType::fromInterval($min, $max)->isSuperTypeOf($modeType)->yes())
```

tests/PHPStan/Analyser/nsrt/count-chars.php
```diff
@@ -0,0 +1,21 @@
+<?php // lint >= 8.0
```
staabm: we need another test-file with expectations for `// lint <= 7.4`

staabm (approved):
Congratulations. You successfully build your first PHPStan extension :-)

u01jmg3:
@staabm: thanks for your patience/guidance 🙏🏻

src/Type/Php/CountCharsFunctionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,58 @@
...
+			return $this->phpVersion->throwsValueErrorForInternalFunctions()
+				? $arrayType
+				: new UnionType([$arrayType, new ConstantBooleanType(false)]);
+		}
+
+		$stringType = new StringType();
+
+		return $this->phpVersion->throwsValueErrorForInternalFunctions()
+			? $stringType
+			: new UnionType([$stringType, new ConstantBooleanType(false)]);
```
zonuexe: Even in PHP 7.x and earlier, [an error would be raised before returning `false`](https://3v4l.org/ueWe1#veol), so I think wrapping the type in a [benevolent union](https://github.com/phpstan/phpstan/discussions/6440#discussioncomment-2026787) is justified.

```suggestion
			: TypeUtils::toBenevolentUnion(new UnionType([$stringType, new ConstantBooleanType(false)]));
```

Similarly for another `new UnionType`.

ondrejmirtes:
Thank you.

PR #3598, opened 2024-11-02
Don't report "is always true/false" for conditionally defined variables

staabm:
closes https://github.com/phpstan/phpstan/issues/6830


ondrejmirtes:
The bug here is that PHPStan complains about "maybe undefined variable". Which it does only in foreach: https://phpstan.org/r/f934106d-8765-4463-8344-40b1c64ffc37

PR #3602, opened 2024-11-04

staabm:
closes https://github.com/phpstan/phpstan/issues/11945

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -46,7 +46,6 @@ use PHPStan\Node\IssetExpr;
 use PHPStan\Node\Printer\ExprPrinter;
 use PHPStan\Node\PropertyAssignNode;
 use PHPStan\Parser\ArrayMapArgVisitor;
-use PHPStan\Parser\ImmediatelyInvokedClosureVisitor;
 use PHPStan\Parser\NewAssignedToPropertyVisitor;
 use PHPStan\Parser\Parser;
 use PHPStan\Php\PhpVersion;
@@ -4795,7 +4794,6 @@ final class MutatingScope implements Scope
 	 * @param Expr\ClosureUse[] $byRefUses
 	 */
 	public function processClosureScope(
-		Expr\Closure $expr,
 		self $closureScope,
 		?self $prevScope,
 		array $byRefUses,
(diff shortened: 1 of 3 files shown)


ruudk:
💙  You are amazing!

After your comment I was trying this too, and was able to fix the _after_ IIFE. But you seem to have fixed both of them.

🎉 💪

staabm:
I can't really say whether this fix is correct. at least it makes the test work ;)

src/Analyser/MutatingScope.php
```diff
@@ -4828,9 +4826,7 @@ public function processClosureScope(
 				$prevVariableType = $prevScope->getVariableType($variableName);
 				if (!$variableType->equals($prevVariableType)) {
 					$variableType = TypeCombinator::union($variableType, $prevVariableType);
-					if ($expr->getAttribute(ImmediatelyInvokedClosureVisitor::ATTRIBUTE_NAME) !== true) {
-						$variableType = self::generalizeType($variableType, $prevVariableType, 0);
-					}
+					$variableType = self::generalizeType($variableType, $prevVariableType, 0);
```
staabm, line 4829: in this file we effectively revert https://github.com/phpstan/phpstan-src/commit/8fbcf5bdc6250cb4da57ad5945f371bb0c65f931

src/Analyser/NodeScopeResolver.php
```diff
@@ -4291,6 +4292,24 @@ private function processClosureNode(
...
+			$nodeCallback(new ClosureReturnStatementsNode(
+				$expr,
+				$gatheredReturnStatements,
+				$gatheredYieldStatements,
+				$statementResult,
+				$executionEnds,
+				array_merge($statementResult->getImpurePoints(), $closureImpurePoints),
+			), $closureScope);
+
+			return new ProcessClosureResult($scope->processClosureScope($intermediaryClosureScope, null, $byRefUses), $statementResult->getThrowPoints(), $statementResult->getImpurePoints(), $invalidateExpressions);
```
staabm: ~~in case we agree this fix is correct, I think we should add new tests for IIFE regarding impure-points and throw-points~~

ruudk:
@ondrejmirtes You're probably very busy but if you have a spare moment, could you check this PR? It's helping TwigStan because Twig's compiled code contains a lot of immediately involved callables that currently produce false positives. 🙏

ondrejmirtes:
Thank you.

PR #3603, opened 2024-11-04
Do not trigger NodeScopeResolver in PHPUnit data provider

ruudk:
When I'm debugging an issue with PHPStan, and want to go in deep into the NodeScopeResolver, I'm always surprised that the first breakpoints are triggered by this test. It's because work is done in the data provider that calls the NodeScopeResolver.

I think no work should be done in the PHPUnit data providers, because these are always executed. Even when you specify a filter to only run a specific test.

The total time of this test is still the same.

### DIFF
--- a/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
@@ -34,6 +34,9 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 	/** @var Scope[][] */
 	private static array $assertTypesCache = [];
 
+	/** @var array<string, Scope> */
+	private static array $fileScopes = [];
+
 	public function testClassMethodScope(): void
 	{
 		self::processFile(__DIR__ . '/data/class.php', function (Node $node, Scope $scope): void {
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Sorry to be pain about this, but legacy is legacy, I prefer not to invest too much into that code. I'd much prefer if this was rewritten for NodeScopeResolverTest. Which means to move the asserted file into tests/PHPStan/Analyser/nsrt, and then transform the asserts performed in LegacyNodeScopeResolverTest into assertType and assertVariableCertainty calls.

All of this should be done with a script in an automated way, so I can review both the script and the results :)

Some test cases will be definitely harder to rewrite, but most of them should be easy.

ruudk:
I understand. 

But writing a script to automatically convert these legacy tests to the new way seems like a lot of work which I can't allocate time for at the moment. Sorry.

Should this be closed?

I feel it's too bad, because it will impact everyone using the debugger in phpstan-src.

ondrejmirtes:
I'm sorry, I worry this wouldn't work too well. Every time a new static property is introduced, something breaks somewhere. For example a file can be tested multiple times with different configuration. I also don't want to invest in modernizing LegacyNodeScopeResolver. I want to get rid of it.

When I'm debugging something, I always run PHPStan analysis on that file by running `bin/phpstan`. Not through PHPUnit.

PR #3605, opened 2024-11-05
xdebug_get_function_stack: fix signature

janedbal:
I'm not sure why it was wrong, I dont see any signature changes [in changelog](https://xdebug.org/updates).

- https://xdebug.org/docs/all_functions#xdebug_get_function_stack

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -13272,7 +13272,7 @@ return [
 'xdebug_get_declared_vars' => ['array'],
 'xdebug_get_formatted_function_stack' => [''],
 'xdebug_get_function_count' => ['int'],
-'xdebug_get_function_stack' => ['array', 'options='=>'array'],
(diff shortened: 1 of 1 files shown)


resources/functionMap.php
```diff
@@ -13272,7 +13272,7 @@
 'xdebug_get_declared_vars' => ['array'],
 'xdebug_get_formatted_function_stack' => [''],
 'xdebug_get_function_count' => ['int'],
-'xdebug_get_function_stack' => ['array', 'message='=>'string', 'options='=>'int'],
+'xdebug_get_function_stack' => ['array', 'options='=>'array'],
```
ondrejmirtes: Would be nice to introduce array shape with the actual options.

ondrejmirtes (changes requested):

resources/functionMap.php
janedbal: Added

ondrejmirtes:
Thank you.

PR #3607, opened 2024-11-05
Introduce UnionType::filterTypes

VincentLanglet:


ondrejmirtes:
Why it would be nullable? Just make it return self.

VincentLanglet:
> Why it would be nullable? Just make it return self.

I cannot return self, but I can return Type. I updated the PR.

ondrejmirtes:
Thank you!

PR #3608, opened 2024-11-05
Make ArrayFindArgVisitor supports array_any() and array_all()

zonuexe:
After https://github.com/phpstan/phpstan-src/pull/3518 was merged, I remembered that these functions could be supported as well.


ondrejmirtes:
Thank you!

PR #3609, opened 2024-11-06
More details php version information in diagnose

staabm:
as requested in https://github.com/phpstan/phpstan-src/pull/3584#issuecomment-2460260650

grafik

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -2108,6 +2108,7 @@ services:
 		arguments:
 			composerAutoloaderProjectPaths: %composerAutoloaderProjectPaths%
 			allConfigFiles: %allConfigFiles%
+			configPhpVersion: %phpVersion%
 		autowired: false
 
 	# Error formatters
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -70,7 +70,9 @@ parameters:
 		maximumNumberOfProcesses: 32
 		minimumNumberOfJobsPerProcess: 2
 		buffer: 134217728 # 128 MB
-	phpVersion: null
+	phpVersion:
+		min: 80303
+		max: 80104
 	polluteScopeWithLoopInitialAssignments: true
 	polluteScopeWithAlwaysIterableForeach: true
 	polluteScopeWithBlock: true
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -70,9 +70,7 @@ parameters:
 		maximumNumberOfProcesses: 32
 		minimumNumberOfJobsPerProcess: 2
 		buffer: 134217728 # 128 MB
-	phpVersion:
-		min: 80303
-		max: 80104
+	phpVersion: null
 	polluteScopeWithLoopInitialAssignments: true
 	polluteScopeWithAlwaysIterableForeach: true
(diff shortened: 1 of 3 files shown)


src/Diagnose/PHPStanDiagnoseExtension.php
```diff
@@ -48,11 +53,36 @@ public function print(Output $output): void
...
+			$output->writeLineFormatted(sprintf(
+				'<info>PHP version for analysis:</info> %s-%s (from %s)',
+				$minVersion->getVersionString(),
+				$maxVersion->getVersionString(),
+				$this->phpVersion->getSourceLabel(),
+			));
+
+		} else {
+			$output->writeLineFormatted(sprintf(
+				'<info>PHP version for analysis:</info> %s (from %s)',
```
mvorisek, line 90: What is this, isn't phpstan using only one version for the analysis?

staabm: Since https://github.com/phpstan/phpstan-src/pull/3584 you can declare a range in the NEON config-file

mvorisek: That's awesome 😍

ondrejmirtes (changes requested):
This shows me we need to do a bit of a refactoring.

Current state: ComposerPhpVersionFactory either reads `phpVersion` from phpstan.neon (if it has min+max), or from `composer.json`. And ConstantResolver uses ComposerPhpVersionFactory for this logic.

Wanted state: ComposerPhpVersionFactory only reads PHP version from `composer.json`, as its name says. And then ConstantResolver decides if it should read phpVersion from phpstan.neon with min/max provided. And if it's not there then it should read it from ComposerPhpVersionFactory.

Feel free to do this in a separate PR before this one. Thanks.

ondrejmirtes (changes requested):
I'd say the composer.json PHP version is irrelevant if there's phpVersion.min+max version in phpstan.neon. The diagnose command should reflect that and print composer.json version only if phpVersion in phpstan.neon is int or null.

ondrejmirtes:
Looking at ComposerPhpVersionFactory usages, I found a bug: VersionCompareFunctionDynamicReturnTypeExtension only looks at composer version, not phpVersion.min+max version.

staabm:
> Looking at ComposerPhpVersionFactory usages, I found a bug: VersionCompareFunctionDynamicReturnTypeExtension only looks at composer version, not phpVersion.min+max version.

I will send a separate PR

ondrejmirtes:
Awesome, now it makes sense to me :)

conf/config.neon
```diff
@@ -2108,6 +2108,7 @@ services:
 		arguments:
 			composerAutoloaderProjectPaths: %composerAutoloaderProjectPaths%
 			allConfigFiles: %allConfigFiles%
+			configPhpVersion: %phpVersion%
 		autowired: false
 
 	# Error formatters
```
CypherReymax96, line 2114: Fix_Resolve_Complete

CypherReymax96 (approved):

PR #3611, opened 2024-11-08
Use named argument in error for variadic types

ruudk:
When a type is variadic, and multiple named arguments are used, it's often hard to tell for which argument the error is coming.

Since we have the named argument, we could use it in this situation.

Example:
```php
$this->doIpsum(1, 2, foo: $this->foo(), bar: $this->bar(), baz: $this->baz());
```

Before the error would be:
> Parameter ...$args of method NamedArgumentsMethod\Foo::doIpsum() expects string, int given.

And one had to inspect the different method calls to see where it might originate.

After:
> Named argument foo for variadic parameter ...$args of method NamedArgumentsMethod\Foo::doIpsum() expects string, int given.

It's immediately clear it's related to `foo` named argument.


ondrejmirtes (changes requested):
1) Rebase this on top of 2.0.x
2) I'd much rather do it in `describeParameter` private method, but please rethink how the message (messages) should change.

You have about 48 hours to do this change, otherwise we'd have to put it behind bleeding edge :)

ruudk:
I can do that, but let's first agree on the error message:

```diff
-Parameter ...$args of method NamedArgumentsMethod\Foo::doIpsum() expects string, int given.
+Named argument foo for parameter ...$args of method NamedArgumentsMethod\Foo::doIpsum() expects string, int given.
```
?

ondrejmirtes:
What about `Named argument foo for variadic parameter ...$args` in this case? To make it super obvious.

ondrejmirtes:
(Only add "variadic" before parameter only for this specific message, not always.)

ruudk:
Updated the PR

> (Only add "variadic" before parameter only for this specific message, not always.)

Seeing this, I don't think it's what you had in mind.

How to deal with this when I need to do this in `describeParameter`?

Introduce a 3rd nullable argument and only set it for `argument.type`?

ondrejmirtes:
No, the point for doing it in `describeParameter` is that it should always change, for all its usages.

The current version is wrong for non-variadic parameters.

ruudk:
> The current version is wrong for non-variadic parameters.

Could you tell me what's wrong? I'm lost now.

ondrejmirtes:
```php
		if (is_int($positionOrNamed)) {
			$parts[] = 'Parameter #' . $positionOrNamed;
		} elseif (is_string($positionOrNamed)) {
			$parts[] = 'Named argument ' . $positionOrNamed . ' for variadic parameter';
		} else {
			$parts[] = 'Parameter';
		}
```

The `elseif` with message `'Named argument ' . $positionOrNamed . ' for variadic parameter'` is triggered also for non-variadic parameters. That's wrong. You can see it from the tests you modified.

For example: `Named argument b for variadic parameter $b` - does not make sense, `$b` is not variadic, I can see that, because there's no `...` in front of `$b`.

ruudk:
Wow, that doesn't make any sense at all indeed.

ruudk:
Alright, I think I have it now.

ondrejmirtes:
Thank you!

PR #3613, opened 2024-11-08
Introduce uppercase-string

pmjones:
This PR introduces `uppercase-string` in all places where `lowercase-string` is recognized, **except** for _SprintfFunctionDynamicReturnTypeExtension_.

The omission is because `sprintf()` format strings cannot be typed reliably as `uppercase-string` because most of the formatting specifiers are lowercase. For example, the format string `'FOO %s BAR'` is treated as a plain `string` rather than `uppercase-string`, because `%s` is lowercase (even though it might represent an uppercase string).

It appears that `lowercase-string` exhibits similar behavior. For example, consider the format string `foo %F`. It formats as the lowercase string 'foo' and a non-locale-aware float; the returned string will be the same after `strtolower()`, so it should be valid as a `lowercase-string`.

However, in `nsrt/lowercase-string-sprintf.php`, if you add `assertType('lowercase-string', sprintf('foo %F', $lowercase));`, it will fail with:

```
Expected: lowercase-string
Actual:   non-falsy-string
```

That's because `%F` is uppercase, even though it represents a float.

If you consider these to be problems, they might be resolved by adding a new type, `printf-string`, that can ignore the formatting specifiers when determining `isLowercase()` and `isUppercase()`, but I imagine that would have rather broad implcations.

Let me know how you'd like to proceed.


src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
```diff
@@ -67,12 +76,18 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+					$urlType->isUppercaseString()->yes(),
+				);
 			}
 
 			$componentType = $componentType->toInteger();
 			if (!$componentType instanceof ConstantIntegerType) {
-				return $this->createAllComponentsReturnType($urlType->isLowercaseString()->yes());
+				return $this->createAllComponentsReturnType(
+					$urlType->isLowercaseString()->yes(),
+					$urlType->isLowercaseString()->yes(),
```
staabm:
```suggestion
					$urlType->isUppercaseString()->yes(),
```

pmjones: Good catch, thanks!

src/Type/Accessory/AccessoryNumericStringType.php
```diff
@@ -312,6 +312,11 @@ public function isLowercaseString(): TrinaryLogic
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isUppercaseString(): TrinaryLogic
+	{
+		return TrinaryLogic::createMaybe();
```
ondrejmirtes, line 317: numeric string might be a lowercase-string because of the scientific "e" notation. But can a numeric string ever be an uppercase string?

ondrejmirtes (changes requested):
This is very complete and very nice, thank you! Just one thing I noticed.

src/Type/Accessory/AccessoryNumericStringType.php
mvorisek: `1e100` and `1E100` are both numeric strings, so both LC/UC are allowed

and php prints scientific notation in UC by default - https://3v4l.org/Tg497 - so if `float` is casted into `string`, `numeric-string&uppercase-string` is the best narrowed type

ondrejmirtes:
One more thing - Windows tests are failing https://github.com/phpstan/phpstan-src/actions/runs/11749776276/job/32736693637?pr=3613

pmjones:
> One more thing - Windows tests are failing https://github.com/phpstan/phpstan-src/actions/runs/11749776276/job/32736693637?pr=3613

Yes; it's related to [this change](https://github.com/phpstan/phpstan-src/pull/3613/files#diff-8c472741b225513b7229cb974a69d9efe6b5b145d8c69dafd0e858afacc9e0b4L69-R71) where two reasons are given instead of one, and the PHP_EOL between them causes the Windows failures.

After thinking about it some more, I believe it should remain as one reason (not two) and am trying to track down where the second reason is coming from. Any help there would be appreciated!

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
```diff
@@ -635,10 +635,6 @@ public function testArrayUdiffCallback(): void
 				'Parameter #3 $data_comp_func of function array_udiff expects callable(1|2|3|4|5|6, 1|2|3|4|5|6): int, Closure(string, string): string given.',
 				6,
 			],
-			[
-				'Parameter #3 $data_comp_func of function array_udiff expects callable(1|2|3|4|5|6, 1|2|3|4|5|6): int, Closure(int, int): (literal-string&lowercase-string&non-falsy-string&numeric-string) given.',
```
ondrejmirtes, line 641: Why did this error disappear?

tests/PHPStan/Analyser/nsrt/isset-coalesce-empty-type.php
```diff
@@ -472,7 +472,7 @@ function coalesce()
 
 		assertType('int<0, max>', rand() ?? false);
 
-		assertType('0|lowercase-string', preg_replace('', '', '') ?? 0);
+		assertType('0|(lowercase-string&uppercase-string)', preg_replace('', '', '') ?? 0);
```
ondrejmirtes, line 475: This type shouldn't be possible. My guess is that it wouldn't survive type normalization. Because from their implementation of isSuperTypeOf, these types are mutually exclusive. For the same reason `string&int` can't be created.

I understand it's possible for a string to comply with both constraints, but I'd prefer if this extension produced just a general string instead of `lowercase-string&uppercase-string`.

tests/PHPStan/Analyser/nsrt/literal-string.php
```diff
@@ -51,13 +52,14 @@ public function doFoo($literalString, string $string, $numericString)
...
+		assertType("literal-string&lowercase-string&non-falsy-string&uppercase-string", str_repeat('    1   ', $x));
+		assertType("literal-string&lowercase-string&non-falsy-string&uppercase-string", str_repeat('+1', $x));
 		assertType("literal-string&lowercase-string&non-falsy-string", str_repeat('1e9', $x));
-		assertType("literal-string&lowercase-string&non-falsy-string&numeric-string", str_repeat('19', $x));
+		assertType("literal-string&non-falsy-string&uppercase-string", str_repeat('1E9', $x));
+		assertType("literal-string&lowercase-string&non-falsy-string&numeric-string&uppercase-string", str_repeat('19', $x));
 
 		$x = rand(0,2);
-		assertType("literal-string&lowercase-string", str_repeat('19', $x));
+		assertType("literal-string&lowercase-string&uppercase-string", str_repeat('19', $x));
```
ondrejmirtes, line 62: Same point here

ondrejmirtes (changes requested):
I found a few more things.

ondrejmirtes:
After you rebase on top of 1.12.x with this commit https://github.com/phpstan/phpstan-src/commit/ccfb4ab7a19151925b9434e3245892006b3d9dcd, the rule error tip should appear only once.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -480,10 +481,15 @@ public function resolveConcatType(Type $left, Type $right): Type
 		if ($leftStringType->isLiteralString()->and($rightStringType->isLiteralString())->yes()) {
 			$accessoryTypes[] = new AccessoryLiteralStringType();
 		}
+
 		if ($leftStringType->isLowercaseString()->and($rightStringType->isLowercaseString())->yes()) {
 			$accessoryTypes[] = new AccessoryLowercaseStringType();
 		}
 
+		if ($leftStringType->isUppercaseString()->and($rightStringType->isUppercaseString())->yes()) {
```
mvorisek, line 489:
```suggestion
		elseif ($leftStringType->isUppercaseString()->and($rightStringType->isUppercaseString())->yes()) {
```

src/Type/FloatType.php
```diff
@@ -239,6 +239,11 @@ public function isLowercaseString(): TrinaryLogic
 		return TrinaryLogic::createNo();
 	}
 
+	public function isUppercaseString(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
```
mvorisek, line 244: should be maybe - see https://github.com/phpstan/phpstan-src/pull/3613/files#r1835336748

pmjones:
> After you rebase on top of 1.12.x with this commit [ccfb4ab](https://github.com/phpstan/phpstan-src/commit/ccfb4ab7a19151925b9434e3245892006b3d9dcd), the rule error tip should appear only once.

That did the trick; thanks!

src/Reflection/InitializerExprTypeResolver.php
VincentLanglet: It shouldnot be `elseif` because a string can be both lowercase and uppercase.

Example:  `'33'` or `'-'`.

src/Rules/Comparison/StrictComparisonOfDifferentTypesRule.php
```diff
@@ -93,6 +94,24 @@ public function processNode(Node $node, Scope $scope): array
 			$verbosity = VerbosityLevel::precise();
 		}
 
+		if (
```
VincentLanglet: I think it can be merged with the previous if.
```
$leftType->isConstantScalarValue()->yes()
&& !$leftType->isString()->no()
&& !$rightType->isConstantScalarValue()->yes()
&& !$rightType->isString()->no()
&& (
    TrinaryLogic::extremeIdentity($leftType->isLowercaseString(), $rightType->isLowercaseString())->maybe()
    || TrinaryLogic::extremeIdentity($leftType->isUppercaseString(), $rightType->isUppercaseString())->maybe()
) || (
    ...
)
```

src/Type/Accessory/AccessoryLowercaseStringType.php
```diff
@@ -309,6 +309,11 @@ public function isLowercaseString(): TrinaryLogic
 		return TrinaryLogic::createYes();
 	}
 
+	public function isUppercaseString(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
```
VincentLanglet: Should be `TrinaryLogic::createMaybe()`.

`'42'` or `'-'` is both lowercase and uppercase...

src/Type/Accessory/AccessoryUppercaseStringType.php
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	public function isLiteralString(): TrinaryLogic
+	{
+		return TrinaryLogic::createMaybe();
+	}
+
+	public function isLowercaseString(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
```
VincentLanglet: Should be `TrinaryLogic::createMaybe();`

src/Type/IntersectionType.php
```diff
@@ -341,10 +342,14 @@ private function describeItself(VerbosityLevel $level, bool $skipAccessoryTypes)
 				|| $type instanceof AccessoryNumericStringType
 				|| $type instanceof AccessoryNonFalsyStringType
 				|| $type instanceof AccessoryLowercaseStringType
+				|| $type instanceof AccessoryUppercaseStringType
 			) {
 				if ($type instanceof AccessoryLowercaseStringType && !$level->isPrecise()) {
 					continue;
 				}
+				if ($type instanceof AccessoryUppercaseStringType && !$level->isPrecise()) {
```
VincentLanglet: Might be merged with previous if (?)

src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
```diff
@@ -59,6 +60,8 @@ public function getTypeFromFunctionCall(
 		$stringType = $scope->getType($args[1]->value);
 		if ($stringType->isLowercaseString()->yes()) {
 			$returnValueType = new IntersectionType([new StringType(), new AccessoryLowercaseStringType()]);
+		} elseif ($stringType->isUppercaseString()->yes()) {
```
VincentLanglet: It should not be elseif.

https://phpstan.org/r/4d463db9-c5ce-43a2-8261-b00274381327

```
explode('-', '01-23-45')
```
should give `non-empty-list<lowercase-string&uppercase-string>`

but will give `non-empty-list<lowercase-string>` because of the order of the condition here.

code should be
```
$accessory = [];
if ($stringType->isLowercaseString()->yes()) {
    $accessory[] = new AccessoryLowercaseStringType();
} 
if ($stringType->isUppercaseString()->yes()) {
    $accessory[] = new AccessoryUppercaseStringType();
}
if (count($accessory) > 0) {
    $returnValueType = new IntersectionType([new StringType(), ...$accessory]);
} else {
    $returnValueType = new StringType();
}

File: src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php [hunk 13]
```diff
@@ -105,10 +123,14 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
 
+		if ($urlType->isUppercaseString()->yes()) {
+			return $this->componentTypesPairedConstantsForUppercaseString[$componentType->getValue()] ?? new ConstantBooleanType(false);
+		}
+
 		return $this->componentTypesPairedConstants[$componentType->getValue()] ?? new ConstantBooleanType(false);
 	}
 
-	private function createAllComponentsReturnType(bool $urlIsLowercase): Type
+	private function createAllComponentsReturnType(bool $urlIsLowercase, bool $urlIsUppercase): Type
```
VincentLanglet: I don't think you handle the case where both $urlIsLowercase and $urlIsUppercase are true.

should be
```
$returnTypes = [
					new ConstantBooleanType(false),
					new NullType(),
					IntegerRangeType::fromInterval(0, 65535),
					new IntersectionType([new StringType(), new AccessoryLowercaseStringType(), new AccessoryUppercaseStringType()]),
					$this->createComponentsArray(true, true),
				];
```

But I'm unsure such a real case exist for url...

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
```diff
@@ -141,7 +179,7 @@ private function createAllComponentsReturnType(bool $urlIsLowercase): Type
 		return $this->allComponentsTogetherType;
 	}
 
-	private function createComponentsArray(bool $urlIsLowercase): Type
+	private function createComponentsArray(bool $urlIsLowercase, bool $urlIsUppercase): Type
```
VincentLanglet: Same, here are not handle the case or both param are true. (if a real case exists for url)

stubs/core.stub
```diff
@@ -69,7 +69,13 @@ function str_shuffle(string $string): string {}
 
 /**
  * @param array<mixed> $result
- * @param-out ($string is lowercase-string ? array<int|string, array<mixed>|lowercase-string> : array<int|string, array<mixed>|string>) $result
+ * @param-out ($string is lowercase-string
```
VincentLanglet: Technically, a first check of 
`string is lowercase-string&uppercase-string` should be done since such character exists. (Like '42' or '-')

But I dunno if real case exists with `parse_str`.

stubs/core.stub
```diff
@@ -315,16 +321,16 @@ function is_callable(mixed $value, bool $syntax_only = false, ?string &$callable
 function abs($num) {}
 
 /**
- * @return ($string is lowercase-string ? lowercase-string : string)
+ * @return ($string is lowercase-string ? lowercase-string : ($string is uppercase-string ? uppercase-string : string))
```
VincentLanglet: Should check `$string is lowercase-string&uppercase-string` first.
https://phpstan.org/r/98dd06f1-2543-45c5-96b9-58c0495e3302

or we will have the following issue:
```
'-23-' // isLowercase TRUE ; isUppercase TRUE
trim('-23-', '-') // isLowercase TRUE ; isUppercase FALSE
```

stubs/core.stub
```diff
@@ -315,16 +321,16 @@ function is_callable(mixed $value, bool $syntax_only = false, ?string &$callable
...
  */
 function trim(string $string, string $characters = " \n\r\t\v\x00"): string {}
 
 /**
- * @return ($string is lowercase-string ? lowercase-string : string)
+ * @return ($string is lowercase-string ? lowercase-string : ($string is uppercase-string ? uppercase-string : string))
```
VincentLanglet: Same, should check `$string is lowercase-string&uppercase-string` first.

stubs/core.stub
```diff
@@ -315,16 +321,16 @@ function is_callable(mixed $value, bool $syntax_only = false, ?string &$callable
...
  */
 function ltrim(string $string, string $characters = " \n\r\t\v\x00"): string {}
 
 /**
- * @return ($string is lowercase-string ? lowercase-string : string)
+ * @return ($string is lowercase-string ? lowercase-string : ($string is uppercase-string ? uppercase-string : string))
```
VincentLanglet: Same, should check `$string is lowercase-string&uppercase-string` first.

tests/PHPStan/Analyser/nsrt/isset-coalesce-empty-type.php
VincentLanglet: Losing `lowercase-string&uppercase-string` would create issue, even for users of only one type

```
/** @param lowercase-string $string */
public function acceptsOnlyLowercase($string) {}

/** @var int $int */
$int = ...;

acceptsOnlyLowercase((string) $int) {} // Will report an error
```

`(string) $int` is both a lowercase and uppercase string, so should be accepted by `acceptsOnlyLowercase` but if we lose those accessory in normalization ; wouldn't we get an error "lowercase-string expected, string passed" ?

VincentLanglet (changes requested):
This PR should be copied too. https://github.com/phpstan/phpstan-src/pull/3510

IntegerRangeType::toString and IntegerType::toString are both lowercase-string and uppercase-string.

And a whole reflection on lowercase-string&uppercase-string seems needed.

Cause if this intersection is lost the current code
```
/** @param lowercase-string $string */
public function acceptsOnlyLowercase($string) {}

/** @var int $int */
$int = ...;

acceptsOnlyLowercase((string) $int); // Will report an error
```
and doesn't actually cf
https://phpstan.org/r/e722e920-af05-4c7b-8290-1205df68c4f8

ondrejmirtes:
In this case we need to verify that `TypeCombinator::intersect()` of both lowercase-string and uppercase-string leads to expected result. And both types need to tell that it's possible it can be the other too.

Should be done in TypeCombinatorTest.

src/Type/Accessory/AccessoryNumericStringType.php
pmjones: @ondrejmirtes @mvorisek Is this comment resolved, or do I need to do something here?

tests/PHPStan/Analyser/nsrt/isset-coalesce-empty-type.php
pmjones: I think @VincentLanglet is right -- strtoupper('11') and strtolower('11') both result in '11', which means '11' is both uppercase and lowercase. Am I missing something?

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
pmjones: Looks like it was because I did not collect the accessory string types correctly; corrected by f319d828c but the error message no longer has `lowercase-string` in it, which I *think* is right but please correct me as needed. @VincentLanglet your thoughts?

tests/PHPStan/Analyser/nsrt/isset-coalesce-empty-type.php
staabm: ondrej seems to be on the same page, see https://github.com/phpstan/phpstan-src/pull/3613#issuecomment-2469975969

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
VincentLanglet: To me, you're loosing the `lowercase-string` because of
```
if ($lower && $upper) {
			return;
		}
```

Which means that "25" is lowercase-string&uppercase-string but `["25", "26"]` is not considered as `array<lowercase-string&uppercase-string>` anymore which seems wrong and will lead to false positive I think. (for same reason than https://github.com/phpstan/phpstan-src/pull/3613#discussion_r1835754973) 

I would say accessory need to be kept. 
Maybe we will want to simplify the `description` for errors later ; but with the changes you made, it's not the description you changed, you just lost the accessory type. So to me the patch is not correct.

Maybe the accessory were lost because of the error I already mentionned, like the fact
- AccessoryLowercaseStringType::isUppercaseString is `TrinaryLogic::createMaybe` (and not No)
- AccessoryUppercaseStringType::isLowercaseString is `TrinaryLogic::createMaybe` (and not No)

pmjones: @VincentLanglet Very good, I will address the createMaybe() comment next. Thanks for your patience.

src/Type/Accessory/AccessoryLowercaseStringType.php
pmjones: Fixed in d472b913b.

src/Type/Accessory/AccessoryUppercaseStringType.php
pmjones: Fixed in d472b913b.

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
pmjones: @VincentLanglet I have addressed createMaybe() in d472b913b and updated the relevant uppercase-string tests; however, the corollary lowercase-string tests did *not* need updating, which makes me think something else might be missing. Your thoughts?

src/Rules/Comparison/StrictComparisonOfDifferentTypesRule.php
pmjones: Fixed in c2523e787.

tests/PHPStan/Analyser/nsrt/uppercase-string-parse.php
```diff
@@ -18,7 +18,7 @@ public function parse(string $uppercase, string $string): void
 		if (array_key_exists('foo', $a)) {
 			$value = $a['foo'];
 			if (\is_string($value)) {
-				assertType('uppercase-string', $value);
+				assertType('lowercase-string|uppercase-string', $value);
```
VincentLanglet: This is wrong. I assume it's because of the dynamicReturnType https://github.com/phpstan/phpstan-src/pull/3613#discussion_r1835753441

Can you try
```
@param-out (
$string is lowercase-string&uppercase-string 
? array<int|string, array<mixed>|lowercase-string&uppercase-string>
: ($string is lowercase-string
    ? array<int|string, array<mixed>|lowercase-string>
    : ($string is uppercase-string
        ? array<int|string, array<mixed>|uppercase-string>
        : array<int|string, array<mixed>|string>
        )
     )
) $result
```

VincentLanglet: Same strategy for
-  https://github.com/phpstan/phpstan-src/pull/3613#discussion_r1835753838
- https://github.com/phpstan/phpstan-src/pull/3613#discussion_r1835753877
- https://github.com/phpstan/phpstan-src/pull/3613#discussion_r1835753900

Might solve the trim issue.

But the change in those test need to be reverted.

VincentLanglet (changes requested):

src/Type/IntersectionType.php
pmjones: Fixed in 85bb2bf2d

tests/PHPStan/Analyser/nsrt/uppercase-string-parse.php
pmjones: > This is wrong.

Will revert and add the @param-out fixes; again, thank you for your patient guidance.

VincentLanglet: No problem, this is not an easy PR.
The cohabitation of both lowercase-string and uppercase-string is harder than introducing only one of those accessory types.

stubs/core.stub
pmjones: Resolved in 2f78f966a

stubs/core.stub
pmjones: Resolved in 2f78f966a

stubs/core.stub
pmjones: Resolved in 2f78f966a

stubs/core.stub
pmjones: Fixed in 0c5e7213c

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -518,6 +518,26 @@ public function resolveConcatType(Type $left, Type $right): Type
 		return new StringType();
 	}
 
+	protected function appendAccessoryCasedStringTypes($leftStringType, $rightStringType, array &$accessoryTypes): void
+	{
+		$lower = $leftStringType->isLowercaseString()->and($rightStringType->isLowercaseString())->yes();
+		$upper = $leftStringType->isUppercaseString()->and($rightStringType->isUppercaseString())->yes();
+
+		if ($lower && $upper) {
```
VincentLanglet: To me, this if should be removed.
You can both add lowercase and uppercase.

tests/PHPStan/Analyser/nsrt/uppercase-string-parse.php
```diff
@@ -0,0 +1,36 @@
...
+	 */
+	public function parse(string $uppercase, string $string): void
+	{
+		$a = [];
+		parse_str($uppercase, $a);
+
+		if (array_key_exists('foo', $a)) {
+			$value = $a['foo'];
+			if (\is_string($value)) {
+				assertType('lowercase-string|uppercase-string', $value);
```
VincentLanglet: The fact tests are green with `lowercase-string|uppercase-string` is weird.

I recommend to write `assertType('uppercase-string', $value);` and we'll need to debug this.

tests/PHPStan/Analyser/nsrt/uppercase-string-trim.php
```diff
@@ -0,0 +1,29 @@
...
+	 * @param uppercase-string $uppercase
+	 */
+	public function doTrim(string $uppercase, string $string): void
+	{
+		assertType('lowercase-string|uppercase-string', trim($uppercase));
+		assertType('lowercase-string|uppercase-string', ltrim($uppercase));
+		assertType('lowercase-string|uppercase-string', rtrim($uppercase));
+		assertType('lowercase-string|uppercase-string', trim($uppercase, $string));
+		assertType('lowercase-string|uppercase-string', ltrim($uppercase, $string));
+		assertType('lowercase-string|uppercase-string', rtrim($uppercase, $string));
```
VincentLanglet: The fact tests are green with `lowercase-string|uppercase-string` is weird.

I recommend to write `assertType('uppercase-string', ...);` and we'll need to debug this.

tests/PHPStan/Type/TypeCombinatorTest.php
```diff
@@ -2009,6 +2010,46 @@ public function dataUnion(): iterable
...
+				UnionType::class,
+				'non-empty-string|uppercase-string',
+			],
+			[
+				[
+					new IntersectionType([new StringType(), new AccessoryLiteralStringType()]),
+					new IntersectionType([new StringType(), new AccessoryUppercaseStringType()]),
+				],
+				UnionType::class,
+				'literal-string|uppercase-string',
```
VincentLanglet, line 2051: You should add a test for lowercase-string|uppercase-string
I assumed this should be simplified to string.

tests/PHPStan/Type/TypeCombinatorTest.php
```diff
@@ -3934,6 +3975,46 @@ public function dataIntersect(): iterable
...
+				IntersectionType::class,
+				'non-empty-string&uppercase-string',
+			],
+			[
+				[
+					new IntersectionType([new StringType(), new AccessoryLiteralStringType()]),
+					new IntersectionType([new StringType(), new AccessoryUppercaseStringType()]),
+				],
+				IntersectionType::class,
+				'literal-string&uppercase-string',
```
VincentLanglet, line 4024: You should add a test for `lowercase-string&uppercase-string` which should stay as if.

VincentLanglet (commented):
The current ToDo-list seems to be:

- This comment need to be addressed 
https://github.com/phpstan/phpstan-src/pull/3613/files#r1835752042

- Same for https://github.com/phpstan/phpstan-src/pull/3613/files#r1835752521

- appendAccessoryCasedStringTypes need to be changed.

- `lowercase-string|uppercase-string` should be simplified as `string`

- We need to understand why `parse_url`/`trim` doesn't produce the expected behavior.

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
VincentLanglet: To me the appendAccessoryCasedStringTypes is not correct you're doing
```
if ($lower && $upper) {
			return;
		}
```
but this if condition need to be removed.

If the string is both lower and upper, both accessory need to be added.

src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
pmjones: Fixed in 3c6a7b3a3 -- note the modifications to the test cases as well.

tests/PHPStan/Analyser/nsrt/literal-string.php
pmjones: I think this also is to be left as-is -- please correct me if I'm wrong here!

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
pmjones: Resolved in b6b7b49d9 -- I think that's the real fix. Note that the error now has both lowercase-string and uppercase-string in it.

src/Reflection/InitializerExprTypeResolver.php
pmjones: Removed in b6b7b49d9

tests/PHPStan/Analyser/nsrt/uppercase-string-parse.php
pmjones: Tests reverted in e99bf4f85

tests/PHPStan/Analyser/nsrt/uppercase-string-parse.php
pmjones: Addressed in e99bf4f85

tests/PHPStan/Analyser/nsrt/uppercase-string-trim.php
pmjones: Addressed in e99bf4f85

VincentLanglet: I feel like the trim issue is because of `uppercaseString::isLowercase` is maybe and how conditional return type works.

I tried and the issue is solved with a real dynamicReturnTypeExtension, you should cherry-pick/copy this https://github.com/phpstan/phpstan-src/commit/22f0cab4b286a0e819056ff07dd3680ccba6e9f6

Maybe the same thing could be done for parse_str ? (and mb_parse_str, I just discovered)...
but I dunno if there is something similar to dynamicReturnTypeExtension for param-out.

Edit: Yes, we need a FunctionParameterOutTypeExtension so 
You should take a look at https://github.com/phpstan/phpstan-src/commit/38151b7667d240b1da9984036c54e6cd180751da for parse_str issue.

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
pmjones: Addressed in d0685570d

> But I'm unsure such a real case exist for url...

Same here.

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
pmjones: Fixed in 554f6140b

The solution may be "too clever" -- please let me know if you'd prefer a different approach.

tests/PHPStan/Analyser/nsrt/uppercase-string-trim.php
VincentLanglet: I saw you resolved this comment ; I just want to be sure you saw my solution with the extension @pmjones

pmjones: I see them now, and will copy -- thanks!

pmjones: @VincentLanglet Copied and added in ebd50a326 -- however, [tests/PHPStan/Analyser/data/param-out.php](https://github.com/phpstan/phpstan-src/pull/3613/files#diff-0718f319732d6bab708878e5f02d9cd20d0f7a135f2f6bb6a3cbc89c6547febe) began to fail, so I modified the test to update the assertion. Let me know if that should be reverted.

VincentLanglet: To me it's ok. The existing param-out was wrong since
```
mb_parse_str('42=foo', $output);
```
gives an array<int, string>
and was inferred as `array<string, string>' (https://phpstan.org/r/9ab67bf6-3269-4585-81b4-1618efcf6d75)

This can be improved but the fact we can compute the result of parse_str and mb_parse_str for constant string should be done in another PR.

src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
```diff
@@ -57,10 +58,17 @@ public function getTypeFromFunctionCall(
...
+		$accessory = [];
 		if ($stringType->isLowercaseString()->yes()) {
-			$returnValueType = new IntersectionType([new StringType(), new AccessoryLowercaseStringType()]);
+		    $accessory[] = new AccessoryLowercaseStringType();
+		}
+		if ($stringType->isUppercaseString()->yes()) {
+		    $accessory[] = new AccessoryUppercaseStringType();
+		}
+		if (count($accessory) > 0) {
+		    $returnValueType = new IntersectionType([new StringType(), ...$accessory]);
```
VincentLanglet:
```
$accessoryTypes[] = new StringType();
$returnValueType = new IntersectionType($accessoryTypes);
```
will maybe avoid lint error on lower php versions.

src/Type/Php/ParseStrParameterOutTypeExtension.php
```diff
@@ -0,0 +1,59 @@
...
+		$stringType = $scope->getType($args[0]->value);
+		$accessory = [];
+		if ($stringType->isLowercaseString()->yes()) {
+			$accessory[] = new AccessoryLowercaseStringType();
+		}
+		if ($stringType->isUppercaseString()->yes()) {
+			$accessory[] = new AccessoryUppercaseStringType();
+		}
+		if (count($accessory) > 0) {
+			$valueType = new IntersectionType([new StringType(), ...$accessory]);
```
VincentLanglet:
```
$accessoryTypes[] = new StringType();
$returnValueType = new IntersectionType($accessoryTypes);
```
will maybe avoid the lint error

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
```diff
@@ -105,19 +123,43 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 			return $this->componentTypesPairedConstantsForLowercaseString[$componentType->getValue()] ?? new ConstantBooleanType(false);
 		}
 
+		if ($urlType->isUppercaseString()->yes()) {
```
VincentLanglet: Dunno if you shouldn't handle the case 
```
$urlType->isLowercaseString()->yes() && $urlType->isUppercaseString()->yes()
```
before ?

tests/PHPStan/Analyser/nsrt/uppercase-string-trim.php
pmjones: Excellent -- resolving this conversation, please unresolve as you see fit.

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
```diff
@@ -105,19 +123,43 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+					new ConstantBooleanType(false),
+					new NullType(),
+					IntegerRangeType::fromInterval(0, 65535),
+					new IntersectionType([new StringType(), new AccessoryLowercaseStringType(), new AccessoryUppercaseStringType()]),
+					$this->createComponentsArray(true, true),
+				];
+
+				$union = TypeCombinator::union(...$returnTypes);
+				$this->allComponentsTogetherTypeForLowercaseString = $union;
+				$this->allComponentsTogetherTypeForUppercaseString = $union;
```
VincentLanglet: It doesn't work this way ; you'll have to create a 
```
allComponentsTogetherTypeForLowercaseAndUppercaseString
```
or you'll cache a wrong value.

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
```diff
@@ -105,19 +123,43 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
 		return $this->componentTypesPairedConstants[$componentType->getValue()] ?? new ConstantBooleanType(false);
 	}
 
-	private function createAllComponentsReturnType(bool $urlIsLowercase): Type
+	private function createAllComponentsReturnType(bool $urlIsLowercase, bool $urlIsUppercase): Type
 	{
+		if ($urlIsLowercase && $urlIsUppercase) {
+			if (
+				$this->allComponentsTogetherTypeForLowercaseString === null
+				&& $this->allComponentsTogetherTypeForLowercaseString === null
```
VincentLanglet: You'll need a `allComponentsTogetherTypeForLowercaseAndUppercaseString` i'm afraid.

Cause you shouldn't rely on allComponentsTogetherTypeForLowercaseString or allComponentsTogetherTypeForUppercaseString.

But maybe it's overcomplicated to handle both lowercase-string and uppercase-string here cause an url with only non-letter caracter maybe doesn't exists ?

WDYT @pmjones ? Maybe we should drop the case lowercase-string&uppercase-string in ParseUrl function.

VincentLanglet (commented):
We also need `lowercase-string|uppercase-string` to be simplified to `string`

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
pmjones: > Maybe we should drop the case lowercase-string&uppercase-string in ParseUrl function.

I am *absolutely* good with that -- I started out by trying to have uppercase available wherever lowercase was, but ParseUrl has become a bit much. I'll put it back the way it was before uppercase.

VincentLanglet: 👍

pmjones: Removed in 356e0f810

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
pmjones: Removed in 356e0f810

src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php
pmjones: Removed in 356e0f810

src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
pmjones: Applied in 4c54a1a44

src/Type/Php/ParseStrParameterOutTypeExtension.php
pmjones: Applied in 4c54a1a44

src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
pmjones: Linting issues resolved.

src/Type/Php/ParseStrParameterOutTypeExtension.php
pmjones: Linting issues resolved.

pmjones:
@VincentLanglet ["This PR should be copied too. https://github.com/phpstan/phpstan-src/pull/3510"](https://github.com/phpstan/phpstan-src/pull/3613#pullrequestreview-2425817843)

That one is merged into 1.12.x, and I have rebased since that PR, so I think it has been included now. Let me know if I am mistaken.

src/Type/Accessory/AccessoryNumericStringType.php
pmjones: Resolving this for now; please unresolve it as necessary.

tests/PHPStan/Type/TypeCombinatorTest.php
pmjones: Added in 96773a4d2

tests/PHPStan/Type/TypeCombinatorTest.php
pmjones: Added in 96773a4d2

pmjones:
@VincentLanglet  ["We also need `lowercase-string|uppercase-string` to be simplified to `string`"](https://github.com/phpstan/phpstan-src/pull/3613#pullrequestreview-2446561125) -- I admit I do not know where to accomplish this. Hints/tips?

VincentLanglet:
> @VincentLanglet ["This PR should be copied too. https://github.com/phpstan/phpstan-src/pull/3510"](https://github.com/phpstan/phpstan-src/pull/3613#pullrequestreview-2425817843)
> 
> That one is merged into 1.12.x, and I have rebased since that PR, so I think it has been included now. Let me know if I am mistaken.

You need to update 
- IntegerType::toString, it returns 
```
return new IntersectionType([
			new StringType(),
			new AccessoryLowercaseStringType(),
			new AccessoryUppercaseStringType(),
			new AccessoryNumericStringType(),
		]);
```

- IntegerRangeType::toString, it returns
```
$finiteTypes = $this->getFiniteTypes();
		if ($finiteTypes !== []) {
			return TypeCombinator::union(...$finiteTypes)->toString();
		}
		$isZero = (new ConstantIntegerType(0))->isSuperTypeOf($this);
		if ($isZero->no()) {
			return new IntersectionType([
				new StringType(),
				new AccessoryLowercaseStringType(),
			    new AccessoryUppercaseStringType(),
				new AccessoryNumericStringType(),
				new AccessoryNonFalsyStringType(),
			]);
		}

		return new IntersectionType([
			new StringType(),
			new AccessoryLowercaseStringType(),
			new AccessoryUppercaseStringType(),
			new AccessoryNumericStringType(),
		]);
```

VincentLanglet:
> @VincentLanglet ["We also need `lowercase-string|uppercase-string` to be simplified to `string`"](https://github.com/phpstan/phpstan-src/pull/3613#pullrequestreview-2446561125) -- I admit I do not know where to accomplish this. Hints/tips?

Might need to be validated with @ondrejmirtes, but I think it's in `TypeCombinator::union` or `TypeCombinator::compareTypesInUnion`

But it seems not easy to do cause
`lowercase-string|uppercase-string` need to be simplified to `string`
`literal-string&lowercase-string|literal-string&uppercase-string` need to be `literal-string` (same for other accessory)
`literal-string&lowercase-string|uppercase-string` should stay untouched (i think).

Depends on how hard it is to handle, and how important it is to solve ; it might be postponed in another PR.

Except my previous comment, I don't think I see anything to add on this PR ; good job

ondrejmirtes:
I reviewed the whole PR again and I like it a lot! Thank you very much.

If you wanted more challenges in the type system now that you proficient in it, there's no shortage of them I can throw at you :)

staabm:
great job!

VincentLanglet:
> I reviewed the whole PR again and I like it a lot! Thank you very much.
> 
> If you wanted more challenges in the type system now that you proficient in it, there's no shortage of them I can throw at you :)

I had a comment left about IntegerType::toString which is both lowercase and uppercase.

I made the PR https://github.com/phpstan/phpstan-src/pull/3651

pmjones:
Thanks all, glad it turned out well!

thg2k:
Do you think it would be doable to introduce something like html-string and sql-string? Wouldn't be too different from uppercase/lowercase-string. Idea being: mysqli_real_escape_string(string): sql-string, and htmlspecialchars(string): html-string, then concat would preserve the semantics (does it?) and stuff like that... The idea being that then mysqli_query(sql-string) and all output(html-string).. just gathering some preliminary input, i'd open a separate issue to discuss that.

ondrejmirtes:
@thg2k I plan to do "nominal string type". Which means you'll be able to do `whatever-string` without any special meaning to PHPStan. But `whatever-string` will only accept another `whatever-string`. And it will be user's job to write their `@phpstan-assert` PHPDocs to narrow the types.

ondrejmirtes:
The implementation would look very similar to this PR.

PR #3616, opened 2024-11-09
Introduce isDeprecated to CallableParametersAcceptor

kamil-zacek:


src/Type/ClosureType.php
```diff
@@ -143,6 +143,11 @@ public function isPure(): TrinaryLogic
 		return $certainCount > 0 ? TrinaryLogic::createNo() : TrinaryLogic::createMaybe();
 	}
 
+	public function isDeprecated(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
```
ondrejmirtes: This is not the case. First-class callables like `Foo::method(...)` are resolved to ClosureType so the deprecated-ness should be carried over.

ondrejmirtes (changes requested):
Also please review all `new CallableType` to make sure we haven't missed a place where it should be added.

src/Type/ClosureType.php
kamil-zacek: I am not sure how to deal with it. Do you think the class ClosureType itself should have a member "isDeprecated" which will be in the constructor and resolved during the creation of instance of ClosureType itself?

staabm:
I guess rhis was rebased against 2.1.x but the PR target needs still to be changed to 2.1.x

kamil-zacek:
> I guess rhis was rebased against 2.1.x but the PR target needs still to be changed to 2.1.x

Yes, I know it. I will close and reopen PR when I receive reply to my previous question.

PR #3617, opened 2024-11-09
Improve `for` loop initial statement scope pollution

herndlm:
Added a similar method on `Mutatingscope` as `foreach` handling has it already. This should keep expressions in the scope we can safely keep while still removing the initial expressions when `polluteScopeWithLoopInitialAssignments` is set to `false`.

Before, it was merging `$finalScope` with `$scope` to get rid of initial expressions which has the downside of also getting rid of expressions that _only_  exist in `$finalScope` because e.g. the loop is iterating at least once and assigns a variable inside.

Closes https://github.com/phpstan/phpstan/issues/9550

Btw a similar thing seems to be going on with foreach and `polluteScopeWithAlwaysIterableForeach` set to `false`.

PR #3618, opened 2024-11-10
Improve `foreach` scope pollution

herndlm:
equivalent of https://github.com/phpstan/phpstan-src/pull/3617 but for `foreach` and `polluteScopeWithAlwaysIterableForeach` set to `false`.

not sure yet if there's a better way of dealing with the keyVar/valueVar variable expression checks..

don't know yet what this might be fixing..

### DIFF
--- a/tests/PHPStan/Analyser/data/foreach-loop-no-scope-pollution.php
+++ b/tests/PHPStan/Analyser/data/foreach-loop-no-scope-pollution.php
@@ -1,0 +1,115 @@
+<?php declare(strict_types=1);
+
+namespace ForEachLoopNoScopePollution;
+
+use PHPStan\TrinaryLogic;
+use function PHPStan\Testing\assertNativeType;
+use function PHPStan\Testing\assertType;
+use function PHPStan\Testing\assertVariableCertainty;
+
+class ForEachLoopNoScopePollution
+{
+
+	/** @param int $b */
+	public function loopThatIteratesAtLeastOnce(int $a, $b): void
+	{
+		$items = [17 => 'foo', 'bar' => 19];
+
+		foreach ($items as $key => $item) {
+			$a = rand(0, 1);
+			$b = rand(0, 1);
+			$c = rand(0, 1);
+		}
+
+		assertType("17|'bar'", $key);
+		assertNativeType("17|'bar'", $key);
+		assertVariableCertainty(TrinaryLogic::createMaybe(), $key);
+
+		assertType("19|'foo'", $item);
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
I don't get it. What kind of bug is there in the foreach that needs fixing?

herndlm:
Setting something inside it that does not exist in scope already basically. Weird edge case, but the same as with `for`.

tests/PHPStan/Analyser/data/foreach-loop-no-scope-pollution.php
```diff
@@ -0,0 +1,115 @@
...
+		assertNativeType('int', $a);
+		assertVariableCertainty(TrinaryLogic::createYes(), $a);
+
+		assertType('int<0, 1>', $b);
+		assertNativeType('int', $b);
+		assertVariableCertainty(TrinaryLogic::createYes(), $b);
+
+		assertType('int<0, 1>', $c);
+		assertNativeType('int', $c);
+		assertVariableCertainty(TrinaryLogic::createYes(), $c);
```
ondrejmirtes, line 42: This should not be yes but maybe. That's the point of `polluteScopeWithAlwaysIterableForeach: false`.

Please if `for` loop currently behaves like that then please fix the behaviour too.

herndlm:
But it indeed doesn't look like this was even reported. Not sure if it's worth it then.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/data/foreach-loop-no-scope-pollution.php
herndlm: Oh, the docs just talk about variables set in the loop init statements or the key and value variables. If so, my previous PR basically needs to be reverted and the issue it fixed was by design? :)

I can do this in the evening btw.

ondrejmirtes: Oh yeah, sorry about that  https://github.com/phpstan/phpstan/issues/9550#issuecomment-2466757351

herndlm: No worries. Should I open a PR for all those tests only? In case they add value, otherwise I don't mind of course.

ondrejmirtes: It's fine, I think it's well tested :)

PR #3620, opened 2024-11-11
fix: check for existence of second arg in CountCharsFunctionDynamicReturnTypeExtension

canvural:
Fixes https://github.com/phpstan/phpstan/issues/11994


ondrejmirtes:
Thank you!

PR #3622, opened 2024-11-11
Fix `for` endless loop detection

herndlm:
Doing this in the generalized finalScope which includes an already filtered bodyScope is too late and causes false-positives.

Fixes https://github.com/phpstan/phpstan/issues/11992
Improves endless loop detection as in cases like https://github.com/phpstan/phpstan/issues/736
And most likely it fixes what was mentioned in the discussion https://github.com/phpstan/phpstan/discussions/11986 too

Follow-up for https://github.com/phpstan/phpstan-src/pull/3573


ondrejmirtes:
Thank you.

PR #3623, opened 2024-11-11
ClassReflection: resolve missing template type to its default rather than bound

jiripudil:
closes phpstan/phpstan#11899


ondrejmirtes:
Ohhh, super nice!

ondrejmirtes:
Thank you!

PR #3624, opened 2024-11-11
get_defined_vars() return type contains know variables

MartinMystikJonas:
Solves https://github.com/phpstan/phpstan/issues/11948
Solves https://github.com/phpstan/phpstan/issues/11740

Credits to @ruudk

### DIFF
--- a/src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
+++ b/src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
@@ -1,0 +1,46 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+final class GetDefinedVarsFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+{
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return $functionReflection->getName() === 'get_defined_vars';
+	}
+
+	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): Type
+	{
+		if ($scope->canAnyVariableExist()) {
+			return new ArrayType(
+				new StringType(),
+				new MixedType(),
+			);
+		}
+
+		$typeBuilder = ConstantArrayTypeBuilder::createEmpty();
+
+		foreach ($scope->getDefinedVariables() as $variable) {
+			$typeBuilder->setOffsetValueType(new ConstantStringType($variable), $scope->getVariableType($variable), false);
+		}
+
+		foreach ($scope->getMaybeDefinedVariables() as $variable) {
+			$typeBuilder->setOffsetValueType(new ConstantStringType($variable), $scope->getVariableType($variable), true);
+		}
+
+		return $typeBuilder->getArray();
+	}
+
+}
--- a/tests/PHPStan/Analyser/nsrt/get-defined-vars.php
+++ b/tests/PHPStan/Analyser/nsrt/get-defined-vars.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace GetDefinedVars;
+
+use function PHPStan\Testing\assertType;
+
+$global = "foo";
+
+assertType('array<string, mixed>', get_defined_vars()); // any variable can exist
+
+function doFoo(int $param) {
+	$local = "foo";
+	assertType('array{param: int, local: \'foo\'}', get_defined_vars());
+	assertType('array{\'param\', \'local\'}', array_keys(get_defined_vars()));
+}
+
+function doBar(int $param) {
+	global $global;
+	$local = "foo";
+	assertType('array{param: int, global: mixed, local: \'foo\'}', get_defined_vars());
+	assertType('array{\'param\', \'global\', \'local\'}', array_keys(get_defined_vars()));
+}
+
+function doConditional(int $param) {
+	$local = "foo";
+	if(true) {
+		$conditional = "bar";
+		assertType('array{param: int, local: \'foo\', conditional: \'bar\'}', get_defined_vars());
+	} else {
+		$other = "baz";
+		assertType('array{param: int, local: \'foo\', other: \'baz\'}', get_defined_vars());
+	}
+	assertType('array{param: int, local: \'foo\', conditional: \'bar\'}', get_defined_vars());
+}
+
+function doRandom(int $param) {
+	$local = "foo";
+	if(rand(0, 1)) {
+		$random1 = "bar";
+		assertType('array{param: int, local: \'foo\', random1: \'bar\'}', get_defined_vars());
+	} else {
+		$random2 = "baz";
+		assertType('array{param: int, local: \'foo\', random2: \'baz\'}', get_defined_vars());
+	}
+	assertType('array{param: int, local: \'foo\', random2?: \'baz\', random1?: \'bar\'}', get_defined_vars());
+}
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -1526,6 +1526,11 @@ services:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension
 
 	-
+		class: PHPStan\Type\Php\GetDefinedVarsFunctionReturnTypeExtension
+		tags:
+			- phpstan.broker.dynamicFunctionReturnTypeExtension
+
+	-
 		class: PHPStan\Type\Php\GetParentClassDynamicFunctionReturnTypeExtension
 		tags:
 			- phpstan.broker.dynamicFunctionReturnTypeExtension


src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,51 @@
...
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\Type;
+
+// based on code by @ruddk
```
ruudk:
```suggestion
```

And it's ruudk 😉

MartinMystikJonas: Sorry. Fixed that.

tests/PHPStan/Analyser/nsrt/get-defined-vars.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace GetDefinedVars;
+
+use function PHPStan\Testing\assertType;
+
+function doFoo(int $param) {
+	$local = "foo";
+	assertType('array{param: int, local: \'foo\'}', get_defined_vars());
+	assertType('array{\'param\', \'local\'}', array_keys(get_defined_vars()));
```
staabm, line 14: I think a test with conditionally defined variables would be helpful

MartinMystikJonas: Good idea. I added that.

tests/PHPStan/Analyser/nsrt/get-defined-vars.php
```diff
@@ -0,0 +1,18 @@
...
+	if(true) {
+		$conditional = "bar";
+	} else {
+		$other = "baz";
```
staabm: please also assert types also inside this if/else branches.

tests/PHPStan/Analyser/nsrt/get-defined-vars.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace GetDefinedVars;
+
+use function PHPStan\Testing\assertType;
+
+function doFoo(int $param) {
+	$local = "foo";
+	assertType('array{param: int, local: \'foo\'}', get_defined_vars());
+	assertType('array{\'param\', \'local\'}', array_keys(get_defined_vars()));
+	if(true) {
```
staabm, line 26: use 

```suggestion
	if(rand(0,1) {
```

to have a condition which cannot be inferred at static analysis time

MartinMystikJonas: I would keep both to test how it behaves when types can be inferred and when not.

MartinMystikJonas: Test added and it revealed that current implementation did not handled maybe defined variables. So I fixed it. Thanks for pointing out that.

tests/PHPStan/Analyser/nsrt/get-defined-vars.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace GetDefinedVars;
+
+use function PHPStan\Testing\assertType;
+
+function doFoo(int $param) {
```
staabm, line 11: looking at examples from https://www.php.net/manual/en/function.get-defined-vars.php

I think we might need to handle things like (super) global vars in the returned array?

MartinMystikJonas: It should be handled by PHPStan but I will add tests for that too. On global level it would be just `array<string, mixed>` because any variable can exists in global scope. Inside functions it can be determined form use of `$global $var` so it should be resolvable too.

tests/PHPStan/Analyser/nsrt/get-defined-vars.php
```diff
@@ -0,0 +1,46 @@
...
+use function PHPStan\Testing\assertType;
+
+$global = "foo";
+
+assertType('array<string, mixed>', get_defined_vars()); // any variable can exist
+
+function doFoo(int $param) {
+	$local = "foo";
+	assertType('array{param: int, local: \'foo\'}', get_defined_vars());
+	assertType('array{\'param\', \'local\'}', array_keys(get_defined_vars()));
```
ruudk, line 14: When escaping single quotes, I always (when possible) switch to double quoted string to prevent that. 

* easier on the eyes
* easier to copy and paste
* easier to search for the unescaped string in the codebase

```suggestion
	assertType("array{param: int, local: 'foo'}", get_defined_vars());
	assertType("array{'param', 'local'}", array_keys(get_defined_vars()));
```

MartinMystikJonas: I do that too but I checked how it is done in other parts of PHPStan and found out that escaping seems to be standard here.

herndlm: I also often do the double quote. but I think in general it's very inconsistent :D not sure if such thing exists, but a coding standard rule might be good for this 😊
UPDATE: on the other hand - in those tests files it wouldn't be enforced anyway I guess..

ruudk: We use https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#symplifyquoteescaperector and it works good

staabm:
(the unrelated lint build errors will be fixed after rebase; please remove the merge commit)

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,68 @@
...
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use function array_filter;
+use function array_map;
+use function array_merge;
+use function array_values;
+use function count;
+use function range;
+
+// based on code by @ruudk
```
ruudk: Could you remove this
```suggestion
```


If you want to credit me, you can use GitHub's standard method for this, and append this to the commit body:

```
Co-Authored-By: Ruud Kamphuis <ruudk@users.noreply.github.com>
```

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,68 @@
...
+		if ($scope->canAnyVariableExist()) {
+			return new ArrayType(
+				new StringType(),
+				new MixedType(),
+			);
+		}
+
+		$variables = array_values(array_filter(
+			$scope->getDefinedVariables(),
+			static fn ($variable) => $variable !== 'this',
```
ruudk: Not sure if we need to remove `$this`. 

I used that in my example snippet, because I didn't want `this` in there. But in general cases, I think you do want to see it when it's there.

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,68 @@
...
+		$values = array_map(
+			static fn ($variable) => $scope->getVariableType($variable),
+			array_merge($variables, $maybeVariables),
+		);
+
+		if ($maybeVariables !== []) {
+			$maybeIndexes = range(count($variables), count($variables) + count($maybeVariables));
+		} else {
+			$maybeIndexes = [];
+		}
```
ruudk:
```suggestion
		$maybeIndexes = [];
		if ($maybeVariables !== []) {
			$maybeIndexes = range(count($variables), count($variables) + count($maybeVariables));
		}
```

Might be a preference, but I always try to prevent the use of `else` when possible.

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
MartinMystikJonas: ok I removed it

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
MartinMystikJonas: ok

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
MartinMystikJonas: ok

MartinMystikJonas:
@staabm Removed merge commit. Should be ready to merge - two failing checks are unrelated.

staabm:
the "Lint / Coding Standard (pull_request) " fail is related to the PR. just run `make cs-fix` locally and commit the fix

MartinMystikJonas:
CS fixed

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,60 @@
...
+			static fn ($variable) => $scope->getVariableType($variable),
+			$allVariables,
+		);
+
+		$maybeIndexes = [];
+		if ($maybeVariables !== []) {
+			$maybeIndexes = range(count($definedVariables), count($definedVariables) + count($maybeVariables));
+		}
+
+		return new ConstantArrayType($keys, $values, [0], $maybeIndexes);
```
ondrejmirtes: Use ConstantArrayTypeBuilder instead.

ondrejmirtes (changes requested):
Also it might be useful to still do it on 1.12.x instead?

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
MartinMystikJonas: Rewritten

MartinMystikJonas:
> Also it might be useful to still do it on 1.12.x instead?

How to do it properly? Can I just change target branch?

staabm:
> How to do it properly? Can I just change target branch?

try locally running a rebase against 1.12.x. if it works cleanly just force push it and change the target branch of the PR
(make sure your 1.12.x local branch is up 2 date before)

MartinMystikJonas:
Well for some reason (maby I messed up) I was a bit more complicated but it seems it is ok now. I will see if check find any issue and fix it eventualy.

MartinMystikJonas:
Ok it should be ready for merge into 1.12.x

ondrejmirtes (approved):

ondrejmirtes:
Perfect, thank you!

PR #3627, opened 2024-11-12
Refactor ComposerPhpVersionFactory, ConstantResolver

staabm:
as requested in https://github.com/phpstan/phpstan-src/pull/3609#pullrequestreview-2428906861

### DIFF
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -45,7 +45,7 @@ final class ConstantResolver
 		private ReflectionProviderProvider $reflectionProviderProvider,
 		private array $dynamicConstantNames,
 		private int|array|null $phpVersion,
-		private ?ComposerPhpVersionFactory $composerPhpVersionFactory,
+		private ComposerPhpVersionFactory $composerPhpVersionFactory,
 	)
 	{
 	}
@@ -370,10 +370,6 @@ final class ConstantResolver
 			return new PhpVersion($this->phpVersion['min']);
 		}
 
-		if ($this->composerPhpVersionFactory === null) {
-			return null;
-		}
-
 		return $this->composerPhpVersionFactory->getMinVersion();
 	}
 
@@ -389,10 +385,6 @@ final class ConstantResolver
 			}
 
 			return new PhpVersion($this->phpVersion['max']);
-		}
-
-		if ($this->composerPhpVersionFactory === null) {
-			return null;
 		}
 
 		return $this->composerPhpVersionFactory->getMaxVersion();
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -3,10 +3,12 @@
 namespace PHPStan\Analyser;
 
 use PhpParser\Node\Name;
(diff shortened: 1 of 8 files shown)


src/Analyser/ConstantResolver.php
```diff
@@ -35,12 +39,13 @@ final class ConstantResolver
...
 	 * @param string[] $dynamicConstantNames
+	 * @param int|array{min: int, max: int}|null $phpVersion
 	 */
 	public function __construct(
 		private ReflectionProviderProvider $reflectionProviderProvider,
 		private array $dynamicConstantNames,
-		private ?PhpVersion $composerMinPhpVersion,
-		private ?PhpVersion $composerMaxPhpVersion,
+		private int|array|null $phpVersion,
+		private ?ComposerPhpVersionFactory $composerPhpVersionFactory,
```
ondrejmirtes: Why is this nullable?

ondrejmirtes (changes requested):

src/Analyser/ConstantResolver.php
staabm: to allow [creation of the ConstantResolver](https://github.com/phpstan/phpstan-src/blob/38630336ab201cb9ecce6d4104ec1ea5b08e8b7e/src/DependencyInjection/ValidateIgnoredErrorsExtension.php#L68) when the DI is not yet constructed

ondrejmirtes: Just pass `new ComposerPhpVersionFactory` with empty `$composerAutoloaderProjectPaths` there.

src/Analyser/ConstantResolver.php
```diff
@@ -116,25 +121,25 @@ public function resolvePredefinedConstant(string $resolvedConstantName): ?Type
...
-				&& $this->composerMaxPhpVersion->getMajorVersionId() === $this->composerMinPhpVersion->getMajorVersionId()
-				&& $this->composerMaxPhpVersion->getMinorVersionId() === $this->composerMinPhpVersion->getMinorVersionId()
+				$this->getMinPhpVersion() !== null
+				&& $this->getMaxPhpVersion() !== null
+				&& $this->getMaxPhpVersion()->getMajorVersionId() === $this->getMinPhpVersion()->getMajorVersionId()
+				&& $this->getMaxPhpVersion()->getMinorVersionId() === $this->getMinPhpVersion()->getMinorVersionId()
 			) {
-				$minRelease = $this->composerMinPhpVersion->getPatchVersionId();
-				$maxRelease = $this->composerMaxPhpVersion->getPatchVersionId();
+				$minRelease = $this->getMinPhpVersion()->getPatchVersionId();
```
ondrejmirtes: You're calling the getter multiple times and it always leads to creating new object. I'd just do `if (in_array($resolvedConstantName, ['PHP_VERSION_ID', 'PHP_MAJOR_VERSION', 'PHP_MINOR_VERSION']))` and save it to a variable first, then use the variable. inside for each constant.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you! Now onto the diagnose command :)

PR #3629, opened 2024-11-13
Remove from final exit points every exception that is catched

Bellangelo:
Fixes https://github.com/phpstan/phpstan/issues/11906


Bellangelo:
@ondrejmirtes Should I open a PR for the 1.12.x branch as well? Or are you gonna back-porting all issues from the 2.0.x back to the 1.12.x?

staabm:
If you want this PR to apply to 1.x you should target the PR like that.

It usually only makes sense if the PR does not conflict too much with 2.x

ondrejmirtes:
I'm not sure if all of those expected errors should disappear. That's why I didn't merge this yet.

VincentLanglet:
https://github.com/phpstan/phpstan/issues/11906 is now solved

PR #3631, opened 2024-11-13
Utilize phpVersion.min+max in VersionCompareFunctionDynamicReturnTypeExtension

staabm:
see https://github.com/phpstan/phpstan-src/pull/3609#issuecomment-2473114108


ondrejmirtes:
Thank you.

PR #3633, opened 2024-11-14
feat: add TypeCombinator::removeTruthy method

canvural:
This PR adds `TypeCombinator::removeTruthy` method.

Did not add tests because the `remove` method already have tests that covers this case.


ondrejmirtes:
Thank you!

PR #3634, opened 2024-11-15
Update curl_setopt string values and allow nullable

xPaw:
Ref: https://github.com/php/php-src/blob/3f4bcd8cba2538dd4e6641a25598221aaf0f82c5/ext/curl/interface.c#L1949


ondrejmirtes (changes requested):
Please add some tests for this.

xPaw:
Done.

src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -918,49 +918,91 @@ private static function getCurlOptValueType(int $curlOpt): ?Type
...
+			'CURLOPT_UNIX_SOCKET_PATH',
+			'CURLOPT_XOAUTH2_BEARER',
+		];
+		foreach ($nullableStringConstants as $constName) {
+			if (defined($constName) && constant($constName) === $curlOpt) {
+				return new UnionType([
+					new NullType(),
+					TypeCombinator::intersect(
+						new StringType(),
+						new AccessoryNonEmptyStringType(),
```
ondrejmirtes, line 943: /cc @staabm Please tell me how important is the non-empty-string type here? The code around has some constants for both non-empty-string and string, which ones should these be? Thanks.

staabm: I am not exactly sure what the question is.

As I understand it, you ask whether we can/should be using the non-empty-string type here, because it is a more precise parameter type (which is kind of a BC break - and something we usually only do in bleeding edge)?

alternatively we could use `string|null` instead of `non-empty-string|null` which is more forgiving...?

staabm: btw: I wonder whether this method should be refactored to something like

```
$intType = new IntegerType();
$stringType = new StringType();
$stringOrNull = new UnionType([new StringType(), new NullType()]);

$curlConstants = [
  'CURLOPT_BUFFERSIZE' => $intType,
  'CURLOPT_CONNECTTIMEOUT' => $intType,
  // ...
  'CURLOPT_PRE_PROXY' => $stringType,
  // ...
  // ... huge array with all constants
];

foreach ($curlConstants as $constName => $type) {
	if (defined($constName) && constant($constName) === $curlOpt) {
		return $type;
	}
}
```

that way we get a huge curl map, similar to the signature files, instead of this chunks for arrays which are hard to read and diffs of PRs changing types would get more obvious.

ondrejmirtes: I wonder why some parameters are typed as string and some as non-empty-string. And which type should be used for the new constants added here.

xPaw: I kept the same non empty string as its specified below the code I added, I'm only allowing null.

ondrejmirtes: That's what I thought, that's why I'm asking if it should be string or non-empty-string.

staabm: > I wonder why some parameters are typed as string and some as non-empty-string

in the initial implementation we defined nearly all params as non-empty-string.

time passed by and people reported a few constants which mistakenly didn't allow empty-string like in
https://github.com/phpstan/phpstan-src/commit/90403cc656aca2c2e0ae8ba176c461ca490ec24d

so in the end we landed what we have right now with all different kind of types.
I think its not super important to know whether its a non-empty-string or not, but since we already have this difference I would not weaken it to be `string` all over the place.

some of the newly added constants use things I never used before and cannot really judge.

this PR looks good to me though.

staabm: in case we are not sure about real world impact of CURL_* constant changes we could add one of the popular curl wrapping http client classes to the community phpstan builds

- https://packagist.org/packages/guzzlehttp/guzzle

ondrejmirtes:
Thank you.

PR #1529, opened 2022-07-20
Support `ConstantArrayType` callables with union values

herndlm:
Closes https://github.com/phpstan/phpstan/issues/2343

I basically just transformed `findTypeAndMethodName(): ?ConstantArrayTypeAndMethod` to `findTypeAndMethodNames(): ConstantArrayTypeAndMethod[]`.

PR #3637, opened 2024-11-16
Fix ConstantArrayType::isCallable

VincentLanglet:
I basically copied the logic `findTypeAndMethodNames` with the following changes:
- I kept the methods with `no` as result of `$type->hasMethod($method->getValue())`
- I ended up with an `TrinaryLogic::extremIdentity`

Closes https://github.com/phpstan/phpstan/issues/12063

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -482,7 +482,7 @@ class ConstantArrayType extends ArrayType implements ConstantType
 				if (!$phpVersion->supportsCallableInstanceMethods()) {
 					$methodReflection = $type->getMethod($method->getValue(), new OutOfClassScope());
 					if ($classOrObject->isString()->yes() && !$methodReflection->isStatic()) {
-						$has = $has->and(TrinaryLogic::createNo());
+						$has = TrinaryLogic::createNo();
 					}
 				} elseif ($this->isOptionalKey(0) || $this->isOptionalKey(1)) {
 					$has = $has->and(TrinaryLogic::createMaybe());
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -458,7 +458,7 @@ class ConstantArrayType extends ArrayType implements ConstantType
 
 	public function isCallable(): TrinaryLogic
 	{
-		$typeAndMethods = $this->findTypeAndMethodNames(false);
+		$typeAndMethods = $this->findTypeAndMethodNames();
 		if ($typeAndMethods === []) {
 			return TrinaryLogic::createNo();
 		}
@@ -468,7 +468,19 @@ class ConstantArrayType extends ArrayType implements ConstantType
 			$typeAndMethods,
 		);
 
-		return TrinaryLogic::extremeIdentity(...$results);
+		$isCallable = TrinaryLogic::createYes()->and(...$results);
+		if ($isCallable->yes()) {
+			$callableArray = $this->getClassOrObjectAndMethods();
+			if ($callableArray !== []) {
+				[$classOrObject, $methods] = $callableArray;
+
+				if (count($methods->getConstantStrings()) !== count($typeAndMethods)) {
+					return TrinaryLogic::createMaybe();
+				}
+			}
+		}
+
+		return $isCallable;
 	}
 
 	public function getCallableParametersAcceptors(ClassMemberAccessAnswerer $scope): array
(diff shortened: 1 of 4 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -458,17 +458,41 @@ public function equals(Type $type): bool
...
+		$hasMethodTrinary = [];
+		$phpVersion = PhpVersionStaticAccessor::getInstance();
+		foreach ($methods->getConstantStrings() as $method) {
+			$has = $type->hasMethod($method->getValue());
+
+			if ($has->yes()) {
+				if (!$phpVersion->supportsCallableInstanceMethods()) {
+					$methodReflection = $type->getMethod($method->getValue(), new OutOfClassScope());
+					if ($classOrObject->isString()->yes() && !$methodReflection->isStatic()) {
+						$has = $has->and(TrinaryLogic::createNo());
```
VincentLanglet: In `findTypeAndMethodNames` there is a `continue;` I don't want to.

src/Type/Constant/ConstantArrayType.php
```diff
@@ -458,17 +458,41 @@ public function equals(Type $type): bool
...
+				} elseif ($this->isOptionalKey(0) || $this->isOptionalKey(1)) {
+					$has = $has->and(TrinaryLogic::createMaybe());
+				}
+			}
+
+			$hasMethodTrinary[] = $has;
+		}
 
-		return TrinaryLogic::createYes()->and(...$results);
+		return TrinaryLogic::extremeIdentity(...$hasMethodTrinary);
```
VincentLanglet: If all no => no
If different trinaryLogic => maybe
If all yes => yes

VincentLanglet:
cc @herndlm, since you implemented it first in https://github.com/phpstan/phpstan-src/pull/1529/files#diff-2e28cc3b285eae2b224df51cda5d6d532675e945e2995e2cd76293bca56799d6R343

src/Type/Constant/ConstantArrayType.php
ondrejmirtes: Why not `$has = TrinaryLogic::createNo()`?

ondrejmirtes (changes requested):

VincentLanglet:
> Why not $has = TrinaryLogic::createNo()?

Indeed ; simplified.

ondrejmirtes (changes requested):
The implementation if `isCallable()` needs to be synced with `getCallableParametersAcceptors` otherwise there are going to be bugs. That's why currently both methods use `$this->findTypeAndMethodNames()`.

VincentLanglet:
> The implementation if `isCallable()` needs to be synced with `getCallableParametersAcceptors` otherwise there are going to be bugs. That's why currently both methods use `$this->findTypeAndMethodNames()`.

I provided another solution in https://github.com/phpstan/phpstan-src/pull/3637/commits/8993085a165f2157c3ba1764d162469118f7db06

I can modify findTypeAndMethodNames directly, but that means that I need to allow calling
```
ConstantArrayTypeAndMethod::createConcrete($type, $method->getValue(), TrinaryLogic::createNo());
```
which impact the usage in
- UnusedPrivateMethodRule
- getCallableParametersAcceptors

So I dunno if:
- `ConstantArrayTypeAndMethod::createConcrete($type, $method->getValue(), TrinaryLogic::createNo());` would make sens
- I need to filter out those value in UnusedPrivateMethodRule and getCallableParametersAcceptors (and add a boolean in findTypeAndMethodNames to filter out automatically)

ondrejmirtes:
I don't understand why we'd need to create `ConstantArrayTypeAndMethod::createConcrete($type, $method->getValue(), TrinaryLogic::createNo())`.

I'm sorry, I don't like the current state of this PR enough to merge it. I'd have to look into it myself.

VincentLanglet:
> I don't understand why we'd need to create `ConstantArrayTypeAndMethod::createConcrete($type, $method->getValue(), TrinaryLogic::createNo())`.

Currently when we call `findTypeAndMethodNames` on a constantArray:
```
array{View, 'existingMethod'|'nonExistingMethod'}
```
the logic done is
- View::existingMethod => exists, we call `createConcrete` on `View,  existingMethod, TrinaryLogic::createYes()`.
- View::nonExistingMethod => doesn't exists, we skip this

Then the array returned is a single `ConstantArrayTypeAndMethod` so checking all the TrinaryLogic gives a Yes, and we think it's always a callable.

We need a way to track that some values were skipped.

> I'm sorry, I don't like the current state of this PR enough to merge it. I'd have to look into it myself.

The other idea I have would be to check that the count of `$methods->getConstantStrings()` is the same than the count of the result of `findTypeAndMethodNames` to be sure no value were skipped (because they are TrinaryLogic::createNo()).
- A count of 0 => No
- A different count => Maybe
- A same count => Keep the existing logic

VincentLanglet:
@ondrejmirtes I repushed the changes with my last idea (based on the count) ; you might prefer it.

VincentLanglet:
Closing in favor of https://github.com/phpstan/phpstan-src/pull/3717 which is in a separate branch

PR #3638, opened 2024-11-16
Use the correct type for final constants

thg2k:
Trying to tame the new level 10.. I want to access class constants as `$var::CONST`, but rightfully so they are of mixed type as they could be overridden...

Accept a `@final` annotation (or the actual native final for php 8.3) to return the correct type.


ondrejmirtes:
Thank you.

PR #3639, opened 2024-11-16
Fix array-shape descriptions with special chars

staabm:
goal is to make sure array-shapes generated via `$constantArrayType->describe(VerbosityLevel::precise())` are parsable with `TypeStringResolver`


ondrejmirtes:
There's `Type::toPhpDocNode()` for that. You can feed that to Printer from phpdoc.parser.

PR #3640, opened 2024-11-16
Fix sprintf dynamic return type

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12065

### DIFF
--- a/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
@@ -29,6 +29,7 @@ use function array_values;
 use function count;
 use function in_array;
 use function intval;
+use function is_array;
 use function is_string;
 use function preg_match;
 use function sprintf;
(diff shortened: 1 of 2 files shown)


src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
```diff
@@ -132,14 +136,14 @@ public function getTypeFromFunctionCall(
 								continue 2;
 							}
 						}
-						$singlePlaceholderEarlyReturn = TypeCombinator::union(...$result);
+						$singlePlaceholderEarlyReturn[] = TypeCombinator::union(...$result);
```
ondrejmirtes: Would be better to do one union at the end instead of these intermediate ones.

ondrejmirtes (changes requested):

src/Type/Php/SprintfFunctionDynamicReturnTypeExtension.php
VincentLanglet: Indeed ; fixed.

ondrejmirtes:
Thank you.

PR #3641, opened 2024-11-17
FunctionCallParametersCheck: Add native parameter type

staabm:
just stumbled over this untyped parameter

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -53,7 +53,7 @@ final class FunctionCallParametersCheck
 	}
 
 	/**
-	 * @param Node\Expr\FuncCall|Node\Expr\MethodCall|Node\Expr\StaticCall|Node\Expr\New_ $funcCall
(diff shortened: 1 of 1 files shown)


src/Rules/FunctionCallParametersCheck.php
```diff
@@ -61,7 +60,7 @@ public function check(
 		ParametersAcceptor $parametersAcceptor,
 		Scope $scope,
 		bool $isBuiltin,
-		$funcCall,
+		Expr\CallLike $funcCall,
```
ondrejmirtes: Please use exactly the union type in the PHPDoc. CallLike is wider than that - there are more classes there that are not supposed to be passed into FunctionCallParametersCheck.

ondrejmirtes (changes requested):

ondrejmirtes:
I rebased on 2.0.x (no reason to do this on 1.12.x, I'd be a bit afraid). But there's a conflict now :)

staabm:
rebased

ondrejmirtes:
Thank you!

PR #3642, opened 2024-11-17
Implement Scope->getPhpVersion()

staabm:
implements the idea described in https://github.com/phpstan/phpstan-src/pull/3642#discussion_r1845465753


staabm:
sending it up for review to sync our expectations. if this works for you, I can add it to all rules.

when rules take the php-version from the `Scope`, we also need to think how we can/want control the `PhpVersion` in tests, so we can test `PhpVersion` dependent behaviour

src/Analyser/MutatingScope.php
```diff
@@ -5721,4 +5721,27 @@ public function getIterableValueType(Type $iteratee): Type
...
+		}
+
+		$scalars = $versionId->getConstantScalarValues();
+		if ($scalars !== []) {
+			$ints = [];
+			foreach($scalars as $scalar) {
+				if (!is_int($scalar)) {
+					throw new ShouldNotHappenException();
+				}
+				$ints[] = $scalar;
```
ondrejmirtes: Looking at this piece of code I realized it's time to do an idea I had for a long time. Please follow my train of thought (while I'm typing this out on a train to Prague :))

Often there's not a single PHP version, but a range of them so returning `bool` from a method like this isn't exactly right:

```php
	public function supportsNullCoalesceAssign(): bool
	{
		return $this->versionId >= 70400;
	}
```

If we're analysing against both PHP 7.3 and 7.4+, we MIGHT support null coalesce assign.

If we're analysing only against PHP 7.3, we do NOT support null coalesce assign.

If we're analysing only against PHP 7.4+, we DO support null coalesce assign.

Sounds familiar? :) We should return TrinaryLogic instead of `bool`.

But I don't want the headache of BC breaks. Adding a new method about PHP version to Scope allows us to return something else than `PhpVersion`. A new unrelated class called something like `TrinaryPhpVersion` or `PhpVersions`. And we do not need to add all methods from PhpVersion on it. We can add them slowly as needed.

So in this PR this new class could have just `producesWarningForFinalPrivateMethods(): TrinaryLogic` correctly computed for all various scenarious of what `PHP_VERSION_ID` can be in that place.

And the rule would correctly handle yes/no/maybe along with some tests.

The constructor of this class could just accept a list of supported PHP versions like `80000, 80100, 80200...`. Because all methods in PhpVersion asking about this number always ask about the minor version, ending in `00`. Meaning that if `PHP_VERSION_ID` is 80210, the `10` at the end would have to be trimmed and exchanged for `00` to correctly answer.

Does this sound compelling to you?

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: yeah I can follow this train :).

> when rules take the php-version from the `Scope`, we also need to think how we can/want control the `PhpVersion` in tests, so we can test `PhpVersion` dependent behaviour

this is still a open question though. how to adjust existing tests so we can force the php version?

wdyt about adding `RuleTestCase->getPhpVersions()` with a default impl and handle it similar to `getCollectors`?

ondrejmirtes: I don't think we need anything like that. If we want to test behaviour for a specific PHP version we will just write the condition `if (PHP_VERSION_ID > ...)` in the tested code.

src/Php/PhpVersion.php
```diff
@@ -248,6 +248,9 @@ public function supportsPassNoneEncodings(): bool
 		return $this->versionId < 70300;
 	}
 
+	/**
+	 * @deprecated use PhpVersions::producesWarningForFinalPrivateMethods() instead
```
ondrejmirtes: I wouldn't deprecate this.

src/Php/PhpVersions.php
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	public function producesWarningForFinalPrivateMethods(): TrinaryLogic
+	{
+		return $this->minPhpVersion(80000);
+	}
+
+	private function minPhpVersion(int $versionId): TrinaryLogic
+	{
+		if ($this->minVersionId >= $versionId) {
```
ondrejmirtes: This isn't sufficient. Let's say something is supported only between 7.2 and 7.3. If our intervals only allow < 7.1 and > 8.0, it's not okay to collapse all the intervals, it loses the information.

I know this is just theoretical but let's address this now before this becomes a problem in the future.

src/Analyser/Scope.php
```diff
@@ -136,4 +137,6 @@ public function filterByFalseyValue(Expr $expr): self;
 
 	public function isInFirstLevelStatement(): bool;
 
+	public function getPhpVersions(): PhpVersions;
```
ondrejmirtes: I'd call this method `getPhpVersion`

src/Analyser/MutatingScope.php
```diff
@@ -5721,4 +5723,34 @@ public function getIterableValueType(Type $iteratee): Type
 		return $iteratee->getIterableValueType();
 	}
 
+	public function getPhpVersions(): PhpVersions
+	{
+		$versionId = $this->getType(new ConstFetch(new Name('PHP_VERSION_ID')));
```
ondrejmirtes: If PHP_VERSION_ID is not specified (hasExpressionType), I'd create PhpVersions based on PhpVersion from the constructor.

src/Php/PhpVersions.php
```diff
@@ -0,0 +1,58 @@
...
+
+use PHPStan\ShouldNotHappenException;
+use PHPStan\TrinaryLogic;
+use function max;
+use function min;
+
+/**
+ * @api
+ */
+final class PhpVersions
```
ondrejmirtes, line 12: This class needs some unit tests.

src/Php/PhpVersions.php
```diff
@@ -0,0 +1,58 @@
...
+
+	private int $maxVersionId;
+
+	/**
+	 * @api
+	 * @param list<int> $phpVersionIds
+	 */
+	public function __construct(
+		array $phpVersionIds,
+	)
```
ondrejmirtes, line 17: I wouldn't cover the constructor with `@api` for now.

ondrejmirtes (changes requested):

src/Php/PhpVersions.php
staabm: I just realized we could just use the all-problem-solving `Type->isSuperTypeOf`, which means the `PhpVersions` class effectively no longer contains real logic, which requires a separate unit-test, right?

ondrejmirtes: I still want a test that tests the public method for no/maybe/yes. Thanks.

tests/PHPStan/Rules/Methods/FinalPrivateMethodRuleTest.php
```diff
@@ -2,49 +2,47 @@
...
+			],
+		]);
 
-	/**
-	 * @dataProvider dataRule
-	 * @param list<array{0: string, 1: int, 2?: string}> $errors
-	 */
-	public function testRule(int $phpVersion, array $errors): void
-	{
-		$this->phpVersionId = $phpVersion;
```
ondrejmirtes, line 47: Please do not touch existing test, I can't read this diff. It'd be better to add a completely new file with the top-level ifs. Thanks.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Methods/FinalPrivateMethodRuleTest.php
staabm: I now tried to keep the old test as much as possible

ondrejmirtes:
Awesome, thank you! Now we can go through a lot of occurences of `PhpVersion` being injected in the constructor, like in rules and dynamic return type extensions, and use `Scope::getPhpVersion()` there instead.

ondrejmirtes:
Also every occurence of `$this->phpVersion` in MutatingScope.

This assumes we'll add the used methods to `PhpVersions`.

src/Analyser/MutatingScope.php
```diff
@@ -5721,4 +5722,14 @@ public function getIterableValueType(Type $iteratee): Type
 		return $iteratee->getIterableValueType();
 	}
 
+	public function getPhpVersion(): PhpVersions
+	{
+		$versionExpr = new ConstFetch(new Name('PHP_VERSION_ID'));
+		if (!$this->hasExpressionType($versionExpr)->yes()) {
+			return new PhpVersions(new ConstantIntegerType($this->phpVersion->getVersionId()));
```
ondrejmirtes, line 5729: We can afford to start reading min/max values from the phpVersion config.

PR #3643, opened 2024-11-17
Use ConstantScalarTypeTrait in NullType

VincentLanglet:
Hi, I wanted to finish (https://github.com/phpstan/phpstan-src/pull/3484) and therefor following your advice on https://github.com/phpstan/phpstan-src/pull/3485#issuecomment-2376086557.

The best option I see would be to introduce
```
public function compareToScalar(
    ConstantScalarType,
    Phpversion,
    callable
)
```
in the ConstantScalarTypeTrait ; but the NullType would require the exact same logic.

So it would be helpful to use the ConstantScalarTypeTrait in the NullType.
Also, there is multiple method duplicated between both trait/class.

The only issue I found is that ConstantScalarTypeTrait sometimes rely on parent method. This can be avoided with
```
get_parent_class($this) !== false
```
checks ; but PHPStan doesn't understand such things.

Should I just ignore the static analysis error or you don't want such PR ?
Also, I didn't know if I should have targeted the 2.0.x branch for such refacto to avoid conflict on merge ; or if it's ok on 1.12.x.

### DIFF
--- a/src/Type/NullType.php
+++ b/src/Type/NullType.php
@@ -11,6 +11,7 @@ use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantFloatType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\Traits\ConstantScalarTypeTrait;
 use PHPStan\Type\Traits\FalseyBooleanTypeTrait;
 use PHPStan\Type\Traits\NonArrayTypeTrait;
 use PHPStan\Type\Traits\NonCallableTypeTrait;
@@ -23,6 +24,7 @@ use PHPStan\Type\Traits\NonRemoveableTypeTrait;
 class NullType implements ConstantScalarType
 {
 
+	use ConstantScalarTypeTrait;
 	use NonArrayTypeTrait;
 	use NonCallableTypeTrait;
 	use NonIterableTypeTrait;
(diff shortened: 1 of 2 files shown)


staabm:
> but PHPStan doesn't understand such things.

Maybe `method_exists()` works?
Or `is_callable()`

VincentLanglet:
> > but PHPStan doesn't understand such things.
> 
> Maybe `method_exists()` works? Or `is_callable()`

The error comes from 
https://github.com/phpstan/phpstan-src/blob/30d20e6ee2dd93206aed2e26187aba0fc679d5c9/src/Rules/Methods/StaticMethodCallCheck.php#L82-L116
so I get an error because 
```
$currentClassReflection->getParentClass() === null
```
in the context of NullType. PHPStan doesn't understand `get_parent_class($this) !== false` is a safeGuard.

ondrejmirtes:
The method shouldn't just accept ConstantScalarType. It should always be asking a Type about any other Type.

I don't love traits and I'd like to see them be used less, not more.

VincentLanglet:
> The method shouldn't just accept ConstantScalarType. It should always be asking a Type about any other Type.

I really don't understand what you're asking.

Those methods already exists, they are
- looseCompare
- isSmallerThan
- isSmallerOrEqualThan

But there is a special logic needed for ConstantScalar which is that for constantScalar we need to know if we check one of the many following condition based on PHP version and type of constants.
```
(string) $a === $b;
$a === (int) $b;
(int) $a === $b;
...
```

This is a special logic needed only for ConstantScalarType when compared to another ConstantScalarType ; so I don't see how and why I would work with a new method to Type about any other Type. This is similar to the existing
```
if ($type instanceof ConstantScalarType) {
     return LooseComparisonHelper::compareConstantScalars($this, $type, $phpVersion);
}
```

I provided a refacto in https://github.com/phpstan/phpstan-src/pull/3485 which seems the simpler way (in my eyes) to have a reusable method in order to 
- cast A
- cast B
- then compare `A == B` or `A <= B` or `A < B`.

Some guidelines would be helpful 🙏

PR #3644, opened 2024-11-18
3rd parameter of htmlentities & htmlspecialchars allows null

sreichel:
- Fixes https://github.com/phpstan/phpstan/issues/12066


ondrejmirtes:
Thank you!

PR #3646, opened 2024-11-19
Fix subtracting enums inside in_array when const set is used

janedbal:
Resolves https://github.com/phpstan/phpstan/issues/12083


src/Type/Php/InArrayFunctionTypeSpecifyingExtension.php
```diff
@@ -110,7 +110,7 @@ public function specifyTypes(FunctionReflection $functionReflection, FuncCall $n
 			$context->true()
 			|| (
 				$context->false()
-				&& count($arrayValueType->getFiniteTypes()) === 1
+				&& count($arrayValueType->getFiniteTypes()) > 0
```
ondrejmirtes, line 113: 1) Tests are failing
2) There are more places in this file that still do `=== 1`, maybe they could be also `> 0`.

ondrejmirtes (changes requested):

src/Type/Php/InArrayFunctionTypeSpecifyingExtension.php
janedbal: 1. I know, will check tomorrow
2. Looking at the diff that broke it, it feels like a typo only here. The other place was === 1 even before. But I will check it tomorrow too.

ondrejmirtes:
Thank you.

PR #3652, opened 2024-11-20
Fix `iterator_to_array` return type with generators

paulbalandan:
Generators allow any type for its keys. When passed to `iterator_to_array()`, the keys are coerced to string or int where possible. For non-scalar keys, the function throws.

Current PHPStan behaviour: https://phpstan.org/r/2489fc17-69f8-4e75-894a-48e9cb1b211f
Expected behaviours:
  - https://3v4l.org/TodlT
  - https://3v4l.org/QgQWa
  - https://3v4l.org/g2hKC


ondrejmirtes:
Thank you!

PR #3653, opened 2024-11-20
TooWidePropertyTypeRule: dont skip always-read properties

janedbal:
- Since this rule reports assignment issues, it is irrelevant if it is always read or not
- Found this by dropping [similar rule](https://github.com/shipmonk-rnd/phpstan-rules/pull/283) of our own and Doctrine entities are no longer reported:

```php
#[Entity]
class FooEntity 
{

    /**
     * @var array<string>|null
     */
    #[Column(type: Types::SIMPLE_ARRAY, nullable: true)]
    private ?array $zones; // null never assigned

    /**
     * @param list<string> $zones
     */
    public function __construct(
        array $zones,
    )
    {
        $this->zones = $zones;
    }

}
```

### DIFF
--- a/src/Rules/TooWideTypehints/TooWidePropertyTypeRule.php
+++ b/src/Rules/TooWideTypehints/TooWidePropertyTypeRule.php
@@ -61,9 +61,6 @@ final class TooWidePropertyTypeRule implements Rule
 				continue;
 			}
 			foreach ($this->extensionProvider->getExtensions() as $extension) {
-				if ($extension->isAlwaysRead($propertyReflection, $propertyName)) {
-					continue 2;
-				}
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

ondrejmirtes:
FYI I had to revert this. It causes false positives for properties like this:

```php
    #[ORM\Column(type: 'json', nullable: true)]
    private ?array $rawContent;
```

After your change PHPStan reports "is never assigned null so it can be removed from the property type" which is not exactly true, because the DB column allows `null` and there can be `null` values coming from the database.

Ideally we'd have an extension that would tell PHPStan bout the meaning of `#[ORM\Column(type: 'json', nullable: true)]`  (and the fact that `null` can be written there) but we don't have that information right now.

PR #3654, opened 2024-11-22
Refactor TryRemove/Accepts for DateTime|DateTimeImmutable and Exception|Error

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12102


src/Type/UnionType.php
```diff
@@ -45,6 +51,12 @@ class UnionType implements CompoundType
 
 	use NonGeneralizableTypeTrait;
 
+	public const EQUAL_UNION_CLASSES = [
+		DateTimeInterface::class => [DateTimeImmutable::class, DateTime::class],
+		Throwable::class => [Error::class, Exception::class],
+		Traversable::class => [IteratorAggregate::class, Iterator::class],
+	];
```
VincentLanglet: Not sure if I need to put this constant somewhere else (like UnionTypeHelper) / as a static function or with a better naming.

I support any number of classes in values, just in case.

stof:
shouldn't this be implemented through the mechanism of `AllowedSubTypesClassReflectionExtension::getAllowedSubTypes` instead of this special handling ?

ondrejmirtes:
@stof We tried that but it didn't work for some reason. I think that rules reading the "allowed subtypes" extensions would report some unwanted errors.

tests/PHPStan/Rules/Classes/ImpossibleInstanceOfRuleTest.php
```diff
@@ -167,11 +167,6 @@ public function testInstanceof(): void
 					388,
 					$tipText,
 				],
-				[
-					'Instanceof between T of Exception and Error will always evaluate to false.',
```
VincentLanglet, line 171: This was wrong
```
/**
	 * @template T of \Exception
	 * @param T $e
	 */
	public function test(\Throwable $t, $e): void {
		if ($t instanceof $e) return;
		if ($e instanceof $t) return;
	}
```

`test(new Exception(), new LogicException());`
- The first if is false
- The second if is true

src/Type/UnionType.php
```diff
@@ -45,6 +51,12 @@ class UnionType implements CompoundType
 
 	use NonGeneralizableTypeTrait;
 
+	public const EQUAL_UNION_CLASSES = [
+		DateTimeInterface::class => [DateTimeImmutable::class, DateTime::class],
+		Throwable::class => [Error::class, Exception::class], // phpcs:ignore SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException
+		Traversable::class => [IteratorAggregate::class, Iterator::class],
```
ondrejmirtes: I'd accept part of this PR - the refactoring into this constant. Without addition of the Traversable entry.

ondrejmirtes (changes requested):

ondrejmirtes:
I changed the base to 2.0.x, please fix the confict. Thanks.

VincentLanglet:
> I changed the base to 2.0.x, please fix the confict. Thanks.

Done

ondrejmirtes:
Thank you.

PR #3655, opened 2024-11-22
Improve `getopt()` function stub

janedbal:
Closes https://github.com/phpstan/phpstan/issues/12103

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -321,6 +321,7 @@ function get_defined_constants(bool $categorize = false): array {}
 
 /**
  * @param array<string> $long_options
+ * @param mixed $rest_index
  * @param-out int $rest_index
  * @return __benevolent<array<string,string>|array<string,false>|array<string,list<mixed>>|false>
  */
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -322,7 +322,7 @@ function get_defined_constants(bool $categorize = false): array {}
 /**
  * @param array<string> $long_options
  * @param mixed $rest_index
- * @param-out int $rest_index
+ * @param-out positive-int $rest_index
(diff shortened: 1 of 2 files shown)


stubs/core.stub
```diff
@@ -318,3 +318,11 @@ function abs($num) {}
  * @return ($categorize is true ? array<string, array<string, mixed>> : array<string, mixed>)
  */
 function get_defined_constants(bool $categorize = false): array {}
+
+/**
+ * @param array<string> $long_options
+ * @param-out int $rest_index
+ * @return __benevolent<array<string,string>|array<string,false>|array<string,list<mixed>>|false>
+ */
+function getopt(string $short_options, array $long_options = [], int &$rest_index = null): array|false {}
```
janedbal: Should I delete the entry from functionMap?

stubs/core.stub
```diff
@@ -318,3 +318,12 @@ function abs($num) {}
  * @return ($categorize is true ? array<string, array<string, mixed>> : array<string, mixed>)
  */
 function get_defined_constants(bool $categorize = false): array {}
+
+/**
+ * @param array<string> $long_options
+ * @param mixed $rest_index
+ * @param-out int $rest_index
```
ondrejmirtes: Can it be positive-int? Or maybe `int<0, max>`?

ondrejmirtes (changes requested):

stubs/core.stub
janedbal: Tested

```php
#!/usr/bin/env php
<?php

getopt('', [], $rest);

var_dump($rest);
```

both via `php foo.php` and `./foo.php`, always giving `int(1)`, so I used `positive-int`.

ondrejmirtes:
Thank you.

PR #3656, opened 2024-11-23
Prevent reporting `property.inInterface` error when PHP version is `8.4` or later

jakubtobiasz:
After reading Ondrej's tweet and playing with PHPStan `2.0` on PHP `8.4`, I've found out it reports the `property.inInterface` error and it is non-ignorable. My first thought was to allow ignoring this error, but I've ended up not reporting this issue at all once the PHP version is `8.4` or later.

### DIFF
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -32,7 +32,7 @@ final class PropertiesInInterfaceRule implements Rule
 
 		if (!$this->phpVersion->supportsPropertyHooks()) {
 			return [
-				RuleErrorBuilder::message('Interfaces may not include properties.')
+				RuleErrorBuilder::message('Interfaces cannot include properties.')
 					->nonIgnorable()
 					->identifier('property.inInterface')
 					->build(),
@@ -41,7 +41,7 @@ final class PropertiesInInterfaceRule implements Rule
 
 		if (!$node->hasHooks()) {
 			return [
-				RuleErrorBuilder::message('Interfaces may only include hooked properties.')
+				RuleErrorBuilder::message('Interfaces can only include hooked properties.')
 					->nonIgnorable()
 					->identifier('property.nonHookedInInterface')
 					->build(),
@@ -50,7 +50,7 @@ final class PropertiesInInterfaceRule implements Rule
 
 		if (!$node->isPublic()) {
 			return [
-				RuleErrorBuilder::message('Interfaces may not include non-public properties.')
+				RuleErrorBuilder::message('Interfaces cannot include non-public properties.')
 					->nonIgnorable()
 					->identifier('property.nonPublicInInterface')
 					->build(),
@@ -59,7 +59,7 @@ final class PropertiesInInterfaceRule implements Rule
 
 		if ($this->hasAnyHookBody($node)) {
 			return [
-				RuleErrorBuilder::message('Interfaces may not include property hooks with bodies.')
+				RuleErrorBuilder::message('Interfaces cannot include property hooks with bodies.')
 					->nonIgnorable()
 					->identifier('property.hookBodyInInterface')
 					->build(),
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -30,15 +30,6 @@ final class PropertiesInInterfaceRule implements Rule
 			return [];
 		}
 
-		if (!$this->phpVersion->supportsPropertyHooks() && $node->hasHooks()) {
-			return [
-				RuleErrorBuilder::message('Property hooks in interfaces are supported only on PHP 8.4 and later.')
-					->nonIgnorable()
-					->identifier('property.unsupportedHooksInInterface')
-					->build(),
-			];
-		}
-
 		if (!$this->phpVersion->supportsPropertyHooks()) {
 			return [
 				RuleErrorBuilder::message('Interfaces may not include properties.')
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -30,6 +30,15 @@ final class PropertiesInInterfaceRule implements Rule
 			return [];
 		}
 
+		if (!$this->phpVersion->supportsPropertyHooks() && $node->hasHooks()) {
+			return [
+				RuleErrorBuilder::message('Property hooks in interfaces are supported only on PHP 8.4 and later.')
+					->nonIgnorable()
+					->identifier('property.unsupportedHooksInInterface')
+					->build(),
+			];
+		}
+
 		if (!$this->phpVersion->supportsPropertyHooks()) {
 			return [
 				RuleErrorBuilder::message('Interfaces may not include properties.')
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -5,6 +5,7 @@ namespace PHPStan\Rules\Properties;
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
 use PHPStan\Node\ClassPropertyNode;
+use PHPStan\Php\PhpVersion;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 
@@ -13,6 +14,10 @@ use PHPStan\Rules\RuleErrorBuilder;
  */
 final class PropertiesInInterfaceRule implements Rule
 {
+
+	public function __construct(private PhpVersion $phpVersion)
+	{
+	}
 
 	public function getNodeType(): string
 	{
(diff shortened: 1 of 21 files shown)


staabm:
You need a bit more research on the topic, e.g. it seems the properties on interfaces are only allowed when public

https://3v4l.org/TBaXF

and you need tests for the new cases

src/Rules/Properties/PropertiesInInterfaceRule.php
```diff
@@ -7,13 +7,16 @@
...
 use PHPStan\Rules\RuleErrorBuilder;
+use const PHP_VERSION_ID;
 
 /**
  * @implements Rule<ClassPropertyNode>
  */
 final class PropertiesInInterfaceRule implements Rule
 {
 
+	private const PHP_8_4_VERSION_ID = 80400;
```
staabm: PhpVersion id-checks need to go into a new method of PhpVersion class

jakubtobiasz:
Hi @staabm,

Thanks for leaving your review 🤠.

> You need a bit more research on the topic, e.g. it seems the properties on interfaces are only allowed when public

Right, I knew about it but missed it here.

So in case of `protected`/`private` properties:

a) We should just report `Interfaces may not include properties.`
b) We should report `Interfaces may include only public properties.` and leave the `property.inInterface`
c) We should create a separate class with message `Interfaces may include only public properties.` and id `property.nonPublicInInterface`
d) anything else

I bet the option `c)`, but I prefer to ask as it's my first time contrib to phpstan :).

---

At the same time, I guess we should leave this implementation as-is (so `property.inInterface` is never reported on PHP `8.4` and above) and just create a new rule as mentioned in `c)`.

ondrejmirtes:
Please wait for my review before further steps 😊

src/Rules/Properties/PropertiesInInterfaceRule.php
```diff
@@ -25,6 +28,10 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		if (PHP_VERSION_ID >= self::PHP_8_4_VERSION_ID) {
```
ondrejmirtes: This logic needs to use PhpVersion - you inject that into the constructor, and you ask a new method `supportsPropertyHooksInInterfaces`.

ondrejmirtes (changes requested):
Please note that in PHP 8.4 not all properties in interface are allowed. Please read the Interfaces section in https://wiki.php.net/rfc/property-hooks.

1) Only public properties can be in an interface
2) Only properties with get or set or both hooks can be in an interface.
3) Property hook in an interface cannot have a body.

The rule should account for all of this.

Please note that the original error message "Interfaces may not include properties." is not exactly correct on PHP 8.4 so all of these situations should have new messages.

Thanks.

tests/PHPStan/Rules/Properties/NonEmptyPropertyHookRuleTest.php
```diff
@@ -0,0 +1,40 @@
...
+	protected function getRule(): Rule
+	{
+		return new NonEmptyPropertyHookRule(new PhpVersion(PHP_VERSION_ID));
+	}
+
+	public function testRule(): void
+	{
+		if (PHP_VERSION_ID < 80400) {
+			$this->markTestSkipped('This test requires at least PHP 8.4.');
+		}
```
jakubtobiasz: I'm not 100% sure about this approach. In theory it is valid; however, maybe we should mock `PhpVersion` or set `PHP_VERSION_ID` by hand.

jakubtobiasz:
@ondrejmirtes I've implemented all suggestions, and I'm ready for the next ones (if needed 😅).

ondrejmirtes (changes requested):
1) We don't need to do this in multiple rules. It kind of makes it hard to see what situations are covered or not. A decision tree (series of if-else conditions) in a single rule would be much more readable and understandable. So please do this in a single rule. The right moment to split a rule might come later but not now.
2) There are cases not covered by your added code. Properties can be abstract even in classes, meaning at least one of their hooks must not have a body.

jakubtobiasz:
I was focused on interfaces and fixing the issue with `2.0.x` not working with PHP `8.4` while creating this PR. Of course, if you decide this PR should cover property hook as a whole, just let me know, and I'll adjust the PR.

However, from my POV, the first and second bullets are conflicted. If we want to add handling abstract classes with the logic inside the `NonEmptyPropertyHookRule`, I'd recommend leaving it as a separate class (as `PropertyInInterface` doesn't sound good in case we add abstract classes there).

In my humble opinion, we should stay with split classes as they follow SRP and new features seem to not fit the `Property in Interface` as we are in the property hook context. 

So my suggestion is to leave the separation and cover abstract classes as a separate PR, as in such a case this PR could be merged as a bugfix for `2.0`, which would be nice (at least from my perspective) once I want to use `8.4` as soon as possible.

Of course, I'm waiting for your decision 🖖🏻.

ondrejmirtes:
I don't want this PR to completely implement property hooks, that's a much bigger project. I just want it to check what I said I want it to check. So to report a wrong declaration that's detectable just by looking at it.

I don't even want this rule to check missing implementation of an abstract hook from a parent. That's a job for a different rule.

Here's a complete list of things that should be reported by this (single) rule class after this PR:

1) On PHP before 8.4: interfaces can't have properties
2) On PHP before 8.4: properties can't have hooks
3) (All following items are on 8.4+). Properties in interfaces must have at least one hook, and all hooks must not have a body.
4) Non-abstract properties in classes must only have hooks with bodies.
5) Abstract property can only be in an abstract class.
6) Abstract property must have at least one hook without a body.
7) Property in an interface must be public.

I'd understand if you decide not to work on this. In that case you can close this PR.

But in my eyes this is the minimal work needed for the rule to make sense if it's supposed to understand "some properties can now be in interface" which is what you want to achieve.

jakubtobiasz:
> I'd understand if you decide not to work on this. In that case you can close this PR.

Nah, that's not the case. I'll do my best to deliver this :).

I'll use the list you provided while reimplementing it. I guess around the end of the next week.

---

Just last thing — where should I put the code touching abstract classes?

jakubtobiasz:
- On PHP before 8.4: interfaces can't have properties
- On PHP before 8.4: properties can't have hooks
- Properties in interfaces must have at least one hook, and all hooks must not have a body.
- Property in an interface must be public.

Are covered by the latest changes :). The rest will be covered later this week.

By the way, I'm bumping the previous questions 😅
> Just last thing — where should I put the code touching abstract classes?

I don't see any existing rule matching the new rules about property hooks, so it seems natural to create `PropertiesInClassRule`, but I prefer to make sure before I dive into it 😅.

ondrejmirtes:
> Just last thing — where should I put the code touching abstract classes?

Right now you're in a rule about ClassPropertyNode. You can get `$node->getClassReflection()` and ask it whether it's abstract.

ondrejmirtes:
But yeah, these items:

* Non-abstract properties in classes must only have hooks with bodies.
* Abstract property can only be in an abstract class.
* Abstract property must have at least one hook without a body.

Could be added to a separate rule, like `PropertyInClassRule`. (I dislike the plural tbh.)

jakubtobiasz:
Despite my approach with setting the `phpVersion` manually failed, and I'll change it to a regular `$this->markAsSkipped()` once PHP is below `8.4` a new error appeared after upgrading the `ondrejmirtes/better-reflection` package to the latest version.

```
 assert($setHookBody !== null) in /Users/jacob/Projects/jakubtobiasz/phpstan-src/vendor/ondrejmirtes/better-reflection/src/Reflection/ReflectionProperty.php:699

 Caused by
 AssertionError: assert($setHookBody !== null)
```

Do you have any idea how to "omit" it or what I can do with it?

P.S. The code itself is finished, so you can take a look if it looks good in general.

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -0,0 +1,109 @@
...
+	public function getNodeType(): string
+	{
+		return ClassPropertyNode::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$classReflection = $node->getClassReflection();
+
+		if (!$classReflection->isClass() || !$this->phpVersion->supportsPropertyHooks()) {
```
ondrejmirtes: This rule is missing an error: Abstract properties are supported only on PHP 8.4 and later.

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -0,0 +1,109 @@
...
+			return [];
+		}
+
+		if (!$classReflection->isAbstract() && $node->hasHooks() && $node->isAbstract()) {
+			return [
+				RuleErrorBuilder::message('Classes may not include abstract hooked properties.')
```
ondrejmirtes: Fix the message: Non-abstract classes may...

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -0,0 +1,109 @@
...
+					->nonIgnorable()
+					->identifier('property.abstractHookedInClass')
+					->build(),
+			];
+		}
+
+		if (!$classReflection->isAbstract() && $node->hasHooks() && !$this->doAllHooksHaveBody($node)) {
```
ondrejmirtes: I'm sorry, these conditions are unreadable for me. From this structure I can't figure out if we're missing a scenario or not. Please change the structure like this:

```php
		if (!$classReflection->isAbstract()) {
			if ($node->hasHooks()) {
				if ($node->isAbstract()) {
					// return error
				}
				if (!$this->doAllHooksHaveBody($node)) {
					// return error
				}
			}

			return [];
		}

		if ($node->isAbstract()) {
			if (!$node->hasHooks()) {
				// return error
			}
			if (!$this->isAtLeastOneHookBodyEmpty($node)) {
				// return error
			}
		} else {
			if ($node->hasHooks()) {
				// return error
			}
		}
```

ondrejmirtes (changes requested):

ondrejmirtes:
> Do you have any idea how to "omit" it or what I can do with it?

The solution is to skip the test based on `PHP_VERSION_ID`, or change the expected errors based on `PHP_VERSION_ID`. Thanks.

jakubtobiasz:
> The solution is to skip the test based on PHP_VERSION_ID, or change the expected errors based on PHP_VERSION_ID. Thanks.

But the test is also failing on `Tests / Tests (8.4, ubuntu-latest) (pull_request)` & `Tests / Tests (8.4, windows-latest) (pull_request)` for the same reason 🤔. So it seems it's not only a problem of a test being executed on the *wrong* version.

src/Rules/Properties/PropertyInClassRule.php
jakubtobiasz: I had the same feeling; being honest, I forgot to ask about extracting it into two separate classes. However, I went the way you recommended :)

jakubtobiasz:
From what I see in the `\PHPStan\BetterReflection\Reflection\ReflectionProperty::computeImmediateVirtual` should be a check whether the property:
- is in an interface
- is in an abstract class and is marked as abstract

P.S. I let myself resolve your comments; however, if your flow assumes you are resolving them after checking if they are fixed, let me know :).
P.S2 I've double checked, and everything seems to be good on my site 🙈. PHP `7.4` tests are failing due to being not able to parse the file. `8.x` are failing due to the reason above 👆🏻.
P.S.3 `8.3` is throwing another magic error; no idea why, but it also doesn't seem to be related to my changes 🙈.

ondrejmirtes:
I'll take it from here, thanks!

ondrejmirtes (approved):

ondrejmirtes:
Feel free to check out the commits I added. There was one bug in logic (https://github.com/phpstan/phpstan-src/pull/3656/commits/7eaed3a5799a0dfb268b408cddb179fa70019bca), the rest is basically my preferences how to organize code.

ondrejmirtes:
Thank you.

jakubtobiasz:
Thanks for the cooperation 🙌🏼

ondrejmirtes:
Please see:

* https://github.com/phpstan/phpstan/discussions/12337
* https://github.com/phpstan/phpstan/issues/12336

PR #3657, opened 2024-11-23
Avoid false inference with instanceof

VincentLanglet:


src/Analyser/TypeSpecifier.php
```diff
@@ -156,14 +156,17 @@ public function specifyTypesInCondition(
 			}
 
 			$classType = $scope->getType($expr->class);
-			$type = TypeTraverser::map($classType, static function (Type $type, callable $traverse): Type {
+			$uncertainty = false;
+			$type = TypeTraverser::map($classType, static function (Type $type, callable $traverse) use (&$uncertainty): Type {
```
VincentLanglet, line 160: I tried to copy the logic from MutatingScope::inferType with an uncertainty boolean:
https://github.com/phpstan/phpstan-src/blob/adcabb3078603f46c811c7dc460174c9df551310/src/Analyser/MutatingScope.php#L960-L989

tests/PHPStan/Analyser/nsrt/bug-12107.php
```diff
@@ -0,0 +1,43 @@
...
+class HelloWorld
+{
+	public function sayHello(Throwable $e1, LogicException $e2): void
+	{
+		if ($e1 instanceof $e2) {
+			return;
+		}
+
+		assertType('Throwable', $e1);
+		assertType('bool', $e1 instanceof $e2); // could be false
```
VincentLanglet, line 19: I'm unsure to say PHPStan that
- $e1 is not `Throwable~LogicException` but still `Throwable`
- the specific expression `$e1 instanceof $e2` is false because already computed.

But I'm not sure it's a big loss.

tests/PHPStan/Analyser/nsrt/instanceof-class-string.php
```diff
@@ -32,7 +32,7 @@ public function doBar(Foo $foo, Bar $bar): void
 		if ($foo instanceof $class) {
 			assertType(self::class, $foo);
 		} else {
-			assertType('InstanceOfClassString\Foo~InstanceOfClassString\Bar', $foo);
+			assertType('InstanceOfClassString\Foo', $foo);
```
VincentLanglet, line 35: If I call `doFar($foo, $superBar)` with superBar a class extending bar, 
$foo is only `Foo~SuperBar` but can still being Bar.

tests/PHPStan/Analyser/nsrt/instanceof.php
```diff
@@ -80,9 +80,9 @@ public function testExprInstanceof($subject, string $classString, $union, $inter
 			assertType('true', $subject instanceof Foo);
 			assertType('bool', $subject instanceof $classString);
 		} else {
-			assertType('mixed~InstanceOfNamespace\Foo', $subject);
-			assertType('false', $subject instanceof Foo);
-			assertType('false', $subject instanceof $classString);
+			assertType('mixed', $subject);
```
VincentLanglet, line 83: if I pass `class-string<SuperFoo>` for `$classString` the subject will only be `mixed~SuperFoo` but not `mixed~Foo`.

tests/PHPStan/Analyser/nsrt/instanceof.php
```diff
@@ -180,8 +180,8 @@ public function testExprInstanceof($subject, string $classString, $union, $inter
 			assertType('InstanceOfNamespace\Foo', $object);
 			assertType('bool', $object instanceof $classString);
 		} else {
-			assertType('object~InstanceOfNamespace\Foo', $object);
-			assertType('false', $object instanceof $classString);
+			assertType('object', $object);
```
VincentLanglet, line 183: Same idea.

tests/PHPStan/Rules/Classes/ImpossibleInstanceOfRuleTest.php
```diff
@@ -167,11 +167,6 @@ public function testInstanceof(): void
 					388,
 					$tipText,
 				],
-				[
-					'Instanceof between T of Exception and Error will always evaluate to false.',
```
VincentLanglet, line 171: Code is 
```
	/**
	 * @template T of \Exception
	 * @param T $e
	 */
	public function test(\Throwable $t, $e): void {
		if ($t instanceof $e) return;
		if ($e instanceof $t) return;
	}
```

if I call it with `test(new \Exception(), new \LogicException())` the second if is true.

ondrejmirtes:
Thank you!

PR #3658, opened 2024-11-23
Check that values passed to array_sum/product are castable to number

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/11883

It's largely a copy-paste of `ParameterCastableToStringRule` (#3132). I checked php-src and as far as I can tell [this](https://github.com/php/php-src/blob/c84b7ed0c52da92d598b2ddf67275e9c9a1e8d81/Zend/zend_operators.c#L353) is the code that's used by `array_sum`/`array_product` to convert objects to numbers. It seems to correspond to `Type::toNumber()` in PHPStan.

It seems that resource worked before PHP 8.3, but I'd report it anyway, since I can't imagine a scenario where you'd want `array_sum`/`array_product` over resources: https://3v4l.org/lF7SE

### DIFF
--- a/conf/config.level5.neon
+++ b/conf/config.level5.neon
@@ -7,7 +7,7 @@ parameters:
 
 conditionalTags:
 	PHPStan\Rules\Functions\ParameterCastableToNumberRule:
-		phpstan.rules.rule: %featureToggles.bleedingEdge%
+		phpstan.rules.rule: %featureToggles.checkParameterCastableToNumberFunctions%
 
 rules:
 	- PHPStan\Rules\DateTimeInstantiationRule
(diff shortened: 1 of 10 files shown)


conf/config.level5.neon
```diff
@@ -5,6 +5,10 @@ parameters:
 	checkFunctionArgumentTypes: true
 	checkArgumentsPassedByReference: true
 
+conditionalTags:
+	PHPStan\Rules\Functions\ParameterCastableToNumberRule:
+		phpstan.rules.rule: %featureToggles.bleedingEdge%
```
ondrejmirtes: Please give it a new feature toggle.

ondrejmirtes (changes requested):

ondrejmirtes:
Awesome, thank you!

PR #3659, opened 2024-11-23
Scope->getPhpVersion() allows array via phpVersion NEON config

staabm:
> We can afford to start reading min/max values from the phpVersion config.

refs https://github.com/phpstan/phpstan-src/pull/3642#discussion_r1855167754


ondrejmirtes:
Thank you.

PR #3661, opened 2024-11-24
Enabling constructor check for class-string variables

sayuprc:
Changed to check constructor arguments when creating an instance from a variable containing a class-string.
Skip checking if T in class-string<T> is abstract or interface.

resolve phpstan/phpstan#12010

### DIFF
--- a/src/Rules/Classes/InstantiationRule.php
+++ b/src/Rules/Classes/InstantiationRule.php
@@ -16,11 +16,16 @@ use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\CompoundType;
 use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\Generic\TemplateGenericObjectType;
+use PHPStan\Type\StaticType;
+use PHPStan\Type\VerbosityLevel;
 use function array_map;
 use function array_merge;
 use function count;
 use function sprintf;
+use function str_starts_with;
 use function strtolower;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Rules/Classes/InstantiationRule.php
```diff
@@ -245,6 +248,18 @@ private function getClassNames(Node $node, Scope $scope): array
 
 		$type = $scope->getType($node->class);
 
+		if (str_starts_with($type->describe(VerbosityLevel::precise()), 'class-string')) {
```
ondrejmirtes: I'm sorry, rule logic can't rely on string type descriptions.

staabm: could this use `Type->isClassStringType()`?

src/Rules/Classes/InstantiationRule.php
```diff
@@ -245,6 +248,18 @@ private function getClassNames(Node $node, Scope $scope): array
...
+			$classStringObjectType = $type->getClassStringObjectType();
+
+			return array_map(
+				static fn (string $name): array => [$name, true],
+				array_filter($classStringObjectType->getObjectClassNames(), function (string $name): bool {
+					$reflectionClass = $this->reflectionProvider->getClass($name);
+					return !$reflectionClass->isAbstract() && !$reflectionClass->isInterface();
+				}),
+			);
```
staabm: could this be shortened with `Type->getObjectClassReflections()`?

sayuprc:
@staabm 
Thank you for your reviewing!
I tried fixing with [298e55e](https://github.com/phpstan/phpstan-src/pull/3661/commits/298e55e79123c44068a009c11c8b89a042b89007)

staabm (approved):
lgtm thanks. 

please give ondej some time to work thru the queue of open PRs

src/Rules/Classes/InstantiationRule.php
```diff
@@ -245,6 +247,20 @@ private function getClassNames(Node $node, Scope $scope): array
 
 		$type = $scope->getType($node->class);
 
+		if ($type->isClassString()->yes()) {
+			$concretes = array_filter(
+				$type->getClassStringObjectType()->getObjectClassReflections(),
+				static fn (ClassReflection $classReflection): bool => !$classReflection->isAbstract() && !$classReflection->isInterface(),
+			);
+
+			if (0 < count($concretes)) {
```
ondrejmirtes:
```suggestion
			if (count($concretes) > 0) {
```

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #3662, opened 2024-11-24
Named argument detection is scope php-version dependent

staabm:
refs https://github.com/phpstan/phpstan-src/pull/3642#issuecomment-2495429969

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/named-arguments.php
+++ b/tests/PHPStan/Rules/Functions/data/named-arguments.php
@@ -1,7 +1,7 @@
 <?php
 
 namespace FunctionNamedArguments;
-
+if (PHP_VERSION_ID < 80000) return;
 function bar(): void
 {
 	foo(i: 1);
--- a/tests/PHPStan/Rules/Functions/data/named-arguments.php
+++ b/tests/PHPStan/Rules/Functions/data/named-arguments.php
@@ -1,7 +1,7 @@
 <?php
 
 namespace FunctionNamedArguments;
-if (PHP_VERSION_ID < 80000) return;
+
 function bar(): void
 {
 	foo(i: 1);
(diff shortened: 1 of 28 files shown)


tests/PHPStan/Rules/Functions/data/named-arguments.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 namespace FunctionNamedArguments;
-
+if (PHP_VERSION_ID < 80000) return;
```
ondrejmirtes: We don't need this here, right? The test should run only on PHP 8+.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Functions/data/named-arguments.php
staabm: the test which uses this file, has error expectations for PHP 7, see

https://github.com/phpstan/phpstan-src/blob/a0e007aee46a0a3ac56d74789f1107a87d0f5740/tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php#L494-L499

ondrejmirtes: It doesn't make sense to me. PHP 7.x does not support named arguments so it doesn't make sense to check them?

staabm: so do you say I should skip the test on php7, or should the test be made to assert 0 errors on php7?
(and I guess the rule needs adjustments for that)

staabm: I went with skipping the tests for now

ondrejmirtes:
Thank you.

PR #3663, opened 2024-11-24
non-capturing catch support-detection is scope php-version dependent

staabm:
refs https://github.com/phpstan/phpstan-src/pull/3642#issuecomment-2495429969


ondrejmirtes:
Thank you.

PR #3664, opened 2024-11-24
Remove sha256 definition

VincentLanglet:
Cf https://github.com/phpstan/phpstan/issues/12048#issuecomment-2496156102 message 
and https://onlinephp.io/c/37b9d0 example, I feel like this function doesn't exists.

That would explain why there is none in PHPStorm stubs.

I'm opened to counter-example.


ondrejmirtes:
Super interesting! Thank you.

PR #3665, opened 2024-11-24
More precise hash return type

VincentLanglet:
Follow up of https://github.com/phpstan/phpstan-src/pull/3541 (cc @staabm)

Currently `md5('foo')` is considered non-falsy-string&lowercase-string, but `hash('md5', 'foo')` is still non-empty-string.

At first, I planned to list hash algo with lowercase output, but 
- I tried 
```
foreach (hash_algos() as $algo) {
     var_dump(hash($algo, ''));
}
```
and get everytime a non-falsy-string and lowercase-string (when binary is false)

- My research on internet/chatGPT give the same result.

So it seems like it's always `non-falsy-string&lowercase-string` with binary false, and `non-falsy-string` when binary true.


resources/functionMap.php
```diff
@@ -3909,18 +3909,18 @@
 'HaruPage::stroke' => ['bool', 'close_path='=>'bool'],
 'HaruPage::textOut' => ['bool', 'x'=>'float', 'y'=>'float', 'text'=>'string'],
 'HaruPage::textRect' => ['bool', 'left'=>'float', 'top'=>'float', 'right'=>'float', 'bottom'=>'float', 'text'=>'string', 'align='=>'int'],
-'hash' => ['non-empty-string|false', 'algo'=>'string', 'data'=>'string', 'raw_output='=>'bool'],
+'hash' => ['(non-falsy-string&lowercase-string)|false', 'algo'=>'string', 'data'=>'string', 'raw_output='=>'bool'],
```
mvorisek: https://www.php.net/manual/en/function.hash.php - must be conditional return, when `$binary` flag is set, the string can be anything (contain UC easily)

VincentLanglet: Thanks ! I updated the PR, please take a new look

mvorisek: I am still not sure if `non-falsy-string` in the functionMap.php is correct as short hash functions like `crc32` can quite easily produce falsy string.

staabm: @mvorisek can you show an example on 3v4l.org ?

mvorisek: https://3v4l.org/Bo2AT

staabm: so your examples shows us a case where it is `non-falsy-string` - which is the opposite of your thesis?

> `crc32` can quite easily produce falsy string

mvorisek: You are right! `'00..'` is `non-falsy-string` per https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting but only `'00..' == 0` is true.

staabm: AFAIK every string with more then 1 char is non-falsy-string

VincentLanglet:
What do you think about this PR @ondrejmirtes ?

If preferred, I can use `non-empty-string` rather than `non-falsy-string` in order to be safer (?)

ondrejmirtes:
Thank you!

PR #3666, opened 2024-11-25
Scope: use `scope->getConstant` instead

staabm:
while doing some local tests, I realized the constant name was wrong.

I am not sure how to test this in isolation, therefore submitting without a test.
I think this will be covered after we use `$scope->getPhpVersion()` in e.g. return type extensions.

I realized there is a bug, as when testing non-namespaced files the PHP_VERSION_ID was not properly found within the scope


src/Analyser/MutatingScope.php
```diff
@@ -5727,15 +5727,15 @@ public function getIterableValueType(Type $iteratee): Type
 
 	public function getPhpVersion(): PhpVersions
 	{
-		$versionExpr = new ConstFetch(new Name('PHP_VERSION_ID'));
-		if (!$this->hasExpressionType($versionExpr)->yes()) {
```
staabm, line 5732: this previous code does not work in files without a namespace, e.g.

```php
<?php // lint >= 7.4

if (PHP_VERSION_ID < 80000) {
	\PHPStan\dumpType(PHP_VERSION_ID);
}
```

the `dumpType` returns the correct value, but the scope uses a different expression name

```
$scope->debug();
array (
  '\\PHP_VERSION_ID (Yes)' => 'int<50207, 79999>',
  'native \\PHP_VERSION_ID (Yes)' => 'int<50207, 79999>',
)
```

vs. (note: file contains a namespace)

```php
<?php // lint >= 7.4

namespace ANamespace;

if (PHP_VERSION_ID < 80000) {
	\PHPStan\dumpType(PHP_VERSION_ID);
}
```

```
$scope->debug();
array (
  'PHP_VERSION_ID (Yes)' => 'int<50207, 79999>',
  'native PHP_VERSION_ID (Yes)' => 'int<50207, 79999>',
)
```

ondrejmirtes (changes requested):
We should:

1) Add (maybe private) method `getGlobalConstantType` in MutatingScope.
2) Refactor `hasConstant` logic that creates the correct ConstFetch
3) Reuse the logic from 2) in our `getPhpVersion` method
4) Also add tests because without tests I don't trust any change :)

src/Analyser/MutatingScope.php
mvorisek: Isn't `\PHP_VERSION_ID` const fetch enough?

staabm: nope, thats the point of this PR

tests/PHPStan/Analyser/ScopePhpVersionTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+class ScopePhpVersionTest extends TypeInferenceTestCase
+{
+
+	public function dataTestPhpVersion(): array
+	{
+		return [
+			[
+				'int<80000, 80499>',
+				__DIR__ . '/data/global-scope-constants.php',
```
ondrejmirtes: And a test for a namespaced file would be nice

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/ScopePhpVersionTest.php
staabm: here you are

ondrejmirtes:
Thank you.

PR #3667, opened 2024-11-25
Implement SocketCreateReturnTypeExtension

staabm:
php-version dependent `socket_create` handling to ease writing cross php version code

refs https://github.com/phpstan/phpstan/discussions/4212

see https://php.watch/versions/8.0/sockets-sockets-addressinfo

### DIFF
--- a/src/Php/PhpVersions.php
+++ b/src/Php/PhpVersions.php
@@ -33,6 +33,11 @@ final class PhpVersions
 		return IntegerRangeType::fromInterval(80000, null)->isSuperTypeOf($this->phpVersions)->result;
 	}
 
+	public function socketFunctionsUseObject(): TrinaryLogic
+	{
+		return IntegerRangeType::fromInterval(80000, null)->isSuperTypeOf($this->phpVersions)->result;
+	}
+
 	public function supportsNamedArguments(): TrinaryLogic
 	{
 		return IntegerRangeType::fromInterval(80000, null)->isSuperTypeOf($this->phpVersions)->result;
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -1830,6 +1830,16 @@ services:
 			- phpstan.broker.dynamicMethodReturnTypeExtension
 
 	-
+		class: PHPStan\Type\Php\SocketObjectReturnTypeFunctionExtension
+		tags:
+			- phpstan.broker.dynamicFunctionReturnTypeExtension
+
+	-
+		class: PHPStan\Type\Php\AddressInfoObjectReturnTypeFunctionExtension
+		tags:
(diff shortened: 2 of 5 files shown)


src/Php/PhpVersions.php
```diff
@@ -28,4 +28,9 @@ public function producesWarningForFinalPrivateMethods(): TrinaryLogic
 		return IntegerRangeType::fromInterval(80000, null)->isSuperTypeOf($this->phpVersions)->result;
 	}
 
+	public function socketCreateUsesResource(): TrinaryLogic
+	{
+		return IntegerRangeType::fromInterval(null, 79999)->isSuperTypeOf($this->phpVersions)->result;
```
ondrejmirtes: I'd write this as "since 8.0" rather than until "7.9999". So rename this method to what it actually returns since 8.0.

ondrejmirtes (changes requested):

src/Php/PhpVersions.php
staabm: sure

staabm:
broken build is unrelated

staabm:
just found https://php.watch/versions/8.0/sockets-sockets-addressinfo which describes all functions which have been changed from `resource` to `Socket`-objects. 
I have double checked the php.net manual for all of them and it turned out they have the same parameter signature so we can re-use the extension for all of them.

ondrejmirtes:
I'm hesitant to merge this. If PHPStan were to properly merge reflection data across multiple supported PHP versions then this PR wouldn't be needed at all. It's really weird to implement this for a single function when the rest of PHPStan doesn't really support this.

thg2k:
> I'm hesitant to merge this. If PHPStan were to properly merge reflection data across multiple supported PHP versions then this PR wouldn't be needed at all. It's really weird to implement this for a single function when the rest of PHPStan doesn't really support this.

I agree. I think PHPStan should have its own meta-type that automatically resolves to the correct type (i.e. resource/object) according to the current php version.

At the moment i'm happily using this hack: https://github.com/phpstan/phpstan/discussions/4212#discussioncomment-284814

The problem is that it works only within the same project, i.e. as the phpstan.neon file is local to the project it does not "export" easily, so for methods facing externally I have to type it as "mixed".

These are the type aliases i'm currently using:

```
	typeAliases: (php 7.x)
		resource_gd: 'resource'
		resource_gdfont: 'int'
		resource_socket: 'resource'

	typeAliases: (php 8.x)
		resource_gd: '\GdImage'
		resource_gdfont: '\GdFont'
		resource_socket: '\Socket'

```

Also note that some types have been introduced in different versions, so for example GdImage and Socket are 8.0 while GdFont is 8.1

PR #3668, opened 2024-11-25
Fixed a bug where the last value was not recognized when passing an associative array as an argument

sayuprc:
For optional arguments, updated the validation to treat them as non-existent when not provided.

resolve phpstan/phpstan#11815


ondrejmirtes:
Thank you!

PR #3670, opened 2024-11-26
Allow gitlab ci to be detected

BackEndTea:
Gitlab CI wasn't being detected by the `CiDetectedErrorFormatter`. This MR allows detection so that the `CiDetectedErrorFormatter` can be used for gitlab as well.

### DIFF
--- a/src/Command/ErrorFormatter/CiDetectedErrorFormatter.php
+++ b/src/Command/ErrorFormatter/CiDetectedErrorFormatter.php
@@ -17,6 +17,7 @@ class CiDetectedErrorFormatter implements ErrorFormatter
 	public function __construct(
 		private GithubErrorFormatter $githubErrorFormatter,
 		private TeamcityErrorFormatter $teamcityErrorFormatter,
+		private GitlabErrorFormatter $gitlabErrorFormatter,
 	)
 	{
 	}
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
AFAIK GitLab still only shows these results in paid tiers: https://docs.gitlab.com/ee/ci/testing/code_quality.html

That's why GitLab CI is not currently detected by default. Maybe it's part of `GITLAB_FEATURES` env variable, I'm not sure. If you research that and detect correctly then I'd merge that.

alexbrouwer:
Gitlab expects the results in a JSON file which can the be passed as a "codequality" artifact. It will automatically be parsed and displayed. Where depends on the "tier" being used, but even the free version shows it as you can read [here](https://docs.gitlab.com/ci/testing/code_quality/#view-code-quality-results).

Currently I have to run the PHPStan analysis twice. Once the get results in a Gitlab format, once to get it in a human readable form. 

```yaml
phpstan:
  # ...
  script:
    - vendor/bin/phpstan analyse --no-progress --error-format=gitlab --memory-limit=2G --quiet > code_quality_phpstan.json || echo "PHPStan found errors, see below."
    - vendor/bin/phpstan analyse --no-progress --memory-limit=2G
  artifacts:
    reports:
      codequality: code_quality_phpstan.json
```

Automatically detecting Gitlab resolves the duplicate analysis, but would require choosing, or adding config for, the name of the output file.

BackEndTea:
@alexbrouwer i created https://github.com/BackEndTea/phpstan-multiple-error-formatter this package to have multiple error formatters at once for my case. Perhaps it could work for you as well.

PR #3671, opened 2024-11-26
Allow multiple error formatters

BackEndTea:
This MR introduces the `errorFormatter.multiple` (name pending). It allows a user to specify multiple output formats.

In a CI environment, you sometimes want to output to multiple tools, and a table output for the logs.
Currently that is only possible by composing multiple error formatters in your own code. When you have multiple projects with the same set up, you either need a package, or duplicated code.

In your config you could specify the following, which mean you'll use both the table and gitlab output when running PHPStan with the flag `--error-format multiple`.

```yaml
services:
    errorFormatter.multiple:
        arguments:
            formatters:
                - @errorFormatter.table
                - @errorFormatter.gitlab
```

The second commit introduces a new parameter for the gitlab output, to allow redirecting this to a file, instead of the stdout.
This means that we can have both a table output in the logs, and a gitlab output written to a file. 

I have not added this paramter to other formatters like the JSON one, as i'm unsure if this is the way to go.

With the gitlab option one could then the following configuration to output the needed json for gitlab, and still have a table output.

```yaml
services:
    errorFormatter.gitlab:
        arguments:
            fileLocation: %currentWorkingDirectory%/phpstan.json
    errorFormatter.multiple:
        arguments:
            formatters:
                - @errorFormatter.table
                - @errorFormatter.gitlab
```

Another option would be change the error-format flag to allow multiple error formatters. I think this way gives users more freedom as they can specify the `multiple` as their default error formatter in the configuration.

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -2174,3 +2174,9 @@ services:
 		class: PHPStan\Command\ErrorFormatter\TeamcityErrorFormatter
 		arguments:
 			relativePathHelper: @simpleRelativePathHelper
+
+	errorFormatter.multiple:
+		class: PHPStan\Command\ErrorFormatter\ChainErrorFormatter
+		arguments:
+			formatters:
+				- @errorFormatter.table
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Hi, I'm against this, it doesn't make much sense to me. If you configure multiple error formatters including the one for CI, then the developers on their local machine get output that is irrelevant (and unreadable) for them.

PR #3672, opened 2024-11-26
Ease debugging BaselineNeonErrorFormatterIntegrationTest

staabm:
I see this test fail more often on PRs lately with output like:

```
php vendor/bin/paratest --runner WrapperRunner --no-coverage --group exec
ParaTest v6.6.3 upon PHPUnit 9.6.20 by Sebastian Bergmann and contributors.

............................E...                                 32 / 32 (100%)

Time: 00:35.534, Memory: 80.00 MB

There was 1 error:

1) PHPStan\Command\ErrorFormatter\BaselineNeonErrorFormatterIntegrationTest::testRunUnixFileWithWindowsBaseline
PHPStan\ShouldNotHappenException: Could not clear result cache.

D:\a\phpstan-src\phpstan-src\tests\PHPStan\Command\ErrorFormatter\BaselineNeonErrorFormatterIntegrationTest.php:74
D:\a\phpstan-src\phpstan-src\tests\PHPStan\Command\ErrorFormatter\BaselineNeonErrorFormatterIntegrationTest.php:54

FAILURES!
Tests: 32, Assertions: 82, Errors: 1.
make: *** [Makefile:9: tests-integration] Error 2
```

lets see whether we can get a more meaningful output which might help to fix the underlying problem

### DIFF
--- a/tests/PHPStan/Command/ErrorFormatter/BaselineNeonErrorFormatterIntegrationTest.php
+++ b/tests/PHPStan/Command/ErrorFormatter/BaselineNeonErrorFormatterIntegrationTest.php
@@ -69,9 +69,9 @@ class BaselineNeonErrorFormatterIntegrationTest extends TestCase
 			throw new ShouldNotHappenException();
 		}
 		chdir(__DIR__ . '/../../../..');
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I don't think we need the debugging. It's obvious what's going on. paratest runs in multiple threads. All these test methods share the same result cache file. Windows doesn't like multiple processes accessing the same file and sometimes fails.

The solution is to have different `tmpDir` for each of these tests.

PR #3680, opened 2024-11-27
Remove shortArraySyntax definiton on Printer::construct()

samsonasik:
It seems already short array syntax by default


samsonasik:
Ready to merge 👍 , failing test is unrelated.

ondrejmirtes:
Thank you!

PR #3682, opened 2024-11-27
Retain list type when assigning to offset 1 of non-empty-list

herndlm:
Fixes https://github.com/phpstan/phpstan/issues/12131

I don't think we can detect this edge case inside `Type::setOffsetValueType()` because we need info of 2 different accessory types, right? Is `IntersectionType` the best place then I suppose?

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/list-type.php
+++ b/tests/PHPStan/Analyser/nsrt/list-type.php
@@ -106,4 +106,29 @@ class Foo
 		assertType('array<int<0, 1>|int<3, max>, int>', $list);
 	}
 
+	/** @param list<int> $list */
+	public function testSetOffsetExplicitlyWithoutGap(array $list): void
+	{
+		assertType('list<int>', $list);
+		$list[0] = 17;
+		assertType('non-empty-list<int>&hasOffsetValue(0, 17)', $list);
+		$list[1] = 19;
+		assertType('non-empty-list<int>&hasOffsetValue(0, 17)&hasOffsetValue(1, 19)', $list);
+		$list[0] = 21;
+		assertType('non-empty-list<int>&hasOffsetValue(0, 21)&hasOffsetValue(1, 19)', $list);
+
+		$list[2] = 23;
+		assertType('non-empty-array<int<0, max>, int>&hasOffsetValue(0, 21)&hasOffsetValue(1, 19)&hasOffsetValue(2, 23)', $list);
+	}
+
+	/** @param list<int> $list */
+	public function testSetOffsetExplicitlyWithGap(array $list): void
+	{
+		assertType('list<int>', $list);
+		$list[0] = 17;
+		assertType('non-empty-list<int>&hasOffsetValue(0, 17)', $list);
+		$list[2] = 21;
+		assertType('non-empty-array<int<0, max>, int>&hasOffsetValue(0, 17)&hasOffsetValue(2, 21)', $list);
+	}
+
 }
--- a/tests/PHPStan/Rules/Properties/data/bug-12131.php
+++ b/tests/PHPStan/Rules/Properties/data/bug-12131.php
@@ -23,7 +23,9 @@ class Test
 	{
 		$this->array[1] = 1;
 	}
-}
 
-$a = new Test();
-$a->setAtOne();
+	public function setAtTwo(): void
+	{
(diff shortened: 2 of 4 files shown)


tests/PHPStan/Rules/Properties/data/bug-12131.php
```diff
@@ -0,0 +1,29 @@
...
+	public function setAtZero(): void
+	{
+		$this->array[0] = 1;
+	}
+
+	public function setAtOne(): void
+	{
+		$this->array[1] = 1;
+	}
+}
```
staabm: IMO this test should have a 

```
	public function setAtTwo(): void
	{
		$this->array[2] = 2;
	}
```

method and define expectations for it

herndlm: you mean you want to see it (still) keep failing here? is that enough? but I can also add some type assertions and load it in NodeScopeResolverTest I guess. just not sure if it's really needed

herndlm: or I add more specific list-related test cases to the generic test file for this rule: `tests/PHPStan/Rules/Properties/data/properties-assigned-types.php`

staabm: > you mean you want to see it (still) keep failing here? is that enough?

yes.

> but I can also add some type assertions and load it in NodeScopeResolverTest I guess

sounds good to me

staabm: > If I have a class with an array property declared as `list<T>` or `non-empty-list<T>` and if I try to assign a value to an existing element with a non-zero index in that list, e.g. `$array[1] = 1`, then I get an assign.propertyType error:
> `non-empty-array<int<0, max>, int> might not be a list.`

reading this from the error report, I also wonder whether its correct just to add hardcoded support for `1`.

staabm: e.g. https://phpstan.org/r/b40e3747-909e-47bb-96df-2a69df6780fb

herndlm: added more tests :)

> e.g. https://phpstan.org/r/b40e3747-909e-47bb-96df-2a69df6780fb

this is weird and seems to be something else. pretty sure the constantarraytype(builder) is taking care of this list properly.. https://phpstan.org/r/5364810d-df16-49e7-a465-ca63f0033f17 strange

herndlm: ah, I see the `MutatingScope` is asked for the type of `$this->array` with a `OriginalPropertyTypeExpr` and that returns not the constant array but, surprise, the original property type :/ the scope would even know what `$this->array` is here, but I guess there is a good reason for doing it like it is done right now hmm.

UPDATE: only a couple of tests start failing if that behaviour is changed, but those cases become slightly weird. E.g. invalid assignments cause still an error, but then that invalid assignment leads to a prop type change which influences future assignments in the scope and either cause false positives/negatives or just different, sometimes weird types in error messages. it might be possible to improve the current behaviour though, maybe I'll keep looking a bit more..

ondrejmirtes:
Thank you.

tests/PHPStan/Analyser/nsrt/list-type.php
```diff
@@ -106,4 +106,29 @@ public function testUnset(array $list): void
...
+		assertType('list<int>', $list);
+		$list[0] = 17;
+		assertType('non-empty-list<int>&hasOffsetValue(0, 17)', $list);
+		$list[1] = 19;
+		assertType('non-empty-list<int>&hasOffsetValue(0, 17)&hasOffsetValue(1, 19)', $list);
+		$list[0] = 21;
+		assertType('non-empty-list<int>&hasOffsetValue(0, 21)&hasOffsetValue(1, 19)', $list);
+
+		$list[2] = 23;
+		assertType('non-empty-array<int<0, max>, int>&hasOffsetValue(0, 21)&hasOffsetValue(1, 19)&hasOffsetValue(2, 23)', $list);
```
herndlm, line 121: wanted to delete this. because that other method below exists 🤦

ondrejmirtes: There's always the next PR :)

PR #3684, opened 2024-11-27
Nuke `OriginalPropertyTypeExpr`

herndlm:
just want to know what this affects outside of PHPStan's test suite..

I assume it will lead to some improvements in edge cases while causing some other edge case issues, especially with invalid offset setting on array props.


ondrejmirtes:
There aren't many integration tests running right now :)

herndlm:
ok, still interesting. looks like this would fix

- https://github.com/phpstan/phpstan/issues/7880
- https://github.com/phpstan/phpstan/issues/6571
- https://github.com/phpstan/phpstan/issues/6398

downside could be some weird types in errors in some cases, especially phpstan's test suite that tries to assign incorrect types in a row. more integration tests would be interesting I guess :)

herndlm:
enough for today. it would be interesting to see if the type from the scope can be used for that `TypesAssignedToPropertiesRule` rule only, while not modifying the property type itself. that would be the best of both worlds I suppose..

herndlm:
Ok, I think I know what I have to do :)

ondrejmirtes:
Finally figured it out https://github.com/phpstan/phpstan-src/commit/a8cae55bcbab4ddb1529abaa83947cd6f0968ecf

herndlm:
> Finally figured it out [a8cae55](https://github.com/phpstan/phpstan-src/commit/a8cae55bcbab4ddb1529abaa83947cd6f0968ecf)

amazing that you still remembered this PR :D

ondrejmirtes:
There are things that haunt me for years until they're solved 😀

PR #3685, opened 2024-11-27
Remove unnecessary test code

herndlm:
just forgot to clean this up in https://github.com/phpstan/phpstan-src/pull/3682

the next method is dealing explicitly with this


ondrejmirtes:
Thank you!

PR #3688, opened 2024-11-28
Add PHP Configurator

jack-worman:
I believe there is no better way to configure a PHP application than with PHP code.

The advantages that I see are:
1. You don't have to learn a new syntax
2. IDE support built in, since it is only PHP code
3. Documentation can be put directly onto options via doc-blocks


ondrejmirtes:
This isn't a project that you can just implement without prior discussion, and expect it to be merged.

Although one day I'd like to switch away from .neon, I'm not ready to tackle this project right now.

Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it.

jack-worman:
I was hoping the PR would be the place for the discussion/proof-of-concept.
I wasn't expecting it to be merged without any thought.

jack-worman:
Would you be open to discussing this via an issue on https://github.com/phpstan/phpstan/issues?

ondrejmirtes:
Yeah, sure.

PR #3691, opened 2024-11-29
Updated signatures for "XML parser"

sreichel:
Updated all signatures according to https://www.php.net/manual/en/function.xml-error-string.php

(hope i did it right with deltas)


ondrejmirtes:
Hi, can you show on phpstan.org/try which bugs this fixes? I feel like we don't need most of the changes as PHPStan will already report them correctly, mostly thanks to https://github.com/phpstan/php-8-stubs.

sreichel:
Hi,

e.g. https://phpstan.org/r/84e5838e-96ed-4111-b332-1424bd6ad513

Stubs look wrong to me ... (?)

All changes are ...

- resource: PHP8+ - "resource" -> "XMLParser"
- handler: all PHP - string is allowed as handler (next to callable)
- handler: PHP8.4 - null is allowed
-

ondrejmirtes:
As I said, most of these are taken care of thanks to php-8-stubs.

Please submit only the changes from `callable` to `callable|string`, those make sense.

ondrejmirtes:
But the "passing string" is being deprecated on 8.4+, so I'm not sure about it:

> Passing a non-[callable](https://www.php.net/manual/en/language.types.callable.php) [string](https://www.php.net/manual/en/language.types.string.php) to handler is now deprecated, use a proper callable for methods, or [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) to reset the handler.

ondrejmirtes:
So it's better to not allow that either.

sreichel:
Mhhh, we are running php7.4 ... not 8.4.

I have read that it will be deprecated, but phpstan also supports older versions. Not?

So i have to baseline that?

ondrejmirtes:
It supports older versions, but you can also fix the deprecated code even before you upgrade to newer version. You eventually want to upgrade from an unsupported version, right :) Having PHPStan report the error forces you to get ready for a newer PHP version sooner.

ondrejmirtes:
Normally I'm fixing these errors, but after all these years no one complained about `string` not being accepted there so I don't think it makes sense to fix that now.

sreichel:
> but after all these years no one complained about string not being accepted

i think these methods arnt used a lot at all but the issue is valid.

Is there anything wrong with that changes?

sreichel:
Thanks.

Tried to help.

PR #3698, opened 2024-11-30
Use native PHPDocs for `Rule` and `RuleTestCase`

paulbalandan:
I'm using intelephense in VS Code and it cannot understand the `@phpstan-` annotated PHPDocs. However, it understands the native PHPDocs, including the generics related. This PR uses those PHPDocs for `Rule` and `RuleTestCase` for better IDE completion on vscode.

Proof:
image

### DIFF
--- a/src/Rules/Rule.php
+++ b/src/Rules/Rule.php
@@ -20,18 +20,18 @@ use PHPStan\Analyser\Scope;
  * Learn more: https://phpstan.org/developing-extensions/rules
  *
  * @api
- * @phpstan-template TNodeType of Node
+ * @template TNodeType of Node
  */
 interface Rule
 {
 
 	/**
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you.

paulbalandan:
Thanks for merging this. Would you be ok for follow up PRs that (1) removes similar `@phpstan-` annotated PHPDocs and replaces them with the native ones, and (2) removes superfluous PHPDocs due to better IDE completions, especially in the rules?

ondrejmirtes:
We need to acknowledge this isn't fixing anything in PHPStan, but rather addressing a shortcoming of another tool, but I'm okay with this, we don't need any prefixes in PHPDocs here.

PR #3702, opened 2024-12-02
Use current PHP version in CallToFunctionParametersRuleTest

herndlm:
Working on something in `FunctionCallParametersCheck` which is tested via `CallToFunctionParametersRuleTest` but realised that the PHP version was hard-coded to 8.0 which is blocking me there. Also: bigger range of tests over multiple PHP versions, yay :) in general there seem to be many rules where similar things could be done.

Unfortunately this causes a small conflict in 2.0.x apparently.

UPDATE: Markus made me aware of https://github.com/phpstan/phpstan-src/pull/3662 and I adapted the named arguments test to be skipped exactly the same way to avoid troubles there.

### DIFF
--- a/tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
+++ b/tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
@@ -28,7 +28,7 @@ class CallToFunctionParametersRuleTest extends RuleTestCase
 		$broker = $this->createReflectionProvider();
 		return new CallToFunctionParametersRule(
 			$broker,
-			new FunctionCallParametersCheck(new RuleLevelHelper($broker, true, false, true, $this->checkExplicitMixed, $this->checkImplicitMixed, true, false), new NullsafeCheck(), new PhpVersion(80000), new UnresolvableTypeHelper(), new PropertyReflectionFinder(), true, true, true, true, true),
+			new FunctionCallParametersCheck(new RuleLevelHelper($broker, true, false, true, $this->checkExplicitMixed, $this->checkImplicitMixed, true, false), new NullsafeCheck(), new PhpVersion(PHP_VERSION_ID), new UnresolvableTypeHelper(), new PropertyReflectionFinder(), true, true, true, true, true),
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


herndlm:
ah, I just saw that `CallMethodsRuleTest` uses the current PHP version dynamically, **but** for the named arguments test it forces it to 8.0. Which method are we preferring? I see it has a `disallow-named-arguments.php` test, maybe that's the way to go then. Or something like marking the test skipped if the PHP version is too low? 🤔 😅

for the bug fix / feature PR I'm preparing it doesn't matter. I just need to be able to have a dynamic PHP version in this test file.

staabm:
Please note: we are working on the same files

https://github.com/phpstan/phpstan-src/pull/3662

herndlm:
> Please note: we are working on the same files
> 
> #3662

Oh, not good.. The scope dependent php version functionality is 2.x only, right? I'm also OK with the skipping approach of course, the constructor is conflicting any way..

staabm:
> The scope dependent php version functionality is 2.x only, right? 

yes

> I'm also OK with the skipping approach of course, the constructor is conflicting any way..

no its fine. lets see which one is merged earlier :)

staabm:
ahh this looks great now, thanks

herndlm:
> ahh this looks great now, thanks

for a moment I thought I broke gitlab though with all my back and forth .. xD  apparently it wasn't happy with me force-pushing back to an old state hmm. anyway, closer to yours now, yeah

Screenshot From 2024-12-02 20-53-31

herndlm:
btw I have a PR prepared which adapts a rule dealing with an high-upvoted bug. but it would need this small test change first. would be great if this can be merged :)

herndlm:
@ondrejmirtes after this I was planning to fix https://github.com/phpstan/phpstan/issues/11418, or do you consider that to be a new feature, then I can just skip this and work on 2.x after https://github.com/phpstan/phpstan-src/pull/3662 was merged.

I don't want to pressure you, just trying to figure out what the best approach is to avoid conflicts/issues.

ondrejmirtes:
This https://github.com/phpstan/phpstan-src/pull/3662 basically fixes the same problem, so in order to avoid conflicts, I'm closing this. Thanks.

PR #3706, opened 2024-12-04
Implement ArrayColumnRule

staabm:
closes https://github.com/phpstan/phpstan/issues/5101
closes https://github.com/phpstan/phpstan/issues/12188

ArrayColumnRule re-uses the logic from the `array_column` return type extension and reports problems when the logic implies the result will be a never-type.

there are more open problems with `array_column` and I would approach those after this PR got merged

### DIFF
--- a/src/Rules/Functions/ArrayColumnRule.php
+++ b/src/Rules/Functions/ArrayColumnRule.php
@@ -1,0 +1,110 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Functions;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\NeverType;
+use PHPStan\Type\Php\ArrayColumnHelper;
+use PHPStan\Type\VerbosityLevel;
+use function count;
+use function sprintf;
+
+/**
+ * @implements Rule<Node\Expr\FuncCall>
+ */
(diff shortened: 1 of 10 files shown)


staabm:
@herndlm any opinions on the rule?

only the last commit is important.  commits 1-3 are pure refactorings, which could land as a separate previous PR, in case ondrej is fine with the approach in general

btw: ~~I have no idea why~~ the test is failling only with PHP <= 8.1  
-> found the cause: PHPStan is pessimistic and allows ["dynamic properties" in PHP < 8.2](https://github.com/phpstan/phpstan-src/blob/28dfac80bb4a05891ad4da675677c2077098d080/src/Php/PhpVersion.php#L256-L259)

src/Rules/Functions/ArrayColumnRule.php
```diff
@@ -0,0 +1,111 @@
...
+				$indexKeyType = $scope->getNativeType($args[2]->value);
+			}
+		}
+
+		$errors = [];
+		if ($columnKeyType->isNull()->no()) {
+			[$returnValueType, $iterableAtLeastOnce] = $this->arrayColumnHelper->getReturnValueType($arrayType, $columnKeyType, $scope);
+			if ($returnValueType instanceof NeverType || $iterableAtLeastOnce->no()) {
+				$errorBuilder = RuleErrorBuilder::message(sprintf(
+					'Cannot access column %s on %s.',
```
staabm, line 77: I think we will need a better message.. maybe mentioning array offsets or property access, but how to name it in unions of array/objects..? 

maybe we can find a better generic message

ondrejmirtes:
I'm looking forward to this, please rebase this on 2.1.x and mark it as ready so I can review it, thanks.

staabm:
will send separate new PRs

PR #3707, opened 2024-12-04
Remove seemingly unnecessary benevolent union in explode extension

herndlm:
I have literally no clue what this does. also: empty string `**NEVER**`/`false` handling exists already elsewhere up and was added _after_ this.


herndlm:
interesting, so it is making explode return type handling simpler on PHP < 8 with mixed delimiters. weird

there is one test that checks this, any ideas if this is really needed? was added in 2018 with ac64acdc

but anyway, I don't have to force-remove this 😊

PR #3708, opened 2024-12-04
Narrow explode return to constant arrays for small positive limits

herndlm:
Related to https://github.com/phpstan/phpstan/discussions/12187

kept this simple and only handle a single constant limit on purpose for now

Needs https://github.com/phpstan/phpstan-src/pull/3709 first

### DIFF
--- a/src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ExplodeFunctionDynamicReturnTypeExtension.php
@@ -11,7 +11,9 @@ use PHPStan\Type\Accessory\AccessoryLowercaseStringType;
 use PHPStan\Type\Accessory\AccessoryUppercaseStringType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\IntegerRangeType;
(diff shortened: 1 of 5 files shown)


staabm (commented):
to cover the initial request of the discussion, I think it would be useful to assert the result of 
`explode(',', $date, 2) + [1 => null]` in the tests

staabm:
I guess you saw it already yourself, but just in case: seems you found a case in which we miss to narrow a constant arrays optional offsets :-)

https://phpstan.org/r/6286a642-020a-4e53-b713-72994a0bbcad

herndlm:
Yes, was creating a similar example in parallel (https://phpstan.org/r/d6b8098a-282b-47cd-be91-38e42401d1eb) :) let's fix that first..

herndlm:
currently: https://phpstan.org/r/a1181543-1503-4108-bcf4-ee8e13abda66

so most of it are things like
> Expected type array{0: string, 1?: string}, actual: non-empty-list<string>

not sure if this PR even adds much value..
some max limits at least I guess hmm, could still be useful potentially?

herndlm:
oh interesting, the 3 failing tests show that the narrower type can cause new `offsetAccess.notFound` errors to appear:

- https://github.com/phpstan/phpstan-src/actions/runs/13833502116/job/38703310885?pr=3708
- https://github.com/phpstan/phpstan-src/actions/runs/13833502116/job/38703311456?pr=3708
- https://github.com/phpstan/phpstan-src/actions/runs/13833502116/job/38703311898?pr=3708 -> https://github.com/doctrine/collections/blob/2790cb57e6510e4404873bdca6688e447550d720/src/Expr/ClosureExpressionVisitor.php#L44

which I think is expected and fine.

herndlm:
What to do about this one? Is it OK or e.g.  bleeding edge toggle maybe?

ondrejmirtes:
The problem with this one is that it promotes errors that would previously be only reported with `reportPossiblyNonexistentGeneralArrayOffset: true` to be always reported.

Does this have an actual benefit? Isn't `non-empty-list<string>` enough?

herndlm:
I was also not sure. If in doubt, maybe it's better not do this. At least the preparation work for this uncovered other count related improvements 😊

PR #3709, opened 2024-12-04
Improve `count()` narrowing of constant arrays

herndlm:
Generalizes the already existing solution (wow, this is way more complicated than I thought!) and uses it outside of union types as well. I kind of had to make the "isNormalCount" determination a bit more dry, which unfortunately made the diff worse..

I did do a performance comparison with `hyperfine` running `make phpstan` and didn't notice any changes, but my system is not the best.

Closes https://github.com/phpstan/phpstan/issues/12190
Closes https://github.com/phpstan/phpstan/issues/3631

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-4700.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-4700.php
@@ -41,9 +41,9 @@ function(array $array, int $count): void {
 	if (isset($array['e'])) $a[] = $array['e'];
 	if (count($a) > $count) {
 		assertType('int<2, 5>', count($a));
-		assertType('array{0: mixed~null, 1?: mixed~null, 2?: mixed~null, 3?: mixed~null, 4?: mixed~null}', $a);
+		assertType('list{0: mixed~null, 1: mixed~null, 2?: mixed~null, 3?: mixed~null, 4?: mixed~null}', $a);
 	} else {
-		assertType('0', count($a));
-		assertType('array{}', $a);
+		assertType('int<0, 5>', count($a)); // Could be int<0, 1>
+		assertType('array{}|array{0: mixed~null, 1?: mixed~null, 2?: mixed~null, 3?: mixed~null, 4?: mixed~null}', $a); // Could be array{}|array{0: mixed~null}
 	}
 };
--- a/tests/PHPStan/Analyser/nsrt/bug-4700.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-4700.php
@@ -40,7 +40,7 @@ function(array $array, int $count): void {
 	if (isset($array['d'])) $a[] = $array['d'];
 	if (isset($array['e'])) $a[] = $array['e'];
 	if (count($a) > $count) {
-		assertType('int<1, 5>', count($a));
+		assertType('int<2, 5>', count($a));
 		assertType('array{0: mixed~null, 1?: mixed~null, 2?: mixed~null, 3?: mixed~null, 4?: mixed~null}', $a);
 	} else {
 		assertType('0', count($a));
--- a/tests/PHPStan/Analyser/nsrt/count-type.php
+++ b/tests/PHPStan/Analyser/nsrt/count-type.php
@@ -64,6 +64,29 @@ class Foo
 		}
 	}
 
+	/** @param array{0: string, 1?: string} $arr */
+	public function doBar(array $arr): void
+	{
+		if (count($arr) <= 1) {
+			assertType('1', count($arr));
+			return;
+		}
+
+		assertType('2', count($arr));
+		assertType('array{string, string}', $arr);
+	}
+
+	/** @param array{0: string, 1?: string} $arr */
+	public function doBaz(array $arr): void
+	{
+		if (count($arr) > 1) {
+			assertType('2', count($arr));
+			assertType('array{string, string}', $arr);
+		}
+
+		assertType('1|2', count($arr));
+	}
+
 }
 
 /**
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1071,10 +1071,10 @@ final class TypeSpecifier
 			}
 
 			$isSizeSuperTypeOfArraySize = $sizeType->isSuperTypeOf($type->getArraySize());
-			if ($context->truthy() && $isSizeSuperTypeOfArraySize->no()) {
+			if ($isSizeSuperTypeOfArraySize->no()) {
 				return new NeverType();
 			}
-			if ($context->falsey() && !$isSizeSuperTypeOfArraySize->yes()) {
+			if ($context->falsey() && $isSizeSuperTypeOfArraySize->maybe()) {
 				return new NeverType();
 			}
 
@@ -1123,10 +1123,10 @@ final class TypeSpecifier
 					return $valueTypesBuilder->getArray();
 				}
 
-				return $type;
+				return $context->truthy() ? $type : new NeverType();
 			}
 
-			return TypeCombinator::intersect($type, new NonEmptyArrayType());
+			return $context->truthy() ? TypeCombinator::intersect($type, new NonEmptyArrayType()) : new NeverType();
 		});
 
 		return $this->create($countFuncCall->getArgs()[0]->value, $resultType, $context, $scope)->setRootExpr($rootExpr);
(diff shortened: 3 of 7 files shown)


herndlm:
I know, I changed that narrow* method to return Type instead of SpecifiesTypes. Maybe I'll change this back again tomorrow.

herndlm:
> I know, I changed that narrow* method to return Type instead of SpecifiesTypes. Maybe I'll change this back again tomorrow.

adapted. looks better now IMO. The master plan is to move more common things for both `count() == xx` and `count > xx` specification into the now renamed `specifyTypesForCountFuncCall()`. Would also be great if `turnListIntoConstantArray()` goes away completely or is only used internally once... But for here this is more than enough IMO.

@staabm what do you think about this?

tests/PHPStan/Analyser/nsrt/bug-4700.php
```diff
@@ -40,7 +40,7 @@ function(array $array, int $count): void {
 	if (isset($array['d'])) $a[] = $array['d'];
 	if (isset($array['e'])) $a[] = $array['e'];
 	if (count($a) > $count) {
-		assertType('int<1, 5>', count($a));
+		assertType('int<2, 5>', count($a));
 		assertType('array{0: mixed~null, 1?: mixed~null, 2?: mixed~null, 3?: mixed~null, 4?: mixed~null}', $a);
```
staabm: can/should this be more precise (separate PR)?

```suggestion
		assertType('array{0: mixed~null, 1: mixed~null, 2?: mixed~null, 3?: mixed~null, 4?: mixed~null}', $a);
```

staabm:
> what do you think about this?

this PR looks great, thanks for cleaning up after me :)

tests/PHPStan/Analyser/nsrt/bug-4700.php
herndlm: yes, it should. I would prefer separate, because I have no clue where to even look :)

apparently it was like this (same array and `int<2, 5>` before you added the count feature already. interesting that it changed then and changed back now.

staabm: ok cool. I guess we should open a new issue for it to get issue-bot support and have it on the radar

herndlm: I could fix this in another commit. looks like the offset was not considered for the truthy scope. falsey can be improved too, but not really sure how, might need bigger changes.

I mean, if the array count is `int<2, 5>` in truthy, then it should most likely be `int<0, 1>` in falsey? I believe the `turnListIntoConstantArray()` helper might just not be able to deal with this yet. But I'm not sure if I'll be able to further improve this here.

herndlm: could simplify it a bit more and found a better fix. added comments to where the types could be better in falsey context, I think this is as far as I can go here. at least it's not wrong any more now, it could be better / narrower though :)

herndlm:
Should stuff like this be based on 1.12.x or latest dev branch? I sometimes have issues deciding 😅

ondrejmirtes:
Big changes to TypeSpecifier I prefer on 2.1.x, to avoid conflicts.

herndlm:
Thx, makes sense. And good call, there was a conflict.. Rebased this and also other of my PRs on 2.1.x and dealt with conflicts 😊

herndlm:
I think this is ready, only that first issue bot response is confusing me slightly: https://github.com/phpstan/phpstan-src/actions/runs/12635127480. but at the same time it looks like those errors are already there now (https://phpstan.org/r/3cb22712-7994-4b1c-9946-3946083ddb27) or am I missing something?

UPDATE: ah wait, it's line 109 of PHP 7.1 - 7.4 which seems to be loosing some generic info I think. I'll check again then..

herndlm:
ok done, this was even simplified further, is now behaving as I expect it to and should be save :)

ondrejmirtes (changes requested):
The levels test JSON needs updating: https://github.com/phpstan/phpstan-src/actions/runs/13762263415/job/38480859499?pr=3709

Just run it locally, it will autoupdate the JSON file, and commit that.

Also it might be pointing out an actual bug, it might be worth reviewing what changed.

herndlm:
@ondrejmirtes thx for checking this. it looks like though those levels test differences are not coming from here but were recently introduced in https://github.com/phpstan/phpstan-src/commit/deb091148b63515ff2765d18525090fcba7baff3

UPDATE: but let me update those levels tests in a dedicated PR anyway, you can review the changes there.

--> https://github.com/phpstan/phpstan-src/pull/3869

tests/PHPStan/Analyser/nsrt/count-type.php
```diff
@@ -64,6 +64,29 @@ public function doFooBar(
 		}
 	}
 
+	/** @param array{0: string, 1?: string} $arr */
+	public function doBar(array $arr): void
+	{
+		if (count($arr) <= 1) {
+			assertType('1', count($arr));
+			return;
```
VincentLanglet, line 72: What about adding `assertType('array{string}', $arr);` ?

ondrejmirtes:
Thank you!

tests/PHPStan/Analyser/nsrt/count-type.php
herndlm: good point, I already forgot. but quickly looked again and apparently it's currently

```php
assertType('array{0: string, 1?: string}', $arr);
```

it ends up using a falsey scope here, because `count($arr) <= 1` is transformed to `2 < count($arr)` and then something in the loop still skips creating a result type because it can't fully deal with falsey scopes unfortunately. If I remove that `continue`, it would basically come up with "cannot be `array{string, string}`" which should make it then `array{string}` by some falsey scope filtering or so irrelevant of the new code. but that does also not work :/ and many tests for (list) counting start to fail :/ would be great to further improve this, but I was kind of stuck there..

ondrejmirtes:
Hi, I found out about a regression because of this PR: https://phpstan.org/r/38051a80-e8ef-4f26-abaa-dac27791b444

The keys shouldn't be optional :)

herndlm:
It looks like this was already like that before but that case is not supported by the changes here? but I agree, they should be non-optional 🤔

herndlm:
aha, that one is not handled because the `$arrayType->isList()` returns maybe. interesting

ondrejmirtes:
Yeah but even if something is not a list, if the total number of keys is <= than the verified count, we know we can make all of them required.

herndlm:
I'll look into this. @staabm initially created all kind of list count tests (luckily!) and some start to fail, so those cases are dangerous. e.g. smth like where the keys have gaps I suppose.

```php
if (count($row) === 2) {
	assertType('array{0: int, 3?: string|null}', $row); // is array{int, *ERROR*} instead
} else {
	assertType('array{0: int, 3?: string|null}|array{string}', $row);
}
```

or, in general, some new feature for this needs to be added or the existing one changed :) should be doable

ondrejmirtes:
FYI this PR broke this snippet https://phpstan.org/r/8b318eac-e881-46c4-a72d-8ae55f2953e5

There's nothing we can deduce for `$this->importedDaySummaryRows` so the type shouldn't be narrowed to Never.

I'm looking into this myself but I'm not sure how successful I'm going to be, so please look into it too. The problem goes away when I change the type `$this->totalExpectedRows` from int to positive-int so it's going to be something about that.

ondrejmirtes:
This diff fixes the problem but makes some tests fail:

```diff
diff --git a/src/Analyser/TypeSpecifier.php b/src/Analyser/TypeSpecifier.php
index c016b2869..e54de7d5e 100644
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1136,8 +1136,10 @@ final class TypeSpecifier
 				$resultTypes[] = $valueTypesBuilder->getArray();
 				continue;
 			}
+		}
 
-			$resultTypes[] = $arrayType;
+		if (count($resultTypes) === 0) {
+			return null;
 		}
 
 		return $this->create($countFuncCall->getArgs()[0]->value, TypeCombinator::union(...$resultTypes), $context, $scope)->setRootExpr($rootExpr);

```

Please look into this as this is currently blocking a release. Thanks.

staabm:
I think I have a working fix. PR incoming

ondrejmirtes:
I have one more blocking regression caused by this PR: https://phpstan.org/r/1fe15afb-62c0-4c3d-8df5-b240ccb36399

The type of `$otherMinPrices` should not degrade to mixed.

herndlm:
preparing a fix, but currently it "unfixes" https://github.com/phpstan/phpstan/issues/1311 again unfortunately.
UPDATE: looks like it was not properly fixed really anyway, I'll open the PR soon with explanations

PR #3710, opened 2024-12-05
Fix htmlspecialchars and htmlentities being not-empty-string without ENT_SUBSTITUTE flag

jack-worman:
Closes https://github.com/phpstan/phpstan/issues/12021


ondrejmirtes:
Thank you!

PR #3713, opened 2024-12-06
Fix `iterator_to_array` to early return when `$preserveKeys` is false

paulbalandan:
Closes https://github.com/phpstan/phpstan/issues/12196


ondrejmirtes:
Thank you!

PR #3716, opened 2024-12-09
Array map on multiple array is a list

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12223


ondrejmirtes:
Thank you!

PR #3717, opened 2024-12-09
Fix ConstantArrayType::isCallable

VincentLanglet:
Reopened from https://github.com/phpstan/phpstan-src/pull/3637 since I messed up with my branches (and the previous changes was on 1.12.x branch...)

Closes https://github.com/phpstan/phpstan/issues/12063

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -481,7 +481,19 @@ class ConstantArrayType implements Type
 			$typeAndMethods,
 		);
 
-		return TrinaryLogic::createYes()->and(...$results);
+		$isCallable = TrinaryLogic::createYes()->and(...$results);
+		if ($isCallable->yes()) {
+			$callableArray = $this->getClassOrObjectAndMethods();
+			if ($callableArray !== []) {
+				[$classOrObject, $methods] = $callableArray;
+
+				if (count($methods->getConstantStrings()) !== count($typeAndMethods)) {
+					return TrinaryLogic::createMaybe();
+				}
+			}
+		}
+
+		return $isCallable;
 	}
 
 	public function getCallableParametersAcceptors(ClassMemberAccessAnswerer $scope): array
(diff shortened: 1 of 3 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -468,7 +468,19 @@ public function isCallable(): TrinaryLogic
 			$typeAndMethods,
 		);
 
-		return TrinaryLogic::createYes()->and(...$results);
+		$isCallable = TrinaryLogic::createYes()->and(...$results);
+		if ($isCallable->yes()) {
+			$callableArray = $this->getClassOrObjectAndMethods();
```
ondrejmirtes, line 486: This is ineffective. This method was already called once inside `findTypeAndMethodNames` above. It's a bit of a mess right now.

I just made sure this kind of mistake isn't possible: https://github.com/phpstan/phpstan-src/commit/30b9cd86c523b59739b609dc4e325d3275d71188

So the fix should probably be done inside `findTypeAndMethodNames` but I'm not sure.

ondrejmirtes (changes requested):

src/Type/Constant/ConstantArrayType.php
VincentLanglet: > So the fix should probably be done inside findTypeAndMethodNames but I'm not sure.

Not sure too. Depends of the purpose of the method.

Currently the `findTypeAndMethodNames` returns all the possible callable with at least a Maybe certainty. Cf https://github.com/phpstan/phpstan-src/pull/3637#issuecomment-2485308668

Currently it's not possible to create
```
ConstantArrayTypeAndMethod::createConcrete($type, $method->getValue(), TrinaryLogic::createNo());
```

So if you have a string 'foo'|'bar' and only 'foo' is a callable name, if it will returns 'foo'.
But it doesn't mean that 'foo'|'bar' is always a callable.

That's why I added the logic in isCallable to check the count of possible callable didn't drop (which mean some value has No as certainty).

PR #3718, opened 2024-12-10
Add-fpow

jack-worman:
Docs: https://www.php.net/manual/en/function.fpow.php

### DIFF
--- a/resources/functionMap_php84delta.php
+++ b/resources/functionMap_php84delta.php
@@ -19,6 +19,7 @@ return [
 		'http_clear_last_response_headers' => ['void'],
 		'mb_lcfirst' => ['string', 'string'=>'string', 'encoding='=>'string'],
 		'mb_ucfirst' => ['string', 'string'=>'string', 'encoding='=>'string'],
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
This doesn't solve the problem. PHPStan already knows about the function. See https://github.com/phpstan/phpstan/issues/12226#issuecomment-2530719733

jack-worman:
> This doesn't solve the problem. PHPStan already knows about the function. See [phpstan/phpstan#12226 (comment)](https://github.com/phpstan/phpstan/issues/12226#issuecomment-2530719733)

Although it may not solve the problem, shouldn't `fpow()` still be added? Why is `mb_ucfirst()` added to the new list?

ondrejmirtes:
Not sure, we don't need it there.

jack-worman:
> Not sure, we don't need it there.

To help my understanding, could you give me an example of a reason to ever add a function to this file?

ondrejmirtes:
1) If it exists on PHP 7 (php-8-stubs are used for PHP 8+)
2) Or if we want to type more specific parameter or return types

PR #3719, opened 2024-12-10
Respect asserts and throws on pure functions that return void

jack-worman:
Fixes some of the problems from https://github.com/phpstan/phpstan/issues/12224

This should now be fixed: https://phpstan.org/r/852dd5b6-e76c-41ed-a0fb-7b1a46b0900f

### DIFF
--- a/tests/PHPStan/Rules/Pure/PureFunctionRuleTest.php
+++ b/tests/PHPStan/Rules/Pure/PureFunctionRuleTest.php
@@ -167,4 +167,14 @@ class PureFunctionRuleTest extends RuleTestCase
 		]);
 	}
 
+	public function testBug12224(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-12224.php'], [
+			[
+				'Function PHPStan\Rules\Pure\data\pureWithThrowsVoid() is marked as pure but returns void.',
+				18,
+			],
+		]);
+	}
+
 }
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Pure/PureFunctionRuleTest.php
```diff
@@ -167,4 +167,11 @@ public function testBug11361(): void
 		]);
 	}
 
+	public function testBug12224(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-12224.php'], [
+			['Function PHPStan\Rules\Pure\data\pureWithThrowsVoid() is marked as pure but returns void.', 18],
```
ondrejmirtes:
```suggestion
			[
				'Function PHPStan\Rules\Pure\data\pureWithThrowsVoid() is marked as pure but returns void.',
				18,
			],
```

tests/PHPStan/Rules/Pure/PureMethodRuleTest.php
```diff
@@ -194,10 +194,12 @@ public function dataBug11207(): array
 		];
 	}
 
-	public function testBug12048(): void
+	public function testBug12224(): void
 	{
 		$this->treatPhpDocTypesAsCertain = true;
-		$this->analyse([__DIR__ . '/data/bug-12048.php'], []);
+		$this->analyse([__DIR__ . '/data/bug-12224.php'], [
```
ondrejmirtes, line 218: I don't get it, why did you delete this test? You should only add a new test.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Pure/PureMethodRuleTest.php
jack-worman: not sure why i did that. Reverted the delete

ondrejmirtes:
Thank you!

PR #3720, opened 2024-12-10
Use native PHPDocs wherever possible

paulbalandan:
Follow up PR to #3698 

Replaced the phpdocs only in `src/` and `stubs/`.


ondrejmirtes:
Thank you.

PR #3721, opened 2024-12-11
RegexArrayShapeMatcher: fix regex wildcard omitted from type

staabm:
closes https://github.com/phpstan/phpstan/issues/12211


ondrejmirtes:
Thank you.

PR #3722, opened 2024-12-11
RegexArrayShapeMatcher: fix regex alternatives in capture group are concatenated

staabm:
closes https://github.com/phpstan/phpstan/issues/12210
closes https://github.com/phpstan/phpstan/issues/12173


ondrejmirtes:
Thank you.

PR #3723, opened 2024-12-11
RegexArrayShapeMatcher: Don't narrow 'J' modifier

staabm:
I think the 'J' modifier - which I have never seen in code until today - is something we don't need to support.
adding support is not impossible, but I don't want to overcomplicate the regex analysis for such an edge case.

this PR just makes sure we don't narrow it in a wrong way.

closes https://github.com/phpstan/phpstan/issues/12126


ondrejmirtes:
Thank you.

PR #3724, opened 2024-12-11
Enhance offset access by defining various access modes, and restrict which ones are available for DOM classes

Girgias:
This is very W.I.P. and I think this is the correct way to fix it, but I don't know if I should update `Type::isOffsetAccessible()` to pass in the `$mode` parameter or not.

This would fix https://github.com/phpstan/phpstan/issues/12235 and do preliminary work to support https://wiki.php.net/rfc/container-offset-behaviour if/when it lands.


VincentLanglet:
https://github.com/phpstan/phpstan/issues/12235 is now solved

PR #3725, opened 2024-12-11
Add support for internal classes that overload offset access

Girgias:
Quick fix for https://github.com/phpstan/phpstan/issues/12235 as `DOMNodeList` already doesn't complain about incorrect writes it makes sense to have the same behaviour for the other internal classes until PHPStan uses the access mode of the offset.

### DIFF
--- a/tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
+++ b/tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
@@ -898,4 +898,14 @@ class NonexistentOffsetInArrayDimFetchRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/bug-2634.php'], []);
 	}
 
+	public function testInternalClassesWithOverloadedOffsetAccess(): void
+	{
+		$this->analyse([__DIR__ . '/data/internal-classes-overload-offset-access.php'], []);
+	}
+
+	public function testInternalClassesWithOverloadedOffsetAccessInvalid(): void
+	{
+		$this->analyse([__DIR__ . '/data/internal-classes-overload-offset-access-invalid.php'], []);
+	}
+
 }
--- a/tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
+++ b/tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
@@ -905,8 +905,8 @@ class NonexistentOffsetInArrayDimFetchRuleTest extends RuleTestCase
 
 	public function testInternalClassesWithOverloadedOffsetAccess84(): void
 	{
-		if (version_compare(PHP_VERSION, '8.4.0', '<')) {
-			$this->markTestSkipped('Test requires PHP version 8.4.');
+		if (PHP_VERSION_ID < 80400) {
+			$this->markTestSkipped('Test requires PHP 8.1.');
 		}
 		$this->analyse([__DIR__ . '/data/internal-classes-overload-offset-access-php84.php'], []);
 	}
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
Please add a regression test to NonexistentOffsetInArrayDimFetchRuleTest, thanks.

Girgias:
Tests added, and rebased to `1.12.x`.

One set of tests are case which should fail, but currently cannot.

tests/PHPStan/Rules/Arrays/data/internal-classes-overload-offset-access-php84.php
```diff
@@ -0,0 +1,74 @@
+<?php // lint >= 8.4
```
Girgias: This doesn't seem to be working? Not sure why however.

ondrejmirtes: This only works in `/nsrt/`. But here you're in a rule test. Which means you need to call `$this->markTestSkipped()` in the test based on PHP_VERSION_ID. There are many existing tests that do this.

thg2k: > This only works in `/nsrt/`. But here you're in a rule test. Which means you need to call `$this->markTestSkipped()` in the test based on PHP_VERSION_ID. There are many existing tests that do this.

Remember I tried to introduce a new annotation syntax that would allow us to use the same skip system for all data files? (main gotcha was data files that required ranges and additional conditions for skipping). The current system is cumbersome and unintuitive. Would you like to revisit my idea? I can provide a sample PR on which to discuss it.

ondrejmirtes: The current system works very well in my opinion :)

One day we might be able to test the rules in a similar way we test inferred types (by discovering and filtering files in a directory) but today is not that day.

Girgias:
@ondrejmirtes is there anything else I need to do? :)

tests/PHPStan/Rules/Arrays/data/internal-classes-overload-offset-access-invalid-php84.php
```diff
@@ -0,0 +1,91 @@
+<?php
+/**
+ * All of these offset accesses are invalid
+ * ++ and -- are also disallowed in general are they operate "by ref"
```
staabm, line 4:
```suggestion
 * ++ and -- are also disallowed in general as they operate "by ref"
```

tests/PHPStan/Rules/Arrays/data/internal-classes-overload-offset-access-invalid.php
```diff
@@ -0,0 +1,52 @@
+<?php
+/**
+ * All of these offset accesses are invalid
+ * ++ and -- are also disallowed in general are they operate "by ref"
```
staabm, line 4:
```suggestion
 * ++ and -- are also disallowed in general as they operate "by ref"
```

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -898,4 +898,30 @@ public function testBug2634(): void
...
 
+	public function testInternalClassesWithOverloadedOffsetAccess(): void
+	{
+		$this->analyse([__DIR__ . '/data/internal-classes-overload-offset-access.php'], []);
+	}
+
+	public function testInternalClassesWithOverloadedOffsetAccess84(): void
+	{
+		if (PHP_VERSION_ID < 80400) {
+			$this->markTestSkipped('Test requires PHP 8.1.');
```
ondrejmirtes:
```suggestion
			$this->markTestSkipped('Test requires PHP 8.4.');
```

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -898,4 +898,30 @@ public function testBug2634(): void
...
+
+	public function testInternalClassesWithOverloadedOffsetAccessInvalid(): void
+	{
+		$this->analyse([__DIR__ . '/data/internal-classes-overload-offset-access-invalid.php'], []);
+	}
+
+	public function testInternalClassesWithOverloadedOffsetAccessInvalid84(): void
+	{
+		if (PHP_VERSION_ID < 80400) {
+			$this->markTestSkipped('Test requires PHP 8.1.');
```
ondrejmirtes:
```suggestion
			$this->markTestSkipped('Test requires PHP 8.4.');
```

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #3726, opened 2024-12-12
Update functionMap.php to fix return types and parameters

vindic:
Fix for https://github.com/phpstan/phpstan/issues/12148#issue-2704647597 

Fixing the signatures for the following functions :

apcu_cache_info()
apcu_sma_info()

And another one

strval()

setting the parameter value to mixed as defined https://www.php.net/manual/en/function.strval.php 

Code snippet that reproduces the problem :
https://phpstan.org/r/f3e4b8de-f919-44d2-8b08-b979c5240693


ondrejmirtes:
Don't change the strval signature, it's actually correct. To find out why, search the issue tracker.

resources/functionMap.php
```diff
@@ -12088,7 +12088,7 @@
 'strtoupper' => ['uppercase-string', 'str'=>'string'],
 'strtr' => ['string', 'str'=>'string', 'from'=>'string', 'to'=>'string'],
 'strtr\'1' => ['string', 'str'=>'string', 'replace_pairs'=>'array'],
-'strval' => ['string', 'var'=>'__stringAndStringable|int|float|bool|resource|null'],
+'strval' => ['string', 'var'=>'mixed'],
```
staabm: I think it was correct before, see e.g. https://github.com/phpstan/phpstan/issues/9295#issuecomment-1542186125

vindic: Ok, so the PHP documentation is incorrect, next time I will check the issue tracker first.

I have reverted and commit it again.

ondrejmirtes:
Thank you.

PR #3728, opened 2024-12-12
Support inferring new templated type from initial null coalesce assignment

netpok:
Fixes https://github.com/phpstan/phpstan/issues/12250

I made a bug report https://github.com/phpstan/phpstan/issues/12250, and this PR is fixing that.

While all the current tests are passing this PR adds a conditional assignment to NewAssignedToPropertyVisitor which only contained non conditional ones so I'm marking it as draft.


netpok:
This was fixed in https://github.com/phpstan/phpstan-src/pull/4996

PR #3730, opened 2024-12-14
Added `false` as possible return type for `shell_exec`

sreichel:
Specs: https://www.php.net/manual/en/function.shell-exec.php

Test: https://phpstan.org/r/43960222-d368-4ea6-9c6e-e4e9baa6ef9d

"Conflicting" rector change:  https://getrector.com/demo/bc16a638-5551-4dd4-8c7e-55680a3c7052


ondrejmirtes:
Thank you.

PR #3732, opened 2024-12-15
Remove incorrect doc leftover from 1.x

AJenbo:
This appears to have come from a now removed function, or at least it doesn't really make sens to say to use accepts() instead of accepts() :)

### DIFF
--- a/src/Type/Type.php
+++ b/src/Type/Type.php
@@ -64,13 +64,6 @@ interface Type
 	/** @return list<ConstantStringType> */
 	public function getConstantStrings(): array;
 
-	/**
-	 * This is like accepts() but gives reasons
-	 * why the type was not/might not be accepted in some non-intuitive scenarios.
-	 *
-	 * In PHPStan 2.0 this method will be removed and the return type of accepts()
-	 * will change to AcceptsResult.
-	 */
 	public function accepts(Type $type, bool $strictTypes): AcceptsResult;
 
 	public function isSuperTypeOf(Type $type): IsSuperTypeOfResult;


ondrejmirtes:
Yeah, sure, thanks :)

AJenbo:
On a side note I'm trying to update Bladestan to PHPStan 2.0, but I can't figure out how to not make use of `instanceof ObjectType`, 

```php
if ($objectType instanceof ObjectType) {
    if ($objectType->isInstanceOf(Component::class)->yes()) return true;
    if ($objectType->isInstanceOf(Mailable::class)->yes()) return true;
    if ($objectType->isInstanceOf(MailMessage::class)->yes()) return true;
}
```

If I use `isSuperTypeOf()` it returns maybe that that just isn't as confidence raising.

If I instead create a new instance of `ObjectType` using the class name from `$objectType` calling `isInstanceOf` then also goes from yes to maybe :(

P.s.s. PHPStan 2.0 has solved the performance issues we where having with Bladestan, so I'm very excited to get this out for everyone to use :)

ondrejmirtes:
The proper way of asking about this is:

```php
$component = new ObjectType(Component::class);
if ($component->isSuperTypeOf($objectType)->yes()) return true;
```

More about that here: https://phpstan.org/developing-extensions/type-system#querying-a-specific-type

AJenbo:
I also tried that, it also gives me a `maybe` response :(

Here is the class being analyzed: https://github.com/TomasVotruba/bladestan/blob/main/tests/Rules/Fixture/laravel-mail_message-method.php

and the code that I'm trying to update: https://github.com/TomasVotruba/bladestan/blob/8baf741ca5cf067c0fa5d1e94498ab7e8c5e9860/src/NodeAnalyzer/BladeViewMethodsMatcher.php#L124

ondrejmirtes:
There's a big difference between:

```
$component->isSuperTypeOf($objectType)
```

And

```
$objectType->isSuperTypeOf($component)
```

If you do the first approach and the class in $objectType extends/implements Component then the answer is yes() being true.

herndlm:
You might need to create a union of those 3 classes and do the super type check with that I assume.

Because, otherwise, PHPStan is correct, if you're only asking for Component, it could still be something else, which is the reason you end up with maybe.

ondrejmirtes:
If you're not sure we can schedule a call and figure it out together.

AJenbo:
> There's a big difference between...

I did try `$component->isSuperTypeOf($objectType)` as well as `$objectType->isSuperTypeOf($component)`, neither worked and I am completely out of ideas.

> You might need to create a union of those 3 classes and do the super type check with that I assume.

I don't think that's the issue here, but that could be an idea for simplifying the code.

> If you're not sure we can schedule a call and figure it out together.

I would appreciate that a lot :)

In case someone want's to give it a go the offending code is covered by a unit test so you can test it by simply running phpunit for the project. PHPStan 2.0 support is found in the [phpstan2](https://github.com/bladestan/bladestan/pull/119) branch, it's fully working ... PHPStan is just less then happy with how things are implemented in a few places.

ondrejmirtes:
I just checked out the project. Some tests are failing for me locally, all of it is about `assert($ruleError instanceof IdentifierRuleError)`.

You're doing:

```php
            $ruleError = RuleErrorBuilder::message($error->getMessage())
                ->file($phpFilePath)
                ->line($phpFileLine)
                ->metadata([
                    'template_file_path' => array_key_first($fileNameAndTemplateLine),
                    'template_line' => current($fileNameAndTemplateLine),
                ])
                ->build();
            assert($ruleError instanceof IdentifierRuleError);
```

But that doesn't make sense. You're supposed to call `->identiifer()` method on RuleErrorBuilder and set an identifier for the reported error.

Here's more about that https://phpstan.org/blog/using-rule-error-builder

ondrejmirtes:
You can get that from `$error->getIdentifier()`.

ondrejmirtes:
I'm working on the repo right now and I will send a PR once the code looks like it should and tests pass.

ondrejmirtes:
Here you go: https://github.com/bladestan/bladestan/pull/120

The main reason why some methods were returning `maybe` instead of `yes` was that those classes were unknown to PHPStan: https://github.com/bladestan/bladestan/pull/120/commits/975b9b52c1db0bed4a223e44f7d03c031524eece

AJenbo:
Arh that makes a lot of sens I was feeling like it somehow just didn't recognize the classes when trying to rebuild the ObjectsType but only when it directly analyzed the code. Thank you so much for helping with this, I feel a lot better about releasing the next version now :D

PR #3734, opened 2024-12-16
10% faster FunctionCallParametersCheck

staabm:
2.0.x before this PR
```
time php bin/phpstan --debug  
79.37s user 0.92s system 99% cpu 1:20.37 total
```

2.0.x after this PR
```
time php bin/phpstan --debug  
76.22s user 0.84s system 99% cpu 1:17.13 total
```

by reducing `getType` calls:

grafik


ondrejmirtes:
Thank you!

PR #3735, opened 2024-12-16
Support # comments in regex

staabm:
closes https://github.com/phpstan/phpstan/issues/12242

see https://www.regular-expressions.info/freespacing.html#freecomment

will look into whitespace normalization for the `x`-modifer in a separate PR

### DIFF
--- a/src/Type/Regex/RegexGroupParser.php
+++ b/src/Type/Regex/RegexGroupParser.php
@@ -23,6 +23,7 @@ use PHPStan\Type\TypeCombinator;
 use function count;
 use function in_array;
 use function is_int;
+use function preg_replace;
 use function rtrim;
 use function sscanf;
 use function str_contains;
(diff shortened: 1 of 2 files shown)


src/Type/Regex/RegexGroupParser.php
```diff
@@ -64,20 +65,28 @@ public function parseGroups(string $regex): ?array
...
-
 		$modifiers = $this->regexExpressionHelper->getPatternModifiers($regex) ?? '';
 		foreach (self::NOT_SUPPORTED_MODIFIERS as $notSupportedModifier) {
 			if (str_contains($modifiers, $notSupportedModifier)) {
 				return null;
 			}
 		}
 
+		// The regex engine ignores everything after the (?# until the first closing parenthesis
+		$regex = preg_replace('/\(\?#[^)]*\)/', '', $regex) ?? '';
```
staabm: general comments which work independent from modifiers

ondrejmirtes:
Thank you!

src/Type/Regex/RegexGroupParser.php
```diff
@@ -64,20 +65,25 @@ public function parseGroups(string $regex): ?array
...
 		$modifiers = $this->regexExpressionHelper->getPatternModifiers($regex) ?? '';
 		foreach (self::NOT_SUPPORTED_MODIFIERS as $notSupportedModifier) {
 			if (str_contains($modifiers, $notSupportedModifier)) {
 				return null;
 			}
 		}
 
+		if (str_contains($modifiers, 'x')) {
+			// in freespacing mode the # character starts a comment and runs until the end of the line
+			$regex = preg_replace('/[^?]#.*/', '', $regex) ?? '';
```
Seldaek, line 77: This will also get rid of the character before # due to `[^?]`. I realize in most sane cases this will be a space and thus it's harmless, but IMO this should rather be a lookbehind like `(?<!\?)`. I assume the goal was to exclude `(?#...)` but those are already ignored/handled by the lexer https://github.com/phpstan/phpstan-src/blob/2.0.x/resources/RegexGrammar.pp#L80-L82

staabm: will look into it, thanks for the heads up.

(I had to use this additional char before the `#` as we would otherwise destroy comments in `(?# ..)` notation). since we replace before lexing we need to make sure we don't turn the comments into something which is no longer a comment

staabm: fix in https://github.com/phpstan/phpstan-src/pull/3739

PR #3737, opened 2024-12-17
Fix `DOMDocument::create*()` return types

staabm:
closes https://github.com/phpstan/phpstan/issues/12278

see https://www.php.net/manual/en/class.domdocument.php

grafik


ondrejmirtes:
I'd rather do benevolent unions here, because the failures are going to be rare.

ondrejmirtes:
Thank you.

dimadeush:
@ondrejmirtes  Sometimes I can see that doc from PHPStorm is not the same as documentation on the php web site. F.e.

```
/**
     * Adds new child at the end of the children
     * @link https://php.net/manual/en/domnode.appendchild.php
     * @param DOMNode $node <p>
     * The appended child.
     * </p>
     * @return DOMNode The node added.
     */
    public function appendChild(DOMNode $node) {}
```

But on the php web-site we can see next:

> public DOMNode::appendChild([DOMNode](https://www.php.net/manual/en/class.domnode.php) $node): [DOMNode](https://www.php.net/manual/en/class.domnode.php)|[false](https://www.php.net/manual/en/language.types.value.php)

Currently I don't know where the "proper" way and maybe we need to communicate JetBrains as of PHPStorm jar plugin related. Will be done with link to this example. Also I don't know if other php functionality can have wrong description.

dimadeush:
Created issue for JetBrains as several products can be related - https://youtrack.jetbrains.com/issue/WI-80068/Wrong-php-documentation-inside-php-jar-plugin

PR #3738, opened 2024-12-17
Readonly classes cannot be combined with `#[AllowDynamicProperties]`

staabm:
closes https://github.com/phpstan/phpstan/issues/12281

### DIFF
--- a/src/Rules/Classes/ClassAttributesRule.php
+++ b/src/Rules/Classes/ClassAttributesRule.php
@@ -8,6 +8,9 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Node\InClassNode;
 use PHPStan\Rules\AttributesCheck;
 use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function count;
+use function sprintf;
 
 /**
  * @implements Rule<InClassNode>
@@ -28,12 +31,39 @@ final class ClassAttributesRule implements Rule
 	{
 		$classLikeNode = $node->getOriginalNode();
 
-		return $this->attributesCheck->check(
+		$errors = $this->attributesCheck->check(
 			$scope,
 			$classLikeNode->attrGroups,
 			Attribute::TARGET_CLASS,
 			'class',
 		);
+
+		$classReflection = $node->getClassReflection();
+		if (
+			$classReflection->isReadOnly()
+			|| $classReflection->isEnum()
+			|| $classReflection->isInterface()
+		) {
+			$typeName = 'readonly class';
+			$identifier = 'class.allowDynamicPropertiesReadonly';
+			if ($classReflection->isEnum()) {
+				$typeName = 'enum';
+				$identifier = 'enum.allowDynamicProperties';
+			}
+			if ($classReflection->isInterface()) {
+				$typeName = 'interface';
+				$identifier = 'interface.allowDynamicProperties';
+			}
+
+			if (count($classReflection->getNativeReflection()->getAttributes('AllowDynamicProperties')) > 0) {
+				$errors[] = RuleErrorBuilder::message(sprintf('Attribute class AllowDynamicProperties cannot be used with %s.', $typeName))
+					->identifier($identifier)
+					->nonIgnorable()
+					->build();
+			}
+		}
+
+		return $errors;
 	}
 
 }
--- a/src/Rules/Traits/TraitAttributesRule.php
+++ b/src/Rules/Traits/TraitAttributesRule.php
@@ -50,7 +50,7 @@ final class TraitAttributesRule implements Rule
 		$classReflection = $this->reflectionProvider->getClass($traitName->toString());
 		if (count($classReflection->getNativeReflection()->getAttributes('AllowDynamicProperties')) > 0) {
 			$errors[] = RuleErrorBuilder::message('Attribute class AllowDynamicProperties cannot be used with trait.')
-				->identifier('class.allowDynamicPropertiesTrait')
+				->identifier('trait.allowDynamicProperties')
 				->nonIgnorable()
 				->build();
 		}
--- a/src/Rules/Traits/TraitAttributesRule.php
+++ b/src/Rules/Traits/TraitAttributesRule.php
@@ -1,0 +1,61 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Traits;
+
+use Attribute;
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\AttributesCheck;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function count;
+
+/**
+ * @implements Rule<Node\Stmt\Trait_>
+ */
+final class TraitAttributesRule implements Rule
+{
+
+	public function __construct(
+		private AttributesCheck $attributesCheck,
+		private ReflectionProvider $reflectionProvider,
+	)
+	{
+	}
+
+	public function getNodeType(): string
+	{
+		return Node\Stmt\Trait_::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$traitName = $node->namespacedName;
+		if ($traitName === null) {
+			return [];
+		}
+
+		if (!$this->reflectionProvider->hasClass($traitName->toString())) {
+			return [];
+		}
+
+		$errors = $this->attributesCheck->check(
+			$scope,
+			$node->attrGroups,
+			Attribute::TARGET_CLASS,
+			'class',
+		);
+
+		$classReflection = $this->reflectionProvider->getClass($traitName->toString());
+		if (count($classReflection->getNativeReflection()->getAttributes('AllowDynamicProperties')) > 0) {
+			$errors[] = RuleErrorBuilder::message('Attribute class AllowDynamicProperties cannot be used with trait.')
+				->identifier('class.allowDynamicPropertiesTrait')
+				->nonIgnorable()
+				->build();
+		}
+
+		return $errors;
+	}
+
+}
--- a/src/Rules/Traits/TraitAttributesRule.php
+++ b/src/Rules/Traits/TraitAttributesRule.php
@@ -4,6 +4,7 @@ namespace PHPStan\Rules\Traits;
 
 use Attribute;
 use PhpParser\Node;
+use PhpParser\Node\Stmt\Class_;
 use PHPStan\Analyser\MutatingScope;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\ReflectionProvider;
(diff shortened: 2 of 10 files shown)


tests/PHPStan/Rules/Classes/data/bug-12281.php
```diff
@@ -0,0 +1,10 @@
+<?php // lint >= 8.2
+
+namespace Bug12281;
+
+#[\AllowDynamicProperties]
+readonly class BlogData { /* … */ }
+
+/** @readonly  */
+#[\AllowDynamicProperties]
+class BlogDataPhpdoc { /* … */ }
```
staabm, line 10: not sure we want the same error for phpdoc readonly, as we have for native readonly...?

looking thru the source we usually only do the "readonly-by-phpdoc" dance for properties, not for classes

src/Rules/Classes/ClassAttributesRule.php
```diff
@@ -28,12 +31,23 @@ public function processNode(Node $node, Scope $scope): array
...
 			$scope,
 			$classLikeNode->attrGroups,
 			Attribute::TARGET_CLASS,
 			'class',
 		);
+
+		$classReflection = $scope->getClassReflection();
+		if ($classReflection !== null && $classReflection->isReadOnly()) {
+			if (count($classReflection->getNativeReflection()->getAttributes('AllowDynamicProperties')) > 0) {
+				$errors[] = RuleErrorBuilder::message(sprintf('Cannot apply #[AllowDynamicProperties] to readonly class %s.', $classReflection->getName()))
```
ondrejmirtes: We don't use this error message format anywhere. I'd follow some error from AttributesCheck. Example:

> Attribute class AllowDynamicProperties cannot be used with readonly classes.

src/Rules/Classes/ClassAttributesRule.php
```diff
@@ -28,12 +31,23 @@ public function processNode(Node $node, Scope $scope): array
...
+					->identifier('class.allowDynamicPropertiesReadonly')
+					->build();
```
ondrejmirtes, line 62: Should be non-ignorable

tests/PHPStan/Rules/Classes/data/bug-12281.php
ondrejmirtes: I wouldn't do that for PHPDoc `@readonly`.

ondrejmirtes (changes requested):
We also need to error when this attribute is used above interfaces. Also please check what happens above enums and traits (I'm not sure).

staabm:
> We also need to error when this attribute is used above interfaces. Also please check what happens above enums and traits (I'm not sure).

I have tested this cases and linked the 3v4l.org examples in https://github.com/phpstan/phpstan/issues/12281#issue-2744490661

TL;DR:
- enums not allowed
- interfaces not allowed
- traits not allowed

what do you think about adding a separate `AllowDynamicPropertiesAttributeRule` which targets `ClassLike` instead of adding this error to the existing rule?
(atm we don't have a rule which covers attributes on traits)

src/Rules/Classes/ClassAttributesRule.php
```diff
@@ -28,12 +31,34 @@ public function processNode(Node $node, Scope $scope): array
...
 
-		return $this->attributesCheck->check(
+		$errors = $this->attributesCheck->check(
 			$scope,
 			$classLikeNode->attrGroups,
 			Attribute::TARGET_CLASS,
 			'class',
 		);
+
+		$classReflection = $scope->getClassReflection();
```
ondrejmirtes: You can ask `$node` for the class reflection, it's not nullable on that.

src/Rules/Classes/ClassAttributesRule.php
```diff
@@ -28,12 +31,34 @@ public function processNode(Node $node, Scope $scope): array
...
+		if (
+			$classReflection !== null
+			&& ($classReflection->isReadOnly() || $classReflection->isEnum() || $classReflection->isInterface())
+		) {
+			$message = 'Attribute class AllowDynamicProperties cannot be used with readonly classes.';
+			if ($classReflection->isEnum()) {
+				$message = 'Attribute class AllowDynamicProperties cannot be used with enums.';
+			}
+			if ($classReflection->isInterface()) {
+				$message = 'Attribute class AllowDynamicProperties cannot be used with interface.';
```
ondrejmirtes: This could by DRYed and cleaned up with sprintf.

src/Rules/Classes/ClassAttributesRule.php
```diff
@@ -28,12 +31,34 @@ public function processNode(Node $node, Scope $scope): array
...
+			}
+
+			if (count($classReflection->getNativeReflection()->getAttributes('AllowDynamicProperties')) > 0) {
+				$errors[] = RuleErrorBuilder::message($message)
+					->identifier('class.allowDynamicPropertiesReadonly')
```
ondrejmirtes: Needs better identifier for enums and interfaces.

ondrejmirtes (changes requested):
Also:

1) Please do new TraitAttributesRule similar to ClassAttributesRule that will call AttributesCheck. We need the rule to hook onto Trait_ and fetch the reflection ourselves from ReflectionProvider.
2) Add AllowDynamicProperties check there too.

tests/PHPStan/Rules/Classes/data/bug-12281.php
```diff
@@ -0,0 +1,19 @@
...
+
+#[\AllowDynamicProperties]
+enum BlogDataEnum { /* … */ }
+
+#[\AllowDynamicProperties]
+interface BlogDataInterface { /* … */ }
+
+#[\AllowDynamicProperties]
+trait BlogDataTrait { /* … */ }
```
ondrejmirtes: This test is irrelevant for this rule so move it elsewhere.

src/Rules/Classes/ClassAttributesRule.php
```diff
@@ -28,12 +31,39 @@ public function processNode(Node $node, Scope $scope): array
...
+		if (
+			$classReflection->isReadOnly()
+			|| $classReflection->isEnum()
+			|| $classReflection->isInterface()
+		) {
+			$typeName = 'readonly class';
+			$identifier = 'class.allowDynamicPropertiesReadonly';
+			if ($classReflection->isEnum()) {
+				$typeName = 'enum';
+				$identifier = 'class.allowDynamicPropertiesEnum';
```
ondrejmirtes: enum.allowDynamicProperties

src/Rules/Classes/ClassAttributesRule.php
```diff
@@ -28,12 +31,39 @@ public function processNode(Node $node, Scope $scope): array
...
+			}
+			if ($classReflection->isInterface()) {
+				$typeName = 'interface';
+				$identifier = 'class.allowDynamicPropertiesInterface';
```
ondrejmirtes: interface.allowDynamicProperties

src/Rules/Traits/TraitAttributesRule.php
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		if (!$this->reflectionProvider->hasClass($traitName->toString())) {
+			return [];
+		}
+
+		$errors = $this->attributesCheck->check(
+			$scope,
+			$node->attrGroups,
+			Attribute::TARGET_ALL, // there is no dedicated trait target
```
ondrejmirtes: TARGET_CLASS should be used https://3v4l.org/DJPUf#v8.4.2

src/Rules/Traits/TraitAttributesRule.php
```diff
@@ -0,0 +1,61 @@
...
+		if ($traitName === null) {
+			return [];
+		}
+
+		if (!$this->reflectionProvider->hasClass($traitName->toString())) {
+			return [];
+		}
+
+		$errors = $this->attributesCheck->check(
+			$scope,
```
ondrejmirtes, line 35: We need to do something with $scope here. Please recreate this problem https://github.com/phpstan/phpstan-src/commit/3447391001f7a5c2bfb77f66c2d0e157242c1dae but with a trait.

We probably need to do `$scope->enterTrait()` here before calling the check. Just do `if !$scope instanceof MutatingScope then throw ShouldNotHappen`.

ondrejmirtes (changes requested):

ondrejmirtes:
Please keep this on 2.0.x. We don't need to be adding new features to 1.12.x anymore.

src/Rules/Traits/TraitAttributesRule.php
staabm: I have adjusted the PR. I think I have reproduced the problem you are after.

adding `$scope->enterTrait($classReflection)` didn't work and throws:
```
2) PHPStan\Rules\Traits\TraitAttributesRuleTest::testBug12011
PHPStan\ShouldNotHappenException: Internal error.

C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\ScopeContext.php:44
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\MutatingScope.php:2935
C:\dvl\Workspace\phpstan-src-staabm\src\Rules\Traits\TraitAttributesRule.php:49
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\FileAnalyser.php:111
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\NodeScopeResolver.php:482
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\NodeScopeResolver.php:354
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\NodeScopeResolver.php:842
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\NodeScopeResolver.php:307
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\FileAnalyser.php:212
C:\dvl\Workspace\phpstan-src-staabm\src\Analyser\Analyser.php:75
C:\dvl\Workspace\phpstan-src-staabm\src\Testing\RuleTestCase.php:177
C:\dvl\Workspace\phpstan-src-staabm\src\Testing\RuleTestCase.php:138
C:\dvl\Workspace\phpstan-src-staabm\tests\PHPStan\Rules\Traits\TraitAttributesRuleTest.php:78
```

at this point I have no idea what todo. could you fix the remaining thing?

staabm: let me fix the build real quick

src/Rules/Traits/TraitAttributesRule.php
```diff
@@ -0,0 +1,61 @@
...
+			$node->attrGroups,
+			Attribute::TARGET_CLASS,
+			'class',
+		);
+
+		$classReflection = $this->reflectionProvider->getClass($traitName->toString());
+		if (count($classReflection->getNativeReflection()->getAttributes('AllowDynamicProperties')) > 0) {
+			$errors[] = RuleErrorBuilder::message('Attribute class AllowDynamicProperties cannot be used with trait.')
+				->identifier('class.allowDynamicPropertiesTrait')
```
staabm: maybe?

```suggestion
				->identifier('trait.allowDynamicProperties')
```

src/Rules/Traits/TraitAttributesRule.php
ondrejmirtes: Push the code that's crashing like that please, I will take a look

staabm: here you are

ondrejmirtes: Oh right, we need to be in a class first before we can enter a trait.

I'd try two things: try to put the trait reflection into enterClass (but a different ShouldNotHappen might be thrown, not sure).

If that fails, we need to come up with some fake class to put into enterClass before we can enterTrait. It might be okay to simulate some anonymous Class_ node to use.

If it's too much work, don't worry and I'll merge the original code that doesn't use MutatingScope at all.

staabm: I have class-faking without a thrown exception working. the test-case is still failling though

Will have another look after sleeping over it

ondrejmirtes:
I changed my mind, I'm going to adjust it myself, thanks!

ondrejmirtes:
See my commit and reasoning in description: https://github.com/phpstan/phpstan-src/pull/3738/commits/b1cebc2e1f5133b1bd68ebb4401947c5f4d0e324

ondrejmirtes:
Please stop force-pushing here, I said I'm taking over. Thanks.

staabm:
> Please stop force-pushing here, I said I'm taking over. Thanks.

oh sorry, did not see your comments. I got it working just a moment ago ;)

ondrejmirtes:
Thank you!

staabm:
thanks

PR #3739, opened 2024-12-17
Fix regex comment support eating too much chars

staabm:
as noted by Jordi in https://github.com/phpstan/phpstan-src/pull/3735/files#r1888302665


ondrejmirtes:
Thank you.

PR #3740, opened 2024-12-19
Fix preg_match() group containing start/end meta characters

staabm:
closes https://github.com/phpstan/phpstan/issues/12297

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-12297.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-12297.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Bug12297;
+
+use function PHPStan\Testing\assertType;
+
+function doFoo($s) {
+	if (preg_match('/(^.+)/', $s, $matches) === 1) {
+		assertType('non-empty-string', $matches[1]);
+	}
+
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/nsrt/bug-12297.php
```diff
@@ -0,0 +1,19 @@
+<?php // lint >= 7.4
```
ondrejmirtes: What happens before 7.4?

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/bug-12297.php
staabm: no difference. I had the impression the nsrt/ tests require a lint php condition

ondrejmirtes: They don't. If it's omitted then it's always analysed.

I'm asking because after merging to 2.0.x I'd have to deleted it anyway. But if the tests passes on 7.2+ please remove this comment.

staabm: done :+1:

ondrejmirtes:
Thank you.

PR #3741, opened 2024-12-20
Add getOriginalNode() method on ClassPropertyNode

samsonasik:


samsonasik:
Ready to merge 👍

ondrejmirtes:
What do you need this for?

samsonasik:
First, it is for consistency with other virtual node, 
Second, I was experimenting to set scope on original node on rector side, when calling `NodeScopeResolver::processNodes()`, while only detected is `ClassPropertyNode` over original `Property` node.

ondrejmirtes:
The point of ClassPropertyNode is that it abstracts away Param and Property. It doesn't make sense to offer getOriginalNode because users would have to make if-else condition in their code about these two node types.

You can still hook your custom rector onto Param or Property, they do not disappear.

samsonasik:
Ok 👍

PR #3742, opened 2024-12-20
Support named arguments after unpacking

herndlm:
Closes https://github.com/phpstan/phpstan/issues/11418, which is apparently the most upvoted bug right now, even though quite recent hmm
Closes https://github.com/phpstan/phpstan/issues/8046

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -27,7 +27,6 @@ use PHPStan\Type\TypeUtils;
 use PHPStan\Type\VerbosityLevel;
 use function array_fill;
 use function array_key_exists;
-use function array_slice;
 use function count;
 use function implode;
 use function in_array;
@@ -82,22 +81,19 @@ final class FunctionCallParametersCheck
 	{
 		$functionParametersMinCount = 0;
 		$functionParametersMaxCount = 0;
-		$functionParameterNames = [];
 		foreach ($parametersAcceptor->getParameters() as $parameter) {
 			if (!$parameter->isOptional()) {
 				$functionParametersMinCount++;
 			}
 
 			$functionParametersMaxCount++;
-
-			$functionParameterNames[] = $parameter->getName();
 		}
 
 		if ($parametersAcceptor->isVariadic()) {
 			$functionParametersMaxCount = -1;
 		}
 
-		/** @var array<int, array{Expr, Type|null, bool, string|null, int}> $arguments */
+		/** @var array<int, array{Expr, Type|null, bool, string|null, int, bool}> $arguments */
 		$arguments = [];
 		/** @var array<int, Node\Arg> $args */
 		$args = $funcCall->getArgs();
@@ -111,6 +107,8 @@ final class FunctionCallParametersCheck
 				$argumentName = $arg->name->toString();
 			}
 
+			$nonUnpackAfterUnpacked = false;
+
 			if ($hasNamedArguments && $arg->unpack) {
 				$errors[] = RuleErrorBuilder::message('Named argument cannot be followed by an unpacked (...) argument.')
 					->identifier('argument.unpackAfterNamed')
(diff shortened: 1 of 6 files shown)


src/Rules/FunctionCallParametersCheck.php
```diff
@@ -109,20 +119,25 @@ public function check(
 					->build();
 			}
 			if ($hasUnpackedArgument && !$arg->unpack) {
-				$errors[] = RuleErrorBuilder::message('Unpacked argument (...) cannot be followed by a non-unpacked argument.')
-					->identifier('argument.nonUnpackAfterUnpacked')
-					->line($arg->getStartLine())
-					->nonIgnorable()
-					->build();
+				if ($argumentName !== null && $scope->getPhpVersion()->supportsNamedArgumentAfterUnpackedArgument()->yes()) {
+					if (in_array($argumentName, array_slice($functionParameterNames, 0, count($arguments)), true)) {
```
ondrejmirtes: What's going on here? Why the slice?

ondrejmirtes (changes requested):

src/Rules/FunctionCallParametersCheck.php
herndlm: `$functionParameterNames` is new and holds all parameter names the function accepts. `$arguments` is what this loop here is and was already building and includes all arguments that were passed via e.g. unpack or directly and so on. The slice based on the count of `$arguments` gives me the names of the parameters that were already "handled". So, if the current `$argumentName`, which is set in the case of handling a named argument for example, is inside there, it was already passed and passing it again would lead to a fatal error.

I had to think hard about this too now, thought about this already too long ago.. :D

ondrejmirtes: > I had to think hard about this too now

Exactly, this could use some refactoring. I don't get how "all handled parameters" tells us if one of them was unpacked or not. Also, we already report if an argument already overwrites an existing argument so instead we should make sure that place handles this instead of introducing new code to handle that.

See the `argument.duplicate` identifier.

herndlm: yeah, that was my initial thinking, but that stuff is somewhere different and handled later. I can make use of this, but need to pass information along this arguments array most likely. it's a bit complicated and ugly and I thought my addition has the least friction.

herndlm: but anyway, I can just try it out I guess, let's see. I most likely need to pass the information if this was a named argument and unpack already happened before it or so. then we should be good. blows up the array a bit with this edge case, but definitely more readable I think!

herndlm: ok, so I did something, WDYT? the tricky part is that the above handled `argument.nonUnpackAfterUnpacked` should not be created for such cases, while the new `argument.namedOverwriteAfterUnpacked` should and it is handled now below next to `argument.duplicate` which is then not necessary.

src/Rules/FunctionCallParametersCheck.php
```diff
@@ -176,6 +183,7 @@ public function check(
 							false,
 							$keyArgumentName,
 							$arg->getStartLine(),
+							$nonUnpackAfterUnpacked,
```
herndlm: could be replaced with `false`, but not sure what's more readable

src/Rules/FunctionCallParametersCheck.php
```diff
@@ -185,6 +193,7 @@ public function check(
 						true,
 						null,
 						$arg->getStartLine(),
+						$nonUnpackAfterUnpacked,
```
herndlm: could be replaced with `false`, but not sure what's more readable

src/Rules/FunctionCallParametersCheck.php
```diff
@@ -196,6 +205,7 @@ public function check(
 				false,
 				$argumentName,
 				$arg->getStartLine(),
+				$nonUnpackAfterUnpacked,
```
herndlm: could be replaced with `$hasUnpackedArgument` but not sure what's more readable

src/Rules/FunctionCallParametersCheck.php
```diff
@@ -607,10 +617,19 @@ private function processArguments(
 				&& !$parameter->isVariadic()
 				&& !array_key_exists($parameter->getName(), $unusedParametersByName)
 			) {
-				$errors[] = RuleErrorBuilder::message(sprintf('Argument for parameter $%s has already been passed.', $parameter->getName()))
-					->identifier('argument.duplicate')
-					->line($argumentLine)
-					->build();
+				if ($nonUnpackAfterUnpacked) {
+					$errors[] = RuleErrorBuilder::message(sprintf('Named parameter cannot overwrite already unpacked argument $%s.', $parameter->getName()))
```
ondrejmirtes: I'd argue we don't need to change this at all. The current error message (argument.duplicate) already describes the situation correctly.

ondrejmirtes (changes requested):
After this it's good to merge!

src/Rules/FunctionCallParametersCheck.php
herndlm: Interesting point. I think the only reason I'm doing this was to have a non-ignorable error because it is a fatal runtime one. But maybe it's not worth the complexity?

ondrejmirtes: I only make errors non-ignorable where I'm certain about it. Here we might depend on info from PHPDoc's array shapes etc. so it's not certain - it should be ignorable.

ondrejmirtes:
Thank you.

staabm:
@herndlm please check the build error on the 2.1.x branch

PR #3745, opened 2024-12-21
Allow process multiple UnreachableStatementNode on NodeScopeResolver::processNodes()

samsonasik:
/cc @TomasVotruba if this is accepted, this should hopefully will allow us to remove `UnreachableStatementNodeVisitor` on rector side.

https://github.com/rectorphp/rector-src/blob/e5d221e40d2c5acc1a92b1ded501321ee690be06/src/PHPStan/NodeVisitor/UnreachableStatementNodeVisitor.php#L17

which used if found a first `UnreachableStatementNode`, as phpstan only process first `UnreachableStatementNode`

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -6536,8 +6536,7 @@ final class NodeScopeResolver
 	}
 
 	/**
-	 * @template T of Node
-	 * @param array<T> $nodes
+	 * @param array<Node> $nodes
 	 * @return list<Node\Stmt>
 	 */
 	private function getNextUnreachableStatements(array $nodes, bool $earlyBinding): array
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -334,9 +334,11 @@ final class NodeScopeResolver
 				$nextStatements[] = $nextStmt;
 			}
 
-			if ($unreachableStatement instanceof Node\Stmt) {
-				$nodeCallback(new UnreachableStatementNode($unreachableStatement, $nextStatements), $scope);
+			if (!$unreachableStatement instanceof Node\Stmt) {
+				continue;
 			}
+
+			$nodeCallback(new UnreachableStatementNode($unreachableStatement, $nextStatements), $scope);
 		}
 	}
 
@@ -442,9 +444,11 @@ final class NodeScopeResolver
 				$nextStatements[] = $nextStmt;
 			}
 
-			if ($unreachableStatement instanceof Node\Stmt) {
-				$nodeCallback(new UnreachableStatementNode($unreachableStatement, $nextStatements), $scope);
+			if (!$unreachableStatement instanceof Node\Stmt) {
+				continue;
 			}
+
+			$nodeCallback(new UnreachableStatementNode($unreachableStatement, $nextStatements), $scope);
 		}
 
 		$statementResult = new StatementResult($scope, $hasYield, $alreadyTerminated, $exitPoints, $throwPoints, $impurePoints);
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -316,12 +316,10 @@ final class NodeScopeResolver
 			}
 
 			$alreadyTerminated = true;
-			$nextStmt = $this->getFirstUnreachableNode(array_slice($nodes, $i + 1), true);
-			if (!$nextStmt instanceof Node\Stmt) {
-				continue;
+			$nextStmts = $this->getNextUnreachableStatements(array_slice($nodes, $i + 1), true);
+			foreach ($nextStmts as $nextStmt) {
+				$nodeCallback(new UnreachableStatementNode($nextStmt), $scope);
 			}
-
-			$nodeCallback(new UnreachableStatementNode($nextStmt), $scope);
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


ondrejmirtes:
I don't want to spam the user with more of these messages, one of them is enough. But I'd accept a PR that adds this information to UnreachableStatementNode as a new getter, something like `getNextStatements(): array<Stmt>`.

samsonasik:
I see, I will try

samsonasik:
@ondrejmirtes implemented 👍 , I added `addNextStatement()` and `getNextStatements()` methods

samsonasik:
@ondrejmirtes I am not sure how unit test error is related:

```
There was 1 failure:

1) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "D:\a\phpstan-src\phpstan-src\tests\PHPStan\Analyser\nsrt\generic-unions.php" ('D:\a\phpstan-src\phpstan-src\...ns.php')
Failed assertions in D:\a\phpstan-src\phpstan-src\tests\PHPStan\Analyser\nsrt\generic-unions.php:

Line 120:
Expected: 3|'foo'
Actual:   mixed
```

could you help verify? thank you.

src/Analyser/NodeScopeResolver.php
```diff
@@ -308,12 +308,23 @@ public function processNodes(
...
+
+			if ($nextStmts === []) {
 				continue;
 			}
 
-			$nodeCallback(new UnreachableStatementNode($nextStmt), $scope);
+			$unreachableStatementNode = null;
+			foreach ($nextStmts as $nextStmt) {
+				if ($unreachableStatementNode instanceof UnreachableStatementNode) {
+					$unreachableStatementNode->addNextStatement($nextStmt);
```
ondrejmirtes: I generally want immutable objects. Please build the array first and then create the object.

tests/PHPStan/Rules/DeadCode/data/multiple_unreachable.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+/**
+ * @param 'foo' $foo
+ */
+function foo($foo)
```
ondrejmirtes, line 8: This file needs a namespace.

tests/PHPStan/Rules/Exceptions/CatchWithUnthrownExceptionRuleTest.php
```diff
@@ -105,6 +105,10 @@ public function testRule(): void
 				'Dead catch - Exception is never thrown in the try block.',
 				398,
 			],
+			[
+				'Dead catch - Exception is never thrown in the try block.',
+				407,
```
ondrejmirtes: Why did this change?

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Exceptions/CatchWithUnthrownExceptionRuleTest.php
samsonasik: I am not sure, without this, it have error:

```diff
There were 2 failures:

1) PHPStan\Rules\Exceptions\CatchWithUnthrownExceptionRuleTest::testRuleWithoutReportingUncheckedException
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 375: Dead catch - Exception is never thrown in the try block.
 380: Dead catch - Exception is never thrown in the try block.
 398: Dead catch - Exception is never thrown in the try block.
+407: Dead catch - Exception is never thrown in the try block.
```

samsonasik: Ok, add namespace seems fix it https://github.com/phpstan/phpstan-src/pull/3745/commits/66596b45eb61df295e3db477ae99e4d45a48edff

tests/PHPStan/Rules/DeadCode/data/multiple_unreachable.php
samsonasik: namespace added https://github.com/phpstan/phpstan-src/commit/66596b45eb61df295e3db477ae99e4d45a48edff

src/Analyser/NodeScopeResolver.php
samsonasik: updated to immutable object https://github.com/phpstan/phpstan-src/commit/66596b45eb61df295e3db477ae99e4d45a48edff

samsonasik:
@ondrejmirtes it seems working ok now, Ready to merge 👍

ondrejmirtes (changes requested):
You're not actually testing `getNextStatements` anywhere.

Please create a new RuleTestCase inside tests/ and write an anonymous class in `getRule()` that would verify this information, for example by putting the next statements into reported errors.

samsonasik:
I will try

samsonasik:
I've added test for `getNextStatements()`  https://github.com/phpstan/phpstan-src/pull/3745/commits/1083c0be52f2b9c454e8a65a94002f38f71de71a

samsonasik:
@ondrejmirtes test added, Ready to merge 👍

samsonasik:
@ondrejmirtes mergeable?

samsonasik:
@ondrejmirtes Rebased 👍

ondrejmirtes:
I just pushed some test adjustments and a failing test for you.

The output for testRuleTopLevel is currently wrong. See: https://phpstan.org/r/7a2aafcd-10b3-405d-9402-90037f780059 - the top level function is not considered unreachable currently by PHPStan. But it's included in your "next statements" which is wrong.

samsonasik:
@ondrejmirtes I've fixed handling top level function https://github.com/phpstan/phpstan-src/pull/3745/commits/2427d866cd5ff65768432e4bf3210dd16f7d23a3 👍

ondrejmirtes:
Thank you.

samsonasik:
Thank you @ondrejmirtes I will test on rector-src 👍

samsonasik:
linked PR on rector-src for future reference, tested the `UnreachableStatementNode::getNextStatements()` and it seems working ok 🎉, see

- https://github.com/rectorphp/rector-src/pull/6642

PR #3748, opened 2024-12-21
Improve loose comparison on integer

staabm:
see https://www.php.net/manual/en/types.comparisons.php

grafik

see https://3v4l.org/395hr

### DIFF
--- a/src/Type/IntegerType.php
+++ b/src/Type/IntegerType.php
@@ -10,6 +10,7 @@ use PHPStan\Type\Accessory\AccessoryLowercaseStringType;
 use PHPStan\Type\Accessory\AccessoryNumericStringType;
 use PHPStan\Type\Accessory\AccessoryUppercaseStringType;
 use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Traits\NonArrayTypeTrait;
 use PHPStan\Type\Traits\NonCallableTypeTrait;
@@ -139,6 +140,18 @@ class IntegerType implements Type
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		if ($type->isArray()->yes()) {
+			return new ConstantBooleanType(false);
+		}
+
+		if (
+			$phpVersion->nonNumericStringAndIntegerIsFalseOnLooseComparison()
+			&& $type->isString()->yes()
+			&& $type->isNumericString()->no()
+		) {
+			return new ConstantBooleanType(false);
+		}
+
 		return new BooleanType();
 	}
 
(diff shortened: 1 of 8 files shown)


tests/PHPStan/Analyser/nsrt/equal.php
```diff
@@ -152,7 +152,6 @@ public function doFoo(string $a, int $b, float $c): void
 		assertType('false', 'a' != 'a');
 		assertType('true', 'a' != 'b');
 
-		assertType('bool', $b == 'a');
```
staabm, line 155: moved this case into the loose-comparison-* files, as it depends on php version

src/Type/Traits/ConstantScalarTypeTrait.php
```diff
@@ -58,7 +58,7 @@ public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 		}
 
 		if ($type->isConstantArray()->yes() && $type->isIterableAtLeastOnce()->no()) {
-			// @phpstan-ignore equal.notAllowed, equal.invalid
+			// @phpstan-ignore equal.notAllowed, equal.invalid, equal.alwaysFalse
```
staabm, line 61: ignores

```
------ ------------------------------------------------------------------------------------------------------------- 
  Line   src/Type/Traits/ConstantScalarTypeTrait.php (in context of class PHPStan\Type\Constant\ConstantIntegerType)  
 ------ ------------------------------------------------------------------------------------------------------------- 
  62     Loose comparison using == between int and array{} will always evaluate to false.                             
         🪪  equal.alwaysFalse                                                                                        
 ------ ------------------------------------------------------------------------------------------------------------- 

```

tests/PHPStan/Rules/Comparison/ConstantLooseComparisonRuleTest.php
```diff
@@ -173,6 +174,24 @@ public function testBug11694(): void
...
+					29,
+					'Because the type is coming from a PHPDoc, you can turn off this check by setting <fg=cyan>treatPhpDocTypesAsCertain: false</> in your <fg=cyan>%configurationFile%</>.',
+				],
+				[
+					"Loose comparison using == between int<10, 20> and '13foo' will always evaluate to false.",
+					30,
+					'Because the type is coming from a PHPDoc, you can turn off this check by setting <fg=cyan>treatPhpDocTypesAsCertain: false</> in your <fg=cyan>%configurationFile%</>.',
+				],
+			]);
+		}
```
staabm, line 192: https://3v4l.org/W28WC

src/Type/IntegerType.php
```diff
@@ -139,6 +140,18 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
```
staabm, line 141: I think we have a new todo for PHPStan 3.x: adjust all `Type` methods which depend on `PhpVersion` and turn them into using `PhpVersions`

ondrejmirtes: Yeah, this should be possible. It's possible to make PhpVersions from PhpVersion (without losing precision), but not vice-versa.

So places that only have PhpVersion could still call these methods by transforming it into PhpVersions first.

ondrejmirtes:
Awesome, thank you!

PR #3749, opened 2024-12-21
Better typing for mb_convert_encoding

jack-worman:
Closes https://github.com/phpstan/phpstan/issues/12309

Also, fixed types related to the array input and output; the output array has the exact same shape as the input array: https://3v4l.org/OrdDb


resources/functionMap.php
```diff
@@ -6312,7 +6312,7 @@
 'mb_check_encoding' => ['bool', 'var='=>'string|array<string>', 'encoding='=>'string'],
 'mb_chr' => ['string|false', 'cp'=>'int', 'encoding='=>'string'],
 'mb_convert_case' => ['string', 'sourcestring'=>'string', 'mode'=>'int', 'encoding='=>'string'],
-'mb_convert_encoding' => ['string|array<int, string>|false', 'val'=>'string|array<int, string>', 'to_encoding'=>'string', 'from_encoding='=>'mixed'],
+'mb_convert_encoding' => ['string|array<scalar|null|array<scalar|null>>|false', 'val'=>'string|array<scalar|null|array<scalar|null>>', 'to_encoding'=>'string', 'from_encoding='=>'mixed'],
```
jack-worman, line 6315: Is there a way to represent a recursive type?
Or is only going 3 layers deep good enough?

src/Type/Php/MbConvertEncodingFunctionReturnTypeExtension.php
```diff
@@ -25,18 +24,16 @@ public function getTypeFromFunctionCall(
...
+		$argType = $scope->getType($args[0]->value);
 		$isString = $argType->isString();
 		$isArray = $argType->isArray();
-		$compare = $isString->compareTo($isArray);
-		if ($compare === $isString) {
-			return new StringType();
-		} elseif ($compare === $isArray) {
-			return new ArrayType(new IntegerType(), new StringType());
+		if ($isString->yes() || $isArray->yes()) {
+			return new UnionType([$argType, new ConstantBooleanType(false)]);
```
zonuexe, line 34: In all versions, the `mb_convert_encoding()` function throws an error for invalid encodings, so it makes sense to use BenevolentUnionType instead of Union. https://3v4l.org/YuXmF

```suggestion
			return new BenevolentUnionType([$argType, new ConstantBooleanType(false)]);
```

Using a benevolent union, we can pass it to function calls expecting strings without any extra assertions, like this: https://phpstan.org/r/fbcfe268-adc6-48cd-9f35-c66b3117c269

jack-worman: I am not sure about this change because I don't know if there are other ways this function can fail.

zonuexe: [As the PHP manual says](https://www.php.net/manual/en/function.mb-convert-encoding.php#refsect1-function.mb-convert-encoding-errors), the function will only raise an `E_WARNING` if an invalid encoding is specified.

> ## Errors/Exceptions
> As of PHP 8.0.0, a [ValueError](https://www.php.net/manual/en/class.valueerror.php) is thrown if the value of to_encoding or from_encoding is an invalid encoding. Prior to PHP 8.0.0, a [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) was emitted instead.

This function behaves this way in all versions, so it will never return `null`, even if an invalid byte sequence is input.

 * https://github.com/php/php-src/blob/PHP-7.3/ext/mbstring/mbstring.c#L3057-L3060
 * https://github.com/php/php-src/blob/PHP-8.4/ext/mbstring/mbstring.c#L2728-L2729

tests/PHPStan/Analyser/nsrt/bug-3336.php
```diff
@@ -1,10 +0,0 @@
-<?php // lint < 8.0
```
VincentLanglet, line 1: Existing bug are generally not removed but updated. (to see the diff)

VincentLanglet (commented):
It's possible to produce a false return value, for instance with
```
mb_convert_encoding("\x80\x81\x82", "UTF-8", 'auto')
```

So I'm unsure about changing the Union to a BenevolentUnion.

People should be advised to add a `false` check on the result.

src/Type/Php/MbConvertEncodingFunctionReturnTypeExtension.php
VincentLanglet: There is still use cases which returns false, like
```
mb_convert_encoding("\x80\x81\x82", "UTF-8", 'auto')
```

So someone calling `mb_convert_encoding($string, $from, $to)` should be still advised to use a false check.

UnionType might be preferred then.

tests/PHPStan/Analyser/nsrt/bug-3336.php
jack-worman: reverted delete

src/Type/Php/MbConvertEncodingFunctionReturnTypeExtension.php
jack-worman: reverted back to regular union type

resources/functionMap.php
VincentLanglet: I think the best would be to use `string|array<mixed>|false`

src/Type/Php/MbConvertEncodingFunctionReturnTypeExtension.php
```diff
@@ -25,18 +24,14 @@ public function getTypeFromFunctionCall(
...
-		$isString = $argType->isString();
-		$isArray = $argType->isArray();
-		$compare = $isString->compareTo($isArray);
-		if ($compare === $isString) {
-			return new StringType();
-		} elseif ($compare === $isArray) {
-			return new ArrayType(new IntegerType(), new StringType());
+		$argType = $scope->getType($args[0]->value);
+		if ($argType->isString()->yes() || $argType->isArray()->yes()) {
+			return new UnionType([$argType, new ConstantBooleanType(false)]);
```
VincentLanglet, line 34:
```
TypeCombinator::union($argType, new ConstantBooleanType(false))
```
should be used. And it's not ok for constants array or constant strings.

VincentLanglet: I think https://github.com/phpstan/phpstan-src/pull/3914 does the job

ondrejmirtes:
Succeeeded by https://github.com/phpstan/phpstan-src/pull/3914

PR #3750, opened 2024-12-22
Improve loose comparison on union type

staabm:
closes https://github.com/phpstan/phpstan/issues/8800

### DIFF
--- a/src/TrinaryLogic.php
+++ b/src/TrinaryLogic.php
@@ -47,18 +47,6 @@ final class TrinaryLogic
 		return self::$registry[$yesNo] ??= new self($yesNo);
 	}
 
-	public static function createFromBooleanType(BooleanType $type): self
-	{
-		if ($type->isTrue()->yes()) {
-			return self::createYes();
-		}
-		if ($type->isFalse()->yes()) {
-			return self::createNo();
-		}
-
-		return self::createMaybe();
-	}
-
 	private static function create(int $value): self
 	{
 		self::$registry[$value] ??= new self($value);
--- a/src/TrinaryLogic.php
+++ b/src/TrinaryLogic.php
@@ -47,6 +47,18 @@ final class TrinaryLogic
 		return self::$registry[$yesNo] ??= new self($yesNo);
 	}
 
+	public static function createFromBooleanType(BooleanType $type): self
+	{
+		if ($type->isTrue()->yes()) {
+			return self::createYes();
+		}
+		if ($type->isFalse()->yes()) {
+			return self::createNo();
+		}
+
+		return self::createMaybe();
+	}
+
 	private static function create(int $value): self
 	{
 		self::$registry[$value] ??= new self($value);
--- a/src/Type/Accessory/AccessoryLowercaseStringType.php
+++ b/src/Type/Accessory/AccessoryLowercaseStringType.php
@@ -11,6 +11,7 @@ use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\CompoundType;
 use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\FloatType;
@@ -326,6 +327,10 @@ class AccessoryLowercaseStringType implements CompoundType, AccessoryType
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		if ($type->isString()->yes() && $type->isLowercaseString()->no()) {
+			return new ConstantBooleanType(false);
+		}
+
 		return new BooleanType();
 	}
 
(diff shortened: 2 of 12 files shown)


ondrejmirtes (changes requested):
Would ve great to do it with IntersectionType too.

I suspect the implementations should resemble TrinaryLogic::minMax and extremeIdentity so I want to compare them with that.

staabm:
I think its pretty neat now :)

edit: just realized I need to fix the implementations in the accessory types

tests/PHPStan/Analyser/nsrt/loose-comparisons.php
```diff
@@ -653,4 +653,85 @@ public function sayInt(
...
+		if ($upper != '') {
+			assertType('false', '' == $upper);
+		}
+		assertType('bool', '0' == $upper);
+		assertType('bool', 'a' == $upper); // should be false
+		assertType('bool', 'abc' == $upper); // should be false
+		assertType('bool', 'aBc' == $upper);
+		assertType('bool', strtoupper($s) == $upper);
+		assertType('bool', strtolower($s) == $upper); // should be false
+		assertType('bool', $upper == $lower); // should be false
```
staabm: @VincentLanglet are you interessted in implementing `looseCompare()` in upper/lower-case string type? :)

I tried but can't get my head arround this atm

staabm:
added tests for non-empty-array and non-empty-string to show the intersection-type loose compare is working

tests/PHPStan/Analyser/nsrt/loose-comparisons.php
VincentLanglet: The tests are not true, I edited with comment 
```
		assertType('bool', 'aBc' == $upper); // => Should be false no ?
		
		assertType('bool', strtolower($s) == $upper); // should be false => NO, it can be true with `$s = '&!<42' or an empty string
		
		assertType('bool', $upper == $lower); // should be false => NO, it can be true with `$upper = ''` or `$upper = '&!<42'`
		
		assertType('bool', 'AbC' == $lower); // => Should be false no ?
		
		assertType('bool', strtoupper($s) == $lower); // should be false => NO, it can be true with `$s = '&!<42' or an empty string
		
		assertType('bool', $lower == $upper); // should be false => NO, it can be true with `$lower = ''` or `$lower = '&!<42'`
```

So maybe it's the reason of your difficulties ?

For the implementation I would go with
https://github.com/phpstan/phpstan-src/commit/ff999299810056a8cb419f818a0bb85e0697b758

I had to comment the `if` condition because there is a bug, you might know how to fix
```
// Here PHPStan consider $lower is `lowercase`
if ($lower != '') {
     assertType('false', '' == $lower);
     // Here PHPStan consider $lower is `lowercase&non-empty-string`
}
// Here PHPStan consider $lower is `string`
```

VincentLanglet: See https://phpstan.org/r/5b9fe2de-f827-4e3b-99ac-9d8b62dca596 @staabm 

So you might consider moving the assertion to avoid this bug https://github.com/phpstan/phpstan-src/commit/36bda290d1487a8705fc802e11cb28e2bfb78964

staabm: thanks for looking into it. I have added your proposal as a separate commit.

src/Type/Accessory/AccessoryLowercaseStringType.php
```diff
@@ -326,6 +327,10 @@ public function isScalar(): TrinaryLogic
 
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
+		if ($type->isString()->yes() && $type->isLowercaseString()->no()) {
```
mvorisek: Does https://3v4l.org/tEobn test pass (and vice versa)?

staabm: Sounds like a very edge case which should be reported as a new issue.

I don't expect it works with this PR

VincentLanglet: No, it's considered as false with the current implementation.

@staabm I think you'll need to add
```
&& ($type->isNumericString()->no() || $this->isNumericString()->no())
```
condition.

mvorisek: It works already https://phpstan.org/r/3b02891b-463b-4d09-b845-428b10f44547 somehow ;-)

mvorisek: I do not see any issue yet - https://phpstan.org/r/745f421e-ae6f-46f1-aa32-d2fe28d88b3f . But this PR will probably break it.

VincentLanglet: You can cherry pick https://github.com/phpstan/phpstan-src/commit/e11991f2dffbf56f2f8c8be9652454547285e6b3 staabm

staabm: cherry picked and added a few more tests, thanks

src/TrinaryLogic.php
```diff
@@ -47,6 +47,18 @@ public static function createFromBoolean(bool $value): self
 		return self::$registry[$yesNo] ??= new self($yesNo);
 	}
 
+	public static function createFromBooleanType(BooleanType $type): self
```
ondrejmirtes: You had the same idea, nice! But I'd actually like `BooleanType::toTrinaryLogic()` more :)

ondrejmirtes (changes requested):

src/TrinaryLogic.php
staabm: hehe I see. done

ondrejmirtes:
Thank you!

PR #3751, opened 2024-12-22
Fix union of lowercase/uppercase string with empty string

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12312

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -3,8 +3,10 @@
 namespace PHPStan\Type;
 
 use PHPStan\Type\Accessory\AccessoryArrayListType;
+use PHPStan\Type\Accessory\AccessoryLowercaseStringType;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\Accessory\AccessoryType;
+use PHPStan\Type\Accessory\AccessoryUppercaseStringType;
 use PHPStan\Type\Accessory\HasOffsetType;
 use PHPStan\Type\Accessory\HasOffsetValueType;
 use PHPStan\Type\Accessory\HasPropertyType;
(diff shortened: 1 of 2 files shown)


staabm (approved):
lgtm - thanks!

src/Type/TypeCombinator.php
```diff
@@ -461,34 +466,55 @@ private static function compareTypesInUnion(Type $a, Type $b): ?array
...
-			return [new StringType(), null];
+			return [self::intersect(
+				new StringType(),
+				...self::getAccessoryCaseStringTypes($a),
+			), null];
 		}
 
 		if (
 			$a instanceof ConstantStringType
 			&& $a->getValue() === '0'
```
staabm, line 477: This looks like we need another test (the change seems untested)

tests/PHPStan/Analyser/nsrt/bug-12312.php
```diff
@@ -0,0 +1,41 @@
...
+use function PHPStan\Testing\assertType;
+
+class HelloWorld
+{
+	/**
+	 * @param lowercase-string $s
+	 */
+	public function sayLowercase(string $s): void
+	{
+		if ($s != '') {
```
staabm, line 14: I think we need another test for the condition with operands in reversed order

staabm, line 14: I think we need another test test for the condition with operands in reversed order

src/Type/TypeCombinator.php
VincentLanglet: Done

tests/PHPStan/Analyser/nsrt/bug-12312.php
VincentLanglet: Done

ondrejmirtes:
Thank you.

PR #3752, opened 2024-12-22
Fix crash on MethodCallableNode on do while

samsonasik:
Fixes https://github.com/phpstan/phpstan/issues/12246

Ref https://phpstan.org/r/178ac73a-e99c-4961-957f-a3c6062d7c44
Ref https://3v4l.org/5s30L

### DIFF
--- a/src/Node/Printer/ExprPrinter.php
+++ b/src/Node/Printer/ExprPrinter.php
@@ -3,6 +3,7 @@
 namespace PHPStan\Node\Printer;
 
 use PhpParser\Node\Expr;
+use PHPStan\Node\MethodCallableNode;
 
 /**
  * @api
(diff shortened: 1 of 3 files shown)


samsonasik:
Ready to merge 👍

ondrejmirtes (changes requested):
1) MethodCallableNode is not the only one with this problem. There's also FunctionCallableNode, InstantiationCallableNode, StaticMethodCallableNode. Please reproduce and fix issues with them too.
2) Crashing examples have regression tests in AnalyserIntegrationTest, not in a specific rule. Because we're not fixing an error in a specific rule.
3) Printing of custom nodes is fixed by adding a method on PHPStan\Node\Printer\Printer, not by modifying ExprPrinter.

Thanks.

samsonasik:
I will try

VincentLanglet:
https://github.com/phpstan/phpstan/issues/12246 is now closed so I think we can close this.

PR #3753, opened 2024-12-22
Implement `OpenSslEncryptParameterOutTypeExtension`

paulbalandan:
Ref: https://www.php.net/manual/en/function.openssl-encrypt.php

### DIFF
--- a/src/Type/Php/OpenSslEncryptParameterOutTypeExtension.php
+++ b/src/Type/Php/OpenSslEncryptParameterOutTypeExtension.php
@@ -1,0 +1,70 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParameterReflection;
+use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
+use PHPStan\Type\FunctionParameterOutTypeExtension;
+use PHPStan\Type\NullType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use function in_array;
+use function openssl_get_cipher_methods;
+use function strtolower;
+use function substr;
+
+final class OpenSslEncryptParameterOutTypeExtension implements FunctionParameterOutTypeExtension
+{
+
+	public function isFunctionSupported(FunctionReflection $functionReflection, ParameterReflection $parameter): bool
+	{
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/nsrt/openssl.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+use function PHPStan\Testing\assertType;
+
+$ciphertext = openssl_encrypt('plain text', 'aes-256-gcm', 'key', OPENSSL_RAW_DATA, 'iv', $tag, '', 16);
+
+assertType('string|false', $ciphertext);
+assertType('string', $tag);
```
jack-worman: Can be null for some cipher_algo's:
```php
$ciphertext = openssl_encrypt(
    '',
    $cipher_algo = 'aes-128-cbc',
    '',
    OPENSSL_RAW_DATA,
    random_bytes(openssl_cipher_iv_length($cipher_algo)),
    $tag,
);
var_export($tag); // NULL
```

src/Type/Php/OpenSslEncryptParameterOutTypeExtension.php
```diff
@@ -0,0 +1,57 @@
...
+	public function getParameterOutTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $funcCall, ParameterReflection $parameter, Scope $scope): ?Type
+	{
+		$args = $funcCall->getArgs();
+		$cipherArg = $args[1] ?? null;
+
+		if ($cipherArg === null) {
+			return null;
+		}
+
+		$cipherType = current($scope->getType($cipherArg->value)->getConstantStrings());
```
staabm: why do you only work on the first cipher type

paulbalandan: Thanks, I just realised you can do unions of literal strings. Updated the code to check all cases.

src/Type/Php/OpenSslEncryptParameterOutTypeExtension.php
```diff
@@ -0,0 +1,79 @@
...
+		$args = $funcCall->getArgs();
+		$cipherArg = $args[1] ?? null;
+
+		if ($cipherArg === null) {
+			return null;
+		}
+
+		$cipherType = $scope->getType($cipherArg->value);
+
+		return TypeTraverser::map($cipherType, static function (Type $type, callable $traverse): Type {
```
staabm: I don't think you need the TypeTraverser here. Couldn't you just iterate over `getConstantStrings()`?

paulbalandan: Updated to just use `foreach`.

ondrejmirtes:
Thank you!

PR #3757, opened 2024-12-26
Improve `preg_split()` function ReturnType

malsuke:
I developed further extended the extension for the `preg_split()` function.

The `preg_split()` function can specify ReturnType with the following cases:

- When the regular expression passed as the `$pattern` argument is invalid.
- When the `$pattern` and `$subject` arguments are constants.
- When the string passed as the `$subject` argument is non-empty-string.
- When the `$flag` argument is set to one or more of the following: `PREG_SPLIT_OFFSET_CAPTURE`, `PREG_SPLIT_NO_EMPTY`, or `PREG_SPLIT_DELIM_CAPTURE`.

The detailed cases are specified in the test cases.

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -9081,7 +9081,7 @@ return [
 'preg_replace' => ['string|array|null', 'regex'=>'string|array', 'replace'=>'string|array', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
 'preg_replace_callback' => ['string|array|null', 'regex'=>'string|array', 'callback'=>'callable(array<int|string, string>):string', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
 'preg_replace_callback_array' => ['string|array|null', 'pattern'=>'array<string,callable>', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
-'preg_split' => ['list<string>|false', 'pattern'=>'string', 'subject'=>'string', 'limit='=>'?int', 'flags='=>'int'],
+'preg_split' => ['__benevolent<list<string>|list<array{string, int<0, max>}>|false>', 'pattern'=>'string', 'subject'=>'string', 'limit='=>'?int', 'flags='=>'int'],
 'prev' => ['mixed', '&rw_array_arg'=>'array|object'],
 'print_r' => ['string|true', 'var'=>'mixed', 'return='=>'bool'],
 'printf' => ['int', 'format'=>'string', '...values='=>'__stringAndStringable|int|float|null|bool'],
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -9081,7 +9081,7 @@ return [
 'preg_replace' => ['string|array|null', 'regex'=>'string|array', 'replace'=>'string|array', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
 'preg_replace_callback' => ['string|array|null', 'regex'=>'string|array', 'callback'=>'callable(array<int|string, string>):string', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
 'preg_replace_callback_array' => ['string|array|null', 'pattern'=>'array<string,callable>', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
-'preg_split' => ['__benevolent<list<string>|list<array{string, int<0, max>}>|false>', 'pattern'=>'string', 'subject'=>'string', 'limit='=>'?int', 'flags='=>'int'],
+'preg_split' => ['list<string>|list<array{string, int<0, max>}>|false', 'pattern'=>'string', 'subject'=>'string', 'limit='=>'?int', 'flags='=>'int'],
 'prev' => ['mixed', '&rw_array_arg'=>'array|object'],
 'print_r' => ['string|true', 'var'=>'mixed', 'return='=>'bool'],
 'printf' => ['int', 'format'=>'string', '...values='=>'__stringAndStringable|int|float|null|bool'],
--- a/src/Type/Php/PregSplitDynamicReturnTypeExtension.php
+++ b/src/Type/Php/PregSplitDynamicReturnTypeExtension.php
@@ -1,4 +1,4 @@
-<?php declare(strict_types = 1);
+<?php declare(strict_types=1);
 
 namespace PHPStan\Type\Php;
 
@@ -7,6 +7,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
+use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\BitwiseFlagHelper;
 use PHPStan\Type\Constant\ConstantArrayType;
@@ -18,9 +19,16 @@ use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\TypeUtils;
+use function count;
+use function is_array;
+use function is_int;
+use function preg_match;
+use function preg_split;
 use function strtolower;
 
 final class PregSplitDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
@@ -53,19 +61,10 @@ final class PregSplitDynamicReturnTypeExtension implements DynamicFunctionReturn
 		$subjectConstantTypes = $subjectType->getConstantStrings();
 
 		if (
-			count($patternConstantTypes) > 0 &&
-			@preg_match($patternConstantTypes[0]->getValue(), "") === false
+			count($patternConstantTypes) > 0
+			&& @preg_match($patternConstantTypes[0]->getValue(), "") === false
 		) {
-
 			return new ErrorType();
-		}
-
-		if ($subjectArg !== null && $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($subjectArg->value, $scope, 'PREG_SPLIT_OFFSET_CAPTURE')->yes()) {
-			$type = new ArrayType(
-				new IntegerType(),
-				new ConstantArrayType([new ConstantIntegerType(0), new ConstantIntegerType(1)], [new StringType(), IntegerRangeType::fromInterval(0, null)], [2], [], TrinaryLogic::createYes()),
-			);
-			return TypeCombinator::union(TypeCombinator::intersect($type, new AccessoryArrayListType()), new ConstantBooleanType(false));
 		}
 
 		if ($limitArg === null) {
@@ -82,15 +81,47 @@ final class PregSplitDynamicReturnTypeExtension implements DynamicFunctionReturn
 			$flags = $flagType->getConstantScalarValues();
 		}
 
-		if (count($patternConstantTypes) === 0 || count($subjectConstantTypes) === 0 || count($flags) === 0) {
+		if (count($patternConstantTypes) === 0 || count($subjectConstantTypes) === 0) {
+			$returnNonEmptyStrings = $flagArg !== null && $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagArg->value, $scope, 'PREG_SPLIT_NO_EMPTY')->yes();
+			if ($returnNonEmptyStrings) {
+				$stringType = TypeCombinator::intersect(
+					new StringType(),
+					new AccessoryNonEmptyStringType()
+				);
+			} else {
+				$stringType = new StringType();
+			}
+
+			if ($flagArg !== null && $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagArg->value, $scope, 'PREG_SPLIT_OFFSET_CAPTURE')->yes()) {
+				$type = new ArrayType(
+					new IntegerType(),
+					new ConstantArrayType([new ConstantIntegerType(0), new ConstantIntegerType(1)], [$stringType, IntegerRangeType::fromInterval(0, null)], [2], [], TrinaryLogic::createYes()),
+				);
+				return TypeUtils::toBenevolentUnion(
+					TypeCombinator::union(
+						TypeCombinator::intersect($type, new AccessoryArrayListType()),
+						new ConstantBooleanType(false)
+					)
+				);
+			}
+
+			if ($flagArg !== null && $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagArg->value, $scope, 'PREG_SPLIT_NO_EMPTY')->yes()) {
+				return TypeUtils::toBenevolentUnion(
+					TypeCombinator::union(
+						TypeCombinator::intersect(new ArrayType(new MixedType(), $stringType), new AccessoryArrayListType()),
+						new ConstantBooleanType(false)
+					)
+				);
+			}
+
 			return null;
 		}
 
 		$resultTypes = [];
 		foreach ($patternConstantTypes as $patternConstantType) {
 			foreach ($subjectConstantTypes as $subjectConstantType) {
-				foreach ($flags as $flag) {
-					foreach ($limits as $limit) {
+				foreach ($limits as $limit) {
+					foreach ($flags as $flag) {
 						$result = @preg_split($patternConstantType->getValue(), $subjectConstantType->getValue(), $limit, $flag);
 						if ($result !== false) {
 							$constantArray = ConstantArrayTypeBuilder::createEmpty();
--- a/src/Type/Php/PregSplitDynamicReturnTypeExtension.php
+++ b/src/Type/Php/PregSplitDynamicReturnTypeExtension.php
@@ -8,6 +8,7 @@ use PHPStan\Reflection\FunctionReflection;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
+use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\BitwiseFlagHelper;
 use PHPStan\Type\Constant\ConstantArrayType;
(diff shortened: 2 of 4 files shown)


src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +48,125 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+			$limitType = $scope->getType($limitArg->value);
+			$limits = $limitType->getConstantScalarValues();
+		}
+
+		if ($flagArg === null) {
+			$flags = [0];
+		} else {
+			$flagType = $scope->getType($flagArg->value);
+			$flags = $flagType->getConstantScalarValues();
+		}
```
zonuexe, line 102: By replacing it as follows, type checking within multiple Constant loops will no longer be necessary.

```php
$flags = [];
$flagType = $scope->getType($flagArg->value);
foreach ($flagType->getConstantScalarValues() as $flag) {
    if (!is_int()) {
        return new ErrorType();
    }

    $flags[] = $flag;
}
```

zonuexe: resolved https://github.com/phpstan/phpstan-src/pull/3757/commits/8cb3030cbe2e0cbf4da1d42bb9198d450d136df9

resources/functionMap.php
```diff
@@ -9081,7 +9081,7 @@
 'preg_replace' => ['string|array|null', 'regex'=>'string|array', 'replace'=>'string|array', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
 'preg_replace_callback' => ['string|array|null', 'regex'=>'string|array', 'callback'=>'callable(array<int|string, string>):string', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
 'preg_replace_callback_array' => ['string|array|null', 'pattern'=>'array<string,callable>', 'subject'=>'string|array', 'limit='=>'int', '&w_count='=>'int'],
-'preg_split' => ['list<string>|false', 'pattern'=>'string', 'subject'=>'string', 'limit='=>'?int', 'flags='=>'int'],
+'preg_split' => ['__benevolent<list<string>|list<array{string, int<0, max>}>|false>', 'pattern'=>'string', 'subject'=>'string', 'limit='=>'?int', 'flags='=>'int'],
```
VincentLanglet: The other preg_ method are not using benevolent union, so I would think more consistent to not use a benevolent union here too.

malsuke: fix 37f9b3e2650da6d6bd16323d8a1609a3eb5caad6

malsuke: @VincentLanglet @ondrejmirtes 

I understand. I would like to remove benevolent union.

On the other hand, I think that preg_split should not return false unless there is an issue with the regular expression. Furthermore, in this PR, I have modified the code so that if the regular expression is incorrect, an error is returned early in the parsing process.

Therefore, if the regular expression is correct, I am considering not adding false as a Union.
(In this case, this bug can also be fixed. https://github.com/phpstan/phpstan-src/blob/76740fd95bbe616331c66de45f489c697163a52b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php#L890-L900)

If you think not to use benevolent union, do you think it would be fine to remove false? I would like to hear your opinion on this. I would like to get your opinion before making any modifications.

VincentLanglet: Like any other preg method I think it can returns false if an internal error occurs like
- some memory limit is reached
- too many recursion
- some invalid encoding

And in the pho.ini there is some config like pcre.recursion_limit or pcre.backtrack_limit.

So I would keep a non-benevolent union AND false.

If we decide to remove false from the signature it should be removed from all the preg methods. But I dont think we should go this way.

malsuke: I understand.
I modified it to keep a non-benevolent union AND false.

37f9b3e2650da6d6bd16323d8a1609a3eb5caad6

ondrejmirtes:
@staabm As this is about regexes, can you please review it? Thank you.

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+		$limitArg = $args[2] ?? null;
+		$flagArg = $args[3] ?? null;
+		$patternType = $scope->getType($patternArg->value);
+		$patternConstantTypes = $patternType->getConstantStrings();
+		$subjectType = $scope->getType($subjectArg->value);
+		$subjectConstantTypes = $subjectType->getConstantStrings();
+
+		if (
+			count($patternConstantTypes) > 0
+			&& @preg_match($patternConstantTypes[0]->getValue(), '') === false
```
staabm: we usually us `Strings::match`

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+			return null;
+		}
+		$patternArg = $args[0];
+		$subjectArg = $args[1];
+		$limitArg = $args[2] ?? null;
+		$flagArg = $args[3] ?? null;
+		$patternType = $scope->getType($patternArg->value);
+		$patternConstantTypes = $patternType->getConstantStrings();
+		$subjectType = $scope->getType($subjectArg->value);
+		$subjectConstantTypes = $subjectType->getConstantStrings();
```
staabm, line 70: I would move the subject type declartion below the early returning if in the next block

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+		}
+
+		$flags = [];
+		if ($flagArg === null) {
+			$flags = [0];
+		} else {
+			$flagType = $scope->getType($flagArg->value);
+			foreach ($flagType->getConstantScalarValues() as $flag) {
+				if (!is_int($flag)) {
+					return new ErrorType();
```
staabm, line 98: this case is missing a test

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+		}
+
+		$limits = [];
+		if ($limitArg === null) {
+			$limits = [-1];
+		} else {
+			$limitType = $scope->getType($limitArg->value);
+			foreach ($limitType->getConstantScalarValues() as $limit) {
+				if (!is_int($limit)) {
+					return new ErrorType();
```
staabm, line 82: this case is missing a test

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+
+			return TypeCombinator::union($returnListType, new ConstantBooleanType(false));
+		}
+
+		$resultTypes = [];
+		foreach ($patternConstantTypes as $patternConstantType) {
+			foreach ($subjectConstantTypes as $subjectConstantType) {
+				foreach ($limits as $limit) {
+					foreach ($flags as $flag) {
+						$result = @preg_split($patternConstantType->getValue(), $subjectConstantType->getValue(), $limit, $flag);
```
staabm: use `Strings::split`

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+			return new ErrorType();
+		}
+
+		$limits = [];
+		if ($limitArg === null) {
+			$limits = [-1];
+		} else {
+			$limitType = $scope->getType($limitArg->value);
+			foreach ($limitType->getConstantScalarValues() as $limit) {
+				if (!is_int($limit)) {
```
staabm: numeric-string $limit is not an error

https://3v4l.org/JuFHj

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
staabm: this needs to check all patterns not only the first

https://3v4l.org/495b1

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+			}
+		}
+
+		$flags = [];
+		if ($flagArg === null) {
+			$flags = [0];
+		} else {
+			$flagType = $scope->getType($flagArg->value);
+			foreach ($flagType->getConstantScalarValues() as $flag) {
+				if (!is_int($flag)) {
```
staabm: to be consistent with limit, this might also allow numeric-string

https://3v4l.org/PqFaA

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+						}
+
+						$resultTypes[] = $constantArray->getArray();
+					}
+				}
+			}
 		}
 
-		return null;
+		return TypeCombinator::union(...$resultTypes);
```
staabm, line 184: I think we are missing `false`. in the `preg_match` inference we decided this can get false even if all args a valid and static analysis time known, because a regex pattern might be super inefficient (or pattern based attacks might trick the regex engine into return false)

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+			$flagType = $scope->getType($flagArg->value);
+			foreach ($flagType->getConstantScalarValues() as $flag) {
+				if (!is_int($flag)) {
+					return new ErrorType();
+				}
+				$flags[] = $flag;
+			}
+		}
+
+		if (count($patternConstantTypes) === 0 || count($subjectConstantTypes) === 0) {
```
staabm: this if-branch might be factored out into a private method for readability

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +47,131 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+		}
+
+		$resultTypes = [];
+		foreach ($patternConstantTypes as $patternConstantType) {
+			foreach ($subjectConstantTypes as $subjectConstantType) {
+				foreach ($limits as $limit) {
+					foreach ($flags as $flag) {
+						$result = @preg_split($patternConstantType->getValue(), $subjectConstantType->getValue(), $limit, $flag);
+						if ($result === false) {
+							continue;
```
staabm: if one of the static analysis time values make `preg_split` return `false` we should give-up instead of ignoring this fact

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
malsuke: I think using Strings::split here is not right because the limit is fixed to -1.

malsuke:
@staabm 
Thanks for your review.
I've made revisions based on the feedback. Please review again.

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +49,156 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
-		if ($flagsArg !== null && $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagsArg->value, $scope, 'PREG_SPLIT_OFFSET_CAPTURE')->yes()) {
-			$type = new ArrayType(
-				new IntegerType(),
-				new ConstantArrayType([new ConstantIntegerType(0), new ConstantIntegerType(1)], [new StringType(), IntegerRangeType::fromInterval(0, null)], [2], [], TrinaryLogic::createYes()),
+		$limits = [];
+		if ($limitArg === null) {
+			$limits = [-1];
+		} else {
+			$limitType = $scope->getType($limitArg->value);
+			if (!$limitType->isInteger()->yes() && !$limitType->isString()->yes()) {
```
staabm: I think this won't work for a limit like `5|'17'`. could work with `->toInteger()` beforehand.

same for $flag

please add tests for this cases

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +49,156 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+				[$returnStringType, IntegerRangeType::fromInterval(0, null)],
+				[2],
+				[],
+				TrinaryLogic::createYes(),
 			);
-			return TypeCombinator::union(TypeCombinator::intersect($type, new AccessoryArrayListType()), new ConstantBooleanType(false));
+
+			$returnInternalValueType = $returnStringType;
+			if ($flagArg !== null) {
+				$flagState = $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagArg->value, $scope, 'PREG_SPLIT_OFFSET_CAPTURE');
```
staabm:
```suggestion
				$capturesOffset = $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagArg->value, $scope, 'PREG_SPLIT_OFFSET_CAPTURE');
```

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +49,156 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+				if (!is_int($flag) && !is_numeric($flag)) {
+					return new ErrorType();
+				}
+				$flags[] = $flag;
+			}
+		}
+
+		if ($this->isPatternOrSubjectEmpty($patternConstantTypes, $subjectConstantTypes)) {
+			$returnNonEmptyStrings = $flagArg !== null && $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagArg->value, $scope, 'PREG_SPLIT_NO_EMPTY')->yes();
+			if ($returnNonEmptyStrings) {
```
staabm: I would inline this only once used variable to ease reading the code

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
staabm: I think the above comment is still true and we are missing the `false` here

staabm (changes requested):
> So, I set __benevolent<list<string>|list<array{string, int<0, max>}>|false> to the basic return type.

it seems this is no longer true

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
malsuke: So, does this mean that every possible result of preg_split includes the possibility of false, and therefore we need to add false to the union type? 

I had implemented it to return an Error if preg_split returns false, as a warning.   
So, does this mean I should include false in all cases, instead of returning Error?

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
malsuke: I realized that using toInteger() beforehand wasn't ideal because it would convert array, null or some inputs into int(0), which is incorrect.  So, I added a check using isConstantScalarValue() to ensure we can accurately handle values like 5|'14'.

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
staabm: The return type of `false` at runtime is necessary because the `preg_split` call can fail even if we know everything IIRC. 

The current "return ErrorType" could be turned into "return null" in case other rules will already report a phpstan error for the code examples.

malsuke: I have fixed about that in the following commit.
307cf54ac64f11d2f7c8b95a7ada046782aa912f

Additionally, since handling for the false case is no longer necessary, I have removed `if ($result === false)`.

yuk1-ishida:
Sorry for the delay since the last commit. Could you please review it again?
@VincentLanglet @ondrejmirtes

malsuke:
Sorry for the delay since the last commit. Could you please review it again?
@VincentLanglet @ondrejmirtes

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +49,161 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+			Strings::match('', $pattern);
+		} catch (RegexpException) {
+			return false;
+		}
+		return true;
+	}
+
+	private function isIntOrStringValue(Type $type): bool
+	{
+		return $type->isInteger()->yes() || $type->isString()->yes() || $type->isConstantScalarValue()->yes();
```
ondrejmirtes: This logic will fail for `int|string`. The proper way to ask this is `(new UnionType([new IntegerType(), new StringType()]))->isSuperTypeOf($type)`. Also remove the `isConstantScalarValue` part (there non-int/non-string scalars as well).

ondrejmirtes (changes requested):
Other than this I'm waiting for a final LGTM from @staabm  😊

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +49,161 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+			} else {
+				$returnStringType = new StringType();
+			}
 
-		if ($flagsArg !== null && $this->bitwiseFlagAnalyser->bitwiseOrContainsConstant($flagsArg->value, $scope, 'PREG_SPLIT_OFFSET_CAPTURE')->yes()) {
-			$type = new ArrayType(
-				new IntegerType(),
-				new ConstantArrayType([new ConstantIntegerType(0), new ConstantIntegerType(1)], [new StringType(), IntegerRangeType::fromInterval(0, null)], [2], [], TrinaryLogic::createYes()),
+			$capturedArrayType = new ConstantArrayType(
+				[new ConstantIntegerType(0), new ConstantIntegerType(1)],
```
ondrejmirtes: Please use ConstantArrayTypeBuilder instead of new ConstantArrayType.

ondrejmirtes (changes requested):

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
```diff
@@ -36,17 +49,161 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+					return new ErrorType();
+				}
+			}
+		}
+
+		$subjectType = $scope->getType($subjectArg->value);
+		$subjectConstantTypes = $subjectType->getConstantStrings();
+
+		$limits = [];
+		if ($limitArg === null) {
```
staabm, line 74: similar to https://github.com/phpstan/phpstan-src/pull/3970/files we should not only handle a absent argument, but also a explicit `null` passed:

e.g. `preg_split($pattern, $subject, null, PREG_SPLIT_NO_EMPTY);`

staabm (approved):

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
yuk1-ishida: fixed 1ee1a792c4d4d681d74e52724e9ccd907f02116b

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
yuk1-ishida: fixed 1ee1a792c4d4d681d74e52724e9ccd907f02116b

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
malsuke: fixed: 1ee1a792c4d4d681d74e52724e9ccd907f02116b

src/Type/Php/PregSplitDynamicReturnTypeExtension.php
malsuke: fixed: 1ee1a792c4d4d681d74e52724e9ccd907f02116b

ondrejmirtes:
Thank you!

PR #3760, opened 2024-12-29
Implement `ArrayAccess->offsetExists` narrowing

staabm:
closes https://github.com/phpstan/phpstan/issues/3323 as described [in comment](https://github.com/phpstan/phpstan/issues/3323#issuecomment-637799286)

### DIFF
--- a/src/Type/Php/ArrayAccessOffsetExistsMethodTypeSpecifyingExtension.php
+++ b/src/Type/Php/ArrayAccessOffsetExistsMethodTypeSpecifyingExtension.php
@@ -1,0 +1,88 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use ArrayAccess;
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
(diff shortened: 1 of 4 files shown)


src/Type/Php/ArrayAccessOffsetExistsMethodTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,96 @@
...
+				continue;
+			}
+
+			return $this->typeSpecifier->create(
+				$node->var,
+				new HasOffsetValueType($keyType, $arrayAccessGenericTypes[1]),
+				$context,
+				$scope,
+			);
+		}
```
staabm: reading the "raw" array-access generic type, because `ObjectType->getValueType` will append `null` for possibly non-existant offsets. the added `null` makes it impossible for us to differentiate a nullable-typed generic from a non-nullable generic.

https://3v4l.org/2uhaE

---

I guess this logic should be moved into the type-system to make it also work for [`isset()` on `ArrayAccess` objects](https://phpstan.org/r/9ffb60c5-c6b2-460d-a2af-7708c3c70004)

PR #3761, opened 2024-12-30
Remove duplicated PHPDoc from InternalScopeFactory classes

herndlm:
was looking through old PRs of mine and did this once in something that had to be reverted again. I guess it does make sense to not duplicate this if not needed, right?


ondrejmirtes:
Thank you!

PR #3762, opened 2024-12-30
Improve expression resolving of superglobals

herndlm:
⚠️ This is a re-do of https://github.com/phpstan/phpstan-src/pull/2012 which caused performance issues. ⚠️

UPDATE: perf. testing is less relevant, the solution was adapted to lazily add global expression types which should avoid performance issues


ondrejmirtes:
I'm not a huge fan of this approach. It's going to pollute the output of `MutatingScope::debug()` (https://phpstan.org/r/e037be34-122b-4846-82f4-487dc0963845) and make it unreadable. Also the values are going to be in the scope and they are going to be mostly the same for all modern apps which isn't great.

Could we add these to the scope only when they are narrowed by some assignments or conditions? I think we already do a similar thing for `PHP_VERSION_ID` (maybe all constants). Thanks.

herndlm:
> Could we add these to the scope only when they are narrowed by some assignments or conditions? I think we already do a similar thing for `PHP_VERSION_ID` (maybe all constants). Thanks.

smart, why haven't I thought about this :) I pushed something, not sure if it's going in the right direction or not though.

herndlm:
I have a feeling that the stuff accessing `Scope::SUPERGLOBAL_VARIABLES` and `MutatingScope::isGlobalVariable()` as well as the looping in `getSuperglobalTypes()` and `getNativeSuperglobalTypes()` can be somehow improved still. I could try refactoring it.

src/Analyser/MutatingScope.php
```diff
@@ -5807,6 +5810,36 @@ public function getConstantReflection(Type $typeWithConstant, string $constantNa
...
+				continue;
+			}
+
+			$superglobalTypes[$exprString] = $typeHolder;
+		}
+		return $superglobalTypes;
+	}
+
+	/** @return array<string, ExpressionTypeHolder> */
+	private function getNativeSuperglobalTypes(): array
```
staabm, line 6009: you could de-duplicate this method by passing in the expressionTypes via parameter

herndlm: right. I stole this basically from `getNativeConstantTypes()` vs `getConstantTypes()` which could be adapted in a similar way then. maybe in a follow-up refactor :)

src/Analyser/MutatingScope.php
```diff
@@ -3272,6 +3273,8 @@ public function enterNamespace(string $namespaceName): self
 			$this->isDeclareStrictTypes(),
 			null,
 			$namespaceName,
+			$this->getSuperglobalTypes(),
+			$this->getNativeSuperglobalTypes(),
 		);
```
herndlm, line 3318: I think this would not be needed anymore now

herndlm: on the other hand.. removing this would be wrong

herndlm:
https://github.com/phpstan/phpstan-src/pull/3871 is doing this in a better way. checked with the tests here which work there as well (except 2 edge cases I'm not sure and are most likely not worth to pursue)

PR #3763, opened 2025-01-01
Fix analysis on `array_map` with named arguments

takaram:
Closes phpstan/phpstan#12317

### DIFF
--- a/src/Parser/ArrayMapArgVisitor.php
+++ b/src/Parser/ArrayMapArgVisitor.php
@@ -6,7 +6,6 @@ use Override;
 use PhpParser\Node;
 use PhpParser\NodeVisitorAbstract;
 use PHPStan\DependencyInjection\AutowiredService;
-use function array_splice;
 use function count;
 
 #[AutowiredService]
@@ -27,14 +26,31 @@ final class ArrayMapArgVisitor extends NodeVisitorAbstract
 			return null;
 		}
 
-		$callbackPos = 0;
-		if ($args[1]->name !== null && $args[1]->name->name === 'callback') {
-			$callbackPos = 1;
+		$callbackArg = null;
+		$arrayArgs = [];
+		foreach ($args as $i => $arg) {
+			if ($callbackArg === null) {
+				if ($arg->name === null && $i === 0) {
+					$callbackArg = $arg;
+					continue;
+				}
+				if ($arg->name !== null && $arg->name->toString() === 'callback') {
+					$callbackArg = $arg;
+					continue;
+				}
+			}
+
+			$arrayArgs[] = $arg;
 		}
-		$callbackArg = $args[$callbackPos];
-		$arrayArgs = $args;
-		array_splice($arrayArgs, $callbackPos, 1);
-		$callbackArg->value->setAttribute(self::ATTRIBUTE_NAME, $arrayArgs);
+
+		if ($callbackArg !== null) {
+			$callbackArg->value->setAttribute(self::ATTRIBUTE_NAME, $arrayArgs);
+			return new Node\Expr\FuncCall(
+				$node->name,
+				[$callbackArg, ...$arrayArgs],
+				$node->getAttributes(),
+			);
+		}
 
 		return null;
 	}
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Functions/data/bug-12317.php
```diff
@@ -0,0 +1,26 @@
...
+	 * @param list<Uuid> $a
+	 *
+	 * @return list<string>
+	 */
+	public function sayHello(array $a): array
+	{
+		$b = array_map(
+			array: $a,
+			callback: static fn(Uuid $c): string => (string) $c,
+		);
```
staabm: please add another test like:

```
$b = array_map(
			callback: static fn(Uuid $c): string => (string) $c,
			array: $a,
		);
```

(parameters in different order, which is allowed for named arguments).

----

I think there is a another case with more then 2 args which needs testing

staabm:
it might make sense to re-use [ArgumentsNormalizer](https://github.com/phpstan/phpstan-src/blob/068be33ffe572a64647d369e7b63115b58ef1d40/src/Analyser/ArgumentsNormalizer.php)  for this fix

ondrejmirtes:
I wouldn't probably do that, we could spin in an infinite recursion maybe.

tests/PHPStan/Rules/Functions/data/bug-12317.php
takaram: I added some more test cases.

takaram:
I'm not sure why integration test fails... `make tests-integration` succeeds on my machine.

src/Parser/ArrayMapArgVisitor.php
```diff
@@ -14,19 +14,41 @@ final class ArrayMapArgVisitor extends NodeVisitorAbstract
...
+		$args = $node->getArgs();
+		if (count($args) < 2) {
+			return null;
+		}
+
+		$callbackPos = 0;
+		if ($args[1]->name !== null && $args[1]->name->name === 'callback') {
+			$callbackPos = 1;
+		}
+		[$callback] = array_splice($args, $callbackPos, 1);
```
ondrejmirtes: I have no idea what's going on at this line. Please make the code easier to read.

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
```diff
@@ -1936,4 +1936,15 @@ public function testBug12051(): void
...
 
+	public function testBug12317(): void
+	{
+		if (PHP_VERSION_ID < 80000) {
+			$this->markTestSkipped('Test requires PHP 8.0.');
+		}
+
+		$this->checkExplicitMixed = true;
+		$this->checkImplicitMixed = true;
+		$this->analyse([__DIR__ . '/data/bug-12317.php'], []);
```
ondrejmirtes: Would be a good idea to add an example that actually produces an error, to verify the args still look like they should.

ondrejmirtes (changes requested):
What we typically do to normalize args like this is to:

1) Add `$callback = null; $array = null;`.
2) Iterate over `$args` and assign these variables either based on `$i` counter in the foreach, or by arg name.
3) Recreate `$args` by doing `[$callback, $array]` in case both of them were found.

Thanks.

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
takaram: Thank you for your suggestion. I added examples and fixed a bug caught by them.

src/Parser/ArrayMapArgVisitor.php
takaram: I made some refactors.

ondrejmirtes (changes requested):
I'm sorry but it seems like you haven't followed my comment (https://github.com/phpstan/phpstan-src/pull/3763#pullrequestreview-2527283982) at all. There's no foreach in the diff so it's wrong. Code using array_slice/splice is very hard to read and I don't know if it's correct.

Also I'd like the array from the visitor to always look the same way (callback first, array 2nd), so that no other code has to be adjusted for named arguments.

takaram:
@ondrejmirtes 
I fix the code to follow your comment.
I'm sorry for the big delay.

ondrejmirtes:
Thank you!

PR #3768, opened 2025-01-04
Fix test

herndlm:
borked this in the last minute commits of https://github.com/phpstan/phpstan-src/pull/3742

thx @staabm


ondrejmirtes:
Thank you.

PR #3769, opened 2025-01-04
NodeScopeResolver: 10x Faster constant array processing

staabm:
`vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php --filter testBug11913` triggers the by-far slowest unit test in the phpstan-src codebase

before this PR it takes 10-11 seconds
after this PR it finishes in 0.3-0.4 seconds
(run on macbook pro m4 with PHP 8.3.15)

---

the code is dominated by calculating intersections in https://github.com/phpstan/phpstan-src/blob/cdf51107e19c8e90a3b64616eaf9bd60d7d8ae5b/src/Analyser/NodeScopeResolver.php#L5783

my thesis is, that when we know a array is constant, it will always intersect with `new ArrayType(new MixedType(), new MixedType())` and therefore don't need to take the slow path

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5771,7 +5771,10 @@ final class NodeScopeResolver
 		foreach (array_reverse($offsetTypes) as $i => $offsetType) {
 			/** @var Type $offsetValueType */
 			$offsetValueType = array_pop($offsetValueTypeStack);
-			if (!$offsetValueType instanceof MixedType) {
+			if (
+				!$offsetValueType instanceof MixedType
+				&& !$offsetValueType->isConstantArray()->yes()
+			) {
 				$types = [
 					new ArrayType(new MixedType(), new MixedType()),
 					new ObjectType(ArrayAccess::class),


ondrejmirtes (changes requested):
1) Is this relevant on 1.12.x as well?
2) What about general array?
3) What I'd like to attempt is doing `TypeCombinator::union(...$types)->isSuperTypeOf($offsetValueType)` and if it's not a yes() only then run the intersect(). Basically we want to run it only if it's not an array, an array access...

staabm:
> * Is this relevant on 1.12.x as well?
> 

the reproducer is equal slow on 1.12.x, but it does not get faster with this patch (bottleneck on 1.12.x is elsewhere):

1.12.x profile of `bug-11913.php`

grafik

---

> * What about general array?

I think general arrays are less of interesst perf wise because they usually don't tend to get equal huge as constant arrays.

using `&& !$offsetValueType->isArray()->yes()`
instead of `&& !$offsetValueType->isConstantArray()->yes()`

is equal fast but yields the following regressions:
```
1) PHPStan\Analyser\Bug9307Test::testFileAsserts with data set "/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-9307.php:34" ('type', '/Users/staabm/workspace/phpst...07.php', 'array', 'array<*ERROR*>', 34)
Expected type array, got type array<*ERROR*> in /Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/data/bug-9307.php on line 34.
--- Expected
+++ Actual
@@ @@
-'array'
+'array<*ERROR*>'

/Users/staabm/workspace/phpstan-src/src/Testing/TypeInferenceTestCase.php:124
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/Bug9307Test.php:25

2) PHPStan\Rules\Functions\CallToFunctionParametersRuleTest::testBug2911
--- Expected
+++ Actual
@@ @@
-'23: Parameter #1 $array of function Bug2911\bar expects array{bar: string}, non-empty-array given.
+'23: Parameter #1 $array of function Bug2911\bar expects array{bar: string}, non-empty-array<mixed> given.
 '

/Users/staabm/workspace/phpstan-src/src/Testing/RuleTestCase.php:161
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php:1280

3) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "tests/PHPStan/Analyser/nsrt/bug-11518-types.php" ('/Users/staabm/workspace/phpst...es.php')
Failed assertions in /Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/nsrt/bug-11518-types.php:

Line 15:
Expected: non-empty-array&hasOffsetValue('thing', 'bla')
Actual:   non-empty-array<mixed>&hasOffsetValue('thing', 'bla')

/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/NodeScopeResolverTest.php:265

4) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "tests/PHPStan/Analyser/nsrt/bug-2911.php" ('/Users/staabm/workspace/phpst...11.php')
Failed assertions in /Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/nsrt/bug-2911.php:

Line 137:
Expected: non-empty-array&hasOffsetValue('bar', 'string')
Actual:   non-empty-array<mixed>&hasOffsetValue('bar', 'string')
```

---
 
> * What I'd like to attempt is doing `TypeCombinator::union(...$types)->isSuperTypeOf($offsetValueType)` and if it's not a yes() only then run the intersect(). Basically we want to run it only if it's not an array, an array access...

tried it, but this did not work. it prevents the slow `TypeCombinator::intersect` but the newly added `TypeCombinator::union(...$types)->isSuperTypeOf($offsetValueType)` is equal slow to the `intersect` call which is prevented

src/Analyser/NodeScopeResolver.php
```diff
@@ -5771,7 +5771,10 @@ private function produceArrayDimFetchAssignValueToWrite(array $offsetTypes, Type
 		foreach (array_reverse($offsetTypes) as $i => $offsetType) {
 			/** @var Type $offsetValueType */
 			$offsetValueType = array_pop($offsetValueTypeStack);
-			if (!$offsetValueType instanceof MixedType) {
+			if (
+				!$offsetValueType instanceof MixedType
+				&& !$offsetValueType->isConstantArray()->yes()
```
mvorisek, line 5776: would be good to comment this is for performance only

ondrejmirtes:
Alright, let's try this! Thank you.

PR #3770, opened 2025-01-04
Support tagged unions in `array_merge`

herndlm:
lucky find combined with some cleanups :)

Closes https://github.com/phpstan/phpstan/issues/7857
Closes https://github.com/phpstan/phpstan/issues/8632
Closes https://github.com/phpstan/phpstan/issues/9559

### DIFF
--- a/src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php
@@ -5,13 +5,14 @@ namespace PHPStan\Type\Php;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\ShouldNotHappenException;
+use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantIntegerType;
+use PHPStan\Type\Constant\ConstantStringType;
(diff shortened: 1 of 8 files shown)


src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php
```diff
@@ -138,4 +135,29 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+		foreach ($keyTypes as $keyType) {
+			$builder->setOffsetValueType(
+				$keyType,
+				$constantArrayType->getOffsetValueType($keyType),
+				!$constantArrayType->hasOffsetValueType($keyType)->yes(),
+			);
+		}
+
+		return $builder->getArray();
+	}
```
herndlm: took me a bit to come up with this and I think it's not bad, but it still feels slightly strange to do that. is there maybe a better or more generic way? :D

src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php
```diff
@@ -39,23 +40,25 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
 			if ($arg->unpack) {
-				if ($argType instanceof ConstantArrayType) {
-					$argTypesFound = $argType->getValueTypes();
+				if ($argType->isConstantArray()->yes()) {
+					$argTypesFound = [];
+					foreach ($argType->getConstantArrays() as $constantArray) {
+						foreach ($constantArray->getValueTypes() as $valueType) {
+							$argTypesFound[] = $valueType;
+						}
+					}
```
herndlm, line 52: This block looks still wrong / dangerous. I'll check it later today.

staabm:
Ondrejs law strikes again: get rid of `instanceof *Type` will fix some bugs you didn't think are related :-)

ondrejmirtes:
This looks like it could be relevant on 1.12.x, can you try that? Thanks!

herndlm:
> This looks like it could be relevant on 1.12.x, can you try that? Thanks!

done, you should be able to upmerge this just fine

ondrejmirtes:
Thank you!

PR #3772, opened 2025-01-05
Support arrays with union value-types in implode()

staabm:
Primary motivation: get rid of `instanceof ConstantScalarType`

closes https://github.com/phpstan/phpstan/issues/11854

### DIFF
--- a/src/Type/Php/ImplodeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ImplodeFunctionReturnTypeExtension.php
@@ -6,6 +6,7 @@ use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Internal\CombinationsHelper;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\InitializerExprTypeResolver;
 use PHPStan\Type\Accessory\AccessoryLiteralStringType;
 use PHPStan\Type\Accessory\AccessoryLowercaseStringType;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
@@ -126,6 +127,10 @@ final class ImplodeFunctionReturnTypeExtension implements DynamicFunctionReturnT
 			foreach ($combinations as $combination) {
 				$strings[] = new ConstantStringType(implode($separatorType->getValue(), $combination));
 			}
+		}
+
+		if (count($strings) > InitializerExprTypeResolver::CALCULATE_SCALARS_LIMIT) {
+			return null;
 		}
 
 		return TypeCombinator::union(...$strings);
--- a/src/Type/Php/ImplodeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ImplodeFunctionReturnTypeExtension.php
@@ -4,6 +4,7 @@ namespace PHPStan\Type\Php;
 
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
+use PHPStan\Internal\CombinationsHelper;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\Accessory\AccessoryLiteralStringType;
 use PHPStan\Type\Accessory\AccessoryLowercaseStringType;
@@ -12,7 +13,6 @@ use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
 use PHPStan\Type\Accessory\AccessoryUppercaseStringType;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantStringType;
-use PHPStan\Type\ConstantScalarType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\StringType;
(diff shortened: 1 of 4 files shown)


src/Type/Php/ImplodeFunctionReturnTypeExtension.php
```diff
@@ -115,13 +115,17 @@ private function inferConstantType(ConstantArrayType $arrayType, ConstantStringT
...
+				$constScalars = $valueType->getConstantScalarValues();
+				if (count($constScalars) === 0) {
 					return null;
 				}
-				$arrayValues[] = $valueType->getValue();
+				$arrayValues[] = $constScalars;
 			}
 
-			$strings[] = new ConstantStringType(implode($separatorType->getValue(), $arrayValues));
+			$combinations = CombinationsHelper::combinations($arrayValues);
```
ondrejmirtes, line 132: Everywhere we use `CombinationsHelper::combinations` there's counting of the number of combinations beforehand, and ending early if it'd go over some limit, to prevent performance degradation.

ondrejmirtes (changes requested):

src/Type/Php/ImplodeFunctionReturnTypeExtension.php
staabm: good catch, fixed

src/Type/Php/ImplodeFunctionReturnTypeExtension.php
```diff
@@ -115,13 +116,21 @@ private function inferConstantType(ConstantArrayType $arrayType, ConstantStringT
...
+			foreach ($combinations as $combination) {
+				$strings[] = new ConstantStringType(implode($separatorType->getValue(), $combination));
+			}
+		}
+
+		if (count($strings) > InitializerExprTypeResolver::CALCULATE_SCALARS_LIMIT) {
```
ondrejmirtes, line 138: Ideally you'd count the combinations before generating them, it can explode as well.

ondrejmirtes (changes requested):

src/Type/Php/ImplodeFunctionReturnTypeExtension.php
```diff
@@ -115,13 +117,21 @@ private function inferConstantType(ConstantArrayType $arrayType, ConstantStringT
...
+				$constScalars = $valueType->getConstantScalarValues();
+				if (count($constScalars) === 0) {
 					return null;
 				}
-				$arrayValues[] = $valueType->getValue();
+				$arrayValues[] = $constScalars;
 			}
 
-			$strings[] = new ConstantStringType(implode($separatorType->getValue(), $arrayValues));
+			if (count($strings) + count($arrayValues, COUNT_RECURSIVE) > InitializerExprTypeResolver::CALCULATE_SCALARS_LIMIT) {
```
ondrejmirtes: You're not counting the combinations properly. This is how I verified it:

```diff
diff --git a/src/Type/Php/ImplodeFunctionReturnTypeExtension.php b/src/Type/Php/ImplodeFunctionReturnTypeExtension.php
index 76afc963c..e45e0146a 100644
--- a/src/Type/Php/ImplodeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ImplodeFunctionReturnTypeExtension.php
@@ -22,6 +22,7 @@ use PHPStan\Type\TypeCombinator;
 use function count;
 use function implode;
 use function in_array;
+use function iterator_count;
 use const COUNT_RECURSIVE;
 
 final class ImplodeFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
@@ -124,10 +125,16 @@ final class ImplodeFunctionReturnTypeExtension implements DynamicFunctionReturnT
 				$arrayValues[] = $constScalars;
 			}
 
+			var_dump(count($strings) + count($arrayValues, COUNT_RECURSIVE));
+
 			if (count($strings) + count($arrayValues, COUNT_RECURSIVE) > InitializerExprTypeResolver::CALCULATE_SCALARS_LIMIT) {
 				return null;
 			}
 
+			var_dump(iterator_count(CombinationsHelper::combinations($arrayValues)));
+
+			var_dump('---');
+
 			$combinations = CombinationsHelper::combinations($arrayValues);
 			foreach ($combinations as $combination) {
 				$strings[] = new ConstantStringType(implode($separatorType->getValue(), $combination));

```

The output:

```
int(33)
int(8)
string(3) "---"
int(41)
int(8)
string(3) "---"
```

Expected output:

```
int(8)
int(8)
string(3) "---"
int(8)
int(8)
string(3) "---"
```

ondrejmirtes:
Thank you!

tests/PHPStan/Analyser/nsrt/implode.php
```diff
@@ -21,4 +21,34 @@ public function constants() {
...
+	}
+
+	/** @param array{10: 1|2|3, xy: 'a'|'b'|'c'} $constArr */
+	public function constArrays5($constArr) {
+		assertType("'1a'|'1b'|'1c'|'2a'|'2b'|'2c'|'3a'|'3b'|'3c'", implode('', $constArr));
+	}
+
+	/** @param array{0: 1, 1: 'a'|'b', 3?: 'c'|'d', 4?: 'e'|'f', 5?: 'g'|'h', 6?: 'x'|'y'} $constArr */
+	public function constArrays6($constArr) {
+		assertType("string", implode('', $constArr));
```
staabm, line 52: this too generic type made me think about the general case and triggerd https://github.com/phpstan/phpstan-src/pull/3774

PR #3773, opened 2025-01-05
MinMaxFunctionReturnTypeExtension: Cleanup `instanceof ConstantScalarType`

staabm:
see
- https://3v4l.org/W3jHH
- https://3v4l.org/ZKB9J


staabm:
There are more instanceof checks in here, which I would tackle separately

ondrejmirtes:
Thank you.

PR #3774, opened 2025-01-06
More precise implode() return type

staabm:


staabm:
//cc @VincentLanglet since its string accessory business :)

src/Type/Php/ImplodeFunctionReturnTypeExtension.php
```diff
@@ -80,22 +80,23 @@ private function implode(Type $arrayType, Type $separatorType): Type
 		}
 
 		$accessoryTypes = [];
+		$valueTypeAsString = $arrayType->getIterableValueType()->toString();
```
staabm, line 84: the actual change is here: we use `toString` as the `implode` function internally casts everything.

that way the string types detection also works for unions of strings and non-strings (e.g. `2|'a'`) and so we get more precise results

VincentLanglet (approved):
Awesome

tests/PHPStan/Analyser/nsrt/non-empty-string.php
```diff
@@ -212,7 +212,7 @@ public function sayHello(int $i): void
 		// coming from issue #5291
 		$s = array(1, $i);
 
-		assertType('non-falsy-string', implode("a", $s));
+		assertType('lowercase-string&non-falsy-string', implode("a", $s));
```
mvorisek, line 215:
```suggestion
		assertType('lowercase-string&non-falsy-string', implode("a", $s));
		assertType('uppercase-string&non-falsy-string', implode("A", $s));
```

tests/PHPStan/Analyser/nsrt/implode.php
```diff
@@ -49,6 +61,6 @@ public function constArrays5($constArr) {
 
 	/** @param array{0: 1, 1: 'a'|'b', 3?: 'c'|'d', 4?: 'e'|'f', 5?: 'g'|'h', 6?: 'x'|'y'} $constArr */
 	public function constArrays6($constArr) {
-		assertType("string", implode('', $constArr));
+		assertType("literal-string&lowercase-string&non-falsy-string", implode('', $constArr));
```
staabm: Memo to me: Literal string is wrong here

tests/PHPStan/Analyser/nsrt/non-empty-string-str-containing-fns.php
```diff
@@ -14,6 +14,16 @@ class Foo {
...
 	{
+		if (str_contains($i, 0)) {
+			assertType('int', $i);
+		}
+		if (str_contains($s, 0)) {
+			assertType('non-empty-string', $s);
+		}
+		if (str_contains($s, 1)) {
+			assertType('non-falsy-string', $s);
+		}
```
staabm, line 25: these cases depends on strict-types.. are we still fine with it?

https://3v4l.org/CHDVW

ondrejmirtes:
Thank you!

PR #3777, opened 2025-01-10
Do not report constructor unused parameter if class implements an Interface that defines a constructor

carlos-granados:
If the class implements an interface that defines a constructor, then the class constructor needs to match the signature of the constructor in the interface. In this case we should not report any unused constructor parameters and this is what this PR achieves

Fixes https://github.com/phpstan/phpstan/issues/11454

### DIFF
--- a/src/Rules/Classes/UnusedConstructorParametersRule.php
+++ b/src/Rules/Classes/UnusedConstructorParametersRule.php
@@ -48,6 +48,12 @@ final class UnusedConstructorParametersRule implements Rule
 			return [];
 		}
 
+		foreach ($node->getClassReflection()->getInterfaces() as $interface) {
+			if ($interface->hasMethod('__construct')) {
+				return [];
+			}
+		}
+
 		$message = sprintf(
 			'Constructor of class %s has an unused parameter $%%s.',
 			SprintfHelper::escapeFormatString($node->getClassReflection()->getDisplayName()),
--- a/src/Rules/Classes/UnusedConstructorParametersRule.php
+++ b/src/Rules/Classes/UnusedConstructorParametersRule.php
@@ -16,7 +16,6 @@ use function array_map;
 use function array_values;
 use function count;
 use function sprintf;
-use function strtolower;
 
 /**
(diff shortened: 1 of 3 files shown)


carlos-granados:
Again there is a failing check which seems unrelated to this PR

src/Rules/Classes/UnusedConstructorParametersRule.php
```diff
@@ -45,6 +45,12 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		foreach ($node->getClassReflection()->getInterfaces() as $interface) {
+			if ($interface->hasMethod('__construct')) {
```
ondrejmirtes: `hasConstructor` would be more precise

ondrejmirtes: And above instead of `strtolower($method->getName()) !== '__construct'` we should use `$method->isConstructor()`.

ondrejmirtes (changes requested):

staabm:
this PR made me think of classes which implement more then one interface and multiple `__construct` inheritance

https://3v4l.org/easGk
https://phpstan.org/r/dfe421ff-5ff8-44fc-8f5d-8782ef4ffb57

maybe thats something to loot at next (followup PR) @carlos-granados :) - as we are missing a reported error

src/Rules/Classes/UnusedConstructorParametersRule.php
carlos-granados: @ondrejmirtes rebased and updated

staabm:
we need a similar fix for abstract base-classes

https://phpstan.org/r/2652fd08-bb25-431a-8f0f-b28b3c096dc6
https://3v4l.org/hdDcZ

carlos-granados:
I rebased this branch and fixed the linting issue

ondrejmirtes:
Thank you.

For completeness it should also look for abstract constructor in parent classes and implemented traits.

PR #3778, opened 2025-01-10
More precise return type for mysqli_fetch_all()

oxan:
`mysqli_fetch_all()` and `mysqli_result::fetch_all()` returns a sequentially indexed array (list) of arrays, so convert its return type from `array` to `list<array>`.

The fullly specified return type for these functions is a bit more complicated, as it depends on the `$mode` parameter:
* If it's `MYSQLI_ASSOC`, it's a list of associative arrays, i.e. `list<array<string, string|int|float|null>>`.
* If it's `MYSQLI_NUM`, it's a list of lists, i.e. `list<list<array|int|float|null>>`.
* If it's `MYSQLI_BOTH`, it's a list of enumerated and associative arrays, i.e. `list<array<string|int, string|int|float|null>>`.

We could specify the return type as `list<array<string|int, string|int|float|null>>` to catch all these cases, but I'm not sure that'd be very helpful, as it might force users to deal with types in the union that will never happen in their case. The `mysqli_fetch_array()` and `mysqli_result::fetch_array()` functions that share the same polymorphic behaviour are also typehinted as just `array`.

I think the CI failure is unrelated.


ondrejmirtes:
Thank you!

PR #3780, opened 2025-01-11
StrlenFunctionReturnTypeExtension: Cleanup `instanceof ConstantString`

staabm:


staabm:
~~lets see whether its smarter to resolve it in the heart of the typesystem instead:~~  cleanup after https://github.com/phpstan/phpstan-src/pull/3781

I looked thru the tests.. we have lots of coverage arround `strlen` and `mb_strlen`.

ondrejmirtes:
Thank you!

PR #3781, opened 2025-01-11
More precise constant variants of BooleanType

staabm:
instead of resolving bool into true/false in the extension, lets do it in the type-system itself

### DIFF
--- a/src/Type/BooleanType.php
+++ b/src/Type/BooleanType.php
@@ -41,9 +41,19 @@ class BooleanType implements Type
 	{
 	}
 
+	public function getConstantScalarTypes(): array
+	{
+		return [new ConstantBooleanType(true), new ConstantBooleanType(false)];
+	}
+
 	public function getConstantStrings(): array
 	{
-		return [];
+		return [new ConstantStringType('1'), new ConstantStringType('0')];
+	}
+
+	public function getConstantScalarValues(): array
+	{
+		return [true, false];
 	}
 
 	public function describe(VerbosityLevel $level): string
--- a/src/Type/BooleanType.php
+++ b/src/Type/BooleanType.php
@@ -46,11 +46,6 @@ class BooleanType implements Type
 		return [new ConstantBooleanType(true), new ConstantBooleanType(false)];
 	}
 
-	public function getConstantStrings(): array
-	{
-		return [new ConstantStringType('1'), new ConstantStringType('0')];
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/nsrt/bug-11201.php
```diff
@@ -53,4 +53,4 @@ function returnsBool(): bool {
 assertType("'                   1'", $s);
 
 $s = sprintf('%20s', returnsBool());
-assertType("lowercase-string&non-falsy-string", $s);
+assertType("'                    '|'                   1'", $s);
```
staabm, line 56: looks great, see https://3v4l.org/ipPIP

staabm:
@herndlm opinions?

herndlm:
Interesting idea! Makes sense IMO for "finite" types like that

herndlm:
Interesting idea! Makes sense IMO for "finite" types like that

staabm:
> "finite" types like that

hmm interessting wording you use here. makes me wonder what the semantic difference of

`getFiniteTypes` and `getConstantScalarTypes` is .. maybe just "enum", integer-range?

herndlm:
I might be also just mixing up things. Don't even know all those related methods that currently exist on the interface. But good point 🤔

herndlm:
Quickly checked, looks like you could implement that finite method as well here

staabm:
> Quickly checked, looks like you could implement that finite method as well here

its [already implemened](https://github.com/phpstan/phpstan-src/blob/eb0e0bcfe2e4947d06c5eb680f5cf568a688ff4a/src/Type/BooleanType.php#L148-L154)

src/Type/BooleanType.php
```diff
@@ -41,9 +41,19 @@ public function __construct()
...
 
+	public function getConstantScalarTypes(): array
+	{
+		return [new ConstantBooleanType(true), new ConstantBooleanType(false)];
+	}
+
 	public function getConstantStrings(): array
 	{
-		return [];
+		return [new ConstantStringType('1'), new ConstantStringType('0')];
```
ondrejmirtes: Nope. This function isn't used for loose type casts like that.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank yuo.

PR #3784, opened 2025-01-16
Fix negative offset false positive on constant string

staabm:
closes https://github.com/phpstan/phpstan/issues/12122

### DIFF
--- a/src/Type/Constant/ConstantStringType.php
+++ b/src/Type/Constant/ConstantStringType.php
@@ -366,7 +366,7 @@ class ConstantStringType extends StringType implements ConstantScalarType
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
 		if ($offsetType->isInteger()->yes()) {
-			$strLenType = IntegerRangeType::fromInterval(0, strlen($this->value) - 1);
+			$strLenType = IntegerRangeType::fromInterval(-strlen($this->value), strlen($this->value) - 1);
 			return $strLenType->isSuperTypeOf($offsetType);
 		}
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
The bug report is about an error, the regression test should be in RuleTestCase.

src/Type/Constant/ConstantStringType.php
```diff
@@ -376,15 +376,16 @@ public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	public function getOffsetValueType(Type $offsetType): Type
 	{
 		if ($offsetType->isInteger()->yes()) {
+			$strLenType = IntegerRangeType::fromInterval(-strlen($this->value), strlen($this->value) - 1);
+
 			if ($offsetType instanceof ConstantIntegerType) {
-				if ($offsetType->getValue() < strlen($this->value)) {
+				if ($strLenType->isSuperTypeOf($offsetType)->yes()) {
 					return new self($this->value[$offsetType->getValue()]);
```
staabm, line 385: this change also fixes a php warning which could happen on too small negative offsets (covered by the tests)

src/Type/Constant/ConstantStringType.php
```diff
@@ -366,7 +366,7 @@ public function isUppercaseString(): TrinaryLogic
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
 		if ($offsetType->isInteger()->yes()) {
-			$strLenType = IntegerRangeType::fromInterval(0, strlen($this->value) - 1);
+			$strLenType = IntegerRangeType::fromInterval(-strlen($this->value), strlen($this->value) - 1);
```
ondrejmirtes: You know you could ask for strlen() just once :)

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #3660, opened 2024-11-23
Support BcMath\Number operator overloading

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/12099

PR #3787, opened 2025-01-19
Handle BcMath\Number operators for simple cases

schlndh:
Partially fixes: https://github.com/phpstan/phpstan/issues/12099
Fixes: https://github.com/phpstan/phpstan/issues/7937, https://github.com/phpstan/phpstan/issues/8555
It's loosely based on #3660
Here is a demonstration of how the operators behave with various types: https://3v4l.org/jYX5J#v8.4.3

I skipped the complicated stuff. The main issue is unions. IMO it's too difficult to handle unions correctly in the extension. It would probably make more sense for PHPStan to handle the unions itself and offload only non-unions to extensions (e.g. `X|int + X|float` needs to know what `int + float` is).

I also ignored a few preexisting bugs/inconsistencies (null and never types). They can be solved later.


src/Php/PhpVersion.php
```diff
@@ -395,4 +395,9 @@ public function substrReturnFalseInsteadOfEmptyString(): bool
 		return $this->versionId < 80000;
 	}
 
+	public function supportsBcMathNumberOperatorOverloading(): bool
```
schlndh, line 403: I copied this from the original PR. But I'm not sure whether it's correct. I assume that the reason we even need this is because earlier PHP versions might use a polyfill (which wouldn't have the operators overloaded). But PHP 8.4 can also use a polyfill, because bcmath is an optional extension.

tests/PHPStan/Analyser/nsrt/pow.php
```diff
@@ -47,11 +47,11 @@ function (): void {
...
 
-	assertType('int<4, 27>|int<16, 81>', pow($range, $x));
-	assertType('int<4, 27>|int<16, 81>', $range ** $x);
+	assertType('int<4, 81>', pow($range, $x));
+	assertType('int<4, 81>', $range ** $x);
 
-	assertType('int<4, 27>|int<16, 64>', pow($x, $range));
-	assertType('int<4, 27>|int<16, 64>', $x ** $range);
+	assertType('int<4, 64>', pow($x, $range));
+	assertType('int<4, 64>', $x ** $range);
```
schlndh: This is a result of `UnionType::getSortedTypes` being called. ~~I haven't looked into it, but I guess it's a bug in `TypeCombinator::union`, because the result of a union shouldn't depend on the order of parameters.~~

EDIT: I had to fix it, because it broke the test on PHP <8.4 due to the extension only running on 8.4 and above.

tests/PHPStan/Rules/Operators/data/bcmath-number.php
```diff
@@ -0,0 +1,406 @@
...
+		assertType('*ERROR*', $a ^ $b);
+		assertType('*ERROR*', $a | $b);
+		assertType('bool', $a && $b);
+		assertType('true', $a || $b);
+		assertType('bool', $a and $b);
+		assertType('bool', $a xor $b);
+		assertType('true', $a or $b);
+	}
+
+	public function bcVsNever(Number $a): void
```
schlndh: There's a bunch of inconsistencies in how never type is handled. I assume that all of these operations should result in `*NEVER*`, but I'm ignoring it for this PR.

tests/PHPStan/Type/TypeCombinatorTest.php
```diff
@@ -1558,6 +1558,34 @@ public function dataUnion(): iterable
...
 				'int<1, 3>|int<7, 9>',
 			],
+			[
+				[
+					IntegerRangeType::fromInterval(4, 9),
+					IntegerRangeType::fromInterval(16, 81),
+					IntegerRangeType::fromInterval(8, 27),
+				],
+				IntegerRangeType::class,
+				'int<4, 81>',
```
schlndh: Was: `int<4, 27>|int<16, 81>`.

tests/PHPStan/Type/TypeCombinatorTest.php
```diff
@@ -1558,6 +1558,34 @@ public function dataUnion(): iterable
...
+			],
+			[
+				[
+					IntegerRangeType::fromInterval(8, 27),
+					IntegerRangeType::fromInterval(4, 6),
+					new ConstantIntegerType(7),
+					IntegerRangeType::fromInterval(16, 81),
+				],
+				IntegerRangeType::class,
+				'int<4, 81>',
```
schlndh: Was: `int<4, 7>|int<8, 81>` (testUnionInversed)

tests/PHPStan/Type/TypeCombinatorTest.php
```diff
@@ -1558,6 +1558,34 @@ public function dataUnion(): iterable
...
+			],
+			[
+				[
+					new IntegerType(),
+					IntegerRangeType::fromInterval(null, -1),
+					IntegerRangeType::fromInterval(1, null),
+				],
+				IntegerType::class,
+				'int',
```
schlndh: Was: `int|int<min, -1>` (testUnionInversed)

ondrejmirtes:
Hi, I'm sorry for taking such a long time to review this. I've separated part of this PR (https://github.com/phpstan/phpstan-src/commit/253903a83bc6a426ef464cee051dc465029b8f4d) because it's not really related to BcMath.

src/Php/PhpVersion.php
ondrejmirtes: I think it's okay. If someone uses operators with BCMath objects, we can safely assume they assume they have the extension.

tests/PHPStan/Rules/Operators/InvalidBinaryOperationRuleTest.php
```diff
@@ -303,6 +303,528 @@ public function testBug5309(): void
...
 
+	public function testBcMathNumber(): void
+	{
+		if (PHP_VERSION_ID < 80400) {
+			self::markTestSkipped('Test requires PHP 8.4.');
+		}
+
+		$this->analyse([__DIR__ . '/data/bcmath-number.php'], [
+			[
+				'Binary operation "<<" between BcMath\\Number and BcMath\\Number results in an error.',
```
ondrejmirtes: Personally I'd remove this method altogether (testBcMathNumber). We don't need to test all of these situations "results in an error" because presumably it's handled by a single code path. If you still want to test this error, please do it with a separate test file only with a single or a few expressions.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Operators/InvalidBinaryOperationRuleTest.php
schlndh: You're right - it's sufficiently covered by NSRT. I removed the test case from InvalidBinaryOperationRuleTest and moved the file to NSRT directory.

ondrejmirtes:
Thank you!

PR #3788, opened 2025-01-19
Fix ImpossibleCheckTypeFunctionCallRule for `is_subclass` and `is_a`

VincentLanglet:
Improved strategy for https://github.com/phpstan/phpstan-src/pull/3404

The main idea is that `always true` error can always be reported if the second param is a Constant class-string (or union of them).

I included a non regression test from
https://github.com/phpstan/phpstan-src/pull/3404#issuecomment-2600896183

Closes https://github.com/phpstan/phpstan/issues/3979
Closes https://github.com/phpstan/phpstan/issues/8464

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeFunctionCallRule.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeFunctionCallRule.php
@@ -8,7 +8,6 @@ use PHPStan\Parser\LastConditionVisitor;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use function sprintf;
-use function strtolower;
 
 /**
  * @implements Rule<Node\Expr\FuncCall>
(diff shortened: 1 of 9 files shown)


gnutix:
I don't know if this PR would fix this, but I just upgraded to the latest PHPStan version and this new error popped up: 

```
Call to method ReflectionClass<object>::isSubclassOf() with 'App\\Application\\Shared\\
Query\\Filter\\ListQuery' will always evaluate to false.
```

for this code :

```php
#[Override]
    public function supports(Model $model): bool
    {
        $type = $model->getType();
        if ('object' !== $type->getBuiltinType()) {
            return false;
        }

        $class = $type->getClassName();
        if (null === $class || !class_exists($class)) {
            return false;
        }

        $reflection = new ReflectionClass($class);

        return $reflection->isSubclassOf(ListQuery::class) ? ... : ...;
```

Should I open a separate bug?

ondrejmirtes:
@gnutix That is unrelated, and most likely a consequence of https://github.com/phpstan/phpstan/issues/12459

VincentLanglet:
Friendly ping @ondrejmirtes ; this PR is ready to be reviewed :)

ondrejmirtes:
Thank you!

PR #3792, opened 2025-01-21
Keep traversing type when we can contain lowercase/upercase strings

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12457

### DIFF
--- a/tests/PHPStan/Rules/PhpDoc/data/bug-12457.php
+++ b/tests/PHPStan/Rules/PhpDoc/data/bug-12457.php
@@ -12,4 +12,13 @@ class HelloWorld
 		/** @var array{numeric-string} $b */
 		$b = $a;
 	}
+
+	/**
+	 * @param callable(): numeric-string&uppercase-string&lowercase-string $a
+	 */
+	public function sayHello2(callable $a): void
+	{
+		/** @var callable(): string $b */
+		$b = $a;
+	}
 }
--- a/tests/PHPStan/Rules/PhpDoc/data/bug-12457.php
+++ b/tests/PHPStan/Rules/PhpDoc/data/bug-12457.php
@@ -1,0 +1,15 @@
+<?php declare(strict_types = 1);
+
+namespace Bug12457;
+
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/PhpDoc/data/bug-12457.php
```diff
@@ -0,0 +1,15 @@
...
+class HelloWorld
+{
+	/**
+	 * @param array{numeric-string&uppercase-string&lowercase-string} $a
+	 */
+	public function sayHello(array $a): void
+	{
+		/** @var array{numeric-string} $b */
+		$b = $a;
+	}
```
staabm, line 14: do we need more tests for different variants of this problem?

like

```suggestion
	}
	
	/**
	 * @param array{numeric-string&uppercase-string&lowercase-string} $a
	 */
	public function sayUpper(array $a): void
	{
		/** @var array{uppercase-string} $b */
		$b = $a;
	}

	/**
	 * @param array{numeric-string&uppercase-string&lowercase-string} $a
	 */
	public function sayLower(array $a): void
	{
		/** @var array{lowercase-string} $b */
		$b = $a;
	}
```

staabm: maybe also variants with `non-empty-string`, `non-falsey-string`?

VincentLanglet: I don't think so since it's not related to numeric-string directly.

The issue was because it was a ConstantArray with a lowercase-string or upercase-string inside.
Because of the early `return $type` we didn't process the content of the array.

VincentLanglet: But I added a test for the `callable` case
https://github.com/phpstan/phpstan-src/pull/3792/commits/ad97094494cfa5ecb6191c3670f5276a4b8e1770

ondrejmirtes:
Thank you!

PR #3794, opened 2025-01-23
Skip function-like and anonymous class nodes when checking if a function or method is a generator

sebkehr:
Fixes: phpstan/phpstan#12462.

As is, when checking if a function (or method) is a generator (via `PhpFunctionFromParserNodeReflection::isGenerator()`) all subnodes are recursively checked for containment of a yield expression. Hence any function or method itself returning a yielding closure or arrow function is falsely determined to be a generator which in turn leads to incorrect return type checks as demonstrated in https://phpstan.org/r/5970e2e2-0aad-4927-8938-3f7d83e73a95.

The proposed fix therefore simply skips `FunctionLike` nodes in the above mentiond recursive check.


src/Reflection/Php/PhpFunctionFromParserNodeReflection.php
```diff
@@ -281,7 +281,10 @@ private function nodeIsOrContainsYield(Node $node): bool
 		foreach ($node->getSubNodeNames() as $nodeName) {
 			$nodeProperty = $node->$nodeName;
 
-			if ($nodeProperty instanceof Node && $this->nodeIsOrContainsYield($nodeProperty)) {
+			if ($nodeProperty instanceof Node &&
+				!$nodeProperty instanceof FunctionLike &&
```
ondrejmirtes, line 292: Would be great to also not descend into anonymous classes.

ondrejmirtes (changes requested):
Please also submit the same fix to https://github.com/Roave/BetterReflection: https://github.com/Roave/BetterReflection/blob/d262672652b9b70a54246adadf340587b535f549/src/Reflection/ReflectionFunctionAbstract.php#L349-L380

src/Reflection/Php/PhpFunctionFromParserNodeReflection.php
sebkehr: @ondrejmirtes Anonymous class nodes should now be skipped as well. I submitted the fix via Roave/BetterReflection#1479.

ondrejmirtes (changes requested):
I just realized there's one more place that needs fixing. Check out CleaningParserTest. The job of CleaningParser is to erase function bodies of not-analyzed code (like in vendor/) before the AST is cached into memory.

I worry the code is faulty and it might do something like this:

Before:

```php
function foo() {
    $fn = fn () => yield;
};
```

After:

```php
function foo() {
    yield;
};
```

Changing function foo() from non-generator to a generator.

Thank you!

sebkehr:
> I just realized there's one more place that needs fixing. Check out CleaningParserTest. The job of CleaningParser is to erase function bodies of not-analyzed code (like in vendor/) before the AST is cached into memory.

Ah ok, I'll look into it.

sebkehr:
@ondrejmirtes

I added two more examples to  _tests/PHPStan/Parser/data/cleaning-1-before.php_. Please also have a look on the expected outcome after cleaning in _tests/PHPStan/Parser/data/cleaning-1-after.php_.

With the proposed changes to _src/Parser/CleaningVisitor.php_ the closure in _ContainsClosure::doFoo()_ will now be erased not leading to any left-over yield statements just like the yielding array-function in _ContainsClosure::doBar()_. Is this the intended behavior?

I also added _Baz::someGenerator3()_ containing _yield_ as a subexpression which will now lead to a _yield_ expression after cleaning.

sebkehr:
@ondrejmirtes 

1. I'm not sure what to do about the [failing check](https://github.com/phpstan/phpstan-src/actions/runs/13310904103/job/37173046112?pr=3794) as it doesn't seem to related to my changes.
2. Should we maybe put the changes to CleaningVisitor into a separate PR?

VincentLanglet:
Hi @sebkehr sorry for the delay, are you still interested in this PR ? Could you solve the conflict ?

PR #3795, opened 2025-01-24
Allow write access to offsets of readonly ArrayAccess properties

pilif:
In case a readonly property is an object which implements ArrayAccess, offet accesses to said property are turned into method calls on PHP's side and are allowed even on read-only properties.

this fixes phpstan/phpstan#8929

### DIFF
--- a/src/Rules/Properties/ReadOnlyPropertyAssignRule.php
+++ b/src/Rules/Properties/ReadOnlyPropertyAssignRule.php
@@ -5,6 +5,8 @@ namespace PHPStan\Rules\Properties;
 use ArrayAccess;
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
+use PHPStan\Node\Expr\SetOffsetValueTypeExpr;
+use PHPStan\Node\Expr\UnsetOffsetExpr;
 use PHPStan\Node\PropertyAssignNode;
 use PHPStan\Reflection\ConstructorsHelper;
 use PHPStan\Reflection\MethodReflection;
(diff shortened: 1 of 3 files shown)


pilif:
one sec - will fix to make this work properly with old PHP versions

pilif:
ok. much better.

I don't think the failing rector integration test is related to this at all

src/Rules/Properties/ReadOnlyPropertyAssignRule.php
```diff
@@ -89,6 +91,10 @@ public function processNode(Node $node, Scope $scope): array
 				continue;
 			}
 
+			if ((new ObjectType(ArrayAccess::class))->isSuperTypeOf($propertyReflection->getNativeType())->yes()) {
+				continue;
```
ondrejmirtes, line 101: This is a wrong fix. You can't say that in general.

If I have `private readonly ArrayAcess $a` and I do `$this->a = new Foo()` twice, PHP does not allow that.

ondrejmirtes (changes requested):

src/Rules/Properties/ReadOnlyPropertyAssignRule.php
pilif: 🤦 of course - you're right.

I have pushed a patch to expand the test to test for your case and to also check that we're dealing with a `SetOffsetValueTypeExpr`

herndlm:
Interesting, wasn't aware of this. Looks like my https://github.com/phpstan/phpstan-src/pull/3817 is doing a very similar thing, but we mention different issues that are fixed. Those might be duplicates, but can't check the details right now..

ondrejmirtes:
Thank you. I liked https://github.com/phpstan/phpstan-src/pull/3817 slightly more.

PR #3797, opened 2025-01-25
Native type does not know anything about purity

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12458

### DIFF
--- a/src/Rules/PhpDoc/VarTagTypeRuleHelper.php
+++ b/src/Rules/PhpDoc/VarTagTypeRuleHelper.php
@@ -11,7 +11,6 @@ use PHPStan\PhpDoc\Tag\VarTag;
 use PHPStan\PhpDoc\TypeNodeResolver;
 use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\RuleErrorBuilder;
-use PHPStan\TrinaryLogic;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Generic\GenericObjectType;
 use PHPStan\Type\MixedType;
@@ -141,15 +140,15 @@ final class VarTagTypeRuleHelper
 				$type = new ArrayType(new MixedType(), new MixedType());
 			}
 
-			return !$this->isSuperTypeOfVarType($type, $varTagType, false);
+			return !$this->isAtLeastMaybeSuperTypeOfVarType($type, $varTagType);
 		}
 
 		if ($expr instanceof Expr\ConstFetch) {
-			return !$this->isSuperTypeOfVarType($type, $varTagType, false);
+			return !$this->isAtLeastMaybeSuperTypeOfVarType($type, $varTagType);
 		}
 
 		if ($expr instanceof Node\Scalar) {
-			return !$this->isSuperTypeOfVarType($type, $varTagType, false);
+			return !$this->isAtLeastMaybeSuperTypeOfVarType($type, $varTagType);
 		}
 
 		if ($expr instanceof Expr\New_) {
@@ -164,18 +163,18 @@ final class VarTagTypeRuleHelper
 	private function checkType(Type $type, Type $varTagType, int $depth = 0): bool
 	{
 		if ($this->strictWideningCheck) {
-			return !$this->isSuperTypeOfVarType($type, $varTagType, true);
+			return !$this->isSuperTypeOfVarType($type, $varTagType);
 		}
 
 		if ($type->isConstantArray()->yes()) {
 			if ($type->isIterableAtLeastOnce()->no()) {
 				$type = new ArrayType(new MixedType(), new MixedType());
-				return !$this->isSuperTypeOfVarType($type, $varTagType, false);
+				return !$this->isAtLeastMaybeSuperTypeOfVarType($type, $varTagType);
 			}
 		}
 
 		if ($type->isIterable()->yes() && $varTagType->isIterable()->yes()) {
-			if (!$this->isSuperTypeOfVarType($type, $varTagType, false)) {
+			if (!$this->isAtLeastMaybeSuperTypeOfVarType($type, $varTagType)) {
 				return true;
 			}
 
@@ -183,31 +182,39 @@ final class VarTagTypeRuleHelper
 			$innerVarTagType = $varTagType->getIterableValueType();
 
 			if ($type->equals($innerType) || $varTagType->equals($innerVarTagType)) {
-				return !$this->isSuperTypeOfVarType($innerType, $innerVarTagType, true);
+				return !$this->isSuperTypeOfVarType($innerType, $innerVarTagType);
 			}
 
 			return $this->checkType($innerType, $innerVarTagType, $depth + 1);
 		}
 
 		if ($depth === 0 && $type->isConstantValue()->yes()) {
-			return !$this->isSuperTypeOfVarType($type, $varTagType, false);
+			return !$this->isAtLeastMaybeSuperTypeOfVarType($type, $varTagType);
 		}
 
-		return !$this->isSuperTypeOfVarType($type, $varTagType, true);
+		return !$this->isSuperTypeOfVarType($type, $varTagType);
 	}
 
-	private function isSuperTypeOfVarType(Type $type, Type $varTagType, bool $strict): bool
+	private function isSuperTypeOfVarType(Type $type, Type $varTagType): bool
 	{
-		$validationCallable = static fn (TrinaryLogic $trinaryLogic): bool => $strict ? $trinaryLogic->yes() : !$trinaryLogic->no();
-
-		$result = $type->isSuperTypeOf($varTagType);
-		if ($validationCallable($result)) {
+		if ($type->isSuperTypeOf($varTagType)->yes()) {
 			return true;
 		}
 
 		$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), new NameScope(null, []));
 
-		return $validationCallable($type->isSuperTypeOf($varTagType));
+		return $type->isSuperTypeOf($varTagType)->yes();
+	}
+
+	private function isAtLeastMaybeSuperTypeOfVarType(Type $type, Type $varTagType): bool
+	{
+		if (!$type->isSuperTypeOf($varTagType)->no()) {
+			return true;
+		}
+
+		$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), new NameScope(null, []));
+
+		return !$type->isSuperTypeOf($varTagType)->no();
 	}
 
 }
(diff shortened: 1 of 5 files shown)


ondrejmirtes (changes requested):
I disagree with this on a conceptual level. With a closure like `fn () => sleep(5)`, only native types are involved and we still know it's impure.

With one-off fixes like that, we'd be chaising our own tail for a long time (finding more stuff to fix all the time, which is frustrating). I propose a different fix. In https://github.com/phpstan/phpstan-src/blob/2.1.x/src/Rules/PhpDoc/VarTagTypeRuleHelper.php there are about 9 isSuperTypeOf calls.

1) Instead of calling them directly on a type, introduce a private method that accepts `$type, $varTagType` and calls isSuperTypeOf inside. That way we only need to change a single place.
2) After calling this `isSuperTypeOf`, in case an error would be reported, run one more check. Take the `$type` (the native type of the expression), do `Type::toPhpDocNode()` and then create a `Type` again with `TypeNodeResolver`. This will "clean up" everything that's not expressible with PHPDocs.
3) Call `isSuperTypeOf` again with the new `$type` that's the result of operation from 2).

VincentLanglet:
> 2. After calling this `isSuperTypeOf`, in case an error would be reported, run one more check. Take the `$type` (the native type of the expression), do `Type::toPhpDocNode()` and then create a `Type` again with `TypeNodeResolver`. This will "clean up" everything that's not expressible with PHPDocs.

`TypeNodeResolver::resolve` needs a `NameScope`. How do I get one here ?

ondrejmirtes:
I think null is fine? Or an empty one. We won't be resolving any relative names, I hope.

ondrejmirtes:
Anyway, this is just a guess what could work, maybe it's still going to break for some cases.

VincentLanglet:
> I think null is fine? Or an empty one. We won't be resolving any relative names, I hope.

Indeed.

I saw there was 5 occurences of `new NameScope(null, [])` do you think it could be useful to provide a 
`NameScope::createEmpty()` method ?

image

src/Rules/PhpDoc/VarTagTypeRuleHelper.php
```diff
@@ -157,36 +164,43 @@ private function shouldVarTagTypeBeReported(Node\Expr $expr, Type $type, Type $v
...
+			return $this->isSuperTypeOfVarType($type, $varTagType)->no();
 		}
 
-		return !$type->isSuperTypeOf($varTagType)->yes();
+		return !$this->isSuperTypeOfVarType($type, $varTagType)->yes();
+	}
+
+	private function isSuperTypeOfVarType(Type $type, Type $varTagType): TrinaryLogic
+	{
+		$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), new NameScope(null, []));
```
VincentLanglet: > After calling this isSuperTypeOf, in case an error would be reported, run one more check. 

Seems like if I do this every time it works. I would say it's avoid two calls to `isSuperTypeOf` (?)

Also since sometimes checks are 
```
!$this->isSuperTypeOfVarType($type, $varTagType)->yes()
```
and sometimes it's
```
$this->isSuperTypeOfVarType($type, $varTagType)->no()
```

It was unclear to me if I should have call `toPhpDocNode` for `maybe` results. So I did it every time.

ondrejmirtes: Please do what I suggested. I don't want to break any existing use-cases, that's why I want to do the "new" thing right before when the original error message would appear. To decrease the number of reported errors, not to potentially incraese it.

ondrejmirtes (changes requested):

src/Rules/PhpDoc/VarTagTypeRuleHelper.php
VincentLanglet: Sure, but I don't understand how I'll know "an error would be reported" easily since sometime the usage will be
`!$this->isSuperTypeOfVarType($type, $varTagType)->yes()` and sometimes it will be
`$this->isSuperTypeOfVarType($type, $varTagType)->no()`.

The following implementation
```
private function isSuperTypeOfVarType(Type $type, Type $varTagType): TrinaryLogic
	{
	    $result = $type->isSuperTypeOf($varTagType);
        if ($result->yes()) {
             return $result;
        }
	
		$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), new NameScope(null, []));

		return $type->isSuperTypeOf($varTagType);
	}
```
is not satisfying since for case where the usage is
```
$this->isSuperTypeOfVarType($type, $varTagType)->no()
```
I use the strategy `toPhpDocNode` for nothing since the first `isSuperTypeOf` wouldn't have report any error.

VincentLanglet: I have the implementation 
https://github.com/phpstan/phpstan-src/pull/3797/commits/d8b9c4841e390a71086473274cec1331bbf4fa9f

But I dunno if it's satisfying enough

ondrejmirtes:
We don't need to be too smart. We csn have two different methods - one for !yes(), one for no ().

VincentLanglet:
> We don't need to be too smart. We csn have two different methods - one for !yes(), one for no ().

I created `isSuperTypeOfVarType` and `isAtLeastMaybeSuperTypeOfVarType`

VincentLanglet:
Friendly ping @ondrejmirtes ; this PR is ready to be re-reviewed :)

ondrejmirtes (changes requested):
I worry the empty `NameScope` might cause problems for generics. Can you please test it with `@template T` above a method and then using the `T` in inline `@var` PHPDoc tag?

With a failing test I'll come up with a solution to the problem.

ondrejmirtes:
We need a test that executes the line with `$type`:

```php
		if ($type->isSuperTypeOf($varTagType)->yes()) {
			return true;
		}

		$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), new NameScope(null, []));

		return $type->isSuperTypeOf($varTagType)->yes();
```

Please debug what `$type` is before and after, ideally with `$type->describe(VerbosityLevel::precise())`.

VincentLanglet:
> With a failing test I'll come up with a solution to the problem.

I fixed the test, it's failing now.

```
var_dump($type->describe(VerbosityLevel::precise()));
$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), new NameScope(null, []));
var_dump($type->describe(VerbosityLevel::precise()));
```
gives `Closure(): list<T>` before and after

ondrejmirtes:
Now instead of `new NameScope(null, [])`, you need to get the right NameScope by doing:

```php
			$function = $scope->getFunction();
			$nameScope = $this->fileTypeMapper->getNameScope(
				$scope->getFile(),
				$scope->isInClass() ? $scope->getClassReflection()->getName() : null,
				$scope->isInTrait() ? $scope->getTraitReflection()->getName() : null,
				$function !== null ? $function->getName() : null,
			);
```

It's a new method I just added in 1.12.x.

VincentLanglet:
> Now instead of `new NameScope(null, [])`, you need to get the right NameScope by doing:
> 
> ```
> 			$function = $scope->getFunction();
> 			$nameScope = $this->fileTypeMapper->getNameScope(
> 				$scope->getFile(),
> 				$scope->isInClass() ? $scope->getClassReflection()->getName() : null,
> 				$scope->isInTrait() ? $scope->getTraitReflection()->getName() : null,
> 				$function !== null ? $function->getName() : null,
> 			);
> ```
> 
> It's a new method I just added in 1.12.x.

Thanks ! It works well.

src/Rules/PhpDoc/VarTagTypeRuleHelper.php
```diff
@@ -151,42 +160,87 @@ private function shouldVarTagTypeBeReported(Node\Expr $expr, Type $type, Type $v
...
+	private function isSuperTypeOfVarType(Scope $scope, Type $type, Type $varTagType): bool
+	{
+		if ($type->isSuperTypeOf($varTagType)->yes()) {
+			return true;
+		}
+
+		try {
+			$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), $this->createNameScope($scope));
+		} catch (NameScopeAlreadyBeingCreatedException) {
+			return false;
```
ondrejmirtes: Does this mean it's going to be reported, or it's not going to be reported? I'd rather opt for failing silently (to not report an error).

src/Rules/PhpDoc/VarTagTypeRuleHelper.php
```diff
@@ -151,42 +160,87 @@ private function shouldVarTagTypeBeReported(Node\Expr $expr, Type $type, Type $v
...
+	private function isAtLeastMaybeSuperTypeOfVarType(Scope $scope, Type $type, Type $varTagType): bool
+	{
+		if (!$type->isSuperTypeOf($varTagType)->no()) {
+			return true;
+		}
+
+		try {
+			$type = $this->typeNodeResolver->resolve($type->toPhpDocNode(), $this->createNameScope($scope));
+		} catch (NameScopeAlreadyBeingCreatedException) {
+			return false;
```
ondrejmirtes: dtto

ondrejmirtes (changes requested):

src/Rules/PhpDoc/VarTagTypeRuleHelper.php
VincentLanglet: I didn't know exactly what this exception means.

Before my PR we were doing just `$type->isSuperTypeOf($varTagType)->no()`.

> Does this mean it's going to be reported,

Yes. But the error was also reported before this PR. This won't introduce new error.
That's why I didn't chose the failing silently solution.

Now, we doing
- If `$type->isSuperTypeOf($varTagType)->no()` was OK before, early return
- Try to create a namescope
	- If we cannot create a namescope => Report an error (But this error is currently already reported by PHPStan because the check `$type->isSuperTypeOf($varTagType)->no()` was not ok).
	- If we can create a namescope => use it to remove some false-positive.

VincentLanglet: Do you still prefer, I return `true` ?

ondrejmirtes: Yes. This would be an error the user cannot do with much. So I don't want to report it.

VincentLanglet: Done

ondrejmirtes:
Perfect, thank you! From the method naming to the logic, this is top-notch!

ondrejmirtes:
Can you please submit these regression tests? Thanks.

- [ ] https://github.com/phpstan/phpstan/issues/11015
- [ ] https://github.com/phpstan/phpstan/issues/10861
- [ ] https://github.com/phpstan/phpstan/issues/7934
- [ ] https://github.com/phpstan/phpstan/issues/11535

VincentLanglet:
Sure, I opened https://github.com/phpstan/phpstan-src/pull/3841

PR #3798, opened 2025-01-27
fix: json error formatter when files are empty

shyim:
Hey,

I am trying to parse the JSON output of PHPStan in Go.

When it is filled I get:

```
"files: { "filename.php": {}}
```

but when not it's
```
"files": []
```

so it changes it type from object to array and it explodes in another languages like Go:

```
Error: json: cannot unmarshal array into Go struct field PhpStanOutput.files of type map[string]struct { Errors int "json:\"errors\""; Messages []struct { Message string "json:\"message\""; Line int "json:\"line\""; Ignorable bool "json:\"ignorable\""; Identifier string "json:\"identifier\"" } "json:\"messages\"" }
```

### DIFF
--- a/src/Command/ErrorFormatter/JsonErrorFormatter.php
+++ b/src/Command/ErrorFormatter/JsonErrorFormatter.php
@@ -5,9 +5,10 @@ namespace PHPStan\Command\ErrorFormatter;
 use Nette\Utils\Json;
 use PHPStan\Command\AnalysisResult;
 use PHPStan\Command\Output;
+use stdClass;
 use Symfony\Component\Console\Formatter\OutputFormatter;
-use function array_key_exists;
 use function count;
+use function property_exists;
 
 final class JsonErrorFormatter implements ErrorFormatter
 {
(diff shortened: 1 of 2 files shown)


src/Command/ErrorFormatter/JsonErrorFormatter.php
```diff
@@ -60,6 +61,11 @@ public function formatErrors(AnalysisResult $analysisResult, Output $output): in
 			$errorsArray['errors'][] = $notFileSpecificError;
 		}
 
+		// When files empty format it as an empty object
+		if ($errorsArray['files'] === []) {
+			$errorsArray['files'] = new stdClass();
```
ondrejmirtes: I'd prefer if this key was always stdClass, for consistency.

ondrejmirtes (changes requested):
You could even go further and make everything that's string-indexed array here being encoded into stdClass. But I don't require that.

ondrejmirtes:
Thank you.

PR #3800, opened 2025-01-29
Add stub for \DOMNode::hasAttributes

Niklan:
I'm not sure if the property should be explicitely declared. It works fine without it. Most likely, it comes from PHPStorm's stubs. Also, I'm not sure if it should be:

```php
@phpstan-assert-if-true DOMNamedNodeMap $this->attributes
```

to be more specific, or is it fine with `!null`?

This PR fixes an issue where PHPStan does not understand chained calls, as discussed here: https://github.com/phpstan/phpstan/issues/12495#issuecomment-2619213520

```php
    if ($node->hasAttributes()) {
      foreach ($node->attributes->getIterator() as $attribute) {
```

```
  22     Cannot call method getIterator() on DOMNamedNodeMap|null.  
         🪪  method.nonObject   
```

### DIFF
--- a/stubs/dom.stub
+++ b/stubs/dom.stub
@@ -30,6 +30,17 @@ class DOMDocument
 class DOMNode
 {
 
+	/**
+	 * @var DOMNamedNodeMap|null
+	 */
+	public $attributes;
+
+	/**
+	 * @phpstan-assert-if-true !null $this->attributes
+	 * @return bool
+	 */
+	public function hasAttributes() {}
+
 }
 
 class DOMElement extends DOMNode


ondrejmirtes:
Thank you.

VincentLanglet:
For the record, it now report `DomElement::hasAttributes` as always true.
https://phpstan.org/r/a2e4727f-a66f-4aee-9d62-18973c8984a8

I think it's because of the bad type of PHPStorm stubs.
I tried https://github.com/JetBrains/phpstorm-stubs/pull/1741

PR #3801, opened 2025-01-30
Imagick::getConfigureOptions() returns array instead of string

blankse:
Should be fix https://phpstan.org/r/650952e7-9859-4db0-ae91-58b8ca25a1ef

See:
https://github.com/JetBrains/phpstorm-stubs/pull/1707
https://phpimagick.com/Imagick/getConfigureOptions

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -4729,7 +4729,7 @@ return [
 'Imagick::getColorspace' => ['Imagick::COLORSPACE_*'],
 'Imagick::getCompression' => ['Imagick::COMPRESSION_*'],
 'Imagick::getCompressionQuality' => ['int'],
-'Imagick::getConfigureOptions' => ['array'],
+'Imagick::getConfigureOptions' => ['array<string, string>'],
 'Imagick::getCopyright' => ['string'],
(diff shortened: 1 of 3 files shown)


resources/functionMap.php
```diff
@@ -4729,7 +4729,7 @@
 'Imagick::getColorspace' => ['Imagick::COLORSPACE_*'],
 'Imagick::getCompression' => ['Imagick::COMPRESSION_*'],
 'Imagick::getCompressionQuality' => ['int'],
-'Imagick::getConfigureOptions' => ['string'],
+'Imagick::getConfigureOptions' => ['array'],
```
ondrejmirtes: Would be great to be more specific here, is it `array<uppercase-string, string>`?

ondrejmirtes (changes requested):

resources/functionMap.php
blankse: @ondrejmirtes I changed it here https://github.com/phpstan/phpstan-src/pull/3801/commits/e92d4786ed6654e39d3d199f8a2178c8d6a23780

It's not just uppercase strings.

ondrejmirtes:
Thank you.

PR #3803, opened 2025-02-02
Prevent declaring hooked properties as readonly

jakubtobiasz:
Fulfills `Hooked properties cannot be readonly` from https://github.com/phpstan/phpstan/issues/12336 :)

### DIFF
--- a/src/Rules/Properties/PropertyInClassRule.php
+++ b/src/Rules/Properties/PropertyInClassRule.php
@@ -72,8 +72,6 @@ final class PropertyInClassRule implements Rule
 						->build(),
 				];
 			}
-
-			return [];
 		}
 
 		if ($node->isReadOnly()) {
@@ -87,13 +85,15 @@ final class PropertyInClassRule implements Rule
 			}
 		}
 
-		if (!$this->doAllHooksHaveBody($node)) {
-			return [
-				RuleErrorBuilder::message('Non-abstract properties cannot include hooks without bodies.')
-					->nonIgnorable()
-					->identifier('property.hookWithoutBody')
-					->build(),
-			];
+		if (!$node->isAbstract() && !$node->isReadOnly()) {
+			if (!$this->doAllHooksHaveBody($node)) {
+				return [
+					RuleErrorBuilder::message('Non-abstract properties cannot include hooks without bodies.')
+						->nonIgnorable()
+						->identifier('property.hookWithoutBody')
+						->build(),
+				];
+			}
 		}
 
 		return [];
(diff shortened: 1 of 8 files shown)


src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -76,6 +76,17 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		if ($node->isReadOnly()) {
+			if ($node->hasHooks()) {
```
herndlm, line 85: You could move both conditions into a single if and avoid a nesting level?

jakubtobiasz: In theory I can, but I wanted to be consistent with how we handled the abstract ones.

herndlm: Ah OK, was just curious about the reason and didn't see that.

ondrejmirtes: Yeah, I prefer this structure. It makes very easy to see conditions that aren't covered by the code. For example that abstract readonly properties will not be reported 😂 https://phpstan.org/r/45f2ba85-1ebe-4924-8b28-5356639b5ad9

ondrejmirtes (changes requested):

ondrejmirtes:
Also please do the same in the interface property rule, thank you.

src/Rules/Properties/PropertyInClassRule.php
ondrejmirtes: Of course please check first whether PHP allows it or not, I might be wrong.

jakubtobiasz:
Hello @ondrejmirtes 👋🏼,

I've covered what you asked for :).

As I'm working on hooked properties and I see your list of `to be done` in that matter, what do you think about splitting `PropertyInClassRule` into `PropertyInClassRule` (non-hooked properties thingies only), `HookedPropertyInAbstractClassRule`, `HookedPropertyInReadonlyClassRule`, `HookedPropertyInClassRule` or something similar? I'm afraid keeping all these rules in a single class might affect readability and maintainability in the future :/.

ondrejmirtes:
I think the current way is better. It's easier to see what is and what isn't covered. Some rules apply to all properties so with your suggestion there would have to be some duplication too.

jakubtobiasz:
Ok, so do you anything more from me in this PR or I can take a next one :D.

ondrejmirtes:
I moved a piece of code for better readability (https://github.com/phpstan/phpstan-src/pull/3803/commits/fdda36ea1097475e347d91467576861d7143b279) which means the error is reported regardless of `$node->isReadOnly()` and no tests have failed. Can you please confirm this still works as intended?

jakubtobiasz:
Seems legit 🤔

ondrejmirtes:
Thank you!

PR #3804, opened 2025-02-03
Teamcity - show rule identifier when verbose output is set

kamil-zacek:
Teamcity Code Inspection - show rule identifier for error when verbose output is set

### DIFF
--- a/src/Command/ErrorFormatter/TeamcityErrorFormatter.php
+++ b/src/Command/ErrorFormatter/TeamcityErrorFormatter.php
@@ -10,6 +10,7 @@ use function array_values;
 use function count;
 use function is_string;
 use function preg_replace;
+use function sprintf;
 use const PHP_EOL;
 
 /**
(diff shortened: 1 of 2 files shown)


kamil-zacek:
@ondrejmirtes i've added rule identifier to teamcity output to show identifier of an error in Teamcity. It should show this identifier in the parentheses when the verbose output is active.

ondrejmirtes:
Thank you!

jom:
Just a note: The emoji seemed to break the TeamCity parsing of the service message in our instance. I'm working around this for now with a custom error reporter that removes the emoji.

kamil-zacek:
> Just a note: The emoji seemed to break the TeamCity parsing of the service message in our instance. I'm working around this for now with a custom error reporter that removes the emoji.

Interesting. We don't have the same problem at our teamcity. But it's true that in teamcity the emoji doesn't look as good as at phpstan command line output, so I think there is no problem to remove it.

PR #3806, opened 2025-02-04
Add generic typings to array_uintersect

axlon:
This PR adds generic typings to `array_uintersect`, I copied the annotations from `array_udiff`

### DIFF
--- a/stubs/arrayFunctions.stub
+++ b/stubs/arrayFunctions.stub
@@ -58,11 +58,11 @@ function uksort(array &$array, callable $callback): bool
 }
 
 /**
- * @template T of mixed
- *
+ * @template T
+ * @template U
  * @param array<T> $one
- * @param array<T> $two
- * @param callable(T, T): int $three
+ * @param array<U> $two
+ * @param callable(T, U): int $three
  */
 function array_udiff(
     array $one,
@@ -72,9 +72,10 @@ function array_udiff(
 
 /**
  * @template T
+ * @template U
  * @param array<T> $one
- * @param array<T> $two
- * @param callable(T, T): int $three
+ * @param array<U> $two
+ * @param callable(T, U): int $three
  */
 function array_uintersect(
     array $one,
--- a/stubs/arrayFunctions.stub
+++ b/stubs/arrayFunctions.stub
@@ -71,6 +71,18 @@ function array_udiff(
 ): int {}
 
 /**
+ * @template T
+ * @param array<T> $one
+ * @param array<T> $two
+ * @param callable(T, T): int $three
+ */
+function array_uintersect(
+    array $one,
+    array $two,
+    callable $three
+): int {}
+
+/**
  * @param array<array-key, mixed> $value
  * @return ($value is list ? true : false)
  */
--- a/stubs/arrayFunctions.stub
+++ b/stubs/arrayFunctions.stub
@@ -58,30 +58,32 @@ function uksort(array &$array, callable $callback): bool
 }
 
 /**
- * @template T
- * @template U
- * @param array<T> $one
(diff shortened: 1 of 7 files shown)


axlon:
This PR is a partial fix for https://github.com/phpstan/phpstan/issues/7707, if this PR is the correct way to do this I can add support for the other functions mentioned as well

stubs/arrayFunctions.stub
```diff
@@ -70,6 +70,18 @@ function array_udiff(
...
+ * @template T
+ * @param array<T> $one
+ * @param array<T> $two
+ * @param callable(T, T): int $three
+ */
+function array_uintersect(
+    array $one,
+    array $two,
+    callable $three
+): int {}
```
ondrejmirtes: It'd be better to update the array_udiff and array_uintersect to actually use two different `@template` for each array (T + U).

ondrejmirtes (changes requested):

schlndh:
@axlon BTW: The same fix is already implemented here: https://github.com/phpstan/phpstan-src/pull/3282

axlon:
> @axlon BTW: The same fix is already implemented here: https://github.com/phpstan/phpstan-src/pull/3282

I see, I missed this when looking through issues. However there's a slight difference in that the arrays can be different per @ondrejmirtes' suggestion.

The same issue with that PR is still present here; the definitions of these functions are wrong, but merging this would get us closer to PHPStan fully understanding these functions

schlndh:
@axlon You're right. Feel free to take whatever you want from the original PR, if you decide to expand this one to cover all similar functions. Alternatively, I'm also around to update my PR if ondrejmirtes prefers that.

stubs/arrayFunctions.stub
```diff
@@ -58,18 +58,31 @@ function uksort(array &$array, callable $callback): bool
...
+ * @template U
+ * @param array<T> $one
+ * @param array<U> $two
+ * @param callable(T, U): int $three
+ */
+function array_uintersect(
+    array $one,
+    array $two,
+    callable $three
+): int {}
```
schlndh: Both `array_udiff` and `array_uintersect` return an array, not int: https://3v4l.org/R0WNi

axlon:
I'll try to take a look at this soon

axlon:
I think I've found a way to document these functions where they're quite accurate (so far), but I hit a roadblock with array shapes; PHPStan seems to add an integer key to the return type whenever the first argument is an array shape. I've added failing tests in the last commit and I will try to figure out why this is happening

schlndh:
@axlon It looks like it might be possible to work around it by using `array<K, V>` and then `K`, instead of `T of array` and `key-of<T>`: https://phpstan.org/r/f5f9cb93-4b83-4042-a91f-cd76d4a1789c

axlon:
> @axlon It looks like it might be possible to work around it by using `array<K, V>` and then `K`, instead of `T of array` and `key-of<T>`: https://phpstan.org/r/f5f9cb93-4b83-4042-a91f-cd76d4a1789c

~I'll check that out!~ Not only does this fix the issue, it also makes the return types more accurate 👌

axlon:
@ondrejmirtes please check this out when you find the time, and let me know if this is the way forward. If it is I'll work on adding the remaining array functions listed in the original issue.

As far as the failing tests go:
- The PHP 7.4 tests probably fail because of mixed (I suppose I should exclude them for PHP 7.4?)
- The other failing workflows seem unrelated (to my eyes)

ondrejmirtes:
https://github.com/phpstan/phpstan-src/pull/3282 seemed a bit more complete to me so I merged that. Thank you.

PR #3809, opened 2025-02-05
Add a test covering a hooked property in a readonly class

jakubtobiasz:
`$node->isReadOnly()` returns `true` when a property lives inside a readonly class, so I've added only a simple test to existing test suite :).

Fulfills `Readonly classes cannot have hooked properties` (ref: https://github.com/phpstan/phpstan/issues/12336)


ondrejmirtes:
Thank you very much! BTW each item in the linked issue does not have to be a separate PR, you can do multiple items at once if they are related :)

PR #3810, opened 2025-02-05
Add handling an attempt to declare a hooked property static

jakubtobiasz:
Fulfills `Static properties cannot be hooked` (ref: https://github.com/phpstan/phpstan/issues/12336)

### DIFF
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -66,6 +66,15 @@ final class PropertiesInInterfaceRule implements Rule
 			];
 		}
 
+		if ($node->isStatic()) {
+			return [
+				RuleErrorBuilder::message('Hooked properties cannot be static.')
+					->nonIgnorable()
+					->identifier('property.hookedStaticInInterface')
+					->build(),
+			];
+		}
+
 		if ($this->hasAnyHookBody($node)) {
 			return [
 				RuleErrorBuilder::message('Interfaces cannot include property hooks with bodies.')
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -70,7 +70,7 @@ final class PropertiesInInterfaceRule implements Rule
 			return [
 				RuleErrorBuilder::message('Hooked properties cannot be static.')
 					->nonIgnorable()
-					->identifier('property.hookedStaticInInterface')
+					->identifier('property.hookedStatic')
 					->build(),
 			];
 		}
--- a/src/Rules/Properties/PropertyInClassRule.php
+++ b/src/Rules/Properties/PropertyInClassRule.php
@@ -92,6 +92,17 @@ final class PropertyInClassRule implements Rule
 			}
 		}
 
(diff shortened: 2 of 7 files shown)


ondrejmirtes (changes requested):
PropertiesInInterfaceRule should cover this too.

jakubtobiasz:
Done :)

src/Rules/Properties/PropertiesInInterfaceRule.php
```diff
@@ -66,6 +66,15 @@ public function processNode(Node $node, Scope $scope): array
 			];
 		}
 
+		if ($node->isStatic()) {
+			return [
+				RuleErrorBuilder::message('Interfaces cannot include static hooked properties.')
+					->nonIgnorable()
+					->identifier('property.staticInInterface')
```
ondrejmirtes: I think we should use the exact same message and identifier as in PropertyInClassRule.

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -92,6 +92,17 @@ public function processNode(Node $node, Scope $scope): array
 			}
 		}
 
+		if ($node->isStatic()) {
+			if ($node->hasHooks()) {
+				return [
+					RuleErrorBuilder::message('Hooked properties cannot be static.')
+						->nonIgnorable()
+						->identifier('property.hookStatic')
```
ondrejmirtes: `hookedStatic` would be better

ondrejmirtes (changes requested):

src/Rules/Properties/PropertiesInInterfaceRule.php
jakubtobiasz: I wanted to keep consistency across the class. In my opinion it looks better once we look at the whole class, not only this particular rule error builder. Of course, the final decision is up to you 🖖🏻.

src/Rules/Properties/PropertyInClassRule.php
jakubtobiasz: As above, it's consistent now. If you really want to change it, let me know 🖖🏻.

src/Rules/Properties/PropertiesInInterfaceRule.php
ondrejmirtes: I think we should use the exact same message and identifier as in PropertyInClassRule.

src/Rules/Properties/PropertyInClassRule.php
ondrejmirtes: `hookedStatic` would be better

src/Rules/Properties/PropertiesInInterfaceRule.php
```diff
@@ -66,6 +66,15 @@ public function processNode(Node $node, Scope $scope): array
 			];
 		}
 
+		if ($node->isStatic()) {
+			return [
+				RuleErrorBuilder::message('Hooked properties cannot be static.')
+					->nonIgnorable()
+					->identifier('property.hookedStaticInInterface')
```
ondrejmirtes:
```suggestion
					->identifier('property.hookedStatic')
```

ondrejmirtes:
Thank you.

PR #3812, opened 2025-02-07
Prevent setting a default value for a virtual property

jakubtobiasz:
Fulfills `Virtual property cannot have default value` (ref: https://github.com/phpstan/phpstan/issues/12336)


ondrejmirtes:
Thank you.

PR #3816, opened 2025-02-10
Return given value for dumpType

alexander-schranz:
This make it possible to add PHPStan\dumpType to nested calls like:

```php
$test = new Test(\PHPStan\dumpType($other));
```

without additional errors returned.

### DIFF
--- a/src/dumpType.php
+++ b/src/dumpType.php
@@ -3,25 +3,27 @@
 namespace PHPStan;
 
 /**
+ * @template T
  * @phpstan-pure
- * @param mixed $value
- * @return mixed
+ * @param T $value
+ * @return T
  *
  * @throws void
  */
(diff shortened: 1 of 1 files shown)


src/dumpType.php
```diff
@@ -3,25 +3,27 @@
...
+ * @template T
  * @phpstan-pure
- * @param mixed $value
- * @return mixed
+ * @param T $value
+ * @return T
  *
  * @throws void
  */
 function dumpType($value) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
```
mvorisek, line 13: IMO the return type should be `void`.

alexander-schranz: The Pull Requests target is to return the value not to make it null or void.

ondrejmirtes:
No, sorry, I think that would be very surprising behaviour. You can always save the value into a variable and dump that.

PR #3817, opened 2025-02-10
Fix readonly property assign with ArrayAccess offset

herndlm:
Closes https://github.com/phpstan/phpstan/issues/8929
Closes https://github.com/phpstan/phpstan/issues/12537

UPDATE: added test for 8929 and dealing with unset after realizing the bug duplicate and also finding https://github.com/phpstan/phpstan-src/pull/3795. I don't mind which one gets merged first, anybody feel free to re-use parts from here in the other MR or directly.


herndlm:
sorry boss, didn't want to hog CI, but thought it would be a good idea to rebase the last 3 non-controversial PRs from me 😊

ondrejmirtes:
Thank you!

PR #3825, opened 2025-02-13
feat: cache the result of ClassReflection::hasMethod method

canvural:
Hello 👋🏽 

This PR caches the result of the `ClassReflection::hasMethod` method. While investigating some performance issues I found that PHPStan calls method class reflection extensions multiple times for the same class and method. And some extensions might do the actual work in `hasMethod` method rather than `getMethod`, for example Larastan does this to cache stuff internally. 

I can add tests if wanted. Couldn't think of an easy way to do it right away 😅

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -142,6 +142,9 @@ class ClassReflection
 	/** @var array<string, true> */
 	private static array $resolvingTypeAliasImports = [];
 
+	/** @var array<string, bool> */
+	private array $hasMethodCache = [];
+
 	/**
 	 * @param PropertiesClassReflectionExtension[] $propertiesClassReflectionExtensions
 	 * @param MethodsClassReflectionExtension[] $methodsClassReflectionExtensions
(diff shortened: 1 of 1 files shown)


src/Reflection/ClassReflection.php
```diff
@@ -473,16 +476,28 @@ public function hasProperty(string $propertyName): bool
 
 	public function hasMethod(string $methodName): bool
 	{
+		$key = sprintf('%s-%s', $methodName, $this->getCacheKey());
```
ondrejmirtes: This cache key doesn't make sense. Here we are inside a specific ClassReflection instance. The cache property isn't static. I believe this could just be `$methodName;` as it's in `getMethod`. In there the `$key` is modified when the `$scope` is in class, but `hasMethod` doesn't have `$scope` parameter so the cache key can simply be `$methodName`.

Remove the `$key` variable altogether, and just use `$methodName` when accessing the array. Thank you.

ondrejmirtes (changes requested):

src/Reflection/ClassReflection.php
canvural: Yep. makes total sense. Updated.

ondrejmirtes:
Thank you.

PR #3827, opened 2025-02-15
Readonly properties cannot be unset()

staabm:
closes https://github.com/phpstan/phpstan/issues/12421

### DIFF
--- a/src/Rules/Variables/UnsetRule.php
+++ b/src/Rules/Variables/UnsetRule.php
@@ -91,14 +91,12 @@ final class UnsetRule implements Rule
 			}
 
 			if ($propertyReflection->isReadOnly() || $propertyReflection->isReadOnlyByPhpDoc()) {
-				$type = $scope->getType($node->var);
-
 				return RuleErrorBuilder::message(
 					sprintf(
-						'Cannot unset %s property %s of %s.',
+						'Cannot unset %s %s::$%s property.',
 						$propertyReflection->isReadOnly() ? 'readonly' : '@readonly',
-						$node->name->name,
-						$type->describe(VerbosityLevel::value()),
+						$propertyReflection->getDeclaringClass()->getDisplayName(),
+						$foundPropertyReflection->getName(),
 					),
 				)
 					->line($node->getStartLine())
--- a/src/Rules/Variables/UnsetRule.php
+++ b/src/Rules/Variables/UnsetRule.php
@@ -75,20 +75,6 @@ final class UnsetRule implements Rule
 		) {
 			$type = $scope->getType($node->var);
 			foreach ($type->getObjectClassReflections() as $classReflection) {
-				if ($classReflection->isReadOnly() || $classReflection->isImmutable()) {
-					return RuleErrorBuilder::message(
-						sprintf(
-							'Cannot unset property %s of %s class %s.',
-							$node->name->name,
-							$classReflection->isReadOnly() ? 'readonly' : 'immutable',
-							$type->describe(VerbosityLevel::value()),
-						),
-					)
-						->line($node->getStartLine())
-						->identifier('unset.readonlyClass')
-						->build();
-				}
-
 				if (!$classReflection->hasNativeProperty($node->name->name)) {
 					continue;
 				}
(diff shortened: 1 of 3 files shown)


src/Rules/Variables/UnsetRule.php
```diff
@@ -69,6 +69,45 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
 			}
 
 			return $this->canBeUnset($node->var, $scope);
+		} elseif (
+			$node instanceof Node\Expr\PropertyFetch
+			&& $node->name instanceof Node\Identifier
+		) {
+			$type = $scope->getType($node->var);
+			foreach ($type->getObjectClassReflections() as $classReflection) {
+				if ($classReflection->isReadOnly() || $classReflection->isImmutable()) {
```
ondrejmirtes: I'd completely leave out this section about readonly/immutable classes. It will be caught by the section about native properties which is okay by me.

src/Rules/Variables/UnsetRule.php
```diff
@@ -69,6 +69,45 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
...
+					return RuleErrorBuilder::message(
+						sprintf(
+							'Cannot unset %s property %s of %s.',
+							$propertyReflection->isReadOnly() ? 'readonly' : '@readonly',
+							$node->name->name,
+							$type->describe(VerbosityLevel::value()),
+						),
+					)
+						->line($node->getStartLine())
+						->identifier('unset.readonlyProperty')
```
ondrejmirtes: We need different identifier for isReadOnly/isReadOnlyByPhpDoc.

ondrejmirtes (changes requested):

staabm:
adjusted per feedback. thank you.

src/Rules/Variables/UnsetRule.php
```diff
@@ -69,6 +69,31 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
 			}
 
 			return $this->canBeUnset($node->var, $scope);
+		} elseif (
+			$node instanceof Node\Expr\PropertyFetch
+			&& $node->name instanceof Node\Identifier
+		) {
```
ondrejmirtes, line 82: We have better ways of getting a property reflection from PropertyFetch. Two actually:

1) PropertyReflectionFinder. FoundPropertyReflection will even tell you if it's a native one or not.
2) `Scope::getPropertyReflection()`

ondrejmirtes (changes requested):

src/Rules/Variables/UnsetRule.php
staabm: ahh nice. 

I was already searching how to know about property hooks so I can implement the error for unsetting a hooked property.

will be a followup PR

src/Rules/Variables/UnsetRule.php
```diff
@@ -69,6 +76,35 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
...
+				return null;
+			}
+
+			$propertyReflection = $foundPropertyReflection->getNativeReflection();
+			if ($propertyReflection === null) {
+				return null;
+			}
+
+			if ($propertyReflection->isReadOnly() || $propertyReflection->isReadOnlyByPhpDoc()) {
+				$type = $scope->getType($node->var);
```
ondrejmirtes: You don't need to do this, just ask for property declaringClass getDisplayName.

src/Rules/Variables/UnsetRule.php
```diff
@@ -69,6 +76,35 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
...
+
+				return RuleErrorBuilder::message(
+					sprintf(
+						'Cannot unset %s property %s of %s.',
```
ondrejmirtes: Usually we do Foo::$name format.

%s of %s sounds like an ancient text: Yennefer of Vengerberg 😀

ondrejmirtes (changes requested):

src/Rules/Variables/UnsetRule.php
ondrejmirtes: I wrote about property hooks here https://github.com/phpstan/phpstan/discussions/12337

ondrejmirtes:
Perfect, thank you!

PR #3828, opened 2025-02-16
ci: add failing test for bug-12585

calebdw:
Hello!

This adds failings tests for https://github.com/phpstan/phpstan/issues/12585.

Please note that there's actually two bugs here:

### 1. Return type

I'm getting errors of the form:

```php
Post::query()->whereHas('user', fn ($q) => $q->where('name', 'test'));
// Anonymous function should return Bug12585\Builder<Bug12585\Model> but returns Bug12585\Builder<Bug12585\User>
```

despite the parent method having `(\Closure(Builder<TRelatedModel>): mixed)|null` as the `$callback` type; and the `MethodParameterClosureTypeExtension` specifying `MixedType` as the return type for the replaced closure type.

It seems like the parameter type of the closure phpdoc is being used as the return type...

### 2. Parameter Type

I am getting errors of the form:

```php
User::query()->whereHas('posts', fn (PostBuilder $q) => $q->where('name', 'test'));
// Parameter #2 $callback of method Bug12585\Builder<Bug12585\User>::whereHas() expects (Closure(Bug12585\Builder<Bug12585\Post>): mixed)|null,             
//         Closure(Bug12585\PostBuilder): Bug12585\PostBuilder given.
```

I think that the `MethodParameterClosureTypeExtension` is only used to resolve the closure type inside the closure, while the PHPDoc is being used when checking method calls.

Thanks!


makroxyz:
Please merge!

ondrejmirtes:
@makroxyz It doesn't make sense to merge failing tests, it needs to be fixed first.

canvural:
Looks similar to #3131

PR #3830, opened 2025-02-16
Hooked properties cannot be both final and private

staabm:
refs https://github.com/phpstan/phpstan/issues/12336

see https://3v4l.org/5ZXOF

### DIFF
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -75,6 +75,15 @@ final class PropertiesInInterfaceRule implements Rule
 			];
 		}
 
+		if ($node->isFinal()) {
+			return [
+				RuleErrorBuilder::message('Interfaces cannot include final properties.')
+					->nonIgnorable()
+					->identifier('property.finalInInterface')
+					->build(),
+			];
+		}
+
 		if ($this->hasAnyHookBody($node)) {
 			return [
 				RuleErrorBuilder::message('Interfaces cannot include property hooks with bodies.')
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -75,6 +75,15 @@ final class PropertiesInInterfaceRule implements Rule
 			];
 		}
 
+		if ($node->isAbstract()) {
+			return [
+				RuleErrorBuilder::message('Property in interface cannot be explicitly abstract.')
+					->nonIgnorable()
+					->identifier('property.hookedStatic')
+					->build(),
+			];
+		}
+
 		if ($node->isFinal()) {
 			return [
 				RuleErrorBuilder::message('Interfaces cannot include final properties.')
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -101,7 +101,7 @@ final class PropertiesInInterfaceRule implements Rule
 			return [
 				RuleErrorBuilder::message('Property hook cannot be both abstract and final.')
 					->nonIgnorable()
-					->identifier('property.hookedAbstractFinal')
+					->identifier('property.abstractFinalHook')
 					->build(),
 			];
 		}
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -79,7 +79,7 @@ final class PropertiesInInterfaceRule implements Rule
 			return [
 				RuleErrorBuilder::message('Property in interface cannot be explicitly abstract.')
 					->nonIgnorable()
-					->identifier('property.hookedAbstract')
+					->identifier('property.abstractInInterface')
 					->build(),
 			];
 		}
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -84,6 +84,19 @@ final class PropertiesInInterfaceRule implements Rule
 			];
 		}
 
+		foreach ($node->getHooks() as $hook) {
+			if (!$hook->isFinal()) {
+				continue;
+			}
+
+			return [
+				RuleErrorBuilder::message('Property hook cannot be both abstract and final.')
+					->nonIgnorable()
+					->identifier('property.abstractFinal')
+					->build(),
+			];
+		}
+
 		if ($this->hasAnyHookBody($node)) {
 			return [
 				RuleErrorBuilder::message('Interfaces cannot include property hooks with bodies.')
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -79,7 +79,7 @@ final class PropertiesInInterfaceRule implements Rule
 			return [
 				RuleErrorBuilder::message('Property in interface cannot be explicitly abstract.')
 					->nonIgnorable()
-					->identifier('property.hookedStatic')
+					->identifier('property.hookedAbstract')
 					->build(),
 			];
 		}
--- a/src/Rules/Properties/PropertiesInInterfaceRule.php
+++ b/src/Rules/Properties/PropertiesInInterfaceRule.php
@@ -101,7 +101,7 @@ final class PropertiesInInterfaceRule implements Rule
 			return [
 				RuleErrorBuilder::message('Property hook cannot be both abstract and final.')
 					->nonIgnorable()
-					->identifier('property.abstractFinal')
+					->identifier('property.hookedAbstractFinal')
 					->build(),
 			];
 		}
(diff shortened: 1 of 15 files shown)


ondrejmirtes (changes requested):
Even non-hooked property cannot be final&private: https://3v4l.org/fAXFZ

Also please check the same error for methods and constants (it might already be implemented, not sure)

Also, please add an error that would report final properties on PHP < 8.4.

Also, please add error for final properties in interfaces.

Also, please add error for final hooks in abstract properties (in both class and interface): https://3v4l.org/hSO0C#v8.4.4

Also, abstract + final property together is not allowed.

staabm:
memo to me: more cases to consider
- https://3v4l.org/fvOaG
- https://3v4l.org/SYBhr

additional: follow discussion in https://github.com/php/php-src/pull/17861

staabm:
> Even non-hooked property cannot be final&private: [3v4l.org/fAXFZ](https://3v4l.org/fAXFZ)

done

> Also please check the same error for methods and constants (it might already be implemented, not sure)

methods are already covered and tests exist for it
- https://phpstan.org/r/19373f40-9ff3-447a-b9c3-da3714410992

constants are missing (separate PR? new rule similar to FinalPrivateMethodRule?)
- https://phpstan.org/r/6a42162e-ada2-479f-aacd-f29bbf717583


> Also, please add an error that would report final properties on PHP < 8.4.
 
done

> Also, please add error for final properties in interfaces.

should I add a new PropertyInInterfaceRule  ?

> Also, please add error for final hooks in abstract properties (in both class and interface): [3v4l.org/hSO0C#v8.4.4](https://3v4l.org/hSO0C#v8.4.4)

done
 
> Also, abstract + final property together is not allowed.

php-src doesn't error on https://3v4l.org/2fSaq#v8.4.4

ondrejmirtes:
> should I add a new PropertyInInterfaceRule ?

There is already `PropertiesInInterfaceRule`.

ondrejmirtes:
Reported bug to PHP https://github.com/php/php-src/issues/17916

ondrejmirtes:
Please add an error for `final abstract` properties as it's now fixed and reported in PHP.

staabm:
I think we are running now in a php-parser bug: https://github.com/nikic/PHP-Parser/issues/1071. will look further

ondrejmirtes:
It's fine if it's a parse error. It means we don't need to implement a custom rule to detect this invalid code.

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -32,6 +32,18 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		if (
+			$node->isFinal()
+			&& !$this->phpVersion->supportsFinalProperties()
+		) {
+			return [
+				RuleErrorBuilder::message('Property cannot be final.')
```
ondrejmirtes: This needs to follow a message pattern of other features that have support in recent PHP versions. Example: Property hooks are supported only on PHP 8.4 and later.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Properties/PropertiesInInterfaceRuleTest.php
```diff
@@ -40,6 +40,10 @@ public function testPhp83AndPropertiesInInterface(): void
 				'Interfaces cannot include properties.',
 				11,
 			],
+			[
+				'Interfaces cannot include properties.',
```
staabm: should this error (pre existing to this PR) also be changed to

```suggestion
				'Interfaces can include properties only on PHP 8.4 and later.',
```

?

ondrejmirtes: Yes please. When I originally added this error it was before PHP 8.4.

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -81,6 +93,54 @@ public function processNode(Node $node, Scope $scope): array
...
+				];
+			}
+
+			foreach ($node->getHooks() as $hook) {
+				if (!$hook->isFinal()) {
+					continue;
+				}
+
+				return [
+					RuleErrorBuilder::message('Property cannot be both final and private.')
```
ondrejmirtes: This message should say it's about the hook. "Private property cannot have a final hook" or something like that.

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -81,6 +93,54 @@ public function processNode(Node $node, Scope $scope): array
...
+						->nonIgnorable()
+						->identifier('property.finalPrivate')
+						->build(),
+				];
+			}
+		}
+
+		if ($classReflection->isAbstract() && $node->isAbstract()) {
```
ondrejmirtes: I'd remove `$classReflection->isAbstract() && ` here.

ondrejmirtes (changes requested):

src/Rules/Properties/PropertyInClassRule.php
staabm: thanks - I am still learning this new property hooks thing :-)

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -81,6 +93,54 @@ public function processNode(Node $node, Scope $scope): array
...
+				];
+			}
+
+			foreach ($node->getHooks() as $hook) {
+				if (!$hook->isFinal()) {
+					continue;
+				}
+
+				return [
+					RuleErrorBuilder::message('Property cannot be both abstract and final.')
```
ondrejmirtes, line 140: This is not an error in PHP: https://3v4l.org/j1ktj#v8.4.4

src/Rules/Properties/PropertiesInInterfaceRule.php
```diff
@@ -75,6 +75,15 @@ public function processNode(Node $node, Scope $scope): array
 			];
 		}
 
+		if ($node->isFinal()) {
+			return [
+				RuleErrorBuilder::message('Interfaces cannot include final properties.')
```
ondrejmirtes, line 89: What about final hooks in interfaces?

ondrejmirtes (changes requested):

ondrejmirtes:
One more thing to report - final hook without a body in a class: https://phpstan.org/r/1d253f61-25a3-4654-8ef0-4f3c4268c770

staabm:
> One more thing to report - final hook without a body in a class: [phpstan.org/r/1d253f61-25a3-4654-8ef0-4f3c4268c770](https://phpstan.org/r/1d253f61-25a3-4654-8ef0-4f3c4268c770)

I think thats already covered in https://github.com/phpstan/phpstan-src/pull/3830/files#diff-2678a88fc82b38e00cf1a626ec3234f0d0e78b202c1324abce258eadfc735e68R3-R10

src/Rules/Properties/PropertiesInInterfaceRule.php
```diff
@@ -75,6 +75,37 @@ public function processNode(Node $node, Scope $scope): array
 			];
 		}
 
+		if ($node->isAbstract()) {
+			return [
+				RuleErrorBuilder::message('Property in interface cannot be explicitly abstract.')
+					->nonIgnorable()
+					->identifier('property.hookedStatic')
```
ondrejmirtes: wrong identifier here

ondrejmirtes (changes requested):

src/Rules/Properties/PropertyInClassRule.php
```diff
@@ -81,6 +93,58 @@ public function processNode(Node $node, Scope $scope): array
...
+
+			foreach ($node->getHooks() as $hook) {
+				if (!$hook->isFinal()) {
+					continue;
+				}
+
+				return [
+					RuleErrorBuilder::message('Private property cannot have a final hook.')
+						->nonIgnorable()
+						->identifier('property.finalPrivateHook')
```
staabm, line 114: maybe

```suggestion
						->identifier('property.hookedPrivateFinal')
```

?

src/Rules/Properties/PropertiesInInterfaceRule.php
ondrejmirtes: Still wrong? `property.hookedStatic` does not make sense for `Property in interface cannot be explicitly abstract.`?

ondrejmirtes (changes requested):

src/Rules/Properties/PropertiesInInterfaceRule.php
staabm: argh, sorry

src/Rules/Properties/PropertiesInInterfaceRule.php
```diff
@@ -75,6 +75,37 @@ public function processNode(Node $node, Scope $scope): array
...
+
+		foreach ($node->getHooks() as $hook) {
+			if (!$hook->isFinal()) {
+				continue;
+			}
+
+			return [
+				RuleErrorBuilder::message('Property hook cannot be both abstract and final.')
+					->nonIgnorable()
+					->identifier('property.hookedAbstractFinal')
```
ondrejmirtes:
```suggestion
					->identifier('property.abstractFinalHook')
```

ondrejmirtes:
Perfect, thank you!

ondrejmirtes:
Something you could do next: abstract + private (https://3v4l.org/es9gD)

PR #3837, opened 2025-02-25
Add support for pseudo-abstract/self-referencing constants

El-Virus:
PHP doesn't support abstract constants, so using self-referencing constants in traits in order to force implementer classes to define the constant is not unheard of.
An example of such an implementation would be:
```php
trait Greeter {
    private const string WHO = self::WHO;

    public static function greeting() : string {
        return "Hello, " . static::WHO;
    }
}

class JackGreeter {
    use Greeter;
    private const string WHO = 'Jake';
}
```

Unfortunately PHPStan does not currently take into consideration that self-referencing constants are (usually) meant to be overridden.

This Pull Request omits the Constant Type check for self-referencing constants.


staabm:
interessting. I have never seen this in practice before.

ondrejmirtes:
I've seen this before in the issue tracker but can't find it now. However, this is an awful hack and I don't plan to support it. Your code is broken. Use abstract method instead.

PR #3838, opened 2025-02-26
Implement FinalPrivateConstantRule

staabm:
as requested in https://github.com/phpstan/phpstan-src/pull/3830#pullrequestreview-2620415500

### DIFF
--- a/src/Rules/Constants/FinalPrivateConstantRule.php
+++ b/src/Rules/Constants/FinalPrivateConstantRule.php
@@ -37,7 +37,7 @@ final class FinalPrivateConstantRule implements Rule
 		$errors = [];
 		foreach ($node->consts as $classConstNode) {
 			$errors[] = RuleErrorBuilder::message(sprintf(
-				'Private constant %s::%s() cannot be final as it is not visible to other classes.',
+				'Private constant %s::%s() cannot be final as it is never overridden by other classes.',
 				$classReflection->getDisplayName(),
 				$classConstNode->name->name,
 			))->identifier('classConstant.finalPrivate')->build();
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/Constants/FinalPrivateConstantRuleTest.php
```diff
@@ -0,0 +1,27 @@
...
+	protected function getRule(): Rule
+	{
+		return new FinalPrivateConstantRule();
+	}
+
+	public function testRule(): void
+	{
+		$this->analyse([__DIR__ . '/data/final-private-const.php'], [
+			[
+				'Private constant FinalPrivateConstants\User::FINAL_PRIVATE() cannot be final as it is not visible to other classes.',
```
staabm: used the error message from https://3v4l.org/oBYII

src/Rules/Constants/FinalPrivateConstantRule.php
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		if (!$node->isPrivate()) {
+			return [];
+		}
+
+		$errors = [];
+		foreach ($node->consts as $classConstNode) {
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'Private constant %s::%s() cannot be final as it is not visible to other classes.',
```
ondrejmirtes: Please follow the method wording: Private method HelloWorld::sayHello() cannot be final as it is never overridden by other classes.

ondrejmirtes (changes requested):

src/Rules/Constants/FinalPrivateConstantRule.php
```diff
@@ -0,0 +1,49 @@
...
+			return [];
+		}
+
+		$errors = [];
+		foreach ($node->consts as $classConstNode) {
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'Private constant %s::%s() cannot be final as it is never overridden by other classes.',
+				$classReflection->getDisplayName(),
+				$classConstNode->name->name,
+			))->identifier('classConstant.finalPrivate')->build();
```
ondrejmirtes: should be non-ignorable

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #3839, opened 2025-02-26
Hooked properties cannot be both abstract and private

staabm:
as requested in https://github.com/phpstan/phpstan-src/pull/3830#issuecomment-2685492387


ondrejmirtes:
Thank you!

PR #3842, opened 2025-02-28
Hooked properties cannot be unset()

staabm:
refs https://github.com/phpstan/phpstan/issues/12336

### DIFF
--- a/src/Rules/Variables/UnsetRule.php
+++ b/src/Rules/Variables/UnsetRule.php
@@ -4,6 +4,7 @@ namespace PHPStan\Rules\Variables;
 
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
+use PHPStan\Php\PhpVersion;
 use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\Properties\PropertyReflectionFinder;
 use PHPStan\Rules\Rule;
@@ -20,6 +21,7 @@ final class UnsetRule implements Rule
 
 	public function __construct(
 		private PropertyReflectionFinder $propertyReflectionFinder,
+		private PhpVersion $phpVersion,
 	)
 	{
 	}
@@ -115,6 +117,23 @@ final class UnsetRule implements Rule
 					->line($node->getStartLine())
 					->identifier('unset.hookedProperty')
 					->build();
+			} elseif ($this->phpVersion->supportsPropertyHooks()) {
+				if (
+					!$propertyReflection->isPrivate()
+					&& !$propertyReflection->isFinal()->yes()
+					&& !$propertyReflection->getDeclaringClass()->isFinal()
+				) {
+					return RuleErrorBuilder::message(
+						sprintf(
+							'Cannot unset %s::$%s property which might get hooked in subclass.',
+							$propertyReflection->getDeclaringClass()->getDisplayName(),
+							$foundPropertyReflection->getName(),
+						),
+					)
+						->line($node->getStartLine())
+						->identifier('unset.maybeHookedProperty')
+						->build();
+				}
 			}
 		}
 
--- a/src/Rules/Variables/UnsetRule.php
+++ b/src/Rules/Variables/UnsetRule.php
@@ -125,7 +125,7 @@ final class UnsetRule implements Rule
 				) {
 					return RuleErrorBuilder::message(
 						sprintf(
-							'Cannot unset %s::$%s property which might get hooked in subclass.',
+							'Cannot unset property %s::$%s because it might have hooks in a subclass.',
 							$propertyReflection->getDeclaringClass()->getDisplayName(),
 							$foundPropertyReflection->getName(),
 						),
--- a/src/Rules/Variables/UnsetRule.php
+++ b/src/Rules/Variables/UnsetRule.php
@@ -131,7 +131,7 @@ final class UnsetRule implements Rule
 						),
 					)
 						->line($node->getStartLine())
-						->identifier('unset.maybeHookedProperty')
+						->identifier('unset.possiblyHookedProperty')
 						->build();
 				}
 			}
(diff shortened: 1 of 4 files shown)


src/Rules/Variables/UnsetRule.php
```diff
@@ -103,6 +103,19 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
 					->identifier($propertyReflection->isReadOnly() ? 'unset.readOnlyProperty' : 'unset.readOnlyPropertyByPhpDoc')
 					->build();
 			}
+
+			if ($propertyReflection->getNativeReflection()->getHooks() !== []) {
```
ondrejmirtes: I'd rather be if you asked this on `$propertyReflection` directly. You can add a method `isHooked(): bool` which does `$this->getHook !== null || $this->setHook !== null`.

ondrejmirtes (changes requested):

src/Rules/Variables/UnsetRule.php
```diff
@@ -103,6 +103,19 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
...
+				return RuleErrorBuilder::message(
+					sprintf(
+						'Cannot unset hooked %s::$%s property.',
+						$propertyReflection->getDeclaringClass()->getDisplayName(),
+						$foundPropertyReflection->getName(),
+					),
+				)
+					->line($node->getStartLine())
+					->identifier('unset.hookedProperty')
+					->build();
```
staabm, line 119: non-ignorable?

ondrejmirtes: Rather not, because the error depends on the type of the expr the property is fetched on. And that might be wrong, so it should be ignorable.

ondrejmirtes (changes requested):
One more thing - if we're on PHP 8.4+ (phpVersion::supportsPropertyHooks), and the property is not final, and the declaring class is not final, and the property is not private, report a similar error even if the property does not have hooks. Preferably with a different explaining error message. Because a subclass might create a hooked declaration for the same property.

We're already doing a similar thing here (accessing/writing a property that might be hooked in a subclass can throw any exception): https://github.com/phpstan/phpstan-src/blob/3d0a6583037b8bd68909e56523d026e173332979/src/Analyser/NodeScopeResolver.php#L4317-L4331

staabm:
> One more thing - if we're on PHP 8.4+ (phpVersion::supportsPropertyHooks), and the property is not final, and the declaring class is not final, and the property is not private, report a similar error even if the property does not have hooks. Preferably with a different explaining error message. Because a subclass might create a hooked declaration for the same property.

this sounds like a edge case which should only be reported via a opt-in flag..?

staabm:
while testing I found https://phpstan.org/r/0299facc-d23d-4ed8-9567-e1c7c8c30536 which I think should not report a error on line 9 on PHP 8.4, right?

https://3v4l.org/ZD4oE

tests/PHPStan/Rules/Variables/UnsetRuleTest.php
```diff
@@ -60,12 +61,16 @@ public function testBug2752(): void
...
 	public function testBug4289(): void
 	{
-		$this->analyse([__DIR__ . '/data/bug-4289.php'], [
-			[
-				'Cannot unset Bug4289\BaseClass::$fields property which might get hooked in subclass.',
-				25,
-			],
-		]);
+		if (PHP_VERSION_ID < 80400) {
+			$this->analyse([__DIR__ . '/data/bug-4289.php'], []);
```
ondrejmirtes: We usually do this by assigning `$errors` to `[]` and then to something on-empty in an if branch. So that there's only a single `analyse` call in the body.

src/Rules/Variables/UnsetRule.php
```diff
@@ -103,6 +105,36 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
...
+				) {
+					return RuleErrorBuilder::message(
+						sprintf(
+							'Cannot unset %s::$%s property which might get hooked in subclass.',
+							$propertyReflection->getDeclaringClass()->getDisplayName(),
+							$foundPropertyReflection->getName(),
+						),
+					)
+						->line($node->getStartLine())
+						->identifier('unset.maybeHookedProperty')
```
ondrejmirtes: I usually use `possibly` instead of `maybe` in an identifier

src/Rules/Variables/UnsetRule.php
```diff
@@ -103,6 +105,36 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
...
+					->build();
+			} elseif ($this->phpVersion->supportsPropertyHooks()) {
+				if (
+					!$propertyReflection->isPrivate()
+					&& !$propertyReflection->isFinal()->yes()
+					&& !$propertyReflection->getDeclaringClass()->isFinal()
+				) {
+					return RuleErrorBuilder::message(
+						sprintf(
+							'Cannot unset %s::$%s property which might get hooked in subclass.',
```
ondrejmirtes:
```suggestion
							'Cannot unset property %s::$%s because it might have hooks in a subclass.',
```

ondrejmirtes (changes requested):

src/Rules/Variables/UnsetRule.php
```diff
@@ -103,6 +105,36 @@ private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
...
+				)
+					->line($node->getStartLine())
+					->identifier('unset.hookedProperty')
+					->build();
+			} elseif ($this->phpVersion->supportsPropertyHooks()) {
+				if (
+					!$propertyReflection->isPrivate()
+					&& !$propertyReflection->isFinal()->yes()
+					&& !$propertyReflection->getDeclaringClass()->isFinal()
+				) {
```
staabm, line 125: I also though about adding 

```php
public function isHookedInSubclass(): TrinaryLogic {
		if (
			!$this->isPrivate()
			&& !$this->isFinal()->yes()
			&& !$this->getDeclaringClass()->isFinal()
		) {
		}

	}
```

to `PhpPropertyReflection`. wdyt?

ondrejmirtes: Not needed right now :) I'd have to this about that more.

ondrejmirtes:
Thank you!

PR #3845, opened 2025-03-02
Fix GetNonVirtualPropertyHookReadRule on abstract property

staabm:
I found https://phpstan.org/r/0299facc-d23d-4ed8-9567-e1c7c8c30536 which I think should not report a error on line 9 on PHP 8.4

https://3v4l.org/ZD4oE

### DIFF
--- a/src/Rules/Properties/GetNonVirtualPropertyHookReadRule.php
+++ b/src/Rules/Properties/GetNonVirtualPropertyHookReadRule.php
@@ -93,6 +93,10 @@ final class GetNonVirtualPropertyHookReadRule implements Rule
 				continue;
 			}
 
+			if ($propertyReflection->isAbstract()->yes()) {
+				continue;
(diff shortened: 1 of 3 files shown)


src/Rules/Properties/GetNonVirtualPropertyHookReadRule.php
```diff
@@ -93,6 +93,10 @@ public function processNode(Node $node, Scope $scope): array
 				continue;
 			}
 
+			if ($propertyReflection->isAbstract()->yes()) {
```
ondrejmirtes: I feel like this is a wrong fix. Instead, the foreach over the hooks above should check if the get hook has a body or not.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3846, opened 2025-03-02
Add `stream_socket_client`’s stub

MatTheCat:
Should fix https://github.com/phpstan/phpstan/issues/12622

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -284,6 +284,17 @@ function str_ireplace($search, $replace, $subject, ?int &$count = null) {}
 function stream_select(?array &$read, ?array &$write, ?array &$except, ?int $seconds, ?int $microseconds = null) {}
 
 /**
+ * @param int $error_code
+ * @param-out int $error_code
+ * @param string $error_message
+ * @param-out string $error_message
+ * @param int-mask<1, 2, 4> $flags
+ * @param resource|null $context
+ * @return resource|false
+ */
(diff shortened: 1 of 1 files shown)


stubs/core.stub
```diff
@@ -283,6 +283,17 @@ function str_ireplace($search, $replace, $subject, ?int &$count = null) {}
  */
 function stream_select(?array &$read, ?array &$write, ?array &$except, ?int $seconds, ?int $microseconds = null) {}
 
+/**
+ * @param int $error_code
+ * @param-out int $error_code
+ * @param string $error_message
+ * @param-out string $error_message
```
ondrejmirtes, line 290: What are these set to if the call does not fail?

ondrejmirtes (changes requested):

stubs/core.stub
MatTheCat: `$error_code` will be zero and `$error_message` an empty string.

```shell
$ php -r 'var_dump(stream_socket_client("udp://127.0.0.1:13", $errorCode, $errorMessage), $errorCode, $errorMessage);'
resource(5) of type (stream)
int(0)
string(0) ""
```

EDIT it seems you can get those values even if the call fails: https://3v4l.org/TmSQQ

MatTheCat:
@ondrejmirtes I’m not sure what changes you requested?

PR #3850, opened 2025-03-03
@readonly property cannot be passed by-ref

staabm:
closes https://github.com/phpstan/phpstan/issues/12676


ondrejmirtes:
This is fine because it's going to be very rare and arguably it's an ommission in a feature added in the past, but for future additions we need to start adding them only to bleeding edge again :)

ondrejmirtes:
Thank you.

PR #3855, opened 2025-03-06
Fix unsetting array item triggers unset.possiblyHookedProperty

staabm:
closes https://github.com/phpstan/phpstan/issues/12695

### DIFF
--- a/src/Rules/Variables/UnsetRule.php
+++ b/src/Rules/Variables/UnsetRule.php
@@ -37,6 +37,67 @@ final class UnsetRule implements Rule
 		$errors = [];
 
 		foreach ($functionArguments as $argument) {
+			if (
+				$argument instanceof Node\Expr\PropertyFetch
+				&& $argument->name instanceof Node\Identifier
+			) {
+				$foundPropertyReflection = $this->propertyReflectionFinder->findPropertyReflectionFromNode($argument, $scope);
+				if ($foundPropertyReflection === null) {
+					continue;
+				}
+
+				$propertyReflection = $foundPropertyReflection->getNativeReflection();
+				if ($propertyReflection === null) {
+					continue;
+				}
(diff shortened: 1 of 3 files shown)


src/Rules/Variables/UnsetRule.php
```diff
@@ -48,7 +48,7 @@ public function processNode(Node $node, Scope $scope): array
 		return $errors;
 	}
 
-	private function canBeUnset(Node $node, Scope $scope): ?IdentifierRuleError
+	private function canBeUnset(Node $node, Scope $scope, bool $inArrayDimFetch = false): ?IdentifierRuleError
```
ondrejmirtes: More logical solution is to exclude `$node instanceof Node\Expr\PropertyFetch` section completely out of `canBeUnset` recursion and just check it directly in processNode.

The other part of this about unsetting on readonly property is also wrong so we need to fix that too:

* https://phpstan.org/r/5c6651e5-cfaf-4688-8b51-0f9ac27c8495
* https://3v4l.org/vONrk#v8.4.4

ondrejmirtes (changes requested):

src/Rules/Variables/UnsetRule.php
staabm: interessting.. 

> Readonly properties cannot be [unset()](https://www.php.net/manual/en/function.unset.php) once they are initialized. However, it is possible to unset a readonly property prior to initialization, from the scope where the property has been declared.

https://www.php.net/manual/en/language.oop5.properties.php

ondrejmirtes:
I force-pushed what I meant instead. THank you.

staabm:
I see now. didn't got it in the first place.

ondrejmirtes:
Thank you!

PR #3856, opened 2025-03-06
Fix typo false-positive

staabm:
should fix build error

```
error: `Excluder` should be `Excluded`
  --> src/DependencyInjection/ValidateIgnoredErrorsExtension.php:159:14
    |
159 |      if (FileExcluder::isAbsolutePath($ignorePath)) {
    |              ^^^^^^^^
    |
```


ondrejmirtes:
Thank you. Please also send a similar fix to phpstan/phpstan: https://github.com/phpstan/phpstan/actions/runs/13677410548/job/38240946978

PR #3857, opened 2025-03-06
Detect accessing static property as non static

jnoordsij:
Closes https://github.com/phpstan/phpstan/issues/12692

Thanks to the pointers provided on the issue, I've managed to copy over some bits from the 'reverse' check and adjust them accordingly:
https://github.com/phpstan/phpstan-src/blob/8d4796d28975b9c60399b228fe23decbe7ef7a1f/src/Rules/Properties/AccessStaticPropertiesRule.php#L213-L228

This seems to work fine, but maybe there's still some configuration to be accounted for, given that while incorrect static access will result in a fatal error, incorrect non static access will only emit an info + warning message. As a consequence maybe this check should be considered less critical/important.

### DIFF
--- a/src/Rules/Properties/AccessPropertiesCheck.php
+++ b/src/Rules/Properties/AccessPropertiesCheck.php
@@ -163,10 +163,10 @@ final class AccessPropertiesCheck
 		if ($propertyReflection->isStatic()) {
 			return [
 				RuleErrorBuilder::message(sprintf(
-					'Non static access to static property %s::$%s.',
+					'Non-static access to static property %s::$%s.',
 					$propertyReflection->getDeclaringClass()->getDisplayName(),
 					$name,
-				))->identifier('property.nonStaticAccess')->build(),
+				))->identifier('staticProperty.nonStaticAccess')->build(),
 			];
 		}
 
--- a/src/Rules/Properties/AccessPropertiesCheck.php
+++ b/src/Rules/Properties/AccessPropertiesCheck.php
@@ -160,6 +160,16 @@ final class AccessPropertiesCheck
 		}
 
 		$propertyReflection = $type->getProperty($name, $scope);
+		if ($propertyReflection->isStatic()) {
+			return [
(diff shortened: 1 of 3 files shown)


src/Rules/Properties/AccessPropertiesCheck.php
```diff
@@ -160,6 +160,16 @@ private function processSingleProperty(Scope $scope, PropertyFetch $node, string
 		}
 
 		$propertyReflection = $type->getProperty($name, $scope);
+		if ($propertyReflection->isStatic()) {
+			return [
+				RuleErrorBuilder::message(sprintf(
+					'Non static access to static property %s::$%s.',
```
ondrejmirtes:
```suggestion
					'Non-static access to static property %s::$%s.',
```

src/Rules/Properties/AccessPropertiesCheck.php
```diff
@@ -160,6 +160,16 @@ private function processSingleProperty(Scope $scope, PropertyFetch $node, string
...
+					$propertyReflection->getDeclaringClass()->getDisplayName(),
+					$name,
+				))->identifier('property.nonStaticAccess')->build(),
```
ondrejmirtes:
```suggestion
				))->identifier('staticProperty.nonStaticAccess')->build(),
```

ondrejmirtes (changes requested):
Please do not re-number test lines like that. Put the new test at the bottom of the file. Or even better - add a completely new test method with a completely new test file. It can be `testBug12692` / `bug-12692.php`. Thanks

src/Rules/Properties/AccessPropertiesCheck.php
jnoordsij: For refrence: I stuck with `non static` over `non-static` to match the PHP runtime message, but I do agree this variant is more readable.

ondrejmirtes:
Thank you!

PR #3858, opened 2025-03-06
Replace error-prone instanceof in Rules classes

zonuexe:
Reduce usage of `$type instanceof *Type` pattern in Rules classes.

refs https://github.com/phpstan/phpstan/issues/8311


src/Rules/TooWideTypehints/TooWideMethodReturnTypehintRule.php
```diff
@@ -82,9 +82,9 @@ public function processNode(Node $node, Scope $scope): array
 
 		$returnType = TypeCombinator::union(...$returnTypes);
 		if (
-			!$method->isPrivate()
-			&& ($returnType->isNull()->yes() || $returnType instanceof ConstantBooleanType)
-			&& !$isFirstDeclaration
+			!$isFirstDeclaration
+			&& !$method->isPrivate()
+			&& ($returnType->isNull()->yes() || $returnType->isTrue()->yes() || $returnType->isFalse()->yes())
```
zonuexe, line 86: Logical operators are short-circuit evaluated, so rearranging expressions can lead to micro-optimizations.

ondrejmirtes (changes requested):
I don't want to replace these calls by just accessing the first element. Union types should be properly handled. Multiple values should be iterated over.

src/Rules/Api/NodeConnectingVisitorAttributesRule.php
```diff
@@ -42,10 +43,11 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 		$argType = $scope->getType($args[0]->value);
-		if (!$argType instanceof ConstantStringType) {
+		if (count($argType->getConstantStrings()) === 0) {
```
staabm: maybe better

```suggestion
		if (count($argType->getConstantStrings()) !== 1) {
```

instead of ignoring possible known values

zonuexe:
Since `ConstantArrayType` and `ConstantStringType` each return `return [$this];` from the `getConstant*()` methods, I thought it would be effective to simply replace them with `count() === 0` or `count() === 1` to minimize the change difference, but I also prefer to replace it with `foreach` instead of `$types[0]`, so I will change my approach and fix it that way.

src/Rules/Api/NodeConnectingVisitorAttributesRule.php
zonuexe: Thanks!

src/Rules/PhpDoc/RequireExtendsCheck.php
```diff
@@ -41,39 +43,48 @@ public function checkExtendsTags(Node $node, array $extendsTags): array
...
 				$errors[] = RuleErrorBuilder::message(sprintf('PHPDoc tag @phpstan-require-extends contains non-object type %s.', $type->describe(VerbosityLevel::typeOnly())))
 					->identifier('requireExtends.nonObject')
 					->build();
 				continue;
 			}
 
-			$class = $type->getClassName();
-			$referencedClassReflection = $type->getClassReflection();
+			$classNames = $type->getObjectClassNames();
+			sort($classNames);
```
zonuexe, line 57: I am sorting this array because tests seem to fail randomly.

 * https://github.com/phpstan/phpstan-src/actions/runs/13726816592/job/38395109333?pr=3858

zonuexe:
`MissingTypehintCheck` was reverted in this branch because the literal substitution seemed to change its behavior. https://github.com/phpstan/phpstan-src/pull/3858/commits/a408a580f8298050a714ae7f6004679f445e3cf0

src/Rules/PhpDoc/RequireExtendsCheck.php
```diff
@@ -41,39 +43,48 @@ public function checkExtendsTags(Node $node, array $extendsTags): array
...
+			if (count($type->getObjectClassNames()) === 0) {
 				$errors[] = RuleErrorBuilder::message(sprintf('PHPDoc tag @phpstan-require-extends contains non-object type %s.', $type->describe(VerbosityLevel::typeOnly())))
 					->identifier('requireExtends.nonObject')
 					->build();
 				continue;
 			}
 
-			$class = $type->getClassName();
-			$referencedClassReflection = $type->getClassReflection();
+			$classNames = $type->getObjectClassNames();
```
staabm: Introduce this var earlier so you don't need to call `getObjectClassNames` twice (methods called on Type can get expensive)

src/Rules/PhpDoc/RequireImplementsDefinitionTraitRule.php
```diff
@@ -49,34 +51,38 @@ public function processNode(Node $node, Scope $scope): array
 		$errors = [];
 		foreach ($implementsTags as $implementsTag) {
 			$type = $implementsTag->getType();
-			if (!$type instanceof ObjectType) {
+			if (count($type->getObjectClassNames()) === 0) {
```
staabm: Introduce a var so you don't need to call `getObjectClassNames` twice

src/Rules/Variables/CompactVariablesRule.php
```diff
@@ -66,25 +66,23 @@ public function processNode(Node $node, Scope $scope): array
...
 
 	/**
-	 * @return array<int, ConstantStringType>
+	 * @return list<ConstantStringType>
 	 */
 	private function findConstantStrings(Type $type): array
 	{
-		if ($type instanceof ConstantStringType) {
-			return [$type];
+		if (count($type->getConstantStrings()) > 0) {
```
staabm: Introduce a var so you don't need to call `getConstantStrings` twice

src/Rules/Arrays/DuplicateKeysInLiteralArraysRule.php
```diff
@@ -70,38 +70,37 @@ public function processNode(Node $node, Scope $scope): array
 				}
 			} else {
 				$keyType = $itemNode->getScope()->getType($key);
-
-				$arrayKeyValue = $keyType->toArrayKey();
-				if ($arrayKeyValue instanceof ConstantIntegerType) {
+				$arrayKeyValues = $keyType->toArrayKey()->getConstantScalarValues();
+				if (count($arrayKeyValues) === 1 && is_int($arrayKeyValues[0])) {
```
VincentLanglet, line 74: Wouldn't it better/possible to handle union of scalar values here (?)

src/Rules/Comparison/ConstantLooseComparisonRule.php
```diff
@@ -57,7 +57,7 @@ public function processNode(Node $node, Scope $scope): array
 			return $ruleErrorBuilder->treatPhpDocTypesAsCertainTip();
 		};
 
-		if (!$nodeType->getValue()) {
+		if (in_array(false, $nodeType->getConstantScalarValues(), true)) {
```
VincentLanglet: I'm not sure but isn't `$nodeType->isFalse()->yes()` better ?

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -68,12 +69,12 @@ public function findSpecifiedType(
 				$functionName = strtolower((string) $node->name);
 				if ($functionName === 'assert' && $argsCount >= 1) {
 					$arg = $node->getArgs()[0]->value;
-					$assertValue = ($this->treatPhpDocTypesAsCertain ? $scope->getType($arg) : $scope->getNativeType($arg))->toBoolean();
-					if (!$assertValue instanceof ConstantBooleanType) {
+					$assertValues = ($this->treatPhpDocTypesAsCertain ? $scope->getType($arg) : $scope->getNativeType($arg))->getConstantScalarValues();
+					if (count($assertValues) !== 1 || !is_bool($assertValues[0])) {
```
VincentLanglet: You could maybe have kept the `toBoolean` and checked
```
$assertValue->isTrue()->yes() || $assertValue->isFalse()->yes()
```

src/Rules/Comparison/ConstantLooseComparisonRule.php
zonuexe: That's correct, thanks!

src/Rules/Variables/CompactVariablesRule.php
zonuexe: 👍

src/Rules/Arrays/DuplicateKeysInLiteralArraysRule.php
zonuexe: Do you mean the following?

```diff
modified   src/Rules/Arrays/DuplicateKeysInLiteralArraysRule.php
@@ -71,10 +71,12 @@ final class DuplicateKeysInLiteralArraysRule implements Rule
 			} else {
 				$keyType = $itemNode->getScope()->getType($key);
 				$arrayKeyValues = $keyType->toArrayKey()->getConstantScalarValues();
-				if (count($arrayKeyValues) === 1 && is_int($arrayKeyValues[0])) {
-					$autoGeneratedIndex = $autoGeneratedIndex === null
-						? $arrayKeyValues[0]
-						: max($autoGeneratedIndex, $arrayKeyValues[0]);
+				foreach ($arrayKeyValues as $arrayKeyValue) {
+					if (is_int($arrayKeyValue)) {
+						$autoGeneratedIndex = $autoGeneratedIndex === null
+							? $arrayKeyValue
+							: max($autoGeneratedIndex, $arrayKeyValue);
+					}
 				}
 			}
```

<details><summary>Or is it type-level arithmetic?</summary>

```diff
modified   src/Rules/Arrays/DuplicateKeysInLiteralArraysRule.php
@@ -42,7 +42,6 @@ final class DuplicateKeysInLiteralArraysRule implements Rule
 		$valueLines = [];
 
 		/**
-		 * @var int|false|null $autoGeneratedIndex
 		 * - An int value represent the biggest integer used as array key.
 		 *   When no key is provided this value + 1 will be used.
 		 * - Null is used as initializer instead of 0 to avoid issue with negative keys.
@@ -63,27 +62,37 @@ final class DuplicateKeysInLiteralArraysRule implements Rule
 				}
 
 				if ($autoGeneratedIndex === null) {
-					$autoGeneratedIndex = 0;
-					$keyType = new ConstantIntegerType(0);
+					$autoGeneratedIndex = new ConstantIntegerType(0);
 				} else {
-					$keyType = new ConstantIntegerType(++$autoGeneratedIndex);
+					$autoGeneratedIndex = $scope->getType(new Plus(
+						new Int_(1),
+						new TypeExpr($autoGeneratedIndex),
+					));
 				}
+				$keyType = $autoGeneratedIndex;
 			} else {
 				$keyType = $itemNode->getScope()->getType($key);
-				$arrayKeyValues = $keyType->toArrayKey()->getConstantScalarValues();
-				if (count($arrayKeyValues) === 1 && is_int($arrayKeyValues[0])) {
+				$arrayKeyType = $keyType->toArrayKey();
+				if ($autoGeneratedIndex !== false && $arrayKeyType->isInteger()->yes()) {
 					$autoGeneratedIndex = $autoGeneratedIndex === null
-						? $arrayKeyValues[0]
-						: max($autoGeneratedIndex, $arrayKeyValues[0]);
+						? $keyType
+						: $scope->getType(new FuncCall(
+							new Name('\max'),
+							[
+								new Arg(new TypeExpr($autoGeneratedIndex)),
+								new Arg(new TypeExpr($arrayKeyType)),
+							]
+						));
 				}
 			}
```

</details>

`$autoGeneratedIndex` is explained as follows:

```php
		/**
		 * - An int value represent the biggest integer used as array key.
		 *   When no key is provided this value + 1 will be used.
		 * - Null is used as initializer instead of 0 to avoid issue with negative keys.
		 * - False means a non-scalar value was encountered and we cannot be sure of the next keys.
		 */
		$autoGeneratedIndex = null;
```

Since the key value [actually changes depending on the input value](https://php-play.dev/?c=DwfgDgFmBQD0BU8AEABMBDATugtkglgHYAuwAjADRJkAMAfEgCRHFLyzQBmAroQMbF8Ae0JJOQoQAoWTFgEoAXEgBuQ-ABMkAb2hI9TdEgC8SANrQAkFeYljDAMz2KlqzTvVnViwCYqsWEgAQgCmfOjcAM7BSEKcsrYA1sEAnlTEENFJySroADbc0WGEhEKsAEaZJQDuhAB0LhZk7k6WALoA3Lr6oRBCSABEWjbEAL7u-VRY2MkA%2BlkRkozocqberVT9ADqE-Z0j0FxCmMHofBBIktiEAObBkpTUNHJI6BFMhM86%2BmISix97QA&v=8.4&f=console), more appropriate typing is possible with type operations. However, since `DuplicateKeysInLiteralArraysRule` has some subtle behavior, I would like to keep the changes in this branch to a minimum.

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -69,12 +68,13 @@ public function findSpecifiedType(
...
-					$assertValues = ($this->treatPhpDocTypesAsCertain ? $scope->getType($arg) : $scope->getNativeType($arg))->getConstantScalarValues();
-					if (count($assertValues) !== 1 || !is_bool($assertValues[0])) {
+					$assertValue = ($this->treatPhpDocTypesAsCertain ? $scope->getType($arg) : $scope->getNativeType($arg))->toBoolean();
+					if (! $assertValue->isTrue()->yes() && ! $assertValue->isFalse()->yes()) {
 						return null;
 					}
 
-					return $assertValues[0];
+					/** @var bool */
+					return $assertValue->getConstantScalarValues()[0];
```
VincentLanglet: You can return `$assertValue->isTrue()->yes()` if you want to avoid the var phpdoc

zonuexe: Makes sense!

ondrejmirtes:
I'd love to merge this, but please fix the conflicts.

ondrejmirtes:
I was able to rebase this myself 🎉

ondrejmirtes:
Thank you.

PR #3859, opened 2025-03-07
Reduce method calls in ExpressionTypeHolder

staabm:
while looking into https://github.com/phpstan/phpstan/issues/12647 I noticed `ExpressionTypeHolder->and` is on a very hot path. 

grafik

lets micro optimize by remove unnecessary method calls in this path

before
```
➜  PhpSpreadsheet git:(master) ✗ hyperfine 'php ../phpstan-src/bin/phpstan analyze src/PhpSpreadsheet/Reader/Xlsx.php --debug' # before
  Time (mean ± σ):      8.948 s ±  0.020 s    [User: 8.774 s, System: 0.170 s]
  Range (min … max):    8.913 s …  8.975 s    10 runs
```

after
```
➜  PhpSpreadsheet git:(master) ✗ hyperfine 'php ../phpstan-src/bin/phpstan analyze src/PhpSpreadsheet/Reader/Xlsx.php --debug' # after
  Time (mean ± σ):      8.775 s ±  0.011 s    [User: 8.608 s, System: 0.163 s]
  Range (min … max):    8.751 s …  8.790 s    10 runs
```

### DIFF
--- a/src/Analyser/ExpressionTypeHolder.php
+++ b/src/Analyser/ExpressionTypeHolder.php
@@ -35,15 +35,15 @@ final class ExpressionTypeHolder
 
 	public function and(self $other): self
 	{
-		if ($this->getType()->equals($other->getType())) {
-			$type = $this->getType();
+		if ($this->type->equals($other->type)) {
+			$type = $this->type;
 		} else {
-			$type = TypeCombinator::union($this->getType(), $other->getType());
+			$type = TypeCombinator::union($this->type, $other->type);
 		}
 		return new self(
 			$this->expr,
 			$type,
-			$this->getCertainty()->and($other->getCertainty()),
+			$this->certainty->and($other->certainty),
 		);
 	}
 


ondrejmirtes:
Usually we can achieve a much better optimization by reducing the number of calls, not optimizating how fast it's going to run when it's called 2 million times. Can you please investigate why is it called so often? Usually there's some other optimization to be had because of huge combinatorial explosion in types or something like that.

staabm:
I am still looking into the case, but I think this method is called so often that the PR is still worth it - even for a lot of other profiles with big unions.
(in addition it helps getting better profiles with blackfire, because the profiler is very sensitive to function/method calls)

its 50-100ms on a single file, so it easily adds up when analying projects with lots of files

ondrejmirtes:
Thank you! But we should still investigate why is it called so many times, and avoid that :)

staabm:
thank you.

debugged the `src/PhpSpreadsheet/Reader/Xlsx.php` case further:

there is one union involved with more then 30 elements, which is resolved from the phpdoc `ZipArchive::ER_*` (from phpstan core signatures for ZipArchive).

the 2nd biggest union has 8 elements and is resolved from this phpdoc
https://github.com/PHPOffice/PhpSpreadsheet/blob/3ef8ec4b1a095f700412c6e44f15b3eaedae23fe/src/PhpSpreadsheet/Worksheet/Worksheet.php#L2713

-> there is no really huge union involved it seems. we spent most of the time with merging scope objects

profile can be seen here: https://blackfire.io/profiles/0293ba31-80fc-403f-9452-bfda15b7c7c7/graph

-> I am still investigating

PR #3863, opened 2025-03-09
Added regression test

staabm:
we did not yet implement the feature requested in bug10717, but at least a bug was fixed which is mentioned in the https://github.com/phpstan/phpstan/issues/10717#issuecomment-2002515197


staabm:
//cc @VincentLanglet

tests/PHPStan/Analyser/nsrt/bug-10717.php
```diff
@@ -0,0 +1,1052 @@
...
+}
+
+function test(string $code): void
+{
+	$country = Languages::DATA[$code]['country'];
+
+	if ($country === 'fo' || $country === 'Faroese' || $country === 'Føroyskt') {
+		// foo
+	} else {
+		assertType('(bool|string)', $country);
```
staabm, line 1049: here is the type assertion. everything else above is just a oversized array of data

ondrejmirtes:
Thank you.

PR #3865, opened 2025-03-09
Try fixing assigning array offsets on possibly undefined nested array offsets

ondrejmirtes:


ondrejmirtes:
```php
<?php

function doFoo(int $i): void
{
	$array = [];

	$array[$i]['bar'] = 1;
	$array[$i]['baz'] = 2;

	// on line 10 PHPStan assigns:
	// 1) $array[$i]['baz'] to 2
	// 2) $array[$i] to array{bar: 1, baz: 2}
	// 3) $array = from non-empty-array<int, array{bar: 1}> to non-empty-array<int, array{bar: 1, baz: 2}>
	// but step 3) is basically wrong for most code because assigning "baz" might be on a different offset than array{bar: 1} is on
	// but somehow we need to continue doing this if this PR is supposed to be successful

	assertType('non-empty-array<int, array{bar: 1, baz: 2}>', $array);
}

```

ondrejmirtes:
I realized the above problem could be fixed with a new accessory "ArraySize" type. When doing `$array[$i]['baz'] = 2;` and by knowing that `$array[$i]` already exists and by knowing that `$array` has size `1`, we could overwrite the array with a new type. So instead of:

```
non-empty-array<int, array{bar: 1, baz?: 2}>
```

We could actually end up with:

```
non-empty-array<int, array{bar: 1, baz: 2}>
```

VincentLanglet:
All the issues listed in this PR are closed
Related to the PR:
https://github.com/phpstan/phpstan-src/pull/3874
https://github.com/phpstan/phpstan-src/pull/4393

PR #2037, opened 2022-11-30
Supports functions whose side effects are flipped by parameters

zonuexe:
closes https://github.com/phpstan/phpstan/issues/8440

PR #3867, opened 2025-03-10
Mark ob_*() functions has side-effects

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/12577

All [`ob_*()` functions](https://www.php.net/manual/ref.outcontrol.php) are impure.

It's possible to implement state like `clearstatcache()` in scopes, but it probably makes sense to just mark these as `['hasSideEffects' => true]` and leave it at that.


ondrejmirtes:
Please follow instructions here https://github.com/phpstan/phpstan-src/blob/2fe4e0f94e75fe8844a21fdb81799f01f0591dfe/resources/functionMetadata.php#L3-L15

zonuexe:
@ondrejmirtes 

The PhpStorm stub marks [`ob_get_contents()`](https://github.com/JetBrains/phpstorm-stubs/blob/7385d3075dc365911c4a3168fa762de6aa4550c9/standard/standard_8.php#L360-L361) as follows:

```php
#[Pure(true)]
function ob_get_contents(): string|false {}
```

[`#[Pure(true)]`](https://github.com/JetBrains/phpstorm-attributes/blob/64de815a4509c29e00d5e3474087fd24c171afc2/src/Pure.php#L17-L19
) is defined as follows:

> ```php
> class Pure
> {
>     /**
>      * @param bool $mayDependOnGlobalScope Whether the function result may be dependendent on anything except passed variables
>      */
>     public function __construct(bool $mayDependOnGlobalScope = false)
>     {
>     }
> }
> ```

`ob_get_contents()` is clearly an impure function with state, but it doesn't make sense to write `ob_get_contents();` on a separate line. `#[Pure(true)]` means that.

This is the same as the https://github.com/phpstan/phpstan-src/pull/2037 vs https://github.com/phpstan/phpstan-src/commit/2b5b3174c7f00e56b7692b4c26552b2ad13c73d5 conflict about `file_get_contents()`.

According to the current semantics of PHPStan, all `#[Pure(true)]` should be declared as `hasSideEffects => true`, but importing them all seems to break many of PHPStan's tests.

ondrejmirtes:
Not needed anymore thanks to https://github.com/phpstan/phpstan-src/pull/3880

PR #3868, opened 2025-03-10
Fix false positives on non-existing-offset's

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/3766#issuecomment-2708842018 with this PR we detect some more array-dim fetch false-positive cases we see with `reportPossiblyNonexistentGeneralArrayOffset`.

closes https://github.com/phpstan/phpstan/issues/11020

### DIFF
--- a/src/Rules/Arrays/NonexistentOffsetInArrayDimFetchRule.php
+++ b/src/Rules/Arrays/NonexistentOffsetInArrayDimFetchRule.php
@@ -14,6 +14,7 @@ use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
 use function count;
 use function in_array;
+use function is_string;
 use function sprintf;
 
 /**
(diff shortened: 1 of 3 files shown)


staabm:
CI errors seem unrelated, except in shipmonk-rnd/dead-code-detector in which we now correctly no longer emit an error and therfore the `// @phpstan-ignore offsetAccess.notFound` phpdocs no longer have an error to ignore

https://github.com/shipmonk-rnd/dead-code-detector/blob/c218837a440e69f6ac231fd2f3d0c514d02a241f/src/Provider/ReflectionUsageProvider.php#L113C59-L113C96
https://github.com/shipmonk-rnd/dead-code-detector/blob/c218837a440e69f6ac231fd2f3d0c514d02a241f/src/Provider/ReflectionUsageProvider.php#L143

ondrejmirtes:
Thank you!

PR #3869, opened 2025-03-10
Update levels tests

herndlm:
Needs updating after https://github.com/phpstan/phpstan-src/pull/3857

looks fine I suppose, but there are union types involved, so not a 100% sure.


ondrejmirtes:
Thank you!

PR #3870, opened 2025-03-11
Ignore negative keys on checking whether a ConstantArray is a list

watasuke102:
In `ConstantArrayTypeBuilder`, `setOffsetValueType()` uses `ConstantIntegerType` of `offsetType` for checking whether a `ConstantArray` is a list, but doesn't check whether it is positive. Because of that, a `ConstantArray` that has negative integer keys is treated as list.

This PR resolves phpstan/phpstan#12708.

### DIFF
--- a/src/Type/Constant/ConstantArrayTypeBuilder.php
+++ b/src/Type/Constant/ConstantArrayTypeBuilder.php
@@ -162,7 +162,7 @@ final class ConstantArrayTypeBuilder
 				$this->keyTypes[] = $offsetType;
 				$this->valueTypes[] = $valueType;
 
-				if ($offsetType instanceof ConstantIntegerType) {
+				if ($offsetType instanceof ConstantIntegerType && $offsetType->getValue() >= 0) {
 					$min = min($this->nextAutoIndexes);
 					$max = max($this->nextAutoIndexes);
 					if ($offsetType->getValue() > $min) {
--- a/src/Type/Constant/ConstantArrayTypeBuilder.php
+++ b/src/Type/Constant/ConstantArrayTypeBuilder.php
@@ -162,19 +162,25 @@ final class ConstantArrayTypeBuilder
 				$this->keyTypes[] = $offsetType;
 				$this->valueTypes[] = $valueType;
 
-				if ($offsetType instanceof ConstantIntegerType && $offsetType->getValue() >= 0) {
+				if ($offsetType instanceof ConstantIntegerType) {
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
A similar fix needs to be done in ArrayType::isList() - see https://phpstan.org/r/47372278-5264-4f01-a12c-dce00a50539f

ondrejmirtes:
Sorry, I'm wrong.

ondrejmirtes:
Disregard my review there.

src/Type/Constant/ConstantArrayTypeBuilder.php
```diff
@@ -162,7 +162,7 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $opt
 				$this->keyTypes[] = $offsetType;
 				$this->valueTypes[] = $valueType;
 
-				if ($offsetType instanceof ConstantIntegerType) {
+				if ($offsetType instanceof ConstantIntegerType && $offsetType->getValue() >= 0) {
```
ondrejmirtes: I'm worried what happens there because suddenly the `nextAutoIndexes` are not updated for negative numbers.

But they should be updated. The behaviour even changed on PHP 8.3+: https://3v4l.org/DWBH9

ondrejmirtes (changes requested):
Please only check this review. Thanks.

watasuke102:
Oh I requested review by mistake, sorry

watasuke102:
It's ready for review, thank you!

ondrejmirtes:
Thank you.

zonuexe:
Probably not a problem with `ArrayType`, but I found a problem with `ConstantArrayType::isList()` by modifying [Ondřej's case](https://github.com/phpstan/phpstan-src/pull/3870#pullrequestreview-2673207566).

https://phpstan.org/r/78b134e0-88b0-4bbf-be11-032bc267650c

ondrejmirtes:
@zonuexe Array shapes are not currently sealed by default. The most upvoted feature request about it is here on GitHub.

zonuexe:
This problem is also related to [sealed array](https://github.com/phpstan/phpstan/issues/8438), but that alone doesn't prevent input like `[1 => 'x', 0 => 'y']`, so we might need a type like `array{0: string, 1: string}&list`

ondrejmirtes:
I get it, it's complicated 😊 Maybe we could enforce the order of keys along with true sealed-ness of arrays.

PR #3871, opened 2025-03-11
Fix false positives on non-existing-offsets of super-globals

staabm:
see https://phpstan.org/r/1eb6a683-733c-486e-ab5a-e38f2d573093

this PR effectively allows to narrow the array type of a super-global variable in scope

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -552,14 +552,12 @@ final class MutatingScope implements Scope
 			throw new UndefinedVariableException($this, $variableName);
 		}
 
-		$defaultType = new MixedType();
-		if ($this->isGlobalVariable($variableName)) {
-			$defaultType = new ArrayType(new BenevolentUnionType([new IntegerType(), new StringType()]), new MixedType(true));
-		}
-
 		$varExprString = '$' . $variableName;
 		if (!array_key_exists($varExprString, $this->expressionTypes)) {
-			return $defaultType;
+			if ($this->isGlobalVariable($variableName)) {
+				return new ArrayType(new BenevolentUnionType([new IntegerType(), new StringType()]), new MixedType(true));
+			}
+			return new MixedType();
 		}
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you!

herndlm:
oh cool, this seems to be a simpler version of https://github.com/phpstan/phpstan-src/pull/3762 which I can close now :)

staabm:
Ohh.. this wasn't intentional 😅

herndlm:
that's ok, this solution is simpler / more elegant :)

herndlm:
Hi, I remember there was a weird issue when I tried to use my prepared superglobals tests here: https://github.com/phpstan/phpstan-src/pull/3762/files#diff-51a54cb478a71bc6cb00a1276e9df23606258635fe80bfff8843430cc254505c

basically, the truthy/falsey scope bled into the null one. I should have debugged this more.

I believe that is related to this change here causing issues like https://github.com/phpstan/phpstan/issues/12771 and the ones linked to it. @staabm are you available and want to / can look into this? If not I can try later too

herndlm:
hmm, looks like the issue is fixed when I set the expression directly as in https://github.com/phpstan/phpstan-src/pull/3762/files#diff-dc817f2bab8672057a95375a542e68599164f6ab2380e4cccd1b50583e295d85R562-R564, which is kind of the only core difference to this PR.

but I'm not a 100% sure how it even worked before tbh. and if I propose the expression fix, then I could also propose the rest I'm doing in that MR (taking over the expressions when entering functions, classes, namespace)? not sure. an opinion from @ondrejmirtes would be interesting before I prepare anything :)

ondrejmirtes:
This introduced a regression: https://phpstan.org/r/86811bd5-067b-476e-958a-bf621ad773c5

Please look into it, thanks.

staabm:
> This introduced a regression: [phpstan.org/r/86811bd5-067b-476e-958a-bf621ad773c5](https://phpstan.org/r/86811bd5-067b-476e-958a-bf621ad773c5)

for posterity: fixed via https://github.com/phpstan/phpstan-src/pull/3901

PR #3872, opened 2025-03-11
Treat array-shapes without keys as tuples

zonuexe:
refs https://github.com/phpstan/phpstan/issues/12725


zonuexe:
I'm a bit surprised that this single change digs up [list-shapes that were previously imported from Psalm](https://github.com/phpstan/phpstan/issues/8789).

```
2) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "tests/PHPStan/Analyser/nsrt/list-shapes.php" ('/home/runner/work/phpstan-src...es.php')
Failed assertions in /home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Analyser/nsrt/list-shapes.php:

Line 21:
Expected: array{'a'}
Actual:   list{'a'}

Line 22:
Expected: array{'a', 'b'}
Actual:   list{'a', 'b'}

Line 23:
Expected: array{0: 'a', 1?: 'b'}
Actual:   list{0: 'a', 1?: 'b'}
```

ondrejmirtes:
We tried something similar here but decided against it https://github.com/phpstan/phpstan-src/pull/3381

ondrejmirtes:
I don't think we should make any difference between `array{int}` and array{0: int}`. It should be the same thing...

zonuexe:
Closing in favour of the plan in phpstan/phpstan#14939: the functionMap migration continues in #6024 (expanded with `hrtime`, `preg_split`, `sys_getloadavg`), and the semantic change will be re-derived on top of the current sealed-array implementation behind bleeding edge. The approach here — treating any explicit key as indeterminate — turned out to over-demote `array{0: T}` and `array{}`, which must stay lists (see the RFC for the denotational rule).

PR #3873, opened 2025-03-11
Add regression test

herndlm:
For the loop/count case of https://github.com/phpstan/phpstan/issues/1311

### DIFF
--- a/tests/PHPStan/Rules/Properties/TypesAssignedToPropertiesRuleTest.php
+++ b/tests/PHPStan/Rules/Properties/TypesAssignedToPropertiesRuleTest.php
@@ -139,6 +139,11 @@ class TypesAssignedToPropertiesRuleTest extends RuleTestCase
 		]);
 	}
 
+	public function testBug1311(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-1311.php'], []);
+	}
+
 	public function testTypesAssignedToPropertiesExpressionNames(): void
 	{
 		$this->analyse([__DIR__ . '/data/properties-from-array-into-object.php'], [
--- a/tests/PHPStan/Rules/Properties/data/bug-1311.php
+++ b/tests/PHPStan/Rules/Properties/data/bug-1311.php
@@ -1,0 +1,24 @@
+<?php declare(strict_types = 1);
+
+namespace Bug1311;
+
+class HelloWorld
+{
+	/**
+	 * @var array<int, string>
+	 */
+	private $list = [];
+
+	public function convertList(): void
+	{
+		$temp = [1, 2, 3];
+
+		for ($i = 0; $i < count($temp); $i++) {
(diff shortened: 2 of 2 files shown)


tests/PHPStan/Rules/Properties/TypesAssignedToPropertiesRuleTest.php
```diff
@@ -139,6 +139,16 @@ public function testBug1216(): void
 		]);
 	}
 
+	public function testBug1311(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-1311.php'], [
+			[
+				'Property Bug1311\HelloWorld::$list (array<int, string>) does not accept array<int, int>.',
```
herndlm: this is still here in a different format, because the test basically has strict rules enabled and the phpdoc is essentially wrong. not fully sure how to deal with this. see also
https://phpstan.org/r/29982739-327a-494d-bc97-06021030576b vs https://phpstan.org/r/a780f206-b942-4d5e-8be2-e3a79a911964

ondrejmirtes: I don't get it. How does this test have strict-rules enabled? And which parameter from strict-rules is causing this error to appear?

herndlm: tbh I was hoping you would tell me 😅 looks like it's coming from the $strictTypes here: https://github.com/phpstan/phpstan-src/blob/c8833df6ffe3de490c641db572fc68a65bf412f2/src/Type/StringType.php#L128-L130

the string array values are not accepting the int values.

but not sure how that is related to strict-rules..

herndlm: aha on the playground the treatPhpDocTypesAsCertain is also somehow involved apparently. I obviously did not invest carefully now but it's confusing :)

ondrejmirtes: My guess is one of the pollute options set to false, that would explain the playground, but it doesn't explain the rule test.

herndlm: you were right, setting `polluteScopeWithLoopInitialAssignments` to `true` makes the error go away in the test (it is `false` there). in general, by default it is `true` and strict-rules sets this to `false` in the playground which makes the error appear.

ondrejmirtes:
I changed it (it surprised me), so please rebase and update this branch https://github.com/phpstan/phpstan-src/commit/1b56b0c1e0fb7ad574d8d13e4ee63fb05716fa9c

herndlm:
done. looking good now 👍

ondrejmirtes:
Thank you.

PR #3874, opened 2025-03-11
Fix false positives on existing-offsets after assign

staabm:
closes https://github.com/phpstan/phpstan/issues/12406
closes https://github.com/phpstan/phpstan/issues/11679

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5480,8 +5480,7 @@ final class NodeScopeResolver
 				}
 
 				if ($originalVar->dim instanceof Variable || $originalVar->dim instanceof Node\Scalar) {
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Thank you!

ondrejmirtes:
I'm sorry I'm supplying you with so much work today 😊 This PR causes these false positives: https://phpstan.org/r/deb8f749-5ef2-41cd-ba7b-e3039a8f692c

staabm:
I had not yet a idea how to fix it.
adding some debug statements it seems the problem is, that the loop analysis already knows all offsets the array will have in the end and the first assignment already assigns the array with all offsets

https://phpstan.org/r/ce5a357e-f957-407a-b173-96ea59652144

I will be AFK until later today or tomorrow and cannot look further into it right now

PR #3876, opened 2025-03-11
Support narrowing a constant array to a list with count

herndlm:
Implementing what was found/suggested in https://github.com/phpstan/phpstan-src/pull/3709#issuecomment-2714931508 and improves very special cases with int ranges. 

PR title is maybe misleading. it's making keys non-optional if it can for list-like cases.

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1061,10 +1061,11 @@ final class TypeSpecifier
 			$isNormalCount = (new ConstantIntegerType(COUNT_NORMAL))->isSuperTypeOf($mode)->result->or($type->getIterableValueType()->isArray()->negate());
 		}
 
+		$isConstantArray = $type->isConstantArray();
 		$isList = $type->isList();
 		if (
 			!$isNormalCount->yes()
-			|| (!$type->isConstantArray()->yes() && !$isList->yes())
+			|| (!$isConstantArray->yes() && !$isList->yes())
 			|| $type->isIterableAtLeastOnce()->no() // array{} cannot be used for further narrowing
 		) {
 			return null;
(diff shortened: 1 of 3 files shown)


ondrejmirtes (changes requested):
I believe this has nothing to do with lists. If you have `array{a?: int, b?: string}` and check count >= 2, you can make them required too. Please add more tests.

If I have `array{a: int, b: string, c?: int}` and check `count > 2` or `count >= 3`, `c` should become required too.

staabm:
> If I have `array{a: int, b: string, c?: int}` and check `count > 2` or `count >= 3`, `c` should become required too.

this sounds like sealed vs. unsealed array shape business

ondrejmirtes:
Yes, in the future we will have to modify the condition to only apply to sealed shapes.

ondrejmirtes:
You're right, we shouldn't change this now before sealed shapes are implemented.

ondrejmirtes:
Thank you.

ondrejmirtes:
We have another regression: https://phpstan.org/r/c97add7b-a241-45c9-88f1-b8e7c76daa09

Please, Martin or Markus, look into it. Thank you!

herndlm:
weird, this stuff is supposed to deal only with lists or constant arrays, I'll try to take a look

herndlm:
yeah, looks not related to recent changes at least. if I did not mess up massively right now, then this is `*NEVER*` already with phpstan 2.0.0

I'll see if I can find out why though..

herndlm:
wait, everything is OK I think 😅 it is comparing a constant int with an array and that will always return false and make the assertion throw an exception. riiight? :) https://3v4l.org/k2LU5

I was already deep-debugging `TypeSpecifier` and my confusion grew by the second 😂

ondrejmirtes:
Oh now I'm confused 😂 I need to look into what was I thinking.

ondrejmirtes:
Alright, this is the right code sample: https://phpstan.org/r/032a2b0f-40bc-48e1-a5ee-a71859e78b04

This really changed before and after this PR. Sorry about the previous confusion.

PR #3877, opened 2025-03-12
ResultCache: allow customization of params not invalidating cache

janedbal:
- Currently, I'm improving our dead-code-detector with some [optional debug output](https://github.com/shipmonk-rnd/dead-code-detector/pull/165/files). 
- This is configured via parameter that is wired to a Rule hooked to `CollectedDataNode`. 
- As this runs after parallel analysis is done, changing such parameter never affects result cache, but it gets invalidated anyway. 
- This PR solves this issue by providing configuration option where even phpstan extensions can mark their parameters as not affecting result cache.

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -189,6 +189,19 @@ parameters:
 			- '1.1.1.2'
 		tmpDir: %sysGetTempDir%/phpstan-fixer
 	__validate: true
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Thank you!

janedbal:
Thank you!

PR #3878, opened 2025-03-13
Implement TypeSpecifierContext->getReturnType()

staabm:
for expression like

`if (doFoo() > 0)` 

we want to pass the context specific return-type information into type-specifying extensions so they can decide not just on true/truethy/false/falsy but also on "remaining return type values" (e.g. preg_match returns `0|1|false`)

this should help to reduce the number of hardcoded hacks within the TypeSpecifier und move over some cases into extensions (which also allows 3rd party extensions todo similar things, as they can't hard-code hack the TypeSpecifier)

ondrej quoted:

> There should be a new method on the context, can be getReturnType
>
> Let’s say we have doFoo(): int
> if (doFoo() > 0), in the then branch, this should return int<1, max>, in the else branch it should return int<min, 0>

### DIFF
--- a/src/Analyser/TypeSpecifierContext.php
+++ b/src/Analyser/TypeSpecifierContext.php
@@ -32,7 +32,7 @@ final class TypeSpecifierContext
 	private static function create(?int $value, ?Type $returnType = null): self
 	{
 		if ($returnType !== null) {
-			// return type bound context is unique for each context and therefore not cachable
+			// return type bound context is unique for each context and therefore not cacheable
 			return new self($value, $returnType);
 		}
 
--- a/src/Analyser/TypeSpecifierContext.php
+++ b/src/Analyser/TypeSpecifierContext.php
@@ -3,7 +3,6 @@
 namespace PHPStan\Analyser;
 
 use PHPStan\ShouldNotHappenException;
-use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 
 /**
@@ -23,19 +22,27 @@ final class TypeSpecifierContext
 	/** @var self[] */
 	private static array $registry;
 
-	private function __construct(private ?int $value)
+	private function __construct(
+		private ?int $value,
+		private ?Type $returnType,
+	)
 	{
 	}
 
-	private static function create(?int $value): self
+	private static function create(?int $value, ?Type $returnType = null): self
 	{
-		self::$registry[$value] ??= new self($value);
+		if ($returnType !== null) {
+			// return type bound context is unique for each context and therefore not cachable
+			return new self($value, $returnType);
+		}
+
+		self::$registry[$value] ??= new self($value, null);
 		return self::$registry[$value];
 	}
 
-	public static function createTrue(): self
+	public static function createTrue(?Type $returnType = null): self
 	{
-		return self::create(self::CONTEXT_TRUE);
+		return self::create(self::CONTEXT_TRUE, $returnType);
 	}
 
 	public static function createTruthy(): self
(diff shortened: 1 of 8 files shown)


tests/PHPStan/Analyser/data/type-specifier-context-return-type.php
```diff
@@ -0,0 +1,19 @@
...
+namespace TypeSpecifierContextReturnTypeTest;
+
+use function PHPStan\Testing\assertType;
+
+class ContextReturnType {
+	public function returnsInt(int $specifiedContextReturnType): int {}
+
+	public function doFoo(int $i) {
+		assertType('int', $i);
+		if ($this->returnsInt($i) > 0) {
```
staabm, line 12: this does not work yet, as we don't call into MethodTypeSpecifyingExtension for arbitrary conditions.. this is only called for a hardcoded list of methods/functions, see https://github.com/phpstan/phpstan-src/blob/7dd7b1426dd24da14eb788b4a476aad588e86dfb/src/Analyser/TypeSpecifier.php#L238-L331

herndlm: just out of curiosity - what's the context of this? why does the "context" get type info?

staabm: for expression like

`if (doFoo() > 0)` 

we want to pass the context specific return-type information into type-specifying extensions so they can decide not just on true/truethy/false/falsy but also on "remaining return type values" (e.g. preg_match returns `0|1|false`)

this should help to reduce the number of hardcoded hacks within the TypeSpecifier und move over some cases into extensions (which also allows 3rd party extensions todo similar things, as they can't hard-code hack the TypeSpecifier)

ondrej quoted:

> There should be a new method on the context, can be getReturnType
>
> Let’s say we have doFoo(): int
> if (doFoo() > 0), in the then branch, this should return int<1, max>, in the else branch it should return int<min, 0>

src/Analyser/TypeSpecifierContext.php
```diff
@@ -21,34 +22,42 @@ final class TypeSpecifierContext
...
+
+		self::$registry[$value] ??= new self($value, null);
 		return self::$registry[$value];
 	}
 
-	public static function createTrue(): self
+	public static function createTrue(?Type $returnType = null): self
 	{
-		return self::create(self::CONTEXT_TRUE);
+		return self::create(self::CONTEXT_TRUE, $returnType);
```
staabm, line 47: todo: bit-field cleanup after the main-point of this PR works

> About true/truthy - let’s get rid of the bit-masking, I think they will be implemented as getReturnType()->isTrue()->yes()
> The other one as getReturnType()->toBoolean()->isTrue()->yes()

src/Analyser/TypeSpecifierContext.php
```diff
@@ -61,7 +72,14 @@ public function negate(): self
 		if ($this->value === null) {
 			throw new ShouldNotHappenException();
 		}
-		return self::create(~$this->value & self::CONTEXT_BITMASK);
+
+		$negatedReturnType = null;
+		if ($this->returnType !== null) {
+			$baseType = $this->returnType->generalize(GeneralizePrecision::lessSpecific());
+			$negatedReturnType = TypeCombinator::remove($baseType, $this->returnType);
```
staabm, line 79: This won't work for unions of int. We will need 2 two return types in the context and toggle between them on negate()

staabm:
had a different try with https://github.com/phpstan/phpstan-src/pull/3881

PR #3879, opened 2025-03-13
Collected data: reduce memory consumption & result cache size

janedbal:
Previous state:
- for **each emit** inside Collector, **filepath and collector class is carried along the raw data**
  - this adds significant amount of bytes transferred and stored in result cache

New state:
- collected data are groupped by file & collector class
  - this causes significant improvement for projects like [Dead Code Detector](https://github.com/shipmonk-rnd/dead-code-detector) where we emit very often

Impact:
- my [old analysis](https://github.com/shipmonk-rnd/dead-code-detector/pull/30) where I made similar optimization in userland was:
  - filesize of `resultCache.php` dropped from `384M` to `213M` (**44%** down)
  - RAM usage dropped from `20.14 GB` to `19.52 GB`

### DIFF
--- a/src/Collectors/CollectedData.php
+++ b/src/Collectors/CollectedData.php
@@ -8,6 +8,8 @@ use ReturnTypeWillChange;
 
 /**
  * @api
+ *
+ * @phpstan-type CollectorData = array<string, array<class-string<Collector<Node, mixed>>, list<mixed>>>
  */
 final class CollectedData implements JsonSerializable
 {
--- a/src/Analyser/Analyser.php
+++ b/src/Analyser/Analyser.php
@@ -3,8 +3,7 @@
 namespace PHPStan\Analyser;
 
 use Closure;
-use PhpParser\Node;
-use PHPStan\Collectors\Collector;
+use PHPStan\Collectors\CollectedData;
 use PHPStan\Collectors\Registry as CollectorRegistry;
 use PHPStan\Rules\Registry as RuleRegistry;
 use Throwable;
@@ -13,6 +12,9 @@ use function array_merge;
 use function count;
 use function memory_get_peak_usage;
 
+/**
+ * @phpstan-import-type CollectorData from CollectedData
+ */
 final class Analyser
 {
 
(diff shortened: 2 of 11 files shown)


janedbal:
I made some new measurements:

| Analysed **17k files**  | Before  | After   |
|----------------------|--------|--------|
| Result cache size   | 813 MB  | 505 MB  |
| Used memory        | 12.76 GB   | 12.55 GB |

ondrejmirtes (changes requested):
Please introduce a type alias for `array<class-string<Collector<Node, mixed>>, list<mixed>>>`. You can put it above CollectedData class for example.

src/Collectors/CollectedData.php
```diff
@@ -8,6 +8,8 @@
 
 /**
  * @api
+ *
+ * @phpstan-type CollectorData = array<string, array<class-string<Collector<Node, mixed>>, list<mixed>>>
```
staabm, line 12: IIRC this `=` is too much

```suggestion
 * @phpstan-type CollectorData array<string, array<class-string<Collector<Node, mixed>>, list<mixed>>>
```

ondrejmirtes: = works and importantly, it's the only syntax recognised by PhpStorm.

janedbal: I had that at first, but then aligned it with existing `@phpstan-type LinesToIgnore = array<string, array<int, non-empty-list<string>|null>>`

staabm: @ondrejmirtes maybe we should use it in the docs then?

https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases

ondrejmirtes: I'd rather be if PhpStorm fixed that on their side.

ondrejmirtes:
Thank you!

src/Collectors/CollectedData.php
janedbal: At least something work in PHPStorm without =:

image

PR #118, opened 2020-01-29
JSON encode/decode should invalidate JSON error state

cs278:
Fixes https://github.com/phpstan/phpstan/issues/2863

PR #3880, opened 2025-03-13
Treat #[Pure(true)] in PhpStorm stubs as hasSideEffects => true

zonuexe:
PhpStorm marks functions that have side effects but whose return value is important as `#[Pure(true)]`, which in current PHPStan functionality means `[hasSideEffects => true]`.

refs https://github.com/phpstan/phpstan-src/pull/3867#issuecomment-2721999811
refs https://github.com/phpstan/phpstan/issues/12738


staabm (changes requested):
before this PR phpstan errored
https://phpstan.org/r/57c356b1-b4f6-4a36-94f8-01a21b8996d4

> Call to function checkdate() on a separate line has no effect.

after this change it no longer does

```
➜  phpstan-src git:(pr/3880) ✗ cat test.php

class HelloWorld
{
        public function sayHello($m): void
        {
                checkdate($m, $m, $m);
        }
}                                                                           
                                                                                                                        

➜  phpstan-src git:(pr/3880) ✗ php bin/phpstan analyze test.php --debug
/Users/staabm/workspace/phpstan-src/test.php
 ------ --------------------------------------------- 
  Line   test.php                                     
 ------ --------------------------------------------- 
  3      Class HelloWorld must be abstract or final.  
         🪪  phpstan.finalClass                       
 ------ --------------------------------------------- 


                                                                                                                        
 [ERROR] Found 1 error                                                                                                  
                                                                                                                        

➜  phpstan-src git:(2.1.x) ✗ php bin/phpstan analyze test.php --debug
/Users/staabm/workspace/phpstan-src/test.php
 ------ ---------------------------------------------------------------- 
  Line   test.php                                                        
 ------ ---------------------------------------------------------------- 
  3      Class HelloWorld must be abstract or final.                     
         🪪  phpstan.finalClass                                          
  7      Call to function checkdate() on a separate line has no effect.  
         🪪  function.resultUnused                                       
 ------ ---------------------------------------------------------------- 


                                                                                                                        
 [ERROR] Found 2 errors                                                                                                 
                                                                                                                        

➜  phpstan-src git:(2.1.x) ✗ 

```

zonuexe:
@staabm Thanks, I've sent a PR upstream since the [`checkdate()`](https://www.php.net/checkdate) function is clearly a pure function.
https://github.com/JetBrains/phpstorm-stubs/pull/1726/files#r1997508160

resources/functionMetadata.php
```diff
@@ -811,47 +811,47 @@
 	'ctype_upper' => ['hasSideEffects' => false],
 	'ctype_xdigit' => ['hasSideEffects' => false],
 	'curl_copy_handle' => ['hasSideEffects' => false],
-	'curl_errno' => ['hasSideEffects' => false],
-	'curl_error' => ['hasSideEffects' => false],
+	'curl_errno' => ['hasSideEffects' => true],
+	'curl_error' => ['hasSideEffects' => true],
```
staabm, line 815: This also looks wrong. I guess you should review all lines which changed from false to true

zonuexe: Although they appear to be pure, they actually rely on some external state associated with the resource, so it makes sense that PhpStorm marks them as `#[Pure(true)]`.

While it may be convenient to make `curl_errno()` pure for convenience in common use cases, it is inconsistent with impure functions like `ob_get_level()`(#12577).

To solve this problem, [`must_use`/`#[NoDiscard]`](https://github.com/phpstan/phpstan/issues/12738) needs to be supported separately from `@pure`/`@impure`.

claudepache:
As I understand, both `#[Pure(false)]` and `#[Pure(true)]` mean that the function has no side-effect, the difference between the two is that a function marked with `#[Pure(true)]` may depend on global state (and therefore is not pure).

In case `"hasSideEffects" => (bool)` doesn’t mean more than what it says (which I have not checked), those functions should be marked with `"hasSideEffects" => false`.

ondrejmirtes:
Thank you!

staabm:
nice, this fixed a few more bugs, see https://github.com/phpstan/phpstan-src/actions/runs/13851103733

regression tests would be awesome

ondrejmirtes:
I don't think we need them, they'd just verify "yeah, this function is in this array", not any real logic 😊

ondrejmirtes:
But yeah, three issues can be closed thanks to this :)

resources/functionMetadata.php
```diff
@@ -1444,8 +1455,8 @@
 	'posix_uname' => ['hasSideEffects' => false],
 	'pow' => ['hasSideEffects' => false],
 	'preg_grep' => ['hasSideEffects' => false],
-	'preg_last_error' => ['hasSideEffects' => false],
-	'preg_last_error_msg' => ['hasSideEffects' => false],
+	'preg_last_error' => ['hasSideEffects' => true],
```
mvorisek, line 1458: https://3v4l.org/O43Rb

what side effects does it have?

ondrejmirtes: Thanks to this line we solved this issue https://github.com/phpstan/phpstan/issues/10342

mvorisek: Sure, I mean we should differentiate between pure and side effects. In my understanding "pure" is about return value, "side effects" are about effects on other functions. So I would expect this function to have no side effects, but be pure instead.

related with #118

PR #3881, opened 2025-03-14
Implement TypeSpecifierContext->getReturnType()

staabm:
trying a minimal approach to see whether the idea works as a whole

---

for expression like

`if (doFoo() > 0)` 

we want to pass the context specific return-type information into type-specifying extensions so they can decide not just on true/truethy/false/falsy but also on "remaining return type values" (e.g. preg_match returns `0|1|false`)

this should help to reduce the number of hardcoded hacks within the TypeSpecifier und move over some cases into extensions (which also allows 3rd party extensions todo similar things, as they can't hard-code hack the TypeSpecifier)

ondrej quoted:

> There should be a new method on the context, can be getReturnType
>
> Let’s say we have doFoo(): int
> if (doFoo() > 0), in the then branch, this should return int<1, max>, in the else branch it should return int<min, 0>


src/Analyser/TypeSpecifierContext.php
```diff
@@ -18,17 +19,15 @@ final class TypeSpecifierContext
 	public const CONTEXT_FALSEY = self::CONTEXT_FALSE | self::CONTEXT_FALSEY_BUT_NOT_FALSE;
 	public const CONTEXT_BITMASK = 0b1111;
 
-	/** @var self[] */
-	private static array $registry;
```
staabm: drop the caching, as context now contains an additional Type (state)

on my machine this did not make a difference in performance

src/Analyser/TypeSpecifier.php
```diff
@@ -466,6 +426,26 @@ public function specifyTypesInCondition(
 				}
 			}
 
+			if (
+				!$context->null()
+				&& $expr->right instanceof Expr\FuncCall
+				&& $expr->right->name instanceof Name
+				&& in_array(strtolower((string) $expr->right->name), ['preg_match', 'strlen', 'mb_strlen'], true)
+			) {
```
staabm, line 436: atm I whitelisted the cases which I dropped above to keep my sanity.
I feel we can drop it after we removed the hard-coded case list and moved the logic into extensions

staabm:
the [case failling in webmozart-assert](https://github.com/phpstan/phpstan-webmozart-assert/blob/f2687ce7638e22ab8c674dff3387ebcdb4b9469e/tests/Type/WebMozartAssert/data/string.php#L142-L143) is effectively

```
<?php

declare(strict_types=1);

class HelloWorld
{
	public function sayHello($m): void
	{
		\PHPStan\dumpType(
			is_string($m)
			&& strlen($m) >= 1
			&& strlen($m) <= 0
		);
	}
}
```

not sure what todo with it atm.

__edit: fixed with [19749e8](https://github.com/phpstan/phpstan-src/pull/3881/commits/19749e8ae772bc052bd30938e8cedd3b7471ecb5)__

staabm:
debugging some more, I think more precisely the problem is here:

```
<?php

function sayHello(string $m): void
{
	if (strlen($m) >= 1) {
		if (strlen($m) <= 0) {
			\PHPStan\dumpType(
				$m
			);
		}
	}
}
```

it dumps `non-empty-string` instead of `*NEVER*`

__edit: fixed with [19749e8](https://github.com/phpstan/phpstan-src/pull/3881/commits/19749e8ae772bc052bd30938e8cedd3b7471ecb5)__

src/Type/Php/StrlenTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,72 @@
...
+
+		$argType = $scope->getType($args[0]->value);
+		if (!$argType->isString()->yes()) {
+			return new SpecifiedTypes();
+		}
+
+		$returnType = $context->getReturnType();
+		if ($returnType instanceof NeverType) {
+			return $this->typeSpecifier->create($args[0]->value, $returnType, $context->negate(), $scope);
+		}
```
staabm, line 57: we might move this into TypeSpecifier instead.

when `$callType instanceof NeverType` turn all non-by-ref-args into `*NEVER*`.
not sure its required though.

staabm:
//cc @herndlm

src/Analyser/TypeSpecifier.php
```diff
@@ -466,6 +426,26 @@ public function specifyTypesInCondition(
...
+				if (!$scope instanceof MutatingScope) {
+					throw new ShouldNotHappenException();
+				}
+				$newScope = $scope->filterBySpecifiedTypes($result);
+				$callType = $newScope->getType($expr->right);
+				$newContext = $context->newWithReturnType($callType);
```
staabm, line 442: In the future we could use this new api also for `==` and `===` not just `<` I think.

ondrejmirtes:
Yeah, this works, but I'd like to "burn more bridges" in the first PR about this change. I don't mean "breaking backward compatibility", but what I mean is to use the new code path as much as possible.

1) TypeSpecifierContext should be created with the return type. So the constructor should accept `?Type`, not `?int $value`.
2) We probably don't need the constants with bitmasks at all. The createTrue etc. methods should simply create the object with a type that makes sense and continues to be correct.
3) The negate method is tricky. Maybe we need to create the object with both the "if" path type and the "else" path type.

ondrejmirtes:
I use this approach often. It forces the code to be correct and generally uncovers any flaws about the changes 😊

PR #3882, opened 2025-03-14
Call all involved class on TypeSpecifierAwareExtension

brambaud:
See [ #12741](https://github.com/phpstan/phpstan/issues/12741)

For the moment, we only want to try a modification to see if existing tests passed. The next steps is to implement a test to reproduce the error then we'll re-implement the modification for real.

cc @ChinaskiJr

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -519,18 +519,27 @@ final class TypeSpecifier
 			$methodReflection = $scope->getMethodReflection($methodCalledOnType, $expr->name->name);
 			if ($methodReflection !== null) {
 				$referencedClasses = $methodCalledOnType->getObjectClassNames();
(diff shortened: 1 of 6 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -519,19 +519,28 @@ public function specifyTypesInCondition(
 			$methodReflection = $scope->getMethodReflection($methodCalledOnType, $expr->name->name);
 			if ($methodReflection !== null) {
 				$referencedClasses = $methodCalledOnType->getObjectClassNames();
```
staabm, line 521: I think this can be simplified using `getObjectClassReflections` instead of `getObjectClassNames`.
that way you no longer need the `reflectionProvider` as you already get the `ClassReflection`s returned.

(same for static call)

PR #3883, opened 2025-03-18
Excluded path: include example for optional path

thePanz:
Make more clear how to mark an excluded path optional, with the use of `(?)` suffix

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -383,11 +383,13 @@ final class CommandHelper
 			if (count($suggestOptional) > 0) {
 				$errorOutput->writeLineFormatted('If the excluded path can sometimes exist, append <fg=cyan>(?)</>');
 				$errorOutput->writeLineFormatted('to its config entry to mark it as optional. Example:');
-				$errorOutput->writeLineFormatted('<fg=cyan>excludePaths:</>');
+				$errorOutput->writeLineFormatted('');
+				$errorOutput->writeLineFormatted('<fg=cyan>parameters:</>');
+				$errorOutput->writeLineFormatted("\t<fg=cyan>excludePaths:</>");
 				foreach ($suggestOptional as $key => $suggestOptionalPaths) {
-					$errorOutput->writeLineFormatted(sprintf('  <fg=cyan>%s:</>', $key));
+					$errorOutput->writeLineFormatted(sprintf("\t\t<fg=cyan>%s:</>", $key));
 					foreach ($suggestOptionalPaths as $suggestOptionalPath) {
-						$errorOutput->writeLineFormatted(sprintf('    - <fg=cyan>%s (?)</>', $suggestOptionalPath));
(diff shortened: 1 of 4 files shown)


src/Command/CommandHelper.php
```diff
@@ -380,7 +380,8 @@ public static function begin(
 			}
 
 			$errorOutput->writeLineFormatted('If the excluded path can sometimes exist, append <fg=cyan>(?)</>');
-			$errorOutput->writeLineFormatted('to its config entry to mark it as optional.');
+			$errorOutput->writeLineFormatted('to its config entry to mark it as optional. Example:');
+			$errorOutput->writeLineFormatted(' <fg=cyan>- path/something (?)</>');
```
ondrejmirtes: Would be great to show the full example including `excludePaths:` and to use the actual path in question, not just `path/something`. Thanks.

ondrejmirtes (changes requested):

src/Command/CommandHelper.php
thePanz: I tried to achieve that as my first contribution to PHPStan :)
I proposed a first change to be able to track which paths should be suggested.

from my understanding: paths adhering to `FileExcluder::isFnmatchPattern()` are not evaluated, thus I opted to *not* suggest those as Optional. Let me know if those should be added too as suggestions.

ondrejmirtes (changes requested):
I've just added a single entry that does not exist like this:

```neon
parameters:
	excludePaths:
		- foo

```

But your code results in the following output:


Screenshot 2025-03-19 at 9 02 25

Please fix that.

ondrejmirtes:
Also I just noticed there's a typo: `analyseAndScan` is missing a colon.

gharlan:
The example should use relative paths.

ondrejmirtes:
It should but right now it's impossible to get the correct relative path.

Because relative path depends on the location of the config file, but at this point we have the whole parameters array decoded from all the config files at once.

I don't need OP to tackle this. I can look at it myself once the PR works.

thePanz:
Some brain dump while trying to implement the feature:
1. in `ValidateExcludePathsExtension::loadConfiguration()` we get the paths resolved and absolute, we would need a way to figure out the original value passed in the configuration
2. the `$this->getContainerBuilder()` returns what I call "fully qualified configuration" where the shortcut mentioned here https://phpstan.org/user-guide/ignoring-errors#excluding-whole-file is already resolved to the full configuration

thePanz:
Got this result: 
image

By using the following configuration:
```neon
includes: [ build/phpstan.neon ]
parameters:
	excludePaths:
		- foo

```

ondrejmirtes:
Please mark it as ready for review if it's done from your side.

ondrejmirtes:
I've pushed two commits, feel free to inspect them 😊

ondrejmirtes:
Thank you!

PR #3884, opened 2025-03-18
Fix `SessionHandlerInterface::read` return type

sayuprc:
Fixed `SessionHandlerInterface::read` could return false

resolve phpstan/phpstan#12748


ondrejmirtes:
Thank you.

PR #3885, opened 2025-03-18
Fix for handling non-stringable types in dynamic access

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/9475

### DIFF
--- a/src/Rules/Classes/ClassConstantRule.php
+++ b/src/Rules/Classes/ClassConstantRule.php
@@ -11,6 +11,7 @@ use PHPStan\Php\PhpVersion;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\ClassNameCheck;
 use PHPStan\Rules\ClassNameNodePair;
+use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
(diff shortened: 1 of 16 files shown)


ondrejmirtes (changes requested):
This is very very valuable, but it needs two things:

1) The new errors can be reported only with bleeding edge.
2) The new code has to employ RuleLevelHelper. The idea is that "always wrong" like `int` is always reported, "sometimes wrong" like `int|string` is reported on level 7+, nullable like `string|null` is reported on level 8+ and mixeds are reported on level 9+. Just use RuleLevelHelper like it's used in all the rules and it will be taken care of automatically.

ondrejmirtes (changes requested):
Also please do this in two separate PRs:

1) Let the rules handle Expr as `$node->name` to continue the error-reporting. So that:

```php
$a = 'Foo';
echo Foo::{$a}
```

Is checked against a non-existent constant. This doesn't need bleeding edge.

2) Introduce new reported errors that check `$a` in `Foo::{$a}` to be a string. This needs bleeding edge and RuleLevelHelper.

Thank you.

ondrejmirtes:
Also within those PRs add relevant regression tests for these issues: https://github.com/phpstan/phpstan-src/actions/runs/13923245791 (some of them might disappear once we do it for bleeding edge so it's nice to have this list now)

zonuexe:
Ok, I'll freeze this branch and submit some PRs.

PR #3886, opened 2025-03-18
Support dynamic name expressions in rules

zonuexe:
resolve https://github.com/phpstan/phpstan-src/pull/3885#pullrequestreview-2694393342

### DIFF
--- a/src/Rules/Methods/CallMethodsRule.php
+++ b/src/Rules/Methods/CallMethodsRule.php
@@ -3,7 +3,9 @@
 namespace PHPStan\Rules\Methods;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr\BinaryOp\Identical;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Scalar\String_;
 use PHPStan\Analyser\Scope;
 use PHPStan\Internal\SprintfHelper;
 use PHPStan\Reflection\ParametersAcceptorSelector;
@@ -11,6 +13,7 @@ use PHPStan\Rules\FunctionCallParametersCheck;
 use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\Rule;
 use PHPStan\Type\Constant\ConstantStringType;
+use function array_column;
 use function array_map;
 use function array_merge;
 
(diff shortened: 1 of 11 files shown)


src/Rules/Methods/CallMethodsRule.php
```diff
@@ -31,12 +33,26 @@ public function getNodeType(): string
...
 	public function processNode(Node $node, Scope $scope): array
 	{
-		if (!$node->name instanceof Node\Identifier) {
-			return [];
+		$errors = [];
+		if ($node->name instanceof Node\Identifier) {
+			$methodNames = [$node->name->name];
+		} else {
+			$callType = $scope->getType($node->name);
+			$methodNames = array_map(static fn ($type): string => $type->getValue(), $callType->getConstantStrings());
```
ondrejmirtes: This applies to all the rules. We need to pair each name with its own Scope.

Let's say we have this code:

```php
if (rand(0, 1)) {
    $foo = 'doFoo';
    $bar = 1;
} else {
    $foo = 'doBar';
    $bar = 'a';
}

$this->$foo($bar);
```

If the first parameter of doFoo accepts int, and the first parameter of doBar accepts string, this code should not report any errors.

You can see the way it should be done here: https://github.com/phpstan/phpstan-src/blob/ef6cc0a1d6e2b60c0328cb6a47db55b4c615056b/src/Analyser/MutatingScope.php#L2149

Also please feel free to do the same thing here: https://github.com/phpstan/phpstan-src/blob/ef6cc0a1d6e2b60c0328cb6a47db55b4c615056b/src/Rules/Properties/AccessPropertiesCheck.php#L47 (I thought it already works but apparently not).

ondrejmirtes (changes requested):

src/Rules/Methods/CallMethodsRule.php
staabm: ohh wow. thats clever.

src/Rules/Methods/CallMethodsRule.php
```diff
@@ -31,12 +37,34 @@ public function getNodeType(): string
...
 	{
-		if (!$node->name instanceof Node\Identifier) {
-			return [];
+		$errors = [];
+		if ($node->name instanceof Node\Identifier) {
+			$methodNameScopes = [$node->name->name => $scope];
+		} else {
+			$nameType = $scope->getType($node->name);
+			$methodNames = array_map(static fn (ConstantStringType $type): string => $type->getValue(), $nameType->getConstantStrings());
+			$methodNameScopes = array_column(array_map(
```
ondrejmirtes: Please write less clever code without multiple array_map and array_column calls. A simple foreach would suffice!

ondrejmirtes (changes requested):

src/Rules/Methods/CallMethodsRule.php
```diff
@@ -31,12 +34,30 @@ public function getNodeType(): string
...
 	{
-		if (!$node->name instanceof Node\Identifier) {
-			return [];
+		$errors = [];
+		if ($node->name instanceof Node\Identifier) {
+			$methodNameScopes = [$node->name->name => $scope];
+		} else {
+			$nameType = $scope->getType($node->name);
+			$methodNameScopes = [];
+			foreach ($nameType->getConstantStrings() as $constantString) {
```
ondrejmirtes, line 43: Nice! Now please do it consistently for all the touched rules. Thanks.

ondrejmirtes (changes requested):

ondrejmirtes:
Perfect, thank you! Please add the new error in a single rule and once we agree how it should work, you can copy the approach to other rules.

VincentLanglet:
I think it solved https://github.com/phpstan/phpstan/issues/2920

Should I write a non regression test or someone already on it ?

And seems like it also solved https://github.com/phpstan/phpstan/issues/7990

For the issue https://github.com/phpstan/phpstan/issues/4608, I'm not sure... Seems like there is a regression since
the last two lines of
```
$c = new class {
	public function abc(): void {}
};

$s = rand(0, 1) ? 'abc' : 'not_abc';

$c->{$s}();
call_user_func([$c, $s]);
[$c, $s]();
```
are not reported anymore. @ondrejmirtes @zonuexe

ondrejmirtes:
https://github.com/phpstan/phpstan/issues/2920 does not need a regression test, there are many examples of tests in this PR with the same value here already. Same for https://github.com/phpstan/phpstan/issues/7990.

In https://github.com/phpstan/phpstan/issues/4608 the lines 10 and 11 stopped being reported in september 2022, unrelated to this PR.

PR #3887, opened 2025-03-18
ResultCacheManager: fix meta key difference for projectConfig

janedbal:
Previously, `projectConfig` was **always reported as different** because if was comparing neon string with array:

image

### DIFF
--- a/src/Analyser/ResultCache/ResultCacheManager.php
+++ b/src/Analyser/ResultCache/ResultCacheManager.php
@@ -312,17 +312,26 @@ final class ResultCacheManager
 	}
 
 	/**
+	 * @param mixed[]|null $projectConfig
+	 */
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

ondrejmirtes:
I'm sorry, I had to revert this. This causes a crash when there's a pre-existing result cache (i.e. all projects).

```
PHP Fatal error:  Uncaught TypeError: PHPStan\Analyser\ResultCache\ResultCacheManager::normalizeMetaProjectConfig(): Argument #1 ($projectConfig) must be of type ?array, string given, called in phar:///home/runner/work/ipsum/ipsum/api/vendor/phpstan/phpstan/phpstan.phar/src/Analyser/ResultCache/ResultCacheManager.php on line 337
```

PR #3888, opened 2025-03-19
Fix always-true detection in in_array with union type haystack

sayuprc:
resolve phpstan/phpstan#12755

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -153,6 +153,10 @@ final class ImpossibleCheckTypeHelper
 							foreach ($haystackArrayTypes as $haystackArrayType) {
 								if ($haystackArrayType instanceof ConstantArrayType) {
 									foreach ($haystackArrayType->getValueTypes() as $i => $haystackArrayValueType) {
+										if ($haystackArrayValueType instanceof UnionType) {
+											continue;
+										}
+
 										if ($haystackArrayType->isOptionalKey($i)) {
 											continue;
 										}
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -153,11 +153,7 @@ final class ImpossibleCheckTypeHelper
 							foreach ($haystackArrayTypes as $haystackArrayType) {
 								if ($haystackArrayType instanceof ConstantArrayType) {
 									foreach ($haystackArrayType->getValueTypes() as $i => $haystackArrayValueType) {
-										if ($haystackArrayValueType instanceof UnionType) {
-											continue;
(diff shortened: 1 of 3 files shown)


src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -147,6 +147,10 @@ public function findSpecifiedType(
 							foreach ($haystackArrayTypes as $haystackArrayType) {
 								if ($haystackArrayType instanceof ConstantArrayType) {
 									foreach ($haystackArrayType->getValueTypes() as $i => $haystackArrayValueType) {
+										if ($haystackArrayValueType instanceof UnionType) {
```
VincentLanglet, line 150: This seems weird.

Does it means that 
```
/**
	 * @param array{
	 *     key1: bool|int,
	 * } $myArray
	 */
	public function testBool(array $myArray): ?\stdClass
	{
	    if (\in_array(null, $myArray, true)) {
	        return null;
	    }

		return (object) $myArray;
	}
```
won't be reported too because `bool|int` is an union.

staabm:
your PR has impact on more open bugs.

please look at the changes and decide whether these are correct/intended
https://github.com/phpstan/phpstan-src/actions/runs/13947226248?pr=3888

please add additional tests to cover this bugs, in case they are fixed

tests/PHPStan/Rules/Comparison/data/bug-12755.php
```diff
@@ -0,0 +1,35 @@
...
+	 * } $myArray
+	 */
+	public function testBool(array $myArray): ?\stdClass
+	{
+	    if (\in_array(null, $myArray, true)) {
+	        return null;
+	    }
+
+		return (object) $myArray;
+	}
```
staabm: you should test a case which is expected to still emit an error after the change (a array which always contains null)

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -147,7 +147,7 @@ public function findSpecifiedType(
 							foreach ($haystackArrayTypes as $haystackArrayType) {
 								if ($haystackArrayType instanceof ConstantArrayType) {
 									foreach ($haystackArrayType->getValueTypes() as $i => $haystackArrayValueType) {
-										if ($haystackArrayType->isOptionalKey($i)) {
+										if ($haystackArrayValueType instanceof UnionType || $haystackArrayType->isOptionalKey($i)) {
```
ondrejmirtes: I believe there's a better way to ask this. Something like `count($haystackArrayValueType->getFiniteTypes()) === 1`.

ondrejmirtes (changes requested):

ondrejmirtes:
Please fix the conflicts.

ondrejmirtes:
Thank you!

PR #3889, opened 2025-03-20
Fix count() regression

staabm:
refs https://github.com/phpstan/phpstan-src/pull/3709#issuecomment-2739840190

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1063,10 +1063,12 @@ final class TypeSpecifier
 
 		$isConstantArray = $type->isConstantArray();
 		$isList = $type->isList();
+		$positiveInt = IntegerRangeType::fromInterval(1, null);
 		if (
 			!$isNormalCount->yes()
 			|| (!$isConstantArray->yes() && !$isList->yes())
 			|| $type->isIterableAtLeastOnce()->no() // array{} cannot be used for further narrowing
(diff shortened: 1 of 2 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -1063,10 +1063,12 @@ private function specifyTypesForCountFuncCall(
 
 		$isConstantArray = $type->isConstantArray();
 		$isList = $type->isList();
+		$positiveInt = IntegerRangeType::fromInterval(1, null);
 		if (
 			!$isNormalCount->yes()
 			|| (!$isConstantArray->yes() && !$isList->yes())
 			|| $type->isIterableAtLeastOnce()->no() // array{} cannot be used for further narrowing
+			|| !$positiveInt->isSuperTypeOf($sizeType)->yes()
```
ondrejmirtes: What would change if we also included zero here?

ondrejmirtes (changes requested):
Awesome, thanks!

src/Analyser/TypeSpecifier.php
staabm: seems to also work

herndlm:
Nice, thx!

ondrejmirtes:
Thank you.

ondrejmirtes:
We have another one 🙁 https://phpstan.org/r/c97add7b-a241-45c9-88f1-b8e7c76daa09

PR #3890, opened 2025-03-20
Fix elapsed time format for result cache restore time

janedbal:
Before: `Result cache restored in 6.400000 seconds.`
Now: `Result cache restored in 6.4 seconds.`


ondrejmirtes:
Thank you!

PR #3891, opened 2025-03-21
Remove MongoDB extension from function map

alcaeus:
With a 2.0 release imminent, having a single set of signatures is set to cause issues for users, regardless of which version of the stubs we use. Since there are multiple sources for function signatures we remove the function map entries, instead relying on either PhpStorm stubs (which have a similar issue as function map entries but allow for more flexibility) or using the reflection checks as fallback.


ondrejmirtes:
The impact of these changes can be seen here https://github.com/phpstan/phpstan-src/actions/runs/13988512408/job/39167077636

TBH it's pretty minimal and the diff could be used to improve PhpStorm stubs.

alcaeus:
Yep, I'll take a look at that. Looking at the first failure, I think that the PhpStorm stubs are out of date, which I'll double-check in the lead-up to our 2.0 release.

Which version would such a change land in? Is it possible to include this in 1.x or would it always land in a 2.x release only?

ondrejmirtes:
I no longer update phpstorm-stubs in 1.12.x.

Which is fine: both the functionMap will be cleared and phpstorm-stubs updated in 2.1.x.

alcaeus:
Sounds good - I was mainly asking to know which branch to target and to know whether we need to update to PHPStan 2.x in Doctrine ODM.

ondrejmirtes:
I'm fine with these changes so merging: https://github.com/phpstan/phpstan-src/actions/runs/15004909289/job/42161071236

PR #3892, opened 2025-03-21
RegexArrayShapeMatcher - turn more details immutable

staabm:
goal: more immutability so we can in the end re-use hoa/ast parse results instead of regex lexing/parsing the same regex over and over (which shows up in perf profiles).

this also paves the way for https://github.com/phpstan/phpstan/issues/12749 for which we need the group list to infer a more precise subject-type result

### DIFF
--- a/src/Type/Php/RegexArrayShapeMatcher.php
+++ b/src/Type/Php/RegexArrayShapeMatcher.php
@@ -123,18 +123,18 @@ final class RegexArrayShapeMatcher
 			$trailingOptionals++;
 		}
 
-		$onlyOptionalTopLevelGroupIndex = $this->getOnlyOptionalTopLevelGroupIndex($groupList);
+		$onlyOptionalTopLevelGroupId = $this->getOnlyOptionalTopLevelGroupId($groupList);
 		$onlyTopLevelAlternation = $this->getOnlyTopLevelAlternation($groupList);
 		$flags ??= 0;
 
 		if (
 			!$matchesAll
 			&& $wasMatched->yes()
-			&& $onlyOptionalTopLevelGroupIndex !== null
+			&& $onlyOptionalTopLevelGroupId !== null
 		) {
 			// if only one top level capturing optional group exists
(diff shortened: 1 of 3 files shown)


src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -115,30 +114,23 @@ private function matchRegex(string $regex, ?int $flags, TrinaryLogic $wasMatched
...
-		$trailingOptionals = 0;
-		foreach (array_reverse($groupList) as $captureGroup) {
-			if (!$captureGroup->isOptional()) {
-				break;
-			}
-			$trailingOptionals++;
-		}
-
-		$onlyOptionalTopLevelGroup = $this->getOnlyOptionalTopLevelGroup($groupList);
-		$onlyTopLevelAlternation = $this->getOnlyTopLevelAlternation($groupList);
```
staabm, line 127: move operations previously did on `array<int, RegexCapturingGroup> $captureGroups` into a dedicated class

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -154,12 +146,10 @@ private function matchRegex(string $regex, ?int $flags, TrinaryLogic $wasMatched
 				);
 			}
 
-			$onlyOptionalTopLevelGroup->clearOverrides();
```
staabm, line 157: no more state clearing/handling

src/Type/Php/RegexArrayShapeMatcher.php
```diff
@@ -199,11 +189,6 @@ private function matchRegex(string $regex, ?int $flags, TrinaryLogic $wasMatched
 				);
 
 				$combiTypes[] = $combiType;
-
-				foreach ($groupCombo as $groupId) {
-					$group = $comboList[$groupId];
-					$group->clearOverrides();
-				}
```
staabm, line 206: no more state clearing/handling

ondrejmirtes:
Awesome, thank you!"

PR #3893, opened 2025-03-22
Fix false positives on existing-offsets after assign

staabm:
fix regression mentioned in https://github.com/phpstan/phpstan-src/pull/3874#issuecomment-2743241539


ondrejmirtes:
Nice, makes sense. Thank you!

Maybe we should do equals() instead?

staabm:
> Maybe we should do equals() instead?

I tried a few cases, but I am not sure when this would make a difference

PR #3895, opened 2025-03-22
Fix `count()` regression

staabm:
fix regression mentioned in https://github.com/phpstan/phpstan-src/pull/3876#issuecomment-2744483174

followup to https://github.com/phpstan/phpstan-src/pull/3889

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1063,7 +1063,7 @@ final class TypeSpecifier
 
 		$isConstantArray = $type->isConstantArray();
 		$isList = $type->isList();
-		$zeroOrMore = IntegerRangeType::fromInterval(0, null);
+		$zeroOrMore = IntegerRangeType::fromInterval(1, null);
 		if (
 			!$isNormalCount->yes()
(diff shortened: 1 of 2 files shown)


staabm:
//cc @herndlm

herndlm:
Wow, you beat me to it it seems. I didn't even have time to get to a computer yet. Thx! But I was also not expecting for the fix to be this tbh 🤔

Maybe you can rename the variable too :)

staabm:
I guess we are narrowing a NonEmptyArray with a count(0) down the line. but I could not pinpoint where exactly.

ondrejmirtes:
Thank you! I'll report back if testing on real-world projects finds more problems. Otherwise I'll be able to release 2.1.9.

herndlm:
@ondrejmirtes just give me a minute, I think I can do a tiny cleanup of an edge case I was handling there which is now covered by @staabm's adaption

because the explanation with non-empty-array felt very very familiar :)

ondrejmirtes:
Yeah, sure :)

herndlm:
--> https://github.com/phpstan/phpstan-src/pull/3896

PR #3896, opened 2025-03-22
Remove count() narrowing handling of empty array

herndlm:
Is already covered by the adaption of https://github.com/phpstan/phpstan-src/pull/3895

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1067,7 +1067,6 @@ final class TypeSpecifier
 		if (
 			!$isNormalCount->yes()
 			|| (!$isConstantArray->yes() && !$isList->yes())
-			|| $type->isIterableAtLeastOnce()->no() // array{} cannot be used for further narrowing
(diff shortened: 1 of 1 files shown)


herndlm:
I realized late that this is slightly different though. the fix from @staabm is checking sizeType, while this was checking type. let's see what CI says, if in doubt, maybe it's better to keep? hmm

on the other hand I do prefer a simpler if here though if possible :)

herndlm:
I'll do a quick check for what I added this initially, can't remember

herndlm:
no clue, it doesn't influence anything. independent of @staabm's sizeType check. even if I remove the whole sizeType check line no new tests start failing. while developing things initially, at some point, I'm sure the isIterableAtLeastOnce()->no() was needed to fix failing tests. but most likely I refactored then and it became irrelevant.

if nobody else has another opinion, then I think it's fine to remove it

staabm:
agree we should not have code which does not fail tests when removed (except perf improvements)

ondrejmirtes:
Thank you.

PR #3898, opened 2025-03-22
Avoid falsely specifying never types via `count()`

herndlm:
Refs https://github.com/phpstan/phpstan-src/pull/3709#issuecomment-2745237530

The issue was that in a falsey scope with `count(array{stdClass}) < int<1, max>` it wasn't able to further narrow down the array, and return it as `sureNotType`, leading to `*NEVER*` in the end. Hope I'm not inversing the logic too much here.. Then it would try to use never with an array operation, leading to `*ERROR*` which basically made it then behave like mixed afterwards.

I hate the complex if conditions. This could be somehow more simplified I think, but there are some special cases in the `list-count.php` test which deals basically with removing arrays from a union of arrays by narrowing via count in both truthy and falsey scopes and I ended up with at least 2 of those not working properly anymore. So this, apparently better, exclusion strategy it is..

Unfortunately "unfixes" https://github.com/phpstan/phpstan/issues/1311, but https://phpstan.org/r/a1ce8144-60c8-43fc-bfac-fcb7086b3a25 shows it was never really properly fixed anyway.


ondrejmirtes:
Thank you!

PR #3900, opened 2025-03-23
Fix condition of fall-through case not used for exhaustive checks

rvanvelzen:
Fix phpstan/phpstan#11064. Unfortunately this causes a regression elsewhere, which by the looks of it already didn't work completely.

The problem there is that the assignment is forgotten because of the `$scopeForBranches = $scopeForBranches->filterByFalseyValue($fullCondExpr)`.

### DIFF
--- a/tests/PHPStan/Rules/Variables/IssetRuleTest.php
+++ b/tests/PHPStan/Rules/Variables/IssetRuleTest.php
@@ -277,10 +277,12 @@ class IssetRuleTest extends RuleTestCase
 				112,
 			],
 			[
-				'Variable $variableInFirstCase in isset() always exists and is not nullable.',
+				// could be Variable $variableInFirstCase in isset() always exists and is not nullable.
+				'Variable $variableInFirstCase in isset() is never defined.',
 				116,
 			],
 			[
+				// could be Variable $variableInSecondCase in isset() always exists and is not nullable.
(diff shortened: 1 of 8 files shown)


tests/PHPStan/Rules/Variables/IssetRuleTest.php
```diff
@@ -277,10 +277,12 @@ public function testVariableCertaintyInIsset(): void
...
 			],
 			[
-				'Variable $variableInFirstCase in isset() always exists and is not nullable.',
+				// could be Variable $variableInFirstCase in isset() always exists and is not nullable.
+				'Variable $variableInFirstCase in isset() is never defined.',
 				116,
 			],
 			[
+				// could be Variable $variableInSecondCase in isset() always exists and is not nullable.
 				'Variable $variableInSecondCase in isset() is never defined.',
```
rvanvelzen, line 286: So this error already was incorrect. The other one is now also broken, because its condition is also used

ondrejmirtes:
From the issue bot it looks like it fixed more issues: https://github.com/phpstan/phpstan-src/actions/runs/14018968530

Please add regression tests for them.

rvanvelzen:
Thanks, I added tests for the listed issues.

ondrejmirtes:
Awesome, thank you!

PR #3901, opened 2025-03-24
Fix narrowing of superglobals

herndlm:
Improvements of https://github.com/phpstan/phpstan-src/pull/3871, pulling in parts of https://github.com/phpstan/phpstan-src/pull/3762 basically

Fixes https://github.com/phpstan/phpstan/issues/12771
Fixes https://github.com/phpstan/phpstan/issues/12772
Maybe fixes https://github.com/phpstan/phpstan/issues/12776

I did not take over the changes where the expressions are passed over through classes, functions and namespaces in scopes. Those would deal with some edge cases, but not sure if it's worth.

fyi @staabm

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4833,6 +4833,8 @@ final class MutatingScope implements Scope
 	private function mergeVariableHolders(array $ourVariableTypeHolders, array $theirVariableTypeHolders): array
 	{
 		$intersectedVariableTypeHolders = [];
+		$globalVariableCallback = fn (Node $node) => $node instanceof Variable && is_string($node->name) && $this->isGlobalVariable($node->name);
+		$nodeFinder = new NodeFinder();
 		foreach ($ourVariableTypeHolders as $exprString => $variableTypeHolder) {
 			if (isset($theirVariableTypeHolders[$exprString])) {
 				if ($variableTypeHolder === $theirVariableTypeHolders[$exprString]) {
@@ -4842,12 +4844,22 @@ final class MutatingScope implements Scope
 
 				$intersectedVariableTypeHolders[$exprString] = $variableTypeHolder->and($theirVariableTypeHolders[$exprString]);
 			} else {
+				$expr = $variableTypeHolder->getExpr();
+				if ($nodeFinder->findFirst($expr, $globalVariableCallback) !== null) {
+					continue;
+				}
+
 				$intersectedVariableTypeHolders[$exprString] = ExpressionTypeHolder::createMaybe($variableTypeHolder->getExpr(), $variableTypeHolder->getType());
 			}
 		}
 
 		foreach ($theirVariableTypeHolders as $exprString => $variableTypeHolder) {
 			if (isset($intersectedVariableTypeHolders[$exprString])) {
+				continue;
+			}
+
+			$expr = $variableTypeHolder->getExpr();
+			if ($nodeFinder->findFirst($expr, $globalVariableCallback) !== null) {
 				continue;
 			}
 
--- a/tests/PHPStan/Analyser/nsrt/superglobals.php
+++ b/tests/PHPStan/Analyser/nsrt/superglobals.php
@@ -1,0 +1,69 @@
+<?php declare(strict_types=1);
+
+namespace Superglobals;
+
+use function PHPStan\Testing\assertNativeType;
+use function PHPStan\Testing\assertType;
+
+class Superglobals
+{
+
+	public function originalTypes(): void
+	{
+		assertType('array<mixed>', $GLOBALS);
+		assertType('array<mixed>', $_SERVER);
+		assertType('array<mixed>', $_GET);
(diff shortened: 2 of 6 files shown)


tests/PHPStan/Analyser/nsrt/superglobals.php
```diff
@@ -0,0 +1,59 @@
...
+	{
+		if (isset($GLOBALS['foo'])) {
+			assertType("non-empty-array&hasOffsetValue('foo', mixed~null)", $GLOBALS);
+			assertNativeType("non-empty-array<mixed>&hasOffset('foo')", $GLOBALS); // https://github.com/phpstan/phpstan/issues/8395
+		} else {
+			assertType('array<mixed>', $GLOBALS);
+			assertNativeType('array<mixed>', $GLOBALS);
+		}
+		assertType('array', $GLOBALS);
+		assertNativeType('array<mixed>', $GLOBALS);
```
herndlm: these 2 fail without the adaptions here

herndlm:
phpstan-doctrine failures are unrelated, those seemed to have appeared because of recent phpstan improvements, opened https://github.com/phpstan/phpstan-doctrine/pull/650 for it

tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
```diff
@@ -1001,4 +1001,9 @@ public function testHashing(): void
 		]);
 	}
 
+	public function testBug12772(): void
```
mvorisek, line 1004: Please add also `$_ENV['CI']` test explicitly, as I want to make sure GitHub CI environment variable is never assumed to exist on runtime - https://github.com/atk4/ui/blob/235416b31066827226e462f1916d9fc99c0d14b4/demos/init-db.php#L556. And thank you very much for looking into this issue!

Our reproducible environment is the same as here, GitHub CI, so as long as the tests passes here, our should pass then too.

herndlm: I get your point, but adaptions here are not related to any environment variables. If Ondřej is fine with this, I expect it to be merged very soon and then we should also see the effects

staabm (approved):
thanks for looking into it

src/Analyser/MutatingScope.php
```diff
@@ -554,10 +554,13 @@ public function getVariableType(string $variableName): Type
...
-			if ($this->isGlobalVariable($variableName)) {
-				return new ArrayType(new BenevolentUnionType([new IntegerType(), new StringType()]), new MixedType(true));
+			if (!$this->isGlobalVariable($variableName)) {
+				return new MixedType();
 			}
-			return new MixedType();
+
+			$superglobalType = new ArrayType(new BenevolentUnionType([new IntegerType(), new StringType()]), new MixedType(true));
+			$this->expressionTypes[$varExprString] = ExpressionTypeHolder::createYes(new Variable($variableName), $superglobalType);
+			$this->nativeExpressionTypes[$varExprString] = ExpressionTypeHolder::createYes(new Variable($variableName), $superglobalType);
```
ondrejmirtes: I don't get why the fix looks like this.

1) This is polluting `$this->expressionTypes` with any superglobal in the analysed code.
2) It's mutating Scope in a getter method which is a big red-flag.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
herndlm: good point, I don't understand how this even worked before without adding an expression tbh :P the idea of setting the expression for proper narrowing came from https://github.com/phpstan/phpstan-src/pull/3762#issuecomment-2574943147 basically.

not sure if relevant for the bug or fix, but after the expression is created it is dealing with late resolveable types still in the same method in https://github.com/phpstan/phpstan-src/blob/a11741d568f968f7b5938fb3c80e852a7952bd14/src/Analyser/MutatingScope.php#L566

herndlm: looks like I can't get around step-debugging this issue :D I assume the expressions are just overriding something that is incorrectly set and retained or so? let's see..

ondrejmirtes: I found a better fix, the root issue has to be addressed in mergeVariableHolder. Stay tuned.

ondrejmirtes: See https://github.com/phpstan/phpstan-src/pull/3901/commits/9710ba52e92e3cd45eb01af2d1761b5cfa5921d2

herndlm: oh cool, thx!

herndlm: those variable type holders were the missing piece for me I guess

ondrejmirtes:
Thank you.

herndlm:
thank _you_ for adapting the fix to be proper :)

PR #3903, opened 2025-03-26
Fix uopz signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12499

`\1` signatures wasn't working because of a typo.

Not sure which PR should still land in 1.12.x.


ondrejmirtes:
Thank you.

PR #3905, opened 2025-03-28
Narrow offsets for lists with HasOffsetType/HasOffsetValueType

staabm:
closes https://github.com/phpstan/phpstan/issues/12605 by implementing the idea described in https://github.com/phpstan/phpstan/issues/12605#issuecomment-2671206308


ondrejmirtes:
Thank you!

PR #3906, opened 2025-03-28
Difference between Instance and Static properties

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12775

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -83,6 +83,12 @@ final class ClassReflection
 	/** @var ExtendedPropertyReflection[] */
 	private array $properties = [];
 
+	/** @var ExtendedPropertyReflection[] */
+	private array $instanceProperties = [];
+
+	/** @var ExtendedPropertyReflection[] */
+	private array $staticProperties = [];
+
 	/** @var RealClassClassConstantReflection[] */
 	private array $constants = [];
 
@@ -148,6 +154,12 @@ final class ClassReflection
 
 	/** @var array<string, bool> */
 	private array $hasPropertyCache = [];
+
+	/** @var array<string, bool> */
+	private array $hasInstancePropertyCache = [];
+
+	/** @var array<string, bool> */
+	private array $hasStaticPropertyCache = [];
 
 	/**
 	 * @param PropertiesClassReflectionExtension[] $propertiesClassReflectionExtensions
@@ -449,6 +461,9 @@ final class ClassReflection
 		return $attributes !== [];
 	}
 
+	/**
+	 * @deprecated Use hasInstanceProperty or hasStaticProperty instead
+	 */
 	public function hasProperty(string $propertyName): bool
 	{
 		if (array_key_exists($propertyName, $this->hasPropertyCache)) {
@@ -468,11 +483,59 @@ final class ClassReflection
 			}
 		}
 
+		// For BC purpose
+		if ($this->getPhpExtension()->hasStaticProperty($this, $propertyName)) {
+			return $this->hasPropertyCache[$propertyName] = true;
+		}
+
 		if ($this->requireExtendsPropertiesClassReflectionExtension->hasProperty($this, $propertyName)) {
 			return $this->hasPropertyCache[$propertyName] = true;
 		}
 
 		return $this->hasPropertyCache[$propertyName] = false;
+	}
+
+	public function hasInstanceProperty(string $propertyName): bool
+	{
+		if (array_key_exists($propertyName, $this->hasInstancePropertyCache)) {
+			return $this->hasInstancePropertyCache[$propertyName];
+		}
+
+		if ($this->isEnum()) {
+			return $this->hasInstancePropertyCache[$propertyName] = $this->hasNativeProperty($propertyName);
+		}
+
+		foreach ($this->propertiesClassReflectionExtensions as $i => $extension) {
+			if ($i > 0 && !$this->allowsDynamicPropertiesExtensions()) {
+				break;
+			}
+			if ($extension->hasProperty($this, $propertyName)) {
+				return $this->hasInstancePropertyCache[$propertyName] = true;
+			}
+		}
+
+		if ($this->requireExtendsPropertiesClassReflectionExtension->hasInstanceProperty($this, $propertyName)) {
+			return $this->hasPropertyCache[$propertyName] = true;
+		}
+
+		return $this->hasPropertyCache[$propertyName] = false;
+	}
+
+	public function hasStaticProperty(string $propertyName): bool
+	{
+		if (array_key_exists($propertyName, $this->hasStaticPropertyCache)) {
+			return $this->hasStaticPropertyCache[$propertyName];
+		}
+
+		if ($this->getPhpExtension()->hasStaticProperty($this, $propertyName)) {
+			return $this->hasStaticPropertyCache[$propertyName] = true;
+		}
+
+		if ($this->requireExtendsPropertiesClassReflectionExtension->hasStaticProperty($this, $propertyName)) {
+			return $this->hasStaticPropertyCache[$propertyName] = true;
+		}
+
+		return $this->hasStaticPropertyCache[$propertyName] = false;
 	}
 
 	public function hasMethod(string $methodName): bool
@@ -619,6 +682,20 @@ final class ClassReflection
 
 			unset($this->properties[$name]);
 		}
+		foreach ($this->instanceProperties as $name => $property) {
+			if (!$property->isPrivate()) {
+				continue;
+			}
+
+			unset($this->instanceProperties[$name]);
+		}
+		foreach ($this->staticProperties as $name => $property) {
+			if (!$property->isPrivate()) {
+				continue;
+			}
+
+			unset($this->staticProperties[$name]);
+		}
 		foreach ($this->methods as $name => $method) {
 			if (!$method->isPrivate()) {
 				continue;
@@ -629,6 +706,7 @@ final class ClassReflection
 		$this->getPhpExtension()->evictPrivateSymbols($this->getCacheKey());
 	}
 
+	/** @deprecated Use getInstanceProperty or getStaticProperty */
 	public function getProperty(string $propertyName, ClassMemberAccessAnswerer $scope): ExtendedPropertyReflection
 	{
 		if ($this->isEnum()) {
@@ -658,6 +736,15 @@ final class ClassReflection
 			}
 		}
 
+		// For BC purpose
+		if ($this->getPhpExtension()->hasStaticProperty($this, $propertyName)) {
+			$property = $this->wrapExtendedProperty($this->getPhpExtension()->getStaticProperty($this, $propertyName));
+			if ($scope->canReadProperty($property)) {
+				return $this->properties[$key] = $property;
+			}
+			$this->properties[$key] = $property;
+		}
+
 		if (!isset($this->properties[$key])) {
 			if ($this->requireExtendsPropertiesClassReflectionExtension->hasProperty($this, $propertyName)) {
 				$property = $this->requireExtendsPropertiesClassReflectionExtension->getProperty($this, $propertyName);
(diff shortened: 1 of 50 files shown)


src/Reflection/RequireExtension/RequireExtendsPropertiesClassReflectionExtension.php
```diff
@@ -5,29 +5,108 @@
...
 		return $property;
 	}
 
-	private function findProperty(ClassReflection $classReflection, string $propertyName): ?ExtendedPropertyReflection
+	public function hasInstanceProperty(ClassReflection $classReflection, string $propertyName): bool
 	{
+		return $this->findProperty(
+				$classReflection,
+				$propertyName,
+				// TODO: Use hasInstanceProperty
```
VincentLanglet, line 47: TODO introducing the methods on the Type.

src/Reflection/Php/PhpClassReflectionExtension.php
```diff
@@ -177,6 +211,7 @@ public function getNativeProperty(ClassReflection $classReflection, string $prop
 		return $this->nativeProperties[$classReflection->getCacheKey()][$propertyName];
 	}
 
+	// TODO: Find the difference between createInstanceProperty and createStaticProperty
```
VincentLanglet: TODO, I need to update the implementation (and understand it before)

VincentLanglet:
I encounter some issues/questions for you @ondrejmirtes.

First, `HasPropertyType` is tricky
- since it's used for both static and non static properties.
- and `property_exists` returns true either it's a static or non static properties

So we have
```
class Foo
{
	public $foo;
	public static $staticFoo;
}


var_dump(property_exists(new Foo(), 'foo')); // TRUE
var_dump(property_exists(new Foo(), 'staticFoo')); // TRUE
Foo::$foo; // CRASH
(new Foo())->foo; // Works
Foo::$staticFoo; // Works
(new Foo())->staticFoo; // CRASH
```

Second, and kinda related, how should be considered properties from ObjectShapeType ?
Are they only instance properties ?

Also if you already have some comment on the existing work.

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -9118,10 +9118,6 @@ public function dataUnionProperties(): array
 				'UnionProperties\Bar|UnionProperties\Foo',
 				'$something->doSomething',
 			],
-			[
-				'UnionProperties\Bar|UnionProperties\Foo',
-				'$something::$doSomething',
```
VincentLanglet, line 8754: This throw an error in PHP, a static access to a instance property is not valid.

tests/PHPStan/Analyser/nsrt/bug-nullsafe-prop-static-access.php
```diff
@@ -26,5 +26,4 @@ function foo(?A $a): void
 	\PHPStan\Testing\assertType('string|null', $a?->b->get());
 
 	\PHPStan\Testing\assertType('int|null', $a?->b::$value);
-	\PHPStan\Testing\assertType('int|null', $a?->b->value);
```
VincentLanglet, line 29: An instance property to a static property is not valid

tests/PHPStan/Rules/Properties/TypesAssignedToPropertiesRuleTest.php
```diff
@@ -43,10 +43,6 @@ public function testTypesAssignedToProperties(): void
 				'Static property PropertiesAssignedTypes\Foo::$staticStringProperty (string) does not accept int.',
 				37,
 			],
-			[
-				'Property PropertiesAssignedTypes\Ipsum::$parentStringProperty (string) does not accept int.',
```
VincentLanglet, line 49: `parent::$parentStringProperty` is not valid since the property is an instance one.

`$this->parentStringProperty` is correctly reported

ondrejmirtes:
Yeah, these kinds of questions are great and generally are a positive effect of deprecating old methods - it forces us to do stuff correctly and fix existing problems along the way.

My thoughts:

* HasPropertyType is used alongside property_exists so representing both instance and static properties is fine. Not sure what kind of problem you have with that now. If you don't know what to do about this we can leave it for later (it's fine to leave some failing tests here for now).
* ObjectShapeType - should work only for instance properties.

src/Reflection/ClassReflection.php
```diff
@@ -683,9 +770,83 @@ public function getProperty(string $propertyName, ClassMemberAccessAnswerer $sco
...
+		}
+
+		if (!isset($this->instanceProperties[$key])) {
+			throw new MissingPropertyFromReflectionException($this->getName(), $propertyName);
+		}
+
+		return $this->instanceProperties[$key];
+	}
+
+	public function getStaticProperty(string $propertyName, ClassMemberAccessAnswerer $scope): ExtendedPropertyReflection
```
ondrejmirtes: I feel like this isn't calling any extension (static magic properties aren't a thing in PHP) so this method shouldn't need `$scope`.

src/Reflection/ClassReflection.php
```diff
@@ -683,9 +770,83 @@ public function getProperty(string $propertyName, ClassMemberAccessAnswerer $sco
...
+			$key = sprintf('%s-%s', $key, $scope->getClassReflection()->getCacheKey());
+		}
+
+		if (!isset($this->staticProperties[$key])) {
+			if ($this->getPhpExtension()->hasStaticProperty($this, $propertyName)) {
+				$property = $this->wrapExtendedProperty($this->getPhpExtension()->getStaticProperty($this, $propertyName));
+				if ($scope->canReadProperty($property)) {
+					return $this->staticProperties[$key] = $property;
+				}
+				$this->staticProperties[$key] = $property;
```
ondrejmirtes: What if this would not check `Scope::canReadProperty` and would always return?

ondrejmirtes (changes requested):

VincentLanglet:
In the new list of todo to see with you @ondrejmirtes I have:
- `PHPClassReflectionExtension::createProperty` to maybe split into createInstanceProperty and createStaticProperty (?) ; I'm not familiar with the implementation yet.
- `HasPropertyType::isSuperTypeOf` => remplace usage of `hasProperty`
- `HasProperty::isSubTypeOf` => remplace usage of `hasProperty`
- Having or not a difference between `HasProperty::hasInstanceProperty` and `HasProperty::hasStaticProperty` and updating the `new HasProperty()` eventually.
- Deciding if the LevelsIntegrationTest change is ok. (It's UnionType related, when one of the class has not the property and the other has the static property of the same name)

> HasPropertyType is used alongside property_exists so representing both instance and static properties is fine. Not sure what kind of problem you have with that now. If you don't know what to do about this we can leave it for later (it's fine to leave some failing tests here for now).

Currently the TypeSpecifier does
```
if (
					$var instanceof PropertyFetch
					&& $var->name instanceof Node\Identifier
				) {
					$types = $types->unionWith(
						$this->create($var->var, new IntersectionType([
							new ObjectWithoutClassType(),
							new HasPropertyType($var->name->toString()),
						]), TypeSpecifierContext::createTruthy(), $scope)->setRootExpr($expr),
					);
				} elseif (
					$var instanceof StaticPropertyFetch
					&& $var->class instanceof Expr
					&& $var->name instanceof Node\VarLikeIdentifier
				) {
					$types = $types->unionWith(
						$this->create($var->class, new IntersectionType([
							new ObjectWithoutClassType(),
							new HasPropertyType($var->name->toString()),
						]), TypeSpecifierContext::createTruthy(), $scope)->setRootExpr($expr),
					);
				}
```
Which I think, means that accessing to `Foo::$bar` makes PHPStan thinks that `foo->bar` exists. (and the opposite)

So should we add a param to the HasPropertyType construct which would be `STATIC|INSTANCE|UNKNOWN`

src/Reflection/ClassReflection.php
VincentLanglet: Scope is not used by extension in other methods but
- the cache key
- `canReadProperty` check

Btw, I currently call `RequireExtendsPropertiesClassReflectionExtension::hasStaticProperty` and `getStaticProperty`. Which is almost like extensions with `hasProperty/getProperty`.

src/Reflection/ClassReflection.php
VincentLanglet: Wouldn't this an issue for `protected/private` static properties ?

This will skip the possible override by requireExtendsPropertiesClassReflectionExtension::getStaticProperty

VincentLanglet:
I think it's ready for a new review @ondrejmirtes.

To me:
- Intregration tests error message failure is ok and need to be updated
- Extension tests failure is ok, deprecation will need to be solve in those.

VincentLanglet:
Rebased and fix, it should be ready for a new review @ondrejmirtes :)

VincentLanglet:
It's rebased and fixed I think.

I know it's not an easy PR to review @ondrejmirtes but it would be awesome if we succeed closing it :) 
Is there something I can do to help the review ?

VincentLanglet:
It's rebased and fixed again.

This PR is really hard to maintain, are you (still ?) interested in it @ondrejmirtes ?
Can I do something to help you to review it ?

src/Reflection/RequireExtension/RequireExtendsPropertiesClassReflectionExtension.php
```diff
@@ -5,28 +5,100 @@
...
-use PHPStan\Reflection\PropertiesClassReflectionExtension;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\TrinaryLogic;
+use PHPStan\Type\Type;
 
-final class RequireExtendsPropertiesClassReflectionExtension implements PropertiesClassReflectionExtension
+final class RequireExtendsPropertiesClassReflectionExtension
 {
 
+	/** @deprecated Use hasInstanceProperty or hasStaticProperty */
```
ondrejmirtes, line 15: We don't need to keep this, RequireExtendsPropertiesClassReflectionExtension is not part of `@api`.

src/Reflection/Php/PhpClassReflectionExtension.php
```diff
@@ -168,6 +185,28 @@ public function getProperty(ClassReflection $classReflection, string $propertyNa
 		return $this->propertiesIncludingAnnotations[$classReflection->getCacheKey()][$propertyName];
 	}
 
+	public function hasStaticProperty(ClassReflection $classReflection, string $propertyName): bool
```
ondrejmirtes: I don't get why this is added. PhpClassReflectionExtension only gives native properties so they can only be or the other. I think code calling this can call `hasProperty`/`getProperty` instead and then ask if it `$property->isStatic()` or not.

src/Reflection/Php/PhpClassReflectionExtension.php
```diff
@@ -168,6 +185,28 @@ public function getProperty(ClassReflection $classReflection, string $propertyNa
...
+	public function getStaticProperty(ClassReflection $classReflection, string $propertyName): ExtendedPropertyReflection
+	{
+		if (!isset($this->staticPropertiesIncludingAnnotations[$classReflection->getCacheKey()][$propertyName])) {
+			$this->staticPropertiesIncludingAnnotations[$classReflection->getCacheKey()][$propertyName] = $this->createProperty($classReflection, $propertyName, true);
+		}
+
+		return $this->staticPropertiesIncludingAnnotations[$classReflection->getCacheKey()][$propertyName];
+	}
+
+	public function hasNativeProperty(ClassReflection $classReflection, string $propertyName): bool
```
ondrejmirtes: Again, calling hasProperty should be sufficient.

src/Reflection/Php/PhpClassReflectionExtension.php
```diff
@@ -156,7 +170,10 @@ public function evictPrivateSymbols(string $classCacheKey): void
 
 	public function hasProperty(ClassReflection $classReflection, string $propertyName): bool
 	{
-		return $classReflection->getNativeReflection()->hasProperty($propertyName);
```
ondrejmirtes: Why is implementation of hasProperty changed?

src/Type/Accessory/HasPropertyType.php
```diff
@@ -72,7 +72,10 @@ public function accepts(Type $type, bool $strictTypes): AcceptsResult
 
 	public function isSuperTypeOf(Type $type): IsSuperTypeOfResult
 	{
-		return new IsSuperTypeOfResult($type->hasProperty($this->propertyName), []);
+		return new IsSuperTypeOfResult(
+			$type->hasInstanceProperty($this->propertyName)->or($type->hasStaticProperty($this->propertyName)),
```
ondrejmirtes, line 76: When is `HasPropertyType` being used to represent static properties?

ondrejmirtes (changes requested):
Really nice, just a few minor questions 😊

src/Reflection/RequireExtension/RequireExtendsPropertiesClassReflectionExtension.php
VincentLanglet: But this method is used by `ClassReflection::getProperty` which is deprecated but part of `@api`.

src/Reflection/Php/PhpClassReflectionExtension.php
VincentLanglet: This is the whole issue.

PhpClassReflectionExtension does not gives only native properties.
The `createProperty` method also read annotation in the code
```
if (
			$includingAnnotations
			&& !$declaringClassReflection->isEnum()
			&& $this->annotationsPropertiesClassReflectionExtension->hasProperty($classReflection, $propertyName)
		) {
			$hierarchyDistances = $classReflection->getClassHierarchyDistances();
			$annotationProperty = $this->annotationsPropertiesClassReflectionExtension->getProperty($classReflection, $propertyName);
			if (!isset($hierarchyDistances[$annotationProperty->getDeclaringClass()->getName()])) {
			}

			$distanceDeclaringClass = $propertyReflection->getDeclaringClass()->getName();
			$propertyTrait = $this->findPropertyTrait($propertyReflection);
			if ($propertyTrait !== null) {
				$distanceDeclaringClass = $propertyTrait;
			}
			if (!isset($hierarchyDistances[$distanceDeclaringClass])) {
			}

			if ($hierarchyDistances[$annotationProperty->getDeclaringClass()->getName()] <= $hierarchyDistances[$distanceDeclaringClass]) {
				return $annotationProperty;
			}
		}
```

So we need a different getProperty and getStaticProperty because the first one use the call
```
$this->createProperty($classReflection, $propertyName, true)
```
while the second should rely only on native type and be called with
```
$this->createProperty($classReflection, $propertyName, false)
```

src/Reflection/Php/PhpClassReflectionExtension.php
VincentLanglet: Related to discussion https://github.com/phpstan/phpstan-src/pull/3906#discussion_r2328641975

src/Reflection/Php/PhpClassReflectionExtension.php
VincentLanglet: Related to discussion https://github.com/phpstan/phpstan-src/pull/3906#discussion_r2328641975

src/Type/Accessory/HasPropertyType.php
VincentLanglet: `property_exists` returns true for both instance and static properties.

It's used in PropertyExistsTypeSpecifyingExtension but also in TypeSpecifier
```
if (
					$var instanceof PropertyFetch
					&& $var->name instanceof Node\Identifier
				) {
					$types = $types->unionWith(
						$this->create($var->var, new IntersectionType([
							new ObjectWithoutClassType(),
							new HasPropertyType($var->name->toString()),
						]), TypeSpecifierContext::createTruthy(), $scope)->setRootExpr($expr),
					);
				} elseif (
					$var instanceof StaticPropertyFetch
					&& $var->class instanceof Expr
					&& $var->name instanceof Node\VarLikeIdentifier
				) {
					$types = $types->unionWith(
						$this->create($var->class, new IntersectionType([
							new ObjectWithoutClassType(),
							new HasPropertyType($var->name->toString()),
						]), TypeSpecifierContext::createTruthy(), $scope)->setRootExpr($expr),
					);
				}
```

Currently, it's seems to complicated to introduce both `HasInstancePropertyType` and `HasStaticPropertyType`

VincentLanglet:
Rebased and I had to introduce new fixes because of recent bugfixes.

Should be re-ready for review @ondrejmirtes

staabm:
nit: PR description contains a wrong link

VincentLanglet:
> nit: PR description contains a wrong link

Thanks, updated to https://github.com/phpstan/phpstan/issues/12775

ondrejmirtes:
Hi, I'm taking over this PR and taking it over the finish line. Already have a local branch with some changes. Stay tuned!

ondrejmirtes:
Thank you!

VincentLanglet:
Thank you too !

PR #3908, opened 2025-03-28
Fix lost list-type if substituted a element via loop

staabm:
closes https://github.com/phpstan/phpstan/issues/12274

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5450,14 +5450,15 @@ final class NodeScopeResolver
 			$offsetValueType = $varType;
 			$offsetNativeValueType = $varNativeType;
 
-			$valueToWrite = $this->produceArrayDimFetchAssignValueToWrite($offsetTypes, $offsetValueType, $valueToWrite);
+			$valueToWrite = $this->produceArrayDimFetchAssignValueToWrite($dimFetchStack, $offsetTypes, $offsetValueType, $valueToWrite, $scope);
 
 			if (!$offsetValueType->equals($offsetNativeValueType) || !$valueToWrite->equals($nativeValueToWrite)) {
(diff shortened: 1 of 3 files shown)


staabm:
I am not sure about the impl, as it e.g. does not work for the regular `for` loop as we don't have `$items[$i]` in the scope expressions (don't know whether that is intentional):

```
/**
 * @param non-empty-list<int> $items
 *
 * @return non-empty-list<int>
 */
function getItemsWithForLoop(array $items): array
{
	for($i = 0; $i < count($items); $i++) {
		$items[$i] = 1;
	}

	assertType('non-empty-list<int>', $items);
	return $items;
}
```

edit: logged as a [new issue](https://github.com/phpstan/phpstan/issues/12807) so its not forgotten

ondrejmirtes (changes requested):
This is nice, I have two ideas for improvements:

1) Try to do this logic in produceArrayDimFetchAssignValueToWrite instead, I suspect it might work for nested assigns (`$a[$i][$j]`) as well.
2) Try calling `setExistingOffsetValueType` instead of `setOffsetValueType` in this scenario. Right now I think it's used for some weird thing around `unset` but I suspect it might be a bit more elegant. We need to fix how `AccessoryArrayListType::setExistingOffsetValueType` is implemented too.

These are just ideas - if they don't work, I'm happy to revert to the current state and merge it as it is. Thanks!

staabm:
I have implemented 1) - it works like you suspected.

regarding 2) - what are the expectations with `setExistingOffsetValueType`? I don't see why it would be more elegant or whether it will cover a situation `setOffsetValueType` won't. what are the differences between this 2 methods?

ondrejmirtes:
> setExistingOffsetValueType

Nevermind, probably not a good idea 😊

ondrejmirtes:
Thank you!

PR #3909, opened 2025-03-28
Set offset on list keeps list if there's HasOffsetType for all preceeding offsets

staabm:
taking the from https://github.com/phpstan/phpstan-src/pull/3905 and put it onto setOffsetValueType, to preserve list-state

### DIFF
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -800,8 +800,30 @@ class IntersectionType implements CompoundType
 
 		$result = $this->intersectTypes(static fn (Type $type): Type => $type->setOffsetValueType($offsetType, $valueType, $unionValues));
 
-		if ($offsetType !== null && $this->isList()->yes() && $this->isIterableAtLeastOnce()->yes() && (new ConstantIntegerType(1))->isSuperTypeOf($offsetType)->yes()) {
-			$result = TypeCombinator::intersect($result, new AccessoryArrayListType());
+		if (
+			$offsetType !== null
+			&& $this->isList()->yes()
+			&& !$result->isList()->yes()
+		) {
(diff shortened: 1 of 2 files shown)


src/Type/IntersectionType.php
```diff
@@ -800,8 +800,30 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
 
 		$result = $this->intersectTypes(static fn (Type $type): Type => $type->setOffsetValueType($offsetType, $valueType, $unionValues));
 
-		if ($offsetType !== null && $this->isList()->yes() && $this->isIterableAtLeastOnce()->yes() && (new ConstantIntegerType(1))->isSuperTypeOf($offsetType)->yes()) {
```
staabm, line 803: this previously existing logic, which was implemented for https://github.com/phpstan/phpstan/issues/12131 - only worked for index `1`. the new logic works for the generic case

staabm:
//cc @herndlm

ondrejmirtes:
Thanks! We need something similar in NodeScopeResolver in AST-based logic.

If we have a $list and we know that $list[$i] exists based on Scope::hasExpressionType (because $i isn't narrow enough, it's just an int), assigning $list[$i] should keep the list-ness of the variable.

staabm:
> If we have a $list and we know that $list[$i] exists based on Scope::hasExpressionType (because $i isn't narrow enough, it's just an int), assigning $list[$i] should keep the list-ness of the variable.

did you see https://github.com/phpstan/phpstan-src/pull/3908 or do you mean someting different?

staabm:
I think I understand now, and you mean: https://phpstan.org/r/323b3300-934b-46f3-bdd4-b428d2e8a9c1

ondrejmirtes:
Yeah, I think it's #3908 actually :D

PR #3910, opened 2025-03-29
Add stringable access check to ClassConstantRule

zonuexe:
refs https://github.com/phpstan/phpstan-src/pull/3885#pullrequestreview-2694393342, https://github.com/phpstan/phpstan-src/pull/3886#issuecomment-2735632469, https://github.com/phpstan/phpstan/issues/13238

In the context of class constant fetching, string keys are not implicitly cast.
* https://3v4l.org/WZTCC

### DIFF
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -17,9 +17,6 @@ conditionalTags:
 
 services:
 	-
-		class: PHPStan\Rules\Classes\ClassConstantRule
-
-	-
 		class: PHPStan\Rules\Classes\NewStaticInAbstractClassStaticMethodRule
 
 	-
--- a/conf/config.level0.neon
+++ b/conf/config.level0.neon
@@ -17,6 +17,9 @@ conditionalTags:
 
 services:
 	-
+		class: PHPStan\Rules\Classes\ClassConstantRule
+
+	-
 		class: PHPStan\Rules\Classes\NewStaticInAbstractClassStaticMethodRule
 
 	-
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -24,6 +24,7 @@ parameters:
 			tooWideThrowType: true
 	featureToggles:
 		bleedingEdge: false
+		checkNonStringableDynamicAccess: false
 		checkParameterCastableToNumberFunctions: false
 		skipCheckGenericClasses: []
 		stricterFunctionMap: false
--- a/src/Rules/Classes/ClassConstantRule.php
+++ b/src/Rules/Classes/ClassConstantRule.php
@@ -76,9 +76,8 @@ final class ClassConstantRule implements Rule
 					static fn (Type $type) => $type->isString()->yes(),
 				);
 
-				$type = $nameTypeResult->getType();
-
-				if (!$type->isString()->yes()) {
+				$nameType = $nameTypeResult->getType();
+				if (!$nameType instanceof ErrorType && !$nameType->isString()->yes()) {
 					$className = $node->class instanceof Name
 						? $scope->resolveName($node->class)
 						: $scope->getType($node->class)->describe(VerbosityLevel::typeOnly());
--- a/src/Rules/Classes/ClassConstantRule.php
+++ b/src/Rules/Classes/ClassConstantRule.php
@@ -67,14 +67,14 @@ final class ClassConstantRule implements Rule
 			}
 
 			if ($this->checkNonStringableDynamicAccess) {
-				$typeResult = $this->ruleLevelHelper->findTypeToCheck(
+				$nameTypeResult = $this->ruleLevelHelper->findTypeToCheck(
 					$scope,
 					$node->name,
 					'',
 					static fn (Type $type) => $type->isString()->yes(),
 				);
 
-				$type = $typeResult->getType();
+				$type = $nameTypeResult->getType();
 
 				if (!$type->isString()->yes()) {
 					$className = $node->class instanceof Name
--- a/src/Rules/Classes/ClassConstantRule.php
+++ b/src/Rules/Classes/ClassConstantRule.php
@@ -81,8 +81,8 @@ final class ClassConstantRule implements Rule
 						? $scope->resolveName($node->class)
 						: $scope->getType($node->class)->describe(VerbosityLevel::typeOnly());
 
-					$errors[] = RuleErrorBuilder::message(sprintf('Cannot fetch constant from %s with a non-stringable type %s.', $className, $nameType->describe(VerbosityLevel::precise())))
-						->identifier('classConstant.fetchInvalidExpression')
+					$errors[] = RuleErrorBuilder::message(sprintf('Class constant name for %s must be a string, but %s was given.', $className, $nameType->describe(VerbosityLevel::precise())))
+						->identifier('classConstant.nameNotString')
 						->build();
 				}
 			}
--- a/src/Rules/Classes/ClassConstantRule.php
+++ b/src/Rules/Classes/ClassConstantRule.php
@@ -5,6 +5,7 @@ namespace PHPStan\Rules\Classes;
 use PhpParser\Node;
 use PhpParser\Node\Expr\BinaryOp\Identical;
 use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Name;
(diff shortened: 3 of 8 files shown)


src/Rules/Classes/ClassConstantRule.php
```diff
@@ -39,6 +39,7 @@ public function __construct(
 		private RuleLevelHelper $ruleLevelHelper,
 		private ClassNameCheck $classCheck,
 		private PhpVersion $phpVersion,
+		private bool $checkNonStringableDynamicAccess = true,
```
zonuexe: I understand that we shouldn't use the default value, but I was getting the following error, probably due to a missing DI configuration, so I temporarily added `= true` to test.

```
% make phpstan
php bin/phpstan clear-result-cache -q && php -d memory_limit=448M bin/phpstan

In Resolver.php line 677:

  Service 'rules.26' (type of PHPStan\Rules\Classes\ClassConstantRule): Parameter $checkNonStringableDynamicAccess in ClassConstantRule::__construct() has no class type or default value, so its value must be
  specified.
```

src/Rules/Classes/ClassConstantRule.php
```diff
@@ -60,6 +61,24 @@ public function processNode(Node $node, Scope $scope): array
 				$name = $constantString->getValue();
 				$constantNameScopes[$name] = $scope->filterByTruthyValue(new Identical($node->name, new String_($name)));
 			}
+
+			if ($this->checkNonStringableDynamicAccess) {
+				$accepts = $this->ruleLevelHelper->accepts(new StringType(), $nameType, true);
```
ondrejmirtes: Delete this line, it's not used anywhere.

src/Rules/Classes/ClassConstantRule.php
```diff
@@ -60,6 +61,24 @@ public function processNode(Node $node, Scope $scope): array
...
+				$typeResult = $this->ruleLevelHelper->findTypeToCheck(
+					$scope,
+					$node->name,
+					'',
+					static fn (Type $type) => $type->toString()->isString()->yes()
+				);
```
ondrejmirtes, line 77: After this call there should be:

```php
		$type = $typeResult->getType();
		if ($type instanceof ErrorType) {
			return [];
		}
```

Some rules return "unknown class errors" but that's irrelevant here.

tests/PHPStan/Rules/Classes/ClassConstantRuleTest.php
```diff
@@ -19,6 +19,8 @@ class ClassConstantRuleTest extends RuleTestCase
 
 	private int $phpVersion;
 
+	private bool $checkNonStringableDynamicAccess;
```
ondrejmirtes: If all the test cases use `true` then this property isn't needed at all. Just use `true` in the `new` expression in `getRule`.

src/Rules/Classes/ClassConstantRule.php
```diff
@@ -60,6 +61,24 @@ public function processNode(Node $node, Scope $scope): array
...
+
+				if (! $typeResult->getType()->isString()->yes() ||
+					$typeResult->getType()->toString()->isNumericString()->yes()
+				) {
+					$errors[] = RuleErrorBuilder::message(sprintf('Cannot fetch class constant with a non-stringable type %s.', $nameType->describe(VerbosityLevel::typeOnly())))
+						->identifier('classConstant.fetchInvalidExpression')
```
ondrejmirtes: The error message and the identifier should reflect we're trying to access a class constant by non-stringable name. That's not obvious.

ondrejmirtes: Also it might be nice to mention the class name here.

conf/config.neon
```diff
@@ -882,6 +883,11 @@ services:
 	-
 		class: PHPStan\Rules\ClassForbiddenNameCheck
 
+	-
+		class: PHPStan\Rules\Classes\ClassConstantRule
+		arguments:
+			checkNonStringableDynamicAccess: %featureToggles.checkNonStringableDynamicAccess%
```
ondrejmirtes: Why are you adding the service here in this file? Remove it please.

ondrejmirtes: Instead, config.level0.neon should be modified:

1) Remove the rule from `rules` section.
2) Add the service same way you did here.
3) But also add `tags` section with the corresponding tag.

src/Rules/Classes/ClassConstantRule.php
ondrejmirtes: My suggestion above will help you get rid of this error.

ondrejmirtes (changes requested):

ondrejmirtes:
I'd love if you could finish this (and the other rules) 😊

src/Rules/Classes/ClassConstantRule.php
zonuexe: Dynamic fetching of constants [doesn't do implicit casting](https://3v4l.org/WZTCC), so it seems appropriate to simply check `isString()->yes()` without `toString()`.

I was wrong in the early stages of this PR.

src/Rules/Classes/ClassConstantRule.php
zonuexe: Fixed it.

conf/config.level0.neon
```diff
@@ -16,6 +16,9 @@ conditionalTags:
 		phpstan.rules.rule: %featureToggles.newStaticInAbstractClassStaticMethod%
 
 services:
+	-
+		class: PHPStan\Rules\Classes\ClassConstantRule
```
ondrejmirtes: Why is this added here? It's not needed at all, the rule has `#[RegisteredRule(level: 0)]` attribute.

src/Rules/Classes/ClassConstantRule.php
ondrejmirtes: 1) Do not make this optional.
2) Put `#[AutowiredParameter(ref: '%featureToggles.checkNonStringableDynamicAccess%')]` above the parameter.

ondrejmirtes: After you `composer dump` it should work.

src/Rules/Classes/ClassConstantRule.php
ondrejmirtes: I'd rename the variable to `$nameTypeResult`.

src/Rules/Classes/ClassConstantRule.php
ondrejmirtes: The error message is wrong regarding the isString/toString point you made above. This is no longer about "stringables", but about "strings". Also the identifier should reflect it's about "name".

ondrejmirtes (changes requested):

src/Rules/Classes/ClassConstantRule.php
```diff
@@ -63,6 +67,27 @@ public function processNode(Node $node, Scope $scope): array
...
+
+			if ($this->checkNonStringableDynamicAccess) {
+				$nameTypeResult = $this->ruleLevelHelper->findTypeToCheck(
+					$scope,
+					$node->name,
+					'',
+					static fn (Type $type) => $type->isString()->yes(),
+				);
+
+				$type = $nameTypeResult->getType();
```
ondrejmirtes: `$type` -> `$nameType`

src/Rules/Classes/ClassConstantRule.php
```diff
@@ -63,6 +67,27 @@ public function processNode(Node $node, Scope $scope): array
...
+
+				if (!$type->isString()->yes()) {
```
ondrejmirtes:
```suggestion
				if (!$type instanceof ErrorType && !$type->isString()->yes()) {
```

ondrejmirtes: All calls to findTypeToCheck follow this pattern. `ErrorType` is returned if unknown classes were encountered there.

ondrejmirtes (changes requested):
After this I'll merge this and you can make the same changes to other rules 👍

zonuexe:
@ondrejmirtes Fixed it. I will prepare a PR for the rules for each node.

ondrejmirtes:
Thank you!

PR #3586, opened 2024-10-27
Fix signature type for default-null parameters

staabm:
closes https://github.com/phpstan/phpstan/issues/7522

PR #3912, opened 2025-03-30
Fix signature type for default-null parameters

### LINKED ISSUE #113: Stub for ob_start missing null type for callback argument for 8.0 <= PHP < 8.4
The stub for `ob_start` is missing `null` in the type specification for the `$callback` argument for 8.0 <= PHP < 8.4.

https://github.com/phpstan/php-8-stubs/blob/main/stubs/ext/standard/ob_start.php

It's not clear if the omission of `null` from the type specification before 8.4 was deliberate. The docs for PHP do not mention this. They also include an example whereby a null is provided as the value for the `$callback` argument, which appears to work fine with 8.0 <= PHP < 8.4.

https://www.php.net/manual/en/function.ob-start.php

So this appears to be a bug in the stub. It is causing PHPStan to report incorrect errors against valid code, including against the example provided in the PHP docs:

https://github.com/phpstan/phpstan/issues/12282

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/7522
Closes https://github.com/phpstan/phpstan/issues/12282 (duplicate)

Closes https://github.com/phpstan/php-8-stubs/issues/113

Closes https://github.com/phpstan/phpstan-src/pull/3586 (cherry-picked from staabm https://github.com/phpstan/phpstan-src/pull/3586#issuecomment-2764622127)

### DIFF
--- a/src/Reflection/SignatureMap/Php8SignatureMapProvider.php
+++ b/src/Reflection/SignatureMap/Php8SignatureMapProvider.php
@@ -3,6 +3,7 @@
 namespace PHPStan\Reflection\SignatureMap;
 
 use PhpParser\Node\AttributeGroup;
+use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\Stmt\ClassConst;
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Here you can also see the effect it has on all signatures: https://github.com/phpstan/phpstan-src/actions/runs/14158052783/job/39659896386

ondrejmirtes:
Thank you!

PR #3913, opened 2025-03-31
Disable purity check for non final methods

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12382

### DIFF
--- a/src/Rules/Pure/FunctionPurityCheck.php
+++ b/src/Rules/Pure/FunctionPurityCheck.php
@@ -92,6 +92,11 @@ final class FunctionPurityCheck
 				count($throwPoints) === 0
 				&& count($impurePoints) === 0
 				&& count($functionReflection->getAsserts()->getAll()) === 0
+				&& (
+					!$functionReflection instanceof ExtendedMethodReflection
+					|| $functionReflection->isFinal()->yes()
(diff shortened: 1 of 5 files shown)


src/Rules/Pure/FunctionPurityCheck.php
```diff
@@ -86,6 +86,11 @@ public function check(
 				count($throwPoints) === 0
 				&& count($impurePoints) === 0
 				&& count($functionReflection->getAsserts()->getAll()) === 0
+				&& (
+					!$functionReflection instanceof ExtendedMethodReflection
```
staabm, line 96: Do we need to support `@final` phpdoc?

VincentLanglet: It would be better indeed.

It's not included in those isFinal methods ?

staabm: I am not sure. A test could tell :-)

VincentLanglet: tests added

ondrejmirtes:
Please rebase and fix the conflicts, then I can merge it.

ondrejmirtes:
Thank you!

PR #3914, opened 2025-03-31
Fix mb_convert_encoding signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12309

Closes https://github.com/phpstan/phpstan-src/pull/3749

Following the example https://3v4l.org/OrdDb
- mb_convert_encoding accepts array of non-string values, and just ignores them
- mb_convert_encoding accepts array of array and works recursively


VincentLanglet:
Friendly ping @ondrejmirtes in case this PR was missed.

ondrejmirtes:
Thank you.

PR #2888, opened 2024-01-23
Improve `date()` return types

zonuexe:
resolve https://github.com/phpstan/phpstan/issues/10468 and https://github.com/phpstan/phpstan/issues/6613

PR #3919, opened 2025-04-05
Fix DateTime::format('u') return type

mvorisek:
fix https://github.com/phpstan/phpstan-src/pull/2888 and https://phpstan.org/r/251087ce-9f59-4c2b-853b-b7d0c91cc584

### DIFF
--- a/src/Type/Php/DateFunctionReturnTypeHelper.php
+++ b/src/Type/Php/DateFunctionReturnTypeHelper.php
@@ -76,8 +76,12 @@ final class DateFunctionReturnTypeHelper
 				return $this->buildNumericRangeType(0, 1, false);
 			case 'u':
 				return $useMicrosec
-					? new IntersectionType([new StringType(), new AccessoryNonFalsyStringType()])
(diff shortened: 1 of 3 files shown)


src/Type/Php/DateFunctionReturnTypeHelper.php
```diff
@@ -76,7 +76,7 @@ public function buildReturnTypeFromFormat(string $formatString, bool $useMicrose
 				return $this->buildNumericRangeType(0, 1, false);
 			case 'u':
 				return $useMicrosec
-					? new IntersectionType([new StringType(), new AccessoryNonFalsyStringType()])
+					? new IntersectionType([new StringType(), new AccessoryNumericStringType()])
```
ondrejmirtes: Please keep the non-falsy type as well. Also, you could add type inference for `v`.

ondrejmirtes (changes requested):

src/Type/Php/DateFunctionReturnTypeHelper.php
mvorisek: done

ondrejmirtes:
Thank you.

PR #3921, opened 2025-04-06
Add `strrev()` to `NonEmptyStringFunctionsReturnTypeExtension`

nsfisis:
# Problem

If `strrev()` takes `non-falsy-string` or `non-empty-string`, it should return `non-falsy-string` or `non-empty-string`, but it returns just a `string`.

https://phpstan.org/r/f9b50c8a-1a20-43ff-8715-42f267a28e24

```php
\PHPStan\dumpType(strrev("abc"));
// => string
```

# Solution

Add `strrev()` to `NonEmptyStringFunctionsReturnTypeExtension`.

### DIFF
--- a/src/Type/Php/NonEmptyStringFunctionsReturnTypeExtension.php
+++ b/src/Type/Php/NonEmptyStringFunctionsReturnTypeExtension.php
@@ -34,6 +34,7 @@ final class NonEmptyStringFunctionsReturnTypeExtension implements DynamicFunctio
 			'preg_quote',
 			'rawurlencode',
 			'rawurldecode',
+			'strrev',
 		], true);
 	}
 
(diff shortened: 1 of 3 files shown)


VincentLanglet:
Hi @nsfisis 

Best would be to have a StrrevFunctionReturnTypeExtension because
- We can know the strrev of a constant string
- A lowercase string will still be lowercase
- A uppercase string will still be uppercase

VincentLanglet:
> Hi @nsfisis
> 
> Best would be to have a StrrevFunctionReturnTypeExtension because
> 
> * We can know the strrev of a constant string
> * A lowercase string will still be lowercase
> * A uppercase string will still be uppercase

I did the full implementation in https://github.com/phpstan/phpstan-src/pull/3960

nsfisis:
Close this PR in favor of #3960. Thank you @VincentLanglet!

PR #3924, opened 2025-04-07
non-empty-array<mixed> does not accept array with hasOffsetValue

staabm:
closes https://github.com/phpstan/phpstan/issues/12847

**edit:** the underlying issue was fixed with https://github.com/phpstan/phpstan-src/pull/3937

### DIFF
--- a/src/Type/Accessory/HasOffsetValueType.php
+++ b/src/Type/Accessory/HasOffsetValueType.php
@@ -122,10 +122,6 @@ class HasOffsetValueType implements CompoundType, AccessoryType
 
 	public function isAcceptedBy(Type $acceptingType, bool $strictTypes): AcceptsResult
 	{
-		if ($acceptingType->isArray()->yes() && $acceptingType->isIterableAtLeastOnce()->yes()) {
-			return AcceptsResult::createYes();
-		}
-
 		return $this->isSubTypeOf($acceptingType)->toAcceptsResult();
 	}
 
--- a/src/Type/Accessory/HasOffsetValueType.php
+++ b/src/Type/Accessory/HasOffsetValueType.php
@@ -122,6 +122,10 @@ class HasOffsetValueType implements CompoundType, AccessoryType
 
 	public function isAcceptedBy(Type $acceptingType, bool $strictTypes): AcceptsResult
 	{
+		if ($acceptingType instanceof NonEmptyArrayType) {
+			return AcceptsResult::createYes();
+		}
+
 		return $this->isSubTypeOf($acceptingType)->toAcceptsResult();
 	}
 
--- a/src/Type/Accessory/HasOffsetValueType.php
+++ b/src/Type/Accessory/HasOffsetValueType.php
@@ -122,7 +122,7 @@ class HasOffsetValueType implements CompoundType, AccessoryType
 
 	public function isAcceptedBy(Type $acceptingType, bool $strictTypes): AcceptsResult
 	{
-		if ($acceptingType instanceof NonEmptyArrayType) {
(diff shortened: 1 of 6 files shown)


src/Type/Accessory/NonEmptyArrayType.php
```diff
@@ -76,6 +76,12 @@ public function getConstantStrings(): array
 
 	public function accepts(Type $type, bool $strictTypes): AcceptsResult
 	{
+		if ($type instanceof HasOffsetType
+			|| $type instanceof HasOffsetValueType
+		) {
+			return AcceptsResult::createYes();
+		}
```
staabm: I am not 100% sure on this fix, as the 2 types do not necessarily mean "offset on array".

did not find a better place to implement it though.

staabm:
I see we have more todo here.

a union of 2 non-empty-arrays with has-offset accessory atm is just `array`, could be at least `non-empty-array` 

https://phpstan.org/r/cfad9fa9-2279-4f97-b9df-580aab7aa844

**edit:** fixed in https://github.com/phpstan/phpstan-src/pull/3937

src/Type/Accessory/NonEmptyArrayType.php
ondrejmirtes: I've been sitting on this one because I don't understand what's happening in PHPStan so that this is necessary.

You know how I feel about `instanceof *Type` 😊

It might be nicer to fix this in (HasOffsetType|HasOffsetValueType)::isAcceptedBy, because technically these types are accepted by non-empty-array.

ondrejmirtes (changes requested):

src/Type/Accessory/NonEmptyArrayType.php
staabm: ahh great idea. moving the logic into the HasOffset* types also resolves my mentioned problem, as I can check for array and non-empty-ness

src/Type/Accessory/HasOffsetValueType.php
```diff
@@ -122,6 +122,10 @@ public function isSubTypeOf(Type $otherType): IsSuperTypeOfResult
 
 	public function isAcceptedBy(Type $acceptingType, bool $strictTypes): AcceptsResult
 	{
+		if ($acceptingType->isArray()->yes() && $acceptingType->isIterableAtLeastOnce()->yes()) {
+			return AcceptsResult::createYes();
```
ondrejmirtes: It should check the iterable value type against `$this->valueType`.

ondrejmirtes (changes requested):

src/Type/Accessory/HasOffsetValueType.php
staabm: I can't check the iteratable-value-type and the non-emptiness in a single IF because `$acceptingType` is invoked once for each element of the intersection.

this means I get the `ArrayType` which I could check against the value-type, but not against non-emptiness
and on the 2nd call I get the `NonEmptyArray` type, which I can check only against the non-emptiness, but not the value type

staabm: I will think more about it.. maybe we need a fix directly on `IntersectionType` then

staabm:
ohh lol, I did not realize that my [recently merged PR](https://github.com/phpstan/phpstan-src/pull/3937) fixed the issue

staabm:
lets use this PR to just land the regression test. [its fixed](https://github.com/phpstan/phpstan-src/pull/3937)

ondrejmirtes:
Thank you.

ondrejmirtes:
BTW it'd still be beneficial to write some actual unit tests for the types in question, maybe we realize the fix is still needed for some situations where the types don't even go through `TypeCombinator` first.

PR #3926, opened 2025-04-08
Inline array_map(fn, $arr) and use foreach-loop instead

staabm:
counterpart of https://github.com/Roave/BetterReflection/pull/1502

[PHPStan callsite logs](https://gist.github.com/staabm/284e5996677f5eda0bbd1f3d1f002a93#file-sorted-log)


ondrejmirtes:
Please provide some real-world benchmark what effect this has. I'm not willing to compromise on how I want the code to look to gain zero point nothing seconds in performance 😊

staabm:
yes absolutely. I am still exploring and use the PR to share my work with nielsdos so we can investigate until something meaningful manifests

staabm:
Will be optimized in 
- https://github.com/php/php-src/pull/18273
- https://github.com/php/php-src/pull/18326

PR #3927, opened 2025-04-08
Update \SNMP class signatures

miken32:
Per the [documentation](https://www.php.net/manual/en/snmp.get.php) for `SNMP::get()`:
>If `objectId` is a string, then `SNMP::get()` will return SNMP object as string. If `objectId` is a array, all requested SNMP objects will be returned as associative array of the SNMP object ids and their values. 

I updated the return types for `SNMP::get()` and `SNMP::getnext()` to reflect this. I also updated the arguments to use their canonical (camel case) names and corrected a mistake in the ordering of two arguments to `SNMP::walk()`.

One change I wasn't sure about: can arguments can accept complex expressions like return types do? The documentation for `SNMP::set()` says:
>If `objectId` is string, both `type` and `value` must be string too. If `objectId` is array, value must be equal-sized array containing corresponding values, `type` may be either string (it's value will be used for all `objectId`-`value` pairs) or equal-sized array with per-OID value.

So I adjusted the argument types to reflect this logic but am not sure if this is ok or not.


resources/functionMap.php
```diff
@@ -10547,13 +10547,13 @@
 'snmp3_walk' => ['array|false', 'host'=>'string', 'sec_name'=>'string', 'sec_level'=>'string', 'auth_protocol'=>'string', 'auth_passphrase'=>'string', 'priv_protocol'=>'string', 'priv_passphrase'=>'string', 'object_id'=>'string', 'timeout='=>'int', 'retries='=>'int'],
 'SNMP::__construct' => ['void', 'version'=>'int', 'hostname'=>'string', 'community'=>'string', 'timeout='=>'int', 'retries='=>'int'],
 'SNMP::close' => ['bool'],
-'SNMP::get' => ['array|string|false', 'object_id'=>'string|array', 'preserve_keys='=>'bool'],
+'SNMP::get' => ['($objectId is array ? array<string,string> : string)|false', 'objectId'=>'string|string[]', 'preserveKeys='=>'bool'],
```
ondrejmirtes: I'm not sure conditional type will work here in functionMap. Please add a test in `tests/PHPStan/Analyser/nsrt` to verify that.

ondrejmirtes (changes requested):

resources/functionMap.php
miken32: Correct, I should not have assumed it would. I updated to use the multiple signature syntax, I also provided more specificity for string parameters and updated the function-based calls as well.

ondrejmirtes:
Thank you!

PR #3928, opened 2025-04-09
Remember value of arguments passed to `{min,max}()`

cs278:
Fixes phpstan/phpstan#12731

Hopefully this is a suitable fix as I started playing around with other ways of trying to achieve this when I found `AlwaysRememberedExpr` which makes this really simple!


staabm:
searching the phpstan-src codebase with `$scope->getType(new` makes me feel we could have the same problem in a few more places.

would make sense to look into them, in case ondrej is happy with the fix (I have no idea whether/when `AlwaysRememberedExpr` should be used)

ondrejmirtes:
Thank you!

PR #3452, opened 2024-09-18
Allow toggling `treatPhpDocTypesAsCertain` tip

ruudk:
Fixes phpstan/phpstan#11689

PR #3929, opened 2025-04-10
Allow togging `discoveringSymbols` tip

herndlm:
Refs: https://github.com/phpstan/phpstan/discussions/12859

very much analogue to https://github.com/phpstan/phpstan-src/pull/3452

small docs adaption in https://github.com/phpstan/phpstan/pull/12862


ondrejmirtes:
Thank you.

PR #3930, opened 2025-04-11
Remember narrowed types from the constructor when analysing other methods

staabm:
closes https://github.com/phpstan/phpstan/issues/12860
closes https://github.com/phpstan/phpstan/issues/10048
closes https://github.com/phpstan/phpstan/issues/11828
closes https://github.com/phpstan/phpstan/issues/9075
closes https://github.com/phpstan/phpstan/issues/6063
closes https://github.com/phpstan/phpstan/issues/12723

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -5,4 +5,3 @@ parameters:
 		skipCheckGenericClasses!: []
 		stricterFunctionMap: true
 		reportPreciseLineForUnusedFunctionParameter: true
-		narrowMethodScopeFromConstructor: true
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -5,3 +5,4 @@ parameters:
 		skipCheckGenericClasses!: []
 		stricterFunctionMap: true
 		reportPreciseLineForUnusedFunctionParameter: true
+		narrowMethodScopeFromConstructor: true
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -306,7 +306,7 @@ final class MutatingScope implements Scope
 			if ($expr instanceof FuncCall) {
 				if (
 					!$expr->name instanceof Name
-					|| !in_array($expr->name->name, ['class_exists'], true)
+					|| !in_array($expr->name->name, ['class_exists', 'function_exists'], true)
 				) {
 					continue;
 				}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -294,41 +294,57 @@ final class MutatingScope implements Scope
 		);
 	}
 
-	public function rememberConstructorScope(): self
+	/**
+	 * @param array<string, ExpressionTypeHolder> $currentExpressionTypes
+	 * @return array<string, ExpressionTypeHolder>
+	 */
+	private function rememberConstructorExpressions(array $currentExpressionTypes): array
 	{
 		$expressionTypes = [];
-		foreach ($this->expressionTypes as $exprString => $expressionTypeHolder) {
+		foreach ($currentExpressionTypes as $exprString => $expressionTypeHolder) {
 			$expr = $expressionTypeHolder->getExpr();
-			if (!$expr instanceof ConstFetch) {
+			if ($expr instanceof PropertyFetch) {
+				if (
+					!$expr->name instanceof Node\Identifier
+					|| !$expr->var instanceof Variable
+					|| $expr->var->name !== 'this'
+					|| !$this->phpVersion->supportsReadOnlyProperties()
+				) {
+					continue;
+				}
+
+				$propertyReflection = $this->propertyReflectionFinder->findPropertyReflectionFromNode($expr, $this);
+				if ($propertyReflection === null) {
+					continue;
+				}
+
+				$nativePropertyReflection = $propertyReflection->getNativeReflection();
+				if ($nativePropertyReflection === null || !$nativePropertyReflection->isReadOnly()) {
+						continue;
+				}
+			} elseif (!$expr instanceof ConstFetch) {
 				continue;
 			}
+
 			$expressionTypes[$exprString] = $expressionTypeHolder;
 		}
 
-		$nativeExpressionTypes = [];
-		foreach ($this->nativeExpressionTypes as $exprString => $expressionTypeHolder) {
-			$expr = $expressionTypeHolder->getExpr();
-			if (!$expr instanceof ConstFetch) {
-				continue;
-			}
-
-			$nativeExpressionTypes[$exprString] = $expressionTypeHolder;
+		if (array_key_exists('$this', $currentExpressionTypes)) {
+			$expressionTypes['$this'] = $currentExpressionTypes['$this'];
 		}
 
-		if (array_key_exists('$this', $this->expressionTypes)) {
-			$expressionTypes['$this'] = $this->expressionTypes['$this'];
-		}
-		if (array_key_exists('$this', $this->nativeExpressionTypes)) {
-			$nativeExpressionTypes['$this'] = $this->nativeExpressionTypes['$this'];
-		}
+		return $expressionTypes;
+	}
 
+	public function rememberConstructorScope(): self
+	{
 		return $this->scopeFactory->create(
 			$this->context,
 			$this->isDeclareStrictTypes(),
 			$this->getFunction(),
 			$this->getNamespace(),
-			$expressionTypes,
-			$nativeExpressionTypes,
+			$this->rememberConstructorExpressions($this->expressionTypes),
+			$this->rememberConstructorExpressions($this->nativeExpressionTypes),
 			$this->conditionalExpressions,
 			$this->inClosureBindScopeClasses,
 			$this->anonymousFunctionReflection,
@@ -3334,7 +3350,7 @@ final class MutatingScope implements Scope
 
 	private function enterFunctionLike(
 		PhpFunctionFromParserNodeReflection $functionReflection,
-		bool $preserveThis,
+		bool                                $preserveConstructorScope,
 	): self
 	{
 		$parametersByName = [];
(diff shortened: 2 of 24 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -294,6 +294,54 @@ public function enterDeclareStrictTypes(): self
...
+		}
+
+		$nativeExpressionTypes = [];
+		foreach ($this->nativeExpressionTypes as $exprString => $expressionTypeHolder) {
+			$expr = $expressionTypeHolder->getExpr();
+			if (!$expr instanceof ConstFetch) {
+				continue;
+			}
+
+			$nativeExpressionTypes[$exprString] = $expressionTypeHolder;
```
ondrejmirtes: I'd say you should go through these expressions and exclude everything that references variables (using NodeFinder), except for `$this`. That way you'd be able to narrow property types as well.

ondrejmirtes: Also: this isn't the only required change in MutatingScope. You need to make sure these things are re-used in enterClassMethod.

src/Analyser/NodeScopeResolver.php
```diff
@@ -925,7 +930,17 @@ private function processStmtNode(
...
 
-			$this->processStmtNodes($stmt, $stmt->stmts, $classScope, $classStatementsGatherer, $context);
+			// analyze static methods first, constructor next and instance methods last so we can carry over the scope
+			$classLikeStatements = $stmt->stmts;
+			usort($classLikeStatements, static function ($a, $b) {
+				if (!$a instanceof Node\Stmt\ClassMethod || !$b instanceof Node\Stmt\ClassMethod) {
+					return 0;
+				}
+
+				return [!$a->isStatic(), $a->name->toLowerString() !== '__construct'] <=> [!$b->isStatic(), $b->name->toLowerString() !== '__construct'];
```
ondrejmirtes: We have a better way to ask for constructor, that works for same-named constructors as class on PHP 7.

ondrejmirtes (changes requested):

staabm:
re-ordering the order in which methods get analyzed affected the ordering of errors in tests... I am not yet sure where/how to compensate that

ondrejmirtes:
Here's this line: https://github.com/phpstan/phpstan-src/blob/375f68eb71490589ec0b1fa810ac190761ef7ba3/src/Testing/RuleTestCase.php#L244

You could call `getErrors` instead of `getUnorderedErrors`. Not sure what the side effects might be, hopefully it's not brutal.

src/Analyser/MutatingScope.php
staabm: > That way you'd be able to narrow property types as well.

I can't think of a scenario in which this would work. when a property is narrowed in `__construct` it could easily be override by a different instance method later on (except for readonly properties maybe).

can you give me an idea with an example ?

ondrejmirtes: Oh right, it'd work only for readonly properties.

staabm: do you think it should still be implemented?

or is it not worth the effort because in the readonly property case its likely combined with property promotion and this is likely already narrowed with a `@param` type

ondrejmirtes: I remember some cases where it was also used for `class_exists`, to remove false positives from the rest of the class. So I'd say this might be useful and it's not as much effort.

Oh: we need to use the original scope (not the narrowed one) for static methods.

ondrejmirtes: Also the narrowed scope is useful for property hook bodies so please test that (on 8.4+).

src/Analyser/NodeScopeResolver.php
staabm: `ExtendedMethodReflection` has no `isConstructor` method.

I took inspiration from https://github.com/phpstan/phpstan-src/blob/375f68eb71490589ec0b1fa810ac190761ef7ba3/src/Analyser/NodeScopeResolver.php#L654 which seems to be also wrong then?

src/Analyser/MutatingScope.php
staabm: > Oh: we need to use the original scope (not the narrowed one) for static methods.

yeah, thats the reason why I sorted static methods to be analyzed first.. that way they work with a clean scope.. then I narrow the scope via constructor and use this scope for instance methods then

staabm:
I have put it behind bleeding edge and added corresponding opt-in methods for the test-base-classes so we can move forward without the need to adjust all tests

staabm:
sorry for push bombing :-) - I am finished now

conf/bleedingEdge.neon
```diff
@@ -5,3 +5,4 @@ parameters:
 		skipCheckGenericClasses!: []
 		stricterFunctionMap: true
 		reportPreciseLineForUnusedFunctionParameter: true
+		narrowMethodScopeFromConstructor: true
```
ondrejmirtes: Why bleeding edge? I understand if we just want to try it on a small user base first, but this is kind of just making analysis smarter, so we'll remove the flag at some point during 2.x.

ondrejmirtes: I think we can enable this for everyone.

If you're just doing this for the tests, I think we could enable it for everyone by passing `true` to NodeScopeResolver in config.neon.

The way this is now just enabled for TypeInferenceTestCase but not RuleTestCase is fine 👍

src/Analyser/NodeScopeResolver.php
```diff
@@ -791,6 +793,10 @@ private function processStmtNode(
 					$classReflection,
 					$methodReflection,
 				), $methodScope);
+
+				if ($isConstructor && $this->narrowMethodScopeFromConstructor) {
+					$scope = $statementResult->getScope()->rememberConstructorScope();
```
ondrejmirtes: I don't think `$statementResult->getScope()` contains what we want here. Consider branches that always end in `throw` - the object isn't actually created for that situation.

Please add a test for this scenario and check what happens. I think what we need is something like this: https://github.com/phpstan/phpstan-src/blob/375f68eb71490589ec0b1fa810ac190761ef7ba3/src/Rules/TooWideTypehints/TooWideParameterOutTypeCheck.php#L34-L51

Getting the types from execution ends and return statements is something we need I think here: https://github.com/phpstan/phpstan-src/blob/375f68eb71490589ec0b1fa810ac190761ef7ba3/src/Rules/TooWideTypehints/TooWideParameterOutTypeCheck.php#L34-L51

ondrejmirtes (changes requested):

ondrejmirtes:
I just realized there's one more thing we could use this for: with natively typed properties (non-nullable), expressions like `isset` or `??` do not report errors. Because these properties might be uninitialized and `isset` or `??` should report an error.

We should preserve PropertyInitializationExpr and then check for that in IssetCheck.

This is for non-readonly properties too.

staabm:
> We should preserve PropertyInitializationExpr and then check for that in IssetCheck.

great idea. done

src/Rules/IssetCheck.php
```diff
@@ -143,6 +145,23 @@ public function check(Expr $expr, Scope $scope, string $operatorDescription, str
 			}
 
 			if ($propertyReflection->hasNativeType() && !$propertyReflection->isVirtual()->yes()) {
+				if (
+					$expr instanceof Node\Expr\PropertyFetch
+					&& $expr->name instanceof Node\Identifier
+					&& $expr->var instanceof Expr\Variable
+					&& $expr->var->name === 'this'
+					&& !TypeCombinator::containsNull($propertyReflection->getNativeType())
```
ondrejmirtes: 1) Doing `Type::isNull()` would make more sense. This example doesn't cover `mixed` (which can include null).
2) But I think we should pay more attention to `$typeMessageCallback` passed to IssetCheck. For both the type check and the error message. Something a little bit different happens in EmptyRule which should be covered by a set of tests. (Because `empty($x)` is `!isset($x) || $x == false`).

src/Rules/IssetCheck.php
```diff
@@ -143,6 +145,23 @@ public function check(Expr $expr, Scope $scope, string $operatorDescription, str
...
+					&& $scope->hasExpressionType(new PropertyInitializationExpr($propertyReflection->getName()))->yes()
+				) {
+					return RuleErrorBuilder::message(
+						sprintf(
+							'%s cannot be null or uninitialized %s.',
+							$this->propertyDescriptor->describeProperty($propertyReflection, $scope, $expr),
+							$operatorDescription,
+						),
+					)->identifier(sprintf('%s.neverNullOrUninitialized', $identifier))->build();
```
ondrejmirtes: This identifier should include `property`

ondrejmirtes (changes requested):
This is getting really close to merging 👍

staabm:
the 2 failling doctrine tests look valid
- https://github.com/doctrine/dbal/blob/19a2b7deb5fe8c2df0ff817ecea305e50acb62ec/src/Driver/PgSQL/Statement.php#L42
- https://github.com/doctrine/dbal/blob/19a2b7deb5fe8c2df0ff817ecea305e50acb62ec/src/Driver/PgSQL/Connection.php#L34

ondrejmirtes:
Perfect, thank you!

src/Analyser/NodeScopeResolver.php
staabm: I looked into adding `ExtendedMethodReflection->isConstructor(): TrinaryLogic`, but this is incompatible with 
[`PhpMethodFromParserNodeReflection->isConstructor(): bool`
](https://github.com/phpstan/phpstan-src/blob/da74773df7198bb61a69b65af26dd68186b1f91b/src/Reflection/Php/PhpMethodFromParserNodeReflection.php#L279) 

any preferences how you want this to be resolved?

ondrejmirtes: I think we can leave as it is. This is mostly useful for readonly properties anyway, which means `__construct()` constructors...

herndlm:
Hi, this looks pretty neat, thx! I was just wondering why e.g. https://phpstan.org/r/9c85e54d-6c87-42ee-84d5-6db7447c3e4e does not work in the same way? Is it possible to remember such expressions too? UPDATE: changed playground snippet, there was a bug..

Sorry, haven't thought too much about how this works, is it related to the fact that other methods might write / change the property of the wrapper object anytime in between the constructor and method call?

staabm:
we should be able to remember anything which cannot change after constructor was executed (e.g. because its readonly or because the runtime does not allow to change it once set by design).

here is the code which decides what will be remembered and what skipped:
https://github.com/phpstan/phpstan-src/pull/3930/files#diff-dc817f2bab8672057a95375a542e68599164f6ab2380e4cccd1b50583e295d85R305-R334

herndlm:
cool, thx! I think my real case I experienced yesterday was not using readonly props, so it wouldn't work.. and this one is somewhat super special. I'll think about it maybe :)

PR #3931, opened 2025-04-11
RuleTestCase: Order errors

staabm:
as suggested in https://github.com/phpstan/phpstan-src/pull/3930#issuecomment-2796294675

just re-ordering error expectations. not other changes otherwise.

### DIFF
--- a/src/Testing/RuleTestCase.php
+++ b/src/Testing/RuleTestCase.php
@@ -241,7 +241,7 @@ abstract class RuleTestCase extends PHPStanTestCase
 		);
 
 		return [
-			$finalizer->finalize($analyserResult, false, true)->getAnalyserResult()->getUnorderedErrors(),
+			$finalizer->finalize($analyserResult, false, true)->getAnalyserResult()->getErrors(),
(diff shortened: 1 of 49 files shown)


ondrejmirtes:
This is going to be painful for everyone 🙁 What about making it optional? We could make a new `protected function shouldOrderErrors(): bool` defaulting to false.

And then after the constructor change, the rules where the error order doesn't make sense, we could opt them into ordering.

And then change it for everyone in 3.0.

staabm:
I think the problem is, that the impl of https://github.com/phpstan/phpstan-src/pull/3930 brings errors in a different order.

no matter whether we sort now, or sort only by flag: it will always affect already existing tests in case we can't "force" the order we had before https://github.com/phpstan/phpstan-src/pull/3930

but it feels like we did not have a "stable sort" order before so its not something we can enforce

staabm:
we could make the "class-like-statements" sorting a feature-flagged only thing though..?
this would turn the whole scope from constructor narrowing a opt-in though

PR #3932, opened 2025-04-11
DeprecationExtensions: allow custom deprecation-marking logic

janedbal:
Might resolve following issue:

- https://github.com/phpstan/phpstan-deprecation-rules/issues/81


conf/config.neon
```diff
@@ -2157,6 +2157,16 @@ services:
 	-
 		class: PHPStan\Reflection\BetterReflection\SourceStubber\ReflectionSourceStubberFactory
 
+	-
+		class: PHPStan\Reflection\Deprecation\DeprecationResolver
+		arguments:
+			propertyDeprecationProviders: tagged(phpstan.propertyDeprecationProvider)
```
ondrejmirtes: `tagged()` is a cool shortcut but I worry it might lead to reference cycles. It's better to have a class that provides access to user extensions and does `$this->container->getServicesByTag()` itself. That breaks those reference cycles.

For example, in the current PR, you can't have an extension that asks for NodeScopeResolver in its constructor.

src/Reflection/Deprecation/ClassConstantDeprecationProvider.php
```diff
@@ -0,0 +1,27 @@
...
+ * services:
+ * 	-
+ *		class: App\PHPStan\MyProvider
+ *		tags:
+ *			- phpstan.classConstantDeprecationProvider
+ * ```
+ *
+ * @api
+ */
+interface ClassConstantDeprecationProvider
```
ondrejmirtes: A note on naming - this has to be an "extension", not a "provider". All interfaces users are supposed to implement are called "Extension".

"Provider" is typically what allows access to those extensions in PHPStan's codebase, and it's just an internal interface to PHPStan, not covered by `@api`.

See: https://github.com/phpstan/phpstan-src/blob/2.1.x/src/DependencyInjection/Type/ParameterClosureTypeExtensionProvider.php as opposed to to https://github.com/phpstan/phpstan-src/blob/2.1.x/src/Type/FunctionParameterClosureTypeExtension.php.

So please, use PhpStorm rename refactoring to change names of those "provider" interfaces to "extension". And rename the service tags as well :)

src/Reflection/Deprecation/Deprecation.php
```diff
@@ -0,0 +1,35 @@
...
+	{
+		return new self();
+	}
+
+	public function getDescription(): ?string
+	{
+		return $this->description;
+	}
+
+	public function withDescription(?string $description): self
```
ondrejmirtes: * This should be static
* Called `createWithDescription`
* $description shouldn't be nullable

src/Reflection/Deprecation/DeprecationResolver.php
```diff
@@ -0,0 +1,107 @@
...
+
+use PHPStan\BetterReflection\Reflection\Adapter\ReflectionClass;
+use PHPStan\BetterReflection\Reflection\Adapter\ReflectionClassConstant;
+use PHPStan\BetterReflection\Reflection\Adapter\ReflectionEnum;
+use PHPStan\BetterReflection\Reflection\Adapter\ReflectionFunction;
+use PHPStan\BetterReflection\Reflection\Adapter\ReflectionMethod;
+use PHPStan\BetterReflection\Reflection\Adapter\ReflectionProperty;
+use PHPStan\BetterReflection\Reflection\ReflectionConstant;
+
+final class DeprecationResolver
```
ondrejmirtes: This could be called a provider and accept `Container $container` in the constructor instead of these arrays 😊

ondrejmirtes (changes requested):

ondrejmirtes (changes requested):
`DeprecationProvider $deprecationResolver` is in a bunch of places. So is it provider or resolver? 😊 Please sync the type and the parameter name.

src/Reflection/Deprecation/DeprecationProvider.php
```diff
@@ -0,0 +1,119 @@
...
+	private array $functionDeprecationExtensions;
+
+	/** @var array<ConstantDeprecationExtension> $constantDeprecationExtensions */
+	private array $constantDeprecationExtensions;
+
+	public function __construct(
+		Container $container,
+	)
+	{
+		$this->propertyDeprecationExtensions = $container->getServicesByTag('phpstan.propertyDeprecationExtension');
```
ondrejmirtes: These should be done lazily when asked for them. Otherwise this doesn't really prevent the circular reference problem. Only when these are fetched after the constructor it breaks the circular reference cycle.

ondrejmirtes: Also: These tags should be in constants somewhere. Usually there's `public const EXTENSION_TAG` in the related interface.

src/Reflection/Deprecation/DeprecationProvider.php
```diff
@@ -0,0 +1,119 @@
...
+			$deprecation = $extension->getClassConstantDeprecation($reflectionConstant);
+			if ($deprecation !== null) {
+				return $deprecation;
+			}
+		}
+
+		return null;
+	}
+
+	public function isClassDeprecated(ReflectionClass|ReflectionEnum $reflection): ?Deprecation
```
ondrejmirtes: This name doesn't follow the pattern of other names. Please sync it.

ondrejmirtes (changes requested):

src/Reflection/Deprecation/DeprecationProvider.php
janedbal: Ok, do you also want those to be part of ConditionalTagsExtension?

ondrejmirtes: Yeah, absolutely!

ondrejmirtes:
Thank you.

PR #3933, opened 2025-04-13
Faster processing of array comparisons with constant offsets

staabm:
step debugging thru the example made me realize that most of the time we were adding the same constant types to the scope again. This PR detects this situation and does a fast exit to prevent unnecessary scope creations.

grafik

---

before this PR
```
time php bin/phpstan analyze bug-12800.php --debug
  13.40s user 0.18s system 99% cpu 13.637 total
```

after this PR
```
time php bin/phpstan analyze bug-12800.php --debug
  2.84s user 0.13s system 99% cpu 2.974 total
```

grafik

closes https://github.com/phpstan/phpstan/issues/12800

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4428,6 +4428,10 @@ final class MutatingScope implements Scope
 
 		if ($originalExprType->equals($nativeType)) {
 			$newType = TypeCombinator::intersect($type, $originalExprType);
+			if ($newType->isConstantScalarValue()->yes() && $newType->equals($nativeType)) {
+				// don't add the same type over and over again to improve performance
+				return $this;
+			}
 			return $this->specifyExpressionType($expr, $newType, $newType, TrinaryLogic::createYes());
 		}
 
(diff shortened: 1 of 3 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -4428,6 +4428,10 @@ public function addTypeToExpression(Expr $expr, Type $type): self
 
 		if ($originalExprType->equals($nativeType)) {
 			$newType = TypeCombinator::intersect($type, $originalExprType);
+			if ($newType->isConstantScalarValue()->yes() && $newType->equals($originalExprType)) {
```
ondrejmirtes, line 4431: I don't get it, why only for `$newType->isConstantScalarValue()->yes()`?

And: can we make it faster for the other `return` in this method too?

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: > I don't get it, why only for `$newType->isConstantScalarValue()->yes()`?

I went with constant scalars as these matched best the slow case at hand

additionally there are types with weird equals-semantics like ...
- `$mixed->equals($error)` => `true` (because `ErrorType extends MixedType`) 
- ObjectType

... for which tests start failing if we take the fast-pass without changing the scope

> And: can we make it faster for the other return in this method too?

I can't think of a case where this would help

ondrejmirtes: These `equals` problems should be fixed, they might lead to other bugfixes as well.

ondrejmirtes:
Thank you!

PR #3934, opened 2025-04-13
Fix MixedType->equals(ErrorType)

staabm:
refs https://github.com/phpstan/phpstan-src/pull/3933#discussion_r2041162189

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4488,7 +4488,7 @@ final class MutatingScope implements Scope
 			$newType = TypeCombinator::intersect($type, $originalExprType);
 			if ($newType->isObject()->no() && $newType->equals($originalExprType)) {
 				// don't add the same type over and over again to improve performance.
-				// objects can get narrowed even though ObjectType->equal() will return true (e.g. via implicit "final" via new())
+				// objects can get narrowed even though ObjectType->equal() will return true (e.g. via implicit "final" via new Foo())
 				return $this;
 			}
 			return $this->specifyExpressionType($expr, $newType, $newType, TrinaryLogic::createYes());
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4486,7 +4486,7 @@ final class MutatingScope implements Scope
 
 		if ($originalExprType->equals($nativeType)) {
 			$newType = TypeCombinator::intersect($type, $originalExprType);
-			if ($newType->isObject()->no() && $newType->equals($originalExprType)) {
+			if (!$newType->isObject()->yes() && $newType->equals($originalExprType)) {
 				// don't add the same type over and over again to improve performance
 				return $this;
 			}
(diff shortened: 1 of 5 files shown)


staabm:
~~I was not sure whether `TemplateMixedType` should be considered equal to `MixedType` or not~~

src/Type/MixedType.php
```diff
@@ -314,6 +314,10 @@ public function equals(Type $type): bool
 			return false;
 		}
 
+		if ($type instanceof ErrorType) {
```
ondrejmirtes: MixedType isn't equal to TemplateMixedType.

A bunch of types use `!$type instanceof static`, maybe we should do it in more cases.

ondrejmirtes (changes requested):

src/Type/MixedType.php
staabm: changing `MixedType->equals()` to return false for `TemplateMixedType` makes [this test fail](https://github.com/phpstan/phpstan-src/blob/cd2ca6483689a9805b8706c22897ea5413729b6d/tests/PHPStan/Generics/TemplateTypeFactoryTest.php#L50-L58).

```
1) PHPStan\Generics\TemplateTypeFactoryTest::testCreate with data set #6 (PHPStan\Type\Generic\TemplateMixedType Object (...), PHPStan\Type\MixedType Object (...))
mixed -> equals(U (function a(), parameter))
```

I am not into generics enough to oversee how this should work

herndlm:
👍 I suppose this improves the weird cases I had before too, where error is accepted as a mixed or at least something like that 😊

src/Type/MixedType.php
ondrejmirtes: Yeah, those tests are wrong. Type is only equal to another if the another type is exactly the same.

staabm: ok, hopefully the new fix is what you had in mind

src/Analyser/MutatingScope.php
```diff
@@ -4489,7 +4489,7 @@ public function addTypeToExpression(Expr $expr, Type $type): self
 
 		if ($originalExprType->equals($nativeType)) {
 			$newType = TypeCombinator::intersect($type, $originalExprType);
-			if ($newType->isConstantScalarValue()->yes() && $newType->equals($originalExprType)) {
+			if ($newType->isObject()->no() && $newType->equals($originalExprType)) {
```
staabm: Will look into whether we can get rid of the isObject() check here today

staabm: ok, I looked into it. I think we cannot drop the `$newType->isObject()->no()` check here, because Objects will be considered equal by `ObjectType->equals()` even though they have slight differences which are important when narrowing, e.g.

`new Foo()` in local-scope vs. `Foo` typed variable (implicit Final)

staabm: most conservative move would be to leave it `$newType->isConstantScalarValue()->yes()` until we get a slow case which benefits from the change and only fix `MixedType->equals(ErrorType)` for now

staabm:
just reduced it to the actual bugfix.

ondrejmirtes:
Thank you.

PR #3937, opened 2025-04-15
TypeCombinator returns `non-empty-array` for union of `isIterableAtLeastOnce()->yes()`

staabm:
a array can be non-empty caused by different accessory types, namely `NonEmptyArrayType`, `HasOffsetType`, `HasOffsetValueType`.

before this PR we only returned a non-empty-array result from `TypeCombinator::union()` when all elements in the union were using the same accessory types, e.g. 
- `union(array&non-empty, array&non-empty)`
- `union(array&hasOffset('a'), array&hasOffset('a'))`

after this PR we in addition return a non-empty-array for unions of arrays which are non-empty caused by different accessories, e.g. 
- `union(array&non-empty, array&hasOffset('a'))`
- `union(array&hasOffset('b'), array&hasOffset('a'))`
- `union(array&hasOffset('b'), array&hasOffsetValue(4, 'a'))`

refs https://github.com/phpstan/phpstan-src/pull/3924#issuecomment-2802237890


src/Type/TypeCombinator.php
```diff
@@ -703,6 +707,10 @@ private static function processArrayAccessoryTypes(array $arrayTypes): array
 			$commonAccessoryTypes[] = $accessoryType[0];
 		}
 
+		if (TrinaryLogic::createYes()->and(...$isIterableAtLeastOnce)->yes()) {
+			$commonAccessoryTypes[] = new NonEmptyArrayType();
```
staabm, line 711: atm this can lead to some arrays, which - while processing - have the `NonEmptyArrayType` 2-times.
when `TypeCombinator::union()` finally ends, this "duplication" gets normalized away.

we could check at this IF-condition here, whether `$commonAccessoryTypes` already contains `NonEmptyArrayType` so we don't add it a 2nd time. 
for readability I have chosen not todo this additionaly check. but I am fine with adding it, in case you think we should do it.

ondrejmirtes:
I'd say this isn't necessary. Intersection type of ArrayType and HasOffsetValueType should already know it's iterable at least once (it's non-empty).

The question is - why the tests you changed already don't result in `non-empty-array` being expected...

staabm:
> The question is - why the tests you changed already don't result in `non-empty-array` being expected...

before this PR the `TypeCombinator` only takes accessory types into the final type, when they are contained in every element of the union (except `OversizedArrayType`)

see https://github.com/phpstan/phpstan-src/blob/5df46172237c9966b51ada815a31404fa5729878/src/Type/TypeCombinator.php#L687-L704

therefore in e.g. `TypeCombinator::union(array&hasOffset('b'), array&hasOffsetValue(4, 'a'))` the resulting Type is just `array` (it is loosing all accessories, and therefore it no longer is e.g. non-empty)

ondrejmirtes:
Ohhh, I get it, this makes sense then!

ondrejmirtes:
Thank you!

PR #3938, opened 2025-04-15
Remove useless assign $parentNode = $parentNode

samsonasik:


samsonasik:
Ready to merge 👍

ondrejmirtes:
It's there so that it's obvious it's for the `/** @var Node\Stmt\Function_|Node\Stmt\ClassMethod|PropertyHookStatementNode|Expr\Closure $parentNode */`.

I'd only find it beneficial if we were able to get rid of the `@var` completely as well.

samsonasik:
it seems `@var` doesn't require assign for definition, ref https://phpstan.org/r/c3ba8ad1-4c11-4615-884a-6f960ee594ec

samsonasik:
@ondrejmirtes I removed the useless `@var` definition too 👍 https://github.com/phpstan/phpstan-src/pull/3938/commits/18e1d27660b0f90260db10fc183f9199470e4812

and it seems working ok

samsonasik:
Compile phar notice seems unrelated, which e2e tried load very old rector 0.*

ondrejmirtes:
Thank you!

PR #3940, opened 2025-04-15
Fix specifying types on nullsafe true/false comparison

rvanvelzen:
Fix phpstan/phpstan#12866


ondrejmirtes:
Thank you!

nikophil:
Thank you @rvanvelzen!

PR #3941, opened 2025-04-16
ResultCache: configurable days causing cache skip

janedbal:
Recently, I implemented a script that allows result cache reuse from CI job. Sometimes, I get back to older MR where I want to continue, but this hardcoded limit slows me down as I cannot reuse the cache anymore for such MR. 

Note: our full PHPStan analysis takes ~10 minutes on 12th Gen Intel Core i7-1255U.


ondrejmirtes:
Thank you.

PR #3942, opened 2025-04-16
Consider in_array comparison as strict when type is the same

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12884


ondrejmirtes:
Thank you.

PR #3944, opened 2025-04-18
`Scope->rememberConstructorScope` should not remember the function scope

### LINKED ISSUE #323: PHPStan 2.1.12 compatibility
PHPStan just released 2.1.12 and I checked out the Dependabot bump going from 2.1.11 to 2.1.12 and noticed that all my rules that disallow classes except when `allowInMethodsWithAttributes` now report issues.

```php
<?php

declare(strict_types=1);

return [
    'parameters' => [
        'disallowedClasses' => [
            [
                'allowInUse' => true,
                'class' => EventRecorder::class,
                'allowInMethodsWithAttributes' => [
                    AsCommandHandler::class,
                ],
                'allowIn' => [
                    '**/tests/*',
                ],
            ],
        ],
    ],
];
```

I'll debug it further, but just wanted to hint it here already.

@spaze Maybe you can kick off a new CI builds that pulls in the latest PHPStan version to confirm?

staabm:
When remembering the scope we should not remember the scope-function, as otherwise analyzing instance-method parameters would see the remembered`$scope->getFunction()` as `__construct` instead of `null`.

we only want to remember the types

closes https://github.com/spaze/phpstan-disallowed-calls/issues/323

### DIFF
--- a/tests/PHPStan/Analyser/InstanceMethodsParameterScopeFunctionRule.php
+++ b/tests/PHPStan/Analyser/InstanceMethodsParameterScopeFunctionRule.php
@@ -12,7 +12,7 @@ use function sprintf;
 /**
  * @implements Rule<FullyQualified>
  */
-class InstanceMethodsParameterRule implements Rule
+class InstanceMethodsParameterScopeFunctionRule implements Rule
 {
 
 	public function getNodeType(): string
--- a/tests/PHPStan/Analyser/InstanceMethodsParameterScopeFunctionRule.php
+++ b/tests/PHPStan/Analyser/InstanceMethodsParameterScopeFunctionRule.php
@@ -23,11 +23,11 @@ class InstanceMethodsParameterScopeFunctionRule implements Rule
 	public function processNode(Node $node, Scope $scope): array
 	{
 		if ($scope->getFunction() !== null) {
-			throw new ShouldNotHappenException('All names in the tests should not have a scope function.');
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/InstanceMethodsParameterScopeFunctionRule.php
```diff
@@ -0,0 +1,34 @@
...
+
+	public function getNodeType(): string
+	{
+		return FullyQualified::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if ($scope->getFunction() !== null) {
+			throw new ShouldNotHappenException('All names in the tests should not have a scope function.');
```
ondrejmirtes: "function scope", not "scope function". Function is the adjective here 😊

tests/PHPStan/Analyser/InstanceMethodsParameterScopeFunctionRule.php
```diff
@@ -0,0 +1,34 @@
...
+		}
+
+		return [
+			RuleErrorBuilder::message(sprintf('Name %s found in scope-function null', $node->toString()))->identifier('test.instanceOfMethodsParameterRule')->build(),
```
ondrejmirtes: function scope

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #3945, opened 2025-04-18
Fix imprecise types after assignment when strict-types=1

staabm:
work with coecered type only when the assigned one is incompatible with the natively defined property type

closes https://github.com/phpstan/phpstan/issues/12902

supercedes https://github.com/phpstan/phpstan-src/pull/3943

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5626,9 +5626,20 @@ final class NodeScopeResolver
 				$nodeCallback(new PropertyAssignNode($var, $assignedExpr, $isAssignOp), $scope);
 				if ($propertyReflection->canChangeTypeAfterAssignment()) {
 					if ($propertyReflection->hasNativeType() && $scope->isDeclareStrictTypes()) {
+						$assignedNativeType = $scope->getNativeType($assignedExpr);
 						$propertyNativeType = $propertyReflection->getNativeType();
 
-						$scope = $scope->assignExpression($var, TypeCombinator::intersect($assignedExprType->toCoercedArgumentType(true), $propertyNativeType), TypeCombinator::intersect($scope->getNativeType($assignedExpr)->toCoercedArgumentType(true), $propertyNativeType));
+						$newAssignedType = TypeCombinator::intersect($assignedExprType, $propertyNativeType);
+						if ($newAssignedType instanceof NeverType) {
+							$newAssignedType = TypeCombinator::intersect($assignedExprType->toCoercedArgumentType(true), $propertyNativeType);
+						}
+
+						$newAssignedNativeType = TypeCombinator::intersect($assignedNativeType, $propertyNativeType);
+						if ($newAssignedNativeType instanceof NeverType) {
+							$newAssignedNativeType = TypeCombinator::intersect($assignedNativeType->toCoercedArgumentType(true), $propertyNativeType);
+						}
+
+						$scope = $scope->assignExpression($var, $newAssignedType, $newAssignedNativeType);
 					} else {
 						$scope = $scope->assignExpression($var, $assignedExprType, $scope->getNativeType($assignedExpr));
 					}
@@ -5697,9 +5708,20 @@ final class NodeScopeResolver
 				$nodeCallback(new PropertyAssignNode($var, $assignedExpr, $isAssignOp), $scope);
 				if ($propertyReflection !== null && $propertyReflection->canChangeTypeAfterAssignment()) {
 					if ($propertyReflection->hasNativeType() && $scope->isDeclareStrictTypes()) {
+						$assignedNativeType = $scope->getNativeType($assignedExpr);
 						$propertyNativeType = $propertyReflection->getNativeType();
 
-						$scope = $scope->assignExpression($var, TypeCombinator::intersect($assignedExprType->toCoercedArgumentType(true), $propertyNativeType), TypeCombinator::intersect($scope->getNativeType($assignedExpr)->toCoercedArgumentType(true), $propertyNativeType));
+						$newAssignedType = TypeCombinator::intersect($assignedExprType, $propertyNativeType);
+						if ($newAssignedType instanceof NeverType) {
+							$newAssignedType = TypeCombinator::intersect($assignedExprType->toCoercedArgumentType(true), $propertyNativeType);
+						}
+
+						$newAssignedNativeType = TypeCombinator::intersect($assignedNativeType, $propertyNativeType);
+						if ($newAssignedNativeType instanceof NeverType) {
+							$newAssignedNativeType = TypeCombinator::intersect($assignedNativeType->toCoercedArgumentType(true), $propertyNativeType);
+						}
+
+						$scope = $scope->assignExpression($var, $newAssignedType, $newAssignedNativeType);
 					} else {
 						$scope = $scope->assignExpression($var, $assignedExprType, $scope->getNativeType($assignedExpr));
 					}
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Analyser/nsrt/bug-12902.php
```diff
@@ -0,0 +1,72 @@
...
+	public function __construct()
+	{
+		self::$i = getInt();
+		assertType('float|int', self::$i); // could be int
+		assertNativeType('float|int', self::$i); // could be int
+
+		$this->impureCall();
+		assertType('float|int', self::$i);
+		assertNativeType('float|int', self::$i);
+	}
```
staabm: the `impureCall` does not make the scope forget about already narrowed types of static property fetches, therefore we don't narrow in this scenario with this PR.

-> we only improve the instance-call case, because non-static fetches will be forgotten as the test in `NarrowsNativeUnion` proves

src/Analyser/NodeScopeResolver.php
```diff
@@ -5697,9 +5708,13 @@ static function (): void {
...
 				if ($propertyReflection !== null && $propertyReflection->canChangeTypeAfterAssignment()) {
 					if ($propertyReflection->hasNativeType() && $scope->isDeclareStrictTypes()) {
+						$assignedNativeType = $scope->getNativeType($assignedExpr);
 						$propertyNativeType = $propertyReflection->getNativeType();
 
-						$scope = $scope->assignExpression($var, TypeCombinator::intersect($assignedExprType->toCoercedArgumentType(true), $propertyNativeType), TypeCombinator::intersect($scope->getNativeType($assignedExpr)->toCoercedArgumentType(true), $propertyNativeType));
+						$newAssignedType = TypeCombinator::intersect($assignedExprType->toCoercedArgumentType(true), $propertyNativeType);
+						$newAssignedNativeType = TypeCombinator::intersect($assignedNativeType->toCoercedArgumentType(true), $propertyNativeType);
+
+						$scope = $scope->assignExpression($var, $newAssignedType, $newAssignedNativeType);
```
staabm: no functional change here.

ondrejmirtes:
I've analyzed this situation. It's much more clear what's going on in this code sample: https://phpstan.org/r/903891e3-323a-494b-83a0-27aff806eb88

It's happening because IntegerType::toCoercedArgumentType() returns `TypeCombinator::union($this, $this->toFloat())`. And we take that, intrsection that with `int|float` and assign the value.

But what should happen first is even before we call toCoercedArgumentType, we should check whether the original value (ConstantIntegerType 1) is a subtype of the native property type.

1) IntegerType->isSuperTypeOf(ConstantIntegerType) -> yes
2) FloatType->isSuperTypeOf(ConstantIntegerType) -> no

Because 1) matches, we don't need to call toCoercedArgumentType at all. We simply assign the value as it is.

To get a list of types to compare, we can call TypeUtils::flattenTypes on the native property type. That way we can compare it in a foreach even when there's just a single type like `int`.

What's nice about this is that it can solve the problem for strict_types = 0 as well. The workflow should be:

1) Go through the types, if we find one that matches, use the assigned expression.
2) If we don't succeed with 1), check if we're in strict_types = 1. If yes, use the current logic in NodeScopeResolver with toCoercedArgumentType.

If done correctly, this will remove the need for https://github.com/phpstan/phpstan-src/pull/3943 as well.

src/Analyser/NodeScopeResolver.php
```diff
@@ -5625,10 +5625,27 @@ static function (): void {
...
+						}
+
+						if (!$assignedTypeIsCompatible && $scope->isDeclareStrictTypes()) {
+							$scope = $scope->assignExpression(
+								$var,
+								TypeCombinator::intersect($assignedExprType->toCoercedArgumentType(true), $propertyNativeType),
+								TypeCombinator::intersect($assignedNativeType->toCoercedArgumentType(true), $propertyNativeType),
+							);
+						} else {
+							$scope = $scope->assignExpression($var, $assignedExprType, $assignedNativeType);
```
ondrejmirtes: This is gonna get executed if assigned type is compatible or if we're not in declare strict types. That's not what we want. We only want to execute this if assigned type is compatible.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: so what to execute if not compatible and not strict types?

staabm: I think it works now as you imagined.

ondrejmirtes: If not compatible and not strict types, we currently can't do anything. Once we properly implement toCoercedArgumentType in the whole typesystem, then we can use it same way we use it now with strict_types=1.

tests/PHPStan/Analyser/nsrt/bug-12902-non-strict.php
```diff
@@ -0,0 +1,88 @@
...
+
+	public function __construct()
+	{
+		self::$i = getInt();
+		assertType('int', self::$i);
+		assertNativeType('int', self::$i);
+
+		$this->impureCall();
+		assertType('int', self::$i); // should be float|int
+		assertNativeType('int', self::$i); // should be float|int
```
staabm, line 76: this is a unrelated bug, right?

src/Testing/TypeInferenceTestCase.php
```diff
@@ -398,7 +398,7 @@ private static function isFileLintSkipped(string $file): bool
 
 			@fclose($f);
 
-			if (preg_match('~<?php\\s*\\/\\/\s*lint\s*([^\d\s]+)\s*([^\s]+)\s*~i', $firstLine, $m) === 1) {
+			if (preg_match('~<?php\\s*(?:declare\\s*\([^)]+\)\\s*;\\s*)?\\/\\/\s*lint\s*([^\d\s]+)\s*([^\s]+)\s*~i', $firstLine, $m) === 1) {
```
ondrejmirtes: This isn't compatible with how php-parallel-lint works.

tests/PHPStan/Analyser/nsrt/bug-12902-non-strict.php
```diff
@@ -0,0 +1,88 @@
+<?php declare(strict_types = 0); // lint >= 8.1
```
ondrejmirtes: The first line has to be `<?php // lint >= 8.1` and the declare statement has to be on line 3.

tests/PHPStan/Analyser/nsrt/bug-12902-non-strict.php
ondrejmirtes: Yeah, probably.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/bug-12902-non-strict.php
staabm: thanks!

ondrejmirtes:
Thank you!

tests/PHPStan/Analyser/nsrt/bug-12902-non-strict.php
staabm: fix in https://github.com/phpstan/phpstan-src/pull/3950

staabm:
I am in the process of updating PHPStan to 2.1.13 and stumble now again over the problem, I initially tried to fix with my first iteration of this PR in https://github.com/phpstan/phpstan-src/pull/3943

see https://phpstan.org/r/aa5d5753-eca2-4735-bebf-6479e1ab1f42

do you think we could remove nullability from the property-type in scope, as the method we call returns a native non-nullable type, even if strict-types=0?

staabm:
proposed a new PR so we can discuss whether its possible and my idea is clear: https://github.com/phpstan/phpstan-src/pull/3965

PR #3946, opened 2025-04-18
Handle preserve_keys in `array_slice()` for normal arrays

herndlm:
Closes https://github.com/phpstan/phpstan/issues/12880

https://www.php.net/manual/en/function.array-slice.php

> array_slice() will reorder and reset the integer array indices by default. This behaviour can be changed by setting preserve_keys to [true](https://www.php.net/manual/en/reserved.constants.php#constant.true). String keys are always preserved, regardless of this parameter.

https://3v4l.org/7njtg

### DIFF
--- a/src/Type/ArrayType.php
+++ b/src/Type/ArrayType.php
@@ -442,6 +442,10 @@ class ArrayType implements Type
 
 	public function sliceArray(Type $offsetType, Type $lengthType, TrinaryLogic $preserveKeys): Type
 	{
+		if ($preserveKeys->no() && $this->keyType->isInteger()->yes()) {
+			return TypeCombinator::intersect(new self(new IntegerType(), $this->itemType), new AccessoryArrayListType());
+		}
+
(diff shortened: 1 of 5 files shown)


src/Type/ArrayType.php
```diff
@@ -442,6 +442,10 @@ public function shuffleArray(): Type
 
 	public function sliceArray(Type $offsetType, Type $lengthType, TrinaryLogic $preserveKeys): Type
 	{
+		if ($preserveKeys->no() && $this->keyType->isInteger()->yes()) {
+			return TypeCombinator::intersect(new self(new IntegerType(), $this->itemType), new AccessoryArrayListType());
```
herndlm, line 446: Could be replaced with `return $this->getValuesArray();` if it should be more DRY. I wasn't sure, but this felt a bit less "hidden" to me.

ondrejmirtes:
Thank you!

PR #3947, opened 2025-04-18
Improve return type of `array_splice()`

herndlm:
Since `array_splice` returns an extract, similar as `array_slice`, we can make use of what we built already for the latter and get better return types with minimal effort. 

https://www.php.net/manual/en/function.array-splice.php

https://3v4l.org/ggnQB

### DIFF
--- a/src/Type/Php/ArraySpliceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArraySpliceFunctionReturnTypeExtension.php
@@ -4,13 +4,21 @@ namespace PHPStan\Type\Php;
 
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
+use PHPStan\Php\PhpVersion;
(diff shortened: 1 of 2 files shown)


herndlm:
This also should fix @Seldaek's example mentioned in https://github.com/phpstan/phpstan/issues/11917#issuecomment-2501369714, but the original issue was dealing with the inferred type of the passed array, so we can't close that yet. For that I have an idea (`Type::spliceArray()` similar as we also did with e.g. `array_pop` and `array_shift` ) but it's out of scope of this PR and will have to deal with many edge cases and such.

See https://github.com/phpstan/phpstan-src/actions/runs/14536946202

herndlm:
also: there's a small change that test results change after merging https://github.com/phpstan/phpstan-src/pull/3946. therefore rebasing this one afterwards would be a good idea :)
but there aren't many `array_splice()` tests yet apparently.

ondrejmirtes:
Thank you!

PR #3949, opened 2025-04-18
Improve `ConstantArrayType::sliceArray()` with non constant integer args

herndlm:
`ConstantArrayType::sliceArray()` is one of the more complex constant array methods I once came up with. But it can only handle constant integer offsets and lengths so far. I tried to add support for e.g. unions, but this would increase complexity even more.

Currently there is an issue where it naively falls back the defaults (offset 0, length null = full array) when it encounters invalid values. Instead of that, I think it's better and safer to degrade to a general array and do the slice there to avoid incorrect types.


ondrejmirtes:
Thank you!

PR #3950, opened 2025-04-19
Fix forgetting static property access after impure call

staabm:
closes https://github.com/phpstan/phpstan/issues/11019
closes https://github.com/phpstan/phpstan/issues/3747
closes https://github.com/phpstan/phpstan/issues/8523
closes https://github.com/phpstan/phpstan/issues/4864
fixes one repro of https://github.com/phpstan/phpstan/issues/8926
closes https://github.com/phpstan/phpstan/issues/4443

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4382,7 +4382,10 @@ final class MutatingScope implements Scope
 			if (
 				$exprStringToInvalidate === '$this'
 				&& $node instanceof Name
-				&& in_array($node->toLowerString(), ['self', 'static', 'parent'], true)
+				&& (
+					in_array($node->toLowerString(), ['self', 'static', 'parent'], true)
+					|| $this->getClassReflection()->is($this->resolveName($node))
+				)
 			) {
 				return true;
 			}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4379,6 +4379,14 @@ final class MutatingScope implements Scope
 		$nodeFinder = new NodeFinder();
 		$expressionToInvalidateClass = get_class($exprToInvalidate);
 		$found = $nodeFinder->findFirst([$expr], function (Node $node) use ($expressionToInvalidateClass, $exprStringToInvalidate): bool {
+			if (
+				$exprStringToInvalidate === '$this'
+				&& $node instanceof Name
+				&& in_array($node->name, ['self', 'static', 'parent'], true)
+			) {
+				return true;
+			}
+
 			if (!$node instanceof $expressionToInvalidateClass) {
 				return false;
 			}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4382,7 +4382,7 @@ final class MutatingScope implements Scope
 			if (
 				$exprStringToInvalidate === '$this'
 				&& $node instanceof Name
-				&& in_array($node->name, ['self', 'static', 'parent'], true)
+				&& in_array($node->toLowerString(), ['self', 'static', 'parent'], true)
 			) {
 				return true;
 			}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -4384,7 +4384,7 @@ final class MutatingScope implements Scope
 				&& $node instanceof Name
 				&& (
 					in_array($node->toLowerString(), ['self', 'static', 'parent'], true)
-					|| $this->getClassReflection()->is($this->resolveName($node))
+					|| ($this->getClassReflection() !== null && $this->getClassReflection()->is($this->resolveName($node)))
 				)
 			) {
 				return true;
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -3031,7 +3031,10 @@ final class NodeScopeResolver
 				$methodReflection !== null
 				&& (
 					$methodReflection->hasSideEffects()->yes()
-					|| $methodReflection->getName() === '__construct'
+					|| (
+						!$methodReflection->isStatic()
+						&& $methodReflection->getName() === '__construct'
+					)
 				)
 				&& $scopeFunction instanceof MethodReflection
 				&& $scope->isInClass()
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2598,6 +2598,16 @@ final class NodeScopeResolver
 			}
 
 			if (
+				(
+					($functionReflection !== null && $functionReflection->hasSideEffects()->yes())
+					|| $parametersAcceptor instanceof ClosureType && $parametersAcceptor->getReturnType()->isVoid()->yes()
+				)
+				&& $scope->isInClass()
+			) {
+				$scope = $scope->invalidateExpression(new Variable('this'), true);
+			}
+
+			if (
 				$functionReflection !== null
 				&& in_array($functionReflection->getName(), ['json_encode', 'json_decode'], true)
 			) {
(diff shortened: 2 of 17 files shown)


tests/PHPStan/Rules/Methods/data/bug-8523b.php
```diff
@@ -0,0 +1,24 @@
...
+
+		$callback();
+
+		var_dump(self::$instance);
+
+		self::$instance?->save();
+	}
+}
+
+(new HelloWorld())->save();
```
staabm, line 24: there is a similar case where we still fail:

```php

class HelloWorld
{
	public static ?HelloWorld $instance = null;
	
    public function save(): void {
        HelloWorld::$instance = new HelloWorld();

        $callback = static function(): void {
			HelloWorld::$instance = null;
		};
		
		$callback();
			
		var_dump(HelloWorld::$instance);
		
        HelloWorld::$instance?->save();
    }
}

(new HelloWorld())->save();
```

note: instead of `self` in this case the FQCN is used.

maybe we should normalize FQCN globally in AST to `self` in such cases, so we don't need to handle this case all over the place?

src/Analyser/NodeScopeResolver.php
```diff
@@ -2597,6 +2597,13 @@ static function (): void {
 				$throwPoints[] = ThrowPoint::createImplicit($scope, $expr);
 			}
 
+			if (
+				$parametersAcceptor instanceof ClosureType && $parametersAcceptor->getReturnType()->isVoid()->yes()
```
staabm: to be consistent with `MethodCall`, `StaticCall` we could add

`$functionReflection !== null && $functionReflection->hasSideEffects()->yes()`

here, but this would regress some types:

```
There were 3 failures:

1) PHPStan\Analyser\LegacyNodeScopeResolverTest::testForeachLoopVariables with data set #19 ('array<string, 1|2|3>', '$this->property', ''afterLoop'')
$this->property at 'afterLoop'
--- Expected
+++ Actual
@@ @@
-'array<string, 1|2|3>'
+'array<int>'

/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9696
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9590
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9597
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:7092

2) PHPStan\Analyser\LegacyNodeScopeResolverTest::testForeachLoopVariables with data set #17 ('array<string, 1|2|3>', '$this->property', ''begin'')
$this->property at 'begin'
--- Expected
+++ Actual
@@ @@
-'array<string, 1|2|3>'
+'array<int>'

/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9696
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9590
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9597
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:7092

3) PHPStan\Analyser\LegacyNodeScopeResolverTest::testForeachLoopVariables with data set #18 ('non-empty-array<string, 1|2|3>', '$this->property', ''end'')
$this->property at 'end'
--- Expected
+++ Actual
@@ @@
-'non-empty-array<string, 1|2|3>'
+'non-empty-array<int>'

/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9696
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9590
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:9597
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php:7092

FAILURES!                                                  
Tests: 12743, Assertions: 85340, Failures: 3, Skipped: 139.
```

wdyt?

src/Analyser/MutatingScope.php
```diff
@@ -4379,6 +4379,14 @@ private function shouldInvalidateExpression(string $exprStringToInvalidate, Expr
 		$nodeFinder = new NodeFinder();
 		$expressionToInvalidateClass = get_class($exprToInvalidate);
 		$found = $nodeFinder->findFirst([$expr], function (Node $node) use ($expressionToInvalidateClass, $exprStringToInvalidate): bool {
+			if (
+				$exprStringToInvalidate === '$this'
+				&& $node instanceof Name
+				&& in_array($node->name, ['self', 'static', 'parent'], true)
```
ondrejmirtes: These are case-insensitive, need to use toLowerString

src/Analyser/NodeScopeResolver.php
ondrejmirtes: For closures we have already analysed the body, there's `ClosureType::getInvalidateExpressions()` so we know exactly what to invalidate.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: for the example in tests/PHPStan/Rules/Methods/data/bug-8523b.php `ClosureType::getInvalidateExpressions()` returns an empty array.

when searching for `new InvalidateExprNode` in the whole codebase it looks like only `MethodCall` and `Arg->value` is considered a invalidate-expression. Should I add `PropertyFetch` and `StaticPropertyFetch` and `StaticCall`?

src/Analyser/MutatingScope.php
```diff
@@ -4379,6 +4379,14 @@ private function shouldInvalidateExpression(string $exprStringToInvalidate, Expr
 		$nodeFinder = new NodeFinder();
 		$expressionToInvalidateClass = get_class($exprToInvalidate);
 		$found = $nodeFinder->findFirst([$expr], function (Node $node) use ($expressionToInvalidateClass, $exprStringToInvalidate): bool {
+			if (
+				$exprStringToInvalidate === '$this'
+				&& $node instanceof Name
+				&& in_array($node->toLowerString(), ['self', 'static', 'parent'], true)
```
ondrejmirtes: I just realized the `self` here should be improved. We should use `Scope::resolveName` or `resolveTypeByName` and check that the class name matches the current scope class name. Because `self` is just an alias to the current class name, so whether we're using `self::` or `Foo::` shouldn't matter.

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Oh, I just mixed invalidating expressions with impure points. Please check `ClosureType::getImpurePoints()` to see whether closure is pure or not.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
```diff
@@ -2597,6 +2597,13 @@ static function (): void {
 				$throwPoints[] = ThrowPoint::createImplicit($scope, $expr);
 			}
 
+			if (
+				$parametersAcceptor instanceof ClosureType && count($parametersAcceptor->getImpurePoints()) > 0
+				&& $scope->isInClass()
```
staabm, line 2602: last open question: do we want to invalidate 'this' in this case for regular function calls with side-effects?
we only handle Closure at this point right now

see https://github.com/phpstan/phpstan-src/pull/3950#discussion_r2051457669

staabm: Scratch that. A function is pretty unlikely to affect `$this` or `self`

ondrejmirtes:
Thank you!

PR #3951, opened 2025-04-19
TableErrorFormatter: visually differentiate phpstan assertion errors from rule errors

staabm:
before this PR

grafik

after this PR

grafik

### DIFF
--- a/src/Command/ErrorFormatter/TableErrorFormatter.php
+++ b/src/Command/ErrorFormatter/TableErrorFormatter.php
@@ -13,6 +13,7 @@ use function array_map;
 use function count;
 use function explode;
 use function getenv;
+use function in_array;
 use function is_string;
 use function ltrim;
 use function sprintf;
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Green isn't the right choice when it's an error :)

src/Command/ErrorFormatter/TableErrorFormatter.php
```diff
@@ -117,6 +118,14 @@ public function formatErrors(
 
 					$message .= "\n✏️  <href=" . OutputFormatter::escape($url) . '>' . $title . '</>';
 				}
+
+				if (
+					$error->getIdentifier() !== null
+					&& in_array($error->getIdentifier(), ['phpstan.type', 'phpstan.nativeType', 'phpstan.variable'], true)
```
ondrejmirtes: Also would make sense for `phpstan.dumpType` and `phpstan.unknownExpectation`.

ondrejmirtes (changes requested):

staabm:
> Green isn't the right choice when it's an error :)

hehe.. I am too positive about my errors ;)

updated the impl and the screenshot

ondrejmirtes:
Thank you!

PR #3952, opened 2025-04-19
Add `Type::spliceArray()`, improve `splice_array()` array type narrowing

herndlm:
~**Needs https://github.com/phpstan/phpstan-src/pull/3959 first**~ DONE

Closes https://github.com/phpstan/phpstan/issues/11917

This is very similar to `Type:: sliceArray()` and it is used in `NodeScopeResolver` in a similar fashion as `Type::popArray()` is. Moving logic to `Type` deals obviously with all accessory types which were not working properly before too like e.g. list or non-empty-array.

First commit is a tiny refactor I didn't want to cause more noise and dependencies with another PR..

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -1022,6 +1022,108 @@ class ConstantArrayType implements Type
 		return $builder->getArray();
 	}
 
+	public function spliceArray(Type $offsetType, Type $lengthType, Type $replacementType): Type
+	{
+		$keyTypesCount = count($this->keyTypes);
+
+		$offset = $offsetType instanceof ConstantIntegerType ? $offsetType->getValue() : null;
+
+		if ($lengthType instanceof ConstantIntegerType) {
+			$length = $lengthType->getValue();
+		} elseif ($lengthType->isNull()->yes()) {
+			$length = $keyTypesCount;
+		} else {
+			$length = null;
+		}
+
+		if ($offset === null || $length === null) {
+			return $this->degradeToGeneralArray()
+				->spliceArray($offsetType, $lengthType, $replacementType);
+		}
+
+		if ($keyTypesCount + $offset <= 0) {
+			// A negative offset cannot reach left outside the array twice
+			$offset = 0;
+		}
+
+		if ($keyTypesCount + $length <= 0) {
+			// A negative length cannot reach left outside the array twice
+			$length = 0;
+		}
+
+		$offsetWasNegative = false;
+		if ($offset < 0) {
+			$offsetWasNegative = true;
+			$offset = $keyTypesCount + $offset;
+		}
+
+		if ($length < 0) {
+			$length = $keyTypesCount - $offset + $length;
+		}
+
+		$extractType = $this->sliceArray($offsetType, $lengthType, TrinaryLogic::createYes());
+
+		$types = [];
+		foreach ($replacementType->toArray()->getArrays() as $replacementArrayType) {
+			$removeKeysCount = 0;
+			$optionalKeysBeforeReplacement = 0;
+
+			$builder = ConstantArrayTypeBuilder::createEmpty();
+			for ($i = 0;; $i++) {
+				$isOptional = $this->isOptionalKey($i);
+
+				if (!$offsetWasNegative && $i < $offset && $isOptional) {
+					$optionalKeysBeforeReplacement++;
+				}
+
+				if ($i === $offset + $optionalKeysBeforeReplacement) {
+					// When the offset is reached we have to a) put the replacement array in and b) remove $length elements
+					$removeKeysCount = $length;
+
+					if ($replacementArrayType instanceof self) {
+						$valuesArray = $replacementArrayType->getValuesArray();
+						for ($j = 0, $jMax = count($valuesArray->keyTypes); $j < $jMax; $j++) {
+							$builder->setOffsetValueType(null, $valuesArray->valueTypes[$j], $valuesArray->isOptionalKey($j));
+						}
+					} else {
+						$builder->degradeToGeneralArray();
+						$builder->setOffsetValueType($replacementArrayType->getValuesArray()->getIterableKeyType(), $replacementArrayType->getIterableValueType(), true);
+					}
+				}
+
+				if (!isset($this->keyTypes[$i])) {
+					break;
+				}
+
+				if ($removeKeysCount > 0) {
+					$extractTypeHasOffsetValueType = $extractType->hasOffsetValueType($this->keyTypes[$i]);
+
+					if (
+						(!$isOptional && $extractTypeHasOffsetValueType->yes())
+						|| ($isOptional && $extractTypeHasOffsetValueType->maybe())
+					) {
+						$removeKeysCount--;
+						continue;
+					}
+				}
+
+				if (!$isOptional && $extractType->hasOffsetValueType($this->keyTypes[$i])->maybe()) {
+					$isOptional = true;
+				}
+
+				$builder->setOffsetValueType(
+					$this->keyTypes[$i]->isInteger()->no() ? $this->keyTypes[$i] : null,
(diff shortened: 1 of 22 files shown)


tests/PHPStan/Analyser/nsrt/array_splice.php
```diff
@@ -21,41 +21,195 @@ function insertViaArraySplice(array $arr): void
...
+	 */
+	array_splice($arr, -1, 0, ['hello']);
+	assertType('array{a?: 0, 0: \'hello\', b?: 1, c?: 2}', $arr); // Could be array{a?: 0, 0: 'hello', b: 1, c?: 2}
+
+	/**
+	 * @see https://3v4l.org/K5RDp
+	 * @var array{a: 0, b: 1, c?: 2} $arr
+	 */
+	array_splice($arr, 0, -1, ['hello']);
+	assertType('array{0: \'hello\', c?: 2}', $arr); // Could be array{0: 'hello', b?: 1, c?: 2}
```
herndlm: Some types with either negative offsets or negative lengths are not optimal but this one is wrong. Not sure what to do here tbh., the handling of those is extremely hard in combination with optional keys. Without optional keys my current approach seems to work fine.

Ideas about this

- Keep and see if it causes any issues
- Degrade to general array if negative offsets/limits are involved in combination with optional keys
- Rewrite and do something semi-crazy with maybe `getFiniteTypes()` and avoid optional keys completely but have some constant array unions instead

What do you think? Or any other ideas?

staabm: I would prefer a degraded general array over a wrong type (which can lead to more false-positives along the line)

src/Type/Constant/ConstantArrayType.php
```diff
@@ -1012,6 +1006,120 @@ public function sliceArray(Type $offsetType, Type $lengthType, TrinaryLogic $pre
...
+		} else {
+			$length = null;
+		}
+
+		if ($offset === null || $length === null) {
+			return $this->degradeToGeneralArray()
+				->spliceArray($offsetType, $lengthType, $replacementType);
+		}
+
+		if ($keyTypesCount + $offset <= 0) {
```
staabm, line 1044: I wonder whether things like this if-condition could be made more general with a `isSuperTypeOf` instead of a operation on `ConstantIntegerType` only

src/Type/Constant/ConstantArrayType.php
```diff
@@ -1012,6 +1006,120 @@ public function sliceArray(Type $offsetType, Type $lengthType, TrinaryLogic $pre
...
+			 * with offset -4, length 2 and replacement array{17, 19} (which would be spliced to array{a: 0, 0: 17, 1: 19, d: 3, e: 4})
+			 *
+			 * is transformed via reversion to
+			 *
+			 * array{e: 4, d: 3, c: 2, b: 1, a: 0}
+			 * with offset 2 and length 2 (which will be spliced to array{e: 4, d: 3, 1: 19, 0: 17, a: 0} and then reversed again)
+			 */
+			$offset *= -1;
+			$reversedLength = min($length, $offset);
+			$reversedOffset = $offset - $reversedLength;
```
staabm: these operations look like they could be expressed on the type-level directly.

but as you said, maybe this would be too much and we leave it less abstract for now and see whether people report new issues

tests/PHPStan/Analyser/nsrt/array_splice.php
herndlm: Thx for the input! I think I know one of the causes, will take another look later again.

src/Type/Constant/ConstantArrayType.php
herndlm: Can we do such calculations too, can you show me an example maybe? I'm not too confident that I can make it work, but might be also a good follow-up.

staabm: maybe https://github.com/phpstan/phpstan-src/blob/e6d9f0c3ec251ea6fad604d3f97e02bedb4d7f42/src/Type/Php/ArraySumFunctionDynamicReturnTypeExtension.php#L36-L56
or `InitializerExprTypeResolver::get(Minus|Plus|...)Type`

tests/PHPStan/Analyser/nsrt/array_splice.php
herndlm: Ok, so I could add support for negative offsets with optional keys, but not for negative lengths with optional keys. I'm explicitly degrading for the latter case too now, should be safer indeed..

herndlm: found unfortunately a problematic case for negative offsets too. those are both excluded now..

herndlm: sorry for the back and forth, but I found a better way using `ConstantArrayType::getAllArrays()` that leads to perfect accuracy and only degrades if the arrays get too complex.

src/Type/Constant/ConstantArrayType.php
herndlm: can't figure out how to do this now unfortunately, there's also no access to scope in the type. but it's way simpler because non-optional keys don't have to be considered and there's a chance it might work somehow I suppose. maybe still a good follow-up :)

tests/PHPStan/Analyser/nsrt/array_splice.php
```diff
@@ -21,41 +21,261 @@ function insertViaArraySplice(array $arr): void
...
+	 */
+	array_splice($arr, -1, 0, ['hello']);
+	assertType('array{a: 0, b: 1, 0: \'hello\', c?: 2}', $arr);
+
+	/**
+	 * @see https://3v4l.org/K5RDp
+	 * @var array{a: 0, b: 1, c?: 2} $arr
+	 */
+	array_splice($arr, 0, -1, ['hello']);
+	assertType('array{0: \'hello\', b: 1}|array{0: \'hello\', c: 2}', $arr);
```
herndlm: not sure how useful it is that PHPStan fully understands this, but I think it's pretty neat 😊

tests/PHPStan/Analyser/nsrt/bug-5017.php
```diff
@@ -12,10 +12,10 @@ public function doFoo()
 		$items = [0, 1, 2, 3, 4];
 
 		while ($items) {
-			assertType('non-empty-array<0|1|2|3|4, 0|1|2|3|4>', $items);
+			assertType('non-empty-list<int<min, 4>>', $items);
 			$batch = array_splice($items, 0, 2);
-			assertType('array<0|1|2|3|4, 0|1|2|3|4>', $items);
-			assertType('non-empty-list<0|1|2|3|4>', $batch);
+			assertType('list<int<min, 4>>', $items);
+			assertType('non-empty-list<int<min, 4>>', $batch);
```
herndlm: the changes here confused me briefly, but it seems to be coming from loop generalization and there's not much I can do about it

src/Type/Constant/ConstantArrayType.php
herndlm: I'd like to change this, but for the same reasons mentioned in https://github.com/phpstan/phpstan-src/pull/3952#discussion_r2052024142 I seem to can't be doing this right now.

herndlm:
Meh, I'll do one more improvement and also add tests for the return types into the same file. Surely can't hurt and hopefully does not discover slice issues...

tests/PHPStan/Analyser/nsrt/array_splice.php
herndlm: it still understands this, but the return array is not a union anymore because the solution changed

herndlm:
Ok, so I figured out a way to get correct types for `array_splice()` without using `getAllArrays()`. But I had to use `sliceArray()` internally once to deal with the optional offsets.

ondrejmirtes:
Thank you!

PR #3954, opened 2025-04-20
More precise `count($array)` after `array_*` functions

staabm:
closes https://github.com/phpstan/phpstan/issues/7804


staabm:
Just found https://github.com/phpstan/phpstan-src/pull/1828#issuecomment-1277121044 again, which tells me this PR might not be acceptable

PR #3958, opened 2025-04-21
Rework ArrayReplaceFunctionReturnTypeExtension

VincentLanglet:
Close https://github.com/phpstan/phpstan/issues/12828

I basically copy the ArrayMergeFunctionReturnTypeExtension with the following change

https://github.com/phpstan/phpstan-src/blob/4c41073c1dcdf44b6c3c90768c4430e45238b522/src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php#L87
`$keyType instanceof ConstantIntegerType ? null : $keyType,` is replaced by `$keyType` because
- array_merge reindex numeric keys
- array_replace keep all the existing keys 

https://github.com/phpstan/phpstan-src/blob/4c41073c1dcdf44b6c3c90768c4430e45238b522/src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php#L106
`if (!(new IntegerType())->isSuperTypeOf($keyType)->yes()) {` is replaced by `if (!$argType->isList()->yes()) {` because of the same reason.

### DIFF
--- a/src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
@@ -5,12 +5,21 @@ namespace PHPStan\Type\Php;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\TrinaryLogic;
+use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
+use PHPStan\Type\Constant\ConstantIntegerType;
+use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use function array_keys;
 use function count;
+use function in_array;
 use function strtolower;
 
 final class ArrayReplaceFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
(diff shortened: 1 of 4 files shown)


herndlm:
I know I'm maybe a bit fast suggesting this, but we could create a `Type::mergeArray()` or `Type::replaceArray()` with a proper `$preserveKeys` or `$reindex` arg or whatever and then us it in both return type extensions. And maybe I should have done this already the last time I touched the merge one. Anyway, can be done in a follow-up too IMO, so this comment should not be considered a blocker IMO 😊 and I can volunteer to do this.

tests/PHPStan/Analyser/nsrt/array-replace.php
```diff
@@ -23,9 +23,9 @@ public function arrayReplace($array1, $array2): void
 	 */
 	public function arrayReplaceArrayShapes($array1, $array2): void
 	{
-		assertType("non-empty-array<'bar'|'foo', '1'|'2'>", array_replace($array1));
-		assertType("non-empty-array<'bar'|'foo', '1'|'2'>", array_replace([], $array1));
-		assertType("non-empty-array<'bar'|'foo', '1'|'2'|'4'>", array_replace($array1, $array2));
+		assertType("array{foo: '1', bar: '2'}", array_replace($array1));
+		assertType("array{foo: '1', bar: '2'}", array_replace([], $array1));
+		assertType("array{foo: '1', bar: '4'}", array_replace($array1, $array2));
 	}
```
herndlm, line 29: Can you add another test cases where `$array2` is a general array, please? And maybe in the other section somewhere also one where `$array1` is a general array but `$array2` is a constant array.

a bit more mixed keys (strings or ints) would be interesting too 😊 that would also help massively with potential future refactoring

VincentLanglet:
> Anyway, can be done in a follow-up too IMO, so this comment should not be considered a blocker IMO 😊 and I can volunteer to do this.

Yeah, I would prefer this to be done in a follow up PR to separate (and validate separately by ondrej)
- The new implementation
- The refactoring

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
```diff
@@ -23,54 +32,107 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
 
 	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
 	{
-		$arrayTypes = $this->collectArrayTypes($functionCall, $scope);
+		$args = $functionCall->getArgs();
```
samsonasik, line 35: is `isFirstClassCallable()` check needed first? The `getArgs()` got assert notice when there is `...` variadic place holder.

https://github.com/nikic/PHP-Parser/blob/20b0d55f6655ef2d73d841f6b7d87117a90f9ee2/lib/PhpParser/Node/Expr/CallLike.php#L32

ondrejmirtes: The extensions are not called with first-class callables at all.

herndlm:
I guess this is fine. And also brings value of course. Let's see if we can refactor array_merge and array_replace stuff a bit as follow-up I'd say :)

herndlm (approved):

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
```diff
@@ -23,54 +32,107 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+		);
+
+		if ($allConstant->yes()) {
+			$newArrayBuilder = ConstantArrayTypeBuilder::createEmpty();
+
+			foreach ($argTypes as $argType) {
+				/** @var array<int|string, ConstantIntegerType|ConstantStringType> $keyTypes */
+				$keyTypes = [];
+				foreach ($argType->getConstantArrays() as $constantArray) {
+					foreach ($constantArray->getKeyTypes() as $keyType) {
```
ondrejmirtes, line 81: I don't think this is what you want. Let's say you have `array{foo: int}|array{bar: string}`.

You'll end up with two values in `$keyTypes` and you'll create a flat array out of them. Also you'll overwrite a value if two or more arrays in a union share the same key but with different type.

ondrejmirtes (changes requested):

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
VincentLanglet: I'm not sure to understand your point.

This is the current behavior we have with array_replace vs array_merge
https://phpstan.org/r/4ec3a6a9-2645-4565-a81e-577b518eb7c3

Since both function are working almost the same way, I copied and adapt the logic from ArrayMergeFunctionReturnTypeExtension
https://github.com/phpstan/phpstan-src/blob/84e231661a9b542d3114e8d9499626b784864282/src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php#L74-L94

In which we're doing
```
foreach ($argType->getConstantArrays() as $constantArray) {
					foreach ($constantArray->getKeyTypes() as $keyType) {
```

With this PR, we're now getting
```
array{bar?: int, foo: 'thing'}
```
with both array_merge and array_replace

ondrejmirtes:
Thank you.

PR #3959, opened 2025-04-21
Fix `array_slice()` edge cases

herndlm:
Noticed this while finishing https://github.com/phpstan/phpstan-src/pull/3952

Edge cases fixed here:

1. Bug fix: we cannot assume that the result of `array_slice($array, 0, <positive int>) ` with a non-empty-array as input will also return a non-empty array because we don't know which of the keys is set. That only can be done for a non-empty-list. The last test change shows this nicely
```diff
                        assertType('non-empty-array<int>', $items);
                        $batch = array_splice($items, 0, 2);
                        assertType('array<int>', $items);
-                       assertType('non-empty-array<int>', $batch);
+                       assertType('array<int>', $batch);
```

2. Improvement: `array_slice` with length set to 0 will always return `array{}`

3. Bug fix: `array_slice` with negative offset and negative length of the same value or smaller will always return `array{}`

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-5017.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-5017.php
@@ -15,7 +15,7 @@ class Foo
 			assertType('non-empty-array<0|1|2|3|4, 0|1|2|3|4>', $items);
 			$batch = array_splice($items, 0, 2);
 			assertType('array<0|1|2|3|4, 0|1|2|3|4>', $items);
-			assertType('non-empty-list<0|1|2|3|4>', $batch);
+			assertType('list<0|1|2|3|4>', $batch);
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/nsrt/bug-5017.php
```diff
@@ -15,7 +15,7 @@ public function doFoo()
 			assertType('non-empty-array<0|1|2|3|4, 0|1|2|3|4>', $items);
 			$batch = array_splice($items, 0, 2);
 			assertType('array<0|1|2|3|4, 0|1|2|3|4>', $items);
-			assertType('non-empty-list<0|1|2|3|4>', $batch);
+			assertType('list<0|1|2|3|4>', $batch);
```
herndlm, line 18: this is unfortunate, but it happens because the `$items` type is not as good as it can be. This will be improved in https://github.com/phpstan/phpstan-src/pull/3952 and then it will be `non-empty-list` again.

So there's unfortunately some kind of coupling here.

herndlm:
I'm happy with this one and can't do much more I think. Just mentioning it because of what I asked in https://github.com/phpstan/phpstan/discussions/12919. I don't have the resources right now to look into that new type methods with all the adaptions needed 😅

ondrejmirtes:
Thank you!

herndlm:
Ok thx, I'll rebase the array_splice stuff right away now

PR #3960, opened 2025-04-21
Introduce StrrevFunctionReturnTypeExtension

VincentLanglet:
Improves https://github.com/phpstan/phpstan-src/pull/3921

### DIFF
--- a/src/Type/Php/StrrevFunctionReturnTypeExtension.php
+++ b/src/Type/Php/StrrevFunctionReturnTypeExtension.php
@@ -30,11 +30,11 @@ final class StrrevFunctionReturnTypeExtension implements DynamicFunctionReturnTy
 		FunctionReflection $functionReflection,
 		FuncCall $functionCall,
 		Scope $scope,
-	): Type
+	): ?Type
 	{
 		$args = $functionCall->getArgs();
 		if (count($args) < 1) {
-			return new StringType();
+			return null;
 		}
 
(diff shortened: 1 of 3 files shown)


src/Type/Php/StrrevFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,73 @@
...
+
+	public function getTypeFromFunctionCall(
+		FunctionReflection $functionReflection,
+		FuncCall $functionCall,
+		Scope $scope,
+	): Type
+	{
+		$args = $functionCall->getArgs();
+		if (count($args) < 1) {
+			return new StringType();
```
ondrejmirtes: `return null` instead

src/Type/Php/StrrevFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,73 @@
...
+			$accessoryTypes[] = new AccessoryUppercaseStringType();
+		}
+
+		if (count($accessoryTypes) > 0) {
+			$accessoryTypes[] = new StringType();
+
+			return new IntersectionType($accessoryTypes);
+		}
+
+		return new StringType();
```
ondrejmirtes: return null instead

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #3963, opened 2025-04-25
More precise string functions return type with replacement array

staabm:
closes https://github.com/phpstan/phpstan/issues/12928


ondrejmirtes:
Thank you!

PR #3964, opened 2025-04-25
Keep list on unset() with nested dim-fetch

staabm:
closes https://github.com/phpstan/phpstan/issues/12927
closes https://github.com/phpstan/phpstan/issues/12330
closes https://github.com/phpstan/phpstan/issues/11171
closes https://github.com/phpstan/phpstan/issues/8282
closes https://github.com/phpstan/phpstan/issues/12754

### DIFF
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -163,18 +163,7 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 
 	public function setExistingOffsetValueType(Type $offsetType, Type $valueType): Type
 	{
-		if ((new ConstantIntegerType(0))->isSuperTypeOf($offsetType)->yes()) {
-			return $this;
-		}
-
-		if (
-			$valueType->isArray()->yes()
-			&& IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($offsetType)->yes()
-		) {
-			return $this;
-		}
-
-		return new ErrorType();
+		return $this;
 	}
 
 	public function unsetOffset(Type $offsetType): Type
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -151,6 +151,13 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 			return $this;
 		}
 
+		if (
+			$valueType->isArray()->yes()
+			&& IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($offsetType)->yes()
+		) {
+			return $this;
+		}
+
 		return new ErrorType();
 	}
 
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -151,13 +151,6 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 			return $this;
 		}
 
-		if (
-			$valueType->isArray()->yes()
-			&& IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($offsetType)->yes()
-		) {
-			return $this;
-		}
-
 		return new ErrorType();
 	}
 
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -160,6 +160,13 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 			return $this;
 		}
 
+		if (
+			$valueType->isArray()->yes()
+			&& IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($offsetType)->yes()
+		) {
+			return $this;
+		}
+
 		return new ErrorType();
 	}
 
--- a/src/Type/ArrayType.php
+++ b/src/Type/ArrayType.php
@@ -362,12 +362,6 @@ class ArrayType implements Type
 	public function setExistingOffsetValueType(Type $offsetType, Type $valueType): Type
 	{
 		if ($offsetType instanceof ConstantStringType || $offsetType instanceof ConstantIntegerType) {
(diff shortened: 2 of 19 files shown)


ondrejmirtes:
Looks like this fixes a few more issues :) https://github.com/phpstan/phpstan-src/actions/runs/14669281251 Please add regression tests.

herndlm:
Nice 👍

src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -151,6 +151,13 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
 			return $this;
 		}
 
+		if (
+			$valueType->isArray()->yes()
+			&& IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($offsetType)->yes()
```
ondrejmirtes: I don't get this part at all.

1) Why does it matter what `$valueType` is?
2) List certainly isn't preserved if I assign any positive-int offset.

src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -160,6 +167,13 @@ public function setExistingOffsetValueType(Type $offsetType, Type $valueType): T
...
+		) {
+			return $this;
+		}
+
 		return new ErrorType();
```
ondrejmirtes: I think this method can just return `$this` in any case. Assigning existing offset on list will preserve list.

src/Type/IntersectionType.php
```diff
@@ -799,6 +799,9 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
 		}
 
 		$result = $this->intersectTypes(static fn (Type $type): Type => $type->setOffsetValueType($offsetType, $valueType, $unionValues));
+		if ($this->isList()->yes() && $valueType->isArray()->yes()) {
+			$result = TypeCombinator::intersect($result, new AccessoryArrayListType());
```
ondrejmirtes: Again, I don't get why does it matter what `$valueType` is.

ondrejmirtes (changes requested):

src/Type/IntersectionType.php
staabm: the idea was:

```
/** @param list<array<int>> $list */
function doFoo($list, int $i) {
  $list[$i][3] = 123;
}
```

when setting offset $i on $list to `array(3 => 123)`, it does not turn $list into array.
(which is only true when `$i` is a existing offset)

ondrejmirtes: But that's true for anything, not just arrays. I feel like we should call `setExistingOffsetValueType` more if we're sure the offset exists.

ondrejmirtes: This just doesn't make sense. A simple code:

```php
/** @param list<array<int>> $list */
function doFoo($list, int $i) {
	$list[$i][3] = 123;
	\PHPStan\dumpType($list);
}
```

We don't know whether `$list[$i]` exists. 2.1.x is correct to dump this as `non-empty-array<int, non-empty-array<int>&hasOffsetValue(3, 123)>`.

Your branch dumps this as `non-empty-list<non-empty-array<int>&hasOffsetValue(3, 123)>` which is wrong.

ondrejmirtes: For example, with code like this:

```php
foreach ($list as $i => $value) {
    $list[$i][3] = 123;
}
```

We should preserve `$list` being a list because `$list[$i]` exists and this information is already carried in Scope (hasExpressionType).

Right now `setExistingOffsetValueType` is only called for `unset()` but we could call it for when we're sure the offset exists.

staabm: I am now using `setExistingOffsetValueType` more, but in the process somehow broke some tests.
need to investigate further, but as I kind of working thru the fog was thinking it would be worth sharing the progress to verify I am working in the right direction

tests/PHPStan/Rules/Variables/data/bug-8113.php
```diff
@@ -34,15 +34,15 @@ function () {
 		];
 		assertType("non-empty-array<array<mixed>>&hasOffsetValue('Review', array{id: null, text: null, answer: null})&hasOffsetValue('SurveyInvitation', non-empty-array&hasOffsetValue('review', null))", $review);
 		unset($review['SurveyInvitation']['review']);
-		assertType("non-empty-array<array<mixed>>&hasOffsetValue('Review', array<mixed~'review', mixed>)&hasOffsetValue('SurveyInvitation', array<mixed~'review', mixed>)", $review);
+		assertType("non-empty-array<array<mixed>>&hasOffsetValue('Review', array{id: null, text: null, answer: null})&hasOffsetValue('SurveyInvitation', array<mixed~'review', mixed>)", $review);
```
staabm, line 37: I think this expectation change is correct.. previously `unset($review['SurveyInvitation']['review']);` affected the offset `Review` (uppercase `R`) which was a bug IMO

src/Type/ArrayType.php
```diff
@@ -356,9 +356,29 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
 
 	public function setExistingOffsetValueType(Type $offsetType, Type $valueType): Type
 	{
-		return new self(
-			$this->keyType,
-			TypeCombinator::union($this->itemType, $valueType),
+		if ($offsetType instanceof ConstantStringType || $offsetType instanceof ConstantIntegerType) {
```
ondrejmirtes: The idea is that we're overwriting an existing offset so `$this->keyType` should stay intact as is in the original version of this method.

Of course setExistingOffsetValueType was used when unsetting an offset so some details somewhere might be wrong, but what you're doing here now is basically the same thing `setOffsetValueType` is doing so it seems wrong.

ondrejmirtes (changes requested):

ondrejmirtes (changes requested):
This looks great! Please fix the conflicts.

ondrejmirtes:
Thank you!

PR #3965, opened 2025-04-28
More precise types after assignment when strict-types=0

staabm:
Improve types on top of https://github.com/phpstan/phpstan-src/pull/3945 but for the non-strict-types case.

see https://phpstan.org/r/aa5d5753-eca2-4735-bebf-6479e1ab1f42

~~do you think we could remove nullability from the property-type in scope, as the method we call returns a native non-nullable type, even if `strict-types=0`...?~~

closes https://github.com/phpstan/phpstan/issues/12946
closes https://github.com/phpstan/phpstan/issues/12940
closes https://github.com/phpstan/phpstan/issues/12950
closes https://github.com/phpstan/phpstan/issues/12947


ondrejmirtes:
I don't get why this is needed and why the current logic from https://github.com/phpstan/phpstan-src/pull/3945 doesn't already work for that. Can you explain please?

staabm:
~~The logic implemented in the already merged PR only takes the native type from the assignment, when one of the property-native-flattened types is a super-type of the assigment type.~~

~~In the given example the native type of the property is not a super-type of the assigment. Thats why the current logic doesn't take effect. Since we are in non-strict type-land the else-if also doesn't narrow the types.~~


~~New: as a last resort we therefore only use the native knowledge about non-nullability of the assignment, to narrow the property type in scope.~~

~~The example only works because of phpdoc types (and these types are not expressable with native types - aka param-out in this case; similar thing could happen in conditional return types)~~

staabm:
scratch the previous comment.

before https://github.com/phpstan/phpstan-src/pull/3945 we always assigned a type, and since https://github.com/phpstan/phpstan-src/pull/3945 we no longer do so. we are missing a ELSE case

src/Analyser/NodeScopeResolver.php
```diff
@@ -5641,15 +5641,15 @@ static function (): void {
 
 						if ($assignedTypeIsCompatible) {
 							$scope = $scope->assignExpression($var, $assignedExprType, $assignedNativeType);
-						} elseif ($scope->isDeclareStrictTypes()) {
```
staabm, line 5644: with the [last commit](https://github.com/phpstan/phpstan-src/pull/3965/commits/48fad996d5eb896c71117323dcae739095003dfb) I made PHPStan always do type-coersion, even if non-strict-types are used.

I think we can do that in the changed areas, because we are working on natively typed properties and these always convert whatever you assign to them (independent from strict-types=0/1)

see e.g. https://3v4l.org/qnsMN

tests/PHPStan/Analyser/nsrt/bug-12393b.php
```diff
@@ -0,0 +1,145 @@
+<?php declare(strict_types = 0);
```
staabm: copied the test 12393 and changed only the strict_types to 0

src/Analyser/NodeScopeResolver.php
ondrejmirtes: See the reason why we can't simply do this:

```php
class Foo
{

	public int $foo;

	public function doFoo(string $s): void
	{
		$this->foo = $s;
		assertType('int', $this->foo);
	}

	public function doBar(): void
	{
		$this->foo = 'foo';
		assertType('int', $this->foo);
	}

}
```

Reports:

```
15     Expected type int, actual: *NEVER*¨
21     Expected type int, actual: *NEVER*
```

On your PR.

I'm fine with the state of assigning from before all the `toCoercedArgumentType` changes when strict_types is 0. It simply looked like this: https://github.com/phpstan/phpstan-src/blob/e6dc705b29b90dcc5f9773377c05aecbfe9fba3a/src/Analyser/NodeScopeResolver.php#L5215

But `->toCoercedArgumentType(true)` should only be called and assigned with strict_types = 1.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: thanks for the pointer. just to double check: in PHPStan 2.0.12 we get this results for your test:

```php
class Foo
{

	public int $foo;

	public function doFoo(string $s): void
	{
		$this->foo = $s;
		assertType('string', $this->foo); // should be int
	}

	public function doBar(): void
	{
		$this->foo = 'foo';
		assertType("'foo'", $this->foo); // should be int
	}

}
```

and this is good enough to get back to, even though its wrong for now, right?

ondrejmirtes:
I just forcepushed here how the code makes sense to me. We just need to fix the failing tests 😂

staabm:
thanks for the hints.. I will look into the errors after a lunch break

staabm:
> We just need to fix the failing tests 😂

do you mean adjusting test expectations or fixing the underlying bugs?

ondrejmirtes:
Fix the bugs

staabm:
ready for another review. added loads of new tests

ondrejmirtes:
I didn't mean you to start changing the toCoerced... method. It we do that, we need to do it right.

What I meant right now is to slightly adjust what the "else" branch in NodeScopeResolver does after my force-push.

It's up to you 😊

staabm:
> What I meant right now is to slightly adjust what the "else" branch in NodeScopeResolver does after my force-push.

I don't see a way how a slight change can fullfill the tests

> It we do that, we need to do it right.

could you give me an idea whats "wrong" in the current PR?
do you see errors, or do you mean more precise results - like non-empty-string -> int cannot be 0 ..?

ondrejmirtes (changes requested):
This looks promising. But you haven't touched `toCoercedArgumentType` in some types. For example in ArrayTypeTrait, HasOffset*Type, accessory array types, NullType, ObjectType (Stringable object can be passed to string imho), ObjectTypeTrait.

For example VoidType should do `(new NullType)->toCoercedArgumentType($strictTypes)`.

staabm:
thanks for the hints.

I have added support for Stringable objects and callable types.
callables are interessting because they can be coerced to array/string in weak and strict mode
https://3v4l.org/WJ8NW

intentionally I did not add array types, as I don't find a example in which they can be coerced.
I have added tests for the cases I could think of, but these did not need adjustments in one of the `toCoercedArgumentType` Type methods.

I have also added tests for void type and these also did not need adjustments in `toCoercedArgumentType` Type methods.

ondrejmirtes:
@staabm These coercions are described on PHP website. See the links near the bottom at https://www.php.net/manual/en/language.types.type-juggling.php

[Converting to boolean](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting)
    [Converting to integer](https://www.php.net/manual/en/language.types.integer.php#language.types.integer.casting)
    [Converting to float](https://www.php.net/manual/en/language.types.float.php#language.types.float.casting)
    [Converting to string](https://www.php.net/manual/en/language.types.string.php#language.types.string.casting)
    [Converting to array](https://www.php.net/manual/en/language.types.array.php#language.types.array.casting)
    [Converting to object](https://www.php.net/manual/en/language.types.object.php#language.types.object.casting)
    [Converting to resource](https://www.php.net/manual/en/language.types.resource.php#language.types.resource.casting)
    [Converting to NULL](https://www.php.net/manual/en/language.types.null.php#language.types.null.casting)

I'm especially worried about those array accessory types because they just return this, I'm not sure what would happen for example when you have non-empty-array and try to assign it into integer, whethey it'd work properly or not.

staabm:
I think the conversion rules on these manual pages describe what happens on  explicit casts.
property assigments doen't apply the same rules.

see e.g. https://3v4l.org/H5XnR while you can cast a `1` to array, you cannot assign `1` to a array typed property

>  for example when you have non-empty-array and try to assign it into integer, whethey it'd work properly or not.

just added tests for the mentioned cases

staabm:
As I understand it, implicit conversion only happens for the listed scalar types

grafik

tests/PHPStan/Analyser/nsrt/bug-12393b.php
```diff
@@ -0,0 +1,701 @@
...
+class FooVoidInt {
+    private ?int $foo;
+    private int $fooNonNull;
+
+    public function doFoo(): void {
+        $this->foo = $this->returnVoid();
+        assertType('null', $this->foo);
+
+        $this->fooNonNull = $this->returnVoid();
+        assertType('int|null', $this->foo); // should be *ERROR*
```
staabm: I think this is a bug regarding void-to-null transformation which should be fixed in a followup PR

its a rare edge case

ondrejmirtes:
Oh interesting, didn't know about that.

tests/PHPStan/Analyser/nsrt/bug-12393.php
```diff
@@ -143,3 +143,23 @@ public function getMixed()
 	}
 
 }
+
+// https://3v4l.org/LK6Rh
+class CallableString {
+	private string $foo;
+
+	public function doFoo(callable $foo): void {
+		$this->foo = $foo; // PHPStorm wrongly reports an error on this line
```
staabm, line 152: opened a PHPStorm bug: https://youtrack.jetbrains.com/issue/WI-81263/Property-must-be-of-type-string-callable-provided-false-positive

src/Type/ObjectType.php
```diff
@@ -706,6 +706,10 @@ public function toArrayKey(): Type
 
 	public function toCoercedArgumentType(bool $strictTypes): Type
 	{
+		if (!$strictTypes) {
+			return TypeCombinator::union($this, $this->toString());
```
ondrejmirtes, line 718: If the object isn't a Stringable, this should just return `$this`. But instead it will return ErrorType.

Also I'm not sure what happens with `BcMath\Number` in strict_types=1 or 0.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/bug-12393.php
ondrejmirtes: To be fair PHPStan also reports an error for that (which is correct, we mostly don't differentiate between strict_types 1 or 0 in `Type::accepts()`.

src/Type/ObjectType.php
staabm: added tests with links to 3v4l.org

ondrejmirtes:
Alright, I'm really excited about this, let's try it out!

VincentLanglet:
Hi @staabm, 

I just saw this and I have some question about this change.

1) Why https://phpstan.org/r/908e9206-cf9a-4148-8e3b-18805be3ab1a is still reported in strict-types 0 but
https://phpstan.org/r/5075f3d0-f6c6-42ed-bf20-7292529cd705 is not reported anymore in strict-types 0 (but still in strict-types 1).

2) When I have an existing codebase I want to migrate from strict_types 0 to 1, I generally rely on PHPStan to prevent futur error. If I understand this PR, it's now not possible now since some error are not reported anymore. What about an option to tell PHPStan to keep consider files like `strict_types=1` even if they currently are `strict_types=0` (and this option could be activated in strict-rules) ?

staabm:
Huhu

1)

The PR is about type inference, meaning PHPStan better know when/which types flow thru the code. It does not change the rules when errors are reported.
(Which type accepts which)

Regarding BcMath\Number. it behaves like that because the php runtime only errors when strict_types=1, but for strict_types=0 it's legale to assign a object which implements __toString to a string. 

2)

 I don't know whether this makes sense. In case you want strict_types=1 you can declare it and work on the errors. ondrey might have a clear opinion on it

staabm:
The 2 different modes have different characteristics and not necessarily everyone want to migrate from weak to strict. 

I have recently seen discussions of php-src devs whether a new mode should be implemented which is a mix of the 2 existing ones, as both have some advantages

ondrejmirtes:
@VincentLanglet This behaviour didn't change with this PR. If PHPStan reports errors in the same situations where PHP reports TypeError, it's correct.

PR #3967, opened 2025-04-28
Remove expired cloudflare certificate

staabm:
This certificate was added with https://github.com/phpstan/phpstan-src/commit/30f68ba7555e88ad9bdb326a1bed735d79fb29ea but expired in february 2025

see  https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/user-side-certificates/

grafik

cloudflare no longer has a "general purpose" certificate, therefore we can't provide a certificate for PHPStan PRO which works for all customers. therefore deleting it.


VincentLanglet:
Every time I try to run `composer install` it's blocked on this patch. Do you encounter the same behavior ?

staabm:
nope, it works for me on macOS.

do you have gpatch installed?

VincentLanglet:
> do you have gpatch installed?

Oh I missed this part in the readme, thanks !

PR #3968, opened 2025-04-28
TypehintHelper: remove unneeded default param value

staabm:
Calling `TypehintHelper->decideType()` without 2 arguments kind of misses the point. we may pass `null` but calling it with only 1 parameter does not make sense.


ondrejmirtes:
Thank you.

PR #3785, opened 2025-01-16
Extract ArrayColumnHelper from ArrayColumnFunctionReturnTypeExtension

staabm:
extracted the 3 refactoring commits from https://github.com/phpstan/phpstan-src/pull/3706

no functional changes

will send the new ArrayColumnRule with a separate PR after this one merged

PR #3970, opened 2025-04-29
Fix `array_column()` with explicit null `$index_key`

herndlm:
Closes https://github.com/phpstan/phpstan/issues/12945

### DIFF
--- a/src/Type/Php/ArrayColumnFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayColumnFunctionReturnTypeExtension.php
@@ -6,6 +6,7 @@ use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\NullType;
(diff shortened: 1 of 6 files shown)


herndlm:
hey @staabm, I only saw your refactor in https://github.com/phpstan/phpstan-src/pull/3785 now because of conflicts :) I think the `$indexType` should not be nullable to avoid confusion / inconsistency with `null` vs `NullType`.

I think your follow-up for the rules does not exist yet, does it? Would it work with a non-nullable indexType too? I did adapt the new helper here with the rebase to keep the fix for the linked issue simple.

@ondrejmirtes I avoid pinging you, but looks like this small fix is a blocker or has conflict potential for that new rule

staabm:
thanks for the ping

> I think the `$indexType` should not be nullable to avoid confusion / inconsistency with `null` vs `NullType`.

IIRC its fine, as the rule - as I had it implemted previously in https://github.com/phpstan/phpstan-src/pull/3706/files - does not rely on the nullability of the parameter.

I will have a closer look when re-submitting the new rule

herndlm:
jfyi, this is an easy one I think and done already for a while

ondrejmirtes:
Thank you!

PR #3972, opened 2025-04-29
non-falsy-string cannot be false

staabm:


staabm:
its already implemented via `TruthyBooleanTypeTrait`

PR #3975, opened 2025-04-30
Fix crash on dynamic numeric-string symbols

staabm:
closes https://github.com/phpstan/phpstan/issues/12949

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-12949.php
+++ b/tests/PHPStan/Analyser/data/bug-12949.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Bug12949;
+
+function doFoo():void {
+	$b = '0';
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/data/bug-12949.php
```diff
@@ -0,0 +1,19 @@
...
+	${$b} = 1;
+
+	echo "";
+}
+
+function doBar(object $o):void {
+	$b = '0';
+	$o->{$b}();
+	$o::{$b}();
+	echo $o::{$b};
```
staabm, line 16: I don't think these cases are realistic, but I fixed them to keep the implementation of the 4 rules in sync

ondrejmirtes:
Perfect, thank you!

PR #3976, opened 2025-04-30
feat: add option to report ignores without comments

calebdw:
Hello!

Closes https://github.com/phpstan/phpstan/issues/11340

This PR adds the following two option (false by default):
- ~`reportIgnoresWithoutIdentifiers`: reports `@phpstan-ignore-line` and `@phpstan-ignore-next-line` usages~
- `reportIgnoresWithoutComments`: reports `@phpstan-ignore` usages without comments

Thanks!

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -118,6 +118,7 @@ parameters:
 		resolvedPhpDocBlockCacheCountMax: 2048
 		nameScopeMapMemoryCacheCountMax: 2048
 	reportUnmatchedIgnoredErrors: true
+	reportIgnoresWithoutComments: false
 	typeAliases: []
 	universalObjectCratesClasses:
 		- stdClass
@@ -226,6 +227,7 @@ parameters:
 		- [parameters, errorFormat]
 		- [parameters, ignoreErrors]
 		- [parameters, reportUnmatchedIgnoredErrors]
+		- [parameters, reportIgnoresWithoutComments]
 		- [parameters, tipsOfTheDay]
 		- [parameters, parallel]
 		- [parameters, internalErrorsCountLimit]
--- a/src/Parser/RichParser.php
+++ b/src/Parser/RichParser.php
@@ -7,12 +7,14 @@ use PhpParser\Node;
 use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitor\NameResolver;
 use PhpParser\Token;
+use PHPStan\Analyser\FileAnalyserResult;
 use PHPStan\Analyser\Ignore\IgnoreLexer;
 use PHPStan\Analyser\Ignore\IgnoreParseException;
 use PHPStan\DependencyInjection\Container;
 use PHPStan\File\FileReader;
 use PHPStan\ShouldNotHappenException;
 use function array_filter;
+use function array_key_last;
 use function array_map;
 use function count;
 use function implode;
@@ -30,6 +32,9 @@ use const T_COMMENT;
 use const T_DOC_COMMENT;
 use const T_WHITESPACE;
 
+/**
+ * @phpstan-import-type Identifier from FileAnalyserResult
+ */
 final class RichParser implements Parser
 {
 
@@ -120,7 +125,7 @@ final class RichParser implements Parser
 
 	/**
 	 * @param Token[] $tokens
-	 * @return array{lines: array<int, non-empty-list<string>|null>, errors: array<int, non-empty-list<string>>}
+	 * @return array{lines: array<int, non-empty-list<Identifier>|null>, errors: array<int, non-empty-list<string>>}
 	 */
 	private function getLinesToIgnore(array $tokens): array
 	{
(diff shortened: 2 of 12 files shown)


calebdw:
CI failures look to be unrelated

NanoSector:
Thanks for providing this & also picking up the case for no comments :raised_hands:

ondrejmirtes:
I'm thinking about deprecating `@phpstan-ignore-line` and `@phpstan-ignore-next-line` completely now and removing them in 3.0. Because people often use them with identifiers by mistake, but these comments ignore all errors anyway. So I don't want to add `reportIgnoresWithoutIdentifiers` now.

We'd have to add a support for new comment `@phpstan-ignore-all-errors` to have a replacement for these though.

As for `reportIgnoresWithoutComments`, yeah, that one makes sense to me and I'd accept it.

calebdw:
Sounds good---I'll update

> We'd have to add a support for new comment @phpstan-ignore-all-errors to have a replacement for these though.

That's good---however, consider adding an option to enable/disable this when the new comment is added. Some projects may want to complete disable this and not allow such a global ignore

derrabus:
> I'm thinking about deprecating `@phpstan-ignore-line` and `@phpstan-ignore-next-line` completely now and removing them in 3.0. Because people often use them with identifiers by mistake, but these comments ignore all errors anyway. So I don't want to add `reportIgnoresWithoutIdentifiers` now.

A deprecation would also work for me. I want to rid my codebases of those annotations, so as long as an error pops up if one of the old annotations is used, I'm happy. ✌🏻

calebdw:
@ondrejmirtes, this is ready for review :+1:

VincentLanglet:
> That's good---however, consider adding an option to enable/disable this when the new comment is added. Some projects may want to complete disable this and not allow such a global ignore

This feel like a very special/opinionated rules which might make more sens in a dedicated repository or just in your personal project IMHO.

And there is already some way to achieve this like using PHP-CS-Fixer
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/phpdoc/general_phpdoc_annotation_remove.rst
and configuring those tag as "to remove".

Certainly the same could exist with PHP CodeSniffer or Rector or any linter.

I personally think the same about `reportIgnoresWithoutComments` but I let ondrej decide

conf/config.neon
```diff
@@ -118,6 +118,7 @@ parameters:
 		resolvedPhpDocBlockCacheCountMax: 2048
 		nameScopeMapMemoryCacheCountMax: 2048
 	reportUnmatchedIgnoredErrors: true
+	reportIgnoresWithoutComments: false
```
VincentLanglet, line 121: The current implementation only apply to phpstan-ignore with identifier.

I wonder if the option should apply to `phpstan-ignore-line` or `phpstan-ignore-next-line` too, until those tag are deprecated.

src/Parser/RichParser.php
```diff
@@ -349,6 +361,7 @@ private function parseIdentifiers(string $text, int $ignorePos): array
 			throw new IgnoreParseException('Missing identifier', 1);
 		}
 
+		/** @phpstan-ignore return.type (return type is correct, not sure why it's being changed from array shape to key-value shape) */
```
VincentLanglet, line 364: Certainly because of
```
$key = array_key_last($identifiers);
if ($key !== null) {
  $identifiers[$key]['comment'] = $comment;
  $comment = null;
}
```

Dunno if it could help to write
```
/** @var array<array{...}> */ 
$identifiers = [];
```
When initializing the array.

Another solution could be a format
`array{identifiers: array<string>, comments: array<string>}`
but I'm not sure if it's better...

tests/PHPStan/Analyser/data/ignore-no-comments.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace IgnoreNoComments;
+
+class Foo
+{
+	public function doFoo(): void
+	{
+		echo $foo; // @phpstan-ignore variable.undefined
+		echo $foo; // @phpstan-ignore wrong.id (comment)
```
VincentLanglet, line 10: I wasn't even aware comment had to be inside parenthesis.

Might be worth adding a test case for
```
// @phpstan-ignore wrong.id commentWithoutParenthesis
```
Also maybe a tip could be added to the reported error ?

tests/PHPStan/Analyser/AnalyserTest.php
```diff
@@ -638,6 +638,28 @@ public function testIgnoreNextLineUnmatched(): void
...
 
+	#[DataProvider('dataTrueAndFalse')]
+	public function testIgnoresWithoutCommentsReported(): void
+	{
+		$expects = [
+			[9, 'variable.undefined'],
+			[12, 'variable.undefined'],
+			[12, 'wrong.id'],
+			[13, 'wrong.id'],
+			[14, 'variable.undefined'],
```
VincentLanglet, line 649: I think this error is kinda opinionated ; you might want to ignore multiple related errors and comment all of them once instead of repeating the same explanation (?)

conf/config.neon
calebdw: > I wonder if the option should apply to phpstan-ignore-line or phpstan-ignore-next-line too, until those tag are deprecated.

No, this was very much intentional---only the `phpstan-ignore` has parsed comments

tests/PHPStan/Analyser/AnalyserTest.php
calebdw: No, that's not currently possible with the `@phpstan-ignore` grammar---every identifier is parsed individually with its own comment.

There's no possible way for us to tell if the last comment should apply to all identifiers or just the last identifier.

If you really want all identifiers to have the same comment then I would just do something like:

```php
/** @phpstan-ignore error.code (some explainantion), error.code (same) */
```

tests/PHPStan/Analyser/data/ignore-no-comments.php
calebdw: > I wasn't even aware comment had to be inside parenthesis.

image


I can add a tooltip to the reported error, however, the example you provide is actually **invalid** and will cause the ignore not to be recognized

src/Parser/RichParser.php
calebdw: It's trying to collapse the keys instead of keeping the array shape

```
  365    Method PHPStan\Parser\RichParser::parseIdentifiers() should return non-empty-list<array{name: string, comment: string|null}> but returns non-empty-list<array<'comment'|'name', string|null>>.  
         🪪  return.type
```

> Another solution could be a format
> array{identifiers: array<string>, comments: array<string>}
> but I'm not sure if it's better...

The comments have to be associated with their identifiers---I'd rather not have to try and match keys between two different arrays, especially when trying to detect if one doesn't have a comment


Ultimately it doesn't matter because the method signature is correct

VincentLanglet (approved):

src/Parser/RichParser.php
staabm: I think the reason is https://github.com/phpstan/phpstan/issues/14081

ondrejmirtes:
Thank you, I really like this!

ondrejmirtes:
Hey, this does not work as I'd expect, e.g. I tried it and it doesn't work.

Example file:

```
class HelloWorld
{

	public function doFoo(): void
	{
		echo []; // @phpstan-ignore echo.nonString
	}

}

```

Example config:

```
parameters:
	reportIgnoresWithoutComments: true
```

I run:

```
bin/phpstan analyse -l 8 test.php -c test.neon
```

And I get No errors.

But the ignore does not have a comment, so it should fail.

calebdw:
Hmm, tests were passing---I'll take a, look

ondrejmirtes:
It only works sometimes.

ondrejmirtes:
An example: I do this here:

```diff
diff --git a/phpstan.neon.dist b/phpstan.neon.dist
index 4da2feb998..2212ddc850 100644
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1 +1,3 @@
 includes: [ build/phpstan.neon ]
+parameters:
+	reportIgnoresWithoutComments: true

```

And I run PHPStan: `make phpstan`

I get some errors from `src/Type/Generic/TemplateTypeTrait.php` and `src/Type/Traits/ConstantScalarTypeTrait.php`.

But there are many more `@phpstan-ignore` examples in phpstan-src that are not reported.

calebdw:
Hmm, it looks like matched errors may have already been removed from the list of errors by the time I try to process them in `src/Analyser/AnalyserResultFinalizer.php`?

ondrejmirtes:
I'm about to do a release. This is a bit undercooked so I'm going to revert this. Please send a PR with the whole feature again. Thanks!

PR #3977, opened 2025-05-02
Check printf parameter types

schlndh:
[Motivation](https://phpstan.org/r/f7af5ab7-b201-4ddf-9080-ffbef44a4ce4): I refactored code to use `BcMath\Number` instead of `float`, and I didn't notice a printf usage with `%f`, which lead to incorrect result. PHPStan didn't complain, because `BcMath\Number` has `__toString()`.

Here is an executed version of the test file: https://3v4l.org/3pXMc

I left some things for possible future PRs to keep things simple:

- Strict mode: enforce `int` for `%d`, `int|float` for `%f`, `__stringandstringable` for `%s`. This may be too strict for default PHPStan, but IMO it would be a good candidate for phpstan-strict-rules.
- Handle multiple possibilities for format string.
- Add support for `vprintf`/`vsprintf`.

### DIFF
--- a/src/Rules/Functions/PrintfHelper.php
+++ b/src/Rules/Functions/PrintfHelper.php
@@ -5,16 +5,28 @@ namespace PHPStan\Rules\Functions;
 use Nette\Utils\Strings;
 use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Php\PhpVersion;
+use PHPStan\Type\ErrorType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\Type;
 use function array_filter;
+use function array_flip;
+use function array_keys;
+use function array_map;
+use function array_reduce;
 use function count;
 use function max;
+use function sort;
 use function sprintf;
 use function strlen;
+use function usort;
 use const PREG_SET_ORDER;
 
+/** @phpstan-type AcceptingTypeString 'strict-int'|'int'|'float'|'string'|'mixed' */
 #[AutowiredService]
 final class PrintfHelper
 {
+
+	private const PRINTF_SPECIFIER_PATTERN = '(?<specifier>[bs%s]|l?[cdeEgfFGouxX])';
 
 	public function __construct(private PhpVersion $phpVersion)
 	{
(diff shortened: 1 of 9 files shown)


src/Rules/Functions/PrintfParameterTypeRule.php
```diff
@@ -0,0 +1,145 @@
...
+					$argType,
+					$scope->isDeclareStrictTypes(),
+				)->result
+			) {
+				continue;
+			}
+
+			$errors[] = RuleErrorBuilder::message(
+				sprintf(
+					'Placeholder #%d of function %s expects %s, %s given',
```
ondrejmirtes: It'd also be really useful to add the actual placeholder (`%f` for example) in the message here.

ondrejmirtes (changes requested):
I'm a bit skeptical here. I think that a lot more types should be allowed in the placeholders. There are probably many combinations allowed by PHP but disallowed by this new rule. For example - you're allowing only int in `%d` but it also allows `float`: https://3v4l.org/XOB5o

ondrejmirtes:
If it's not the case, the message `'Placeholder #1 of function printf expects int, PrintfParamTypes\\FooStringable given'` (the first assertion) is misleading.

schlndh:
I improved the error messages. The rule allows passing `float` to `%d` (and hopefully other meaningfully usable cases as well), it was just not clear from the previous (quite poor) error messages.

Some edge-cases:

- One parameter can be used by multiple placeholders. I decided to generate a separate error for each placeholder, instead of trying to combine them into one (especially since the placeholder is now a part of the error message).
- One placeholder can reference multiple parameters (e.g. `%*.*f` - the `*`s each consume a positional parameter). So for cases where a `*` is used, I clarified whether it's the width, precision or value that's the problem. E.g. `"%*.*f" (width)`.

[Tests (8.2, ubuntu-latest)](https://github.com/phpstan/phpstan-src/actions/runs/17357223173/job/49272016035?pr=3977#logs) fail, but it seems to be unrelated to my PR (`"UnitEnum" could not be found in the located source`).

src/Rules/Functions/PrintfHelper.php
```diff
@@ -44,34 +55,83 @@ private function getPlaceholdersCount(string $specifiersPattern, string $format)
...
-			} else {
-				$maxOrdinaryNumber++;
-			}
+			$parsedPlaceholders[] = new PrintfPlaceholder(
+				sprintf('"%s"', $placeholder[0]) . ($showValueSuffix ? ' (value)' : ''),
+				isset($placeholder['position']) && $placeholder['position'] !== ''
+					? $placeholder['position'] - 1
+					: $parameterIdx++,
+				$placeholderNumber,
+				$this->getAcceptingTypeBySpecifier($placeholder['specifier'] ?? ''),
```
ondrejmirtes, line 98: This says "accepting" but the property it's being passed to is "accepted".

"Accepting" and "accepted" are opposite sides of the equation so please update it to reflect what side is this.

src/Rules/Functions/PrintfPlaceholder.php
```diff
@@ -0,0 +1,40 @@
...
+			case 'int':
+				return ! $argumentType->toInteger() instanceof ErrorType;
+			case 'float':
+				return ! $argumentType->toFloat() instanceof ErrorType;
+			// The function signature already limits the parameters to stringable types, so there's
+			// no point in checking string again here.
+			case 'string':
+			case 'mixed':
+			default:
+				return true;
```
ondrejmirtes, line 36: I'd rather be if the `default` was removed (PHPStan should be fine with it, but if it's not, the default should throw ShouldNotHappenException).

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #3978, opened 2025-05-02
Only set return native type from null return type when declaring class is built in on __toString()

samsonasik:
Start from PHPStan 2.1.14, in `Rector`, we got notice when downgrade code:

```php
class ParentClassWithToStringMixedReturn
{
    public function __toString()
    {
        return 'a';
    }
}

class WithParentMixedReturn extends ParentClassWithToStringMixedReturn
{
    public function __toString()
    {
        return 'value';
    }
}

echo (new WithParentMixedReturn());
```

This add `string` return to `WithParentMixedReturn::__toString()` see https://3v4l.org/IrE2q#v7.4.33

The issue is when **code consumed by user**: https://3v4l.org/kdcEh#v7.4.33 it got error:

```
Fatal error: Declaration of Consumer::__toString() must be compatible with WithParentMixedReturn::__toString(): string in /in/kdcEh on line 21
```

On our rector-src code, I patch there with use our AstResolver to specify on `__toString` when no return and non-built in

- https://github.com/rectorphp/rector-src/pull/6879

to return `MixedType`

@TomasVotruba if this accepted, we may can rollback to original implementation 👍

### DIFF
--- a/src/Reflection/Php/PhpMethodReflection.php
+++ b/src/Reflection/Php/PhpMethodReflection.php
@@ -330,11 +330,9 @@ final class PhpMethodReflection implements ExtendedMethodReflection
 					return $this->nativeReturnType = new VoidType();
 				}
 				if ($name === '__tostring') {
-					if ($this->declaringClass->isBuiltin()) {
-						return $this->nativeReturnType = new StringType();
-					} else {
-						return $this->nativeReturnType = new MixedType();
-					}
+					return $this->nativeReturnType = $this->declaringClass->isBuiltin()
+						? new StringType()
+						: new MixedType();
 				}
 				if ($name === '__isset') {
 					return $this->nativeReturnType = new BooleanType();
--- a/src/Reflection/Php/PhpMethodReflection.php
+++ b/src/Reflection/Php/PhpMethodReflection.php
@@ -324,7 +324,7 @@ final class PhpMethodReflection implements ExtendedMethodReflection
 	{
 		if ($this->nativeReturnType === null) {
 			$returnType = $this->reflection->getReturnType();
-			if ($returnType === null) {
+			if ($returnType === null && $this->declaringClass->isBuiltin()) {
 				$name = strtolower($this->getName());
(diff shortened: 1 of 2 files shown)


samsonasik:
@ondrejmirtes ready for review, not sure about the php 7.4 notice:

```
There was 1 failure:

1) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "tests/PHPStan/Analyser/nsrt/bug-2471.php" ('/home/runner/work/phpstan-src...71.php')
Failed assertions in /home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Analyser/nsrt/bug-2471.php:

Line 26:
Expected: array<string, null>
Actual:   array<null>
```

https://github.com/phpstan/phpstan-src/actions/runs/14802288573/job/41563600230?pr=3978#step:6:226

since other php versions green, could you give me hint to make it green? Thank you.

src/Reflection/Php/PhpMethodReflection.php
```diff
@@ -330,7 +330,13 @@ private function getNativeReturnType(): Type
 					return $this->nativeReturnType = new VoidType();
 				}
 				if ($name === '__tostring') {
-					return $this->nativeReturnType = new StringType();
+					return $this->nativeReturnType = $this->declaringClass->isBuiltin()
```
ondrejmirtes, line 333: 1) How is `__toString()` different from other methods handled by this method?
2) Personally I'd like a test + revert of the commit that changed this (https://github.com/phpstan/phpstan-src/commit/3854cbc5748a7cb51ee0b86ceffe29bd0564bc98) instead of this change.

ondrejmirtes (changes requested):

src/Reflection/Php/PhpMethodReflection.php
samsonasik: 1. we found by case, for current case is `__toString()` so don't know what other needs
2. yes, I think revert is better for this case, as native is native on overridden method.

samsonasik: https://github.com/phpstan/phpstan-src/pull/3979

PR #3971, opened 2025-04-29
non-falsy-string cannot be converted to 0

staabm:
small drive by optimization realized while working on https://github.com/phpstan/phpstan-src/pull/3965

PR #3980, opened 2025-05-04
Fix DateTime::format('u') return type II

mvorisek:
fix https://github.com/phpstan/phpstan/issues/12965 and https://github.com/phpstan/phpstan-src/pull/3919

### DIFF
--- a/src/Type/Php/DateFunctionReturnTypeHelper.php
+++ b/src/Type/Php/DateFunctionReturnTypeHelper.php
@@ -76,11 +76,11 @@ final class DateFunctionReturnTypeHelper
 				return $this->buildNumericRangeType(0, 1, false);
 			case 'u':
 				return $useMicrosec
-					? new IntersectionType([new StringType(), new AccessoryNonFalsyStringType(), new AccessoryNumericStringType()])
+					? new IntersectionType([new StringType(), new AccessoryNumericStringType()])
 					: new ConstantStringType('000000');
 			case 'v':
 				return $useMicrosec
-					? new IntersectionType([new StringType(), new AccessoryNonFalsyStringType(), new AccessoryNumericStringType()])
+					? new IntersectionType([new StringType(), new AccessoryNumericStringType()])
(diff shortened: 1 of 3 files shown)


mvorisek:
@ondrejmirtes can this be merged?

VincentLanglet:
Can it be at least a non-empty-string @mvorisek ?

mvorisek:
> Can it be at least a non-empty-string @mvorisek ?

That is implied by "numeric-string" already.

mvorisek:
@ondrejmirtes can I have your review here?

ondrejmirtes:
Do these ever return a single `'0'`? https://3v4l.org/AEu9V

mvorisek:
Good point @ondrejmirtes!

So my assumption this issue is caused by https://github.com/phpstan/phpstan-src/pull/3971 was correct.

@staabm in https://github.com/phpstan/phpstan-src/pull/3971 you claim "non-falsy-string cannot be converted to 0" but that is not true. `'00'` is `non-falsy-string`, but after `(int)` cast it can be `0`, ie. falsy - https://phpstan.org/r/371ef912-16a9-485d-82b5-933f8d0ec799

staabm:
you are right, that I missed `(int) '00'`. please create a reverting commit and add a corresponding comment at the test-expectations so don't do the same error again.

thanks

mvorisek:
Please submit from your side. Thanks.

ondrejmirtes:
@mvorisek Don't ask other people to do work for you.

mvorisek:
@ondrejmirtes please read the 1 post above, ok?

ondrejmirtes:
I read it, but it doesn't mean @staabm has to fix it. Your demeanor is infuriating to a lot of people and you should understand that. You've been blocked because of that multiple times.

VincentLanglet:
I opened the revert https://github.com/phpstan/phpstan-src/pull/3997 if it can help.

Not sure which comment you want me to add @staabm

PR #3983, opened 2025-05-08
Fix callable-string must be non-empty-string

aivchen:
Fixes https://github.com/phpstan/phpstan/issues/12979

### DIFF
--- a/src/Type/Accessory/AccessoryNonEmptyStringType.php
+++ b/src/Type/Accessory/AccessoryNonEmptyStringType.php
@@ -76,6 +76,9 @@ class AccessoryNonEmptyStringType implements CompoundType, AccessoryType
 
 	public function acceptsWithReason(Type $type, bool $strictTypes): AcceptsResult
 	{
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
Instead of this approach, please explore modifying `IntersectionType::isNonEmptyString()` based on `$this->isString()->yes()` and `$this->isCallable()->yes()`. Thanks.

staabm:
might also be worth checking whether we have a similar problem with `callable-array` and `non-empty-array`

aivchen:
> Instead of this approach, please explore modifying `IntersectionType::isNonEmptyString()` based on `$this->isString()->yes()` and `$this->isCallable()->yes()`. Thanks.

@ondrejmirtes I tried to implement your suggestion. Have a look please.

aivchen:
> might also be worth checking whether we have a similar problem with `callable-array` and `non-empty-array`

Let's do it in another issue, please. As I see the logic for arrays is completely different.

ondrejmirtes:
Thank you!

PR #3984, opened 2025-05-11
Support iterable as template type

RobertMe:
Fixes phpstan/phpstan#12214

Note: I'm targeting 1.12 branch. But I'm not sure whether that is allowed (in this instance). But I tested merging upwards and that seems to work fine (-ish, just a merge conflict with the added unit test methods as more methods are added in the test class).

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1043,6 +1043,11 @@ parameters:
 		-
 			message: "#^Doing instanceof PHPStan\\\\Type\\\\IntersectionType is error\\-prone and deprecated\\.$#"
 			count: 3
+			path: src/Type/Generic/TemplateIterableType.php
+
+		-
+			message: "#^Doing instanceof PHPStan\\\\Type\\\\IntersectionType is error\\-prone and deprecated\\.$#"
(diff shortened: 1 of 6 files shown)


RobertMe:
Static analysis error is in line with other (baselined) errors for which I followed the same logic. So probably irrelevant (/can be baselined, not sure if I'm expected to do that or it happens automatically one way or another after / during merging).

Furthermore locally I also get errors in `TemplateTypeTrait` (which is used in the new class). So not sure why those aren't reported in the CI checks.

ondrejmirtes:
1) Please baseline the errors.
2) Please move the test to a different place in the same file so that no conflict happens.

Thanks!

RobertMe:
Updated and tested a merge upwards again. Sadly though it still results in a conflict. This time in the baseline (if I'm seeing it correctly due to the quote change (`"` on 1.x and `'` on 2.x) and identifiers being added in 2.x).

ondrejmirtes:
Baseline is fine, I can regenerate that.

ondrejmirtes:
Thank you!

PR #3987, opened 2025-05-12
Validate checkedExceptionClasses config

janedbal:


src/Rules/Exceptions/DefaultExceptionTypeResolver.php
```diff
@@ -27,6 +28,16 @@ public function __construct(
 		private array $checkedExceptionClasses,
 	)
 	{
+		foreach ($this->checkedExceptionClasses as $checkedExceptionClass) {
+			if (!$this->reflectionProvider->hasClass($checkedExceptionClass)) {
+				throw new ShouldNotHappenException('Class ' . $checkedExceptionClass . ' used in \'checkedExceptionClasses\' does not exist.');
```
janedbal: Not sure if this is the preferred way of config validation.

src/Rules/Exceptions/DefaultExceptionTypeResolver.php
```diff
@@ -27,6 +28,16 @@ public function __construct(
...
+			}
+		}
+		foreach ($this->uncheckedExceptionClasses as $uncheckedExceptionClass) {
+			if (!$this->reflectionProvider->hasClass($uncheckedExceptionClass)) {
+				throw new ShouldNotHappenException('Class ' . $uncheckedExceptionClass . ' used in \'uncheckedExceptionClasses\' does not exist.');
+			}
+		}
```
janedbal: Do you want that behind featureFlag / bleedingEdge? I think this will be pretty rare.

src/Rules/Exceptions/DefaultExceptionTypeResolver.php
ondrejmirtes: This is a BC break. Also I'd prefer it if happened in ContainerFactory. There's already `validateParameters` method.

ondrejmirtes: So yeah, please put this only in bleeding edge.

ondrejmirtes (changes requested):

src/Rules/Exceptions/DefaultExceptionTypeResolver.php
janedbal: Changed, see last commit.

src/DependencyInjection/ContainerFactory.php
```diff
@@ -323,6 +324,34 @@ private function validateParameters(array $parameters, array $parametersSchema):
 		if ($phpVersion['max'] < $phpVersion['min']) {
 			throw new InvalidPhpVersionException('Invalid PHP version range: phpVersion.max should be greater or equal to phpVersion.min.');
 		}
+
+		if (array_key_exists('featureToggles', $parameters) && is_array($parameters['featureToggles']) && (bool) $parameters['featureToggles']['bleedingEdge']) {
+			$this->validateExceptionClasses($parameters);
```
ondrejmirtes: You can put these conditions right next to:

```
		$container = $configurator->createContainer()->getByType(Container::class);
		$this->validateParameters($container->getParameters(), $projectConfig['parametersSchema']);
```

And pass ReflectionProvider properly from `$container` instead of obtaining it statically.

ondrejmirtes: Meaning do not call `validateExceptionClasses` from inside `validateParameters`, but instead from `create`.

ondrejmirtes (changes requested):
And fix the conflicts please 😊

src/DependencyInjection/ContainerFactory.php
janedbal: Adjusted based on your comments.

PR #3988, opened 2025-05-12
Fix inconsistent analysis on case-insensitive filesystems

staabm:
closes https://github.com/phpstan/phpstan/issues/12972

since on macos filesystem paths are case-insensitive before this PR we ran into the following error:
```
➜  bug-12972 git:(bug12972) ../../bin/phpstan analyse --debug 

 ! [NOTE] The Xdebug PHP extension is active, but "--xdebug" is not used.                                               
 !        The process was restarted and it will not halt at breakpoints.                                                
 !        Use "--xdebug" if you want to halt at breakpoints.                                                            

Note: Using configuration file /Users/m.staab/Documents/GitHub/phpstan-src/e2e/bug-12972/phpstan.dist.neon.
/Users/m.staab/Documents/GitHub/phpstan-src/e2e/bug-12972/src/folder/file2.php
/Users/m.staab/Documents/GitHub/phpstan-src/e2e/bug-12972/src/other/file.php
 ------ --------------------------------------------------------------------------------------------- 
  Line   other/file.php                                                                               
 ------ --------------------------------------------------------------------------------------------- 
  6      Method other12972\MyClass::doSomething() should return int but return statement is missing.  
 ------ --------------------------------------------------------------------------------------------- 
```

-> PHPStan is using the simple-parser instead of the rich-parser as it does not handle the analyzed path case-insentive [as macOS would do](https://github.com/phpstan/phpstan/issues/12972#issuecomment-2851171317) (same for windows)

see

```sh
➜  phpstan-src git:(bug12972) ls -l e2e/bug-12972/src/                 
total 0
drwxr-xr-x@ 3 m.staab  staff  96 May 12 14:59 folder
drwxr-xr-x@ 3 m.staab  staff  96 May 12 14:52 other
➜  phpstan-src git:(bug12972) ls e2e/bug-12972/src/other/file.php                                                                           
e2e/bug-12972/src/other/file.php
➜  phpstan-src git:(bug12972) ls e2e/bug-12972/src/OTHER/file.php
e2e/bug-12972/src/OTHER/file.php
➜  phpstan-src git:(bug12972) php -r 'var_dump(is_file("e2e/bug-12972/src/OTHER/file.php"));'
bool(true)
➜  phpstan-src git:(bug12972) php -r 'var_dump(is_file("e2e/bug-12972/src/other/file.php"));'
bool(true)
➜  phpstan-src git:(bug12972) php -v                                                         
PHP 8.4.7 (cli) (built: May  6 2025 12:31:58) (NTS)
Copyright (c) The PHP Group
Built by Homebrew
Zend Engine v4.4.7, Copyright (c) Zend Technologies
```

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -308,7 +308,7 @@ jobs:
 
   e2e-tests:
     name: "E2E tests"
-    runs-on: "ubuntu-latest"
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 60
 
     strategy:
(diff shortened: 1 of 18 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -287,84 +287,116 @@ jobs:
...
     steps:
       - name: "Checkout"
         uses: actions/checkout@v4
 
+      - name: "Install GNU Patch on macOS" # see https://github.com/cweagans/composer-patches/issues/326
+        if: runner.os == 'macOS'
+        run: |
+          brew install gpatch
+          echo "/opt/homebrew/opt/gpatch/libexec/gnubin" >> $GITHUB_PATH
+
```
staabm, line 440: on macOS we need a GNU compatible `patch` command, as the apple-shipped `patch` command cannot apply our patch files:

```
Error: Cannot apply patch 0 (patches/cloudflare-ca.patch)!

In Patches.php line 331:
                                                       
  Cannot apply patch 0 (patches/cloudflare-ca.patch)!  
```

src/Analyser/AnalysedFilesResolver.php
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	public function isInAnalyzedFiles(string $file): bool
+	{
+		if ($this->caseInsensitiveFilesystem) {
+			$file = strtolower($file);
+		}
+
+		return isset($this->analysedFiles[$file]);
+	}
```
staabm, line 42: actual fix is here, everything else are adjustments to make sure this new logic is used all over the place

src/Analyser/AnalysedFilesResolver.php
```diff
@@ -0,0 +1,47 @@
...
+	private array $analysedFiles;
+
+	private bool $caseInsensitiveFilesystem;
+
+	/**
+	 * @param string[] $files
+	 */
+	public function __construct(array $files = [])
+	{
+		$this->caseInsensitiveFilesystem = PHP_OS_FAMILY === 'Darwin';
```
staabm: instead of hardcoding this for macOS, we could also use a feature detect (which requires a bit of IO)

.github/workflows/e2e-tests.yml
```diff
@@ -317,84 +317,127 @@ jobs:
...
+              ../bashunit -a contains 'e2e/bug-12972/src/OTHER/file.php' "$OUTPUT"
+            os: "ubuntu-latest"
+          - script: |
+              cd e2e/bug-12972
+              ../../bin/phpstan analyze
+            os: "macos-latest"
+          - script: |
+              cd e2e/bug-12972
+              ../../bin/phpstan analyze
+            os: "windows-latest"
```
ondrejmirtes, line 429: I don't get it. Why do you expect different outcomes on Linux vs. the rest?

ondrejmirtes (changes requested):

.github/workflows/e2e-tests.yml
staabm: because linux does not have a case-insensitive filesystem.

ondrejmirtes: So you have a project that deliberately works only on case-insensitive filesystems?

staabm: the test mimics what php-src will do at runtime.
it will happily accept a different-cased path on windows/macos but doesn't on linux (acutally depending on the filesystem case-sensitivy not the OS itself).

ondrejmirtes: I'd like if the e2e test mimiced how your project actually worked. This just looks like PHPStan fixes macOS but breaks Linux.

staabm: ok, lets look into https://github.com/phpstan/phpstan-src/pull/4003 which mimics our project setup more closely

PR #3990, opened 2025-05-13
unset.possiblyHookedProperty is reported for properties annotated @final

lpd-au:
Resolve https://github.com/phpstan/phpstan/issues/12999

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4373,7 +4373,7 @@ final class NodeScopeResolver
 		if (!$propertyReflection->hasHook($hookName)) {
 			if (
 				$propertyReflection->isPrivate()
-				|| $propertyReflection->isFinalByKeyword()->yes()
+				|| $propertyReflection->isFinal()->yes()
 				|| $declaringClass->isFinal()
 			) {
 				return [];
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4373,7 +4373,7 @@ final class NodeScopeResolver
 		if (!$propertyReflection->hasHook($hookName)) {
 			if (
 				$propertyReflection->isPrivate()
-				|| $propertyReflection->isFinal()->yes()
+				|| $propertyReflection->isFinalByKeyword()->yes()
 				|| $declaringClass->isFinal()
 			) {
 				return [];
--- a/src/Reflection/ExtendedPropertyReflection.php
+++ b/src/Reflection/ExtendedPropertyReflection.php
@@ -38,7 +38,7 @@ interface ExtendedPropertyReflection extends PropertyReflection
 
 	public function isAbstract(): TrinaryLogic;
 
-	public function isFinal(): TrinaryLogic;
+	public function isFinalByKeyword(): TrinaryLogic;
 
 	public function isVirtual(): TrinaryLogic;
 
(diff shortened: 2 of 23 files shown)


ondrejmirtes:
For classes there's `isFinal` and `isFinalByKeyword`. I'd like to follow the same convention here.

Also, there should be a new test for OverridingPropertyRule checking when a `@final` property is overriden. Also, OverridingMethodRule changes the error message and identifier based on whether it's `final` or `@final` so I'd like for OverridingPropertyRule to do the same.

Thanks.

src/Analyser/NodeScopeResolver.php
```diff
@@ -4381,7 +4381,7 @@ private function getThrowPointsFromPropertyHook(
 		if (!$propertyReflection->hasHook($hookName)) {
 			if (
 				$propertyReflection->isPrivate()
-				|| $propertyReflection->isFinal()->yes()
+				|| $propertyReflection->isFinalByKeyword()->yes()
```
ondrejmirtes: I think it'd be fine to keep `isFinal` here, right?

src/Reflection/ExtendedPropertyReflection.php
```diff
@@ -38,7 +38,7 @@ public function getNativeType(): Type;
 
 	public function isAbstract(): TrinaryLogic;
 
-	public function isFinal(): TrinaryLogic;
+	public function isFinalByKeyword(): TrinaryLogic;
```
ondrejmirtes, line 41: You can't rename a method, that's a BC break. You have to add a method.

isFinal will return true if there's a native `final` or PHPDoc `@final`.

isFinalByKeyword will return true if there's a native `final`.

ondrejmirtes (changes requested):

src/Reflection/WrappedExtendedPropertyReflection.php
```diff
@@ -109,6 +109,11 @@ public function isAbstract(): TrinaryLogic
...
 	}
 
+	public function isFinalByKeyword(): TrinaryLogic
+	{
+		return TrinaryLogic::createNo();
+	}
+
 	public function isFinal(): TrinaryLogic
 	{
 		return TrinaryLogic::createNo();
```
lpd-au, line 119: I don't really know what this class represents, hope it's okay

src/Rules/Properties/OverridingPropertyRule.php
```diff
@@ -145,6 +145,16 @@ public function processNode(Node $node, Scope $scope): array
...
 				->nonIgnorable()
 				->build();
+		} elseif ($prototype->isFinal()->yes()) {
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'Property %s::$%s overrides @final property %s::$%s.',
+				$classReflection->getDisplayName(),
+				$node->getName(),
+				$prototype->getDeclaringClass()->getDisplayName(),
+				$node->getName(),
+			))->identifier('property.parentPropertyAnnotatedFinal')
```
ondrejmirtes: Please follow the same convention as in OverridingMethodRule. There's `method.parentMethodFinal` and `method.parentMethodFinalByPhpDoc` so here it should be:

* property.parentPropertyFinal
* property.parentPropertyFinalByPhpDoc

Also, this error has to be ignorable.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #3991, opened 2025-05-13
Fix IterableType->equals(TemplateIterableType)

staabm:
I think this change is necessary, as since https://github.com/phpstan/phpstan-src/commit/8368b0c27e2ae919e1cb874d18d19b2e14a43723 we have `TemplateIterableType extends IterableType`, which we hadn't before.

not sure how to properly test it.


ondrejmirtes:
Thank you!

PR #3995, opened 2025-05-16
Bleeding edge - improve comparison against BcMath\Number

schlndh:
Fixes: https://github.com/phpstan/phpstan/issues/13001

Behavior test: https://3v4l.org/K7THf

### DIFF
--- a/src/Rules/Operators/InvalidComparisonOperationRule.php
+++ b/src/Rules/Operators/InvalidComparisonOperationRule.php
@@ -4,7 +4,10 @@ namespace PHPStan\Rules\Operators;
 
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
+use PHPStan\DependencyInjection\AutowiredParameter;
 use PHPStan\DependencyInjection\RegisteredRule;
+use PHPStan\DependencyInjection\Type\OperatorTypeSpecifyingExtensionRegistryProvider;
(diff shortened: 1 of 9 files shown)


src/Rules/Operators/InvalidComparisonOperationRule.php
```diff
@@ -53,6 +59,22 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		$result = $this->operatorTypeSpecifyingExtensionRegistryProvider->getRegistry()->callOperatorTypeSpecifyingExtensions(
+			$node,
+			$scope->getType($node->left),
+			$scope->getType($node->right),
```
schlndh, line 69: This might cause some issues below level 7, but I'm not sure how to improve it. The operator type specifying extensions don't support union well anyway.

ondrejmirtes:
Please fix the conflicts. Meanwhile the way how rules are registered have changed. Instead of entries in config.neon, parameters are now autowired with an attribute like this: `#[AutowiredParameter(ref: '%featureToggles.stricterFunctionMap%')]`

schlndh:
@ondrejmirtes Done. Thank you for the tip about AutowiredParameter (one more thing that took me a few minutes was that it's also necessary to run `composer dump-autoload` after adding it).

ondrejmirtes:
Thank you!

PR #3997, opened 2025-05-16
Revert "non-falsy-string cannot be converted to 0"

VincentLanglet:
This reverts commit f615b1a39a70d9ec9e49963474c7f9f9577cf88b.

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/non-falsy-string.php
+++ b/tests/PHPStan/Analyser/nsrt/non-falsy-string.php
@@ -11,7 +11,7 @@ class Foo {
 	 * @param truthy-string $truthyString
 	 */
 	public function bar($nonFalseyString, $truthyString) {
-		assertType('int', (int) $nonFalseyString);
+		assertType('int', (int) $nonFalseyString); // Do not remove `0` since `(int) '00'` is still `0`
 		// truthy-string is an alias for non-falsy-string
 		assertType('non-falsy-string', $truthyString);
 	}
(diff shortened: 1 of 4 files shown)


staabm (approved):
thanks :+1:

tests/PHPStan/Analyser/nsrt/non-falsy-string.php
```diff
@@ -11,7 +11,7 @@ class Foo {
 	 * @param truthy-string $truthyString
 	 */
 	public function bar($nonFalseyString, $truthyString) {
-		assertType('int<min, -1>|int<1, max>', (int) $nonFalseyString);
+		assertType('int', (int) $nonFalseyString);
```
mvorisek:
```suggestion
		assertType('int', (int) $nonFalseyString); // Do not remove `0` since `(int) '00'` is still `0`.
```

I prefer to have the comment also in at least one test in case the impl. is possibly replaced in the future.

And thank you for this PR! ❤

VincentLanglet:
Friendly ping @ondrejmirtes

This PR might be forgotten since I assume it's a quick review.
It revert https://github.com/phpstan/phpstan-src/commit/f615b1a39a70d9ec9e49963474c7f9f9577cf88b as asked in https://github.com/phpstan/phpstan-src/pull/3980

`(int) '00'` can still be `0`.

staabm:
> (int) '00' can still be 0.

Please add a test for this case so we see it next time earlier

VincentLanglet:
> > (int) '00' can still be 0.
> 
> Please add a test for this case so we see it next time earlier

How do you want to add such test ?

There is already a test checking that (int) non-falsy-string is an int (and I added a comment).

If I add a test for the special case '00' it will be useless since the methode used will be from ConstantStringType and not AccessoryNonFalsyStringType

staabm:
I see, thanks for looking into it.

ondrejmirtes:
Thank you.

PR #3998, opened 2025-05-18
Handle class without namespace for internal rule

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13041


ondrejmirtes:
Thank you!

PR #1544, opened 2022-07-25
`str_split` returns an empty array since PHP 8.2

herndlm:
Refs: https://github.com/php/php-src/pull/8945

See also https://3v4l.org/XfZO4 and https://3v4l.org/XfZO4/rfc#vgit.master

I never touched the delta files, is this correct like this? One of the tests at least seems to verify it. And the comments in that file are obviously outdated, but I did not touch them because they're like that in other delta files as well.

UPDATE: sorry, was doing too many things at once and forgot to cleanly run make again.. I thought I could avoid nested ternaries with the match expression 😅

PR #1496, opened 2022-07-06
Fix handling of `str_split` / `mb_str_split` string arg compound types + default return type for `mb_str_split`

herndlm:
Closes https://github.com/phpstan/phpstan/issues/7580

There were 2 issues basically which is why I split this into 2 commits
- `str_split` / `mb_str_split` did not handle compound types for the string type correctly (e.g. union of constant strings)
- `mb_str_split` can actually return an empty array. This was adapted in https://github.com/phpstan/phpstan-src/pull/633 which is based on a psalm adaption, but psalm actually reverted it back in https://github.com/vimeo/psalm/pull/6493 because `str_split` and `mb_str_split` differ in behaviour there. See also https://3v4l.org/r6sMe

`StrSplitFunctionReturnTypeExtension` could need a bit more cleanup, but I wanted to avoid rewriting it here.. Maybe I'll adapt some more in follow-ups.

PR #3999, opened 2025-05-18
Improve StrSplit returnType

VincentLanglet:
Hi @herndlm 

I saw you worked multiple times on this extension with
- https://github.com/phpstan/phpstan-src/pull/1496
- https://github.com/phpstan/phpstan-src/pull/1544

### DIFF
--- a/src/Type/Php/StrSplitFunctionReturnTypeExtension.php
+++ b/src/Type/Php/StrSplitFunctionReturnTypeExtension.php
@@ -10,6 +10,7 @@ use PHPStan\Reflection\FunctionReflection;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
+use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantArrayType;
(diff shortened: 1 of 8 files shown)


VincentLanglet:
If it's ok, I opened https://github.com/phpstan/phpstan-src/pull/4000 first

src/Type/Php/StrSplitFunctionReturnTypeExtension.php
```diff
@@ -67,47 +71,70 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+			isset($splitLength)
+			&& ($functionReflection->getName() === 'str_split' || $encoding !== null)
+		) {
+			$constantStrings = $stringType->getConstantStrings();
+			if (count($constantStrings) > 0) {
+				$results = [];
+				foreach ($constantStrings as $constantString) {
+					$value = $constantString->getValue();
+
+					if ($encoding === null && $value === '') {
```
VincentLanglet, line 98: This logic exists to avoid the same issue than https://github.com/phpstan/phpstan/issues/13129

src/Php/PhpVersion.php
```diff
@@ -236,7 +236,7 @@ public function deprecatesDynamicProperties(): bool
 		return $this->versionId >= 80200;
 	}
 
-	public function strSplitReturnsEmptyArray(): bool
+	public function strSplitReturnsEmptyArrayOnEmptyString(): bool
```
ondrejmirtes: This class is `@api`, you can't rename methods here.

ondrejmirtes (changes requested):
This one could also be only for 2.1.x.

VincentLanglet:
@ondrejmirtes The HasPropertyTypeTest::testIsSuperTypeOf with data set #4 seems random,
On this PR it fails on PHP 8.4 ubuntu, 
But on another PR it onl fails on PHP 8.3 ubuntu
https://github.com/phpstan/phpstan-src/actions/runs/16353254353/job/46205186388?pr=3906
And it changes sometimes when I repush.

So I don't think, it's related to the PR ; this is now ready.

ondrejmirtes:
Thank you!

PR #4002, opened 2025-05-20
Add Grouped Error Formatter

FeBe95:
## Description

This PR introduces a new error formatter that groups all errors by their identifier. Groups are sorted in descending order based on the number of occurrences. The output format is a hybrid of the raw and table error formatters.

All error lines are clickable for easy code navigation. As with the table error formatter, the `file:line` format can be customized via the `editorUrl` and `editorUrlTitle` parameters.

## Usage
```sh
vendor/bin/phpstan analyse --error-format=grouped
```

## Use Case
This formatter is particularly useful when working with a large number of errors, for example, when introducing PHPStan to an existing codebase. Grouping by error identifier instead of by file helps speed up the review and fixing process.

## Example Output
```
Note: Using configuration file C:\projects\test-project\phpstan.neon.
 123/123 [============================] 100%

[property.notFound] (4x):
    - C:\projects\test-project\Bar.php:21: Access to an undefined property App\Models\MyModel::$email.
    - C:\projects\test-project\Bar.php:71: Access to an undefined property App\Models\MyModel::$deleted_at.
    - C:\projects\test-project\Foo.php:20: Access to an undefined property App\Models\MyModel::$id.
    - C:\projects\test-project\Foo.php:55: Access to an undefined property App\Models\MyModel::$name.

[method.notFound] (3x):
    - C:\projects\test-project\Bar.php:25: Call to an undefined method App\Helper\SdkClientV3::delete().
    - C:\projects\test-project\Foo.php:276: Call to an undefined method App\Helper\SdkClientV3::update().
    - C:\projects\test-project\Foo.php:760: Call to an undefined method App\Helper\SdkClientV3::query().

[phpDoc.parseError] (2x):
    - C:\projects\test-project\Foo.php:13: PHPDoc tag @param has invalid value (int countryId ,): Unexpected token "countryId", expected variable at offset 93 on line 4
    - C:\projects\test-project\Foo.php:14: PHPDoc tag @param has invalid value (int productId ,): Unexpected token "productId", expected variable at offset 130 on line 5

[arguments.count] (1x):
    - C:\projects\test-project\Bar.php:46: Method Illuminate\Database\Eloquent\Builder<App\Models\MyModel>::orWhereNull() invoked with 2 parameters, 1 required.

                                                                                                                        
 [ERROR] Found 10 errors                                                                                                
                                                                                                                        
```

## Screenshot

![grafik](https://github.com/user-attachments/assets/05116983-bf7d-4370-a8a9-da158f375d12)


mvorisek:
How is this helpful, should't the grouping be done on filename basis?

The reason is many different errors are related, thus errors for LoC close to each other should be reported together.

FeBe95:
We introduced PHPStan to our codebase a few weeks ago. We started with level 0 and added all existing errors to the baseline file.

Then our workflow is as follows:

1. Delete all ignored errors from the baseline
2. Use this grouped error formatter for error analysis
3. Assign one developer to each error type
4. Fix the errors one at a time
5. Increase the level by 1
6. Repeat

> Shouldn't the grouping be done by filename?

We have found it much easier to do research on one specific error type at a time. Often, you can find an answer right away on how to fix the underlying bugs, and then apply it to all similar ones.

PR #4003, opened 2025-05-21
Reproduce autoloader bug

staabm:
this test demonstrates how PHPStan analyzes code differently than it would happen at runtime.

---

runtime: run `php real-world.php`
you see the program runs, without the custom autoloader of `e2e/bug-12972b/autoloader.php` beeing invoked.
all involved classes are autoloaded by the composer class loader. the custom class loader is not invoked for `\other12972\MyClass::class`.

---

static analysis time: run `../../bin/phpstan analyze`

you see PHPStan invokes the custom class loader for class `\other12972\MyClass::class`, which does not happen at regular runtime. the composer class loader is not invoked for `\other12972\MyClass::class` (like it is at runtime).

because of the difference in autoloading functions order in runtime vs. analysis time we see errors which cannot happen at runtime - see https://github.com/phpstan/phpstan/issues/12972

### DIFF
--- a/e2e/bug-12972b/autoloader.php
+++ b/e2e/bug-12972b/autoloader.php
@@ -1,0 +1,9 @@
+<?php
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+spl_autoload_register(function($class) {
+	if ($class === \other12972\MyClass::class) {
+		throw new LogicException('this should not happen');
+	}
+});
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -566,7 +566,10 @@ jobs:
                 echo "FAIL: tmp was created by a worker, meaning sys_temp_dir='tmp~1' was incorrectly evaluated to 'tmp'"
                 exit 1
               fi
-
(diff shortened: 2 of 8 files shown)


e2e/bug-12972b/autoloader.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+spl_autoload_register(function($class) {
+	if ($class === \other12972\MyClass::class) {
+		throw new LogicException('this should not happen');
```
staabm, line 7: in our real world project this autoloader is scanning some paths in the filesystem (similar to how a include-path worked in the past / or how the %PATH% variable works in a shell).

since it scans the filesystem it finds some files and includes them. this in turn leads to the problem described in https://github.com/phpstan/phpstan/issues/12972 -> so invoking the autoloader leads side-effects, as files get included, which has impact on `ReflectionClass->getFileName()` invoked later on in PHPStan

staabm:
replaced by https://github.com/phpstan/phpstan-src/pull/6069

PR #4007, opened 2025-05-22
Allow getenv(null) for PHP > 8

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13065


tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
```diff
@@ -2102,4 +2102,18 @@ public function testBug12499(): void
 		$this->analyse([__DIR__ . '/data/bug-12499.php'], []);
 	}
 
+	public function testBug13065(): void
+	{
+		if (PHP_VERSION_ID < 80000) {
+			$this->analyse([__DIR__ . '/data/bug-13065.php'], [
+				[
+					'Parameter #1 $varname of function getenv expects string, null given.',
```
ondrejmirtes: You can just conditionally define `$errors` and still call `analyse` one time.

src/Type/Php/GetenvFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,49 @@
...
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\UnionType;
+use function count;
+
+final class GetenvFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
```
ondrejmirtes: Could you try to solve this only in functionMap delta files and not via an extension?

ondrejmirtes (changes requested):

src/Type/Php/GetenvFunctionReturnTypeExtension.php
VincentLanglet: I dunno how.

I tried https://github.com/phpstan/phpstan-src/pull/4007/commits/26db34a8ea005476e067fd2fe79bd3596b4935e3 but it doesn't work.

tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -57,6 +57,12 @@ private static function findTestFiles(): iterable
 			yield __DIR__ . '/data/explode-php80.php';
 		}
 
+		if (PHP_VERSION_ID < 80000) {
+			yield __DIR__ . '/data/getenv-php74.php';
+		} else {
+			yield __DIR__ . '/data/getenv-php80.php';
```
ondrejmirtes: This doesn't need to be in NodeScopeResolverTest. Move the files to nsrt/ and label them with proper `// lint` conditions.

tests/PHPStan/Analyser/data/getenv-php74.php
```diff
@@ -0,0 +1,17 @@
...
+
+class Foo
+{
+
+	public function test()
+	{
+		assertType('string|false', getenv(null));
+		assertType('array<string, string>', getenv());
+		assertType('string|false', getenv('foo'));
+	}
```
ondrejmirtes: Please add a test for an argument that's both, something like `getenv($mixed)`. To both PHP 7.4 and PHP 8.0.

ondrejmirtes (changes requested):
I tried the delta change myself, it works.

tests/PHPStan/Analyser/NodeScopeResolverTest.php
VincentLanglet: There is already similar conditions checks like above
```
if (PHP_VERSION_ID < 80000) {
			yield __DIR__ . '/data/explode-php74.php';
		} else {
			yield __DIR__ . '/data/explode-php80.php';
		}
```

The expectated type are different based on the PHP version, that's why I split the files.

The `// lint` condition is only used for the linter, not for the tests right ?

tests/PHPStan/Analyser/data/getenv-php74.php
VincentLanglet: Test added

tests/PHPStan/Analyser/NodeScopeResolverTest.php
ondrejmirtes: Files in nsrt/ when testing type inference are filtered based on`// lint` as well.

VincentLanglet: Oh ! Awesome !

src/Php/PhpVersion.php
```diff
@@ -353,4 +353,9 @@ public function substrReturnFalseInsteadOfEmptyString(): bool
 		return $this->versionId < 80000;
 	}
 
+	public function getenvAcceptsNull(): bool
```
ondrejmirtes: Dead code, you can remove this

tests/PHPStan/Analyser/nsrt/getenv-php74.php
```diff
@@ -0,0 +1,20 @@
+<?php // lint <= 7.4
+
+namespace GetenvPHP74;
+
+use function PHPStan\Testing\assertType;
+
+class Foo
+{
+
+	public function test(string|null $stringOrNull, mixed $mixed)
```
ondrejmirtes: This isn't valid code before PHP 8, remove the `mixed` typehint

ondrejmirtes: The unions aren't either

ondrejmirtes: You can move it to PHPDocs

tests/PHPStan/Analyser/nsrt/getenv-php74.php
```diff
@@ -0,0 +1,20 @@
+<?php // lint <= 7.4
```
ondrejmirtes: `lint < 8.0`

tests/PHPStan/Analyser/nsrt/getenv-php80.php
```diff
@@ -0,0 +1,20 @@
+<?php // lint >= 80
```
ondrejmirtes: `lint >= 8.0`

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #4008, opened 2025-05-22
Fix var php doc inheritance for constant

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10911


ondrejmirtes:
Thank you.

PR #4009, opened 2025-05-22
fix: add updated signature for `getenv`

msbit:
Add the following signature for `getenv`:

     getenv(?string $name = null, bool $local_only = false): string|array|false

to the function map (specifically for PHP 8 and above). This replaces all existing signatures, and should fix [13065][1].

[1]: https://github.com/phpstan/phpstan/issues/13065


msbit:
@ondrejmirtes let me know if this is what you were expecting. It shows no error for the following:

```
<?php

getenv();
getenv(null);
getenv('TERM');
getenv(null, false);
getenv(null, true);
getenv('TERM', false);
getenv('TERM', true);
```

msbit:
@ondrejmirtes I'm not sure about the interactions of optional arguments (`name=`) to allow this to replicate the no-argument override in the base function map. Do you have any advice?

VincentLanglet:
@msbit Fixed by https://github.com/phpstan/phpstan-src/pull/4007

PR #4010, opened 2025-05-22
Improve mb_convert_encoding return type

VincentLanglet:
Not sure if it's enough to close https://github.com/phpstan/phpstan/issues/13064
The `AccessoryEncodingString` would require lot more changes.

### DIFF
--- a/src/Type/Php/MbConvertEncodingFunctionReturnTypeExtension.php
+++ b/src/Type/Php/MbConvertEncodingFunctionReturnTypeExtension.php
@@ -52,12 +52,12 @@ final class MbConvertEncodingFunctionReturnTypeExtension implements DynamicFunct
 
 		$result = TypeCombinator::intersect($initialReturnType, $this->generalizeStringType($argType));
 		if ($result instanceof NeverType) {
-			return null;
+			$result = $initialReturnType;
 		}
 
 		if ($this->phpVersion->throwsValueErrorForInternalFunctions()) {
 			if (!isset($functionCall->getArgs()[2])) {
-				return $result;
+				return TypeCombinator::remove($result, new ConstantBooleanType(false));
 			}
 			$fromEncodingArgType = $scope->getType($functionCall->getArgs()[2]->value);
 
@@ -90,7 +90,7 @@ final class MbConvertEncodingFunctionReturnTypeExtension implements DynamicFunct
 			}
 
 			if (!$returnFalseIfCannotDetectEncoding) {
-				return $result;
+				return TypeCombinator::remove($result, new ConstantBooleanType(false));
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/data/mb-convert-encoding-php8.php
```diff
@@ -0,0 +1,45 @@
...
+	string $constantString,
+	string $string,
+	array $mixedArray,
+	array $structuredArray,
+	array $stringList,
+	array $intList,
+	string|array|bool $union,
+	int $int,
+): void {
+	\PHPStan\Testing\assertType('array|string|false', mb_convert_encoding($mixed, 'UTF-8'));
```
stof: should probably be `array|string` here, as no encoding guessing can happen

tests/PHPStan/Analyser/data/mb-convert-encoding-php8.php
```diff
@@ -0,0 +1,45 @@
...
+	\PHPStan\Testing\assertType('string', mb_convert_encoding($constantString, 'UTF-8'));
+	\PHPStan\Testing\assertType('string', mb_convert_encoding($string, 'UTF-8'));
+	\PHPStan\Testing\assertType('array', mb_convert_encoding($mixedArray, 'UTF-8'));
+	\PHPStan\Testing\assertType('array{foo: string, bar: int, baz: string}', mb_convert_encoding($structuredArray, 'UTF-8'));
+	\PHPStan\Testing\assertType('list<string>', mb_convert_encoding($stringList, 'UTF-8'));
+	\PHPStan\Testing\assertType('list<int>', mb_convert_encoding($intList, 'UTF-8'));
+	\PHPStan\Testing\assertType('array{foo: string, bar: int, baz: string}|string|false', mb_convert_encoding($union, 'UTF-8'));
+	\PHPStan\Testing\assertType('array|string|false', mb_convert_encoding($int, 'UTF-8'));
```
stof: should maybe be `string` (considering that `int` is cast to string by the union type) or an error (for strict types)

src/Type/Php/MbConvertEncodingFunctionReturnTypeExtension.php
```diff
@@ -49,6 +55,45 @@ public function getTypeFromFunctionCall(
...
+			}
+			$fromEncodingArgType = $scope->getType($functionCall->getArgs()[2]->value);
+
+			$mayNotDetectEncoding = false;
+			if (!$fromEncodingArgType->isArray()->no()) {
+				$constantArrays = $fromEncodingArgType->getConstantArrays();
+				if (count($constantArrays) > 0) {
+					foreach ($constantArrays as $constantArray) {
+						if (count($constantArray->getValueTypes()) > 1) {
+							$mayNotDetectEncoding = true;
```
stof: the name looks weird to be here assuming the condition is correct. An array with multiple values will detect encoding.

VincentLanglet: Does `returnFalseIfCannotDetectEncoding` is better ?

stof: I would say yes.

stof (approved):

zonuexe:
It would be great to get this merged in the next release.
It'll make using `mb_convert_encoding()` with PHPStan 2.1.15+ much easier.

VincentLanglet:
Do you prefer that I target 2.1.x @ondrejmirtes ?

ondrejmirtes:
Thank you.

PR #4011, opened 2025-05-24
Improve support switch on ::class

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13069

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1656,13 +1656,11 @@ final class TypeSpecifier
 		if (
 			$leftType instanceof ConstantScalarType
 			&& !$rightExpr instanceof ConstFetch
-			&& !$rightExpr instanceof ClassConstFetch
 		) {
 			return [$binaryOperation->right, $leftType, $rightType];
 		} elseif (
 			$rightType instanceof ConstantScalarType
 			&& !$leftExpr instanceof ConstFetch
-			&& !$leftExpr instanceof ClassConstFetch
 		) {
 			return [$binaryOperation->left, $rightType, $leftType];
 		}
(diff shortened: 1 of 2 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -1656,13 +1656,11 @@ private function findTypeExpressionsFromBinaryOperation(Scope $scope, Node\Expr\
 		if (
 			$leftType instanceof ConstantScalarType
 			&& !$rightExpr instanceof ConstFetch
-			&& !$rightExpr instanceof ClassConstFetch
```
VincentLanglet, line 1659: The findTypeExpressionFromBinaryOperation is used two times:
- Inside resolveIdentical
- Inside resolveEqual

Inside resolveIdentical the `ClassConstFetch` scenario is not needed. But removing the check does not have impact since then the following code
```
$expressions = $this->findTypeExpressionsFromBinaryOperation($scope, $expr);
		if ($expressions !== null) {
			$exprNode = $expressions[0];
			$constantType = $expressions[1];

			$unwrappedExprNode = $exprNode;
			if ($exprNode instanceof AlwaysRememberedExpr) {
				$unwrappedExprNode = $exprNode->getExpr();
			}

			$specifiedType = $this->specifyTypesForConstantBinaryExpression($unwrappedExprNode, $constantType, $context, $scope, $expr);
```
gives a NULL $specifiedType.

Inside resolveEqual the `ClassConstFetch` is needed for an easy check
```
$context->true()
				&& $exprNode instanceof ClassConstFetch
				&& $exprNode->name instanceof Node\Identifier
				&& strtolower($exprNode->name->toString()) === 'class'
				&& $constantType->isString()->yes()
```
without checking both the right side and the left side.

Not sure if you're ok with such change.
I could had a boolean parameter to remove or not the `ClassConstFetch` check but I'm not sure it's worth it.

ondrejmirtes:
Thank you.

PR #4012, opened 2025-05-24
Introduce allowBoolFloatNullAsArrayKey parameter

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12589
Closes https://github.com/phpstan/phpstan/issues/7884
Closes https://github.com/phpstan/phpstan/issues/7864

Naming is opened to discussion.

I think we could enable this in phpstan-strict-rules.


conf/config.neon
```diff
@@ -65,6 +65,7 @@ parameters:
 	reportStaticMethodSignatures: false
 	reportWrongPhpDocTypeInVarTag: false
 	reportAnyTypeWideningInVarTag: false
+	reportCastedArrayKey: false
```
ondrejmirtes: `Casted` is not a correct past tense/adjective form of `cast`. It's `cast`. Also it might be better to use it as a noun: `reportArrayKeyCast`.

ondrejmirtes (changes requested):
Also would be nice if it observed rule levels. Always wrong type like `null` to be always reported, but a partially wrong type like `string|stdClass` only on level 7+...

This is achieved with RuleLevelHelper::findTypeToCheck.

VincentLanglet:
Seems like the `RuleLevelHelper::findTypeToCheck` was not used at all (or wrongly used) on theses rules.

So I fixed it and added tests with and without the reportArrayKeyCast option.

VincentLanglet:
Current status, waiting https://github.com/phpstan/phpstan-src/pull/4166 first

VincentLanglet:
This is finally ready to be reviewed @ondrejmirtes :)

VincentLanglet:
Friendly ping @ondrejmirtes ; I would be interested enabling this on phpstan-strict-rules :)

ondrejmirtes:
So now this is only about the boolean. I don't think it's worth it adding a new config parameter just for that. Maybe it'd be better to advocate for deprecating that in PHP 8.6?

VincentLanglet:
> So now this is only about the boolean. I don't think it's worth it adding a new config parameter just for that.

It's for boolean or people which are using PHP < 8.5 (8.1) and still want the array key casted deprecated.
Since PHP supports 7.4 it might have some benefit.

Also I feel like waiting for 8.6 might be a little bit long for an important bug like the example of https://github.com/phpstan/phpstan/issues/12589 ...

> Maybe it'd be better to advocate for deprecating that in PHP 8.6?

It's would still be a good idea but I have no knowledge in C to implement this in PHP codebase

ondrejmirtes:
So deprecating bools was considered but rejected in the RFC process. Here's the comment and some discussion above it https://externals.io/message/127849#128184.

VincentLanglet:
> So deprecating bools was considered but rejected in the RFC process. Here's the comment and some discussion above it [externals.io/message/127849#128184](https://externals.io/message/127849#128184).

Then based on the fact it will be unlikely deprecated soon on PHP side, do you confirm that such options still make sens on PHPStan side ?

ondrejmirtes:
I don't like the name of the config parameter. You're not reporting array key casts. For example if someone uses `'1'`, you're not reporting that although it's still cast to `1`.

This is more like `allowFloatBoolNullAsArrayKey` with default set to `true`.

VincentLanglet:
> This is more like `allowFloatBoolNullAsArrayKey` with default set to `true`.

I renamed with `allowFloatBoolNullAsArrayKey`, but have some suggestions:
- allowBoolFloatNullAsArrayKey => Type are alphabetically ordered
- allowLooseArrayKeys 
- strictArrayKeyTypes (default false)

Any preferences ?

VincentLanglet:
Just found the issue https://github.com/phpstan/phpstan/issues/7564 and I wonder if it should be included in this option or in a separate one.

If so, rather than introducing `allowFloatBoolNull` param in AllowedArrayKeysTypes it could be something like checking if
```
$type->toArrayKeyType()->isSuperTypeOf($type)->yes();
```
(to avoid reporting numeric-string array keys)

VincentLanglet:
> > This is more like `allowFloatBoolNullAsArrayKey` with default set to `true`.
> 
> I renamed with `allowFloatBoolNullAsArrayKey`, but have some suggestions:
> 
> * allowBoolFloatNullAsArrayKey => Type are alphabetically ordered
> * allowLooseArrayKeys
> * strictArrayKeyTypes (default false)
> 
> Any preferences ?

Maybe you'll have some suggestion about the naming @staabm ?

staabm:
I would go with one of the allowFloatBoolNullAsArrayKey variants, as otherwise we need to document somewhere what "loose" or "strict" means in this context

VincentLanglet:
Any way to move this forward @ondrejmirtes ?

Do you prefer `allowBoolFloatNullAsArrayKey` to have the types sorted alphabetically ?

VincentLanglet:
I'm really interested by this feature @ondrejmirtes @staabm and I feel like only the naming is delaying it, can we find an agreement ?

I see
- allowFloatBoolNullAsArrayKey as proposed by ondrej
- allowBoolFloatNullAsArrayKey, the same but sorted alphabetically
- allowLooseArrayKeys
- strictArrayKeyTypes
- allowNonStringIntegerArrayKeys
...

staabm:
> * allowFloatBoolNullAsArrayKey as proposed by ondrej
> 
> * allowBoolFloatNullAsArrayKey, the same but sorted alphabetically

I am fine with these 2. 

[arguments in my previous comment](https://github.com/phpstan/phpstan-src/pull/4012#issuecomment-3430559644).

VincentLanglet:
I sorted alphabetically to allowBoolFloatNullAsArrayKey then.

conf/config.neon
```diff
@@ -80,6 +80,7 @@ parameters:
 	reportStaticMethodSignatures: false
 	reportWrongPhpDocTypeInVarTag: false
 	reportAnyTypeWideningInVarTag: false
+	allowBoolFloatNullAsArrayKey: true
```
staabm: should the more strict behaviour enabled by default in bleeding edge when strict-rules are installed?

ondrejmirtes:
I think the name of the option is a bit obsolete since the float type and null type are always rejected on newer PHP versions.

VincentLanglet:
> I think the name of the option is a bit obsolete since the float type and null type are always rejected on newer PHP versions.

But an option just called "allowBool" would be confusing for old php version...

Maybe should we look for the opposite like restrictToIntAndString (i need to discuss with my favorite bot to have some name proposition)

conf/config.neon
VincentLanglet: Sure

VincentLanglet: But I think it's configured on phpstan-strict-rules side like https://github.com/phpstan/phpstan-strict-rules/blob/f6f5b3ed51450697b857a06e2de0fa934dee2144/rules.neon#L3-L14

VincentLanglet:
> I think the name of the option is a bit obsolete since the float type and null type are always rejected on newer PHP versions.

I have trouble to find a name satisfying.
Looking at the issues it's said `using non-array-key as an array key` or `array key is not int or string`, so:
- something with `NonStandardArrayKeyTypes` like `allowNonStandardArrayKeyTypes` or `forbidNonArrayKeyTypes`
- something with `IntOrString` like `enforceIntOrStringKeys`

Really hard to name this thing...

conf/config.neon
staabm: right. we can decide whether we want this new rule one of 
- bleeding edge only (meaning it will be default-enabled on next major of phpstan-src)
- strict rules only
- bleeding edge + strict rules only (meaning it will be default-enabled on next major of strict rules)
- opt-in only

ondrejmirtes:
The hardest thing in computer science, right 😊

ondrejmirtes:
What about `reportNonIntStringArrayKey`?

VincentLanglet:
> What about `reportNonIntStringArrayKey`?

I'm ok with this ;

I used `IntOrString` because I was wondering if someone could understand `IntString` as a portion of `numeric-string` like `'1'`

Cf `non-int-string` in your message https://github.com/phpstan/phpstan/issues/10239#issuecomment-1837571316

ondrejmirtes:
I think we can go with `reportNonIntStringArrayKey` (default false). I don't know how compose a config parameter name from `IntOrString`. We'd need something like "nor int nor string" to get a negative.

VincentLanglet:
> reportNonIntStringArrayKey

Renamed !

staabm (approved):
love it.

please make sure to formulate the commit message when 'squash and merge' so it reflects the state this PR ended up with.

maybe issue-bot will report some new issues affected by it after its rebased.

ondrejmirtes:
1) Please send a docs update after merging.
2) Please update phpstan-strict-rules and set this to `%featureToggles.bleedingEdge%`.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

janedbal:
Nice! Did somebody check if this fully outbuilds our [forbidUnsafeArrayKey](https://github.com/shipmonk-rnd/phpstan-rules?tab=readme-ov-file#forbidunsafearraykey)?

ondrejmirtes:
@janedbal Mixed is reported by PHPStan even without this option on L9+. This new option reports null, float, bool, they have that in common. PHPStan does not offer the `reportMixed` and `reportInsideIsset` options.

PR #4013, opened 2025-05-24
Fix HasOffsetValueType::searchArray

VincentLanglet:
See https://onlinephp.io/c/192c51

We're sure that the function cannot return false, but we cannot sure it's not an early key.

cc @herndlm since you implemented it.


herndlm (approved):
Makes sense, thx!

ondrejmirtes:
Thank you.

PR #4014, opened 2025-05-24
Improve array_search inference

VincentLanglet:
First commit is extracted here: https://github.com/phpstan/phpstan-src/pull/4013

### DIFF
--- a/src/Type/Accessory/AccessoryArrayListType.php
+++ b/src/Type/Accessory/AccessoryArrayListType.php
@@ -221,7 +221,7 @@ class AccessoryArrayListType implements CompoundType, AccessoryType
 		return new MixedType();
 	}
 
-	public function searchArray(Type $needleType, TrinaryLogic $strict): Type
+	public function searchArray(Type $needleType, ?TrinaryLogic $strict = null): Type
 	{
 		return new MixedType();
 	}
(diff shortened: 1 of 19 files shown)


src/Type/Accessory/AccessoryArrayListType.php
```diff
@@ -241,7 +241,7 @@ public function reverseArray(TrinaryLogic $preserveKeys): Type
 		return new MixedType();
 	}
 
-	public function searchArray(Type $needleType): Type
+	public function searchArray(Type $needleType, TrinaryLogic $strict): Type
```
ondrejmirtes: Can't be done, it's `@api`.

ondrejmirtes (changes requested):
Also please base this on top of 2.1.x.

src/Type/Accessory/AccessoryArrayListType.php
VincentLanglet: So I need to make it nullable ? `?TrinaryLogic $strict = null`

ondrejmirtes: Yes please. And with null passed keep the current behaviour.

VincentLanglet: Done

ondrejmirtes:
Thank you!

PR #4016, opened 2025-05-24
Attribute `#[AutowiredService]` for slimmer config.neon

ondrejmirtes:


composer.json
```diff
@@ -72,10 +72,17 @@
 		"platform-check": false,
 		"sort-packages": true,
 		"allow-plugins": {
-			"cweagans/composer-patches": true
+			"cweagans/composer-patches": true,
+			"ondrejmirtes/composer-attribute-collector": true,
+			"vaimo/composer-patches": true
```
staabm, line 77: why do we need 2 composer-patches packages?

ondrejmirtes: We don't, it's probably a mistake and one of those lines can be removed.

PR #4019, opened 2025-05-25
More `DynamicFunctionReturnTypeExtension` converted to `#[AutowiredService]` attribute usage

staabm:
bulk add the attribute where possible. on the way I discovered a few special cases which I left in the config.neon and will look into in separate PRs


ondrejmirtes:
😍 Love it, thanks!

PR #4021, opened 2025-05-25
Map all BrokerFactory tags in AutowiredAttributeServicesExtension

staabm:
only did minimal changes to config.neon to reduce risk for conflicts with https://github.com/phpstan/phpstan-src/pull/4019

Inspired by the list in `BrokerFactory` and `ConditionalTagsExtension`.


ondrejmirtes:
There's a conflict.

ondrejmirtes:
Thank you!

PR #4023, opened 2025-05-25
Improve ReplaceFunctionsDynamicReturnTypeExtension

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9870

### DIFF
--- a/src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
@@ -22,7 +22,6 @@ use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeUtils;
-use PHPStan\Type\UnionType;
 use function array_key_exists;
 use function count;
 use function in_array;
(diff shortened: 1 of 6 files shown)


src/Type/Php/ReplaceFunctionsDynamicReturnTypeExtension.php
```diff
@@ -92,68 +103,89 @@ private function getPreliminarilyResolvedTypeFromFunctionCall(
...
-				if ($subjectArgumentType->isUppercaseString()->yes() && $replaceArgumentType->isUppercaseString()->yes()) {
-					$accessories[] = new AccessoryUppercaseStringType();
-				}
+			$subjectArgumentType = $subjectArgumentType->tryRemove($stringArgumentType);
+		}
 
-				if (count($accessories) > 0) {
-					$accessories[] = new StringType();
-					return new IntersectionType($accessories);
+		if ($subjectArgumentType->isArray()->yes()) {
```
VincentLanglet: When we get `string|array<string>`

I didn't find a better way:
- First, to work with the `string` part first (in order to know if it's non-empty, lowercase, etc)
- Secondly to work only with the `array` part (in order to know the shape, if it's a list, non empty, the values, etc)

To access the string part, the best I found was 
```
$stringArgumentType = TypeCombinator::intersect(new StringType(), $subjectArgumentType);
```

Same idea for array because
- $subjectArgumentType->getArrays() does not return array on union
- TypeUtiles::toBenevolent($subjectArgumentType)->getArrays() is returning arrays but is loosing the non-empty and the list accessory.

ondrejmirtes:
Thank you.

PR #4024, opened 2025-05-25
More service converted to `#[AutowiredService]` attribute usage

staabm:


staabm:
Pretty interessting that most `*ClassReflectionExtension` in the `config.neon` don't have any tags added.
seems it might make sense to drop these extension type from `AutowiredAttributeServicesExtension`.

see 
- https://github.com/phpstan/phpstan-src/pull/4024/commits/0cdc824109c87c6636dd08d767ebe66923dae6d1
- https://github.com/phpstan/phpstan-src/pull/4024/commits/3f3092893272f26c5213ea622df7077e6bc49f89
in which I reverted back to the previous style to fix a end-2-end test

ondrejmirtes:
Yeah, if there's a service of a class implementing an interface, but the service doesn't have a tag, it currently cannot use the attribute.

This is about extensions manually retrieved in LazyClassReflectionExtensionRegistryProvider to ensure a certain ordering.

ondrejmirtes:
Thank you!

PR #4028, opened 2025-05-27
Improve TrimDynamicReturnType for constantString

VincentLanglet:


src/Type/Php/LtrimFunctionReturnTypeExtension.php
```diff
@@ -53,12 +54,27 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+
+			foreach ($trimConstantStrings as $trimConstantString) {
+				if (count($stringConstantStrings) > 0) {
+					foreach ($stringConstantStrings as $stringConstantString) {
+						$result[] = new ConstantStringType(
+							ltrim($stringConstantString->getValue(), $trimConstantString->getValue()),
+							true,
+						);
+					}
+				} elseif ($trimConstantString->getValue() === '\\' && $string->isClassString()->yes()) {
```
staabm, line 70: this IF/ELSE-IF construction makes me feel we need a test for a type like `'foo'|class-string<Bar>`

VincentLanglet: When you check `'foo'|class-string<Bar>` constantStrings, it returns [].
When you check if `'foo'|class-string<Bar>` is classString, it returns maybe.

So there is no bug.
I added a test: https://github.com/phpstan/phpstan-src/pull/4028/commits/4c8bfe480bc79844d0e7eec9791fb068ececc6a1

We could improve the type for a better understanding of `ltrim('foo'|class-string<Bar>, '\\')` but
- I don't think it's worth it
- It's kinda harder to handle return type of union type

staabm: yeah, I am fine with the current result. just wanted to make sure we have the current state recorded in a test, so in case it changes we see it immediately

ondrejmirtes:
Thank you.

PR #4029, opened 2025-05-27
Fix DOMElement::$attributes stub

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13076

I added 2 tests of non-regresions.


ondrejmirtes:
Thank you.

PR #4030, opened 2025-05-27
fix: False match.unhandled on array with multiple booleans

ping-yee:
Fix https://github.com/phpstan/phpstan/issues/13029

### DIFF
--- a/src/Rules/Comparison/MatchExpressionRule.php
+++ b/src/Rules/Comparison/MatchExpressionRule.php
@@ -132,6 +132,7 @@ final class MatchExpressionRule implements Rule
 				!$remainingType instanceof NeverType
 				&& !$this->isUnhandledMatchErrorCaught($node)
 				&& !$this->hasUnhandledMatchErrorThrowsTag($scope)
+				&& !$remainingType->isArray()->yes()
 			) {
 				$errors[] = RuleErrorBuilder::message(sprintf(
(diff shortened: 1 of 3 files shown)


VincentLanglet:
If you ignore the error for every remaining array, I assume you won't have an error if you miss an "array" case

Like
- matching array(int) with only some constant-int
- matching array(bool, bool) with only 3 cases (instead of four)


Can you add tests to prove it works ?

src/Rules/Comparison/MatchExpressionRule.php
```diff
@@ -132,6 +132,7 @@ public function processNode(Node $node, Scope $scope): array
 				!$remainingType instanceof NeverType
 				&& !$this->isUnhandledMatchErrorCaught($node)
 				&& !$this->hasUnhandledMatchErrorThrowsTag($scope)
+				&& !$remainingType->isArray()->yes()
```
ondrejmirtes, line 135: This is a bad bugfix. This shuts up the rule for all arrays.

ondrejmirtes (changes requested):

src/Rules/Comparison/MatchExpressionRule.php
ping-yee: I'm trying to find the right way to fix this. Thanks for the review! @ondrejmirtes  @VincentLanglet

PR #3821, opened 2025-02-12
TableErrorFormatter: improve formatting of error tips

FeBe95:
This PR adds some small improvements to the error output, when using the "table" error formatter:

1. Align tip text after `💡` emoji
2. Trim trailing new-line of multi-line tips

<table>
  <thead>
  <tr>
    <th width="50%">Before</th>
    <th width="50%">After</th>
  </tr>
  </thead>
  <tbody>
  <tr>
    <td valign="top">

```
 ------ ----------------------------
  Line   FooBar.php
 ------ ----------------------------
  36     I am an error message.
         🪪  foobar.buz
         💡 I am an error tip.
         💡 I span multiple lines.

         ✏️  FooBar.php
  40     I am an error message.
         🪪  foobar.buz
         💡 I am an error tip.
         💡 I span multiple lines.

         ✏️  FooBar.php
 ------ ----------------------------
```
</td>
    <td valign="top">

```
 ------ ----------------------------
  Line   FooBar.php
 ------ ----------------------------
  36     I am an error message.
         🪪  foobar.buz
         💡  I am an error tip.
         💡  I span multiple lines.
         ✏️  FooBar.php
  40     I am an error message.
         🪪  foobar.buz
         💡  I am an error tip.
         💡  I span multiple lines.
         ✏️  FooBar.php
 ------ ----------------------------


```

</td>
  </tr>
  </tbody>
</table>

PR #4031, opened 2025-05-27
TableErrorFormatter: Use two spaces instead of one

FeBe95:
## Description

This PR reverts commit 59a003a4bf84b319d94fd2a4ff43a5870c316923 which was part of #3821.

## Why?

Before merging #3821, I was asked to replace the seemingly incorrect double spaces with a single space.

After upgrading to the latest version of PHPStan today, I noticed that the double spaces from before were used on purpose, because an emoji character may exceed the width of a single monospace character. Depending on the terminal engine in use, the following characters are either overlapped or shifted to the right.

Using two spaces instead of one makes sure that the output looks good in all terminals (see "**After**" examples).

## Example 1: PhpStorm
The emoji overlaps the next character.

<table>
  <thead>
  <tr>
    <th width="50%">Before</th>
    <th width="50%">After</th>
  </tr>
  </thead>
  <tbody>
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>
  </tbody>
</table>

## Example 2: Windows Terminal
The emoji pushes all following characters to the right.

<table>
  <thead>
  <tr>
    <th width="50%">Before</th>
    <th width="50%">After</th>
  </tr>
  </thead>
  <tbody>
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>
  </tbody>
</table>

### DIFF
--- a/src/Command/ErrorFormatter/TableErrorFormatter.php
+++ b/src/Command/ErrorFormatter/TableErrorFormatter.php
@@ -84,7 +84,7 @@ final class TableErrorFormatter implements ErrorFormatter
 				$filePath = $error->getTraitFilePath() ?? $error->getFilePath();
 				if ($error->getIdentifier() !== null && $error->canBeIgnored()) {
 					$message .= "\n";
-					$message .= '🪪 ' . $error->getIdentifier();
+					$message .= '🪪  ' . $error->getIdentifier();
 				}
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #4032, opened 2025-05-28
Fix "Named arguments are supported only on PHP 8.0 and later." false positive

staabm:
- running https://phpstan.org/r/8df32644-1f4b-4b46-934d-5134753fbd47 in a codebase outside of phpstan-src did not reproduce the error

- putting the reproducer under `CallMethodsRuleTest` did also not reproduce the error, as `ignore-by-php-version.neon.php` is not loaded in the test-suite

### DIFF
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -34,7 +34,7 @@ use const PHP_INT_SIZE;
 final class ConstantResolver
 {
 
-	public const PHP_MIN_VERSION_ID = 50207;
+	public const PHP_MIN_ANALYZABLE_VERSION_ID = 50207;
 
 	/** @var array<string, true> */
 	private array $currentlyResolving = [];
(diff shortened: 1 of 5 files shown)


staabm:
I think the underlying problem is, that we remember PHP_VERSION_ID when narrowing and therefore fallback to a wrong scope-state.

see the scope-dump before/after the IF in this snippet:
https://phpstan.org/r/3462410d-1695-429c-ab2a-b552d340bb76

---

my current fix is wrong though

src/Analyser/ConstantResolver.php
```diff
@@ -34,6 +34,8 @@
 final class ConstantResolver
 {
 
+	public const PHP_MIN_ANALYZABLE_VERSION_ID = 50207;
```
staabm, line 37: this constant might need a better name. couldn't come up with something better.

staabm:
the remaining failing tests are the same ones as [in another just rebased PR of mine](https://github.com/phpstan/phpstan-src/pull/3934), therefore I think the errors are not related to this PR

ondrejmirtes:
Thank you!

PR #4033, opened 2025-05-28
Improved fix "Named arguments are supported only on PHP 8.0 and later" false positive"

staabm:
This reverts commit 1c565b5.

better fix for https://github.com/phpstan/phpstan-src/pull/4032

---

repro https://phpstan.org/r/8df32644-1f4b-4b46-934d-5134753fbd47


ondrejmirtes:
Thank you!

PR #4040, opened 2025-06-02
Improve NAN inferences

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13097

This was introduced in https://github.com/phpstan/phpstan-src/commit/29fcf80e1676f1a13ee04e88db6dbeb8685b58e1 but seems like the `return NAN` doesn't exists anymore so we can improve NAN inferences without regression.


ondrejmirtes:
Thank you.

PR #2955, opened 2024-03-05
Update Box to 4.5.1@1e10a1e 2023-11-04 17:57:48 UTC

ondrejmirtes:

PR #4042, opened 2025-06-03
AutowiredParameter attribute

ondrejmirtes:


staabm:
root cause for the PHP 7.4 only CI errors is, that the transformed code contains lines like
```
		#[AutowiredParameter] 		bool $checkClassCaseSensitivity,
		#[AutowiredParameter]		bool $checkThisOnly
```

while for PHP7.x compat the attributes need to stay on there own line:

```
		#[AutowiredParameter]
 		bool $checkClassCaseSensitivity,
		#[AutowiredParameter]
		bool $checkThisOnly
```

ondrejmirtes:
I know. nikic/PHP-Parser 5.5.0 even addresses this, I think I'm setting it up correctly but it doesn't work yet.

staabm:
narrowed it further down. it only happens in `__construct` methods. if renaming the method to `abc` it works like expected

ondrejmirtes:
You don't need to look into it, I already solved it.

ondrejmirtes:
Blocked on https://github.com/phpstan/phpstan-src/pull/2955 + https://github.com/humbug/php-scoper/issues/1122

PR #4043, opened 2025-06-04
Add more precise return types for the openssl cipher functions

stof:
Those functions return false (and trigger a warning) only when the argument is an unknown algorithm, which is not something worth checking when using them with a known algorithm.

I'm providing this improvement only on PHP 8+, because PHP 7 can have other kind of warnings (that have been upgraded to `ValueError` in PHP 8)

Replaces https://github.com/phpstan/phpstan-src/pull/3582

### DIFF
--- a/src/Type/Php/OpensslCipherFunctionsReturnTypeExtension.php
+++ b/src/Type/Php/OpensslCipherFunctionsReturnTypeExtension.php
@@ -1,0 +1,89 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Php\PhpVersion;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use function array_map;
+use function array_unique;
+use function count;
+use function function_exists;
+use function in_array;
+use function is_null;
+use function openssl_get_cipher_methods;
+use function strtoupper;
+
+#[AutowiredService]
+final class OpensslCipherFunctionsReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+{
+
+	/** @var string[]|null */
+	private ?array $supportedAlgorithms = null;
+
+	public function __construct(private PhpVersion $phpVersion)
+	{
+	}
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return in_array($functionReflection->getName(), ['openssl_cipher_iv_length', 'openssl_cipher_key_length'], true);
+	}
+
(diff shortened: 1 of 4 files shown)


src/Type/Php/OpensslCipherFunctionsReturnTypeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+		$strings = $scope->getType($functionCall->getArgs()[0]->value)->getConstantStrings();
+		$results = array_unique(array_map(fn (ConstantStringType $algorithm): bool => $this->isSupportedAlgorithm($algorithm->getValue()), $strings));
+
+		if (count($results) === 1) {
+			return $results[0]
+				? TypeCombinator::remove($returnType, new ConstantBooleanType(false))
+				: new ConstantBooleanType(false);
+		}
+
+		return $returnType;
```
staabm: instead of 

```
$returnType = ParametersAcceptorSelector::selectFromArgs(
			$scope,
			$functionCall->getArgs(),
			$functionReflection->getVariants(),
		)->getReturnType();
```

you can just `return null;` for the cases where the extension cannot provide a better type then the default.
that way you can simplify the extension a bit.

src/Type/Php/OpensslCipherFunctionsReturnTypeExtension.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	private function isSupportedAlgorithm(string $algorithm): bool
+	{
+		return in_array(strtoupper($algorithm), $this->getSupportedAlgorithms(), true);
+	}
+
+	/** @return string[] */
```
staabm, line 73: might work to use a more precise return type

```suggestion
	/** @return uppercase-string[] */
```

stof: I inspired from MbFunctionsReturnTypeExtensionTrait, which also uses simply `string[]`

src/Type/Php/OpensslCipherFunctionsReturnTypeExtension.php
stof: I refactored the code to return null for all such cases.

staabm:
Needs cs fix

stof:
@staabm done

conf/config.neon
```diff
@@ -1580,6 +1580,11 @@ services:
 		tags:
 			- phpstan.dynamicFunctionThrowTypeExtension
 
+	-
```
ondrejmirtes: Please put this on 2.1.x and use `AutowiredService` attribute instead of changing config.neon. Thanks.

ondrejmirtes (changes requested):

conf/config.neon
stof: @ondrejmirtes sorry for the delay. This is now done.

ondrejmirtes:
Thank you.

stof:
@ondrejmirtes I discovered that in PHP <8.5, the `openssl_get_cipher_methods` function uses a different source of truth than the resolution of cipher algorithms in other functions, which means that some algorithms returned in this function are actually not supported in `openssl_cipher_iv_length`. See https://github.com/php/php-src/issues/19994

Maybe this return type extension should be restricted to run only on PHP 8.5+ instead of PHP 8.0+. Or maybe it should perform actual filtering by attempting to read the iv length for PHP 8.0 to 8.4.

ondrejmirtes:
Please open an issue, this comment would get lost.

PR #4045, opened 2025-06-05
Update RedisCluster signatures

RobiNN1:
I synchronized signatures with https://github.com/JetBrains/phpstorm-stubs/blob/master/redis/RedisCluster.php


ondrejmirtes:
Hey, this is too many changes and it's impossible to review. What actual bugs are you experiencing?

We could also remove these entries altogether and just rely on phpstorm-stubs. "Reflection golden test" workflow will tell us what actual data would change.

RobiNN1:
I added changes from from https://github.com/JetBrains/phpstorm-stubs/commit/348c656b52079cf7e4fc7f16a32b36c66b883f14 

The current signatures are outdated, and more than 50 functions are missing. It was outdated in phpstorm stubs and here too

RobiNN1:
I'm working on new PR and tried to keep the original parameter names, but the difference is still huge. The old signatures of many functions don't even have defined return types.

ondrejmirtes:
Here's a PR where I'm removing all the signatures: https://github.com/phpstan/phpstan-src/pull/4055

Here are the changes in reflection data: https://github.com/phpstan/phpstan-src/actions/runs/15553088664/job/43787529381

You could go through them and maybe improve the stubs in phpstorm-stubs based on that so that we could remove the signatures on PHPStan's side. Thank you.

PR #4046, opened 2025-06-06
PHPUnit 10

ondrejmirtes:


staabm:
I guess this means 1.12.x will see only critical bugfixes as of now, as merging tests between 1.12.x and 2.1.x will be a hard job?

ondrejmirtes:
Yeah, the tests will diverge even more with what I have planned.

PR #4060, opened 2025-06-11
Remove leftover local variable in ClassReflectionPropertyHooksTest

staabm:


ondrejmirtes:
I'd pause on this a bit. For me it'd be best if the core issue was solved in PHPUnit and I wouldn't need to change the tests at all. There's a lot of tests affected.

PR #4061, opened 2025-06-13
Resolve substr based on configured PHP version

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13129


tests/PHPStan/Analyser/NodeScopeResolverPhp7Test.php
```diff
@@ -4,7 +4,7 @@
 
 use PHPStan\Testing\TypeInferenceTestCase;
 
-class LooseConstComparisonPhp7Test extends TypeInferenceTestCase
+class NodeScopeResolverPhp7Test extends TypeInferenceTestCase
```
VincentLanglet: Since the neon file is only
```
parameters:
	phpVersion: 70400 # PHP 7.4
```
we could use those TestCase for all the configured-phpVersion-related tests instead of creating a TestCase every time.

tests/PHPStan/Analyser/data/bug-13129-php7.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
staabm, line 1: I think you can add the `lint` comment which makes the file beeing analyzed in the proper version

staabm:
btw: since phpunit got upgraded on the PHPStan 2.x branch, we need to be careful to send PRs to PHPStan 1.12.x only when easy mergabe into 2.x (attributes got added etc)

tests/PHPStan/Analyser/data/bug-13129-php7.php
VincentLanglet: The lint comment is for the runtime PHP version no ? (Like it's done for a lot of NodeScopeResolverPHPTest).

Here I want to allow Analysing this file on both PHP 7 and PHP 8, but telling PHPStan to consider he's in PHP 7.

VincentLanglet:
> btw: since phpunit got upgraded on the PHPStan 2.x branch, we need to be careful to send PRs to PHPStan 1.12.x only when easy mergabe into 2.x (attributes got added etc)

Only change is this attribute
https://github.com/phpstan/phpstan-src/blob/2.1.x/tests/PHPStan/Analyser/LooseConstComparisonPhp7Test.php#L24

I dunno if it will even create a conflict but if ondrej ask, I can target 2.1.x instead :)

ondrejmirtes:
I'd prefer this one only for 2.1.x :) Thanks.

VincentLanglet:
Rebased on 2.1.x @ondrejmirtes

tests/PHPStan/Analyser/NodeScopeResolverPhp7Test.php
```diff
@@ -5,7 +5,7 @@
 use PHPStan\Testing\TypeInferenceTestCase;
 use PHPUnit\Framework\Attributes\DataProvider;
 
-class LooseConstComparisonPhp7Test extends TypeInferenceTestCase
+class NodeScopeResolverPhp7Test extends TypeInferenceTestCase
```
ondrejmirtes: I don't like these class names. I don't want anything else to show up when I start typing "NodeScopeResolverTest" in my IDE. Sorry :)

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/NodeScopeResolverPhp7Test.php
VincentLanglet: The bug-13129 is not really related to looseComparison since it's related to substr.

The purpose of this class is to test with the given php version in the config rather than the runtime one. Any suggestion for the name then ?

ondrejmirtes: I'd add new couple of classes with "Substr" in their name. Instead of touching existing tests.

ondrejmirtes:
Thank you!

PR #4067, opened 2025-06-20
Fix assertSuperType logic

vojtech-dobes:
So this should actually make `assertSuperType` work in the wilderness, but I fail to see how to test it. It seems like the `FileAssertRule` needs to reimplement the logic for itself.


ondrejmirtes (changes requested):
Would be nice if this had an actual test (failing without changes in src/). There's TypeInferenceTestCaseTest you can use.

vojtech-dobes:
> Would be nice if this had an actual test (failing without changes in src/). There's TypeInferenceTestCaseTest you can use.

Tests added.

ondrejmirtes:
Thank you.

PR #4068, opened 2025-06-20
Improve return type of array map

VincentLanglet:
From https://github.com/phpstan/phpstan-src/pull/3156

Closes https://github.com/phpstan/phpstan/issues/11180
Closes https://github.com/phpstan/phpstan-src/pull/3156

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -159,7 +159,6 @@ use function is_numeric;
 use function is_string;
 use function ltrim;
 use function md5;
-use function preg_match;
 use function sprintf;
 use function str_starts_with;
 use function strlen;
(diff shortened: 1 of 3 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -2329,25 +2330,43 @@ static function (Node $node, Scope $scope) use ($arrowScope, &$arrowFunctionImpu
...
-				)->getReturnType();
+				if ($node->name instanceof String_) {
+					/** @var non-empty-string $name */
+					$name = $node->name->value;
+					$functionName = new Name($name);
+				} elseif (
+					$node->name instanceof FuncCall
+					&& $node->name->isFirstClassCallable()
+					&& $node->name->getAttribute('phpstan_cache_printer') !== null
+					&& preg_match('/\A(?<name>\\\\?[^()]+)\(...\)\z/', $node->name->getAttribute('phpstan_cache_printer'), $m) === 1
```
ondrejmirtes: Why the original code with:

```php
				return ParametersAcceptorSelector::selectFromArgs(
					$this,
					$node->getArgs(),
					$calledOnType->getCallableParametersAcceptors($this),
					null,
				)->getReturnType();
```

does not work here? That should be addressed instead.

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
VincentLanglet: I took the code from the previous PR but it might be improved indeed so I re-tried.

I tried to simplify/rewrite the code in order to show what was missing in the original code. 
Now it should be more clear with https://github.com/phpstan/phpstan-src/pull/4068/files

There is nothing wrong with
```
ParametersAcceptorSelector::selectFromArgs(
					$this,
					$node->getArgs(),
					$calledOnType->getCallableParametersAcceptors($this),
					null,
				)
```
but the issue was the early `getReturnType ` and the fact the DynamicReturnTypeExtension was not used in this case.

Is it better ?

ondrejmirtes:
Nice, thank you!

PR #4070, opened 2025-06-23
Fix "Option 'indent' must either be all spaces or a single tab"

staabm:
closes https://github.com/phpstan/phpstan/issues/13180


ondrejmirtes:
Thank you!

PR #4071, opened 2025-06-23
Describe array shape key in quotes unless valid identifier

vojtech-dobes:
This should make all array shape description parseable as valid phpDoc (unless truncated due to too may items).


ondrejmirtes:
Thank you!

PR #4073, opened 2025-06-26
Avoid false positif in_array always true

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13151
Closes https://github.com/phpstan/phpstan/issues/8818
Closes https://github.com/phpstan/phpstan/issues/12755
Closes https://github.com/phpstan/phpstan/issues/12412

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -157,6 +157,11 @@ final class ImpossibleCheckTypeHelper
 											continue;
 										}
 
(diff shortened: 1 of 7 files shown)


VincentLanglet:
I don't think it's enough to close https://github.com/phpstan/phpstan-src/pull/4073 @ondrejmirtes but at least I hope it will help.

Seems like this fix:
- Avoid false positive
- Do not produce any tested regression (and even if there is, I think it's better for PHPStan to have false-negative rather than false-positive)

ondrejmirtes:
Thank you!

PR #4074, opened 2025-06-26
Playground: Implement PhpdocCommentRule

staabm:
as discussed in https://github.com/phpstan/phpstan/discussions/13189#discussioncomment-13584928

### DIFF
--- a/src/Rules/Playground/PhpdocCommentRule.php
+++ b/src/Rules/Playground/PhpdocCommentRule.php
@@ -18,7 +18,7 @@ final class PhpdocCommentRule implements Rule
 
 	public function getNodeType(): string
 	{
-		return Node::class;
+		return Node\Stmt::class;
 	}
 
 	public function processNode(Node $node, Scope $scope): array
--- a/src/Rules/Playground/PhpdocCommentRule.php
+++ b/src/Rules/Playground/PhpdocCommentRule.php
@@ -1,0 +1,55 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Playground;
+
+use PhpParser\Comment;
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\VirtualNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\ShouldNotHappenException;
+use function str_contains;
+use function str_starts_with;
+
+/**
+ * @implements Rule<Node>
+ */
+final class PhpdocCommentRule implements Rule
+{
+
+	public function getNodeType(): string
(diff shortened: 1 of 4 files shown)


src/Rules/Playground/PhpdocCommentRule.php
```diff
@@ -0,0 +1,56 @@
...
+
+		$comments = $node->getAttribute('comments', []);
+
+		$errors = [];
+		foreach ($comments as $comment) {
+			if (!$comment instanceof Comment) {
+				throw new ShouldNotHappenException();
+			}
+
+			if (!str_contains($comment->getText(), '@')) {
```
staabm: do we need something more sophiscated, like "contains a real phpdoc tag"?
if so, how to find out?

mvorisek: `(\s|^)@\w+(\s|$)` would be good to exclude these:
- email, ex. `x@x.cz`
- `x @ y`, ex. `10 amps @ 1 volt`

staabm: thanks

src/Rules/Playground/PhpdocCommentRule.php
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @implements Rule<Node>
+ */
+final class PhpdocCommentRule implements Rule
+{
+
+	public function getNodeType(): string
+	{
+		return Node::class;
```
ondrejmirtes: I'd do it only for Node\Stmt.

ondrejmirtes (changes requested):

src/Rules/Playground/PhpdocCommentRule.php
staabm: tbh there are so many different node types, that I don't know what the differences are in this case.

adjusted it per request :)

src/Rules/Playground/PhpdocCommentRule.php
```diff
@@ -0,0 +1,52 @@
...
+use Nette\Utils\Strings;
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\VirtualNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function str_starts_with;
+
+/**
+ * @implements Rule<Node>
```
ondrejmirtes: Not synced with getNodeType

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you. Please add it to https://github.com/phpstan/phpstan/blob/2.1.x/playground-runner/playground.neon once it's built.

PR #4075, opened 2025-06-27
Improve property access when type hinting declared with @property

kayw-geek:
### Problem
Previously, accessing properties declared with `@property-read` annotations would incorrectly trigger " Access to an undefined property" error.

```PHP

use function PHPStan\dumpType;

/**
 * @property-read \stdClass $getCreator
 */
class Model
{
}

class Defaults
{
    public function defaults(Model $model): void
    {
        $creator = $model->getCreator;  // Access to an undefined property Model::$getCreator.

    }
}
```

Playground: [phpstan.org/r/7c5fd6b0-da88-4298-8e72-59bebdd48d08](https://phpstan.org/r/7c5fd6b0-da88-4298-8e72-59bebdd48d08)

### Solution
Enhanced property access checking to properly recognize properties declared via PHPDoc annotations.

### DIFF
--- a/src/Rules/Properties/AccessPropertiesCheck.php
+++ b/src/Rules/Properties/AccessPropertiesCheck.php
@@ -23,9 +23,11 @@ use PHPStan\Type\ErrorType;
 use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
+use function array_keys;
 use function array_map;
 use function array_merge;
 use function count;
+use function in_array;
 use function sprintf;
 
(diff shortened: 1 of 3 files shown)


canvural:
Isn't this expected behavior from PHP 8.2+? You need [`AllowDynamicProperties`](https://phpstan.org/blog/solving-phpstan-access-to-undefined-property#php-8.2%3A-add-%23%5Ballowdynamicproperties%5D) attribute.

kayw-geek:
@canvural Thank you for the clarification. I didn't know about this. I'll close this PR.

PR #4076, opened 2025-06-28
TableErrorFormatter: Link path in PhpStorm

vrana:
PhpStorm's internal terminal is very peculiar about paths it link. Pretty much only "at path:line" is linked: https://youtrack.jetbrains.com/issue/IJPL-103639/#focus=Comments-27-3994543.0-0.

I know that I can write my custom error formatter but I see a value in this working by default. Also replicating all features of TableErrorFormatter is not trivial in custom formatter.

Before:
image

After:
image

### DIFF
--- a/src/Command/ErrorFormatter/TableErrorFormatter.php
+++ b/src/Command/ErrorFormatter/TableErrorFormatter.php
@@ -130,7 +130,7 @@ final class TableErrorFormatter implements ErrorFormatter
 						$title = $this->relativePathHelper->getRelativePath($filePath);
 					}
 
-					$message .= "\n✏️  <href=" . OutputFormatter::escape($url) . '>' . $title . '</>';
+					$message .= "\nat  <href=" . OutputFormatter::escape($url) . '>' . $title . '</>';
 				}
 
 				if (
--- a/src/Command/ErrorsConsoleStyle.php
+++ b/src/Command/ErrorsConsoleStyle.php
@@ -100,7 +100,7 @@ final class ErrorsConsoleStyle extends SymfonyStyle
 		foreach ($rows as $i => $column) {
 			$columnRows = explode("\n", $column);
 			foreach ($columnRows as $k => $columnRow) {
-				if (str_starts_with($columnRow, '✏️')) {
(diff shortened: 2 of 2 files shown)


ondrejmirtes:
Thank you for the PR. I have an idea: Can we detect we're running in PhpStorm's internal terminal via some environment variable or other means, and output a clickable compatible clickable path regardless of user's `editorUrl` setting? Thank you!

vrana:
PhpStorm sets `TERMINAL_EMULATOR=JetBrains-JediTerm` which we could use. However, I SSH to a LXC container from PhpStorm and run PhpStan there. Thus this variable is not available.

ondrejmirtes:
Have you tried setting various `editorUrl` / `editorUrlTitle` values to make this work on your machine without changing PHPStan?

See the docs: https://phpstan.org/user-guide/output-format#opening-file-in-an-editor

I'm pretty sure you could add the word `at ` as part of those values.

vrana:
As visible at the screenshot, PhpStan doesn't create link if the line doesn't start with _at_.

ondrejmirtes:
I feel like the root issue in on JetBrains' shoulders and not on PHPStan's. I don't want to compromise the output of table error formatter for everyone because someone didn't fix a bug for 5 years on their side

vrana:
Makes sense.

vrana:
Would you accept a configuration parameter?

ondrejmirtes:
Configuration parameter for what exactly? Something that would remove all emojis from output?

vrana:
I need an `at` instead of the emoji. But I guess I can also live with `| sed 's~✏️~at~'`.

ondrejmirtes:
Could you set `TERMINAL_EMULATOR=JetBrains-JediTerm` in your custom terminal, and we could use that environment variable to detect that we're inside PhpStorm for both you and users that use it directly?

vrana:
OK. Can you reopen the pull request or should I create a new one?

vrana:
OK.

ondrejmirtes:
A new one, the implementation will be different. Thanks! 😊

PR #4078, opened 2025-06-30
Resolve ObjectType::getArraySize() with count() only if defined

vojtech-dobes:
Fix for phpstan/phpstan#13187

### DIFF
--- a/src/Type/ObjectType.php
+++ b/src/Type/ObjectType.php
@@ -946,6 +946,10 @@ class ObjectType implements TypeWithClassName, SubtractableType
 			return new ErrorType();
 		}
 
+		if ($this->hasMethod('count')->yes() === false) {
+			return IntegerRangeType::fromInterval(0, null);
+		}
+
 		return RecursionGuard::run($this, fn (): Type => $this->getMethod('count', new OutOfClassScope())->getOnlyVariant()->getReturnType());
 	}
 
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

uuf6429:
Based on the example in the test, I don't quite get how this change makes sense.

```php
interface Baz {}

function doBaz(Baz $baz): void {
     assertType('int<0, max>', count($baz));   // should be *ERROR* IMO
}
```

Even if `$baz` was an instance that implemented `\Countable` (and `Baz`), PHPStan should still complain at that point that it might not be countable. In this case, since there are no classes or interfaces implementing/extending `Baz`, it's also _definitely_ not countable.

Am I missing something?

ondrejmirtes:
@uuf6429 PHPStan is still going to report `Parameter #1 $value of function count expects array|Countable, Baz given.`: https://phpstan.org/r/b9e378ae-02fc-4788-910e-51a081a8fe4c

No reason to make the return type ErrorType.

uuf6429:
Hmm, I was confused on the basis of the test. If I understand it well such warnings are collected separately and in that case ignored?

I must admit that I tried the playground too, but assumed the change might not have been released (although I suppose that's what "bleeding edge" is for).

Anyway, all good, thanks for checking.

PR #4079, opened 2025-07-02
Correctly retrieve hasSideEffects metadata for inherited methods

zonuexe:
Revert e97439c5987fc7747b68d3617a128fad68102642 to fix https://github.com/phpstan/phpstan/issues/13201


stubs/BackedEnum.stub
```diff
@@ -0,0 +1,16 @@
...
+interface BackedEnum extends UnitEnum
+{
+	/**
+	 * @return static
+	 * @pure
+	 */
+	public static function from(int|string $value): static;
+
+	/**
+	 * @return ?static
```
zonuexe: While debugging with @nsfisis, he pointed out that this DocBlock causes the return value to be a `mixed` type.

Although it is redundant, in the scope of this PR it makes sense to duplicate the type declarations in the PHPDoc.

ondrejmirtes (changes requested):
Can you please debug why these lines are not sufficient? https://github.com/phpstan/phpstan-src/blob/599ac2ba2a7a420396f65142878ece9707a3c086/resources/functionMetadata.php#L18-L19 - they should be.

zonuexe:
I discovered that the functionMetadataMap wasn't being accessed in this process, so in this PR I only fixed the metadata.

I agree that this should be the case, so I'll investigate the underlying issue.

ondrejmirtes:
The code path that creates these methods should use the metadata too.

There's a chance it will fix more than just enums.

zonuexe:
<details><summary>Here are some notes I've taken while looking at the code:</summary>

`PHPStan\Reflection\ExtendedMethodReflection::isPure()` is the authoritative source.

I checked it by injecting the following code into `PHPStan\PhpDoc\PhpDocBlock::resolveMethodPhpDocBlockFromClass()`:

```diff
modified   src/PhpDoc/PhpDocBlock.php
@@ -343,6 +343,8 @@ final class PhpDocBlock
 			if ($parentReflection->isPrivate()) {
 				return null;
 			}
+			var_dump(["{$classReflection->getName()}::{$name}" => $parentReflection->isPure()->describe()]);
+			$isPure = $parentReflection->isPure()->maybe() ? null : $parentReflection->isPure()->yes();
 
 			$classReflection = $parentReflection->getDeclaringClass();
 			$traitReflection = null;
```

`ExtendedMethodReflection` is degenerated by `$classReflection = $parentReflection->getDeclaringClass()` before being passed to the `PhpDocBlock`, so there's no way to get it out of here.

To bring out `$isPure` from here would require breaking backwards compatibility and extending `PhpDocBlock` or `MethodReflection`. But, `PhpDocBlock::isPure()` is unnatural and should not be extended.

Again, we need to pass an appropriate `$isPure` directly to `PHPStan\Reflection\Php\PhpMethodReflection` from `PHPStan\Reflection\Php\PhpClassReflectionExtension::createUserlandMethodReflection()`.

Here's the problem: `PhpDocBlock::resolveMethodPhpDocBlockFromClass()` calls `ClassReflection::getMethod()` to get reflection of the parent class, which relies on `PhpClassReflectionExtension` contained in `MethodsClassReflectionExtension[] $methodsClassReflectionExtensions`.

Therefore, injecting the following code will result in an infinite loop due to recursive calls.

```diff
modified   src/Reflection/Php/PhpClassReflectionExtension.php
@@ -839,7 +839,16 @@ final class PhpClassReflectionExtension
 		}
 		$isInternal = $resolvedPhpDoc->isInternal();
 		$isFinal = $resolvedPhpDoc->isFinal();
-		$isPure = $resolvedPhpDoc->isPure();
+		$isPure = null;
+		if ($actualDeclaringClass->hasNativeMethod($methodReflection->getName())) {
+			var_dump("{$actualDeclaringClass->getName()}::{$methodReflection->getName()}");
+			$parentReflection = $actualDeclaringClass->getNativeMethod($methodReflection->getName());
+			if (!$parentReflection->isPrivate()) {
+				$isPure = $parentReflection->isPure()->maybe() ? null : $parentReflection->isPure()->yes();
+			}
+		}
+
+		$isPure ??= $resolvedPhpDoc->isPure();
 		$asserts = Assertions::createFromResolvedPhpDocBlock($resolvedPhpDoc);
 		$acceptsNamedArguments = $resolvedPhpDoc->acceptsNamedArguments();
 		$selfOutType = $resolvedPhpDoc->getSelfOutTag() !== null ? $resolvedPhpDoc->getSelfOutTag()->getType() : null;
```

We can get the metadata directly as follows, but first we need to identify the parent class name:

```php
			if ($this->signatureMapProvider->hasMethodMetadata($declaringClassName, $methodReflection->getName())) {
				$hasSideEffects = TrinaryLogic::createFromBoolean($this->signatureMapProvider->getMethodMetadata($declaringClassName, $methodReflection->getName())['hasSideEffects']);
			} else {
				$hasSideEffects = TrinaryLogic::createMaybe();
			}
```

Checking for side effects from ancestor metadata.
</details>

ondrejmirtes:
Perfect, thank you!

PR #4081, opened 2025-07-04
Resolve self/parent/static in Closure::bind() scope

zonuexe:
Resolves `self` / `parent` / `static` against the `Closure::bind()` scope (the 3rd argument) inside the bound closure body, for both type inference and the related rules.

Rebased onto `2.2.x` and reworked: the original inline patch in `MutatingScope::getType()` no longer exists after expression handling was split into `ExprHandler` services, so the logic now lives where each expression/rule resolves its class scope.

### What works now

Inside `Closure::bind(fn () => …, $newThis, Scope::class)`:

- **Class constants** — `self::CONST`, `parent::CONST` resolve against the bound scope.
- **`::class`** — `self::class`, `parent::class`.
- **Static method calls** — `self::method()`, `parent::method()` return types.
- **Static property access** — `self::$prop`, `parent::$prop` types.
- **Instantiation** — `new self`, `new parent`, `new static`.
- **Rules** no longer emit `Using self outside of class scope.` etc. and check accessibility against the bound class: `ClassConstantRule`, `StaticMethodCallCheck`, `AccessStaticPropertiesCheck`, `InstantiationRule`.

This works even when `Closure::bind()` is called outside a class context (the closure body type is inferred in the enclosing scope, where the closure-bind scope classes are otherwise unavailable — hence the class-name nodes are annotated by `ClosureBindArgVisitor`).

### Approach

- `ClosureBindArgVisitor` annotates `self`/`parent`/`static` class-name nodes **inside the closure body only** (1st argument) with the bind scope argument. The `$newThis`/`$newScope` arguments are intentionally not rescoped — this keeps `Closure::bind(self::cond() ? … : …, …)` correct (see #6319).
- `ClosureBindScopeResolver` (new) resolves that annotation to a `ClassReflection`, shared by `ClassConstFetchHandler` and the rules.
- `MutatingScope::resolveName()` / `resolveTypeByName()` honour the annotation so static calls / properties / `new` resolve without touching each handler.

### Issues

- Advances phpstan/phpstan#5987 — `self::`/`parent::` constant, `::class`, static method, and private-member cases now resolve. (Remaining: `static::` late static binding and the bound `$this` object type — a separate concern, overlaps #2972.)
- Regression coverage for phpstan/phpstan#7675, #6319, #4865 (all already fixed; unchanged here).

### Not included (follow-ups)

- `Closure::bindTo()` (phpstan/phpstan#8569) — separate call, separate visitor path.
- `static::` LSB + bound `$this` type (phpstan/phpstan#5987 remainder) — overlaps the stalled #2972; better decided after review of this approach.

### DIFF
--- a/src/Analyser/ClosureBindScopeResolver.php
+++ b/src/Analyser/ClosureBindScopeResolver.php
@@ -1,0 +1,55 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Analyser;
+
+use PhpParser\Node\Expr;
+use PhpParser\Node\Name;
+use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Parser\ClosureBindArgVisitor;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\ReflectionProvider;
+use function count;
+
+/**
+ * Resolves the class scope that a `self`/`parent`/`static` class-name node is bound to
+ * by a surrounding `Closure::bind()` call. {@see ClosureBindArgVisitor} annotates such
+ * nodes with the bind scope argument (the 3rd argument of `Closure::bind()`).
+ */
+#[AutowiredService]
+final class ClosureBindScopeResolver
+{
+
+	public function __construct(
+		private ReflectionProvider $reflectionProvider,
+	)
+	{
+	}
+
+	/**
+	 * Returns the class the given class-name node is bound to via `Closure::bind()`, or
+	 * null when the node is not inside a bound closure or the bind scope argument does not
+	 * resolve to a single known class (e.g. the default "static" scope).
+	 */
+	public function resolveScopeClass(Scope $scope, Name $class): ?ClassReflection
+	{
+		$scopeArg = $class->getAttribute(ClosureBindArgVisitor::SCOPE_ATTRIBUTE_NAME);
+		if (!$scopeArg instanceof Expr) {
+			// Either the node is not inside a bound closure, or the attribute is null for
+			// the default "static" scope. Both keep the enclosing class.
+			return null;
+		}
+
+		$objectClassNames = $scope->getType($scopeArg)->getClassStringObjectType()->getObjectClassNames();
+		if (count($objectClassNames) !== 1) {
+			return null;
+		}
+
+		$className = $objectClassNames[0];
+		if (!$this->reflectionProvider->hasClass($className)) {
+			return null;
+		}
+
+		return $this->reflectionProvider->getClass($className);
+	}
+
+}
--- a/src/Analyser/ExprHandler/ClassConstFetchHandler.php
+++ b/src/Analyser/ExprHandler/ClassConstFetchHandler.php
@@ -5,6 +5,7 @@ namespace PHPStan\Analyser\ExprHandler;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\ClassConstFetch;
 use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
 use PhpParser\Node\Stmt;
 use PHPStan\Analyser\ExpressionContext;
 use PHPStan\Analyser\ExpressionResult;
@@ -18,10 +19,14 @@ use PHPStan\Analyser\SpecifiedTypes;
 use PHPStan\Analyser\TypeSpecifier;
 use PHPStan\Analyser\TypeSpecifierContext;
 use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Parser\ClosureBindArgVisitor;
+use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\InitializerExprTypeResolver;
+use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use function array_merge;
+use function count;
 
 /**
  * @implements ExprHandler<ClassConstFetch>
@@ -33,6 +38,7 @@ final class ClassConstFetchHandler implements ExprHandler
 	public function __construct(
 		private InitializerExprTypeResolver $initializerExprTypeResolver,
 		private ExpressionResultFactory $expressionResultFactory,
+		private ReflectionProvider $reflectionProvider,
 	)
 	{
 	}
@@ -48,12 +54,49 @@ final class ClassConstFetchHandler implements ExprHandler
 			return new MixedType();
 		}
 
+		$classReflection = $scope->isInClass() ? $scope->getClassReflection() : null;
+		$bindScopeReflection = $this->resolveClosureBindScopeReflection($scope, $expr->class);
+		if ($bindScopeReflection !== null) {
+			$classReflection = $bindScopeReflection;
+		}
+
 		return $this->initializerExprTypeResolver->getClassConstFetchTypeByReflection(
 			$expr->class,
 			$expr->name->name,
-			$scope->isInClass() ? $scope->getClassReflection() : null,
+			$classReflection,
 			static fn (Expr $e): Type => $scope->getType($e),
 		);
+	}
+
+	/**
+	 * Resolves the `Closure::bind()` scope class annotated on a `self`/`parent`/`static`
+	 * class name node by {@see ClosureBindArgVisitor}. Returns null when the node is not
+	 * inside a bound closure or the scope argument does not resolve to a single known class.
+	 */
+	private function resolveClosureBindScopeReflection(MutatingScope $scope, Expr|Name $class): ?ClassReflection
+	{
+		if (!$class instanceof Name || !$class->hasAttribute(ClosureBindArgVisitor::SCOPE_ATTRIBUTE_NAME)) {
+			return null;
+		}
+
+		$scopeArg = $class->getAttribute(ClosureBindArgVisitor::SCOPE_ATTRIBUTE_NAME);
+		if (!$scopeArg instanceof Expr) {
+			// null attribute means the default "static" scope: keep the enclosing class.
+			return null;
+		}
+
+		$scopeArgType = $scope->getType($scopeArg);
+		$objectClassNames = $scopeArgType->getClassStringObjectType()->getObjectClassNames();
+		if (count($objectClassNames) !== 1) {
+			return null;
+		}
+
+		$className = $objectClassNames[0];
+		if (!$this->reflectionProvider->hasClass($className)) {
+			return null;
+		}
+
+		return $this->reflectionProvider->getClass($className);
 	}
 
 	public function processExpr(NodeScopeResolver $nodeScopeResolver, Stmt $stmt, Expr $expr, MutatingScope $scope, ExpressionResultStorage $storage, callable $nodeCallback, ExpressionContext $context): ExpressionResult
--- a/src/Analyser/ExprHandler/ClassConstFetchHandler.php
+++ b/src/Analyser/ExprHandler/ClassConstFetchHandler.php
@@ -7,6 +7,7 @@ use PhpParser\Node\Expr\ClassConstFetch;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
 use PhpParser\Node\Stmt;
+use PHPStan\Analyser\ClosureBindScopeResolver;
 use PHPStan\Analyser\ExpressionContext;
 use PHPStan\Analyser\ExpressionResult;
 use PHPStan\Analyser\ExpressionResultFactory;
@@ -19,14 +20,10 @@ use PHPStan\Analyser\SpecifiedTypes;
 use PHPStan\Analyser\TypeSpecifier;
 use PHPStan\Analyser\TypeSpecifierContext;
 use PHPStan\DependencyInjection\AutowiredService;
-use PHPStan\Parser\ClosureBindArgVisitor;
-use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\InitializerExprTypeResolver;
-use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use function array_merge;
-use function count;
 
 /**
  * @implements ExprHandler<ClassConstFetch>
@@ -38,7 +35,7 @@ final class ClassConstFetchHandler implements ExprHandler
 	public function __construct(
 		private InitializerExprTypeResolver $initializerExprTypeResolver,
 		private ExpressionResultFactory $expressionResultFactory,
-		private ReflectionProvider $reflectionProvider,
+		private ClosureBindScopeResolver $closureBindScopeResolver,
 	)
 	{
 	}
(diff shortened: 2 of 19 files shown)


staabm:
@SanderMuller please review :)

SanderMuller:
Reviewed. I ran it rather than only reading it, and the approach holds up: the rescoping is confined to the closure body, everything the description advertises resolves, and the cases that fall outside it fail safe.

Verified against `81e06a583` at level 9, with `Closure::bind(fn () => …, null, Target::class)` and `Target extends Base`:

| expression | before | after |
| --- | --- | --- |
| `self::C` | `Using self outside of class scope.` | `'target'` |
| `parent::C` | error | `'base'` |
| `self::SECRET` (private const) | error | `42` |
| `self::$prop` (private static) | error | `string` |
| `self::priv()` (private static) | error | `int` |
| `new self()` | error | `P\Target` |
| `self::class` | error | `'P\Target'` |

A closure nested inside a bound closure inherits the bind scope and resolves to `'target'`, which matches PHP: the inner closure is created while the outer one already carries the rebound scope.

The confinement claim is real, which was the part I most wanted to check. Inside a method of `Ctx`, `self::cond()` in the first argument stays `'ctx'`, `self::obj()` in the `$newThis` argument stays `P\Ctx`, `self::C` inside the body is `'target'`, and `self::C` after the call is `'ctx'` again. So #6319 stays fixed.

Gates on the head: full suite green (21288 tests, 96849 assertions), self-analysis clean. Nothing under `turbo-ext`, and `MutatingScope` is not shadowed by it, so there is no native mirror to keep in step.

Three cases sit outside the visitor. All of them decline to rescope rather than picking the wrong class, which is the right direction, but two are worth knowing about:

1. Named arguments in non-positional order. `Closure::bind(closure: fn () => self::C, newScope: Target::class, newThis: null)` still reports `Using self outside of class scope.`, because `getArgs()` is read positionally, so `$args[2]` is `newThis` there. A non-class expression lands in the resolver and it returns null, so nothing is misresolved. `ArgumentsNormalizer` would cover it if you want it, otherwise a comment saying positional-only would help.
2. An object as the scope, `Closure::bind(fn () => self::C, $o, $o)`, which PHP accepts. `getClassStringObjectType()` yields no class names for an object type, so it is left alone.
3. A closure not written inline at the call site. Inherent to the approach, no complaint.

Three small code notes:

- `MutatingScope::resolveClosureBindScopeClassName()` repeats what `ClosureBindScopeResolver::resolveScopeClass()` already does. Is that to avoid a dependency cycle in the scope, or could the scope take the resolver?
- That method carries two doc blocks, the prose one followed by `/** @return non-empty-string|null */`. Only the second attaches, so the explanation is orphaned. Worth merging into one.
- The annotation path now sits beside the existing `enterClosureBind()` / `inClosureBindScopeClasses` one. The description explains why the scope-based path cannot reach the out-of-class case, so both earn their place, but a line saying which one owns which case would save the next reader the trip.

The single red job is `Compile Turbo Extension (8.4, macos-x86_64)` at 35m0s, which is the runner timeout on a compile step, and this PR has no C++ in it.


On performance, since the visitor calls `spl_object_id()` and an array lookup on every node entered and left, I measured instead of assuming. Three interleaved rounds on a doctrine/symfony codebase: 139.0s median on the base against 138.6s with the PR, inside a within-build spread of about 4s. No measurable cost.

staabm:
> 1. Named arguments in non-positional order. `Closure::bind(closure: fn () => self::C, newScope: Target::class, newThis: null)` still reports `Using self outside of class scope.`, because `getArgs()` is read positionally, so `$args[2]` is `newThis` there. A non-class expression lands in the resolver and it returns null, so nothing is misresolved. `ArgumentsNormalizer` would cover it if you want it, otherwise a comment saying positional-only would help.

this one should be fixed

> * `MutatingScope::resolveClosureBindScopeClassName()` repeats what `ClosureBindScopeResolver::resolveScopeClass()` already does. Is that to avoid a dependency cycle in the scope, or could the scope take the resolver?

if possible it should be de-duplicated

> * That method carries two doc blocks, the prose one followed by `/** @return non-empty-string|null */`. Only the second attaches, so the explanation is orphaned. Worth merging into one.

should be fixed

zonuexe:
I addressed all three notes:

- Named arguments: the visitor now resolves closure/newThis/newScope by name and position. The tests cover reordered and partially-named calls.
- The duplication: the scope already holds the container, so the method now fetches ClosureBindScopeResolver from it behind the cheap attribute pre-check. No duplication left, the factories stay untouched, and the hot path still costs one `hasAttribute()` call. To answer your question: no cycle, I had missed the shortcut.
- I merged the orphaned doc block and added the ownership note: `enterClosureBind()` owns member accessibility while NodeScopeResolver processes the closure body; the annotations own the cases where the body's type is inferred in the enclosing scope.

The object-as-scope case keeps declining to rescope, and the non-inline closure is inherent to the approach, as you said.

SanderMuller:
Notes 2 and 3 are done, and I agree with both resolutions. Fetching the resolver off the container behind the `hasAttribute()` pre-check is fine here — `MutatingScope` already does the same thing on the `getType()` path (`ExprHandlerRegistry::resolve($node, $this->container)`, line 1109), so it is not a new pattern in this class. The merged doc block reads well, and the split it spells out is the right way to describe it.

Note 1 is only half fixed, though, and the remaining half is the part the original report was about.

The visitor now resolves the scope argument by name, so `self` *resolves* correctly. But member accessibility is decided elsewhere, and that path is still positional, so the false positive does not go away — it changes message. Six forms, all of which should be silent:

| # | call | before | after |
|---|---|---|---|
| 1 | `bind($c, null, Target::class)` | clean | clean |
| 2 | `bind(closure: $c, newThis: null, newScope: Target::class)` | clean | clean |
| 3 | `bind(closure: $c, newScope: Target::class, newThis: null)` | `outOfClass.self` | `classConstant.protected` |
| 4 | `bind(newScope: Target::class, closure: $c, newThis: null)` | `outOfClass.self` | `classConstant.protected` |
| 5 | `bind($c, null, newScope: Target::class)` | clean | clean |
| 6 | `bind($c, newScope: Target::class, newThis: null)` | `outOfClass.self` | `classConstant.protected` |

(`$c` is `static fn () => self::C`, `Target::C` protected, level 8, analysed outside any class.)

Worth noting rows 1, 2 and 5 were already clean before the commit: named arguments in canonical order line up with the positions anyway, so `$args[0]`/`$args[2]` happened to be right. The forms that were broken are exactly the ones still broken. At rule level the commit's only observable effect is which error you get.

So it works exactly when the written order happens to match the positional order. Two separate positional readers, both pre-existing:

- `NodeScopeResolver::processArgs()` applies `$closureBindScopeFactory` only at `$i === 0`, so form 4 never enters the bind scope at all.
- The factory in `StaticCallHandler` reads `$expr->getArgs()[1]` and `[2]` for `newThis`/`newScope`, so forms 3 and 6 read the scope as `$newThis` and `null` as the scope, and `$scopeClasses` ends up empty.

Since it is false-positive-for-false-positive on the same lines, I would not call it a regression. But I do not think it is fixed either, and the current test shape hides that. Type resolution *is* genuinely fixed by the commit — `dumpType()` on the six calls gives `*ERROR*` for 3, 4 and 6 before and `'c'` for all six after — so the new `nsrt` file passes, while a rule test over the same forms fails today.

Two ways out, your call which fits:

1. Finish it. `StaticCallHandler` computes the combined acceptor a few lines above the factory, so `ArgumentsNormalizer::reorderStaticCallArguments()` can be applied there and `[1]`/`[2]` become correct — the closure just needs to capture the acceptor as well as `$expr`. Then the `$i === 0` gate becomes "the arg carrying `ClosureBindArgVisitor::ATTRIBUTE_NAME`" instead of a position.
2. Scope it back. Keep the visitor positional-only, document it, and the behaviour is at least consistent with the rest of the bind handling.

Either way I would add forms 3, 4 and 6 to `ClassConstantRuleTest` — the PR already touches that class, and it is the level the report was written at.

One smaller thing, pre-existing and not blocking: `ParametersAcceptorSelector` (around line 535) also reads `$args[0]` for the `ATTRIBUTE_NAME` check. Before this PR the attribute was always on `$args[0]`, so a reordered call could match it against the wrong variable; now the attribute moves to the real closure arg, so that check just no-ops instead. An improvement by accident, but it is the third positional reader if you go with option 1.

On the two red checks: both are the `phpbench` regression gate, and I do not think they are attributable to this PR. Nearly every variant shifts +4% to +18% at once, which is the signature of the committed baseline rather than a hot spot. I ran the two failing variants locally, interleaved, three rounds against the merge base: `bug-13352` 1.556s vs 1.539s and `bug-14624` 1.448s vs 1.434s (medians, base vs PR), so flat to marginally faster, against a spread of up to 3.7% inside a single build. CI reported +17.5% for `bug-14624`. For what it is worth, the same gate went green and then red on an unrelated branch of mine within four hours, with variants reporting up to +787%.

Gates I ran on PR head, all green: the four rule suites the PR touches (130 tests), the `nsrt` closure-bind datasets including the new file, self-analysis, and `phpcs`.

SanderMuller:
Correction to my previous comment: I got the root cause partly wrong, and the fix is smaller than I made it sound.

I claimed two positional readers, one of them the `$i === 0` gate in `processArgs()`. That is not it. `StaticCallHandler` already normalizes before it calls `processArgs()`:

```php
$normalizedExpr = ArgumentsNormalizer::reorderStaticCallArguments($parametersAcceptor, $expr) ?? $expr;
$argsResult = $nodeScopeResolver->processArgs(..., $normalizedExpr, ...);
```

so `$i === 0` really is the closure argument even for a reordered call, and no change is needed there. The single cause is that the factory closure a hundred lines above captures the **un-normalized** `$expr`, so its `getArgs()[1]`/`[2]` are the written positions. That accounts for all three broken forms, including form 4.

Which makes it a one-line fix:

```diff
-$closureBindScopeFactory = static function (MutatingScope $boundScope) use ($expr): MutatingScope {
+$closureBindScopeFactory = static function (MutatingScope $boundScope) use ($expr, $parametersAcceptor): MutatingScope {
+	$expr = ArgumentsNormalizer::reorderStaticCallArguments($parametersAcceptor, $expr) ?? $expr;
```

With that on top of your branch, all six forms from the table are clean, and the full suite passes (21290 tests, 96853 assertions). `ArgumentsNormalizer` is already imported in that file.

Both halves are needed, which I checked by reverting each: with your visitor commit but without the normalization, forms 3, 4 and 6 report `classConstant.protected`; with the normalization but without your visitor commit, the same three go back to `outOfClass.self`. So your commit is necessary, it just was not sufficient on its own.

Also withdrawing the "third positional reader" framing on `ParametersAcceptorSelector` — the `$args[0]` indexing there is a separate pre-existing pattern shared with other arg visitors, and nothing to do with this bug. Sorry for the noise.

Everything else in the previous comment stands: the six-form table, notes 2 and 3 being done, and the bench reds not being attributable.

zonuexe:
You are right — I had fixed the resolution and left the accessibility path positional. Both readers are gone now:

- The bind-scope factory is built from the normalized call (the acceptor was already computed a few lines up, as you said), so `$newThis`/`$newScope` are found at their parameter positions.
- The enter/restore gates in `processArgs()` now key on the argument carrying the `closureBindArg` attribute instead of `$i === 0`.

All six forms are silent at level 8, and `ClassConstantRuleTest` now covers them; without the fix the test fails on forms 3, 4 and 6 exactly.

I left the `ParametersAcceptorSelector` reader alone since it pre-dates this PR and now no-ops as you observed.

zonuexe:
Confirmed. You are right that the gate in `processArgs()` is not part of it. I put `$i === 0` back on top of the normalization and checked: all six forms stay clean, because `processArgs()` receives the already-normalized call.

I dropped my change there along with the import it needed, and kept only the normalization inside the factory. The source diff is now +5/-1 in `StaticCallHandler`, plus the rule test.

SanderMuller (approved):
Confirmed, and thanks for chasing the last one down properly rather than keeping my wrong suggestion. Dropping the `processArgs()` change is right: it receives the already-normalized call, so the `$i === 0` gate was never part of the problem.

What I verified on `c469c18b8`:

- **The rule test pins the regression.** Reverting only `StaticCallHandler.php` and keeping the test fails with exactly three errors, on the three reordered forms (`Access to protected constant C of class ClosureBindNamedArguments\Target` at lines 20, 21 and 23), and passes with the fix. So it fails for the right reason, not incidentally.
- All six argument orders are clean, and `dumpType()` gives `'c'` for every one of them.
- Full suite green (21292 tests, 96857 assertions), self-analysis clean, `phpcs` clean on both changed files.
- **Perf:** the normalization runs inside the bind-scope factory, so worst case is one `reorderStaticCallArguments()` per `Closure::bind()` call site with an inline closure, and it early-returns before any work when the call has no named arguments. Measured it anyway on a file with 1500 `Closure::bind()` call sites, two thirds of them named, interleaved three rounds against the parent commit: 17.06s vs 16.86s median. No cost.

On the four red checks, none of them look attributable to this PR:

- `Test (PHP 8.5)` is the phpbench gate, and it is failing on `2.2.x` itself.
- `Run with Turbo Extension (macos-latest, 8.5, nts, make phpstan)` is a worker OOM at the 450M limit; it fails the same way on unrelated branches right now.
- `Integration tests (windows-latest)` is `AnalyseCommandTest::testConfigurationAutoDiscovery` failing with `Nette\IOException: Unable to include ...\Temp/phpstan/cache/nette.configurator/Container_*.php`. It passed on `dc5df8eb8` eleven minutes earlier, and the only difference since is four lines in `StaticCallHandler` plus a rule test that is not in that suite.
- `extension-tests / phpstan-doctrine (8.0)` is `EntityColumnRuleTest::testRuleWithAllowedNullableProperty` reporting one extra `DateTime`/`DateTimeImmutable` mapping error for `$six`. That suite passed on `5444d4a56`, which already contained every analysis change in this PR, and it went fail → pass → fail across pushes on this branch, so it is not tracking the diff.

I would still get a maintainer to glance at that last one before merge, since it is the only red that is a behaviour difference rather than infrastructure, but I could not construct any path from argument normalization inside `Closure::bind()` to a Doctrine column type mapping.

LGTM.

src/Analyser/ClosureBindScopeResolver.php
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * Returns the class the given class-name node is bound to via `Closure::bind()`, or
+	 * null when the node is not inside a bound closure or the bind scope argument does not
+	 * resolve to a single known class (e.g. the default "static" scope).
+	 */
+	public function resolveScopeClass(Scope $scope, Name $class): ?ClassReflection
+	{
+		if (!$class->hasAttribute(ClosureBindArgVisitor::SCOPE_ATTRIBUTE_NAME)) {
+			return null;
+		}
```
staabm: can't we drop this hasAttribute call, without affecting the logic?

zonuexe: Dropped it. `getAttribute()` returns null when the key is absent, and the `instanceof Expr` check below already rejects that, so the two cases collapse into one. I folded the comment to say both: not inside a bound closure, or a null attribute for the default `static` scope.

I kept the `hasAttribute()` in `MutatingScope::resolveClosureBindScopeClass()` though. That one is not a logic guard but a pre-check that keeps the common path free of the container lookup behind it.

SanderMuller (approved):

PR #4082, opened 2025-07-04
wip: Fix ParserNodeTypeToPHPStanType::resolve() so that it can resolve some of static/self/parent

nsfisis:
TODO

Related: this issue was found while debugging #4079

### DIFF
--- a/src/Type/ParserNodeTypeToPHPStanType.php
+++ b/src/Type/ParserNodeTypeToPHPStanType.php
@@ -23,6 +23,24 @@ final class ParserNodeTypeToPHPStanType
 	{
 		if ($type === null) {
 			return new MixedType();
+		} elseif ($type instanceof Identifier && $type->isSpecialClassName()) {
+			// !!!
+			$typeClassName = (string) $type;
+			$lowercasedClassName = strtolower($typeClassName);
+			if ($classReflection !== null && in_array($lowercasedClassName, ['self', 'static'], true)) {
+				if ($lowercasedClassName === 'static') {
+					return new StaticType($classReflection);
+				}
+				$typeClassName = $classReflection->getName();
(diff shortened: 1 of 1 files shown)


nsfisis:
`\PHPStan\Type\ParserNodeTypeToPHPStanType::resolve()` returns `MixedType` when an `Identifier` of `static` is passed as `$type`.

In contrast, `\PHPStan\BetterReflection\Reflection\ReflectionClass` hardcodes types for `BackedEnum::from()`, and it returns `new Identifier('static')`. `BackedEnum::tryFrom()` is the same except that it returns a nullable type.

https://github.com/ondrejmirtes/BetterReflection/blob/61b25baaca1ea904447f46d975a4ae7c99b722e6/src/Reflection/ReflectionClass.php#L695-L705

I couldn't determine which of the following approaches would be more appropriate:

* `ParserNodeTypeToPHPStanType::resolve()` should handle `Identifier`s of `static` in the same way it handles `Name`s of `static`. (this pull request)
* `ReflectionClass` should change the return type of `BackedEnum::from()` from `Identifier` of `static` to `Name` of `static`.

ondrejmirtes:
Hi, where you're encountering this problem? Without a test I don't know what you're trying to fix.

nsfisis:
> Hi, where you're encountering this problem? Without a test I don't know what you're trying to fix.

This pull request addresses a potential issue that we encountered while debugging #4079. However, the original pull request #4079 has changed its approach to fix the problem, so this pull request could become unnecessary.

I will keep this pull request as a draft until the root cause and fix for #4079 become clear.

nsfisis:
Due to the change in the implementation approach of PR #4079, it appears that this potential issue will no longer cause any actual problems. I will close this PR.

PR #4084, opened 2025-07-06
Simplify property memoization for Flyweight pattern by replacing it with `??=`

zonuexe:
Now that PHP 7.2 has been dropped, this pattern can be replaced.

### DIFF
--- a/build/PHPStan/Build/MemoizationPropertyRule.php
+++ b/build/PHPStan/Build/MemoizationPropertyRule.php
@@ -1,0 +1,150 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Build;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\AssignOp\Coalesce;
+use PhpParser\Node\Expr\BinaryOp\Identical;
+use PhpParser\Node\Expr\ConstFetch;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\StaticPropertyFetch;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\Expression;
+use PhpParser\Node\Stmt\If_;
+use PhpParser\Node\VarLikeIdentifier;
+use PHPStan\Analyser\Scope;
+use PHPStan\File\FileHelper;
+use PHPStan\Node\InClassMethodNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use function count;
+use function dirname;
+use function is_string;
+use function str_starts_with;
+use function strcasecmp;
+
+/**
+ * @implements Rule<If_>
+ */
+final class MemoizationPropertyRule implements Rule
+{
+
+	public function __construct(private FileHelper $fileHelper, private bool $skipTests = true)
+	{
+	}
+
+	public function getNodeType(): string
+	{
+		return If_::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$ifNode = $node;
+
+		if (count($ifNode->stmts) !== 1
+			|| !$ifNode->stmts[0] instanceof Expression
+			|| count($ifNode->elseifs) !== 0
+			|| $ifNode->else !== null
+			|| !$ifNode->cond instanceof Identical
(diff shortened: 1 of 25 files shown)


ondrejmirtes:
Would be great to have an auto-fixable rule that would enforce this under https://github.com/phpstan/phpstan-src/tree/2.1.x/build/PHPStan/Build. Do you think it'd be easy for you to implement that? And verify it by actually using it to make these changes. And compare it to your manual changes 😊

(Yeah, RuleErrorBuilder has `fixNode` method on 2.1.x, a few rules already use it.)

zonuexe:
I hadn't noticed that they added `RuleErrorBuilder::fixNode()` yet, but this is very interesting! It probably looks very easy, so I'll give it a try.

ondrejmirtes:
Yeah, just detect if === null and assignment to the same thing in the only statement.

zonuexe:
@ondrejmirtes 
I think this PR now works perfectly, applying the rules replaced all the things I had missed.

The refactoring feature integrated into PHPStan is a game changer. 🎉

build/PHPStan/Build/MemoizationPropertyRule.php
```diff
@@ -0,0 +1,112 @@
...
+			return [];
+		}
+
+		if (!$ifNode instanceof If_
+			|| count($ifNode->stmts) !== 1
+			|| !$ifNode->stmts[0] instanceof Expression
+			|| count($ifNode->elseifs) !== 0
+			|| $ifNode->else !== null
+			|| !$ifNode->cond instanceof Identical
+			|| !$ifNode->cond->left instanceof PropertyFetch
```
staabm, line 69: I think you can do the same for `StaticPropertyFetch`

build/PHPStan/Build/MemoizationPropertyRule.php
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		$stmts = $methodNode->getStmts();
+		if ($stmts === null || count($stmts) !== 2) {
+			return [];
+		}
+
+		[$ifNode, $returnNode] = $stmts;
+		if (!$returnNode instanceof Return_
+			|| !$returnNode->expr instanceof PropertyFetch
```
staabm: I think you can do the same for `StaticPropertyFetch`

zonuexe: Thanks for the suggestion, I implemented it.

build/PHPStan/Build/MemoizationPropertyRule.php
```diff
@@ -0,0 +1,182 @@
...
+final class MemoizationPropertyRule implements Rule
+{
+
+	public function __construct(private FileHelper $fileHelper, private bool $skipTests = true)
+	{
+	}
+
+	public function getNodeType(): string
+	{
+		return InClassMethodNode::class;
```
ondrejmirtes: I think the rule should apply to `If_`, not to a method body.

ondrejmirtes (changes requested):
I described the logic as following:

> Yeah, just detect if === null and assignment to the same thing in the only statement.

So even if there's `if` in the middle of a function body, this rule could detect and change it.

It doesn't matter if there's `$this->foo = $foo;` or `return $this->foo = $foo;`. The rule should be able to handle both.

build/PHPStan/Build/MemoizationPropertyRule.php
zonuexe: Ok, it has been reflected in the rule.

build/PHPStan/Build/MemoizationPropertyRule.php
```diff
@@ -0,0 +1,150 @@
...
+
+		if ($this->areNodesNotEqual($ifNode->cond->left, [$ifThenNode->var])) {
+			return [];
+		}
+
+		if ($this->skipTests && str_starts_with($this->fileHelper->normalizePath($scope->getFile()), $this->fileHelper->normalizePath(dirname(__DIR__, 3) . '/tests'))) {
+			return [];
+		}
+
+		$errorBuilder = RuleErrorBuilder::message('Memoization property can be simplified.')
```
ondrejmirtes: The message could be better. Something like "This initializing if statement can be replaced with null coalescing assignment operator (??=)."

ondrejmirtes (changes requested):
Otherwise great 👍

build/PHPStan/Build/MemoizationPropertyRule.php
zonuexe: I changed the message and squashed the commits!

ondrejmirtes:
Thank you!

PR #4085, opened 2025-07-10
Update locale_get_display_x() return types

cfahner:
For issue [#12942](https://github.com/phpstan/phpstan/issues/12942) I updated the signatures of the following functions:

* `locale_get_display_language()` and `Locale::getDisplayLanguage()`
* `locale_get_display_name()` and `Locale::getDisplayName()`
* `locale_get_display_region()` and `Locale::getDisplayRegion()`
* `locale_get_display_script()` and `Locale::getDisplayScript()`
* `locale_get_display_variant()` and `Locale::getDisplayVariant()`

The `_language()` and `_name()` variants never seem to return an empty string (at least I could not find any case for which they did), so I left them at `non-empty-string`.

The `_region()` function returns an empty string when you give it a language code without a region.

When you pass in a very long string these methods all return `false` ([C source code](https://heap.space/xref/PHP-8.4/ext/intl/locale/locale_methods.c?r=cb0b2a27e3f6293076f584b878ed733a22cd75ff#576)).

I added my quick testing script as an attachment. This is my first PR here so let me know if there are any other requirement.

[test.zip](https://github.com/user-attachments/files/21167583/test.zip)


ondrejmirtes:
Thank you!

PR #4087, opened 2025-07-10
Fix SplFileObject::fwrite return type

VincentLanglet:
Close https://github.com/phpstan/phpstan/issues/13208


ondrejmirtes:
Thank you.

PR #4088, opened 2025-07-11
Fix return type of bzdecompress()

jlherren:
It mostly returns `int` error codes on failures and `false` only when it cannot initialize.

Source: [php-src](https://github.com/php/php-src/blob/master/ext/bz2/bz2.c#L489-L558), test code: `var_dump(bzdecompress("invalid"));`


ondrejmirtes:
Thank you!

PR #4090, opened 2025-07-11
Fix missing detection of dead code in expressions

staabm:
closes https://github.com/phpstan/phpstan/issues/13232
closes https://github.com/phpstan/phpstan/issues/11909

### DIFF
--- a/src/Analyser/ExpressionResult.php
+++ b/src/Analyser/ExpressionResult.php
@@ -24,11 +24,11 @@ final class ExpressionResult
 	public function __construct(
 		private MutatingScope $scope,
 		private bool $hasYield,
+		private bool $isAlwaysTerminating,
 		private array $throwPoints,
 		private array $impurePoints,
 		?callable $truthyScopeCallback = null,
 		?callable $falseyScopeCallback = null,
-		private bool $isAlwaysTerminating = false,
 	)
 	{
 		$this->truthyScopeCallback = $truthyScopeCallback;
--- a/src/Analyser/ExpressionResult.php
+++ b/src/Analyser/ExpressionResult.php
@@ -28,6 +28,7 @@ final class ExpressionResult
 		private array $impurePoints,
 		?callable $truthyScopeCallback = null,
 		?callable $falseyScopeCallback = null,
+		private bool $isAlwaysTerminating = false,
 	)
 	{
 		$this->truthyScopeCallback = $truthyScopeCallback;
@@ -88,6 +89,11 @@ final class ExpressionResult
 		$callback = $this->falseyScopeCallback;
 		$this->falseyScope = $callback();
 		return $this->falseyScope;
+	}
+
+	public function isAlwaysTerminating(): bool
+	{
+		return $this->isAlwaysTerminating;
 	}
 
 }
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2935,7 +2935,7 @@ final class NodeScopeResolver
 				$exprResult->getThrowPoints(),
 				$exprResult->getImpurePoints(),
 				static fn (): MutatingScope => $scope->filterByTruthyValue($expr),
-				static fn (): MutatingScope => $scope->filterByFalseyValue($expr)
+				static fn (): MutatingScope => $scope->filterByFalseyValue($expr),
 			);
 		} elseif ($expr instanceof StaticCall) {
 			$hasYield = false;
@@ -3247,6 +3247,7 @@ final class NodeScopeResolver
 					$hasYield = $hasYield || $keyResult->hasYield();
 					$throwPoints = array_merge($throwPoints, $keyResult->getThrowPoints());
 					$impurePoints = array_merge($impurePoints, $keyResult->getImpurePoints());
+					$isAlwaysTerminating = $isAlwaysTerminating || $keyResult->isAlwaysTerminating();
 					$scope = $keyResult->getScope();
 				}
 
@@ -3254,6 +3255,7 @@ final class NodeScopeResolver
 				$hasYield = $hasYield || $valueResult->hasYield();
 				$throwPoints = array_merge($throwPoints, $valueResult->getThrowPoints());
 				$impurePoints = array_merge($impurePoints, $valueResult->getImpurePoints());
+				$isAlwaysTerminating = $isAlwaysTerminating || $valueResult->isAlwaysTerminating();
 				$scope = $valueResult->getScope();
 			}
 			$nodeCallback(new LiteralArrayNode($expr, $itemNodes), $scope);
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1915,7 +1915,7 @@ final class NodeScopeResolver
 							$nodeCallback($node, $scope);
 						},
 						ExpressionContext::createDeep(),
-						static fn (MutatingScope $scope): ExpressionResult => new ExpressionResult($scope, false, [], []),
+						static fn (MutatingScope $scope): ExpressionResult => new ExpressionResult($scope, false, false, [], []),
 						false,
 					)->getScope();
 				} elseif ($var instanceof PropertyFetch) {
@@ -2475,7 +2475,7 @@ final class NodeScopeResolver
 						$scope = $scope->exitExpressionAssign($expr->expr);
 					}
 
-					return new ExpressionResult($scope, $hasYield, $throwPoints, $impurePoints, isAlwaysTerminating: $isAlwaysTerminating);
+					return new ExpressionResult($scope, $hasYield, $isAlwaysTerminating, $throwPoints, $impurePoints);
 				},
 				true,
 			);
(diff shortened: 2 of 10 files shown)


staabm:
hmm I wonder whether I should add this `$isAlwaysTerminating = $isAlwaysTerminating || $result->isAlwaysTerminating();` handling everywhere, where we have 
```
			$hasYield = $hasYield || $result->hasYield();
			$throwPoints = array_merge($throwPoints, $result->getThrowPoints());
			$impurePoints = array_merge($impurePoints, $result->getImpurePoints());
```

etc...?

edit: I added it in all places where I could think of a example which I was able to test

staabm:
I think the new behaviour makes sense but the errors are unfortunate

https://github.com/laravel/framework/blob/43993ed92af54aa8620d8e779a7dcd658f44364c/types/Support/Helpers.php#L44-L60

```
------ ------------------------------------------------------- 
  Line   Support/Helpers.php                                    
 ------ ------------------------------------------------------- 
  47     Unreachable statement - code above always terminates.  
         🪪  deadCode.unreachable                               
  66     Unreachable statement - code above always terminates.  
         🪪  deadCode.unreachable                               
 ------ ------------------------------------------------------- 
```

ondrejmirtes (changes requested):
There is StatementResultTest that tests "is always terminating". Please add a similar test for ExpressionResult. Thank you.

src/Analyser/ExpressionResult.php
```diff
@@ -28,6 +28,7 @@ public function __construct(
 		private array $impurePoints,
 		?callable $truthyScopeCallback = null,
 		?callable $falseyScopeCallback = null,
+		private bool $isAlwaysTerminating = false,
```
ondrejmirtes: Doesn't have to be optional and doesn't have to be at the end. I'd prefer it to be after `hasYield`, for consistency with StatementResult.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
```diff
@@ -2408,6 +2413,7 @@ public function processExprNode(Node\Stmt $stmt, Expr $expr, MutatingScope $scop
 			return $this->processExprNode($stmt, $newExpr, $scope, $nodeCallback, $context);
 		}
 
+		$isAlwaysTerminating = false;
```
ondrejmirtes: Remove the variable assign from here. It should be set inside each `if` about each node type.

src/Analyser/NodeScopeResolver.php
```diff
@@ -3192,6 +3218,7 @@ static function (): void {
 				$hasYield = $hasYield || $result->hasYield();
 				$throwPoints = array_merge($throwPoints, $result->getThrowPoints());
 				$impurePoints = array_merge($impurePoints, $result->getImpurePoints());
+				$isAlwaysTerminating = $isAlwaysTerminating || $result->isAlwaysTerminating();
```
ondrejmirtes, line 3235: This is not initialized before the foreach

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
```diff
@@ -2527,6 +2537,7 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context): Exp
 			$functionReflection = null;
 			$throwPoints = [];
 			$impurePoints = [];
+			$isAlwaysTerminating = false;
```
ondrejmirtes, line 2542: Also use the value from `$nameResult`.

src/Analyser/NodeScopeResolver.php
```diff
@@ -2922,6 +2942,7 @@ static function (): void {
 			$hasYield = false;
 			$throwPoints = [];
 			$impurePoints = [];
+			$isAlwaysTerminating = false;
```
ondrejmirtes, line 2948: Should be taken from $classResult in case class is Expr

src/Analyser/NodeScopeResolver.php
```diff
@@ -3076,12 +3099,14 @@ static function (): void {
...
 			$throwPoints = array_merge($throwPoints, $result->getThrowPoints());
 			$impurePoints = array_merge($impurePoints, $result->getImpurePoints());
+			$isAlwaysTerminating = $isAlwaysTerminating || $result->isAlwaysTerminating();
 		} elseif ($expr instanceof PropertyFetch) {
 			$scopeBeforeVar = $scope;
 			$result = $this->processExprNode($stmt, $expr->var, $scope, $nodeCallback, $context->enterDeep());
 			$hasYield = $result->hasYield();
 			$throwPoints = $result->getThrowPoints();
 			$impurePoints = $result->getImpurePoints();
+			$isAlwaysTerminating = false;
```
ondrejmirtes: should be taken from `$result` and also updated in `$expr->name instanceof Expr` branch

src/Analyser/NodeScopeResolver.php
```diff
@@ -3193,12 +3225,14 @@ static function (): void {
...
 				$throwPoints = array_merge($throwPoints, $result->getThrowPoints());
 				$impurePoints = array_merge($impurePoints, $result->getImpurePoints());
+				$isAlwaysTerminating = $isAlwaysTerminating || $result->isAlwaysTerminating();
 				$scope = $result->getScope();
 			}
 		} elseif ($expr instanceof ArrayDimFetch) {
 			$hasYield = false;
 			$throwPoints = [];
 			$impurePoints = [];
+			$isAlwaysTerminating = false;
```
ondrejmirtes, line 3242: Basically this should be updated every time `$hasYield` is updated. Please review all branches in `processExprNode` like that.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: I adjusted it in most cases.

I left out e.g. null-safe operations, as these would only terminate for always-non-nullables and similar stuff.

staabm:
I think this should be ready to land.

thanks for the feedback

tests/PHPStan/Analyser/ExpressionResultTest.php
```diff
@@ -0,0 +1,149 @@
...
+				'empty(exit());',
+				true,
+			],
+			[
+				'isset(exit());',
+				true,
+			],
+			[
+				'$x ? "abc" : exit();',
+				true,
```
ondrejmirtes: Why is this always terminating?

tests/PHPStan/Analyser/ExpressionResultTest.php
```diff
@@ -0,0 +1,149 @@
...
+			],
+			[
+				'$x ? exit() : "abc";',
```
ondrejmirtes, line 75: Why is this always terminating?

tests/PHPStan/Analyser/ExpressionResultTest.php
```diff
@@ -0,0 +1,149 @@
...
+				true,
+			],
+			[
+				'fn() => yield (exit());',
```
ondrejmirtes, line 79: Why is this always terminating? Only execution of such closure would be always terminating, but not closure itself.

tests/PHPStan/Analyser/ExpressionResultTest.php
```diff
@@ -0,0 +1,149 @@
...
+				true,
+			],
+			[
+				'@exit();',
+				true,
+			],
+			[
+				'$x && exit();',
```
ondrejmirtes, line 87: The 2nd part might never be executed

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/ExpressionResultTest.php
staabm: great feedback, thank you.

src/Analyser/NodeScopeResolver.php
```diff
@@ -3289,7 +3289,7 @@ static function (): void {
 			return new ExpressionResult(
 				$leftMergedWithRightScope,
 				$leftResult->hasYield() || $rightResult->hasYield(),
-				false,
+				$leftResult->isAlwaysTerminating() || $rightResult->isAlwaysTerminating(),
```
ondrejmirtes: The right side might never be executed so this should only mention the left side.

ondrejmirtes: `$x && exit();` should be false

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: it is `false`, see https://github.com/phpstan/phpstan-src/pull/4090/files#diff-fa85cdb8d6b73f343f042bbbe433ecc95119bf3dc5e76cc7291563d1c49cbd0dR87

I guess you was looking at an outdated version of the PR while commenting

ondrejmirtes:
I swear GitHub has some kind of caching problem with the new "Files changed" UI. Sorry.

ondrejmirtes:
Thank you!

PR #4091, opened 2025-07-11
More precise return type for `posix_getpwnam()`

zonuexe:
Fixes https://github.com/phpstan/phpstan/issues/12254


ondrejmirtes:
Thank you!

PR #4092, opened 2025-07-12
stream_get_contents() does not return FALSE unless an invalid position is provided

staabm:
closes https://github.com/phpstan/phpstan/issues/3396


ondrejmirtes:
Thank you!

PR #4094, opened 2025-07-13
Add stringable access check to AccessPropertiesCheck

zonuexe:
refs https://github.com/phpstan/phpstan/issues/13238

In the context of property fetching, string keys are implicitly cast.

 * https://3v4l.org/0noRl

Property names starting with a number cannot be defined, but there is no need to exclude numeric, as they are accessible via the `__get()` magic method.

-----

For work reasons, this branch is based on https://github.com/phpstan/phpstan-src/pull/3910, so it will be rebased once that is merged.


zonuexe:
(I'll get home in a few hours and fix it.)

ondrejmirtes:
Thank you!

PR #4095, opened 2025-07-13
Add phpstan-sealed support

VincentLanglet:


src/Rules/Classes/SealedRule.php
```diff
@@ -0,0 +1,69 @@
...
+			$sealedTags = $parent->getSealedTags();
+			foreach ($sealedTags as $sealedTag) {
+				$type = $sealedTag->getType();
+				if ($type->isSuperTypeOf(new ObjectType($className))->yes()) {
+					continue;
+				}
+
+				$errors[] = RuleErrorBuilder::message(
+					sprintf(
+						'%s %s is sealed and only permits %s as subtypes, %s given.',
```
ondrejmirtes: Instead of this rule, you should implement this extension https://apiref.phpstan.org/2.1.x/PHPStan.Reflection.AllowedSubTypesClassReflectionExtension.html. Which actually make this `@phpstan-sealed` tag useful for analysis, and also already checks what you're checking here, through AllowedSubTypesRule.

ondrejmirtes (changes requested):

ondrejmirtes:
Some issues are related to sealed classes and after this is implemented, it'd be easier to reproduce them and in the end, fix them:

* https://github.com/phpstan/phpstan/issues/12882
* https://github.com/phpstan/phpstan/issues/12241
* https://github.com/phpstan/phpstan/issues/10148

VincentLanglet:
The failure seems unrelated to the PR but to https://github.com/phpstan/phpstan-src/commit/fe4bf2c91e5b212a3d302af30d74dae3531efba3

src/PhpDoc/PhpDocNodeResolver.php
```diff
@@ -524,6 +525,24 @@ public function resolveRequireImplementsTags(PhpDocNode $phpDocNode, NameScope $
...
 	}
 
+	/**
+	 * @return array<SealedTypeTag>
+	 */
+	public function resolveSealedTags(PhpDocNode $phpDocNode, NameScope $nameScope): array
+	{
+		$resolved = [];
+
+		foreach (['@psalm-inheritors', '@phpstan-sealed'] as $tagName) {
```
mvorisek, line 535: I would go with `phpstan-inheritors` as psalm did. The reason is "sealed" term is too broad and can be easily confused with sealed array.

Also I think "inheritors" is a better name as we specify what direct child classes can implement it, but we do not restrict further extending, ie. we do not seal it per se.

VincentLanglet: This has already be decided on the phpdoc-parser library pr by ondrej

VincentLanglet: See https://github.com/phpstan/phpdoc-parser/pull/273

tests/PHPStan/Rules/Classes/data/sealed.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Sealed;
+
+/**
+ * @phpstan-sealed FooClass|BarClass
+ */
+class BaseClass {}
+class FooClass extends BaseClass {}
+class BarClass extends BaseClass {}
```
mvorisek, line 10:
```suggestion
class BarClass extends BaseClass {}
class BarClassChild extends BarClass {}
```

Should be tested for all cases and pass.

VincentLanglet: Added

src/Rules/Classes/SealedRule.php
ondrejmirtes: With SealedAllowedSubTypesClassReflectionExtension implemented, this rule can be deleted.

ondrejmirtes (changes requested):

src/Rules/Classes/SealedRule.php
VincentLanglet: My bad, dunno why I forget to remove it ; it's done.

src/Rules/PhpDoc/SealedDefinitionClassRule.php
```diff
@@ -0,0 +1,100 @@
...
+use function array_map;
+use function array_merge;
+use function count;
+use function sprintf;
+
+/**
+ * @implements Rule<InClassNode>
+ */
+#[RegisteredRule(level: 0)]
+final class SealedDefinitionClassRule implements Rule
```
ondrejmirtes, line 26: Should be level 2. That's where PHPDocs are being checked.

ondrejmirtes (changes requested):

src/Rules/PhpDoc/SealedDefinitionClassRule.php
```diff
@@ -0,0 +1,100 @@
...
+use function array_column;
+use function array_map;
+use function array_merge;
+use function count;
+use function sprintf;
+
+/**
+ * @implements Rule<InClassNode>
+ */
+#[RegisteredRule(level: 0)]
```
ondrejmirtes:
```suggestion
#[RegisteredRule(level: 2)]
```

ondrejmirtes: This is what I want to do

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

ondrejmirtes:
Please submit docs change for "PHPDocs basics" page.

VincentLanglet:
> Please submit docs change for "PHPDocs basics" page.

What should we do about this page https://phpstan.org/developing-extensions/allowed-subtypes now it could be simply done with `@phpstan-sealed` ?

ondrejmirtes:
We can put a blue box there explaining there's an easier way, but the extension is still relevant if people want to do it with their custom attributes or other logic.

ondrejmirtes:
Please send the documentation PR. Thanks.

VincentLanglet:
> Please send the documentation PR. Thanks.

I didn't forget, it's on my TODO but I have multiple contributions on multiple repository to maintain.
(And english is not my best language...)

I assume you wait for it before a new release ?

ondrejmirtes:
I'm not blocking release on that but would be nice to have it in the near future :)

Anyway, you've contributed a lot to the upcoming releases, thank you so much!

VincentLanglet:
> Anyway, you've contributed a lot to the upcoming releases, thank you so much!

Thanks for your review too! 
It feels good to get all those PRs out of the way. I'll try to finish the existing one, and find new one to open :)

PR #4096, opened 2025-07-13
Fix SimpleXMLElement::xpath return type

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13236

The initial purpose of the dynamic return type was to remove the false return type.


ondrejmirtes:
Thank you.

PR #4098, opened 2025-07-14
Simplify by replacing chained `$var === const` with `in_array()`

zonuexe:
This is the same change as the previous one at https://github.com/phpstan/phpstan-src/pull/2618, but it will be automatically refactored by rules 🤖.

Currently, `return $var === 'foo' || $var === 'bar';` and `$isFooOrBar = $var === 'foo' || $var === 'bar';` are not covered. It also does not support formulas replaced by [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

### DIFF
--- a/build/PHPStan/Build/OrChainIdenticalComparisonToInArrayRule.php
+++ b/build/PHPStan/Build/OrChainIdenticalComparisonToInArrayRule.php
@@ -1,0 +1,163 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Build;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\ArrayItem;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\BinaryOp\BooleanOr;
+use PhpParser\Node\Expr\BinaryOp\Identical;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\ConstFetch;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Name;
+use PhpParser\Node\Scalar;
+use PhpParser\Node\Stmt\If_;
+use PHPStan\Analyser\Scope;
+use PHPStan\DependencyInjection\AutowiredParameter;
+use PHPStan\File\FileHelper;
+use PHPStan\Node\Printer\ExprPrinter;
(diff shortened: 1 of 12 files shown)


build/PHPStan/Build/OrChainIdenticalComparisonToInArrayRule.php
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * @implements Rule<If_>
+ */
+final class OrChainIdenticalComparisonToInArrayRule implements Rule
+{
+
+	public function __construct(
+		#[AutowiredParameter]
+		private PhpPrinter $printer,
```
ondrejmirtes: Do not make this a DI service. Ask for ExprPrinter instead which is already a DI service.

ondrejmirtes (changes requested):
Great idea! I'll merge it after this fix.

build/PHPStan/Build/OrChainIdenticalComparisonToInArrayRule.php
zonuexe: Thanks, I replaced it with `ExprPrinter`.

zonuexe:
I see, the error that occurred after rebasing seems to be independent of the changes in this PR.

```
Error: Catching internal class PHPUnit\Framework\ExpectationFailedException.
Error: Catching internal class PHPUnit\Framework\ExpectationFailedException.
Error: Call to method getComparisonFailure() of internal class PHPUnit\Framework\ExpectationFailedException from outside its root namespace PHPUnit.
Error: Call to method getComparisonFailure() of internal class PHPUnit\Framework\ExpectationFailedException from outside its root namespace PHPUnit.
```

build/PHPStan/Build/OrChainIdenticalComparisonToInArrayRule.php
```diff
@@ -0,0 +1,163 @@
...
+use function str_starts_with;
+
+/**
+ * @implements Rule<If_>
+ */
+final class OrChainIdenticalComparisonToInArrayRule implements Rule
+{
+
+	public function __construct(
+		#[AutowiredParameter]
```
ondrejmirtes:
```suggestion
```

ondrejmirtes:
Thank you!

An idea: Generalize this rule to also account for `if ($a !== 'Foo' && $a !== 'Bar')` into `!in_array(....`.

PR #4099, opened 2025-07-14
feat: add api to get ObjectType from ClassReflection and EnumCaseReflection

calebdw:
Hello!

This PR adds the following two methods to the api:
- `ClassReflection::getObjectType(): ObjectType`
- `EnumCaseReflection:getEnumCaseObjectType(): EnumCaseObjectType`

### Motivation

It's very easy to get all the `ClassReflection`s from a `Type` with `$type->getObjectClassReflections()`. However, there have been several occasions where I've needed to go the other direction (get an `ObjectType` from a `ClassReflection`) and have not been able to do so easily.

I was digging around in the code, and I discovered a private static method  `ObjectType::createFromReflection()` that already did just this. So I opted to *move* the logic in `ObjectType::createFromReflection()` to `ClassReflection::getObjectType()` instead of simply making the private method public as I didn't think allowing child classes to statically call this method would be a good idea (e.g., `TemplateObjectType::createFromReflection()`)

Thanks!

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -1422,6 +1422,19 @@ final class ClassReflection
 		return $flags;
 	}
 
+	public function getObjectType(): ObjectType
+	{
+		if (!$this->isGeneric()) {
+			return new ObjectType($this->getName());
+		}
+
+		return new GenericObjectType(
+			$this->getName(),
+			$this->typeMapToList($this->getActiveTemplateTypeMap()),
+			variances: $this->varianceMapToList($this->getCallSiteVarianceMap()),
+		);
+	}
+
 	public function getTemplateTypeMap(): TemplateTypeMap
 	{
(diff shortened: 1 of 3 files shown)


calebdw:
@ondrejmirtes,

In the original logic, you have the class reflection instance, but you don't pass it into the ObjectType and instead opt to leave it null so that the class reflection is assigned the next time `getClassReflection()` is called on the type. 

What is the reasoning for this? I would think that it would be better to pass along the class reflection instance since you already have it, but maybe I'm missing something...


https://github.com/phpstan/phpstan-src/blob/5ab9acce23fd4a42b1e033cc99c83e12f1b551ed/src/Type/ObjectType.php#L136-L147

ondrejmirtes:
Thank you, this is okay. Omitting of passing class reflection to ObjectType in the original code might change some behaviour for the better or worse as there will be slight differences in some scenario, we'll see.

PR #4103, opened 2025-07-17
Added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/13210

was fixed by https://github.com/phpstan/phpstan-src/pull/3998


ondrejmirtes:
Thank you!

PR #4104, opened 2025-07-17
Update brianium/paratest to fix parallel tests build

staabm:


staabm:
this PR fixes most CI errors we currently see in all pull requests.

I am not sure about the few remaining ones, therefore leaving them in error state for now.

ondrejmirtes:
Thank you very much!

PR #4106, opened 2025-07-17
Fix IntersectionType::isIterableOnce

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13248

I feel like non-iterable should be excluded since as soon as one subtype is iterable the whole intersection is iterable.

### DIFF
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -1266,10 +1266,9 @@ class IntersectionType implements CompoundType
 	}
 
 	/**
-	 * @param callable(Type $type): Type        $getType
-	 * @param (callable(Type $type): Type)|null $filter
+	 * @param callable(Type $type): Type $getType
 	 */
-	private function intersectTypes(callable $getType, ?callable $filter = null): Type
+	private function intersectTypes(callable $getType): Type
 	{
 		$operands = array_map($getType, $this->types);
 		return TypeCombinator::intersect(...$operands);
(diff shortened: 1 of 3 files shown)


src/Type/IntersectionType.php
```diff
@@ -1243,16 +1247,29 @@ public function getFiniteTypes(): array
...
+		return TrinaryLogic::lazyMaxMin($types, $getResult);
 	}
 
 	/**
-	 * @param callable(Type $type): Type $getType
+	 * @param callable(Type $type): Type        $getType
+	 * @param (callable(Type $type): Type)|null $filter
 	 */
-	private function intersectTypes(callable $getType): Type
+	private function intersectTypes(callable $getType, ?callable $filter = null): Type
```
ondrejmirtes: You've added a filter here but it's not used. Please remove this signature change.

ondrejmirtes (changes requested):
This is great, thank you.

src/Type/IntersectionType.php
VincentLanglet: My bad, fixed

ondrejmirtes:
Also, there might be more candidates to also use a filter in IntersectionType: hasOffsetValueType/getOffsetValueTypesetOffsetValueType (should use isOffsetAccessible)

Feel free to explore that with a failing test in a separate PR. Thank you!

ondrejmirtes:
But maybe not because ObjectType::isOffsetAccessible() says "maybe" for non-final classes which is correct.

PR #4109, opened 2025-07-18
Add non regression test for 13043

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13043

This was resolved when improving the array_map return type.

Also, I tried doing some triage and I think some issues which might be closed:
- https://github.com/phpstan/phpstan/issues/11680 is closed by https://github.com/phpstan/phpstan-src/pull/3326
- https://github.com/phpstan/phpstan/issues/13070 is an PEST issue not a PHPStan one
- https://github.com/phpstan/phpstan/issues/13078 is a valid error related to https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static that someone wants to ignore
- https://github.com/phpstan/phpstan/issues/13080 is a valid error to me (as explained to the author)
- https://github.com/phpstan/phpstan/issues/13134 which is valid (and the user got an answer why)
- https://github.com/phpstan/phpstan/issues/13207 which was explained

Dunno if it helps you (?)

### DIFF
--- a/tests/PHPStan/Rules/Methods/ReturnTypeRuleTest.php
+++ b/tests/PHPStan/Rules/Methods/ReturnTypeRuleTest.php
@@ -1239,6 +1239,11 @@ class ReturnTypeRuleTest extends RuleTestCase
 		]);
 	}
 
+	public function testBug13043(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-13043.php'], []);
+	}
+
 	public function testBug12928(): void
 	{
 		$this->analyse([__DIR__ . '/data/bug-12928.php'], [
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
I'm thankful for the triage you did but please do not mention it in a non-related PR, instead post a comment in each issue.

ondrejmirtes:
Thank you!

VincentLanglet:
> I'm thankful for the triage you did but please do not mention it in a non-related PR, instead post a comment in each issue.

Ok.

Do I need to ping you in each issue ? Or just commenting is enough for you to get a notification ?

Also I assume I don't have to ping you for the ones already listed here (?)

ondrejmirtes:
I see all comments, even if they don't mention me. Yes, please say in each one of them that they can be closed, because when I close them the author will know why.

PR #4110, opened 2025-07-18
Add non regression tests for strict rules issues

### LINKED ISSUE #141: Wrong symbol shown for error report on `and`/`or` non-boolean parameters
https://phpstan.org/r/e3b1da08-3844-4f2a-ac58-a6514af25849

`and` is used, but the error message shows `&&`. Similar problem for `or` and `||`.

### LINKED ISSUE #143: Array key always exists false positive with composite keys
https://phpstan.org/r/50a1171b-7243-4e7d-a9d4-9d0b21cc1d02

"Call to function array_key_exists() with non-empty-string and array will always evaluate to true."

I disabled the following error message using checkMissingIterableValueType: false:
"Function foo() has parameter $arr with no value type specified in iterable type array."
Adding a typehint for for the array makes the other error go away, but that's not the point here.

### LINKED ISSUE #147: Call to function array_is_list() with will always evaluate to true.
Hi!

As you can se there is a false positive issue in the link:

https://phpstan.org/r/4a61f25b-34eb-4935-a054-868192f50aea

ps: if you know a better way to check if an array is already sorted please let me know. :)

VincentLanglet:
Closes https://github.com/phpstan/phpstan-strict-rules/issues/147
Closes https://github.com/phpstan/phpstan-strict-rules/issues/143

Also, you can manually close https://github.com/phpstan/phpstan-strict-rules/issues/141,
because you already solve the issue in https://github.com/phpstan/phpstan-strict-rules/commit/88e714c05658eb18a45ab6a6f4d17a29d44cc065

### DIFF
--- a/tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
+++ b/tests/PHPStan/Rules/Comparison/ImpossibleCheckTypeFunctionCallRuleTest.php
@@ -1007,6 +1007,18 @@ class ImpossibleCheckTypeFunctionCallRuleTest extends RuleTestCase
 		]);
 	}
 
+	public function testBugStrictRule147(): void
+	{
+		$this->treatPhpDocTypesAsCertain = true;
+		$this->analyse([__DIR__ . '/data/bug-strict-147.php'], []);
+	}
+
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you!

PR #4111, opened 2025-07-18
Introduce `rawMessage` key in `ignoreErrors`

gharlan:
refs https://github.com/phpstan/phpstan/issues/11178#issuecomment-2167343722

This pull request introduces the `rawMessage` key in `ignoreErrors` for non-regex raw error messages:

```neon
parameters:
	ignoreErrors:
		-
			rawMessage: 'Raw message without regex escaping etc.'
			identifier: ...
```

### DIFF
--- a/conf/parametersSchema.neon
+++ b/conf/parametersSchema.neon
@@ -114,6 +114,7 @@ parametersSchema:
 			structure([
 				?message: string()
 				?messages: listOf(string())
+				?rawMessage: string()
 				?identifier: string()
 				?identifiers: listOf(string())
 				?path: string()
--- a/src/Analyser/Ignore/IgnoredError.php
+++ b/src/Analyser/Ignore/IgnoredError.php
@@ -30,6 +30,8 @@ final class IgnoredError
 		$message = '';
 		if (isset($ignoredError['message'])) {
 			$message = $ignoredError['message'];
+		} elseif (isset($ignoredError['rawMessage'])) {
+			$message = '"' . $ignoredError['rawMessage'] . '"';
 		}
 		if (isset($ignoredError['identifier'])) {
 			if ($message === '') {
@@ -71,6 +73,7 @@ final class IgnoredError
 		FileHelper $fileHelper,
 		Error $error,
 		?string $ignoredErrorPattern,
+		?string $ignoredErrorMessage,
 		?string $identifier,
 		?string $path,
 	): bool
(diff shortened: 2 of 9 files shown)


conf/parametersSchema.neon
```diff
@@ -143,6 +143,25 @@ parametersSchema:
...
+				path: string()
+				?identifier: string()
+				?reportUnmatched: bool()
+			]),
+			structure([
+				rawMessage: string()
+				paths: listOf(string())
+				?identifier: string()
+				?reportUnmatched: bool()
+			]),
```
ondrejmirtes: This is quickly getting out of hand. You're missing some combinations, like with `identifiers` key.

Maybe we could validate these entries manually in https://github.com/phpstan/phpstan-src/blob/74ac94c93e580bdc5b39b6c8a704508fab927251/src/DependencyInjection/ContainerFactory.php#L303 instead of trying to have a huge declarative schema describing all combinations.

ondrejmirtes (changes requested):

conf/parametersSchema.neon
gharlan: > This is quickly getting out of hand. You're missing some combinations, like with identifiers key.

`identifiers` didn't exist when I created the pull request. Added them now.

> Maybe we could validate these entries manually in [...]

Should this be done here in the pr? Also reducing the existing  combinations?

ondrejmirtes: You can try sending a separate PR that would move the validation. And once it's merged, rebase this PR and adjust the new logic 😊 Thank you.

gharlan: > Maybe we could validate these entries manually in
> 
> https://github.com/phpstan/phpstan-src/blob/74ac94c93e580bdc5b39b6c8a704508fab927251/src/DependencyInjection/ContainerFactory.php#L303

Or maybe adding them to `IgnoredErrorHelper` here?

https://github.com/phpstan/phpstan-src/blob/2.1.x/src/Analyser/Ignore/IgnoredErrorHelper.php#L43-L49

ondrejmirtes: I'd rather them to be validated in ContainerFactory.

gharlan:
In a follow-up PR, I would like to change the baseline generation to use `rawMessage` instead of `message`.
Can I just change that? Or should it be configurable? Should it be part of bleedingEdge?

ondrejmirtes:
Thank you!

Yes, the change should be part of bleeding edge. Since both BaselinePhpErrorFormatter and BaselineNeonErrorFormatter are instantiated in AnalyseCommand, you should add a new featureToggles parameter and fetch it there through `$container->getParameter()`.

szepeviktor:
This is a great improvement.

PR #1497, opened 2022-07-06
Fix #7531 - "Unreachable statement" false positive for html whitespace

murilohpucci:
Hey, first time here, I hope this can help you.

Fixing: [#7531](https://github.com/phpstan/phpstan/issues/7531)

PR #4112, opened 2025-07-18
Fix false positive unreachable statement

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/7531

Closes https://github.com/phpstan/phpstan-src/pull/1497


ondrejmirtes:
Thank you!

PR #4113, opened 2025-07-18
Fix PDO quote return type

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10512

Closes https://github.com/phpstan/phpstan-src/pull/2926

Use a benevolent union as recommended here: https://github.com/phpstan/phpstan-src/pull/2926#issuecomment-2282567015.


ondrejmirtes:
Thank you!

PR #4114, opened 2025-07-18
Fix "Unsafe use of new static when trait enforces constructor via abstract method"

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9654

Closes https://github.com/phpstan/phpstan-src/pull/2767

### DIFF
--- a/src/Rules/Classes/NewStaticRule.php
+++ b/src/Rules/Classes/NewStaticRule.php
@@ -9,6 +9,7 @@ use PHPStan\Php\PhpVersion;
 use PHPStan\Reflection\Php\PhpMethodReflection;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\TrinaryLogic;
 use function strtolower;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Rules/Classes/NewStaticRule.php
```diff
@@ -77,6 +84,20 @@ public function processNode(Node $node, Scope $scope): array
...
 
+		if (
+			$this->phpVersion->supportsAbstractTraitMethods()
+			&& $scope->isInTrait()
+		) {
+			$traitReflection = $scope->getTraitReflection();
+			if ($traitReflection->hasConstructor()) {
+				$traitConstructor = $traitReflection->getConstructor();
+
+				if ($traitConstructor instanceof PhpMethodReflection && $traitConstructor->isAbstract()) {
```
ondrejmirtes: Why it has to be instanceof `PhpMethodReflection`? Such code is usually very brittle.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4115, opened 2025-07-19
feat: add optional rule to report properties that should be promoted

calebdw:
Hello!

Not sure if this would be better suited for [phpstan-strict-rules](https://github.com/phpstan/phpstan-strict-rules), but I figured I'd try it here first.

This adds an optional rule that reports properties that can easily be promoted, which have the following form:

```php
class Foo
{
    public int $count;
    public function __construct(int $count)
    {
        $this->count = $count;
    }
}

// should be

class Foo
{
    public function __construct(public int $count)
    {
    }
}
```

Thanks!

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -69,6 +69,7 @@ parameters:
 	reportPossiblyNonexistentGeneralArrayOffset: false
 	reportPossiblyNonexistentConstantArrayOffset: false
 	checkMissingOverrideMethodAttribute: false
+	reportPropertiesThatShouldBePromoted: false
 	mixinExcludeClasses: []
 	scanFiles: []
 	scanDirectories: []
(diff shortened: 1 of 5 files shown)


VincentLanglet:
> Not sure if this would be better suited for [phpstan-strict-rules](https://github.com/phpstan/phpstan-strict-rules?rgh-link-date=2025-07-19T01%3A55%3A00Z), but I figured I'd try it here first.

This seems to me really opinionated indeed. Also
- It does not really provide any benefit on the code quality or analysis
- You will report unresolvable errors on PHP < 8.0

Such behavior is better on a automated fixable tool, like php-cs-fixer or rector IMHO.
And rector already have such rule https://getrector.com/rule-detail/class-property-assign-to-constructor-promotion-rector

calebdw:
Ah good to know,  I've been meaning to take a look at rector but just haven't done it yet. Thanks!

ondrejmirtes:
Personally I'm using "slevomat/coding-standard" for this.

PR #4117, opened 2025-07-19
Fix AccessPropertiesCheck for checkThisOnly

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13271

### DIFF
--- a/src/Rules/Properties/AccessPropertiesCheck.php
+++ b/src/Rules/Properties/AccessPropertiesCheck.php
@@ -42,6 +42,8 @@ final class AccessPropertiesCheck
 		private bool $checkDynamicProperties,
 		#[AutowiredParameter(ref: '%featureToggles.checkNonStringableDynamicAccess%')]
 		private bool $checkNonStringableDynamicAccess,
+		#[AutowiredParameter]
+		private bool $checkThisOnly,
 	)
 	{
 	}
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Why is this error happening only on level 1 right now? I don't get it.

ondrejmirtes:
Also I'm worried that if you just silence this situation for checkThisOnly=false, it will still happen in: https://phpstan.org/r/c822e0d3-9d45-4531-aced-23ce661afdcb

ondrejmirtes:
I just don't get why is this happening, please explain how you debugged this :)

VincentLanglet:
> Also I'm worried that if you just silence this situation for checkThisOnly=false, it will still happen in: [phpstan.org/r/c822e0d3-9d45-4531-aced-23ce661afdcb](https://phpstan.org/r/c822e0d3-9d45-4531-aced-23ce661afdcb)

The "special check" for `thisOnly` already exists for some other property-related rules
https://github.com/phpstan/phpstan-src/blob/cb331f5698ab83eeda672f151f772917ec901380/src/Rules/Properties/ReadingWriteOnlyPropertiesRule.php#L45-L51

The fact is the line
```
$nameTypeResult = $this->ruleLevelHelper->findTypeToCheck(
			$scope,
			$nodeName,
			'',
			static fn (Type $type) => $type->toString()->isString()->yes(),
		);
$nameType = $nameTypeResult->getType();
```
- Return an ErrorType on level 1
- Return the correctType on level > 1

So then the early return
```
		if (
			!$nameType instanceof ErrorType
			&& !$nameType->toString() instanceof ErrorType
			&& $nameType->toString()->isString()->yes()
		) {
			return [];
		}
```
only works for level > 1.

And why do we get `ErrorType` on level 1 ?
Because of the early return
```
if ($this->checkThisOnly && !$this->isThis($var)) {
			return new FoundTypeResult(new ErrorType(), [], [], null);
		}
```
in
https://github.com/phpstan/phpstan-src/blob/cb331f5698ab83eeda672f151f772917ec901380/src/Rules/RuleLevelHelper.php#L168-L181

I tried first to change the early return in the rule to
```
if (
			$nameType instanceof ErrorType
			|| (!$nameType->toString() instanceof ErrorType && $nameType->toString()->isString()->yes())
		) {
			return [];
		}
```
but some tests are failing which missing error like
```
Property name for $this(DynamicStringableNullsafeAccess\Foo) must be a string, but object was given
```

ondrejmirtes:
I noticed that the conditions in AccessPropertiesCheck are wrong.

This is the concerning line: https://github.com/phpstan/phpstan-src/blob/74e843267a80f2a06a13d53331b428f0d5cf7124/src/Rules/Properties/AccessPropertiesCheck.php#L68

The error CANNOT be reported when `$nameType instanceof ErrorType`. There was a missing `!`.

Additionally, I synced up the rest of the condition with the callback passed to `findTypeToCheck`.

So now, the callback says:

```php
static fn (Type $type) => !$type->toString() instanceof ErrorType && $type->toString()->isString()->yes()
```

And the whole condition after is: 

```php
				if (
					!$nameType instanceof ErrorType
					&& (
						$nameType->toString() instanceof ErrorType
						|| !$nameType->toString()->isString()->yes()
					)
				) {
```

See the commit: https://github.com/phpstan/phpstan-src/commit/81a8d3495d76f8f8f075a8917f2d16cc1dc2ce05

This is how all the other occurences of findTypeToCheck work. So the issue had nothing to do with checkThisOnly.

The fact that errors about `object` disappeared is correct, they're supposed to be reported only on level 7+ and that's not what the test checks.

PR #4119, opened 2025-07-19
Fix ocdb function signatures

VincentLanglet:
- Every function returning a ressource can return false
Currently there is an inconsistent behavior https://phpstan.org/r/6ef0920a-7e33-4c93-a08c-82a76e6a7557
(In php 8.4 it returns ressource|false but in lower php version it returns false.)

- `odbc_do` was forgotten in the functionMap_delta since it's an alias of `odbc_exec`


ondrejmirtes (changes requested):
Please do not update the parameter names. PHPStan already knows the correct ones thanks to https://github.com/phpstan/php-8-stubs. The only thing your change would do would be breaking baselines for people on PHP 7.x.

VincentLanglet:
> Please do not update the parameter names. PHPStan already knows the correct ones thanks to [phpstan/php-8-stubs](https://github.com/phpstan/php-8-stubs?rgh-link-date=2025-07-21T12%3A36%3A40Z). The only thing your change would do would be breaking baselines for people on PHP 7.x.

I simplified the PR

ondrejmirtes:
Thank you.

PR #4120, opened 2025-07-19
Fix pg_lo_import definition

VincentLanglet:
Unlike the doc is saying 
https://www.php.net/manual/en/function.pg-lo-import.php

pg_lo_import seems to be able to return `int|string|false` 
and this is already what PHPStan understand on PHP 8.1+
https://phpstan.org/r/f0e1cd27-4475-4d74-a8e7-555cf014f25b

For reference:
- php/doc PR https://github.com/php/doc-en/pull/4785
- php/src definition
  - https://github.com/php/php-src/blame/dd3a098a9bf967831e889d2e19e873d09c71c9b9/ext/pgsql/pgsql_arginfo.h#L265
  - https://github.com/php/php-src/blob/1748b8111e97f14f709553c7a5498fcdfcd6c754/ext/pgsql/pgsql.stub.php#L348


ondrejmirtes:
Thank you.

PR #4121, opened 2025-07-19
Add missing ssh2_* signatures

VincentLanglet:
Cf https://www.php.net/manual/en/function.ssh2-forward-accept.php
and https://www.php.net/manual/en/function.ssh2-forward-listen.php
and https://www.php.net/manual/en/function.ssh2-poll.php
and https://www.php.net/manual/en/function.ssh2-send-eof.php


ondrejmirtes:
I don't think we need this? https://phpstan.org/r/16bc844a-19ef-458e-a77a-ef78aed7cf0f

VincentLanglet:
> I don't think we need this? [phpstan.org/r/16bc844a-19ef-458e-a77a-ef78aed7cf0f](https://phpstan.org/r/16bc844a-19ef-458e-a77a-ef78aed7cf0f)

The definition seems really wrong:
https://phpstan.org/r/90b89c62-07b7-4a91-9d7f-609c59bf50ca

- `Function ssh2_forward_accept invoked with 1 parameter, 0 required.` => ssh2_forward_accept require 1 param
- `Function ssh2_forward_listen invoked with 2 parameters, 0 required.` => 2 param should be required 
- `ssh2_forward_accept`, `ssh2_forward_listen` and `ssh2_poll` are returning mixed for phpstan when it should be resource|false
- `ssh2_poll` are returning mixed for phpstan when it should be int

I opened the PR on phpstorm to fix this https://github.com/JetBrains/phpstorm-stubs/pull/1758

But I'm not sure to understand how phpstan works. I thought
- It used phpstorm-stubs on PHP 8+
- It use the functionMap on PHP 7.4-

So I opened the PR to fix this on PHP 7.4- ; but it seems like the phpstorm-stubs are also already used on Php 7.4- (?)

ondrejmirtes:
phpstorm-stubs is always used.

php-8-stubs is only used on PHP 8+ (hence the name).

When an error message says "Function ssh2_forward_accept invoked with 1 parameter, 0 required.", it means PHPStan knows the function, but it has wrong signature. Therefore diff ADDING an entry to functionMap.php is suspicious.

I'd prefer waiting for phpstorm-stubs fix to be merged.

VincentLanglet:
Solved by PHPStorm stubs

PR #4122, opened 2025-07-19
Fix lowercase string usage in template

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12739

Seems like the issue is related to the fact that
`ScalarType<string>` and `ScalarType<lowercase-string>` had the same cacheKey or something like that.

`AccessoryLowercaseStringType` and `AccessoryUppercaseStringType` was lost for `level::cache` because the check was only `isPrecise`.

So I added `$level->isCache()` and updated every `$level->isPrecise()` checks (with tests).
But I dunno if it's not better to do something like
```
public function isPrecise(): bool
{
     return $this->value === self::PRECISE || $this->value === self::CACHE;
}
```
(And eventually rename the method to `isAtLeastPrecise` ? Your call @ondrejmirtes.

### DIFF
--- a/tests/PHPStan/Rules/Methods/data/bug-12739.php
+++ b/tests/PHPStan/Rules/Methods/data/bug-12739.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types = 1);
+
+/**
+ * @template TScalar of scalar
+ */
+abstract class ScalarType {
+	/**
+	 * @return class-string<ScalarType<TScalar>>
+	 */
+	abstract public static function getClassIdentify() : string;
+}
+
+/**
+ * @extends ScalarType<int>
+ */
+class MyInt extends ScalarType {
+	public static function getClassIdentify() : string {
+		return MyInt::class;
+	}
+}
+
+/**
+ * @extends ScalarType<string>
+ */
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Rules/Methods/data/bug-12739.php
```diff
@@ -0,0 +1,49 @@
...
+ * @extends ScalarType<string>
+ */
+class MyString extends ScalarType {
+	public static function getClassIdentify() : string {
+		return MyString::class;
+	}
+}
+
+/**
+ * @extends ScalarType<lowercase-string>
```
staabm, line 34: Does the fix also affect other string types, which should be tested here?

VincentLanglet: Yes, uppercaseString.
I added a test.

ondrejmirtes:
There's a conflict, please fix it.

Thank you for the fix. This API is not used often enough for me to be able to come up with a better design, most places use `handle` function. So this PR is fine, just rebase it 😊 Thank you.

staabm (approved):

ondrejmirtes:
Thank you!

PR #4123, opened 2025-07-19
Improve array filter return type on php8

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12689


ondrejmirtes:
Thank you.

PR #4124, opened 2025-07-19
Update eio function return type

VincentLanglet:
All the eio_ function return false on failure.

The description say `resource` as https://www.php.net/manual/en/function.eio-busy.php#refsect1-function.eio-busy-description
but everytime it's say it can return false on failure
https://www.php.net/manual/en/function.eio-busy.php#refsect1-function.eio-busy-returnvalues

(On the phpstorm-stubs, it was documented `resource|bool` but I also updated the return type in https://github.com/JetBrains/phpstorm-stubs/pull/1759)


ondrejmirtes:
Thank you.

PR #4125, opened 2025-07-19
Update stream_filter_ signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12380

Cf https://github.com/php/php-src/blob/f5aa69a5fc760463af8ffdd474119fc27eee83c1/ext/standard/basic_functions.stub.php#L3423

I think it can be anything (ie mixed) because people might write custom filter working with any shape of params.


ondrejmirtes:
Thank you.

PR #4130, opened 2025-07-20
Improve trim on numeric string

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10860


ondrejmirtes:
Thank you.

PR #4131, opened 2025-07-20
Nothing is lower than null and object is never equal to null

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10719
Closes https://github.com/phpstan/phpstan/issues/9141

$a < null is always false.
$a <= null is always when $a is an object.

It needs https://github.com/phpstan/phpstan-src/pull/4132 first.

### DIFF
--- a/src/Type/NullType.php
+++ b/src/Type/NullType.php
@@ -110,6 +110,10 @@ class NullType implements ConstantScalarType
 			return $otherType->isGreaterThan($this, $phpVersion);
 		}
 
+		if ($otherType->isObject()->yes()) {
+			return TrinaryLogic::createYes();
+		}
+
 		return TrinaryLogic::createMaybe();
 	}
 
@@ -121,6 +125,10 @@ class NullType implements ConstantScalarType
 
 		if ($otherType instanceof CompoundType) {
 			return $otherType->isGreaterThanOrEqual($this, $phpVersion);
+		}
+
+		if ($otherType->isObject()->yes()) {
+			return TrinaryLogic::createYes();
 		}
 
 		return TrinaryLogic::createMaybe();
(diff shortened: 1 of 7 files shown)


src/Type/Traits/UndecidedComparisonCompoundTypeTrait.php
```diff
@@ -13,11 +13,19 @@ trait UndecidedComparisonCompoundTypeTrait
 
 	public function isGreaterThan(Type $otherType, PhpVersion $phpVersion): TrinaryLogic
 	{
+		if ($otherType->isNull()->yes() && $this->isObject()->yes()) {
```
VincentLanglet, line 16: This add some logic in the trait but I didn't find another way except creating an ObjectComparisonCompoundTypeTrait and an ObjectComparisonTypeTrait.

Do you prefer to create two new traits ?

src/Type/NullType.php
```diff
@@ -110,6 +110,10 @@ public function isSmallerThan(Type $otherType, PhpVersion $phpVersion): TrinaryL
 			return $otherType->isGreaterThan($this, $phpVersion);
 		}
 
+		if ($otherType->isObject()->yes()) {
```
ondrejmirtes, line 113: The PR title says "Nothing is lower than null" but this change sounds to me like the opposite.

The method asks "is null smaller than object?" and you return "yes".

Maybe I'm running on fumes with only half the brain but I'd like this to be explained. Thanks.

ondrejmirtes (changes requested):

src/Type/NullType.php
VincentLanglet: My bad, I missed some explanation.

Nothing is lower then null so
`null <= $anything` is always true
`$anything < null` is always false

And an object cannot be equal to null  so
`null < $object` is always true
`$object <= null` is always false

ondrejmirtes:
Thank you!

PR #4132, opened 2025-07-20
Fix intersection with substracted

VincentLanglet:
Without this fix, the intersection of
```
new ObjectType(DateTime::class),
new MixedType(subtractedType: new NullType()),
```
was `DateTime~null`.

Extracted from https://github.com/phpstan/phpstan-src/pull/4131


ondrejmirtes:
Thank you.

PR #4134, opened 2025-07-20
Understand at least one argument for non empty array unpacked

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10020

Closes https://github.com/phpstan/phpstan/issues/8506

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -188,6 +188,16 @@ final class FunctionCallParametersCheck
 							$arg->getStartLine(),
 						];
 					}
+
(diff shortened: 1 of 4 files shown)


src/Rules/FunctionCallParametersCheck.php
```diff
@@ -188,6 +188,16 @@ public function check(
 							$arg->getStartLine(),
 						];
 					}
+
+					if (count($arguments) === 0 && $type->isIterableAtLeastOnce()->yes()) {
```
VincentLanglet, line 192: This happen when the constant array is non-empty but all the keys are optional.

ondrejmirtes:
Thank you!

PR #4135, opened 2025-07-20
Add non regression test for 9706

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9706

This was solved by https://github.com/phpstan/phpstan-src/blob/2.1.x/patches/dom_c.patch


ondrejmirtes:
Thank you!

PR #4136, opened 2025-07-20
Add non regression test for 9401

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9401

The issue was about the `array<int<0, max>, int<0, max>> might not be a list.` hint which 
- was incorrect
- which is not here anymore (https://phpstan.org/r/f6667d80-d551-419b-9e03-fd8a1508daba)


ondrejmirtes:
Thank you!

PR #4137, opened 2025-07-20
Update DomNode return type

VincentLanglet:
Cf https://www.php.net/manual/en/class.domnode.php

I copied https://github.com/phpstan/phpstan-src/pull/3737 and used benevolent union.


ondrejmirtes:
Thank you!

PR #4140, opened 2025-07-21
Handle empty array case via `ConstantArrayType::spliceArray()`

herndlm:
This basically acts as recursion guard because a bit later the following piece of code

```php
		if ($offset === null || $length === null) {
			return $this->degradeToGeneralArray()
				->spliceArray($offsetType, $lengthType, $replacementType);
		}
```

would degrade the empty constant array to the same empty constant array, calling the same method again, which then leads to infinite recursion.
We have that exact same code for `sliceArray()` as well btw.

Closes https://github.com/phpstan/phpstan/issues/13279

Was introduced in https://github.com/phpstan/phpstan-src/pull/3952


ondrejmirtes:
Perfect, thank you!

PR #4143, opened 2025-07-21
Check for identifiers before identifier

VincentLanglet:


ondrejmirtes (changes requested):
Please add a test. "stringifyPattern" is used to produce messages like `Ignored error pattern %s was not matched in reported errors.` so the best place to test that is in AnalyserTest.

ondrejmirtes (changes requested):
On a 2nd look, I don't get it. An entry with both `identifier` and `identifiers` should be considered invalid.

I'm arguing in https://github.com/phpstan/phpstan-src/pull/4111 we should move the ever-exploding parameters schema validation of `ignoreErrors` to an actual imperative code in ContainerFactory.

Could you look at it instead? Thank you.

VincentLanglet:
Closing in favor of https://github.com/phpstan/phpstan-src/pull/4309 (I think)

PR #4145, opened 2025-07-22
Fix args are mistakenly handled as immediately-invoked

staabm:
closes https://github.com/phpstan/phpstan/issues/13288
closes https://github.com/phpstan/phpstan/issues/13311
closes https://github.com/phpstan/phpstan/issues/13331
closes https://github.com/phpstan/phpstan/issues/13307
closes https://github.com/phpstan/phpstan/issues/12119

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5139,21 +5139,20 @@ final class NodeScopeResolver
 				$scopeToPass = $closureBindScope;
 			}
 
-			if (
-				$parameterType === null
-				|| $parameterType instanceof MixedType
-				|| $parameterType->isCallable()->no()
-			) {
-				$callCallbackImmediately = false;
-			} elseif ($parameter instanceof ExtendedParameterReflection) {
+			$parameterCallableType = null;
+			if ($parameterType !== null) {
+				$parameterCallableType = TypeUtils::findCallableType($parameterType);
+			}
+
+			if ($parameter instanceof ExtendedParameterReflection) {
 				$parameterCallImmediately = $parameter->isImmediatelyInvokedCallable();
 				if ($parameterCallImmediately->maybe()) {
-					$callCallbackImmediately = $calleeReflection instanceof FunctionReflection;
+					$callCallbackImmediately = $parameterCallableType !== null && $calleeReflection instanceof FunctionReflection;
 				} else {
 					$callCallbackImmediately = $parameterCallImmediately->yes();
 				}
 			} else {
-				$callCallbackImmediately = $calleeReflection instanceof FunctionReflection;
+				$callCallbackImmediately = $parameterCallableType !== null && $calleeReflection instanceof FunctionReflection;
 			}
 
 			if ($arg->value instanceof Expr\Closure) {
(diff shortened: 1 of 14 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -5139,7 +5139,13 @@ private function processArgs(
 				$scopeToPass = $closureBindScope;
 			}
 
-			if ($parameter instanceof ExtendedParameterReflection) {
+			if (
+				$parameterType === null
+				|| $parameterType instanceof MixedType
+				|| $parameterType->isCallable()->no()
```
staabm: to support signatures like
`public function filter(?callable $callback = null): Deque {}`

and call-sites like
```php
			try {
				$deque->filter(function () {
				});
			} catch (\InvalidArgumentException $e) {

			}
```

which is tested in `tests/PHPStan/Rules/Exceptions/data/catch-with-unthrown-exception-stubs.php`

this did not work using `!$parameterType->isCallable()->yes()`

staabm: the fix also has impact on whether/when impurePoints, throwPoints are evaluated in cases when arrow functions/closures are passed to non immediately invoking functions.

I think it needs a additional test-case, but I did not yet found the place to test it

edit: fixed ✅

ondrejmirtes: I think we need a new `TypeUtils` method. `findCallableType` perhaps. It'd return the type for `isCallable()->yes()`. Additionally it should recursively inspect unions for types like `?callable` or `?Closure`.

ondrejmirtes: But I think this logic is wrong. What mainly has to be handled is `$parameterCallImmediately->maybe()`. That's where you need to ask about this. And also in the `else` section where it's `$callCallbackImmediately = $calleeReflection instanceof FunctionReflection;` right now.

ondrejmirtes: This logic fails for a couple reasons:

1) If the parameter is `mixed` but marked with `@param-immediately-invoked-callable` it'd get silently ignored.
2) It'd still mark parameter as immediately invoked if the type is for example `string` (which is maybe-callable).

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: thanks for the in-detail feedback. there are a lot of things to consider, I did not see yet.
additionally https://github.com/phpstan/phpstan/issues/13331#issuecomment-3136382085 was useful.

I tried adding more tests and adding comments about expectations, of what I understood from the comments.

thank you

src/Type/TypeUtils.php
```diff
@@ -170,6 +170,24 @@ public static function findThisType(Type $type): ?ThisType
 		return null;
 	}
 
+	public static function findCallableType(Type $type): ?Type
+	{
+		if ($type->isCallable()->yes()) {
+			return $type;
+		}
+
+		if ($type instanceof UnionType || $type instanceof IntersectionType) {
```
ondrejmirtes: I don't think we need to go into intersections. Once one of the intersected types is callable, the whole type is callable.

ondrejmirtes (changes requested):
Additionally please add a regression test for 12119

ondrejmirtes:
Thank you!

PR #4146, opened 2025-07-22
use benevolent union for `stream_get_contents()` to prevent strict-rule errors

staabm:
closes https://github.com/phpstan/phpstan/issues/13289


ondrejmirtes:
Thank you.

PR #4147, opened 2025-07-22
Fix missing detection of dead code in arrow functions

staabm:
followup to https://github.com/phpstan/phpstan-src/pull/4090


ondrejmirtes:
Thank you.

PR #4148, opened 2025-07-22
Fix missing detection of dead code in closures

staabm:
followup to https://github.com/phpstan/phpstan-src/pull/4090

requires https://github.com/phpstan/phpstan-src/pull/4145

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4690,7 +4690,7 @@ final class NodeScopeResolver
 				array_merge($statementResult->getImpurePoints(), $closureImpurePoints),
 			), $closureScope);
 
-			return new ProcessClosureResult($scope, $statementResult->getThrowPoints(), $statementResult->getImpurePoints(), $invalidateExpressions);
+			return new ProcessClosureResult($scope, $statementResult->getThrowPoints(), $statementResult->getImpurePoints(), $invalidateExpressions, $statementResult->isAlwaysTerminating());
 		}
 
 		$count = 0;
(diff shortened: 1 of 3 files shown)


staabm:
@janedbal could you have a look at the shipmonk results of this PR?

I had a look at the shipmon/phpstan-rules results, and I think these are correct, as only the [1st callable is immediately invoked](https://github.com/shipmonk-rnd/phpstan-rules/blob/9a2a033d12a516a73056bb215e6e4dec0dd33701/tests/Rule/data/ForbidCheckedExceptionInCallableRule/code.php#L106-L112) - but the tests looks like they expect also the 2nd callable would be immediately invoked..?

janedbal:
@staabm I'm not sure what you mean. The rule denies throwing checked exception in closures unless such closure is passed directly to callable param marked immediately invoked. Thus:

```php
/**
 * @param-immediately-invoked-callable $callable
 */
public function immediateThrow(
    callable $callable,  // this callable can be closure with checked exception thrown
    ?callable $denied = null // this cannot, as we dont know when it will be called, and propagation of such exception cannot be ensured by PHPStan
): void
```

But I'm not sure why the test is failing since this MR. I do pass always-terminating closure there as second arg:

```php
$this->immediateThrow(
    function () {},
    function () {
        throw new CheckedException(); // rule error, not reported anymore
    },
);
```

I believe it should still be reported. I didnt breakpoint it, but it looks like [ThrowPoints of the closure might be gone](https://github.com/shipmonk-rnd/phpstan-rules/blob/9a2a033d12a516a73056bb215e6e4dec0dd33701/src/Rule/ForbidCheckedExceptionInCallableRule.php#L201)

staabm:
I think the problem is that

```
		$self = $this; // self is unknown variable in scope of the closure
		$self->immediateThrow(function () {
			throw new CheckedException();
		});
```

immediately throws and therefore all following statements are dead code.
therefore we get no errors for the lines below that (after this PR)

janedbal:
I see, let me adjust the tests there so that we can have it green here.

ondrejmirtes:
Thank you.

PR #4149, opened 2025-07-22
Fix count inferences

VincentLanglet:
This https://phpstan.org/r/ed2e2eec-bb85-48b8-8f8c-8ffad4719790 shows the issue.

The constant array shouldnt be narrowed.
Currently the code is narrowing list and constant array which looks like a list.

But one think we can do is telling the array is non empty.

Closes https://github.com/phpstan/phpstan/issues/13111


ondrejmirtes:
Thank you!

PR #4150, opened 2025-07-23
Fix MethodExistsTypeSpecifyingExtension for union types

staabm:
closes https://github.com/phpstan/phpstan/issues/13272


staabm:
@liamduckett I just realized that the fix of this PR produces a false-positive (see the newly added testcase).
I do not yet know how to fix the initial problem without also regressing the other example.

just to let you know, as https://github.com/phpstan/phpstan-src/pull/4153 has the same problem

tests/PHPStan/Analyser/nsrt/bug-13272.php
```diff
@@ -0,0 +1,32 @@
...
+use function PHPStan\Testing\assertType;
+
+function foo(object $bar): void
+{
+	foreach (['qux', 'quux'] as $method) {
+		assertType("'quux'|'qux'", $method);
+
+		if (!method_exists($bar, $method)) {
+			throw new \Exception;
+		}
```
staabm, line 14: I think the we could build a better type in a `TypeSpecifyingExtension` if the `TypeSpecifierContext` would contain information, whether the context is always-terminating.

liamduckett:
> @liamduckett I just realized that the fix of this PR produces a false-positive (see the newly added testcase). I do not yet know how to fix the initial problem without also regressing the other example.
> 
> just to let you know, as #4153 has the same problem

Thanks. I'm not sure I understand the new test case (probably due to my inexperience) - why would `$bar` being narrowed here, be a problem?

staabm:
the initial types were wrongly narrowed:

```
/**
 * @param 'quux'|'qux' $constUnion
 */
function fooBar(object $bar, string $constUnion): void
{
	if (!method_exists($bar, $constUnion)) {
		throw new \Exception;
	}

    // $bar does not have both methods, it has only either 'quux' or 'qux' but not both
}
```

---


while in the initial loop check example it is correct that $bar will have both methods at once:

```
foreach (['qux', 'quux'] as $method) {
		assertType("'quux'|'qux'", $method);

		if (!method_exists($bar, $method)) {
			throw new \Exception;
		}
}
```

PR #4151, opened 2025-07-23
Fix `array_key_exists` narrows `$key` too much

staabm:
closes https://github.com/phpstan/phpstan/issues/11724

---

for `isset($a[$f])` it already works correctly
- https://phpstan.org/r/b39a70df-b2cd-4ce1-8509-2cd92ea3e641
- https://3v4l.org/9Sv5F#veol


ondrejmirtes:
Thank you!

PR #4152, opened 2025-07-23
More precise intval and floatval signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan-src/pull/1005
Closes https://github.com/phpstan/phpstan/issues/6560

This will give the same behavior for cast vs intval/floatval functions in 
https://phpstan.org/r/850196c1-5f6c-479d-a25a-dd303e9cd047

Not sure if it's worth a non-regression test... should I write a test to expect all the `[argument.type](https://phpstan.org/error-identifiers/argument.type)` errors ?


resources/functionMap.php
```diff
@@ -5614,7 +5614,7 @@
 'intltz_to_date_time_zone' => ['DateTimeZone|false', 'obj'=>''],
 'intltz_use_daylight_time' => ['bool', 'obj'=>''],
 'intlz_create_default' => ['IntlTimeZone'],
-'intval' => ['int', 'var'=>'scalar|array|resource|null', 'base='=>'int'],
+'intval' => ['int', 'var'=>'scalar|array|null', 'base='=>'int'],
```
staabm, line 5078: should we reflect a similar signature in `boolval()` and `strval()`?

VincentLanglet: If you play with https://phpstan.org/r/850196c1-5f6c-479d-a25a-dd303e9cd047 you'll see that
- (string) resource
- (bool) resource

Is not reported.

So `strval` and `boolval` shouldn't too.

ondrejmirtes (changes requested):
Looking at https://phpstan.org/r/62904f7f-1538-422c-a6c4-f3fb53fc40a4 I don't think this is enough. We should have regression test in both CallToFunctionParametersRuleTest and InvalidCastRuleTest making sure they behave the same.

VincentLanglet:
> Looking at [phpstan.org/r/62904f7f-1538-422c-a6c4-f3fb53fc40a4](https://phpstan.org/r/62904f7f-1538-422c-a6c4-f3fb53fc40a4) I don't think this is enough. We should have regression test in both CallToFunctionParametersRuleTest and InvalidCastRuleTest making sure they behave the same.

The pr wasn't good at all, I reworked it with https://github.com/phpstan/phpstan-src/pull/4314.

And the whole issue is not closed yet indeed.

PR #4153, opened 2025-07-23
Fix PropertyExistsTypeSpecifyingExtension for union types

liamduckett:
closes [https://github.com/phpstan/phpstan/issues/13304](https://github.com/phpstan/phpstan/issues/13304)

@staabm it'd be great to [take you up](https://github.com/phpstan/phpstan/issues/13272#issuecomment-3106612286) on any guidance - especially because this PR is so heavily based on your changes in https://github.com/phpstan/phpstan-src/pull/4150. Thanks!

I believe my changes allow narrowing based on an array of constant strings, unless:
1. Any of them are empty
2. Any of them aren't native

I was trying to reason as to whether it'd make sense to try and narrow based on parts of the union (for example, if it contains an empty string, but also a non empty string?) but I think I'm out of brainpower for the day.

### DIFF
--- a/src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
+++ b/src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
@@ -68,10 +68,6 @@ final class PropertyExistsTypeSpecifyingExtension implements FunctionTypeSpecify
 
 		$hasPropertyTypes = [];
 		foreach ($propertyNameTypes as $propertyNameType) {
-			if($propertyNameType->getValue() === '') {
-				return new SpecifiedTypes([], []);
-			}
-
 			$hasPropertyTypes[] = new HasPropertyType($propertyNameType->getValue());
 		}
 
@@ -82,6 +78,10 @@ final class PropertyExistsTypeSpecifyingExtension implements FunctionTypeSpecify
 			$propertyNodes = [];
 
 			foreach($propertyNameTypes as $propertyNameType) {
+				if($propertyNameType->getValue() === '') {
+					return new SpecifiedTypes([], []);
+				}
+
 				$propertyNodes[] = new PropertyFetch(
 					$node->getArgs()[0]->value,
 					new Identifier($propertyNameType->getValue()),
--- a/src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
+++ b/src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
@@ -68,6 +68,10 @@ final class PropertyExistsTypeSpecifyingExtension implements FunctionTypeSpecify
 
 		$hasPropertyTypes = [];
 		foreach ($propertyNameTypes as $propertyNameType) {
+			if($propertyNameType->getValue() === '') {
+				return new SpecifiedTypes([], []);
+			}
+
 			$hasPropertyTypes[] = new HasPropertyType($propertyNameType->getValue());
 		}
 
--- a/src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
+++ b/src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
@@ -55,8 +55,9 @@ final class PropertyExistsTypeSpecifyingExtension implements FunctionTypeSpecify
 		TypeSpecifierContext $context,
 	): SpecifiedTypes
 	{
-		$propertyNameType = $scope->getType($node->getArgs()[1]->value);
-		if (!$propertyNameType instanceof ConstantStringType) {
+		$propertyNameTypes = $scope->getType($node->getArgs()[1]->value)->getConstantStrings();
+
+		if ($propertyNameTypes === []) {
 			return $this->typeSpecifier->create(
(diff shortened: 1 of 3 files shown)


liamduckett:
Looks like I need to remove something from the baseline, and probably rejig my empty string check to be later down.

src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
```diff
@@ -65,34 +66,45 @@ public function specifyTypes(
...
-			$propertyNode = new PropertyFetch(
-				$node->getArgs()[0]->value,
-				new Identifier($propertyNameType->getValue()),
-			);
+			$propertyNodes = [];
+
+			foreach($propertyNameTypes as $propertyNameType) {
+				$propertyNodes[] = new PropertyFetch(
+					$node->getArgs()[0]->value,
+					new Identifier($propertyNameType->getValue()),
```
staabm: identifier expecs a non-emtpy-string and within your loop you did not make sure `$propertyNameType->getValue()` is always non-empty.

you can fix this in the same way you already did in a similar loop above.

staabm (changes requested):
hey,

thanks for the pull request. it looks mostly good already.

I see the following todos atm
- look at the PHPStan errors of the PHP8.x builds first (the PHP7.x errors are likely reported on the wrong line, as the source code is transformed for PHP7.x compat). Usually most of the PHP7.x errors go away after you fix the PHP8 ones.
- after you fixed all newly reported PHPStan errors, you can run `make phpstan-generate-baseline` to cleanup the "Ignored error pattern XXX" errors
- run `make cs-fix` to fix the [Lint / Coding Standard](https://github.com/phpstan/phpstan-src/actions/runs/16480124697/job/46591931380?pr=4153) problems
- run `make` locally to verify all sanity checks are green
- some of the GitHub Action checks fail unrelated to this PR. to get an idea which, its oftentimes usefull to look at other recent PRs and see whether these report the same problems.

liamduckett:
Thanks @staabm. I think most of the failing action checks are from other PRs, as you described?

The exception to this is [doctrine/orm](https://github.com/doctrine/orm) ([see action](https://github.com/phpstan/phpstan-src/actions/runs/16503796366/job/46669418224?pr=4153)), in which I think we've found legitimate issues to report. I believe the simplest example of this is [here](https://github.com/doctrine/orm/blob/62ca8424d872e1ef8c443cf089be0687b7663a9d/src/Mapping/DiscriminatorColumnMapping.php#L60), but they all follow this general pattern.

src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
```diff
@@ -65,24 +66,37 @@ public function specifyTypes(
...
 
-		if ($propertyNameType->getValue() === '') {
-			return new SpecifiedTypes([], []);
+		$hasPropertyTypes = [];
+		foreach ($propertyNameTypes as $propertyNameType) {
+			$hasPropertyTypes[] = new HasPropertyType($propertyNameType->getValue());
 		}
 
 		$objectType = $scope->getType($node->getArgs()[0]->value);
 		if ($objectType instanceof ConstantStringType) {
```
staabm: I guess this check can be replaced with `isString()`  similar to the else if

staabm: Maybe it can even be dropped and the else-case is sufficient

staabm:
> The exception to this is doctrine/orm (see action), in which I think we've found legitimate issues to report. I believe the simplest example of this is here, but they all follow this general pattern.

These look good to me

src/Type/Php/PropertyExistsTypeSpecifyingExtension.php
liamduckett: Sorry, I'm not sure I follow. What could be replaced / dropped?

staabm: `if ($objectType instanceof ConstantStringType) {` looks redundant with the `else` part


I guess we can remove the if-branch

liamduckett:
Thanks again!

staabm (approved):

VincentLanglet:
You might want to mark this PR as ready @liamduckett ?

tests/PHPStan/Analyser/nsrt/bug-13304.php
```diff
@@ -0,0 +1,16 @@
...
+use function PHPStan\Testing\assertType;
+
+function foo(object $foo): void
+{
+	foreach (['qux', 'quux'] as $property) {
+		if (!property_exists($foo, $property)) {
+			throw new \Exception;
+		}
+
+		assertType("object&hasProperty(quux)&hasProperty(qux)", $foo);
```
ondrejmirtes, line 14: Hey, this doesn't make sense. The assertType only makes sense below the foreach.

`property_exists($foo, $property)` with `'qux'|'quux'` does not guarantee that both properties are accessible on the object.

ondrejmirtes (changes requested):

PR #4154, opened 2025-07-24
Fix ImpossibleInstanceOfRule

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10036

Closes https://github.com/phpstan/phpstan-src/pull/2769

Seems like it was implemented but ignored because of an early return.
I just fixed it and add it the usage of the rule levelhelper.

### DIFF
--- a/tests/PHPStan/Rules/Classes/data/bug-10036.php
+++ b/tests/PHPStan/Rules/Classes/data/bug-10036.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Bug10036;
+
+use stdClass;
+
+class HelloWorld
+{
+	public function sayHello(?string $class): void
+	{
+		var_dump(new stdClass instanceof $class);
+	}
+
+	/**
+	 * @param string|null $class
+	 */
+	public function sayWorld($class): void
+	{
+		var_dump(new stdClass instanceof $class);
+	}
+
+	public function sayString(string $class): void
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Classes/data/bug-10036.php
```diff
@@ -0,0 +1,31 @@
...
+	{
+		var_dump(new stdClass instanceof $class);
+	}
+
+	public function sayString(string $class): void
+	{
+		var_dump(new stdClass instanceof $class);
+	}
+
+	public function sayMixed(mixed $class): void
```
staabm, line 27: `array` passed into `instanceof` should also error, see https://3v4l.org/6h0tt (needs test)

staabm: please also test, that a `object` can be used in `instanceof`:

https://3v4l.org/JVXHS

staabm (commented):
thanks for taking over!

tests/PHPStan/Rules/Classes/data/bug-10036.php
VincentLanglet: I added the two tests.

staabm (approved):

ondrejmirtes:
Thank you.

PR #4155, opened 2025-07-24
Replace instanceof checks in JsonThrowOnErrorDynamicReturnTypeExtension

Khartir:
Hope this is OK, I just cloned this rule for https://github.com/thecodingmachine/phpstan-safe-rule/pull/61 and found these deprecated checks.

### DIFF
--- a/src/Type/Php/JsonThrowOnErrorDynamicReturnTypeExtension.php
+++ b/src/Type/Php/JsonThrowOnErrorDynamicReturnTypeExtension.php
@@ -12,12 +12,12 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\BitwiseFlagHelper;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantStringType;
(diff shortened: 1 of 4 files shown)


staabm (approved):

src/Type/Php/JsonThrowOnErrorDynamicReturnTypeExtension.php
```diff
@@ -109,7 +115,7 @@ private function isForceArray(FuncCall $funcCall, Scope $scope): bool
 		}
 
 		$secondArgType = $scope->getType($args[1]->value);
-		$secondArgValue = $secondArgType instanceof ConstantScalarType ? $secondArgType->getValue() : null;
+		$secondArgValue = count($secondArgType->getConstantScalarValues()) === 1 ? $secondArgType->getConstantScalarValues()[0] : null;
```
ondrejmirtes: `$secondArgType->getConstantScalarValues()` with more than one value should also be properly handled here

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4157, opened 2025-07-25
Fix Promoted vs. regular property inconsistency with generics

sayuprc:
resolve phpstan/phpstan#13049

Fixed readonly for constructor promotions.


tests/PHPStan/Analyser/Bug13049Test.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @extends RuleTestCase<PropertyVarianceRule>
+ */
+class Bug13049Test extends RuleTestCase
+{
+
+	protected function getRule(): Rule
+	{
+		return new PropertyVarianceRule(new VarianceCheck());
```
ondrejmirtes: The fix is great, but please put this into existing PropertyVarianceRuleTest instead of adding a new test case. Thank you.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/Bug13049Test.php
sayuprc: Thank you!
I fixed in [4e32f77](https://github.com/phpstan/phpstan-src/pull/4157/commits/4e32f77c7f9fd23d91f2ae705d7c8580230c1dc4).

ondrejmirtes:
Thank you!

PR #4159, opened 2025-07-26
Improve implode signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/5760
Closes https://github.com/phpstan/phpstan/issues/12079

### DIFF
--- a/src/Reflection/ParametersAcceptorSelector.php
+++ b/src/Reflection/ParametersAcceptorSelector.php
@@ -205,7 +205,7 @@ final class ParametersAcceptorSelector
 			}
 
 			if (count($args) <= 2 && (bool) $args[0]->getAttribute(ImplodeArgVisitor::ATTRIBUTE_NAME)) {
-				$acceptor = $parametersAcceptors[0];
+				$acceptor = $namedArgumentsVariants[0] ?? $parametersAcceptors[0];
 				$parameters = $acceptor->getParameters();
 				if (isset($args[1]) || ($args[0]->name !== null && $args[0]->name->name === 'array')) {
 					$parameters = [
--- a/src/Reflection/ParametersAcceptorSelector.php
+++ b/src/Reflection/ParametersAcceptorSelector.php
@@ -209,16 +209,15 @@ final class ParametersAcceptorSelector
 				$parameters = $acceptor->getParameters();
 				if (isset($args[1]) || ($args[0]->name !== null && $args[0]->name->name === 'array')) {
 					$parameters = [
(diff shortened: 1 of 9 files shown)


src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -203,6 +204,31 @@ public static function selectFromArgs(
 				];
 			}
 
+			if (isset($args[0]) && (bool) $args[0]->getAttribute(ImplodeArgVisitor::ATTRIBUTE_NAME)) {
```
staabm: I don't think the isset is necessary

```suggestion
			if ((bool) $args[0]->getAttribute(ImplodeArgVisitor::ATTRIBUTE_NAME)) {
```

see also https://github.com/phpstan/phpstan-src/pull/4160

src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -203,6 +204,32 @@ public static function selectFromArgs(
...
+				if (isset($args[1]) || ($args[0]->name !== null && $args[0]->name->name === 'array')) {
+					$parameters = [
+						new NativeParameterReflection($parameters[0]->getName(), false, new StringType(), PassedByReference::createNo(), false, null),
+						new NativeParameterReflection($parameters[1]->getName(), false, new ArrayType(new MixedType(), new MixedType()), PassedByReference::createNo(), false, null),
+					];
+				} else {
+					$parameters = [
+						new NativeParameterReflection($parameters[0]->getName(), false, new ArrayType(new MixedType(), new MixedType()), PassedByReference::createNo(), false, null),
+					];
+				}
```
staabm, line 219: should the signature depend on php version?
https://www.php.net/manual/en/function.implode.php

grafik

VincentLanglet: Phpstan 2.1 only support php 7.4+ and currently we already have an error for the old signature (because it's deprecated)
https://phpstan.org/r/1c458f6f-d34e-433b-9f54-4d9d7e508a7d

Cf https://github.com/phpstan/phpstan-src/commit/bea339a399b7bfa2c6037eb894bc1bb88683d4b1

So I don't think I have anything more to do, unless we already change the behavior for implode on 7.4

ondrejmirtes (changes requested):
Please fix the conflict. It'd probably be the best if you manually check out ParametersAcceptorSelector currently on top of 2.1.x and try to apply your changes in it manually. Because the Git conflict might get hairy.

ondrejmirtes:
Thank you.

PR #4160, opened 2025-07-26
Simplify ParametersAcceptorSelector

staabm:
[review with whitespaces disabled](https://github.com/phpstan/phpstan-src/pull/4160/files?w=1), makes it obvious what changed

### DIFF
--- a/src/Reflection/ParametersAcceptorSelector.php
+++ b/src/Reflection/ParametersAcceptorSelector.php
@@ -267,8 +267,7 @@ final class ParametersAcceptorSelector
 
 			$closureBindToVar = $args[0]->getAttribute(ClosureBindToVarVisitor::ATTRIBUTE_NAME);
 			if (
-				$closureBindToVar !== null
-				&& $closureBindToVar instanceof Node\Expr\Variable
+				$closureBindToVar instanceof Node\Expr\Variable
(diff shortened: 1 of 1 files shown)


src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -234,7 +234,7 @@ public static function selectFromArgs(
 				];
 			}
 
-			if (isset($args[0]) && (bool) $args[0]->getAttribute(ArrayFindArgVisitor::ATTRIBUTE_NAME)) {
```
staabm, line 237: we already work on line 82 with `$arrayMapArgs = $args[0]->value->getAttribute(ArrayMapArgVisitor::ATTRIBUTE_NAME);` without a outer `isset($args[0])`, meaning the following `isset` checks are useless

staabm: looking more closely: we don't need the `isset` as we have an outer `count($args) > 0`

https://github.com/phpstan/phpstan-src/pull/4160/files#diff-c3b9f9c6c2992760688d174886d5d1ef9b5d4d1097b1175d6d5af43f9dc92192R79

ondrejmirtes:
Thank you!

PR #4161, opened 2025-07-26
Use `Type->isInteger()` over `$dimType instanceof ConstantIntegerType`

staabm:


src/Analyser/MutatingScope.php
```diff
@@ -4238,7 +4238,7 @@ public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType,
 						new ObjectType(ArrayAccess::class),
 						new NullType(),
 					];
-					if ($dimType instanceof ConstantIntegerType) {
+					if ($dimType->isInteger()->yes()) {
```
staabm, line 4241: `$dimType instance ConstantIntegerType` is already narrowed in the outer if

ondrejmirtes (changes requested):
Hey, this change isn't really valuable. With the original code it's obvious what's going on. We just want the `if` to be executed only when the left part of BooleanOr is true.

It'd also be nice to change the line with `$dimType instanceof ConstantIntegerType || $dimType instanceof ConstantStringType`. I understand that it might be difficult, in that case feel free to close the PR.

PR #4162, opened 2025-07-26
Fixed `HasOffsetValueType` accessory missing main-type

staabm:
this PR does not fix the initial reported error of the OP in https://github.com/phpstan/phpstan/issues/13270#issue-3242908977 (which I think is a valid error)

but I as I understood ondrej's examples in https://github.com/phpstan/phpstan/issues/13270#issuecomment-3101734893 it fixes the typing problems

closes https://github.com/phpstan/phpstan/issues/13270


src/Type/MixedType.php
```diff
@@ -169,7 +169,15 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
 
 	public function setExistingOffsetValueType(Type $offsetType, Type $valueType): Type
 	{
-		return new self($this->isExplicitMixed);
+		$types = [
+			new ArrayType(new MixedType(), new MixedType()),
+			new ObjectType(ArrayAccess::class),
+			new NullType(),
+		];
+		if ($offsetType->isInteger()->yes()) {
```
VincentLanglet: I think it should be
```suggestion
		if (!$offsetType->isInteger()->no()) {
```

I set $mixed1[$mixed2] ; $mixed1 could still be a string.

Also if I set $mixed1[$intOrString] ; $mixed1 can stil be a string because maybe $intOrString is never a string finally...

src/Type/MixedType.php
```diff
@@ -169,7 +169,15 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
 
 	public function setExistingOffsetValueType(Type $offsetType, Type $valueType): Type
 	{
-		return new self($this->isExplicitMixed);
+		$types = [
+			new ArrayType(new MixedType(), new MixedType()),
+			new ObjectType(ArrayAccess::class),
+			new NullType(),
```
VincentLanglet: $null = null;
$null[0] = 1;
var_dump($null); // Gives `[1]`

So you can never have `null` after setting something no ?

tests/PHPStan/Analyser/nsrt/composer-array-bug.php
```diff
@@ -19,22 +19,22 @@ public function doFoo(): void
 			foreach ($this->config['authors'] as $key => $author) {
 				if (!is_array($author)) {
 					$this->errors[] = 'authors.'.$key.' : should be an array, '.gettype($author).' given';
-					assertType("mixed", $this->config['authors']);
```
VincentLanglet, line 22: That's half related to the PR, but since we have `!empty($this->config['authors'])` ; shouldn't we already know it's 
```
assertType("mixed~(0|0.0|'0'|''|array{}|false|null)",
```
or I misunderstand something ?

tests/PHPStan/Analyser/nsrt/composer-array-bug.php
```diff
@@ -19,22 +19,22 @@ public function doFoo(): void
 			foreach ($this->config['authors'] as $key => $author) {
```
VincentLanglet, line 20: Since we're doing `foreach ($this->config['authors'] as $key => $author)` we should be able to reduce the type of 
```
$this->config['authors']
```
to `array|object`.

This would avoid having later
```
assertType("array|ArrayAccess|string", $this->config['authors']);
```
when it cannot be a string.

tests/PHPStan/Analyser/nsrt/composer-array-bug.php
VincentLanglet: We could even be more precise with https://github.com/phpstan/phpstan-src/pull/4162#discussion_r2232995435

staabm: yeah, it should be more precise.

I guess its something similar to https://github.com/phpstan/phpstan/issues/13312

tests/PHPStan/Analyser/nsrt/composer-array-bug.php
staabm: > Since we're doing `foreach ($this->config['authors'] as $key => $author)` we should be able to reduce the type


I think even `non-empty-array|Traversable`.

looking into it right now

VincentLanglet: > I think even `non-empty-array|Traversable`.

You can technically foreach on an object I think (https://github.com/php/doc-en/issues/4790)
So non-empty-array|object ; maybe ?

But since phpstan forbids it... https://phpstan.org/r/86edf6b6-5e8e-4b7b-a313-fba55fb8c8f5
I think you're right about non-empty-array|Traversable.

> looking into it right now

Nice, because I feel like merging this PR without the `non-empty-array|object` inference will create more false-positive.

When iterating on mixed and setting value, it will become array|ArrayAccess|string and you won't be able to iterate on it anymore.

staabm: just committed a first iteration. it doesn't work in all cases yet.

tests/PHPStan/Analyser/nsrt/bug-13312.php
```diff
@@ -0,0 +1,41 @@
...
+	assertType('list<mixed>', $arr);
+}
+
+
+function fooBar(mixed $mixed): void {
+	assertType('mixed', $mixed);
+	foreach ($mixed as $v) {
+		assertType('non-empty-array|Traversable', $mixed);
+	}
+	assertType('array|object', $mixed);
```
staabm, line 40: parts of the errors are triggered because this is not `array|Traversable`

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -5816,7 +5816,7 @@ public static function dataIterable(): array
 				'$iterableWithoutTypehint[0]',
 			],
 			[
-				'iterable',
+				'array|Traversable',
```
VincentLanglet: If `array|Traversable` and `iterable` are the same, wouldn't it better to specify the type to `iterable` instead in order to produce less change on existing resolved type (and avoid modifying existing error messages).

That would say we should use
```
private function getForeachType(): Type
	{
		return new IterableType(new MixedType(), new MixedType());
	}
```
?

tests/PHPStan/Analyser/nsrt/composer-array-bug.php
staabm: ok, I think we are in a usable state now.

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
VincentLanglet: Wdyt @staabm ?

staabm: changing it to `iterable` indedd leads to less type changes.
[some test start failling so](https://phpstan.org/r/dd6129ef-bbd1-4820-a7b9-5f1f2ba9f88d), which I think is another bug I need to investigate.

staabm:
required a bit exploring, and back and forth. reduced the scope of the PR a bit and will send followups instead to keep this one simple. 

@ondrejmirtes I think we are good to go now :)

ondrejmirtes:
Thank you!

ondrejmirtes:
I decided for a different approach here: https://github.com/phpstan/phpstan-src/commit/56a30c2e06c33f6af2639aa7a58c901962539c31

My reasoning is in the commit description.

PR #4163, opened 2025-07-26
fix bug 13292 (Invalid UTF-8)

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/13292

The patch is from https://github.com/symfony/symfony/pull/61242 rebased onto symfony/console 5.4. The PR contains a BC break (not relevant for PHPStan), so I'm not sure whether they'll accept it. And since PHPStan uses v5.4, it wouldn't get the fix anyway.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -129,6 +129,9 @@
 			],
 			"react/http": [
 				"patches/Sender.patch"
+			],
+			"symfony/console": [
+				"patches/OutputFormatter.patch"
 			]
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you!

parijke:
bug confirmed on 2.1.19

parijke:
Confirm fixed in 2.1.20

PR #4164, opened 2025-07-26
Detect more possibly-impossible offset

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/1061

### DIFF
--- a/tests/PHPStan/Levels/data/stringOffsetAccess-7.json
+++ b/tests/PHPStan/Levels/data/stringOffsetAccess-7.json
@@ -1,5 +1,10 @@
 [
     {
+        "message": "Offset int|object might not exist on 'foo'.",
+        "line": 19,
+        "ignorable": true
+    },
+    {
         "message": "Offset 'foo' might not exist on array|string.",
         "line": 27,
         "ignorable": true
@@ -10,8 +15,18 @@
         "ignorable": true
     },
     {
+        "message": "Offset int|object might not exist on array|string.",
+        "line": 35,
+        "ignorable": true
+    },
+    {
         "message": "Possibly invalid array key type int|object.",
         "line": 35,
+        "ignorable": true
+    },
+    {
+        "message": "Offset int|object might not exist on array{baz: 21}|array{foo: 17, bar: 19}.",
+        "line": 55,
         "ignorable": true
     },
     {
--- a/tests/PHPStan/Levels/data/stringOffsetAccess-7.json
+++ b/tests/PHPStan/Levels/data/stringOffsetAccess-7.json
@@ -25,11 +25,6 @@
         "ignorable": true
     },
     {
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
I feel like the error could always be reported on level 7+.

The difference is that the `reportPossiblyUndefined...` option is for offsets like `string`, but in this case this is about a union and one of the union parts never exists on the constant array so it'd be a good fit for L7+ without any extra option.

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -127,6 +131,10 @@ public function testRule(): void
 				'Offset int does not exist on array<string, string>.',
 				312,
 			],
+			[
+				'Offset int|null might not exist on array<string, string>.',
```
ondrejmirtes, line 135: Well this is totally wrong, `null` gets cast to `''` in an array key and that can exist on `array<string, string>.`. Not sure where to do the fix.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
VincentLanglet: You misunderstood the error message.

You can see that phpstan already report on line 312 (the error above)
```
'Offset int does not exist on array<string, string>.',
```

So here we're reporting the same message with "might not exist" certainty because
- int does not exists
- null can exist

tests/PHPStan/Levels/data/stringOffsetAccess-7.json
```diff
@@ -9,11 +14,21 @@
...
+        "line": 35,
+        "ignorable": true
+    },
     {
         "message": "Possibly invalid array key type int|object.",
         "line": 35,
         "ignorable": true
     },
+    {
+        "message": "Offset int|object might not exist on array{baz: 21}|array{foo: 17, bar: 19}.",
```
ondrejmirtes: This looks wrong for `array{foo: 17, bar: 19}|array{baz: 21}`. `int|object` never exists on these arrays.

ondrejmirtes (changes requested):
I'm still not so sure about this, the impact seems high...

tests/PHPStan/Levels/data/stringOffsetAccess-7.json
VincentLanglet: Indeed, I think this is because of https://github.com/phpstan/phpstan-src/pull/4272 which need to be fix/merged first then.

VincentLanglet:
Since they are kinda related, https://github.com/phpstan/phpstan-src/pull/4166 might be better to merge first (?)

ondrejmirtes:
Thank you.

PR #4165, opened 2025-07-26
Add DateIntervalFormatDynamicReturnTypeExtension

VincentLanglet:
Related to https://github.com/phpstan/phpstan/issues/1452

but does not close it unfortunately since the `%a` is complicated.

### DIFF
--- a/src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
+++ b/src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
@@ -7,16 +7,20 @@ use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Reflection\MethodReflection;
-use PHPStan\Type\Accessory\AccessoryLiteralStringType;
+use PHPStan\Type\Accessory\AccessoryLowercaseStringType;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
-use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
+use PHPStan\Type\Accessory\AccessoryNumericStringType;
+use PHPStan\Type\Accessory\AccessoryUppercaseStringType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
-use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use function count;
+use function is_numeric;
+use function strtolower;
+use function strtoupper;
 
 #[AutowiredService]
 final class DateIntervalFormatDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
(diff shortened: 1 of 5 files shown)


src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,89 @@
...
+		if (count($constantStrings) === 0) {
+			if ($arg->isNonEmptyString()->yes()) {
+				return new IntersectionType([new StringType(), new AccessoryNonEmptyStringType()]);
+			}
+
+
+			return null;
+		}
+
+		// The worst case scenario for the non-falsy-string check is that every number are 0.
```
staabm:
```suggestion
		// The worst case scenario for the non-falsy-string check is that every number is 0.
```

src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,89 @@
...
+				$accessories[] = new AccessoryNonFalsyStringType();
+			} elseif ($value !== '') {
+				$accessories[] = new AccessoryNonEmptyStringType();
+			}
+			if (strtolower($value) === $value) {
+				$accessories[] = new AccessoryLowercaseStringType();
+			}
+			if (strtoupper($value) === $value) {
+				$accessories[] = new AccessoryUppercaseStringType();
+			}
```
staabm, line 79: this duplicates all logic on how to make accessories based on a scalar string.

couldn't you use a `ConstantStringType` and `generalize` it instead?

VincentLanglet: I tried but I ended up with AccessoryLiteralString too ; which is obviously not here.

VincentLanglet: I'll try with TypeCombinator::remove

VincentLanglet: So it's not possible ; the issue is that TypeCombinator is not made to remove an Accessory.

ondrejmirtes (changes requested):
Please add regression test for https://phpstan.org/r/66bd52bc-2c1c-4b6e-ac27-57c2a7261982.

I'd like both a test in InvalidBinaryOperationRule, and `assertType` for `%a`.

VincentLanglet:
> Please add regression test for [phpstan.org/r/66bd52bc-2c1c-4b6e-ac27-57c2a7261982](https://phpstan.org/r/66bd52bc-2c1c-4b6e-ac27-57c2a7261982).
> 
> I'd like both a test in InvalidBinaryOperationRule, and `assertType` for `%a`.

This doesn't solve the InvalidBinaryOperationRule issue indeed, I just added a nsrt test to show why.

`%a` can return either a numeric string or `(unknown)`, and we cannot know the DateTimeInterval is coming from a `DateTime::diff` call or was just instantiated with something like `new DateInterval(...)`.

So this PR is still an improvement but doesn't solve the issue.

src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,91 @@
...
+		if (!isset($arguments[0])) {
+			return null;
+		}
+
+		$arg = $scope->getType($arguments[0]->value);
+
+		$constantStrings = $arg->getConstantStrings();
+		if (count($constantStrings) === 0) {
+			if ($arg->isNonEmptyString()->yes()) {
+				return new IntersectionType([new StringType(), new AccessoryNonEmptyStringType()]);
```
staabm, line 52: `TypeCombinator::intersect()` instead?

src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
staabm: could we extract the logic into a helper class?

staabm (changes requested):

src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
VincentLanglet: I'm not against refacto but I would have prefer un-directly-related request change from ondrej because
he doesn't like the helper class/method/naming/etc

Maybe it could be better to merge this and introduce the method in a specific PR ?

Also, for this method to be useful in an helper it has to be re-used (multiple times). Is it really the case ?
Currently it's 
- one time with literalstring
- one time without

And where (in which class) do you want to put such method ?

src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
VincentLanglet: We generally use directly IntersectionType for string type + accessories because it doesn't require any extra manipulation while intersect does lot of different things useless in this case

src/Type/Php/DateIntervalFormatDynamicReturnTypeExtension.php
staabm: thats fine. I can have a look after this PR merged

ondrejmirtes:
Thank you!

PR #4166, opened 2025-07-26
Respect phpstan level in InvalidKeyInArrayDimFetchRule

VincentLanglet:
Extracted from https://github.com/phpstan/phpstan-src/pull/4012

It will be easier to review the CastArrayKeyRule after


ondrejmirtes (changes requested):
If you're doing something to respect levels better, you need to add some cases to LevelsIntegrationTest. You can just add the "topic" php file and run the test, it will generate JSONs with the current result.

It'd actually be great to see the current behaviour in the first commit and then see how things get corrected in the second commit with the code changes.

The expectations are:

* level < 7 - report things that are always wrong
* level 7 - report things that might be wrong, like `int|string` only where `int` is expected
* level 8 - same but for `int|null`
* level 9 - explicit mixed
* level 10 - implicit (untyped) mixed

ondrejmirtes:
The best way to get LevelsIntegrationTest results fast is to comment out all the other "topics" when you're working on a single topic. Otherwise it takes a long time for the tests to run.

Also you could add some upload-artifact action call to `tests-levels` in `tests.yml` so that when PR run fails, we can download the updated JSON files and there's no need to run the test locally and slowly.

VincentLanglet:
Ok, so here you have

- First commit introduce the integration test with the current behavior
https://github.com/phpstan/phpstan-src/pull/4166/commits/5e15e2e3479d63092e50b921382168b6d79d0343

- Second commit is my fix of InvalidKeyInArrayDimFetchRule and the impact
https://github.com/phpstan/phpstan-src/pull/4166/commits/ae06b483841a9e0c93bb019f67786d521ba230ae

- Funny things, the message is different between level 7 and 8, and I found the fix needed was in RuleLevelHelper:
https://github.com/phpstan/phpstan-src/pull/4166/commits/38c316f122cede93b6221aed9298032c6a108910
-> You should not remove `null` if it's an accepted type in the Union (because if you do so, you might transform a "maybe error" into a "100% sure error".

Then last commit was just to fix the lint 7.4.

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -41,22 +40,27 @@ public function processNode(Node $node, Scope $scope): array
...
-
 		$varType = $this->ruleLevelHelper->findTypeToCheck(
 			$scope,
 			$node->var,
 			'',
-			static fn (Type $varType): bool => $varType->isArray()->no() || AllowedArrayKeysTypes::getType()->isSuperTypeOf($dimensionType)->yes(),
+			static fn (Type $varType): bool => $varType->isArray()->no(),
 		)->getType();
 
 		if ($varType instanceof ErrorType || $varType->isArray()->no()) {
```
ondrejmirtes: What feels wrong here is that you're doing separate var and dim checks.

If the var check returns ErrorType (because we're on level 3 and not 7+ for example), you're not even going to run the dim check - which might actually report something wrong on level 3+.

ondrejmirtes (changes requested):

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
VincentLanglet: It's already the case no ? The rule is already reporting nothing about $dimensionType if $varType is an errortype or not an array. And to me it make sens, we shouldn't report an invalid array offset, if the variable is not an array.

varType can be an ErrorType 
- when it's mixed on level 7- ; we shouldn't report anything about the offset since we dunno the var type
- when it's not existing class ; not an array
- when it's an object and union is not checked ; not an array
- I'm didn't found other case in RuleLevelHelper.

Do you have an exemple of case which should be reported on level 3 and is not with the current implement ? Cause all the existing tests are green, and I really think the change is correct as showed by the integration tests AND necessary bugfix for https://github.com/phpstan/phpstan-src/pull/4012

ondrejmirtes: You're right. This rule is only about arrays so the behaviour is correct.

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
```diff
@@ -41,22 +40,27 @@ public function processNode(Node $node, Scope $scope): array
...
-		if ($dimensionType instanceof MixedType) {
-			return [];
-		}
-
 		$varType = $this->ruleLevelHelper->findTypeToCheck(
 			$scope,
 			$node->var,
 			'',
-			static fn (Type $varType): bool => $varType->isArray()->no() || AllowedArrayKeysTypes::getType()->isSuperTypeOf($dimensionType)->yes(),
+			static fn (Type $varType): bool => $varType->isArray()->no(),
```
ondrejmirtes, line 47: I'm still wondering about what's the effect of the `$varType` going through RuleLevelHelper. Does something change if we get rid of it only continue if `$scope->getType($node->var)->isArray()->yes()`?

It's weird to me that this rule checks maybe-arrays. It's okay for `array|null` for example, but I'm not sure if it ever should report something on `mixed`.

ondrejmirtes (changes requested):

src/Rules/Arrays/InvalidKeyInArrayDimFetchRule.php
VincentLanglet: I added more test case here with the commit https://github.com/phpstan/phpstan-src/pull/4166/commits/6f99f04c397e57edd81fcfc811a43aced9eac317

You can see that
- array|foo is checked
- explicit and implicit mixed are not checked (because transformed to ErrorType or to a `StrictMixedType` in which `isArray()` return no.

But it was right to report I need to check the result for maybe array.
The error message was `Invalid array key type DateTimeImmutable.`, I think it should be `Possibly invalid array key type DateTimeImmutable.` then. And I fix that in the commit https://github.com/phpstan/phpstan-src/pull/4166/commits/7e0f73b1e2b4001c649550e64662df6a9a26f5e8

ondrejmirtes:
Thank you!

PR #4167, opened 2025-07-26
Fix non final object cast to array

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/2730

### DIFF
--- a/src/Type/ObjectType.php
+++ b/src/Type/ObjectType.php
@@ -36,6 +36,7 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\Accessory\AccessoryNumericStringType;
+use PHPStan\Type\Accessory\HasOffsetValueType;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantBooleanType;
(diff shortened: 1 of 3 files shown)


src/Type/ObjectType.php
```diff
@@ -693,8 +694,17 @@ public function toArray(): Type
...
-		if (!$isFinal && count($arrayKeys) === 0) {
-			return new ArrayType(new MixedType(), new MixedType());
+		if (!$isFinal) {
+			if (count($arrayKeys) === 0) {
+				return new ArrayType(new MixedType(), new MixedType());
+			}
+
+			$types = [new ArrayType(new MixedType(), new MixedType())];
+			foreach ($arrayKeys as $i => $arrayKey) {
+				$types[] = new HasOffsetValueType($arrayKey, $arrayValues[$i]);
```
ondrejmirtes, line 704: I'd like to have this limited to a few items, to prevent combinatorial explosion. 16 should be fine.

If there's more, just return plain array type.

ondrejmirtes (changes requested):

src/Type/ObjectType.php
VincentLanglet: Done with https://github.com/phpstan/phpstan-src/pull/4167/commits/ddfa73462af569b0d79522005c9942e6cd951295

ondrejmirtes:
Thank you.

PR #4168, opened 2025-07-26
Fix InvalidComparisonOperationRule for UnionType

VincentLanglet:
I'm not sure about the purpose of the InvalidComparisonOperationRule which seams to forbid comparison between number and object or number and arrays. Technically something like `$int == $array` is supported by php (and 0 == [] for instance).
Maybe the message 
```
Comparison operation "!=" between array and 0 results in an error.
```
should be improved, but that's another topic.

Looking at the implementation, both `isPossiblyNullableObjectType` and `isPossiblyNullableArrayType` shouldn't use the RuleLevelHelper, because it will removed non-array/non-object on lower level and this is the reason why an error happen on level 6 (and then disappear in level 7).

Finally, I notice that currently a comparison between `int` and `object|array` is not reported.
This is easy to fix, but since I dunno the whole purpose of this rule, I prefer to ask you first @ondrejmirtes if I should fix it too.

Closes https://github.com/phpstan/phpstan/issues/3364

### DIFF
--- a/src/Rules/Operators/InvalidComparisonOperationRule.php
+++ b/src/Rules/Operators/InvalidComparisonOperationRule.php
@@ -62,7 +62,7 @@ final class InvalidComparisonOperationRule implements Rule
 
 		$isLeftNumberType = $this->isNumberType($scope, $node->left);
 		$isRightNumberType = $this->isNumberType($scope, $node->right);
-		if (($isLeftNumberType && $isRightNumberType) || (!$isLeftNumberType && !$isRightNumberType)) {
+		if ($isLeftNumberType === $isRightNumberType) {
 			return [];
 		}
 
--- a/src/Rules/Operators/InvalidComparisonOperationRule.php
+++ b/src/Rules/Operators/InvalidComparisonOperationRule.php
@@ -12,13 +12,14 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\ShouldNotHappenException;
-use PHPStan\Type\BenevolentUnionType;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\FloatType;
 use PHPStan\Type\IntegerType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\NullType;
 use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\Type;
-use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
 use function get_class;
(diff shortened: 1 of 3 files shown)


src/Rules/Operators/InvalidComparisonOperationRule.php
```diff
@@ -59,7 +60,9 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
-		if ($this->isNumberType($scope, $node->left) && $this->isNumberType($scope, $node->right)) {
+		$isLeftNumberType = $this->isNumberType($scope, $node->left);
+		$isRightNumberType = $this->isNumberType($scope, $node->right);
+		if (($isLeftNumberType && $isRightNumberType) || (!$isLeftNumberType && !$isRightNumberType)) {
```
ondrejmirtes: What about `xor` instead? :D

ondrejmirtes (changes requested):

src/Rules/Operators/InvalidComparisonOperationRule.php
VincentLanglet: I never used it, i'll try.


Offtopic @ondrejmirtes I saw you merged/review multiple of my PRs. I dunno if you plan a release. I just wanted to inform you I dont have access to a computer until Wednesday. So if you want to include one of my pr (with requested changes) feel free to push on it and if any of these PR introduce a regression I wont be able to fix it before Wednesday.

It's maybe not a big issue for you, just wanted to inform you

ondrejmirtes: It's fine, thank you for letting me know. I don't think there will be a major issue, PHPStan is continuously tested on several real-world projects (millions of LoC) so I know about the real issues pretty soon.

If something still comes up that a lot of people have problem with, they can wait on a one-before-last version before they upgrade.

VincentLanglet: Simplification was easy finally 
https://github.com/phpstan/phpstan-src/pull/4168/commits/e14567b1ca42c83ed1f30c32d43be471238a79cb

ondrejmirtes:
Thank you!

staabm:
@VincentLanglet please see issue bot results for related regression tests:
https://github.com/phpstan/phpstan-src/actions/runs/17613710896

VincentLanglet:
> @VincentLanglet please see issue bot results for related regression tests: [phpstan/phpstan-src/actions/runs/17613710896](https://github.com/phpstan/phpstan-src/actions/runs/17613710896)

Yeah, I sometime forget...
Would been great to have the bot posting on the PR ^^' 

Do you think it closes https://github.com/phpstan/phpstan/issues/9386, it's unclear to me...

ondrejmirtes:
@VincentLanglet Unfortunately I don't know how to do it in a secure way.

ondrejmirtes:
Looks like it does close https://github.com/phpstan/phpstan/issues/9386.

staabm:
> Yeah, I sometime forget...

me too. no worries.

staabm:
> Unfortunately I don't know how to do it in a secure way.

@ondrejmirtes I guess you fear someone could steal the bot token or similar things?

maybe the issue-bot could just emit a pull request annotation with a link to the job-summary in case changes are present. 
that way we would see that results exist.

this can be done just with a "echo", and without any further api tokens

PR #4169, opened 2025-07-27
Ignore PHPUnit cache-file

staabm:
when running phpunit with `--cache-result`, e.g.

`vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php --cache-result --order-by=defects,random --stop-on-failure`

it creates a result-cache file.

---

lets make sure this file is ignored and not mistakenly committed into the codebase

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@
 /tests/PHPStan/Reflection/data/golden/
 tmp/.memory_limit
 e2e/bashunit
-.phpunit.cache/test-results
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Is there downside to always generate the cache? And put it under tmp/ that's already ignored?

staabm:
no. I assumed there is a reason why phpunit.xml was not configured to use the tmp/ dir.

after configuring it for tmp/ we don't need the additional ignore

PR #4170, opened 2025-07-27
fix unnecessary wrapping in TableErrorFormatter

schlndh:
Fixes https://github.com/phpstan/phpstan/issues/13317

The change was submitted to symfony in https://github.com/symfony/symfony/pull/61268.

This time I tested the changes with a large project that I have available (~15k PHPStan errors) in two terminal sizes (94 and 382). And I did found one more regressions (first one is with 2.1.18 and the second one is with this PR, don't mind the :pencil2: - I'm using spaces for the editor link):

```
  87     Access to an undefined property object::$id.                                          
         🪪  property.notFound                                                                 
         💡  Learn more:                                                                       
            https://phpstan.org/blog/solving-phpstan-access-to-undefined-property              
         ✏️                  


  87     Access to an undefined property object::$id.                                         
         🪪  property.notFound                                                                
         💡  Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-propert  
         y                                                                                    
         ✏️     
```

As you can see the link is broken (i.e. when I click it it goes to the clipped URL) at this particular size (as well as any smaller size). Removing `<fg=cyan>` from it would fix the issue. Alternatively, it could be shortened. What do you think? I'm not particularly eager to continue fixing more symfony/console issues until the previous PRs are accepted. :sweat_smile:

### DIFF
--- a/patches/OutputFormatter.patch
+++ b/patches/OutputFormatter.patch
@@ -31,23 +31,37 @@
 
          return strtr($output, ["\0" => '\\', '\\<' => '<', '\\>' => '>']);
      }
-@@ -253,8 +256,8 @@ class OutputFormatter implements WrappableOutputFormatterInterface
+@@ -253,8 +256,20 @@ class OutputFormatter implements WrappableOutputFormatterInterface
          }
 
          if ($currentLineLength) {
 -            $prefix = substr($text, 0, $i = $width - $currentLineLength)."\n";
 -            $text = substr($text, $i);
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you.

Please continue sending fixes for bigs that currently affect PHPStan. And queue them as Symfony PRs. So that when we're able to update Symfony Console, we'll be able to remove all patches.

PR #4172, opened 2025-07-27
Add non regression test for 3474

VincentLanglet:
Seems like it's correctly normalized since both single quote and double quote have the correct type.

Closes https://github.com/phpstan/phpstan/issues/3474


ondrejmirtes:
Thank you.

PR #4173, opened 2025-07-27
Add non regression test for 3506

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/3506

The issue doesn't exists anymore 
https://phpstan.org/r/8f0535db-38ce-4ba3-830f-39b6b19e7884


ondrejmirtes:
Thank you!

PR #4174, opened 2025-07-27
Add non regression test for 3589

VincentLanglet:
The error is reported on level 9 (which checks mixed)
https://phpstan.org/r/66416d1c-881a-4511-a108-898a27993605

Closes https://github.com/phpstan/phpstan/issues/3589


ondrejmirtes:
Thank you.

PR #4175, opened 2025-07-27
Add non regression test for 3616

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/3616

There is no error on https://phpstan.org/r/cb3fafd4-7c39-47c9-ac6a-972d110b47d5
with and without the phpdoc `/** @var array<string, callable():void> */`.
- https://phpstan.org/r/7db4d2c7-2a51-4ee2-a553-89ad2b357397
- https://phpstan.org/r/34eb20b5-0605-4427-b80c-c84dc880ad09

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-3616.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-3616.php
@@ -1,0 +1,31 @@
+<?php declare(strict_types = 1);
+
+namespace Bug3616;
+
+class Factory {
+	public static function a(): void { echo 'A'; }
+	public static function b(): void { echo 'B'; }
+}
+
+class HelloWorld
+{
+	/** @var array<string, callable():void> */
+	const FACTORIES = [
+		'a' => [Factory::class, 'a'],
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Rules/Functions/data/bug-3616.php
```diff
@@ -0,0 +1,31 @@
...
+namespace Bug3616;
+
+class Factory {
+	public static function a(): void { echo 'A'; }
+	public static function b(): void { echo 'B'; }
+}
+
+class HelloWorld
+{
+	/** @var array<string, callable():void> */
```
ondrejmirtes, line 12: Please add HelloWorld2 without the PHPDoc.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Functions/data/bug-3616.php
VincentLanglet: Done

ondrejmirtes:
Thank you.

PR #4176, opened 2025-07-27
Add non regression test for 3761

VincentLanglet:
Rules about unreachable branch was removed in
https://github.com/phpstan/phpstan-src/compare/7d74a638011774d9e59aa1f9472c39c00182365c...cc7ff8b217021402dcbe817f7002b29a48722811

The error doesn't exists anymore 
https://phpstan.org/r/788cdf82-69e5-450b-8851-f475793aadcc

I added the test regression to ensure we won't get the following error
https://phpstan.org/r/b888b1ec-063b-49dc-8db8-daa1a0495200

Closes https://github.com/phpstan/phpstan/issues/3761

### DIFF
--- a/tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
+++ b/tests/PHPStan/Rules/Comparison/StrictComparisonOfDifferentTypesRuleTest.php
@@ -1001,6 +1001,11 @@ class StrictComparisonOfDifferentTypesRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/bug-10884.php'], []);
 	}
 
+	public function testBug3761(): void
+	{
+		$this->analyse([__DIR__ . '/data/bug-3761.php'], []);
+	}
+
 	public function testBug13208(): void
 	{
 		$this->analyse([__DIR__ . '/data/bug-13208.php'], []);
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Please reference the issue with "Closes ..." in the PR description.

ondrejmirtes:
Also I don't really get it - playground still reports the error https://phpstan.org/r/b888b1ec-063b-49dc-8db8-daa1a0495200, how could you add a regression test if it still needs fixing?

VincentLanglet:
> Also I don't really get it - playground still reports the error [phpstan.org/r/b888b1ec-063b-49dc-8db8-daa1a0495200](https://phpstan.org/r/b888b1ec-063b-49dc-8db8-daa1a0495200), how could you add a regression test if it still needs fixing?

I edited to change
```
if($tag instanceof $class){
```
into 
```
if($tag instanceof NamedTag){
```
to show what would happen with the previous understanding of PHPStan. (and what message we would have got, and therefor in which testCase i have to put the test)

VincentLanglet:
> Please reference the issue with "Closes ..." in the PR description.

Sorry, I forgot, I added the https://github.com/phpstan/phpstan/issues/3761 reference.

ondrejmirtes:
Thank you.

PR #4177, opened 2025-07-27
Add non regression test for 4284

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/4284

https://phpstan.org/r/dc24e5cb-ca3c-45f5-a49e-7425604e7ac9 is understood.


ondrejmirtes:
I'd expect the opposite. If `$browserName` is truthy then `$upgradeBrowserLink` is truthy as well. PHPStan should detect that in both conditions.

PR #4178, opened 2025-07-27
Fix version_compare return type

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/4457

Also avoid the following error
https://phpstan.org/r/e5138f79-6672-4a17-9cd2-531791bc016d

### DIFF
--- a/src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
@@ -10,21 +10,41 @@ use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Php\ComposerPhpVersionFactory;
 use PHPStan\Php\PhpVersion;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\BooleanType;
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Thank you.

src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
```diff
@@ -81,11 +104,21 @@ public function getTypeFromFunctionCall(
...
+						$operatorValue = $operatorString->getValue();
+						if (!in_array($operatorValue, self::VALID_OPERATORS, true)) {
+							if (!$this->phpVersion->throwsValueErrorForInternalFunctions()) {
+								$canBeNull = true;
+							}
+
+							continue;
+						}
+
+						$value = version_compare($version1String->getValue(), $version2String->getValue(), $operatorValue);
```
ondrejmirtes, line 121: Hey, here's `bool` and then it's used in `$types[$value]` which gets picked up with `reportNonIntStringArrayKey: true`. Can you please fix it? I' ignoring this issue in 2.1.x.

VincentLanglet: Sure: https://github.com/phpstan/phpstan-src/pull/4871

PR #4181, opened 2025-07-28
Add non regression test for 4809

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/4809

Cf https://phpstan.org/r/dd639e0b-0c5d-45f1-a22e-7727f04cf471


ondrejmirtes:
Thank you.

PR #4182, opened 2025-07-28
Add non regression test for 4993

VincentLanglet:
So the issue https://github.com/phpstan/phpstan/issues/4993 is unclear to me
- It ask for no error on https://phpstan.org/r/5d0dc8ed-59c6-4b31-90a8-9c8bb85705fa but
  - There is already no error on PHP < 8
  - On PHP 8+ the error about NULL is legit since it's not a possible value. cf (https://github.com/phpstan/phpstan-src/pull/3712)
  
Also the issue talk about the fact the returned array is supposed to be non-empty ; which seems true and supported by psalm https://psalm.dev/r/3176073f08 so I updated the definition.

So I think that
Closes https://github.com/phpstan/phpstan/issues/4993


ondrejmirtes:
Thank you.

PR #4183, opened 2025-07-28
Add non regression test for 5206

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/5206


ondrejmirtes:
Thank you.

PR #4184, opened 2025-07-28
Fix 13296

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13296

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2347,7 +2347,11 @@ final class MutatingScope implements Scope
 					/** @var non-empty-string $name */
 					$name = $node->name->value;
 					$functionName = new Name($name);
-				} elseif ($node->name instanceof FuncCall && $node->name->name instanceof Name) {
+				} elseif (
+					$node->name instanceof FuncCall
+					&& $node->name->name instanceof Name
+					&& $node->name->isFirstClassCallable()
+				) {
 					$functionName = $node->name->name;
 				}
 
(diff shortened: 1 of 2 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -2347,7 +2347,11 @@ static function (Node $node, Scope $scope) use ($arrowScope, &$arrowFunctionImpu
 					/** @var non-empty-string $name */
 					$name = $node->name->value;
 					$functionName = new Name($name);
-				} elseif ($node->name instanceof FuncCall && $node->name->name instanceof Name) {
+				} elseif (
+					$node->name instanceof FuncCall
+					&& $node->name->name instanceof Name
+					&& $node->name->isFirstClassCallable()
```
VincentLanglet, line 2353: Seems like the purpose of this if was to resolve node like
`strtoupper(...)($foo)` 
which is resolved to `strtoupper($foo)`
but it messed up with `foo($bar)($baz)`
which was resolved to `foo($baz)`.

tests/PHPStan/Analyser/nsrt/bug-13296.php
```diff
@@ -0,0 +1,26 @@
...
+
+class Foo
+{
+	/**
+	 * @param non-empty-array<Bar> $a
+	 * @param non-empty-array<int> $b
+	 */
+	public function test($a, $b)
+	{
+		assertType('string', current($a)($b));
```
VincentLanglet, line 15: Without the fix, the type is understood as `current($b)` which is int.

AJenbo:
Thanks

ondrejmirtes:
Please use more descriptive PR titles than "Fix 13296" next time.

ondrejmirtes:
Thank you!

PR #4185, opened 2025-07-28
Remove couchbase from functionmap

VincentLanglet:
This would
Closes https://github.com/phpstan/phpstan/issues/5642

Maybe I didn't fully understood the gain of having those method in the functionMap but
- the current functionMap documents the outdated couchbase version 2.
- phpstorm-stubs correctly documents the couchbase version 3.
- there is a lot of difference between the two version (class/function  removed, class/function added, lot of signature changed)

As a proof I get the correct error in my non regression test.


ondrejmirtes:
Thank you!

This might be an idea for a project: remove huge chunks of functionMap for things that are already nicely covered by phpstorm-stubs.

We can't remove signatures that use special PHPStan types like non-empty-string or benevolent union, but I still think we can make it much smaller. Feel free to send PRs that remove one extension at a time.

PR #4186, opened 2025-07-28
feat: dirty files check

Edgaraszs:
This is a partially completed feature that analyzes files with uncommitted changes.

Can this possibly be added to library? If yes I could continue to finish this just need to make sure before continuing.

Thanks for any feedback :)


ondrejmirtes:
This doesn't make sense and is opposite of what you should do. See https://phpstan.org/blog/why-you-should-always-analyse-whole-project.

PHPStan is fast by default. Read the article to learn why.

Edgaraszs:
> This doesn't make sense and is opposite of what you should do. See https://phpstan.org/blog/why-you-should-always-analyse-whole-project.
> 
> PHPStan is fast by default. Read the article to learn why.

Okay, got it. Thanks for your feedback.

PR #4189, opened 2025-07-29
Add non regression test for 13267

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13267

I fixed in on phpstorm-stubs.


ondrejmirtes (changes requested):
Are you sure this reported an error in CallMethodsRule?

VincentLanglet:
Indeed it was CallStaticMethodsRule (I tried to revert the phpstorm stub to reproduce the error).

Fixed (Linting error is unrelated)

ondrejmirtes:
Thank you.

PR #4190, opened 2025-07-29
Exclude first-party stubs from validation

blyxxyz:
Stub files from vendor directories aren't validated. But the stub files that ship with PHPstan itself weren't excluded, and they were hitting the `checkMissingOverrideMethodAttribute` check.

This didn't happen when PHPstan was installed through composer, since then the stub files happened to be in a vendor directory. But it did happen when PHPstan ran as a phar.

Resolves https://github.com/phpstan/phpstan/issues/11907.


ondrejmirtes:
Thank you!

PR #4192, opened 2025-07-29
Allow 8.5 as PHP version no. 2

ondrejmirtes:


Flyingmana:
The currently failing tests in PHP 7.4 regarding PHPStan\Reflection\ClassReflectionTest::testIsAttributeClass and the changed \Attribute::TARGET_ALL value might be related to this PR here.

will probably be resolved with an update of the Polyfill after this is resolved and released: https://github.com/symfony/polyfill/issues/537

PR #4194, opened 2025-08-01
Fix session_set_cookie_params signature

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/7772

From the changelog of https://www.php.net/manual/en/function.session-set-cookie-params.php
```
8.0.0 | path, domain, secure and httponly are nullable now.
```


ondrejmirtes:
Thank you.

PR #4195, opened 2025-08-01
Fix RedisCluster::scan signature

DKhalil:
Updating the RedisCluster::scan signature to match the documented one:

https://phpredis.github.io/phpredis/RedisCluster.html#method_scan

Broken test:

https://phpstan.org/r/99997323-9bb9-43c7-b1ba-9d32236b2d68


ondrejmirtes:
Thank you.

PR #4197, opened 2025-08-01
Remove DomNode::ownerDocument stub

VincentLanglet:
1) This seems to still be nullable:
For instance https://3v4l.org/mMQj8#v8.4.11

2) According to https://github.com/phpstan/phpstan/issues/8837 this wasn't working anyway.

You also might want to close the issue https://github.com/phpstan/phpstan/issues/8837 since it asked to infer `DomElement::ownerDocument` as non nullable while the 3v4l example is showing it can be null.


ondrejmirtes:
Please fix conflicts.

ondrejmirtes:
Thank you.

PR #4198, opened 2025-08-02
Escape tag in githubErrorFormatter

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10051

### DIFF
--- a/tests/PHPStan/Command/ErrorFormatter/RawErrorFormatterTest.php
+++ b/tests/PHPStan/Command/ErrorFormatter/RawErrorFormatterTest.php
@@ -84,7 +84,7 @@ Bar2
 		yield [
 			'message' => 'One file error with tip',
 			'exitCode' => 1,
-			'numFileErrors' => [5, 6],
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Command/ErrorFormatter/RawErrorFormatterTest.php
```diff
@@ -84,7 +84,7 @@ public static function dataFormatterOutputProvider(): iterable
 		yield [
 			'message' => 'One file error with tip',
 			'exitCode' => 1,
-			'numFileErrors' => [5, 6],
```
VincentLanglet, line 87: This was wrong, this is supposed to be
- The offset (5)
- The length (1), not 6

This is not a list of the offset to take.

ondrejmirtes:
Please do the same in issue-bot/ that posts comments. Assuming we can also do this in plain Markdown, not just in GitHub Actions.

ondrejmirtes:
Thank you.

PR #4200, opened 2025-08-03
Improve TypeCombinator::unionWithSubtractedType

VincentLanglet:
While thinking about https://github.com/phpstan/phpstan/issues/12390 I just decovered the weird behavior
https://phpstan.org/r/0850c2cb-3c81-4451-b2f6-b570d1d41a30

I expect `mixed~object~stdclass & object` to be `stdClass`.

### DIFF
--- a/src/Type/Traits/SubstractableTypeTrait.php
+++ b/src/Type/Traits/SubstractableTypeTrait.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Traits;
+
+use PHPStan\Type\SubtractableType;
+use PHPStan\Type\Type;
+use PHPStan\Type\UnionType;
+use PHPStan\Type\VerbosityLevel;
+use function sprintf;
+
+trait SubstractableTypeTrait
+{
+
+	public function describeSubtractedType(?Type $subtractedType, VerbosityLevel $level): string
+	{
+		if ($subtractedType === null) {
+			return '';
+		}
+
+		if (
+			$subtractedType instanceof UnionType
+			|| ($subtractedType instanceof SubtractableType && $subtractedType->getSubtractedType() !== null)
+		) {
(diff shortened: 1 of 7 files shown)


tests/PHPStan/Analyser/nsrt/subtracted.php
```diff
@@ -0,0 +1,31 @@
...
+	{
+		if(is_object($date)){
+
+		} else {
+			assertType('mixed~object', $date);
+
+			if ($foo) {
+				$date = new \stdClass();
+			}
+			assertType('mixed~object~stdClass', $date);
```
ondrejmirtes: I don't know what this means. There should be parentheses to disambiguate.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/subtracted.php
VincentLanglet: This is already the current behavior https://phpstan.org/r/0850c2cb-3c81-4451-b2f6-b570d1d41a30

I'll see what I can do

src/Type/Traits/SubstractableTypeTrait.php
```diff
@@ -0,0 +1,30 @@
...
+use PHPStan\Type\SubtractableType;
+use PHPStan\Type\Type;
+use PHPStan\Type\UnionType;
+use PHPStan\Type\VerbosityLevel;
+use function sprintf;
+
+trait SubstractableTypeTrait
+{
+
+	public function describeSubtractedType(?Type $subtractedType, VerbosityLevel $level): string
```
VincentLanglet, line 14: I dunno if it's worth to add this method to the interface (I didn't) but since the logic was used 5 times I thought it could been useful to have a specific method (for later updates ?)

ondrejmirtes:
Thank you.

PR #4201, opened 2025-08-03
Fix array_column inference

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10367

Fix is the first commit https://github.com/phpstan/phpstan-src/pull/4201/commits/5e252bd06b7acf3c37ae7cefa5f2a1e2122ba739

Second commit is just a refacto to reduce the number of method calls https://github.com/phpstan/phpstan-src/pull/4201/commits/26476e034f688b32be8b73db7dcc044b195effd3

Since the method is almost always doing logic like
```
$type = $this->getOffsetOrProperty(...., false);
if ($type !== null) {
     return type;
} else {
     $type = $this->getOffsetOrProperty(...., true);
     // do things
}
```
It's better to directly return the type and the certainty and write
```
[$type, $certainty] = $this->getOffsetOrProperty(...., false);
if ($certainty->yes()) {
     return type;
} else {
     // do things
}
```

### DIFF
--- a/src/Type/Php/ArrayColumnHelper.php
+++ b/src/Type/Php/ArrayColumnHelper.php
@@ -39,14 +39,10 @@ final class ArrayColumnHelper
 		}
 
 		$iterableValueType = $arrayType->getIterableValueType();
-		$returnValueType = $this->getOffsetOrProperty($iterableValueType, $columnType, $scope, false);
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you.

PR #4205, opened 2025-08-04
Move isOffsetAccessLegal to MaybeOffsetAccessibleTypeTrait

VincentLanglet:
Is there a reason for not having this offset-related method in the MaybeOffsetAccessibleTypeTrait when all the implementation returns maybe (?)


ondrejmirtes:
I'm fine with this, although I'm not a fan of these traits at all.

I can tolerate them, but I'd really like some custom rules around them (like that if a class is using the trait, it has to use the whole trait and not override any of its methods).

ondrejmirtes:
Thank you.

PR #4206, opened 2025-08-05
Failng test: Trait error ignored with used class path fails with partial analysis

janedbal:
- The support for this ignore was added in https://github.com/phpstan/phpstan-src/commit/14f08c18dbfdd88105e1cafef64336961574a5a7
- Baseline generation uses class path


ondrejmirtes:
This is in line with other "not matched in reported errors": https://github.com/phpstan/phpstan-src/commit/cd6b3fd9717279d07c9a063ced10465121eab585

janedbal:
Thank you!

PR #4207, opened 2025-08-05
fix: place UniversalObjectCratesClassReflectionExtension last

calebdw:
Hello!

This moves the UniversalObjectCratesClassReflectionExtension to the very end of the extensions to give other extensions a chance to override.

Closes https://github.com/phpstan/phpstan/issues/13342

Thanks!

### DIFF
--- a/conf/services.neon
+++ b/conf/services.neon
@@ -69,6 +69,11 @@ services:
 		class: PHPStan\Reflection\Annotations\AnnotationsPropertiesClassReflectionExtension
 
 	-
+		class: PHPStan\Reflection\Php\UniversalObjectCratesClassReflectionExtension
+		arguments:
+			classes: %universalObjectCratesClasses%
+
+	-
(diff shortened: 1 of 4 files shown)


calebdw:
None of the failed jobs look like they came from these changes? Let me know if I need to change anything though!

staabm:
Fix the lint job by `make cs-fix`

calebdw:
@staabm, I can do that, but that's from a file upstream that wasn't touched in this PR

ondrejmirtes:
The changes are fine, I just don't understand why this has to be E2E test. It could be just a test for a specific rule. The config doesn't even add anything, stdClass is already universal object crate.

calebdw:
Ah, I just wasn't sure where to put the test, is there a specific location you would like it to go?

ondrejmirtes:
Looking at the bug report simply putting the file in tests/PHPStan/Analyser/nsrt should be fine

calebdw:
Awesome, I've updated

ondrejmirtes:
Thank you.

PR #4208, opened 2025-08-05
Fix lint

VincentLanglet:
Just in case it wasn't fixed by another PR.

This will avoid everyone to fix it on their PR.


ondrejmirtes:
Thank you!

PR #4209, opened 2025-08-05
do not skip Generator for MissingTypehintCheck

Flyingmana:
references [#7498](https://github.com/phpstan/phpstan/issues/7498)

only a draft for now, I want to see which test are all failing and for what reasons.

### DIFF
--- a/src/Rules/MissingTypehintCheck.php
+++ b/src/Rules/MissingTypehintCheck.php
@@ -43,7 +43,7 @@ final class MissingTypehintCheck
 		Traversable::class,
 		Iterator::class,
 		IteratorAggregate::class,
-		Generator::class,
+		//Generator::class,
 	];
 
 	/**
--- a/src/Rules/MissingTypehintCheck.php
+++ b/src/Rules/MissingTypehintCheck.php
@@ -43,7 +43,7 @@ final class MissingTypehintCheck
 		Traversable::class,
 		Iterator::class,
 		IteratorAggregate::class,
-		//Generator::class,
+		Generator::class,
 	];
 
 	/**
@@ -54,6 +54,8 @@ final class MissingTypehintCheck
 		private bool $checkMissingCallableSignature,
 		#[AutowiredParameter(ref: '%featureToggles.skipCheckGenericClasses%')]
 		private array $skipCheckGenericClasses,
+		#[AutowiredParameter(ref: '%featureToggles.checkGenericIterableClasses%')]
+		private bool $checkGenericIterableClasses,
 	)
 	{
 	}
@@ -118,7 +120,10 @@ final class MissingTypehintCheck
 				if ($classReflection === null) {
 					return $type;
 				}
-				if (in_array($classReflection->getName(), self::ITERABLE_GENERIC_CLASS_NAMES, true)) {
+				if (
+					$this->checkGenericIterableClasses !== true &&
+					in_array($classReflection->getName(), self::ITERABLE_GENERIC_CLASS_NAMES, true)
+				) {
 					// checked by getIterableTypesWithMissingValueTypehint() already
 					return $type;
 				}
--- a/src/Rules/MissingTypehintCheck.php
+++ b/src/Rules/MissingTypehintCheck.php
@@ -121,7 +121,7 @@ final class MissingTypehintCheck
 					return $type;
 				}
 				if (
-					$this->checkGenericIterableClasses !== true &&
+					!$this->checkGenericIterableClasses &&
 					in_array($classReflection->getName(), self::ITERABLE_GENERIC_CLASS_NAMES, true)
 				) {
 					// checked by getIterableTypesWithMissingValueTypehint() already
--- a/src/Rules/MissingTypehintCheck.php
+++ b/src/Rules/MissingTypehintCheck.php
@@ -121,8 +121,11 @@ final class MissingTypehintCheck
 					return $type;
 				}
 				if (
-					!$this->checkGenericIterableClasses &&
-					in_array($classReflection->getName(), self::ITERABLE_GENERIC_CLASS_NAMES, true)
+					$classReflection->getName() === Traversable::class // already covered by getIterableTypesWithMissingValueTypehint
+					|| (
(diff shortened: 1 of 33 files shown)


ondrejmirtes:
Explained the wanted changes here https://github.com/phpstan/phpstan/issues/7498#issuecomment-3239381890

VincentLanglet:
I would love to this moving forward. Are you still interested in your PR @Flyingmana ?

I think you just need to
- Rebase/Fix conflicts
- Describe the iterable type in some stubs/tests

Flyingmana:
Yes, might have time for it later today, but latest next weekend

tests/PHPStan/Analyser/nsrt/bug-8886.php
```diff
@@ -9,5 +9,5 @@ function testPDOStatementGetIterator(): void {
 	$pdo = new PDO('sqlite::memory:');
     $stmt = $pdo->query('SELECT 1');
 
-	assertType('Iterator', $stmt->getIterator());
+	assertType('Iterator<mixed, array<int|string, mixed>>', $stmt->getIterator());
```
Flyingmana, line 12: not sure if there is an alternative method for the type check, but having a more explicit check is probably ok (I hope)

Flyingmana:
I think now all is fixed which belongs to this repository, the rest is separate repositories

src/Rules/MissingTypehintCheck.php
```diff
@@ -118,7 +120,10 @@ public function getNonGenericObjectTypesWithGenericClass(Type $type): array
 				if ($classReflection === null) {
 					return $type;
 				}
-				if (in_array($classReflection->getName(), self::ITERABLE_GENERIC_CLASS_NAMES, true)) {
+				if (
+					$this->checkGenericIterableClasses !== true &&
```
ondrejmirtes: Please do `!$this->checkGenericIterableClasses` instead

tests/PHPStan/Rules/Functions/MissingFunctionParameterTypehintRuleTest.php
```diff
@@ -71,6 +71,11 @@ public function testRule(): void
 				148,
 				MissingTypehintCheck::MISSING_ITERABLE_VALUE_TYPE_TIP,
 			],
+			[
+				'Function MissingFunctionParameterTypehint\missingTraversableTypehint() has parameter $traversable with generic interface Traversable but does not specify its types: TKey, TValu',
```
ondrejmirtes: Pretty sure this will fail with a typo - TValu vs. TValue

ondrejmirtes (changes requested):
I'd like to see more tests with the other no-longer-skipped types like Generator.

Flyingmana:
> I'd like to see more tests with the other no-longer-skipped types like Generator.

I added new tests/examples for generics in 
tests/PHPStan/Rules/Functions/data/missing-function-return-typehint.php

stubs/runtime/ReflectionAttribute.php
```diff
@@ -22,10 +22,16 @@ public function isRepeated(): bool
 		{
 		}
 
+		/**
+		 * @return array<array-key, mixed>;
```
ondrejmirtes: Why the semicolon?

stubs/runtime/ReflectionAttribute.php
```diff
@@ -22,10 +22,16 @@ public function isRepeated(): bool
...
+		 */
 		public function getArguments(): array
 		{
 		}
 
+		/**
+		 * @return self
```
ondrejmirtes: What? That does not make sense.

ondrejmirtes (changes requested):
I may have forgotten some context here but why is this necessary when for example `Traversable` return type already gets `return type has no value type specified in iterable type Traversable`? https://phpstan.org/r/53c35e3b-1a18-40c9-9ca1-6199bdfa1ed5

Flyingmana:
> I may have forgotten some context here but why is this necessary when for example `Traversable` return type already gets `return type has no value type specified in iterable type Traversable`? https://phpstan.org/r/53c35e3b-1a18-40c9-9ca1-6199bdfa1ed5

might have slipped my mind that it was already covered properly when I worked on it the last time. Will remove this part again. (and fix the conflict, check the tests again)

stubs/runtime/ReflectionAttribute.php
Flyingmana: fixed/removed

stubs/runtime/ReflectionAttribute.php
Flyingmana: no idea anymore, but you are correct, that does not make sense, have removed it again

Flyingmana:
another try, Hope its now good :) 
it seems the failing tests are all unrelated.
If more is to do, I currently have more time available in the evenings

ondrejmirtes (changes requested):
Thank you, this makes more sense to me now. Can you please add tests for a pre-existing `getIterableTypesWithMissingValueTypehint` function (for example in MissingMethodReturnTypehintRule) that would check all ITERABLE_GENERIC_CLASS_NAMES to make sure that each of them is either covered by getIterableTypesWithMissingValueTypehint or by the code you've written here?

I'm a little bit worried that the current differentiation for Traversable vs. the others is brittle so I want to be sure it doesn't break in the future. The method `getIterableTypesWithMissingValueTypehint` does not specifically mention `Traversable`, it rests only on `$type->isIterable()->yes()` + `$iterableValue instanceof MixedType && !$iterableValue->isExplicitMixed()`.

Flyingmana:
like the tests I added in `tests/PHPStan/Rules/Functions/MissingFunctionReturnTypehintRuleTest.php`? or a more plain phpunit test? (then an example would be helpfull, as I can not find a test which calls processNode() currently)

ondrejmirtes:
Yes, it's okay. Thank you!

ondrejmirtes (approved):

ondrejmirtes:
Please send PRs to the failing extension repositories. Thanks!

PR #4210, opened 2025-08-06
POC: Use SimultaneousTypeTraverser in RuleLevelHelper::transformAcceptedType

VincentLanglet:
Static analysis is failing because TypeTraverserInstanceofVisitor doesn't support SimultaneousTypeTraverser (not sure it should).

The only current test failing is
```
16: Function ReturnListNullables\doFoo() should return array<string>|null but returns list<string|null>.
```
but I feel like we could reproduce more errors because...

UnionType/Intersection types doesn't work well with the SimultaneousTypeTraverser because 
- when it's called with `A&B&C` and `A` we're not traversing because second type is not an Intersection
- when it's called with `A&B&C` and `A&C` we're not traversing because types count are different

The main issue would be, when traversing `A&B&C` and `A&C`, to match types in order to
- traverse (A, A)
- traverse (B, Never)
- traverse (C, C)

### DIFF
--- a/src/Parser/TypeTraverserInstanceofVisitor.php
+++ b/src/Parser/TypeTraverserInstanceofVisitor.php
@@ -13,6 +13,11 @@ final class TypeTraverserInstanceofVisitor extends NodeVisitorAbstract
 
 	public const ATTRIBUTE_NAME = 'insideTypeTraverserMap';
 
+	private const TYPE_TRAVERSER_CLASSES = [
+		'phpstan\\type\\typetraverser',
+		'phpstan\\type\\simultaneoustypetraverser',
+	];
+
 	private int $depth = 0;
 
 	#[Override]
@@ -33,7 +38,7 @@ final class TypeTraverserInstanceofVisitor extends NodeVisitorAbstract
 		if (
 			$node instanceof Node\Expr\StaticCall
 			&& $node->class instanceof Node\Name
-			&& $node->class->toLowerString() === 'phpstan\\type\\typetraverser'
+			&& \in_array($node->class->toLowerString(), self::TYPE_TRAVERSER_CLASSES, true)
 			&& $node->name instanceof Node\Identifier
 			&& $node->name->toLowerString() === 'map'
 		) {
@@ -49,7 +54,7 @@ final class TypeTraverserInstanceofVisitor extends NodeVisitorAbstract
 		if (
 			$node instanceof Node\Expr\StaticCall
 			&& $node->class instanceof Node\Name
-			&& $node->class->toLowerString() === 'phpstan\\type\\typetraverser'
+			&& \in_array($node->class->toLowerString(), self::TYPE_TRAVERSER_CLASSES, true)
 			&& $node->name instanceof Node\Identifier
 			&& $node->name->toLowerString() === 'map'
 		) {
--- a/src/Rules/RuleLevelHelper.php
+++ b/src/Rules/RuleLevelHelper.php
@@ -6,6 +6,7 @@ use PhpParser\Node\Expr;
 use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\AutowiredParameter;
 use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Reflection\ParametersAcceptor;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\CallableType;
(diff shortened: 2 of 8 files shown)


ondrejmirtes:
The output from issue-bot is yummy here! https://github.com/phpstan/phpstan-src/actions/runs/16773630245

Please add regression tests for those.

I feel like we could improve `IntersectionType::traverseSimultaneously()` for all the common pseudotypes like `non-empty-string`, `list` etc. represented with IntersectionType.

I feel that even when there's `non-empty-string` on left side and `string` on the right side and vice versa, the callback should still be invoked for them.

This needs extensive tests of course.

ondrejmirtes:
Linking https://github.com/phpstan/phpstan/issues/9096

VincentLanglet:
> The output from issue-bot is yummy here! [phpstan/phpstan-src/actions/runs/16773630245](https://github.com/phpstan/phpstan-src/actions/runs/16773630245)
> 
> Please add regression tests for those.

Sure, but shouldn't we fix the actual tests first ?

> I feel like we could improve `IntersectionType::traverseSimultaneously()` for all the common pseudotypes like `non-empty-string`, `list` etc. represented with IntersectionType.
> 
> I feel that even when there's `non-empty-string` on left side and `string` on the right side and vice versa, the callback should still be invoked for them.

Yeah, but I dunno how to improve the implementation of the existing traverseSimultaneously, it's not easy :/

VincentLanglet:
I feel like https://github.com/phpstan/phpstan-src/commit/6597ef616bbb4de57f5946ed02a45a3f82e712c5#r167671380 is helping us.

And maybe same should be done for IntersectionType

ondrejmirtes:
Yeah, I'm thinking about how to implement it.

ondrejmirtes:
Right now I have this failing test:

```diff
diff --git a/tests/PHPStan/Type/SimultaneousTypeTraverserTest.php b/tests/PHPStan/Type/SimultaneousTypeTraverserTest.php
index 76dc038873..750a7a56bd 100644
--- a/tests/PHPStan/Type/SimultaneousTypeTraverserTest.php
+++ b/tests/PHPStan/Type/SimultaneousTypeTraverserTest.php
@@ -125,6 +125,13 @@ class SimultaneousTypeTraverserTest extends PHPStanTestCase
 			$chooseScalarSubtype,
 			"'aaa'|Foo",
 		];
+
+		yield [
+			'list<string|null>',
+			'array<string>',
+			$chooseScalarSubtype,
+			'list<string>',
+		];
 	}
 
 	/**
```

ondrejmirtes:
Alright here's something https://github.com/phpstan/phpstan-src/commit/0693e5aff60577310b71bfe10a1bf2c2d09f86d4

But I don't think it works well for constant arrays for example or other situations, but some use cases could be solved now.

VincentLanglet:
> Alright here's something [0693e5a](https://github.com/phpstan/phpstan-src/commit/0693e5aff60577310b71bfe10a1bf2c2d09f86d4)
> 
> But I don't think it works well for constant arrays for example or other situations, but some use cases could be solved now.

Isn't it weird to only have the case `if ($this->isArray()->yes() && $right->isArray()->yes()) {` handled in the IntersectionType::traverseSimultaneously ? I feel like you lost all the other possible intersection like
- string + Accessory
- object like A&B
- ...

ondrejmirtes:
Well it might not even be needed, depends on how the callback is written. You can send me some failing tests with expectations.

VincentLanglet:
Current status, 2 errors:
```
1) PHPStan\Rules\Functions\ReturnTypeRuleTest::testListWithNullablesUnchecked
--- Expected
+++ Actual
@@ @@
-'
+'16: Function ReturnListNullables\doFoo() should return array<string>|null but returns list<string|null>.
 '

/home/runner/work/phpstan-src/phpstan-src/src/Testing/RuleTestCase.php:179
/home/runner/work/phpstan-src/phpstan-src/tests/PHPStan/Rules/Functions/ReturnTypeRuleTest.php:186

2) PHPStan\Rules\Functions\ReturnTypeRuleTest::testBug8846
--- Expected
+++ Actual
@@ @@
-'
+'23: Function Bug8846\test() should return Bug8846\Foo::THREE|Bug8846\Foo::TWO but returns Bug8846\Foo.
 '
```

First one is maybe an Intersection type with `$this->isArray()->yes()` (list) but `$right->isArray()->maybe()` (array|null)...

VincentLanglet:
Every tests pass with the following fix @ondrejmirtes 
https://github.com/phpstan/phpstan-src/pull/4210/commits/993eae2

But I'd like your opinion on it before starting writing non regression tests.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/pull/6392

PR #4211, opened 2025-08-06
Make DOMNamedNodeMap generic

stefanfisk:
Resolves https://github.com/phpstan/phpstan/issues/13365.

### DIFF
--- a/stubs/dom.stub
+++ b/stubs/dom.stub
@@ -31,12 +31,12 @@ class DOMNode
 {
 
 	/**
-	 * @var DOMNamedNodeMap|null
+	 * @var DOMNamedNodeMap<DOMAttr>|null
 	 */
 	public $attributes;
 
 	/**
-	 * @phpstan-assert-if-true =DOMNamedNodeMap $this->attributes
+	 * @phpstan-assert-if-true =DOMNamedNodeMap<DOMAttr> $this->attributes
 	 * @return bool
 	 */
 	public function hasAttributes() {}
@@ -46,7 +46,7 @@ class DOMNode
 class DOMElement extends DOMNode
 {
 
-	/** @var DOMNamedNodeMap */
+	/** @var DOMNamedNodeMap<DOMAttr> */
 	public $attributes;
 
 	/** @var DOMDocument */
@@ -69,10 +69,9 @@ class DOMElement extends DOMNode
 
 /**
  * @template-covariant TNode as DOMNode
- * @implements Traversable<int, TNode>
  * @implements IteratorAggregate<int, TNode>
  */
-class DOMNodeList implements Traversable, IteratorAggregate, Countable
+class DOMNodeList implements IteratorAggregate, Countable
 {
 
 	/**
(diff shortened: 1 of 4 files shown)


stefanfisk:
Am I correct in assuming that I need to make `DOMNodeList` generic and use `DOMNodeList<DOMAttr> to resolve this in a way that works with the e2e tests?

stubs/dom.stub
```diff
@@ -155,11 +155,37 @@ class DOMProcessingInstruction
...
 
 /**
+ * @template-covariant TNode as DOMNode
+ * @implements Traversable<int, TNode>
+ * @implements IteratorAggregate<int, TNode>
+ *
  * @property-read int $length
  */
-class DOMNamedNodeMap
+class DOMNamedNodeMap implements Traversable, IteratorAggregate, Countable
```
VincentLanglet: It shouldn't need to implements both Traversable and IteratorAggregate in the stub.

See
https://github.com/JetBrains/phpstorm-stubs/blob/6c034ae018ed7ac2cacf2474f824aede73658afa/dom/dom_c.php#L1381

stefanfisk: @VincentLanglet I just copied it from `DOMNodeList` at https://github.com/phpstan/phpstan-src/blob/c0e6f0e0c00aa1e7c8ffe395ffdf4f881985a6c9/stubs/dom.stub#L75

Do you want me to fix `DOMNodeList` as well?

VincentLanglet: If it's done this way for DOMNodeList you can just ignore my remark I think

ondrejmirtes: Please fix both DOMNodeList and DOMNamedNodeMap here.

There are some classes where it'd make sense because of the logic here https://github.com/ondrejmirtes/BetterReflection/blob/61b25baaca1ea904447f46d975a4ae7c99b722e6/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php#L528-L574, but it's not true for any `DOM*` classes.

ondrejmirtes (changes requested):
Additionally please add DOMNamedNodeMap to https://github.com/phpstan/phpstan-src/blob/5a3990227e64a66b058a96753887e4aa7d411b92/conf/config.neon#L29 `skipCheckGenericClasses` array.

stubs/dom.stub
stefanfisk: I've now removed `Traversable` from all DOM classes.

stefanfisk:
I added `DOMNamedNodeMap` to `skipCheckGenericClasses`. What does it do?

ondrejmirtes:
> I added DOMNamedNodeMap to skipCheckGenericClasses. What does it do?

Normally generic classes are required to specify the generic tempalte types in all typehints. So `@param DOMNamedNodeMap $a` is not sufficient, you need to type `@param DOMNamedNodeMap<X> $a`.

But making a non-generic class generic in a non-major PHPStan version is a BC break, that's where skipCheckGenericClasses comes in. It skips this requirement for the listed classes.

ondrejmirtes:
Thank you.

PR #4213, opened 2025-08-07
[FIX] dynamic variable undefined inside `isset`

ga-devfront:
Fix https://github.com/phpstan/phpstan/issues/13353
This is the fix for the issue above. If you see any changes to make, please let me know; this is my first contribution to the project.

### DIFF
--- a/src/Rules/Variables/DefinedVariableRule.php
+++ b/src/Rules/Variables/DefinedVariableRule.php
@@ -41,6 +41,11 @@ final class DefinedVariableRule implements Rule
 	public function processNode(Node $node, Scope $scope): array
 	{
 		$errors = [];
+
+		if ($scope->isUndefinedExpressionAllowed($node)) {
+			return [];
+		}
+
 		if (is_string($node->name)) {
 			$variableNameScopes = [$node->name => $scope];
 		} else {
@@ -78,7 +83,7 @@ final class DefinedVariableRule implements Rule
 			}
 		}
 
-		if ($scope->isInExpressionAssign($node) || $scope->isUndefinedExpressionAllowed($node)) {
+		if ($scope->isInExpressionAssign($node)) {
 			return [];
 		}
 
--- a/tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
+++ b/tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
@@ -1169,4 +1169,13 @@ class DefinedVariableRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/bug-8719.php'], []);
 	}
 
+	public function testDynamicVariable(): void
+	{
+		$this->cliArgumentsVariablesRegistered = true;
+		$this->polluteScopeWithLoopInitialAssignments = true;
+		$this->checkMaybeUndefinedVariables = true;
+		$this->polluteScopeWithAlwaysIterableForeach = true;
+
+		$this->analyse([__DIR__ . '/data/dynamic-variable.php'], []);
+	}
 }
--- a/tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
+++ b/tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
@@ -1169,14 +1169,14 @@ class DefinedVariableRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/bug-8719.php'], []);
 	}
 
-	public function testDynamicVariableInsideIsset(): void
+	public function testBug13353(): void
 	{
 		$this->cliArgumentsVariablesRegistered = true;
 		$this->polluteScopeWithLoopInitialAssignments = true;
 		$this->checkMaybeUndefinedVariables = true;
 		$this->polluteScopeWithAlwaysIterableForeach = true;
 
-		$this->analyse([__DIR__ . '/data/dynamic-variable-inside-isset.php'], [
+		$this->analyse([__DIR__ . '/data/bug-13353.php'], [
 			[
 				'Variable $bar might not be defined.',
 				9,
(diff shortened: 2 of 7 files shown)


tests/PHPStan/Rules/Variables/data/dynamic-variable.php
```diff
@@ -0,0 +1,7 @@
+<?php declare(strict_types = 1);
+
+$foo = 'bar';
```
VincentLanglet: It's often better to have a namespace in tests file if possible.

Like `namespace DynamicVariable;` or `namespace Bug13353`

tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
```diff
@@ -1169,4 +1169,14 @@ public function testBug8719(): void
 		$this->analyse([__DIR__ . '/data/bug-8719.php'], []);
 	}
 
+	public function testDynamicVariable(): void
```
VincentLanglet: I often name the test `testBug13353` and the file `bug-13353.php` when it's related to an issue.
Not sure if it will be wanted by ondrej

tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
```diff
@@ -1169,4 +1169,19 @@ public function testBug8719(): void
...
+	{
+		$this->cliArgumentsVariablesRegistered = true;
+		$this->polluteScopeWithLoopInitialAssignments = true;
+		$this->checkMaybeUndefinedVariables = true;
+		$this->polluteScopeWithAlwaysIterableForeach = true;
+
+		$this->analyse([__DIR__ . '/data/bug-13353.php'], [
+			[
+				'Variable $bar might not be defined.',
+				9,
```
VincentLanglet:
```suggestion
				11,
```
with the namespace it changed the line ;)

ga-devfront: Really thank you for your feedbacks !

VincentLanglet (approved):

tests/PHPStan/Rules/Variables/data/bug-13353.php
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Bug13353;
+
+$foo = 'bar';
+
+if (!isset($$foo)) {
+	echo 'Wololo';
+}
+
+echo $$foo;
```
ondrejmirtes, line 11: Please also test the same code inside an anonymous function. Scope behaves a bit differently about undefined variables there.

ondrejmirtes (changes requested):
Please look at this problem more thoroughly. There's IssetCheck (https://github.com/phpstan/phpstan-src/blob/5a3990227e64a66b058a96753887e4aa7d411b92/src/Rules/IssetCheck.php#L45) used by several rules where only Variable with string name property is being handled, but the code might find some improvements to the logic useful.

ga-devfront:
> Please look at this problem more thoroughly. There's IssetCheck (
> 
> https://github.com/phpstan/phpstan-src/blob/5a3990227e64a66b058a96753887e4aa7d411b92/src/Rules/IssetCheck.php#L45
> 
> ) used by several rules where only Variable with string name property is being handled, but the code might find some improvements to the logic useful.


It was my first idea to fix the problem in the IssetCheck, however, the error being returned actually comes from DefinedVariableRule. Being new to the project, I don't yet understand how it all works. However, if you could tell me a little more about what you'd like to see added, there's no problem; I'm open to any suggestions.
I've added the tests as requested.

ondrejmirtes:
IssetCheck is used by IssetRule and others. Look at IssetRuleTest what it detects.

You should be able to produce some tests with `$$` to see what's currently wrong about the behaviour.

ga-devfront:
> IssetCheck is used by IssetRule and others. Look at IssetRuleTest what it detects.
> 
> You should be able to produce some tests with `$$` to see what's currently wrong about the behaviour.

As you can see, I added the tests in IssetRuleTest.php, but nothing is thrown.
Maybe I missed something, feel free to let me know.

VincentLanglet:
The branch has conflict @ga-devfront ; are you interested in rebasing/finishing the PR ?

ga-devfront:
> The branch has conflict @ga-devfront ; are you interested in rebasing/finishing the PR ?

Rebase done. But don't know what to do more for finishing it.

ga-devfront:
@ondrejmirtes up please :heart:

src/Rules/Variables/DefinedVariableRule.php
```diff
@@ -41,6 +41,11 @@ public function getNodeType(): string
 	public function processNode(Node $node, Scope $scope): array
 	{
 		$errors = [];
+
+		if ($scope->isUndefinedExpressionAllowed($node)) {
```
ondrejmirtes, line 45: Why does this fix work exactly? Looking at the code I'm not sure why moving this condition makes a difference.

ondrejmirtes (changes requested):

src/Rules/Variables/DefinedVariableRule.php
ga-devfront: @ondrejmirtes  It's been a while since I wrote this pull request, but from what I remember:
If you look closely at where `isUndefinedExpressionAllowed` was, it ends up after the expression evaluation, so an error is thrown before it even reaches that point. We put it before because in this specific case, being undefined is accepted, so we don't throw an error. I don't know if I'm being clear; the tests might speak for themselves.

ondrejmirtes: I'm a bit skeptical here. `$node` is the same, but `$scope` is after `$scope->filterByTruthyValue...` call. It would look like that this call does not preserve `isUndefinedExpressionAllowed` but it should. So maybe the code of this rule should not change, but we should look into whether Scope answers how we expect it to?

ondrejmirtes:
This PR references already closed issue.

PR #4217, opened 2025-08-08
Remove some AMQP stubs for functionMap

VincentLanglet:
Same idea than https://github.com/phpstan/phpstan-src/pull/4185
I checked the signature one by one.

More could be removed when PHPStorm will be fixed
https://github.com/JetBrains/phpstorm-stubs/pull/1775

The only change in reflection golden tests is
```
PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD AMQPExchange::delete" ('METHOD AMQPExchange::delete', 'Is built-in: Yes\nHas side-ef... mixed')
--- Expected
+++ Actual
@@ @@
 Visibility: public
 Variants: 1
     /**
-     * @param string $exchangeName
+     * @param string|null $exchangeName
      * @param int $flags
      * @return void
      */
     function delete(mixed $exchangeName = null, mixed $flags = 0): mixed'

```
Which seems legit when looking at
https://github.com/php-amqp/php-amqp/blob/3618db5f781aa3d371b630c666bc90845d7c77d4/stubs/AMQPExchange.php#L114


ondrejmirtes:
Thank you.

PR #4218, opened 2025-08-08
Remove some Ds stubs for functionMap

VincentLanglet:
Same idea than https://github.com/phpstan/phpstan-src/pull/4185
I checked the signature one by one.

More could be removed when PHPStorm will be fixed
https://github.com/JetBrains/phpstorm-stubs/pull/1776

Reflection goldent test is lot of 
```
-Has side-effects: Maybe
+Has side-effects: Yes
```
Wanted, there is a `@throws` on the method.

And one
```
PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD Ds\Deque::sorted" ('METHOD Ds\Deque::sorted', 'Is built-in: Yes\nHas side-ef...\Deque')
--- Expected
+++ Actual
@@ @@
 Variants: 1
     /**
      * @param (callable(TValue, TValue): int)|null $comparator
-     * @return Ds\Sequence<TValue (class Ds\Deque, parameter)>
+     * @return Ds\Deque
      */
     function sorted((callable(): mixed)|null $comparator = null): Ds\Deque'

```

Then I discover there is also an `ext-ds` stub file in PHPStan so I updated it and now the change is an improvement
```
PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD Ds\Deque::sorted" ('METHOD Ds\Deque::sorted', 'Is built-in: Yes\nHas side-ef...\Deque')
--- Expected
+++ Actual
@@ @@
 Variants: 1
     /**
      * @param (callable(TValue, TValue): int)|null $comparator
-     * @return Ds\Sequence<TValue (class Ds\Deque, parameter)>
+     * @return Ds\Deque<TValue (class Ds\Deque, parameter)>
      */
     function sorted((callable(): mixed)|null $comparator = null): Ds\Deque'

```

In another PR (and after PHPStorm PR will be merged), I'll try to get rid of the `ext-ds.stub`.


ondrejmirtes:
Thank you.

PR #4222, opened 2025-08-08
Resolve static in assert phpdoc

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12548

I didn't exactly copy-paste the snippet given because I don't understand how you can expect
```
/**
     * @template T of object
     *
     * @param T $object
     *
     * @return (T is static ? T : static)
     *
     * @phpstan-assert-if-true =static $object
     */
    public static function assertInstanceOf(object $object)
    {
        if (!$object instanceof static) {
            throw new \Error('Object is not an instance of static class');
        }

        return $object;
    }
```
to correctly work: `phpstan-assert-if-true` only work if the method returns a boolean.

Closes https://github.com/phpstan/phpstan/issues/11289
Closes https://github.com/phpstan/phpstan/issues/12376

### DIFF
--- a/src/Reflection/Dummy/ChangedTypeMethodReflection.php
+++ b/src/Reflection/Dummy/ChangedTypeMethodReflection.php
@@ -27,6 +27,7 @@ final class ChangedTypeMethodReflection implements ExtendedMethodReflection
 		private ?array $namedArgumentsVariants,
 		private ?Type $selfOutType,
 		private ?Type $throwType,
+		private Assertions $assertions,
 	)
 	{
 	}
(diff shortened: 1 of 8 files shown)


ondrejmirtes:
Aren't both https://github.com/phpstan/phpstan/issues/11289 and https://github.com/phpstan/phpstan/issues/12376 fixed too?

VincentLanglet:
> Aren't both [phpstan/phpstan#11289](https://github.com/phpstan/phpstan/issues/11289) and [phpstan/phpstan#12376](https://github.com/phpstan/phpstan/issues/12376) fixed too?

Indeed, I added non regression tests

ondrejmirtes:
Thank you.

PR #4223, opened 2025-08-08
Fix wrong inference on preg_match != false

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11708
Closes https://github.com/phpstan/phpstan/issues/11908

cc @staabm

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-11293.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-11293.php
@@ -30,7 +30,7 @@ class HelloWorld
 	public function sayHello4(string $s): void
 	{
 		if (preg_match('/data-(\d{6})\.json$/', $s, $matches) <= 0) {
-			assertType('array{}', $matches);
+			assertType('list{0?: string, 1?: non-falsy-string&numeric-string}', $matches);
 
 			return;
 		}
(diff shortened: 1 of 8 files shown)


tests/PHPStan/Analyser/nsrt/bug-11293.php
```diff
@@ -30,7 +30,7 @@ public function sayHello3(string $s): void
 	public function sayHello4(string $s): void
 	{
 		if (preg_match('/data-(\d{6})\.json$/', $s, $matches) <= 0) {
-			assertType('array{}', $matches);
+			assertType('list{0?: string, 1?: non-falsy-string&numeric-string}', $matches);
```
VincentLanglet, line 33: This might be considered as a "regression" but:
- The main goal of the issue 11293 was the inference line 38
- It will be consistent with `== 0` check https://phpstan.org/r/fb9edc94-e56a-4e9e-8270-a6ae06570bb1 

BTW `array{}` is already technically not true since if preg_match return false, `$matches` is null.

staabm:
Ondrey wanted this problem solved via a more general approach - see https://github.com/phpstan/phpstan-src/pull/3881

ondrejmirtes:
Thank you! I'd like improvements to TypeSpecifierContext but we already have many hardcoded fixes in TypeSpecifier so this is just one more. It can be refactored later.

PR #4225, opened 2025-08-09
Resolve array~list

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9734

I'll need your decision on the condition to chose in ArrayType::tryRemove @ondrejmirtes 

We could have 
```
if ($typeToRemove->isSuperTypeOf(new ConstantArrayType([], []))->yes()) {
     return TypeCombinator::intersect($this, new NonEmptyArrayType());
}
```
but it means that now a 
```
@phpstan-assert-if-true string[] $array
```
will infer a non empty array for the `false` case.

It could make sens since in order to say that all the values are not string, it require to access to at least one value.

If preferred we could restrict such behavior to the list with an extra `$typeToRemove->isList()->yes()` check.

Or, to just solve `array_is_list` ; the check can be something like `$typeToRemove->describe(...) === 'list'`...

### DIFF
--- a/src/Type/ArrayType.php
+++ b/src/Type/ArrayType.php
@@ -609,6 +609,10 @@ class ArrayType implements Type
 			return TypeCombinator::intersect($this, new NonEmptyArrayType());
 		}
 
+		if ($typeToRemove->isSuperTypeOf(new ConstantArrayType([], []))->yes()) {
+			return TypeCombinator::intersect($this, new NonEmptyArrayType());
+		}
+
 		if ($typeToRemove instanceof NonEmptyArrayType) {
 			return new ConstantArrayType([], []);
 		}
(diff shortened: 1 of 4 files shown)


staabm (commented):
Do we have a test somewhere which asserts the types for `array_is_list(mixed)`?

VincentLanglet:
> Do we have a test somewhere which asserts the types for `array_is_list(mixed)`?

Dunno, but what do you have in mind ?
It gives
```
/**
	 * @param mixed $a
	 * @return void
	 */
	public function doFoo4($a): void
	{
		if (array_is_list($a)) {
			assertType('list', $a);
		} else {
			assertType('mixed~list', $a);
		}
	}
```
I added a test.

staabm:
Looks good - just making sure we have it covered

ondrejmirtes:
Thank you.

PR #4226, opened 2025-08-09
Handle default value in phpstan-assert

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9435

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-9435.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-9435.php
@@ -1,4 +1,4 @@
-<?php
+<?php // lint >= 8.0
 
 declare(strict_types=1);
 
--- a/tests/PHPStan/Analyser/nsrt/bug-9435.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-9435.php
@@ -1,0 +1,112 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Bug9435;
+
+use function PHPStan\Testing\assertType;
+
+function x(): mixed
+{
+	return null;
+}
+
+/** @phpstan-assert ($allow_null is true ? string|null : string) $input */
+function trueCheck(mixed $input, bool $allow_null = false): void
+{
+}
+
+$a = x();
+trueCheck($a);
+assertType('string', $a); // incorrect: should be string but is string|null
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/nsrt/bug-9435.php
```diff
@@ -0,0 +1,112 @@
...
+$a = x();
+trueCheck($a);
+assertType('string', $a); // incorrect: should be string but is string|null
+
+$a = x();
+trueCheck($a, false);
+assertType('string', $a); // correct (string)
+
+$a = x();
+trueCheck($a, allow_null: false);
```
staabm, line 28: Add a test where params are named and out-of-order

tests/PHPStan/Analyser/nsrt/bug-9435.php
```diff
@@ -0,0 +1,112 @@
...
+
+use function PHPStan\Testing\assertType;
+
+function x(): mixed
+{
+	return null;
+}
+
+/** @phpstan-assert ($allow_null is true ? string|null : string) $input */
+function trueCheck(mixed $input, bool $allow_null = false): void
```
staabm, line 15: Maybe test a function expecting variadics

VincentLanglet: What do you mean ? Could you give a phpstan.org snippet ?

tests/PHPStan/Analyser/nsrt/bug-9435.php
VincentLanglet: Done

tests/PHPStan/Analyser/nsrt/bug-9435.php
staabm: Something like

```
function variadic(mixed $input, bool $allow_null = false, ...$more): void
```

VincentLanglet: Like this  https://github.com/phpstan/phpstan-src/pull/4226/commits/d365526c3e5d7dbab412de99c01f77de488b8e85 ?

ondrejmirtes:
Thank you.

PR #4227, opened 2025-08-09
Consider assert as side effect

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12224


ondrejmirtes (changes requested):
What about CallToConstructorStatementWithoutSideEffectsRule?

VincentLanglet:
> What about CallToConstructorStatementWithoutSideEffectsRule?

The issue only appear if you use both `@phpstan-pure` and `@phpstan-assert`.
I dunno if it make sens to have a pure constructor.

Anyway it doesn't really cost to add the check, so I updated the PR.

ondrejmirtes:
Thank you.

PR #4228, opened 2025-08-09
Better support for -0.0

VincentLanglet:
This could be a solution for the issue https://github.com/phpstan/phpstan/issues/12225
- Stop merging `0.0` and `-0.0` since they have a different toString value
- Consider `=== 0.0` like `=== 0.0 || === -0.0`
- Same for `== 0`

The other solution I had in mind for the issue was to "just" update ConstantFloat::toString but it has the drawback that
```
(string) 0.0
```
would have been considered as `'0.0'|'-0.0'` while here the constant was really known.

Closes https://github.com/phpstan/phpstan/issues/12225

What could be done to have less changes, would be to modify UnionType::describe to merge 0.0 and -0.0.
But I feel like it's better to be explicit, (since 0.0 and -0.0 have a different behavior) and also, there is already some `-0.0` displayed by PHPStan like in the following snippet (and you'll see the already existing inconsistency).
https://phpstan.org/r/0b3cdd4f-b228-4272-a685-621afcf21dd5

cc @staabm (cause I feel like you worked a lot on strict/loose comparison.


staabm:
in https://github.com/phpstan/phpstan/issues/12225#issuecomment-2614062692 you mentioned a always-true check.

will this error be reported after this PR and needs to be test-covered?

VincentLanglet:
> in [phpstan/phpstan#12225 (comment)](https://github.com/phpstan/phpstan/issues/12225#issuecomment-2614062692) you mentioned a always-true check.
> 
> will this be reported after this PR and needs to be test-covered?

No, but it's finally another issue because you already has the issue with something like
https://phpstan.org/r/7edaff8c-cf6f-4b62-b6e9-e7a25e4b7300

ondrejmirtes:
What if instead of all of these changes, we'd just change `ConstantFloatType::toString()`?

VincentLanglet:
The other solution I had in mind for the issue indeed was to "just" update ConstantFloat::toString but it has the drawback that
 
 ```
 (string) 0.0
 ```
 
 would have been considered as `'0.0'|'-0.0'` while here the constant was really known.

While with this implementation I have the behavior
```
assertType("'0'", (string) 0.0);
assertType("'-0'", (string) -0.0);
```

Your call.

ondrejmirtes:
Yeah, I'm probably fine with that. We could improve that in the future by adding `bool $positive` into ConstantFloatType constructor (but not now).

VincentLanglet:
Closing in favor of https://github.com/phpstan/phpstan-src/pull/4278 then

PR #4229, opened 2025-08-10
Introduce checkTooWideParameterOutInProtectedAndPublicMethods

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12080

I copied the strategy from TooWideMethodReturnTypehintRule
https://github.com/phpstan/phpstan-src/blob/745ae5a988b7e16185f927663ecd7cdd64f63425/src/Rules/TooWideTypehints/TooWideMethodReturnTypehintRule.php#L45-L56

But I didn't add the `isFirstDeclaration` check yet because it removes almost every reported errors... (and no early return was added before). Also, I already don't really understand the check in TooWideMethodReturnTypehintRule

Cause we're getting
```
class WithoutChild
{
     public function foo(): ?int // No error, but it "should" be reported with checkPublicAndProtected
	{
		return 42;
	}
}

class A
{
	public function foo(): int|string // No error cause it's the first declaration
	{
		return 42;
	}
}

class B extends A
{
	public function foo(): int|string // never returns string so it can be removed from the return type => but it will crash class C.
	{
		return 42;
	}
}

class C extends B
{
	public function foo(): int|string // never returns int so it can be removed from the return type
	{
		return '42';
	}
}
```

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -54,6 +54,7 @@ parameters:
 	checkPhpDocMethodSignatures: false
 	checkExtraArguments: false
 	checkMissingTypehints: false
+	checkTooWideParameterOutInProtectedAndPublicMethods: false
 	checkTooWideReturnTypesInProtectedAndPublicMethods: false
 	checkUninitializedProperties: false
 	checkDynamicProperties: false
--- a/conf/parametersSchema.neon
+++ b/conf/parametersSchema.neon
@@ -58,6 +58,7 @@ parametersSchema:
 	checkPhpDocMethodSignatures: bool()
 	checkExtraArguments: bool()
 	checkMissingTypehints: bool()
+	checkTooWideParameterOutInProtectedAndPublicMethods: bool()
 	checkTooWideReturnTypesInProtectedAndPublicMethods: bool()
 	checkUninitializedProperties: bool()
 	checkDynamicProperties: bool()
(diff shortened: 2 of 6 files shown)


ondrejmirtes:
Thank you!

Yeah, the `$isFirstDeclaration` is definitely weird.

ondrejmirtes:
Please contribute docs update to Config Reference.

ondrejmirtes:
I'm waiting on that docs update :) Thank you.

VincentLanglet:
> I'm waiting on that docs update :) Thank you.

Yes, on my todo list ; I'm back from my 2-3 day vacations

ondrejmirtes:
No problem, no pressure 😊

VincentLanglet:
https://github.com/phpstan/phpstan/pull/13480

VincentLanglet:
> Yeah, the `$isFirstDeclaration` is definitely weird.

Should we try to remove it from `TooWideMethodReturnTypehintRule` ?

VincentLanglet:
> > Yeah, the `$isFirstDeclaration` is definitely weird.
> 
> Should we try to remove it from `TooWideMethodReturnTypehintRule` ?

Edit: I tried and seems like it's made to not report thing like
```
public function ipsum(): ?string {
		return null;
	}
```
for method which are made to be overrided

PR #4234, opened 2025-08-10
Declare assert-if-true for filesystem functions

staabm:
closes https://github.com/phpstan/phpstan/issues/6788

simplified version of https://github.com/phpstan/phpstan-src/pull/1993 with reduced scope to fix the build


ondrejmirtes:
Thank you!

PR #4237, opened 2025-08-11
 Fix false positive of function.alreadyNarrowedType (function call variable assignment)

staabm:
closes https://github.com/phpstan/phpstan/issues/13268
closes https://github.com/phpstan/phpstan/issues/12087
closes https://github.com/phpstan/phpstan/issues/9666
closes https://github.com/phpstan/phpstan/issues/9445
closes https://github.com/phpstan/phpstan/issues/7773

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -70,13 +70,6 @@ final class ImpossibleCheckTypeHelper
 			}
 			$args = $node->getArgs();
 			$argsCount = count($args);
-
-			foreach ($args as $arg) {
-				if ($arg->value instanceof Expr\Assign) {
-					return null;
-				}
-			}
-
 			if ($node->name instanceof Node\Name) {
 				$functionName = strtolower((string) $node->name);
 				if ($functionName === 'assert' && $argsCount >= 1) {
@@ -286,6 +279,16 @@ final class ImpossibleCheckTypeHelper
 
 		$results = [];
 
+		$assignedInCallVars = [];
+		if ($node instanceof Expr\CallLike) {
+			foreach ($node->getArgs() as $arg) {
+				if (!$arg->value instanceof Expr\Assign) {
+					continue;
+				}
+
+				$assignedInCallVars[] = $arg->value;
+			}
+		}
 		foreach ($sureTypes as $sureType) {
 			if (self::isSpecified($typeSpecifierScope, $node, $sureType[0])) {
 				$results[] = TrinaryLogic::createMaybe();
@@ -300,6 +303,14 @@ final class ImpossibleCheckTypeHelper
 
 			/** @var Type $resultType */
 			$resultType = $sureType[1];
+
+			foreach ($assignedInCallVars as $assignedInCallVar) {
+				if ($sureType[0] !== $assignedInCallVar->var) {
+					continue;
+				}
+
+				$argumentType = $scope->getType($assignedInCallVar->expr);
+			}
 
 			$results[] = $resultType->isSuperTypeOf($argumentType)->result;
 		}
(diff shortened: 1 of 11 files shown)


staabm:
Looking at issuebot results tells me that I need a different approach

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -278,6 +279,16 @@ public function findSpecifiedType(
...
 		$results = [];
 
+		$assignedInCallVars = [];
+		if ($node instanceof Expr\CallLike) {
+			foreach ($node->getArgs() as $arg) {
+				if (!$arg->value instanceof Expr\Assign) {
+					continue;
+				}
+
+				$assignedInCallVars[] = $arg->value;
```
ondrejmirtes: Can be nested multiple assigns in themselves. Also it might be useful to check and unwrap `Expr\AssignOp\Coalesce` (`??=`) along with tests.

ondrejmirtes (changes requested):

src/Rules/Comparison/ImpossibleCheckTypeHelper.php
ondrejmirtes: Multiple nested assigns, meaning `$foo = $bar = ...`.

staabm: > Can be nested multiple assigns in themselves.

fixed

> Also it might be useful to check and unwrap `Expr\AssignOp\Coalesce` (`??=`) along with tests.

added some basic tests which work as expected. 
more complicated with a mix of assign and coalesce feel alien to me and is pretty hard to read. I would not add support for such non-readable code if possible.

staabm:
coming from this PR here and looking at the issue in https://github.com/phpstan/phpstan/issues/12234 it seems like a better approach would be to have a per argument Scope object.

It looks like the scope object passed to the rules is not seeing the assignment early enough and therefore reasons with a not updated scope. If I am right it might mean we need a better fix here and it could also cover https://github.com/phpstan/phpstan/issues/12234

wdyt?

ondrejmirtes:
Yeah, this is kind of a problem with how PHPStan is written on principle. We have multiple places like NodeScopeResolver, TypeSpecifier, MutatingScope::resolveType, and rules. All of these places need to use the appropriate Scope for the AST property they're asking about. And sometimes they do duplicate work.

I was able to fix some bugs by doing the same thing in TypeSpecifier and MutatingScope that's already been done in NodeScopeResolver. Example: https://github.com/phpstan/phpstan-src/commit/989dd6fab8b0bf66777e20c533c90c7841d05c3a

Sometimes I'm able to solve this with virtual nodes. LiteralArrayNode and LiteralArrayItem exists so that a rule can see the correct Scope for each array item.

The bug in https://phpstan.org/r/8d3536cc-d647-455c-8919-300cc35d7abf is present because although NodeScopeResolver updates the Scope when processing each argument, and providing the Scope for the next argument, the rules that call FunctionCallParametersCheck simply hook onto FuncCall or MethodCall, and use the Scope before the call to get all argument types.

I wonder: If we were to write the analyser from scratch today, how would we go about it for this problem to never happen? I've long thought that NodeScopeResolver/TypeSpecifier/MutatingScope::resolveType should maybe be one thing that just goes through AST once. But I don't know how to tackle that we don't know beforehand what type they will ask for.

ondrejmirtes:
Thank you!

ondrejmirtes:
In regard to my comment: https://github.com/phpstan/phpstan-src/pull/4237#issuecomment-3262472961

There would of course be a great performance improvement to the fact that `processExprNode` would have to happen only a single time for each expression, not multiple times for Ternary, Boolean* and Logical*...

PR #4241, opened 2025-08-13
Fix false-positive: Method never returns XX so it can be removed from the return type

staabm:
closes https://github.com/phpstan/phpstan/issues/10312


staabm:
PHP8.5 nette deprecations are fixed upstream with https://github.com/nette/di/pull/325

ondrejmirtes:
Should fix the conflicts by moving the changes to TooWideTypeCheck :)

ondrejmirtes:
Thank you.

PR #4242, opened 2025-08-13
Improve `str_replace` return type

andersonamuller:
When the config `checkBenevolentUnionTypes` is enabled the `str_replace` starts to always reporting the return as `string|list<string>` but in reality the return depends on the `$subject` param type.

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -256,7 +256,7 @@ function preg_filter($pattern, $replacement, $subject, int $limit = -1, &$count 
  * @param array<string>|string $replace
  * @param array<string>|string $subject
  * @param-out int $count
- * @return list<string>|string
+ * @return ($subject is array ? list<string> : string)
  */
 function str_replace($search, $replace, $subject, ?int &$count = null) {}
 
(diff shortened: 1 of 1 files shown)


staabm:
which error is this PR trying to solve? can you reproduce the error/problem on [phpstan.org/try](https://phpstan.org/try) ?

andersonamuller:
How can I create a try reproducible with the config `checkBenevolentUnionTypes` enabled?
https://phpstan.org/r/e850992d-7494-49b0-9ccb-0af488bd1710

staabm:
I see. you cannot set a config flag on phpstan.org/try.

please add a test-case in ReturnTypeRuleTest which fails without your fix

VincentLanglet:
For the record, the computed type is done in a dynamic return type extension and there is already a PR which does some improvement on it https://github.com/phpstan/phpstan-src/pull/4023

That might not solve your issue thought, but be aware of conflicts.
Also a reproducer would be good first.

VincentLanglet:
Hi @andersonamuller do you still encounter an issue with the latest version now https://github.com/phpstan/phpstan-src/pull/4023 is merged ?

andersonamuller:
I don't have the issue anymore on the latest version. Thanks @VincentLanglet

PR #4243, opened 2025-08-14
Report bool properties as too wide when only used with true/false

staabm:
closes https://github.com/phpstan/phpstan/issues/13384

### DIFF
--- a/build/phpstan.neon
+++ b/build/phpstan.neon
@@ -115,7 +115,6 @@ parameters:
 			count: 1
 			path: ../src/Diagnose/PHPStanDiagnoseExtension.php
 		- '#^Short ternary operator is not allowed#'
-
 	reportStaticMethodSignatures: true
 	tmpDir: %rootDir%/tmp
 	stubFiles:
--- a/build/phpstan.neon
+++ b/build/phpstan.neon
@@ -115,6 +115,7 @@ parameters:
 			count: 1
 			path: ../src/Diagnose/PHPStanDiagnoseExtension.php
 		- '#^Short ternary operator is not allowed#'
+
 	reportStaticMethodSignatures: true
 	tmpDir: %rootDir%/tmp
 	stubFiles:
--- a/src/Rules/TooWideTypehints/TooWideTypeCheck.php
+++ b/src/Rules/TooWideTypehints/TooWideTypeCheck.php
@@ -86,22 +86,22 @@ final class TooWideTypeCheck
 			return [];
 		}
 
-		$functionReturnTypes = [];
+		$returnTypes = [];
 		foreach ($returnStatements as $returnStatement) {
 			$returnNode = $returnStatement->getReturnNode();
 			if ($returnNode->expr === null) {
-				$functionReturnTypes[] = new VoidType();
+				$returnTypes[] = new VoidType();
 				continue;
 			}
 
-			$functionReturnTypes[] = $returnStatement->getScope()->getType($returnNode->expr);
+			$returnTypes[] = $returnStatement->getScope()->getType($returnNode->expr);
 		}
 
 		if (!$statementResult->isAlwaysTerminating()) {
-			$functionReturnTypes[] = new VoidType();
+			$returnTypes[] = new VoidType();
 		}
 
-		$returnType = TypeCombinator::union(...$functionReturnTypes);
+		$returnType = TypeCombinator::union(...$returnTypes);
 
 		if (
 			$returnType->isConstantScalarValue()->yes()
--- a/src/Rules/TooWideTypehints/TooWideTypeCheck.php
+++ b/src/Rules/TooWideTypehints/TooWideTypeCheck.php
@@ -69,7 +69,7 @@ final class TooWideTypeCheck
 	): array
 	{
 		$functionReturnType = TypeUtils::resolveLateResolvableTypes($functionReturnType);
-		if (!$functionReturnType instanceof UnionType) {
+		if (!$functionReturnType instanceof UnionType && !$functionReturnType->isBoolean()->yes()) {
 			return [];
 		}
 		$statementResult = $node->getStatementResult();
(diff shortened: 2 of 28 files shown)


build/phpstan.neon
```diff
@@ -115,6 +115,11 @@ parameters:
 			count: 1
 			path: ../src/Diagnose/PHPStanDiagnoseExtension.php
 		- '#^Short ternary operator is not allowed#'
+		-
+			identifier: property.unusedType
+			message: '#\(bool\) is never assigned (false|true) so it can be removed from the property type#'
+			path: ../tests/PHPStan/*RuleTest.php
+
```
staabm: we cannot use true/false property types in tests, as the simple-downgrader cannot downgrade them.
therefore ignore the errors

ondrejmirtes (changes requested):
1) This doesn't have to be limited to properties
2) This doesn't have to be limited to PHP versions with native standalone `true`/`false` types support. We have PHPDocs after all.

There are many `TooWide*` rules with duplicated logic and maybe some variances in the logic. First we should refactor and deduplicate the logic where possible.

Then we could introduce a new bleeding edge toggle to enable reporting "too wide bool".

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -440,11 +440,13 @@ public function testBug4715(): void
 	public function testBug4734(): void
 	{
 		$errors = $this->runAnalyse(__DIR__ . '/data/bug-4734.php');
-		$this->assertCount(3, $errors);
+		$this->assertCount(5, $errors); // could be 3
```
staabm, line 444: this is a topic for a followup: [`NodeScopeResolver`'s `ClassStatementsGatherer`](https://github.com/phpstan/phpstan-src/blob/5a3990227e64a66b058a96753887e4aa7d411b92/src/Analyser/NodeScopeResolver.php#L977-L1000) works on a per class level, which means access to [private properties from a closure-scope (from within a different class)](https://github.com/phpstan/phpstan-src/blob/5a3990227e64a66b058a96753887e4aa7d411b92/tests/PHPStan/Analyser/data/bug-4734.php#L36-L43) are not seen when analyzing the class which declares the property

staabm:
I think we are good to go

tests/PHPStan/Rules/Comparison/LogicalXorConstantConditionRuleTest.php
```diff
@@ -11,8 +11,10 @@
 class LogicalXorConstantConditionRuleTest extends RuleTestCase
 {
 
+	/** @var true */
```
VincentLanglet: Rather than adding `@var true` I wonder
- If we should a test with false
- If we should just remove the var

Same for others...

src/Rules/TooWideTypehints/TooWideTypeCheck.php
```diff
@@ -65,11 +66,15 @@ public function checkFunction(
 		Type $functionReturnType,
 		string $functionDescription,
 		bool $checkDescendantClass,
+		bool $reportTooWideBool,
```
ondrejmirtes: This should be a constructor parameter, not a function parameter.

ondrejmirtes (changes requested):

src/Rules/TooWideTypehints/TooWideTypeCheck.php
staabm: done

ondrejmirtes (changes requested):
Hey, you're going to kill me 😅 

I'm a bit worried about the pedantry of this rule. I realized I don't want it to force people to be adding `@return false` when they just have a `: bool` return type above their functions.

So I think it should work like this:

* On all PHP versions: When the function has `@return bool`, the rule can enforce to change it to `@return true` or `@return false` (same for properties with `@var`)
* On PHP 8.2+: When the function does not have a PHPDoc `bool` type, only a native `bool` type, the rule can enforce it being changed to `true` or `false`.

The places that call TooWideTypeCheck have easy access to getPhpDocReturnType / getNativeReturnType so it can be easily passed to `check*` methods as arguments.

Please let me know if you want to change the logic, else I can tackle it myself.

ondrejmirtes:
Also: we should give these errors different error messages (that tell the user what they need to do) with differenr error identifiers.

staabm:
> Hey, you're going to kill me 😅

;)

> So I think it should work like this:

should we keep the bleeding edge toggle then again?

ondrejmirtes:
Yeah, all of this has to happen only with bleeding edge.

The only exception would be if we went back in time and managed to implement the part with native types before 8.2 was released. (New rules for PHP 8.5 today are always reported, even without bleeding edge)

staabm:
~~adjusted per feedback~~

ohh I missed this one:

> Also: we should give these errors different error messages (that tell the user what they need to do) with differenr error identifiers.

staabm:
> Also: we should give these errors different error messages (that tell the user what they need to do) with differenr error identifiers.

here we go. I guess next step is a auto-fixer?

ondrejmirtes:
> here we go. I guess next step is a auto-fixer?

I'm not sure about that. I'd like the auto-fixes to be a slam dunk, to be always correct. I'd say they are only suitable for opinionated "coding standard" rules. Like I'm doing for phpstan-src here: https://github.com/phpstan/phpstan-src/tree/2.1.x/build/PHPStan/Build

Let's take your original issue (https://phpstan.org/r/5bbec0ea-63ee-46e0-8584-2f232c522df3) - the solution isn't to change `private static bool $registered = false;` to `private static false $registered = false;` but different.

staabm:
I see. Makes sense.

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
ondrejmirtes: Yeah of course, static analysis is not fit for these shenanigans 😊 Especially for private properties.

ondrejmirtes:
Perfect, thank you!

vrana:
I think that this change caused several false positives in our codebase. I'll try to find a minimal reproducible case.

vrana:
This is an example: https://phpstan.org/r/1695567f-4869-428f-8cb6-fdd8cb147be1

It seems that the problem is not caused by this change but by something else. It looks like the result of the first `$driver->isConnected()` is cached. If I delete it then it works as expected. PHPStan probably thinks that this method is _pure_.

vrana:
It's https://github.com/phpstan/phpstan/issues/10215.

PR #4244, opened 2025-08-14
get rid of couple instanceof *Type checks in TypeCombinator

canvural:
Hello 👋🏽 

It is hard to describe how I came up with this patch. But I thought maybe migrating from `instanceof *Type` to method calls is a good enough reason to get it accepted 😅 

There was one test failing. But I think this updated version is more correct.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1720,25 +1720,7 @@ parameters:
 			path: src/Type/TypeCombinator.php
 
 		-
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
What do you think about the issue-bot report? https://github.com/phpstan/phpstan-src/actions/runs/16970952217

canvural:
> What do you think about the issue-bot report? https://github.com/phpstan/phpstan-src/actions/runs/16970952217

Yeah I saw it. But it looks it's the same case we have in the test file I've changed. The array key became more precise, which is correct IMO.

ondrejmirtes:
Thank you!

PR #4246, opened 2025-08-17
Fix incorrect type inference for @phpstan-assert-if-true on $this with union types

yt-catpaw:
closes https://github.com/phpstan/phpstan/issues/13358
closes https://github.com/phpstan/phpstan/issues/11441

### Summary
`@phpstan-assert-if-true` on `$this` with union types produced incorrect type inference.

### Repro
See added test: `tests/PHPStan/Rules/Methods/AssertIfTrueOnThisTest.php`.

### Cause
Union handling for `$this` assertions wasn’t propagated into `TypeSpecifier` (assert-based refinement path).

### Fix
Normalize/refine `$this` union members before producing `SpecifiedTypes` from assert info.

### Tests
- Added/updated tests in `AssertIfTrueOnThisTest.php`.
- Full test suite passes locally.
- 
### Also fixes #11441

- Added regression test in `tests/PHPStan/Analyser/nsrt/bug-11441.php`.
- This ensures that `@phpstan-assert !null $this->getParam()` works correctly when called on union types (`Foo|Bar`), where each class defines its own method.
- Previously, the `null` type was not removed in such cases. After the fix, `int|string` is correctly inferred.


staabm:
please have a look at the [issue-bot github action job summary](https://github.com/phpstan/phpstan-src/actions/runs/17018504744?pr=4246). it seems this PR also affects other issues.

please add regression tests for those which are getting fixed

yt-catpaw:
Thanks, @staabm. I’ll review the issue-bot GitHub Action job summary and add regression tests for the additionally affected issues. This is my first time contributing to open source (and my first PR to PHPStan), so I’ll update this PR in small batches as I add tests. If you prefer a specific location or pattern for these tests, please let me know.

staabm:
Welcome 🤗 

Type related tests go usually into the https://github.com/phpstan/phpstan-src/tree/2.1.x/tests/PHPStan/Analyser/nsrt/ directory. 

Files located in this directory will automatically checked against using assertType().

ondrejmirtes:
Regression tests are typically added to existing test classes for rules that reported the (disappeared) errors.

Alternatively covering some bugs with simple type inference tests (in `tests/PHPStan/Analyser/nsrt`) is good enough.

tests/PHPStan/Rules/Methods/data/bug-13358.php
```diff
@@ -0,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace Bug13358;
```
staabm, line 3: You can move this also into nsrt/ folder

tests/PHPStan/Rules/Methods/AssertIfTrueOnThisTest.php
```diff
@@ -0,0 +1,24 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Methods;
+
+use PHPStan\Testing\TypeInferenceTestCase;
+
+class AssertIfTrueOnThisTest extends TypeInferenceTestCase
```
staabm: This file can be dropped when everything else moved into nsrt/

yt-catpaw: Done.
https://github.com/phpstan/phpstan-src/pull/4246/commits/31b55ac355cc16dd2c18ebaa4165135a2c181dbe

tests/PHPStan/Rules/Methods/data/bug-13358.php
yt-catpaw: Done.
https://github.com/phpstan/phpstan-src/pull/4246/commits/31b55ac355cc16dd2c18ebaa4165135a2c181dbe

yt-catpaw:
@staabm
Hi, I’ve moved the test case for bug-13358 into tests/PHPStan/Analyser/nsrt/bug-13358.php, but now I’m not sure how to run just this specific test file.

The reason I want to run it individually is because I’d like to debug a specific function and check only the information related to this test case, without running everything else.

Could you tell me how to do that?

Thanks!

yt-catpaw:
Added regression test for #11441 (bug-11441.php) and updated the description accordingly.
Let me know if anything else is needed!

src/Reflection/Type/UnionTypeMethodReflection.php
```diff
@@ -177,7 +177,13 @@ public function getDocComment(): ?string
 
 	public function getAsserts(): Assertions
 	{
-		return Assertions::createEmpty();
+		$assertions = Assertions::createEmpty();
+
+		foreach ($this->methods as $method) {
+			$assertions = $assertions->intersectWith($method->getAsserts());
```
ondrejmirtes, line 183: This will merge assertions from both types. Which is a wrong operation. (The method has a misleading name but it's used in IntersectionTypeMethodReflection where it's appropriate.)

In a union type what needs to be done is we need to look at assertions from all places and do some kind of intersection, figure out the common assertions.

Let's say we have `Foo|Bar` and we call a method that's on both types.

One method does some kind of assertion and the other method does a different assertion. The correct performed assertion on the union is none because they have nothing in common.

So we need to look at on which expression the assertion is performed (in this case `$this->getParam()`) and if all involved methods on the union are doing an assertion on the same expression.

Then we need to take a look at the asserted type and also take it into account only if it's the same. (It's possible that they don't have to be the same but we'd have to figure out the right operation that we can be sure about.)

ondrejmirtes (changes requested):

yt-catpaw:
It looks like this has been addressed in another PR, so I'll close this one. Thanks!

PR #4251, opened 2025-08-20
Update RangeFunctionReturnTypeExtension to include known integer ranges for keys

PrinsFrank:
Right now, the `range` function returns a list Type for large numbers. we can narrow this down, because we do actually know the upper and lower bounds of the keys, by subtracting the lower bound from the upper bound and dividing that by the step number.

I think this covers all scenarios, let me know if I've missed anything!


staabm:
Misses a type-inference test which fails without the patch using `assertType`

staabm:
@PrinsFrank please rebase and add a test

PR #4254, opened 2025-08-23
Update xml_*_handler signature

VincentLanglet:
Follow up of https://github.com/JetBrains/phpstorm-stubs/pull/1773/files

Closes https://github.com/phpstan/phpstan/issues/9970


VincentLanglet:
Tests fail for PHP < 8.4 because of php-src stub.
I try to know if it's an easy on php-src side https://github.com/php/php-src/pull/12340/files#r2311900015

If the stub are wrong, I'm unsure how it will be needed to be fixed on PHPStan (extractor ?) side...

VincentLanglet:
Confirmed from https://github.com/php/php-src/pull/12340#discussion_r2312543714 that the stub was wrong on php-src side.

ondrejmirtes:
Thank you.

PR #4258, opened 2025-08-26
fix build

staabm:
According to [php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-errors](https://www.php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-errors)
since PHP 8, createFromFormat can throws ValueError when the datetime contains NULL-bytes.

see https://github.com/JetBrains/phpstorm-stubs/pull/1760


ondrejmirtes:
Thank you!

PR #4261, opened 2025-08-26
Added proc_open() stub for param-out

staabm:
closes https://github.com/phpstan/phpstan/issues/13197
closes https://github.com/phpstan/phpstan/issues/11777

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -364,13 +364,13 @@ function header_register_callback(callable $callback): bool {}
 function register_tick_function(callable $callback, mixed ...$args): bool {}
 
 /**
- * @template P of int = int
+ * @template P of int
  *
  * @param string|list<string> $command
  * @param array<P, list<string>|resource> $descriptor_spec
- * @param array<mixed> $pipes
- * @param array<string, mixed> $env_vars
- * @param array<string, bool> $options
+ * @param mixed $pipes
+ * @param null|array<string, mixed> $env_vars
+ * @param null|array<string, bool> $options
  *
  * @param-out array<P, resource> $pipes
  *
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -364,13 +364,15 @@ function header_register_callback(callable $callback): bool {}
 function register_tick_function(callable $callback, mixed ...$args): bool {}
 
 /**
- * @param string|array<string> $command
- * @param list<list<string>|resource> $descriptor_spec
- * @param array<resource> $pipes
- * @param array<mixed> $env_vars
+ * @template P of int = int
+ *
+ * @param string|list<string> $command
+ * @param array<P, list<string>|resource> $descriptor_spec
+ * @param array<mixed> $pipes
+ * @param array<string, mixed> $env_vars
  * @param array<string, bool> $options
  *
- * @param-out list<resource> $pipes
+ * @param-out array<P, resource> $pipes
  *
  * @return resource|false
  */
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -362,3 +362,8 @@ function header_register_callback(callable $callback): bool {}
  * @param-later-invoked-callable $callback
  */
 function register_tick_function(callable $callback, mixed ...$args): bool {}
+
+/**
+ * @param-out array<resource> $pipes
+ */
+function proc_open(array|string $command, array $descriptor_spec, &$pipes, ?string $cwd, ?array $env_vars, ?array $options): resource|false {}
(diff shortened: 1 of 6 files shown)


stubs/core.stub
```diff
@@ -362,3 +362,16 @@ function header_register_callback(callable $callback): bool {}
  * @param-later-invoked-callable $callback
  */
 function register_tick_function(callable $callback, mixed ...$args): bool {}
+
+/**
+ * @param string|array<string> $command
+ * @param list<list<string>|resource> $descriptor_spec
+ * @param list<resource> $pipes
```
mvorisek, line 369: In my understanding this can be any array with non-negative int keys.

stubs/core.stub
```diff
@@ -362,3 +362,16 @@ function header_register_callback(callable $callback): bool {}
...
 function register_tick_function(callable $callback, mixed ...$args): bool {}
+
+/**
+ * @param string|array<string> $command
+ * @param list<list<string>|resource> $descriptor_spec
+ * @param array<resource> $pipes
+ * @param array<mixed> $env_vars
+ * @param array<string, bool> $options
+ *
+ * @param-out list<resource> $pipes
```
mvorisek:
```suggestion
 * @template P of int = int
 *  
 * @param string|list<string> $command
 * @param array<P, <list<string>|resource> $descriptor_spec
 * @param array<string, mixed> $env_vars
 * @param array<string, bool> $options
 *
 * @param-out array<P, resource> $pipes
```

reasoning:

## `@template / $descriptor_spec / $pipes`

`$descriptor_spec` describes the pipes returned in `$pipes` and in my understanding the keys always match.

`$pipes` param-in type is not wanted as it can be anything, not only array, but also null/uninitialized variable. Like `$matches` in `preg_match`.

## `$env_vars`

Only string keys should be allowed, as at least in bash keys cannot start with number.

ondrejmirtes (changes requested):
You're closing an issue, please add a regression test for https://phpstan.org/r/dc1d4f91-b22d-4eaa-b796-34a353062477

ondrejmirtes:
(I don't think the issue is solved so that's why I'm asking for a test for the new behaviour)

staabm:
@ondrejmirtes test added. I think it works.

tests/PHPStan/Analyser/nsrt/bug-13197.php
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	$pipes = [];
+
+	$process = @proc_open(
+		$command,
+		[
+			['pipe', 'rb'],
+			['pipe', 'wb'], // stdout
+			['pipe', 'wb'], // stderr
```
mvorisek, line 23:
```suggestion
			['pipe', 'wb'], // stderr
			4 => ['pipe', 'wb'], // https://stackoverflow.com/questions/28909347/is-it-possible-to-connect-more-than-the-two-standard-streams-to-a-terminal-in-li#28909376
```

to shown/test the result can be also not a `list` and `@template` the keys is helpful.

tests/PHPStan/Analyser/nsrt/bug-13197b.php
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	$pipes = [];
+
+	$process = @proc_open(
+		$command,
+		[
+			['pipe', 'rb'],
+			3 => ['pipe', 'wb'], // stdout
+			5 => ['pipe', 'wb'], // stderr
```
mvorisek:
```suggestion
			3 => ['pipe', 'wb'], // https://stackoverflow.com/questions/28909347/is-it-possible-to-connect-more-than-the-two-standard-streams-to-a-terminal-in-li#28909376
			5 => ['pipe', 'wb'],
```

small comment improvement, `stdout` is more or less always `1` pipe index

stubs/core.stub
```diff
@@ -362,3 +362,18 @@ function header_register_callback(callable $callback): bool {}
...
+ * @param array<P, list<string>|resource> $descriptor_spec
+ * @param mixed $pipes
+ * @param null|array<string, mixed> $env_vars
+ * @param null|array<string, bool> $options
+ *
+ * @param-out array<P, resource> $pipes
+ *
+ * @return resource|false
+ */
+function proc_open($command, array $descriptor_spec, &$pipes, ?string $cwd, ?array $env_vars, ?array $options) {}
```
mvorisek:
```suggestion
function proc_open($command, array $descriptor_spec, &$pipes, ?string $cwd = null, ?array $env_vars = null, ?array $options = null) {}
```

not sure if needed in stubs

theofidry:
Thank you @staabm

PR #4264, opened 2025-08-27
Exception.stub - better `getTrace()` return type

keradus:
not sure why subset of PHAR tests failing.
plese, do not hesitate to take over the PR if you agree with suggested change.


ondrejmirtes (changes requested):
Please search for `::getTrace` in `resources/functionMap.php` and fix it in all 29 occurrences. Thank you.

ondrejmirtes:
Thank you.

PR #4265, opened 2025-08-28
Prevent "Unsafe usage of new static() in abstract class" with final constructor

staabm:


ondrejmirtes:
I don't get it. Having a final constructor does not prevent the error: https://3v4l.org/gilGt

Please note that NewStaticRule and NewStaticInAbstractClassStaticMethodRule are different rules for different purposes.

staabm:
hmm I see. its not about the constructor signature

PR #4266, opened 2025-08-28
Fix plus operator with constant arrays

SplotyCode:


SplotyCode:
Fixed https://github.com/phpstan/phpstan/issues/11912

tests/PHPStan/Analyser/data/bug-11912.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Bug11912;
+
+/**
+ * @param array<string, mixed> $results
+ * @param list<string> $names
+ */
+function appendResults(array $results, array $names): bool {
+	// Make sure 'names' comes first in array
+	$results = ['names' => $names] + $results;
+	\PHPStan\Testing\assertType("list<string>", $results['names']);
```
ondrejmirtes: This does not need to be in AnalyserIntegrationTest. Simply move it to `tests/PHPStan/Analyser/nsrt/bug-11912.php` and it will be automatically run with NodeScopeResolverTest.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4267, opened 2025-08-28
Fix/nullable templates

SplotyCode:
Fixed https://github.com/phpstan/phpstan/issues/13048. Essentially, `T of (A|null)` subtracted by `A` is `T of mixed`

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-12894.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-12894.php
@@ -1,18 +1,25 @@
-<?php declare(strict_types = 1);
+<?php declare(strict_types = 1); // lint >= 8.0
 
 namespace Bug12894;
+
+use Closure;
 
 /**
  * @template TValue of object|null
  */
-interface Dependency {
+interface Dependency
+{
+
 	/**
 	 * @return TValue
 	 */
 	public function __invoke(): object|null;
+
 }
 
-interface DependencyResolver {
+interface DependencyResolver
+{
+
 	/**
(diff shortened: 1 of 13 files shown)


ondrejmirtes (changes requested):
1) This fixes a bunch of other issues, please check https://github.com/phpstan/phpstan-src/actions/runs/17309059596 and add regression tests for them too.
2) This has to be modified https://github.com/phpstan/phpstan-src/blob/07a29ab8cf744d6396d3b8e88984e67da57908fb/src/Rules/Generics/TemplateTypeCheck.php#L119-L142 along with a test because `of null` is now supported.

SplotyCode:
This also fixes https://github.com/phpstan/phpstan/issues/12894 and https://github.com/phpstan/phpstan/issues/12989

ondrejmirtes:
> This also fixes https://github.com/phpstan/phpstan/issues/12894 and https://github.com/phpstan/phpstan/issues/12989

This text has to be in the main PR description in order to auto-close the issues. I'll do it manually for this one.

tests/PHPStan/Analyser/nsrt/bug-12894.php
```diff
@@ -0,0 +1,51 @@
+<?php declare(strict_types = 1); // lint >= 8.0
```
ondrejmirtes:
```suggestion
<?php // lint >= 8.0

declare(strict_types = 1);
```

ondrejmirtes:
Thank you.

PR #4268, opened 2025-08-29
Fix duplicate work in AnnotationsPropertiesClassReflectionExtension->findClassReflectionWithProperty()

staabm:
closes https://github.com/phpstan/phpstan/issues/13425

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -394,7 +394,7 @@ jobs:
               ../../bin/phpstan analyze test.php --level=0
           - script: |
               cd e2e/bug13425
-              ../bashunit -a exit_code "1" ../../bin/phpstan analyze src/ plugins/
+              timeout 15 ../bashunit -a exit_code "1" ../../bin/phpstan analyze src/ plugins/
 
     steps:
       - name: "Checkout"
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -392,6 +392,9 @@ jobs:
               cd e2e/composer-version-config
               composer install
               ../../bin/phpstan analyze test.php --level=0
+          - script: |
+              cd e2e/bug13425
+              ../../bin/phpstan analyze src/ plugins/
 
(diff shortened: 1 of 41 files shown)


src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php
```diff
@@ -81,13 +81,11 @@ private function findClassReflectionWithProperty(
 		}
 
 		$parentClass = $classReflection->getParentClass();
-		while ($parentClass !== null) {
+		if ($parentClass !== null) {
 			$methodWithDeclaringClass = $this->findClassReflectionWithProperty($parentClass, $parentClass, $propertyName);
```
staabm, line 85: this recursive call will already work its way to the top parent class. we don't need another loop thru all parents.

see how the number of calls increased on each parent class before this PR:

grafik

ondrejmirtes:
The reproduction repo https://github.com/verfriemelt-dot-org/coverage-fail looks useful. Please copy the directory structure to e2e/ and try to run the analysis in E2E tests.

staabm:
added the e2e test with a timeout. I don't think it makes much sense to put expectations on the output of found errors.

.github/workflows/e2e-tests.yml
```diff
@@ -392,6 +392,9 @@ jobs:
               cd e2e/composer-version-config
               composer install
               ../../bin/phpstan analyze test.php --level=0
+          - script: |
+              cd e2e/bug13425
+              timeout 15 ../bashunit -a exit_code "1" "../../bin/phpstan analyze src/ plugins/"
```
ondrejmirtes, line 397: I couldn't find where is this `timeout` command coming from?

ondrejmirtes:
Thank you!

.github/workflows/e2e-tests.yml
staabm: its a linux command line tool, usually contained in stock installations.

`brew install coreutils` installs it on macos

PR #4269, opened 2025-08-29
Remove unused `shouldGeneralizeInferredType` method from template typ…

SplotyCode:
All classes are final


ondrejmirtes:
Thank you!

PR #4270, opened 2025-08-30
Fix setting known offset

SplotyCode:


src/Analyser/MutatingScope.php
```diff
@@ -750,6 +750,13 @@ public function getType(Expr $node): Type
 			return $this->getType($node->getVar())->unsetOffset($this->getType($node->getDim()));
 		}
 		if ($node instanceof SetOffsetValueTypeExpr) {
+			$var = $node->getVar() instanceof OriginalPropertyTypeExpr ? $node->getVar()->getPropertyFetch() : $node->getVar();
```
SplotyCode, line 753: This feels wrong. I don't really get the reason we generally wrap with OriginalPropertyTypeExpr

PR #4272, opened 2025-08-30
Fix hasOffsetValueType with object

VincentLanglet:
Cf https://www.php.net/manual/en/language.types.array.php

[Array](https://www.php.net/manual/en/language.types.array.php)s and [object](https://www.php.net/manual/en/language.types.object.php)s can not be used as keys. Doing so will result in a warning: Illegal offset type.

So ObjectType::toArrayKey should be an error.

### DIFF
--- a/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
@@ -1801,7 +1801,7 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 				'$this->resource',
 			],
 			[
-				'mixed',
+				'*ERROR*',
 				'$this->yetAnotherAnotherMixedParameter',
 			],
 			[
--- a/src/Type/ArrayType.php
+++ b/src/Type/ArrayType.php
@@ -268,6 +268,9 @@ class ArrayType implements Type
 	public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 	{
 		$offsetType = $offsetType->toArrayKey();
+		if ($offsetType instanceof ErrorType) {
+			return TrinaryLogic::createNo();
+		}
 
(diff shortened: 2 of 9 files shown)


VincentLanglet:
@ondrejmirtes you reported in https://github.com/phpstan/phpstan-src/pull/4164

> This looks wrong for array{foo: 17, bar: 19}|array{baz: 21}. int|object never exists on these arrays.

The reason was the fact that `array[$object]` was considered as "Maybe" rather than no,
as shown by https://phpstan.org/r/c6822927-88f1-4a23-a2c0-bb2fb2662e2a

I tried to fix this, but now I'm getting the a LegacyNodeScope test failure because
```
@var array[array]
```
is now understood as an error and not mixed ; it's related to
https://github.com/phpstan/phpstan-src/blob/5a3990227e64a66b058a96753887e4aa7d411b92/src/PhpDoc/TypeNodeResolver.php#L1230

Does the change seems ok to you ? If not any idea how I should fix it ?

VincentLanglet:
What I don't get currently is why

```
@param array[array]
```
is considered as mixed by PHPStan, and why
```
@var array[array]
```
is considered as `*ERROR*`.

Changing this might be a correct way to finish this PR ; I'm open to any advice ondrej.

VincentLanglet:
The behavior for `array[array]` is now consistent with 
https://phpstan.org/r/712fa80d-be94-4efb-b5e7-c700c9c1a76e

So I had to update the test.

tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -1801,7 +1801,7 @@ public static function dataProperties(): array
 				'$this->resource',
 			],
 			[
-				'mixed',
+				'*ERROR*',
```
VincentLanglet, line 1804: Phpdoc use `array[array]`

ondrejmirtes:
Thank you!

PR #4276, opened 2025-09-01
add phpdoc for iterators in unit tests

Flyingmana:
I found these places via https://github.com/phpstan/phpstan-src/pull/4209


ondrejmirtes:
Thank you.

PR #4277, opened 2025-09-02
Fix hasOffsetValueType for arrays

VincentLanglet:
I feel like we should only check for AllowedArrayKeysTypes since others types are checked by InvalidArrayKey rule.

### DIFF
--- a/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
+++ b/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
@@ -1801,7 +1801,7 @@ class LegacyNodeScopeResolverTest extends TypeInferenceTestCase
 				'$this->resource',
 			],
 			[
-				'*ERROR*',
+				'mixed',
 				'$this->yetAnotherAnotherMixedParameter',
 			],
 			[
--- a/tests/PHPStan/Levels/data/stringOffsetAccess-3.json
+++ b/tests/PHPStan/Levels/data/stringOffsetAccess-3.json
@@ -18,10 +18,5 @@
         "message": "Invalid array key type stdClass.",
         "line": 59,
         "ignorable": true
-    },
-    {
-        "message": "Offset stdClass does not exist on array{baz: 21}|array{foo: 17, bar: 19}.",
-        "line": 59,
-        "ignorable": true
     }
 ]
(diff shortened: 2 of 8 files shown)


tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php
```diff
@@ -1801,7 +1801,7 @@ public static function dataProperties(): array
 				'$this->resource',
 			],
 			[
-				'*ERROR*',
+				'mixed',
```
VincentLanglet, line 1804: This change revert the one in https://github.com/phpstan/phpstan-src/pull/4272

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -193,10 +193,6 @@ public function testStrings(): void
 				'Offset 12.34 might not exist on array|string.',
 				28,
 			],
-			[
-				'Offset int|object might not exist on array|string.',
```
VincentLanglet, line 197: This error was added by https://github.com/phpstan/phpstan-src/pull/4272

It make sens to not have it since
- int is a valid offset and produce no error
- object is not a valid offset (and produce a php crash) and is already reported by the InvalidArrayOffsetRule.

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -932,10 +928,6 @@ public function testBugObject(): void
 				'Offset int|object does not exist on array{baz: 21}|array{foo: 17, bar: 19}.',
 				12,
 			],
-			[
-				'Offset object does not exist on array<string, int>.',
```
VincentLanglet, line 936: Same, already reported by InvalidArrayOffsetRule and this was added by https://github.com/phpstan/phpstan-src/pull/4272

VincentLanglet:
The 7.4 failure already exists on 2.1.x https://github.com/phpstan/phpstan-src/actions/runs/17398668843/job/49386473327

tests/PHPStan/Levels/data/stringOffsetAccess-3.json
```diff
@@ -18,10 +18,5 @@
         "message": "Invalid array key type stdClass.",
         "line": 59,
         "ignorable": true
-    },
-    {
-        "message": "Offset stdClass does not exist on array{baz: 21}|array{foo: 17, bar: 19}.",
```
ondrejmirtes, line 23: We're losing these useful errors. What are they replaced with?

ondrejmirtes (changes requested):

tests/PHPStan/Levels/data/stringOffsetAccess-3.json
VincentLanglet: There is already (above) 
```
Invalid array key type stdClass.
```

ondrejmirtes:
Thank you!

PR #4280, opened 2025-09-03
Add non regression tests for 11761

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11761


ondrejmirtes:
Thank you.

PR #4281, opened 2025-09-03
Fix 7.4 tests

VincentLanglet:


tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php
```diff
@@ -101,6 +101,7 @@ public function testFirstClassCallables(): void
 		]);
 	}
 
+	#[RequiresPhp('>= 8.0')]
```
VincentLanglet, line 105: The sideEffect (exception thrown) was introduced in PHP8

tests/PHPStan/Analyser/nsrt/bug-13365.php
```diff
@@ -1,4 +1,4 @@
-<?php declare(strict_types = 1);
+<?php declare(strict_types = 1); // lint >= 8.0
```
VincentLanglet:
```
assertType('Iterator<mixed, DOMAttr>', $attributes->getIterator());
```
is not true before PHP 8.

PR #4282, opened 2025-09-03
Update ObjectShape native type

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11609

Explanations in https://github.com/phpstan/phpstan/issues/11609#issuecomment-3172138287

### DIFF
--- a/src/Type/ObjectShapePropertyReflection.php
+++ b/src/Type/ObjectShapePropertyReflection.php
@@ -61,12 +61,12 @@ final class ObjectShapePropertyReflection implements ExtendedPropertyReflection
 
 	public function hasNativeType(): bool
 	{
-		return false;
+		return true;
 	}
 
 	public function getNativeType(): Type
 	{
-		return new MixedType();
+		return $this->type;
 	}
 
 	public function getReadableType(): Type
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Nope, it's wrong. The IssetCheck should look at whether the certainty of a property is "maybe", not whether it has a native type or not.

VincentLanglet:
> Nope, it's wrong. The IssetCheck should look at whether the certainty of a property is "maybe", not whether it has a native type or not.

I'm not fixing
```
if (isset($a->world)) { // shouldn't report any error
```
but the null check
```
if ($a->hello !== null) { // should report notIdentical.alwaysTrue 
```

There is nothing related to IssetCheck.

`StrictComparisonOfDifferentTypesRule` is calling
```
$this->richerScopeGetTypeHelper->getIdenticalResult($this->treatPhpDocTypesAsCertain ? $scope : $scope->doNotTreatPhpDocTypesAsCertain(), $node);
```
which does 
```
if (
			(
				$expr->left instanceof Node\Expr\PropertyFetch
				|| $expr->left instanceof Node\Expr\StaticPropertyFetch
			)
			&& $rightType->isNull()->yes()
			&& !$scope->hasPropertyNativeType($expr->left)
		) {
			return new TypeResult(new BooleanType(), []);
		}
```

So when you check `$object->foo !== null` it will not considered as alwaysTrue when `foo` has no native type, because code might be
```
class MyObject 
{
     /** @var string */
     public $foo; // This is initialized with NULL since no native type is used.
}
```

ondrejmirtes:
I'm working on a better fix. Stay tuned 😊

PR #4283, opened 2025-09-04
Fix `RedisCluster::multi()` return type

sayuprc:
Fixed `RedisCluster::multi()` incorrectly returned `Redis`.

While fixing this issue, I noticed that some other return types in `RedisCluster` differ from the latest official stubs.
Should I open a separate PR to address these discrepancies?

resolve phpstan/phpstan#13392

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -1524,6 +1524,12 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug13392(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/nsrt/bug-13392.php');
+		$this->assertNoErrors($errors);
+	}
+
 	/**
 	 * @param string[]|null $allAnalysedFiles
 	 * @return Error[]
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Rules/Debug/data/bug-13392.php
```diff
@@ -0,0 +1,19 @@
...
+	public function get(): RedisCluster;
+}
+
+function func(Client $client): void
+{
+	$redisCluster = $client->get();
+	\PHPStan\dumpType($redisCluster);
+
+	$transaction = $redisCluster->multi();
+	\PHPStan\dumpType($transaction);
```
ondrejmirtes: Instead of this, use `assertType` function and put this under tests/PHPStan/Analyser/nsrt. That's the usual place for type inference tests.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Debug/data/bug-13392.php
sayuprc: I fixed in ec5642d.

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1524,6 +1524,12 @@ public function testBug13310(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug13392(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/nsrt/bug-13392.php');
+		$this->assertNoErrors($errors);
```
ondrejmirtes: No need to have this here. The file will be picked up automatically by NodeScopeResolverTest.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #4285, opened 2025-09-04
fix: check magic methods on final classes for allow dynamic properties

calebdw:
Hello!

This closes https://github.com/phpstan/phpstan/issues/13450

Thanks!

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -410,33 +410,6 @@ final class ClassReflection
 			return true;
 		}
 
-		if ($this->isReadOnly()) {
-			return false;
-		}
-
-		if (UniversalObjectCratesClassReflectionExtension::isUniversalObjectCrate(
-			$this->reflectionProvider,
-			$this,
-		)) {
-			return true;
-		}
-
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -285,6 +285,9 @@ jobs:
           - script: |
               cd e2e/bug-10483
               ../../bin/phpstan
+          - script: |
```
ondrejmirtes: There's nothing justifying this to be an E2E test. Just put the file in tests/PHPStan/Analyser/nsrt.

ondrejmirtes (changes requested):

.github/workflows/e2e-tests.yml
calebdw: Done!

ondrejmirtes (changes requested):
Readonly class can have magic methods (https://3v4l.org/PNmbe) so that check should be moved above `$class = $this`.

calebdw:
Good catch, I also added a test for this

ondrejmirtes:
Thank you.

PR #4289, opened 2025-09-05
Add ConstFetchNode support in array shape

VincentLanglet:
Follow https://github.com/phpstan/phpdoc-parser/pull/275

Closes https://github.com/phpstan/phpstan/issues/5773
and
Closes https://github.com/phpstan/phpstan/issues/6989

### DIFF
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -20,6 +20,7 @@ use PHPStan\PhpDocParser\Ast\ConstExpr\ConstExprNullNode;
 use PHPStan\PhpDocParser\Ast\ConstExpr\ConstExprStringNode;
 use PHPStan\PhpDocParser\Ast\ConstExpr\ConstExprTrueNode;
 use PHPStan\PhpDocParser\Ast\ConstExpr\ConstFetchNode;
+use PHPStan\PhpDocParser\Ast\Type\ArrayShapeItemNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayShapeNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\CallableTypeNode;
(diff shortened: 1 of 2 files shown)


ondrejmirtes (changes requested):
I don't want this huge refactoring. I don't think it makes sense to support `Foo::*` in an array shape key, because the result would no longer be ConstantArrayType, but ArrayType only.

The handling of `Foo::BAR` in an array shape key needs new code, not reusing of an existing code. For example supporting enums does not make sense there either.

Also, I realized that support for `static::` is currently insufficient. It's not going to do what people expect in subclasses. But that's a story for another day.

VincentLanglet:
> I don't want this huge refactoring. I don't think it makes sense to support `Foo::*` in an array shape key, because the result would no longer be ConstantArrayType, but ArrayType only.
> 
> The handling of `Foo::BAR` in an array shape key needs new code, not reusing of an existing code. For example supporting enums does not make sense there either.
> 
> Also, I realized that support for `static::` is currently insufficient. It's not going to do what people expect in subclasses. But that's a story for another day.

Ok so I tried another way.

I still created a method `resolveArrayShapeOffsetType` because there is a lot of early return in the ConstFetchNode handling code.

Then I used a simpler version for the ConstFetchNode by
- removing `*` support
- removing Enum support

src/PhpDoc/TypeNodeResolver.php
```diff
@@ -1081,6 +1073,69 @@ private function resolveArrayShapeNode(ArrayShapeNode $typeNode, NameScope $name
...
+							}
+
+							$className = $classReflection->getParentClass()->getName();
+						}
+						break;
+				}
+			}
+
+			if (!isset($className)) {
+				$className = $nameScope->resolveStringName($constExpr->className);
```
ondrejmirtes, line 1112: I think this is a bit undertested - all these branches - self/static/parent/normal name should have a test in `tests/PHPStan/Analyser/nsrt/bug-6989.php`.

ondrejmirtes (changes requested):

src/PhpDoc/TypeNodeResolver.php
VincentLanglet: I added a test for all the possible case I think.

ondrejmirtes:
Thank you.

PR #4290, opened 2025-09-05
feat: add dynamic parameter type extensions

calebdw:
Closes https://github.com/phpstan/phpstan/issues/11707, closes https://github.com/phpstan/phpstan/issues/12585

Supersedes https://github.com/phpstan/phpstan-src/pull/3828, supersedes https://github.com/phpstan/phpstan-src/pull/3131, supersedes https://github.com/phpstan/phpstan-src/pull/3823

Hello!

This adds generalized dynamic parameter type extensions and deprecates the parameter closure type extensions per https://github.com/phpstan/phpstan/issues/11707#issuecomment-2358932350.

This also fixes the `return.type` and the `argument.type` errors described in https://github.com/phpstan/phpstan/issues/12585 when the parameter type is overridden via the extension.

> [!NOTE]
> This currently only works for array and closure arguments, if support is desired for other argument types in the future then that will have to be added.

CC: @canvural, @Neol3108

Thanks!


canvural:
I tested it by porting my previous implementation for Larastan to this one, and looks like it works fine! Will try to implement more use cases and see if I can find any bugs. Also will try to test it on real projects.But so far so good. Thanks for this!

I'll also try to review this PR (though Ondrej would do a better job 😄 )

e2e/parameter-type-extension/src/ParameterTypeExtension.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App;
```
ondrejmirtes, line 5: I'm not sure what the e2e/parameter-type-extension directory is for. These directories are supposed to execute something in `.github/workflows/e2e-tests.yml` but this one does not. Either move this under `e2e/bug-12585/` if it's used there, or delete it.

src/Analyser/ExpressionContext.php
```diff
@@ -12,18 +12,19 @@ private function __construct(
 		private ?string $inAssignRightSideVariableName,
 		private ?Type $inAssignRightSideType,
 		private ?Type $inAssignRightSideNativeType,
+		private ?Type $overriddenType = null,
```
ondrejmirtes: Does not have to be optional, this is not under BC promise.

src/Analyser/ExpressionContext.php
```diff
@@ -60,4 +61,14 @@ public function getInAssignRightSideNativeType(): ?Type
 		return $this->inAssignRightSideNativeType;
 	}
 
+	public function withOverriddenType(?Type $type): self
+	{
+		return new self($this->isDeep, $this->inAssignRightSideVariableName, $this->inAssignRightSideType, $this->inAssignRightSideNativeType, $type);
+	}
+
+	public function getOverriddenType(): ?Type
```
ondrejmirtes: Why this does have to be here actually at all? Why hasn't been the `passedToType` parameter kept? Why is this named "overriden type"? What is it overriding?

ondrejmirtes (changes requested):
This looks solid, just a few questions as a start, will do deep review later 👍

canvural:
@calebdw @ondrejmirtes Sorry to bother you, but I wanted to ask how we can move this forward? I'm really interested in this feature.

calebdw:
No worries, just been busy with personal life, I'll try to get to the comments soon

src/Analyser/ExpressionContext.php
calebdw: I made it optional because I didn't want to go update all the usages :sweat_smile:. However, I have made non-api methods required and added them.

e2e/parameter-type-extension/src/ParameterTypeExtension.php
calebdw: I've added this to the github workflow

src/Analyser/ExpressionContext.php
calebdw: I've refactored this PR to simply pass the variable around through the methods instead of trying to use the `ExpressionContext`.

> Why hasn't been the passedToType parameter kept? Why is this named "overriden type"?

It has been kept, I just renamed `$passedToType` to `$overriddenType`---"passed to type" didn't make any sense so I renamed it.

> What is it overriding?

This is the overriding type coming from the dynamic parameter type extensions. The extensions are "overriding" the type that phpstan thinks they should be. If you have a better name in mind then I'm all ears.

calebdw:
@ondrejmirtes, @canvural, sorry it took so long to address these comments---just been busy with personal life but hoping to move this forward now :pray:

CamKemBell:
@calebdw - is this still planned on going ahead? Does this fix the issue with using whereHas with custom builders?

I've seen multiple projects now look to using custom builders since the implementation of the attribute & official support from Laravel last year, only to find issues when trying to type the arguments in eloquent method for type hints / auto complete, etc.

https://laravel-news.com/defining-a-dedicated-query-builder-in-laravel-12-with-php-attributes

In the meantime, we have been using whereRelation, which compiles to whereHas under the hood, but it would be nice to use whereHas directly.

image

dnyg:
My team are also eagerly awaiting the outcome of this PR! Hope it can be pushed forward 🙏

calebdw:
@CamKemBell, yes I just rebased and resolved all the conflicts

@ondrejmirtes, this is ready to review :+1:

dnyg:
What is the status on this PR?

VincentLanglet:
Hi @calebdw sorry for the delay there is now more review for the PHPStan codebase.

Are you still interested by the PR ?
There is a lot of conflict, can you solve them please ? 

Thanks !

calebdw:
@VincentLanglet, yes, the conflicts have been resolved!

e2e/parameter-type-extension/src/ParameterTypeExtension.php
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		foreach ($type->getKeyTypes() as $keyType) {
+			$relationType = $this->getRelationTypeFromModel($model, (string) $keyType->getValue(), $scope);
+			if (!$relationType) {
+				continue;
+			}
+
+			$newType = new ClosureType([
+				/** @phpstan-ignore phpstanApi.constructor */
```
VincentLanglet: I assume the error is reported for a valid reason.

I feel like it would be annoying to update this test every time the NativeParameterReflection constructor is modified. So it would be great to
- Write the code differently to avoid this error
- Or to convince ondrej to add `@api` to the constructor

src/Analyser/NodeScopeResolver.php
```diff
@@ -2592,6 +2594,7 @@ public function processExprNode(
 		ExpressionResultStorage $storage,
 		callable $nodeCallback,
 		ExpressionContext $context,
+		?Type $overriddenType,
```
VincentLanglet, line 2771: What about 
```suggestion
		?Type $overriddenType = null,
```
to avoid passing null every time ? (also a way to avoid conflict)

calebdw: I'm not opposed to this, but sounds like Ondre prefers to explicitly pass parameters (even null):

image

calebdw: @VincentLanglet, which direction should I go here?

VincentLanglet: Ondrej always have the last word ^^

VincentLanglet:
Not sure if all the failure are related to the pr ; but there is more failure than on other PR.
You should try fix level test + test with php 7.4 (and maybe reduce the number of escaped mutant)

calebdw:
@VincentLanglet,

- I've fixed the PHP 7.4 tests
- most of the integration tests are failing due to the deprecation warning from the old interfaces---not sure there's any way to fix these besides the respective projects updating their baselines after this has been released 
- the Level test failures appear unrelated
- I'm not too sure how to prevent the escape mutants, some of these would already be skipped (e.g., if an array is not a constant array) so having a test case for every mutant is tricky

phpstan-baseline.neon
```diff
@@ -1611,6 +1725,12 @@ parameters:
 			count: 1
 			path: src/Type/Php/NumberFormatFunctionDynamicReturnTypeExtension.php
 
+		-
+			rawMessage: Class PHPStan\Type\Php\PregReplaceCallbackClosureTypeExtension implements deprecated interface PHPStan\Type\FunctionParameterClosureTypeExtension.
```
VincentLanglet: Shouldn't PregReplaceCallbackClosureTypeExtension implements the new interface then ?

VincentLanglet (commented):
- Please rebase for the conflict

- This will need a PR on https://github.com/phpstan/phpstan-nette after to use the new non-deprecated interface (should we wait before deprecating the old one (?))

- The `integration-tests / Integration - shipmonk-rnd/dead-code-detector` failure seems not normal https://github.com/phpstan/phpstan-src/actions/runs/22313350451/job/64551526195?pr=4290

ondrejmirtes:
What is this one waiting for? 😊

VincentLanglet:
I think it's mainly conflict and targeting 2.2.x

ondrejmirtes:
@VincentLanglet @staabm Once you give a green light and you're happy with this PR, I will take over, take a look and bring it over the finish line 😊

src/Type/DynamicFunctionParameterTypeExtension.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace PHPStan\Type;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParameterReflection;
+
+/**
+ * This is the interface for dynamic parameter type extensions for functions.
```
staabm, line 12: I am looking at this for the first time, and my initial question is:

where to draw the line between `FunctionTypeSpecifyingExtension`, `FunctionParameterOutTypeExtension`, `DynamicFunctionParameterTypeExtension`? when to use which?
(same for Method and StaticMethod ofc)

all of the above can specify/narrow/change types of funtion calls (or function call parameters).

I feel extension implementators will have this question and after we properly discussed the differences, it might make sense to document these

canvural: This is more of a documentation comment right? Not blocking this PR.

staabm: its not blocking, but I am interessted in the anwser. it would help me to get a better understanding for the review

canvural: FunctionTypeSpecifyingExtension is for narrowing the types after a function call. We do not influence the type itself inside that function. And here we have access to the context so we know if we are in a `if` statement.

FunctionParameterOutTypeExtension I always thought for parameters that are passed by reference, but documentation gives an example without reference 😄 So I guess, it works for normal parameters too. But again we specify how the passed variables type changes after the function call.

DynamicFunctionParameterTypeExtension is when you want the variable type changed for that function body. So it'll be analyzed by PHPStan with the changed type. So compared to others this affects the types for the function body itself. Not after the call. And currently it only works for arrays and closures.


These are my understandings. Hope it's clear and I'm not wrong 😄

canvural: Is everything clear here? 👀 

Sorry, if I come up as very pushy but I really want this to move forward. Especially when Ondrej said he can take it over the finish line after approval from other maintainers.

src/Analyser/ExprHandler/ArrayHandler.php
```diff
@@ -66,26 +67,40 @@ public function resolveType(MutatingScope $scope, Expr $expr): Type
...
-			$valueResult = $nodeScopeResolver->processExprNode($stmt, $arrayItem->value, $scope, $storage, $nodeCallback, $context->enterDeep());
+			$overriddenValueType = null;
+			if ($overriddenType !== null && $overriddenType->hasOffsetValueType($keyType)->yes()) {
+				$overriddenValueType = $overriddenType->getOffsetValueType($keyType);
+			}
+
+			if ($arrayItem->key === null) {
+				$nextAutoIndex++;
+			} elseif ($keyType instanceof ConstantIntegerType) {
+				$nextAutoIndex = $keyType->getValue() + 1;
```
staabm, line 107: could keyType be a integer range here?
or e.g. `null|4`

e2e/bug-12585/src/test.php
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/** @extends Builder<Post> */
+class PostBuilder extends Builder
+{
+}
+
+function test(): void
+{
+	User::query()->whereHas('posts', function ($query) {
```
staabm, line 150: we should also have a test with
- nullsafe-method-call
- the variable beeing nullOrObject (a union)
- for "new"
- for static calls

staabm (commented):
I think it looks mostly good, a few questions
- if I did not miss it, it seems none of the tests is covering a error case?
- $overriddenType is passed thru a lot of handlers. I think we should have a few more tests which show that the overridden type gets correctly passed thru complex expressions
- I wonder if we really need end-to-end test-cases only or whether we could at least have some in-phpunit suite tests (these are easier to run, debug and maintain)
- you should look into issue-bot results and judge whether these are expected/related (some of them might go-away after rebase)
- we need this PR to target 2.2.x

canvural:
@calebdw Would you have time to address the comments? Or if you allow, I can open another PR with your code and fix there.

calebdw:
Yes, sorry been busy---I'll get to this shortly

canvural:
Tested the latest PHAR, still works alright 👍🏽 Static analysis fails on CI looks legit though. But didn't happen at runtime 🤔 Maybe didn't hit that code path.

.github/workflows/e2e-tests.yml
```diff
@@ -1,3 +1,4 @@
+---
```
staabm: typo

CamKemBell:
Would love to see this over the line...

canvural:
Can I sponsor/pay someone to bring this over the finish line? Resolving conflicts <--> wait for review loop doesn't seem to end 😄 

// @ondrejmirtes @staabm @VincentLanglet

src/Analyser/ExprHandler/ArrayHandler.php
calebdw: The `instanceof ConstantIntegerType` on line 106 is only used for tracking `$nextAutoIndex` (PHP's auto-incrementing array key behavior). It's not related to the `$overriddenType` lookup -- that happens on line 100 using `$keyType` directly via `hasOffsetValueType()`, which handles any type correctly.

Since this processes literal `Expr\Array_` nodes, the keys are typically constant values. But even if `$keyType` were a union or range, the fallback behavior (not updating `$nextAutoIndex`) is safe -- it just means subsequent auto-indexed items won't have their index tracked precisely, which is the same behavior as the existing upstream code in `MutatingScope`.

calebdw:
@staabm, 

> if I did not miss it, it seems none of the tests is covering a error case?

I've added tests

> $overriddenType is passed thru a lot of handlers. I think we should have a few more tests which show that the overridden type gets correctly passed thru complex expressions

Same here, added tests

> I wonder if we really need end-to-end test-cases only or whether we could at least have some in-phpunit suite tests (these are easier to run, debug and maintain)

The existing `DynamicParameterTypeExtensionClosuresTest` and `DynamicParameterTypeExtensionArraysTest` are already PHPUnit `TypeInferenceTestCase` tests (not just e2e). The e2e tests are supplementary for integration validation.

calebdw:
@ondrejmirtes, @staabm, @VincentLanglet, I've rebased this PR and resolved the conflicts

Can we try to get this over the line :pray:? Let me know if there's something we need to discuss, explain, or revisit implementation details. I'm always willing to hop on a call as well if that helps

canvural:
Hi,

To update, I solved the main pain point on Larastan side with custom PHPDoc type and a small MethodParameterClosureTypeExtension. Very small bit is left uncovered but https://github.com/phpstan/phpstan-src/pull/6417 (or an equivalent) would fix it.

So I no longer need this solution. Our solution at Larastan is a bit more general in a sense that the things that allowed me to achieve this will be used somewhere else in the code base too. So it wasn't only focused on this thing, but it's a very happy side effect 😄 That means Larastan _could_ use this extension if it is available, but it wouldn't strictly need it cause it'd not replace any existing code.

So I'll leave the decision to maintainers to see if the extension could benefit other developers.

ondrejmirtes:
@CamKemBell @dnyg @calebdw Please let me know if you wanted this PR because of Larastan issues (that are now solved) or if you still need it for some other purpose. Thank you!

calebdw:
@ondrejmirtes, this pr still has value because it's a more generalized form of the ClosureParameter extensions that work for any parameter

dnyg:
> @CamKemBell @dnyg @calebdw Please let me know if you wanted this PR because of Larastan issues (that are now solved) or if you still need it for some other purpose. Thank you!

The issue we were facing was already resolved in a later version of Laravel - so we're no longer eagerly awaiting this 🙌

PR #4293, opened 2025-09-07
Fix lost type when assigning variable in method argument

staabm:
closes https://github.com/phpstan/phpstan/issues/12234
closes https://github.com/phpstan/phpstan/issues/12735

### DIFF
--- a/src/Rules/FunctionCallParametersCheck.php
+++ b/src/Rules/FunctionCallParametersCheck.php
@@ -100,6 +100,10 @@ final class FunctionCallParametersCheck
 			$functionParametersMaxCount = -1;
 		}
 
+		if (!$scope instanceof MutatingScope) {
+			throw new ShouldNotHappenException();
+		}
+
 		/** @var array<int, array{Expr, Type|null, bool, string|null, int}> $arguments */
 		$arguments = [];
 		/** @var array<int, Node\Arg> $args */
@@ -112,6 +116,10 @@ final class FunctionCallParametersCheck
 			if ($arg->name !== null) {
 				$hasNamedArguments = true;
 				$argumentName = $arg->name->toString();
+			}
+
+			if ($arg->value instanceof Expr\Assign) {
+				$scope = $scope->assignExpression($arg->value->var, $scope->getType($arg->value->expr), $scope->getNativeType($arg->value->expr));
 			}
 
 			if ($hasNamedArguments && $arg->unpack) {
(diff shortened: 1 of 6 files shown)


ondrejmirtes (changes requested):
I don't like this fix, it's very one-off. There are many things in expressions that aren't assigns, that would still be broken.

As I said, this either needs virtual nodes, or ambitious refactoring.

staabm:
> As I said, this either needs virtual nodes

so you mean e.g. a `CallLikeArg` virtual node so we would hook a rule onto every arg and rework all callers of `FunctionCallParametersCheck::check`?

staabm:
> ambitious refactoring.

this would mean something like a new `MutatingScope->enterCallLike()` method, right?

ondrejmirtes:
The args need to be considered together. Looking at this the problem goes really deep and it might not be so easy to fix. Looking at places where FunctionCallParametersCheck is called, ParametersAcceptorSelector is called:

```php
ParametersAcceptorSelector::selectFromArgs(
	$scope,
	$node->getArgs(),
	$constructorReflection->getVariants(),
	$constructorReflection->getNamedArgumentsVariants(),
)
```

The considered types inside are going to be wrong too because we're only passing along a single scope.

And of course this method is one of the most used methods across the codebase. Changing that would have ripple effects across the ecosystem so we can't do that now.

Also, `FunctionCallParametersCheck::check()` accepts a single scope and `$funcCall` where it gets its arguments. This might be easier to fix. We need a new virtual node for each of `Node\Expr\FuncCall|Node\Expr\MethodCall|Node\Expr\StaticCall|Node\Expr\New_ $funcCall`. And inside of them we need `$args` to be a new virtual node array, something like `Argument[]`. The Argument would contain Scope too.

The proper Argument virtual nodes objects would need to be made inside `NodeScopeResolver::processArgs()`.

ondrejmirtes:
So my conclusion is that we can fix `FunctionCallParametersCheck::check()` now and it might work in most of the cases.

If people do assignment shenanigans and they have wrong generics resolved because of that, it's not so easy to fix.

ondrejmirtes:
> this would mean something like a new MutatingScope->enterCallLike() method, right?

No, I meant the really ambitious refactoring that I described here https://github.com/phpstan/phpstan-src/pull/4237#issuecomment-3262472961 and that I don't know how the code should look like after yet 😅

staabm:
> Also, FunctionCallParametersCheck::check() accepts a single scope and $funcCall where it gets its arguments. This might be easier to fix. We need a new virtual node for each of Node\Expr\FuncCall|Node\Expr\MethodCall|Node\Expr\StaticCall|Node\Expr\New_ $funcCall. And inside of them we need $args to be a new virtual node array, something like Argument[]. The Argument would contain Scope too.
> 
> The proper Argument virtual nodes objects would need to be made inside NodeScopeResolver::processArgs().

While this makes sense it sounds to be a big undertaking to support things I don't like to use at all myself (assigns in expressions).

I prefer to spend my time somewhere on something more valuable

ondrejmirtes:
Yeah, sure 👍 

Keeping this open because the tests are valuable.

ondrejmirtes:
This PR references already closed issues.

PR #4294, opened 2025-09-07
Fix incorrect analyzing of array_shift with non-empty-list property

staabm:
closes https://github.com/phpstan/phpstan/issues/13438
closes https://github.com/phpstan/phpstan/issues/2888
closes https://github.com/phpstan/phpstan/issues/3387
closes https://github.com/phpstan/phpstan/issues/2560
closes https://github.com/phpstan/phpstan/issues/2457
closes https://github.com/phpstan/phpstan/issues/11846

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2664,9 +2664,10 @@ final class NodeScopeResolver
 				$arrayArgNativeType = $scope->getNativeType($arrayArg);
 
 				$isArrayPop = $functionReflection->getName() === 'array_pop';
+				$newType = $isArrayPop ? $arrayArgType->popArray() : $arrayArgType->shiftArray();
 				$scope = $scope->invalidateExpression($arrayArg)->assignExpression(
 					$arrayArg,
-					$isArrayPop ? $arrayArgType->popArray() : $arrayArgType->shiftArray(),
+					$newType,
 					$isArrayPop ? $arrayArgNativeType->popArray() : $arrayArgNativeType->shiftArray(),
 				);
 
@@ -2674,7 +2675,7 @@ final class NodeScopeResolver
 					$scope,
 					$stmt,
 					$arrayArg,
-					$arrayArg,
+					new TypeExpr($newType),
 					static function (Node $node, Scope $scope) use ($nodeCallback): void {
 						if (!$node instanceof PropertyAssignNode && !$node instanceof VariableAssignNode) {
 							return;
@@ -2703,7 +2704,7 @@ final class NodeScopeResolver
 					$scope,
 					$stmt,
 					$arrayArg,
-					$arrayArg,
+					new TypeExpr($arrayType),
 					static function (Node $node, Scope $scope) use ($nodeCallback): void {
 						if (!$node instanceof PropertyAssignNode && !$node instanceof VariableAssignNode) {
 							return;
@@ -2729,9 +2730,10 @@ final class NodeScopeResolver
 				&& $functionReflection->getName() === 'shuffle'
 			) {
 				$arrayArg = $expr->getArgs()[0]->value;
+				$newType = $scope->getType($arrayArg)->shuffleArray();
 				$scope = $scope->assignExpression(
 					$arrayArg,
-					$scope->getType($arrayArg)->shuffleArray(),
+					$newType,
 					$scope->getNativeType($arrayArg)->shuffleArray(),
 				);
 
@@ -2739,7 +2741,7 @@ final class NodeScopeResolver
 					$scope,
 					$stmt,
 					$arrayArg,
-					$arrayArg,
+					new TypeExpr($newType),
 					static function (Node $node, Scope $scope) use ($nodeCallback): void {
 						if (!$node instanceof PropertyAssignNode && !$node instanceof VariableAssignNode) {
 							return;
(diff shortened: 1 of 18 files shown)


ondrejmirtes (changes requested):
1) I'm worried that other virtual nodes still not fire for this, like VariableAssignNode.
2) This is still not going to work on `$this->foo[$x]` for example. This will still have a false positive: https://phpstan.org/r/a6e895c1-0e8e-4e49-8cd9-e7a51655a506

Correct solution in these places is to construct proper `$this->processAssignVar` call. Which might be a bit complicated to get right but here's the easiest inspiration probably:

https://github.com/phpstan/phpstan-src/blob/02116b7405309ece5f6ff6ba437a4c6834835789/src/Analyser/NodeScopeResolver.php#L5382-L5397

src/Analyser/NodeScopeResolver.php
```diff
@@ -2669,6 +2669,23 @@ static function (): void {
 					$isArrayPop ? $arrayArgType->popArray() : $arrayArgType->shiftArray(),
 					$isArrayPop ? $arrayArgNativeType->popArray() : $arrayArgNativeType->shiftArray(),
 				);
+
```
staabm, line 2673: I have also tried removing aboves `$scope->invalidateExpression($arrayArg)->assignExpression($arrayArg, ...)` but this leads to different native vs. phpdoc type results

```diff
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2659,22 +2659,14 @@ final class NodeScopeResolver
                                && count($expr->getArgs()) >= 1
                        ) {
                                $arrayArg = $expr->getArgs()[0]->value;
-
                                $arrayArgType = $scope->getType($arrayArg);
-                               $arrayArgNativeType = $scope->getNativeType($arrayArg);
-
                                $isArrayPop = $functionReflection->getName() === 'array_pop';
-                               $scope = $scope->invalidateExpression($arrayArg)->assignExpression(
-                                       $arrayArg,
-                                       $isArrayPop ? $arrayArgType->popArray() : $arrayArgType->shiftArray(),
-                                       $isArrayPop ? $arrayArgNativeType->popArray() : $arrayArgNativeType->shiftArray(),
-                               );
 
                                $scope = $this->processAssignVar(
                                        $scope,
                                        $stmt,
                                        $arrayArg,
-                                       $arrayArg,
+                                       new TypeExpr($isArrayPop ? $arrayArgType->popArray() : $arrayArgType->shiftArray()),
                                        static function (Node $node, Scope $scope) use ($nodeCallback): void {
                                                if (!$node instanceof PropertyAssignNode && !$node instanceof VariableAssignNode) {
```

ondrejmirtes:
Issue bot jackpot 🎉 

Isn't https://github.com/phpstan/phpstan/issues/11846 also fixed?

staabm:
> Isn't [phpstan/phpstan#11846](https://github.com/phpstan/phpstan/issues/11846)

I wasn't sure, because the reporter wanted it to dump `true`.
but `bool` at least fixes the wrong `false` we had before.

so we are better but not perfect yet.

will add a test

src/Analyser/NodeScopeResolver.php
```diff
@@ -2669,6 +2669,23 @@ static function (): void {
...
+				$scope = $this->processAssignVar(
+					$scope,
+					$stmt,
+					$arrayArg,
+					$arrayArg,
```
ondrejmirtes: These arguments would mean that we're doing `$foo = $foo;` which is weird.

It might not influence anything but `$assignedExpr` should be `new TypeExpr` with the thing passed to 2nd parameter of `assignExpression`. In this case it's `$isArrayPop ? $arrayArgType->popArray() : $arrayArgType->shiftArray()` (but of course first save it into a variable).

src/Analyser/NodeScopeResolver.php
```diff
@@ -2681,6 +2698,23 @@ static function (): void {
 
 				$arrayArg = $expr->getArgs()[0]->value;
 				$scope = $scope->invalidateExpression($arrayArg)->assignExpression($arrayArg, $arrayType, $arrayNativeType);
+
+				$scope = $this->processAssignVar(
+					$scope,
+					$stmt,
+					$arrayArg,
+					$arrayArg,
```
ondrejmirtes: For example here it should be `new TypeExpr($arrayType)`

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
staabm: ok, so it seems you are fine with what I suggested in https://github.com/phpstan/phpstan-src/pull/4294#discussion_r2329530486 ?

src/Analyser/NodeScopeResolver.php
ondrejmirtes: I'm fine with that as a temporary step.

Maybe we can experiment with a new `TypeExpr`-like class in another PR that would carry `$type` and `$nativeType` - two constructor arguments. A bunch of places that currently do `new TypeExpr` might get better behaviour from that.

staabm: after doing https://github.com/phpstan/phpstan-src/pull/4294#discussion_r2329668928 we get the same type changes, as I saw with the above suggestion.

so I went ahead and did this simplification in https://github.com/phpstan/phpstan-src/pull/4294/commits/a53b7a940bb9948006a48947d46ab44d15c49057

if you are not fine with it, I can revert the changes (but it does not affect results of tests)


> Maybe we can experiment with a new `TypeExpr`-like class in another PR that would carry `$type` and `$nativeType` - two constructor arguments


agree :+1:

ondrejmirtes: I don't want this simplification. We'd have to put the code back anyway after we do the "TypeExpr with native type" again. Please revert it.

ondrejmirtes:
Perfect, thank you!

VincentLanglet:
Good job @staabm you're really good to close issues by batch !

PR #4295, opened 2025-09-07
non-empty-array might be empty after array_keys() with filter

staabm:
closes https://github.com/phpstan/phpstan/issues/13377


ondrejmirtes (changes requested):
AccessoryArrayListType might have the same problem, can you try to reproduce it and fix it? Thanks.

staabm:
> AccessoryArrayListType might have the same problem, can you try to reproduce it and fix it?

when using a filter `array_keys` still returns a list, see https://3v4l.org/LCT5q

ondrejmirtes:
Oh, of course 😊

ondrejmirtes:
Thank you!

PR #4302, opened 2025-09-08
Implemented NativeTypeExpr

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/4294#discussion_r2329702345 we add a new VirtualNode type which can carry a native-type in tandem with a phpdoc type.

before this PR, we only had `TypeExpr` which was not able to provide 2 types, but just a single one.

so this PR fixes a few shortcomings, which we accepted with a recent change in https://github.com/phpstan/phpstan-src/pull/4294/commits/4702c2b2e2ff562e887571b492a62c34fdcb41de


ondrejmirtes:
It seems like two more issues are getting fixed.

ondrejmirtes:
Thank you.

PR #4306, opened 2025-09-09
Added regression test

staabm:
cover a different variant of https://github.com/phpstan/phpstan/issues/13385 as the loop problem did already exist with simpler while conditions just using `$operators !== []` instead of `$operators !== [] && end($operators)->priority() >= $child->priority()`


ondrejmirtes:
Thank you.

PR #4307, opened 2025-09-09
Added regression test

staabm:
cover https://github.com/phpstan/phpstan/issues/13385 with different loop types

- outer loop with for() {} instead of foreach() {}
- inner loop with for() {} instead of while() {}

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-13385c.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-13385c.php
@@ -1,0 +1,51 @@
+<?php declare(strict_types = 1);
+
+namespace Bug13385c;
+
+use function PHPStan\Testing\assertType;
+
+interface Operator {
+	public function priority(): int;
+	public function calculate(int $a, int $b): int;
+}
+
+class HelloWorld
+{
+	/**
+	 * @param list<Operator|int> $children
+	 */
+	public function calculate(array $children): int {
+		$operands  = [];
+		$operators = [];
+
+		foreach ($children as $child) {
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/nsrt/bug-13385c.php
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * @param list<Operator|int> $children
+	 */
+	public function calculate(array $children): int {
+		$operands  = [];
+		$operators = [];
+
+		foreach ($children as $child) {
+			if ($child instanceof Operator) {
+				for(;$operators !== [] && end($operators)->priority() >= $child->priority();) {
```
ondrejmirtes, line 23: The test isn't faithful enough because there are no changes in NodeScopeResolver. And the same fix I did for While_ also needs to be done for For_ and many others.

tests/PHPStan/Analyser/nsrt/bug-13385d.php
```diff
@@ -0,0 +1,51 @@
...
+class HelloWorld
+{
+	/**
+	 * @param list<Operator|int> $children
+	 */
+	public function calculate(array $children): int {
+		$operands  = [];
+		$operators = [];
+
+		for ($i = 0; $i < count($children); $i++) {
```
ondrejmirtes, line 21: I don't think the type of outer loop matters. We don't need more than one test for the same change and reported bug in src/.

ondrejmirtes (changes requested):

PR #4308, opened 2025-09-09
Added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/7225


tests/PHPStan/Rules/Methods/ReturnTypeRuleTest.php
```diff
@@ -1259,4 +1259,9 @@ public function testBug12928(): void
 		]);
 	}
 
+	public function testBug7225(): void
```
ondrejmirtes, line 1263: Needs RequiresPhp >= 8.0

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4310, opened 2025-09-09
Added regression test

staabm:
fixed via https://github.com/Roave/BetterReflection/pull/1524

closes https://github.com/phpstan/phpstan/issues/13469


ondrejmirtes:
Thank you.

PR #4312, opened 2025-09-09
While loop-like fixes

ondrejmirtes:


staabm:
this loop business is so complicated, I tried find more things but its really hard.

one case which errors is
https://phpstan.org/r/d883e978-f368-4bf4-afae-8683f7f5ebd3

but not sure whether its a bug or intentional

ondrejmirtes:
I think in the linked example the for loop never executes which is reported by `Strict comparison using !== between array{} and array{} will always evaluate to false.`.

A closer example to the original one is https://phpstan.org/r/02d50b61-9e19-4826-a593-b08f44bd0a48 which behaves as expected.

I just realized most of the code works as it should (it doesn't merge scope from dead code) and I can revert most of the changes.

ondrejmirtes:
That's why there's no impact of the changes I made on open issues and real-world code.

PR #4314, opened 2025-09-10
Allow to cast resource to float

VincentLanglet:
Part of https://github.com/phpstan/phpstan/issues/6560
but not all, (I leave https://github.com/phpstan/phpstan/issues/6560#issuecomment-3274299888)

https://github.com/phpstan/phpstan-src/pull/4152 was wrong, since we can cast a resource to float the same way we can cast it to int.


ondrejmirtes:
Thank you!

PR #4313, opened 2025-09-09
Remember narrowed readonly types in anonymous functions

staabm:
closes https://github.com/phpstan/phpstan/issues/13321

requires https://github.com/phpstan/phpstan-src/pull/4315

PR #4315, opened 2025-09-10
Remember narrowed types from the constructor from deep PropertyFetches

staabm:
types which get narrowed in __construct() on [`readonly` properties can be useful when analyzing other methods](https://staabm.github.io/2025/04/15/phpstan-remember-constructor-types.html).

improve support initially added with https://github.com/phpstan/phpstan-src/pull/3930 by allowing property-fetch-chains, when all involved properties are readonly.

---

preparational work for https://github.com/phpstan/phpstan-src/pull/4313#discussion_r2335791627

commit1 extracts 1:1 a new method
commit2+3 adds support for property-fetch chains
commit4 re-use the new method for shouldInvalidateExpression()

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -315,23 +315,8 @@ final class MutatingScope implements Scope
 					continue;
 				}
 			} elseif ($expr instanceof PropertyFetch) {
-				if (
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #4316, opened 2025-09-10
Fix "Crash: str_increment(): Argument #1 ($string) must be composed only of alphanumeric ASCII characters"

staabm:
I thought about different variants on how this could be solved.
most important point IMO is we don't crash PHPStan on unexpected input.

I think it makes sense to report a *ERROR* type for PHP 8.3+ (so when `function_exists('str_increment')`; and [PHP started to emit deprecations](https://3v4l.org/jtTtD)) when invalid input is provided.

closes https://github.com/phpstan/phpstan/issues/13481


samsonasik:
`str_increment()` call exists in `StrIncrementDecrementFunctionReturnTypeExtension` as well

https://github.com/phpstan/phpstan-src/blob/e9556fb7d2425d889a3a6dd70b9c04db3329c407/src/Type/Php/StrIncrementDecrementFunctionReturnTypeExtension.php#L114

probably worth check there as well?

staabm:
The added test covers this line

src/Analyser/MutatingScope.php
```diff
@@ -1750,7 +1751,11 @@ static function (Node $node, Scope $scope) use ($arrowScope, &$arrowFunctionImpu
 							&& !is_numeric($varValue)
 							&& function_exists('str_increment')
 						) {
-							$varValue = str_increment($varValue);
+							try {
```
VincentLanglet: This feels kinda weird because
- On PHP 8.3 & 8.4 we return ErrorType while it's still valid code (even if deprecated)
- On PHP 8.5+ we're still returning `'1'` for `''++` while it's invalid code

Should we introduce a method inside `PHPVersion` class ?

Then you could keep `@++$varValue` for PHPVersion <= 8.5

staabm: this was one of the variants I considered. the ++$varValue is consistent across all versions (it differs only by emitting a deprecation warning).

if we are fine with swalling the deprecation warning the impl could be simplified.

I think we wouldn't even need per version expectations

ondrejmirtes (changes requested):
I wonder: how come we only use `str_increment` in our code but not `str_decrement` at all? I feel like places doing `--` possibly on strings should also have similar conditions around that code.

ondrejmirtes:
Forgot to link it - I tested it and `--` is deprecated too https://3v4l.org/q5Fp3/rfc#vgit.master

staabm:
what do you think about swallowing the deprecations as there is currently no need to use the str_increment/derecment functions..?
(and these are no good replacement)

ondrejmirtes:
What's wrong with calling them and catching ValueError?

staabm:
its "a lot of work" without a gain

it will also lead to different results depending on the used php runtime version

VincentLanglet:
> What's wrong with calling them and catching ValueError?

The only wrong thing I see is the fact that we get a ValueError on PHP 8.3/8.4 (and considered the result as ErrorType) while ++ and -- are supposed to be still valid (and working, even if deprecated).

ondrejmirtes:
Do you think this will work or am I missing something? https://github.com/phpstan/phpstan-src/commit/e4c5865d6650f7ae1cc092c30bd5fd7b2f336ef0

staabm:
> Do you think this will work or am I missing something?

as is I think it misses some cases, e.g. `++'ab c1'` (this PRs test)

ondrejmirtes:
I'll cherry pick the tests from here and we'll see. Right now I'm fixing the tests which I forgot to run before pushing the commit 🤦

ondrejmirtes:
See https://github.com/phpstan/phpstan-src/commit/392e58debbd30a81e9a5db76d2d69d0f48299844 - I know it's not perfect and it's a weird quirk that for some strings `++`/`--` keeps them the same but `str_increment` throws a ValueError, but I think I can live with this now.

PR #4317, opened 2025-09-10
bleedingEdge: use `rawMessage` in baseline generators

gharlan:
https://github.com/phpstan/phpstan-src/pull/4111#issuecomment-3276198676

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -11,3 +11,4 @@ parameters:
 		newStaticInAbstractClassStaticMethod: true
 		checkExtensionsForComparisonOperators: true
 		reportTooWideBool: true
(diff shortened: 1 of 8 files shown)


staabm:
Would be interesting whether a raw message based baseline is faster than a regular one, because we need less regex matching..?

Could you compare the runtime?

ondrejmirtes:
For huge baselines there has to be a difference for sure

ondrejmirtes:
Thank you!

staabm:
> Could you compare the runtime?

just ran `time make phpstan` before and after this PR on phpstan-src on macOS M4 Pro multiple times:

before: ~29-30s
after: ~27-28s

PR #4318, opened 2025-09-10
Treat Redis::connect as non-deterministic

vrana:
Fixes https://phpstan.org/r/1695567f-4869-428f-8cb6-fdd8cb147be1.

### DIFF
--- a/bin/functionMetadata_original.php
+++ b/bin/functionMetadata_original.php
@@ -170,6 +170,10 @@ return [
 	'DateTimeImmutable::getTimestamp' => ['hasSideEffects' => false],
 	'DateTimeImmutable::getTimezone' => ['hasSideEffects' => false],
 
+	// affects isConnected()
+	'Redis::connect' => ['hasSideEffects' => true],
+	'Redis::pconnect' => ['hasSideEffects' => true],
+
 	'SplDoublyLinkedList::pop' => ['hasSideEffects' => true],
 	'SplDoublyLinkedList::shift' => ['hasSideEffects' => true],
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
1) I'd say connect method has to change instead, that will take care of invalidating remembered isConnected when called.
2) You edited the correct file but you also need to run bin/ script to regenerate the metadats file that doesn't have "original" in its name.

If you're interested, you can check out this part of my talk about pure/impure and so on here: https://youtu.be/AFjr3RlDOZQ?si=dp_XJWfenGJBRQhp (from about 28:50 timestamp)

vrana:
Makes sense. Please suggest how can I test this.

staabm:
you can add a new `testXX` method into `tests/PHPStan/Rules/TooWideTypehints/TooWidePropertyTypeRuleTest.php`
and put your code from https://phpstan.org/r/1695567f-4869-428f-8cb6-fdd8cb147be1 into the tests `data/` directory.

then assert no errors

similar to https://github.com/phpstan/phpstan-src/blob/99efae39a5118af487858194dd66dcf4793ef900/tests/PHPStan/Rules/TooWideTypehints/TooWidePropertyTypeRuleTest.php#L104-L107

vrana:
I've added the test but I couldn't reproduce the original behavior before this change.

staabm:
Needs `$this->reportTooWideBool = true;` in the rule test class

vrana:
Thanks, hallelujah!

staabm:
don't get confused by the CI results.. [we currently investigate spurious CI errors](https://github.com/phpstan/phpstan/issues/13494)

ondrejmirtes:
Thank you!

PR #4319, opened 2025-09-11
MatchExpressionRule - ignore reportAlwaysTrueInLastCondition

vrana:
There are other cases where this rule is harmful, I show them in tests.

Doc: https://github.com/phpstan/phpstan/pull/13489


ondrejmirtes:
Hi, I've said in the past that `reportAlwaysTrueInLastCondition` has been added for completeness and to match how PHPStan behaved by default in the past, but personally I think no one should use it.

What's your reason to enable it? Where do you see the benefit?

vrana:
The example from the test seems reasonable to me:

```php
	public function doFoo(int $i, \stdClass $std)
	{
		if ($i) {
		} elseif ($std) { // error only with reportAlwaysTrueInLastCondition
		}
	}
```

ondrejmirtes:
Thank you.

PR #4320, opened 2025-09-11
Lock container creation

RobinGeuze:
This solves errors like these:
```
 Child process error (exit code 1): PHP Warning:  rename(/app/tmp/cache/phpstan/cache/nette.configurator/Container_37633cd6c1.php.tmp,/app/tmp/cache/phpstan/cache/nette.configurator/Container_37633cd6c1.php): No such file or   
     directory in phar:///app/vendor/phpstan/phpstan/phpstan.phar/vendor/nette/di/src/DI/ContainerLoader.php on line 68                                                                                                                
     Warning: rename(/app/tmp/cache/phpstan/cache/nette.configurator/Container_37633cd6c1.php.tmp,/app/tmp/cache/phpstan/cache/nette.configurator/Container_37633cd6c1.php): No such file or directory in                              
     phar:///app/vendor/phpstan/phpstan/phpstan.phar/vendor/nette/di/src/DI/ContainerLoader.php on line 68                                                                                                                             
                                                                                                                                                                                                                                       
     In ContainerLoader.php line 71:                                                                                                                                                                                                   
                                                                                                                                                                                                                                       
       Unable to create file '/app/tmp/cache/phpstan/cache/nette.configurator/Container_37633cd6c1.php'.                                                                                                                               
                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                       
     worker [-c|--configuration CONFIGURATION] [-l|--level LEVEL] [-a|--autoload-file AUTOLOAD-FILE] [--memory-limit MEMORY-LIMIT] [--xdebug] [--port PORT] [--identifier IDENTIFIER] [--tmp-file TMP-FILE] [--instead-of INSTEAD-OF]  
     [--] [<paths>...]                                                                                                                                                                                                                 
                                                                                                                                                                                                                                       
      while running parallel worker                 
```

This is caused by the journal code deleting the file (because it hasn't been added to the journal yet) while another process is creating it. Since the container php files are quite large the file_put_contents call lasts quite long, and an unlink will work while file_put_contents is running. This then causes the rename to fail since the file has been deleted.

### DIFF
--- a/src/DependencyInjection/Configurator.php
+++ b/src/DependencyInjection/Configurator.php
@@ -191,7 +191,7 @@ final class Configurator extends \Nette\Bootstrap\Configurator
 				continue;
 			}
 			$fileName = $fileInfo->getFilename();
-			if ($fileName === 'container.journal') {
+			if ($fileName === 'container.journal' || $fileName === 'container.lock') {
 				continue;
 			}
 			if (!str_ends_with($fileName, '.php')) {
--- a/src/DependencyInjection/Configurator.php
+++ b/src/DependencyInjection/Configurator.php
@@ -15,6 +15,9 @@ use function array_keys;
 use function count;
 use function error_reporting;
 use function explode;
+use function fclose;
+use function flock;
+use function fopen;
 use function implode;
 use function in_array;
 use function is_dir;
@@ -29,6 +32,8 @@ use function time;
 use function trim;
 use function unlink;
 use const E_USER_DEPRECATED;
+use const LOCK_EX;
+use const LOCK_UN;
 use const PHP_RELEASE_VERSION;
 use const PHP_VERSION_ID;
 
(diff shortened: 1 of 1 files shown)


staabm:
do you know steps how to reproduce the reported problem? would be great to have it documented for the future in case we need to look into similar things again.

RobinGeuze:
Unfortunately not, it is a race condition. Once it happens it seems to keep happening, but I've never been able to pinpoint precisely what causes it. It also seems to specifically happen when running phpstan within a docker desktop on mac container with the primary filesystem being a volume mount, but I've got no concrete proof that is required for the error to occur. The race condition in the code is quite obvious though.

ondrejmirtes:
I tried to fix this a different way: https://github.com/phpstan/phpstan-src/commit/8fd0fdac262bc9c28d8b975ca6f29cf238a5da8d

Please test it. You can require `phpstan/phpstan:2.1.x-dev` in your composer.json. Or wait for an imminent release.

RobinGeuze:
Hey Ondrej, I got the error once directly after upgrading to 2.1.x-dev@0693e5a, but haven't had it since, I'll keep monitoring it.

ondrejmirtes:
@RobinGeuze You can delete all containers on your disk and then try clearing the result cache and re-run PHPStan again.

Then you can list which container classes have been generated in your tmp directory. If there are two classes with similar size, please diff the contents and post them here. Those two containers are created by the parent process and worker processes. Which causes this issue. There should only be one container.

RobinGeuze:
I've closed the PR for now.

PR #4321, opened 2025-09-11
Improve idate() return types

vrana:
My first implementation also modified `DateFunctionReturnTypeHelper` to call `IdateFunctionReturnTypeHelper` but it was quite messy and it wasn't obvious what's coming from where.

### DIFF
--- a/src/Type/Php/IdateFunctionReturnTypeExtension.php
+++ b/src/Type/Php/IdateFunctionReturnTypeExtension.php
@@ -1,0 +1,42 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\Type;
+
+#[AutowiredService]
+final class IdateFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+{
+
+	public function __construct(private IdateFunctionReturnTypeHelper $idateFunctionReturnTypeHelper)
+	{
+	}
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return $functionReflection->getName() === 'idate';
+	}
+
+	public function getTypeFromFunctionCall(
+		FunctionReflection $functionReflection,
+		FuncCall $functionCall,
+		Scope $scope,
+	): ?Type
+	{
+		$args = $functionCall->getArgs();
+		if ($args === []) {
+			return new ConstantBooleanType(false);
+		}
+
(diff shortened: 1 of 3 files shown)


src/Type/Php/IdateFunctionReturnTypeHelper.php
```diff
@@ -0,0 +1,53 @@
...
+
+	public function getTypeFromFormatType(Type $formatType): Type
+	{
+		$types = [];
+		foreach ($formatType->getConstantStrings() as $formatString) {
+			$types[] = $this->buildReturnTypeFromFormat($formatString->getValue());
+		}
+
+		if ($types === []) {
+			$types = [new IntegerType(), new ConstantBooleanType(false)];
```
staabm: to fallback to the default format you could also `return null` here and turn
`IdateFunctionReturnTypeExtension->getTypeFromFunctionCall()` into returning `?Type` - so PHPStan would internally determine the default return type

src/Type/Php/IdateFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,42 @@
...
+	public function getTypeFromFunctionCall(
+		FunctionReflection $functionReflection,
+		FuncCall $functionCall,
+		Scope $scope,
+	): Type
+	{
+		$args = $functionCall->getArgs();
+		if ($args === []) {
+			return new ConstantBooleanType(false);
+		}
```
staabm, line 35: usually we `return null` when args are missing

vrana: In this case, we know that it will return false.

src/Type/Php/IdateFunctionReturnTypeHelper.php
```diff
@@ -0,0 +1,53 @@
...
+			return null;
+		}
+
+		return TypeCombinator::union(...$types);
+	}
+
+	public function buildReturnTypeFromFormat(string $formatString): Type
+	{
+		// see https://www.php.net/idate
+		return match ($formatString) {
```
ondrejmirtes: Unfortunately match expression isn't supported here. Although phpstan-src is written for PHP 8.2+ and the code is downgraded during the PHAR building to PHP 7.4+ with https://github.com/ondrejmirtes/simple-downgrader, it's still kind of pseudo-PHP 8.2+.

Please change this to a switch statement.

ondrejmirtes (changes requested):

src/Type/Php/IdateFunctionReturnTypeHelper.php
staabm: do we want a phpstan-src only rule to detect nodes which are not supported?
do you have in mind which node types are relevant?

ondrejmirtes: We don't need that, many existing CI workflows already point it out https://github.com/phpstan/phpstan-src/actions/runs/17640960108/job/50127724654?pr=4321

vrana: Done.

ondrejmirtes:
Thank you!

PR #4324, opened 2025-09-12
Add missing use statement for the AutowireService attribute

stof:
I forgot the use statement when rebasing https://github.com/phpstan/phpstan-src/pull/4043

The CI caught the mistake but the PR was merged with a failing CI.


ondrejmirtes:
Thank you.

PR #4325, opened 2025-09-12
Added regression test

staabm:
covers a snippet [mentioned in a comment](https://github.com/phpstan/phpstan/issues/7280#issuecomment-2447420232)

was fixed by https://github.com/phpstan/phpstan-src/pull/4168

the PR does not close the issue, therefore not using a "closing"-link


ondrejmirtes:
Thank you.

PR #4326, opened 2025-09-12
Issue-Bot: Emit GitHub Action annotation when issues are affected

staabm:
with this PR we get a issue-bot 'pull request annotation' in the 'files changed' tab, which looks like:

grafik

this annotation is only visible when issue bot finds affected issues, otherwise no annotation is showing up.
intentionally I used a message type "notice" to get a gentle hint which distracts as less as possible.

Goal: make it more obvious when issue bot has helpful results. before this PR we sometimes missed to check the issue bot results as they are hidden deep in the github.com UI

---

triggered by a discussion in https://github.com/phpstan/phpstan-src/pull/4168#issuecomment-3285074537

### DIFF
--- a/issue-bot/src/Console/EvaluateCommand.php
+++ b/issue-bot/src/Console/EvaluateCommand.php
@@ -33,9 +33,9 @@ use function unserialize;
 class EvaluateCommand extends Command
 {
 
-	const EXIT_ERROR = 1;
-	const EXIT_AFFECTS_ISSUES = 2;
-	const EXIT_NO_AFFECTED_ISSUES = 0;
+	private const EXIT_ERROR = 1;
+	private const EXIT_AFFECTS_ISSUES = 2;
+	private const EXIT_NO_AFFECTED_ISSUES = 0;
 
 	public function __construct(
 		private TabCreator $tabCreator,
--- a/issue-bot/src/Console/EvaluateCommand.php
+++ b/issue-bot/src/Console/EvaluateCommand.php
@@ -33,6 +33,10 @@ use function unserialize;
 class EvaluateCommand extends Command
 {
 
+	const EXIT_ERROR = 1;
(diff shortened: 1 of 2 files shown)


staabm:
@ondrejmirtes @VincentLanglet any opinion on this? is it acceptable?

if so, I would cleanup the PR and we can give it a try.
if it doesn't work we can remove the annotation just by removing the commit.. so it does not cost much to test it

issue-bot/src/Console/EvaluateCommand.php
```diff
@@ -133,9 +137,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			}
 		}
 
+		$exitCode = self::EXIT_AFFECTS_ISSUES;
 		if (count($toPost) === 0) {
+			$exitCode = self::EXIT_NO_AFFECTED_ISSUES;
 			$output->writeln(sprintf('No changes in results in %d code snippets from %d GitHub issues. :tada:', $totalCodeSnippets, count($issueCache->getIssues())));
 		}
+		// XXX debug  test
+		$exitCode = self::EXIT_AFFECTS_ISSUES;
```
staabm: todo: remove this debug line, which atm enforce rendering of the github action annotation message

VincentLanglet:
Personally I like it

ondrejmirtes:
I like it too!

The reason why it's not safe to post a comment is because the only event that has access to secrets is `pull_request_target`. And you can't check out the repository if you want to be safe. There's a lot of attacks possible if you do that. There's a big warning here: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

staabm:
ok cool - lets give it a try then

ondrejmirtes:
Awesome, thank you!

PR #4327, opened 2025-09-12
Update stream_get_meta_data return

miken32:
Per [documentation](https://www.php.net/manual/en/function.stream-get-meta-data.php#refsect1-function.stream-get-meta-data-returnvalues), the `crypto` index is added for tls://, https://, and ftps:// streams. Indices of this array aren't documented, but they're [in the source](https://github.com/php/php-src/blob/e6c6333c8e7c2f247e74d45e8b3ab442d6db3ebc/ext/openssl/xp_ssl.c#L2326)

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -10360,7 +10360,7 @@ return [
 'stream_get_contents' => ['__benevolent<string|false>', 'source'=>'resource', 'maxlen='=>'int', 'offset='=>'int'],
 'stream_get_filters' => ['list<string>'],
 'stream_get_line' => ['string|false', 'stream'=>'resource', 'maxlen'=>'int', 'ending='=>'string'],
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

miken32:
You're welcome! I'm still having problems though. It seems like this signature is being ignored; the previous changes from 29c440e are also not showing. https://phpstan.org/r/5dd5aaef-80d2-47f1-bede-2956396ebc0b

miken32:
Ok just found this: https://github.com/phpstan/phpstan-src/blob/2.1.x/resources/functionMap_php83delta.php

It seems to be overriding the definitions for some reason; the commit message for 885d46b says `uri` is always there in 8.3, but [it's not](https://github.com/php/php-src/blob/php-8.4.12/ext/standard/streamsfuncs.c#L581).

And I'm not sure what version it was, but this `crypto` key was added in https://github.com/php/php-src/commit/676a5059b215c2d725ed219d4258d8221759ee0c as a result of conversation at https://github.com/php/php-src/pull/1038#issuecomment-77713046 over ten years ago, so in PHP 5.6 days. I think it was only documented years later though.

So I think that entry in `functionMap_php83delta.php` can just be removed and that will resolve the problem.

PR #4328, opened 2025-09-13
fix: checking optional properties in object shapes

calebdw:
Hello!

Closes https://github.com/phpstan/phpstan/issues/13511

By the time the `getProperty()` method is called, the `$hasProperty` has already been narrowed to `maybe` or it's been determined to be optional and overridden to `yes`. We should simply continue instead of returning `no`

Thanks!

### DIFF
--- a/src/Type/ObjectShapeType.php
+++ b/src/Type/ObjectShapeType.php
@@ -176,15 +176,7 @@ class ObjectShapeType implements Type
 			try {
 				$otherProperty = $type->getProperty((string) $propertyName, $scope);
 			} catch (MissingPropertyFromReflectionException) {
-				return AcceptsResult::createNo(
-					[
-						sprintf(
-							'%s does not have property $%s.',
-							$type->describe(VerbosityLevel::typeOnly()),
-							$propertyName,
-						),
-					],
-				);
+				continue;
 			}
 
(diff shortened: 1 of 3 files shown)


staabm:
don't get confused by the CI results.. [we currently investigate spurious CI errors](https://github.com/phpstan/phpstan/issues/13494)

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1541,6 +1541,12 @@ public function testBug13310(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug13511(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-13511.php');
```
ondrejmirtes: Test needs to be in CallToFunctionParametersRuleTest.

AnalyserIntegrationTest is reserved for crashes and performance issues.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
calebdw: Good to know, thank!

ondrejmirtes:
Thank you!

PR #4329, opened 2025-09-13
feat: add isOptionalProperty method to ObjectShapeType

calebdw:
Hello!

While I was digging around, I noticed that the ConstantArrayType has a similar method and figured it would be nice to have here as well.

Thanks!

### DIFF
--- a/src/Type/ObjectShapeType.php
+++ b/src/Type/ObjectShapeType.php
@@ -95,7 +95,7 @@ class ObjectShapeType implements Type
 			return TrinaryLogic::createNo();
 		}
 
-		if (in_array($propertyName, $this->optionalProperties, true)) {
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
What's your use case? Also, take a look where the ConstantArrayType method is used, and think about whether we're missing something with ObjectShapeType or not.

calebdw:
I don't have a specific use case, I just saw this on the array type and thought it might clean up all the `in_array` calls, if you'd rather not have it then feel free to close

> Also, take a look where the ConstantArrayType method is used, and think about whether we're missing something with ObjectShapeType or not.

I'm not sure what you're trying to get at here, it looks like `isOptionalKey()` is used in several places like when building a new array

PR #4231, opened 2025-08-10
Simplify Redis functionMap

VincentLanglet:
Lot of method have different signatures than phpstorm and I now nothing about the truth.

I only remove the one which was the same.

No changes in reflection golden tests 
https://github.com/phpstan/phpstan-src/actions/runs/16859904605/job/47758612800?pr=4231

PR #4331, opened 2025-09-14
Fixed functionMap.php definition to match RedisCluster stub

sayuprc:
The definition in functionMap.php was inconsistent with the RedisCluster stub.
This PR updates it to match the [stub](https://github.com/phpredis/phpredis/blob/92137ffd3f9b3a763be5afbdf14d14f01aa2c193/redis_cluster.stub.php).

ref: #4283


VincentLanglet:
Hi @sayuprc, the diff of the PR is pretty huge, and then difficult for ondrej to review it.

Also, be aware there is no need to describe every method here since PHPStan rely on PHPStorm stubs
that's why I already removed some definition in https://github.com/phpstan/phpstan-src/pull/4231
For instance, you're adding `_compress` definition 
```
'RedisCluster::_compress' => ['string', 'value'=>'string'],
```
but it's useless because it already exists (and is correct) on PHPStorm side
https://github.com/JetBrains/phpstorm-stubs/blob/d74d654cf966aaa56fd508c1f727c46313903aa2/redis/RedisCluster.php#L3653

So:
- A good way to improve the method definition is to update PHPStorm stubs https://github.com/JetBrains/phpstorm-stubs/blob/d74d654cf966aaa56fd508c1f727c46313903aa2/redis/RedisCluster.php#L12
- If the method definition is wrong on PHPStan, rather than fixing it, it's maybe better to removing it if the one on phpstorm stubs is correct
- If the method definition needs some PHPStan-related types like `__benevolent`, you'll have no choice to add a definition on phpstan side
- The smaller is the PR (you may open multiple PR instead), the easier it will be to be reviewed/merged

Also you can see the impact of the PR with the 
[Reflection golden test / Reflection golden test (8.5) (pull_request)](https://github.com/phpstan/phpstan-src/actions/runs/17709015682/job/50325167678?pr=4331) and looking at the Reflection Golden test section

You'll see diff like
```
1) PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "METHOD RedisCluster::pfMerge" ('METHOD RedisCluster::pfMerge', 'Is built-in: Yes\nHas side-ef...luster')
--- Expected
+++ Actual
@@ @@
     /**
      * @param string $key
      * @param array $keys
-     * @return bool
+     * @return (bool|RedisCluster)
      */
     function pfmerge(string $key, array $keys): bool|RedisCluster'

```
to check if it's ok.

Currently the golden test reflection reports 321 changes ; it's too many to be reviewed

ondrejmirtes:
Yeah, there's no need to "match a stub". If you want to fix some specific bugs (false positives) reported by PHPStan feel free to do so in a smaller focused PR. Thank you!

PR #4332, opened 2025-09-14
update PHAR compilation readme

canvural:
When I want to test how a change in PHPStan would affect a project, I compile the PHAR locally. But noticed the README was outdated. I was doing it by memory (and terminal history) 😄 This PR updates it.

### DIFF
--- a/compiler/README.md
+++ b/compiler/README.md
@@ -4,7 +4,9 @@
 
 ```bash
 composer install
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

Best way is to submit a draft PR and grab the compiled PHAR from Actions artifacts though 😊

PR #4336, opened 2025-09-15
Avoid crashing on enum without value

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/7927

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -965,12 +965,8 @@ final class ClassReflection
 		$initializerExprContext = InitializerExprContext::fromClassReflection($this);
 		foreach ($this->reflection->getCases() as $case) {
 			$valueType = null;
-			if ($case instanceof ReflectionEnumBackedCase) {
-				try {
-					$valueType = $this->initializerExprTypeResolver->getType($case->getValueExpression(), $initializerExprContext);
-				} catch (LogicException) {
-					// Enum case does not have a value
-				}
+			if ($case instanceof ReflectionEnumBackedCase && $case->hasValueExpression()) {
+				$valueType = $this->initializerExprTypeResolver->getType($case->getValueExpression(), $initializerExprContext);
 			}
 			$caseName = $case->getName();
 			$attributes = $this->attributeReflectionFactory->fromNativeReflection($case->getAttributes(), InitializerExprContext::fromClass($this->getName(), $this->getFileName()));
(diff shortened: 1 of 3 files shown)


src/Reflection/ClassReflection.php
```diff
@@ -945,7 +946,11 @@ public function getEnumCases(): array
 		foreach ($this->reflection->getCases() as $case) {
 			$valueType = null;
 			if ($case instanceof ReflectionEnumBackedCase) {
-				$valueType = $this->initializerExprTypeResolver->getType($case->getValueExpression(), $initializerExprContext);
+				try {
+					$valueType = $this->initializerExprTypeResolver->getType($case->getValueExpression(), $initializerExprContext);
+				} catch (LogicException) {
```
staabm: Catch and swallow of such a generic exception could backfire easily in the future. 

Can't we detect the situation somehow? Maybe by using the Reflection class ?

VincentLanglet: I thought about it.

I hesitated about introducing a hasValue method in the betterException class. But I dunno if ondrej wants to add extra method in his fork

What I can do is to just catch the needed call

ondrejmirtes (changes requested):
I don't want to catch LogicException. The right way to prevent this is to contribute something like `hasValueExpression` first to https://github.com/ondrejmirtes/BetterReflection to prevent the error.

Please note that contributing to that fork is a bit tricky. The last two commits are always "[GENERATED] Renamed" and "[GENERATED] Downgrading".

You need to start by `git reset --hard` before these two commits. Then make your changes. Then run `make downgrade-inner` (and commit that) and `make rename-inner` (and commit that).

Please note that the CI there isn't green. The library could appreciate some love but given it's only just me working on the fork I didn't prioritize that yet.

VincentLanglet:
I will need https://github.com/Roave/BetterReflection/releases/tag/6.65.0

ondrejmirtes:
I actually disagree with merging https://github.com/Roave/BetterReflection/pull/1527 instead of https://github.com/Roave/BetterReflection/pull/1528.

Since `hasValueExpression` was already added on https://github.com/Roave/BetterReflection/pull/1526 it should have just been added on the adapter as well.

I'll rebase the fork and adjust it to my taste.

ondrejmirtes:
Updated here.

VincentLanglet:
> Updated here.

Thanks, rebased and updated :)

ondrejmirtes:
Thank you!

PR #4339, opened 2025-09-17
Fix tests

ondrejmirtes:
Closes https://github.com/phpstan/phpstan/issues/13494

### DIFF
--- a/src/DependencyInjection/ValidateIgnoredErrorsExtension.php
+++ b/src/DependencyInjection/ValidateIgnoredErrorsExtension.php
@@ -24,6 +24,7 @@ use PHPStan\PhpDocParser\Parser\ConstExprParser;
 use PHPStan\PhpDocParser\Parser\TypeParser;
 use PHPStan\PhpDocParser\ParserConfig;
 use PHPStan\Reflection\InitializerExprTypeResolver;
+use PHPStan\Reflection\MissingStaticAccessorInstanceException;
 use PHPStan\Reflection\PhpVersionStaticAccessor;
 use PHPStan\Reflection\ReflectionProvider\DirectReflectionProviderProvider;
 use PHPStan\Reflection\ReflectionProvider\DummyReflectionProvider;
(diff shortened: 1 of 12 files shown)


src/DependencyInjection/ValidateIgnoredErrorsExtension.php
```diff
@@ -66,120 +67,143 @@ public function loadConfiguration(): void
...
 
-		throw new InvalidIgnoredErrorPatternsException($errors);
+			throw new InvalidIgnoredErrorPatternsException($errors);
+		} finally {
+			if ($originalReflectionProvider !== null) {
+				ReflectionProviderStaticAccessor::registerInstance($originalReflectionProvider);
+			}
+			if ($originalPhpVersion !== null) {
+				PhpVersionStaticAccessor::registerInstance($originalPhpVersion);
+			}
```
staabm, line 205: as the code stands it looks like `PhpVersionStaticAccessor::registerInstance()` might be called without also calling `ReflectionProviderStaticAccessor::registerInstance`, which might mean `PhpVersionStaticAccessor::registerInstance()` should also clear the cache

ondrejmirtes: I'm doing ObjectType::resetCaches only in ReflectionProvider accessor because that's what it depends on. I don't think there's any cache depending on PhpVersionStaticAccessor?

staabm:
nice!

ondrejmirtes:
You were right about the PhpVersion, I just saw a failure that could be explained by that:

```
1) PHPStan\Type\Generic\GenericObjectTypeTest::testIsSuperTypeOf with data set #2 (PHPStan\Type\Generic\GenericObjectType Object (...), PHPStan\Type\Generic\GenericObjectType Object (...), PHPStan\TrinaryLogic Object (...))
ReflectionClass<object> -> isSuperTypeOf(ReflectionClass<stdClass>)
--- Expected
+++ Actual
@@ @@
-'No'
+'Yes'

/Users/ondrej/Development/phpstan/tests/PHPStan/Type/Generic/GenericObjectTypeTest.php:270

2) PHPStan\Type\Generic\GenericObjectTypeTest::testIsSuperTypeOf with data set #3 (PHPStan\Type\Generic\GenericObjectType Object (...), PHPStan\Type\Generic\GenericObjectType Object (...), PHPStan\TrinaryLogic Object (...))
ReflectionClass<stdClass> -> isSuperTypeOf(ReflectionClass<object>)
--- Expected
+++ Actual
@@ @@
-'No'
+'Maybe'
```

Gonna reset the caches in any case. Thanks!

VincentLanglet:
Awesome !

PR #4340, opened 2025-09-17
Narrow l/r/trim() arg on comparison to ''

staabm:
closes https://github.com/phpstan/phpstan/issues/12973

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -1311,8 +1311,6 @@ final class TypeSpecifier
 			&& isset($exprNode->getArgs()[0])
 			&& $constantStringValue === ''
 		) {
-			$argType = $scope->getType($exprNode->getArgs()[0]->value);
-
 			return $this->create(
 				$exprNode->getArgs()[0]->value,
 				TypeCombinator::intersect(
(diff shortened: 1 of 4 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -1303,6 +1303,26 @@ private function specifyTypesForConstantStringBinaryExpression(
...
+			$context->false()
+			&& $exprNode instanceof FuncCall
+			&& $exprNode->name instanceof Name
+			&& in_array(strtolower((string) $exprNode->name), ['trim', 'ltrim', 'rtrim'], true)
+			&& isset($exprNode->getArgs()[0])
+			&& $constantStringValue === ''
+		) {
+			$argType = $scope->getType($exprNode->getArgs()[0]->value);
+
+			if ($argType->isString()->yes()) {
```
ondrejmirtes: I'd say that if you omit `if ($argType->isString()->yes()) {` and do `new IntersectionType([new StringType(), new AccessoryNonEmptyStringType()])`, we could also narrow `mixed` to `non-empty-string`.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4343, opened 2025-09-18
PHPUnit 10.x is compatible with PHP 8.1

staabm:
Don't downgrade to PHPUnit 9.6.x


staabm:
seems not worth the effort.. compatibility varies with paraunit and phpunit combinations

PR #4344, opened 2025-09-18
Mark Memcached:: methods as impure.

biozshock:
Fixes phpstan/phpstan#13444


bin/functionMetadata_original.php
```diff
@@ -170,6 +170,41 @@
...
+	'Memcached::fetch' => ['hasSideEffects' => true],
+	'Memcached::fetchAll' => ['hasSideEffects' => true],
+	'Memcached::flush' => ['hasSideEffects' => true],
+	'Memcached::get' => ['hasSideEffects' => true],
+	'Memcached::getByKey' => ['hasSideEffects' => true],
+	'Memcached::getDelayed' => ['hasSideEffects' => true],
+	'Memcached::getDelayedByKey' => ['hasSideEffects' => true],
+	'Memcached::getMulti' => ['hasSideEffects' => true],
+	'Memcached::getMultiByKey' => ['hasSideEffects' => true],
+	'Memcached::getServerByKey' => ['hasSideEffects' => true],
```
VincentLanglet, line 194: Does get method has side effects too ?

biozshock: @VincentLanglet I'm sorry, which "get" method?
The `getServerByKey`, where your comment is located has the side effect: https://www.php.net/manual/en/memcached.getserverbykey.php#refsect1-memcached.getserverbykey-returnvalues

VincentLanglet: I meant all the getFoo method sonce they arr just getter

But if I understand correctly they all have side effect because they have impact on the getResultCode

biozshock: @VincentLanglet exactly. I think that i've marked all methods that change `getResultCode`.

VincentLanglet:
Might be better to wait for https://github.com/phpstan/phpstan-src/pull/4422 and use the same strategy

biozshock:
> Might be better to wait for #4422 and use the same strategy

Not all methods are impure, so that wouldn't work.

ondrejmirtes:
@biozshock It might be still fewer lines of information to mark all as impure and the exclude a few with `@phpstan-pure`.

VincentLanglet:
https://github.com/phpstan/phpstan/issues/13444 was solved

PR #4347, opened 2025-09-18
More precise `microtime()` return type

takaram:
Return value of `microtime(false)` is never falsy.

### DIFF
--- a/src/Type/Php/MicrotimeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/MicrotimeFunctionReturnTypeExtension.php
@@ -6,9 +6,11 @@ use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Accessory\AccessoryNonFalsyStringType;
 use PHPStan\Type\BenevolentUnionType;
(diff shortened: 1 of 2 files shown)


src/Type/Php/MicrotimeFunctionReturnTypeExtension.php
```diff
@@ -26,8 +28,10 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
 
 	public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): Type
 	{
+		$stringType = new IntersectionType([new StringType(), new AccessoryNonFalsyStringType()]);
```
staabm, line 31: not sure it will be helpful, but I think we could in addition make it a `AccessoryLowercaseStringType`

mvorisek: Why? It is neither LC or UC, thus it should be accepted also by `uppercase-string` type.

ondrejmirtes:
Thank you!

src/Type/Php/MicrotimeFunctionReturnTypeExtension.php
VincentLanglet: > Why? It is neither LC or UC, thus it should be accepted also by uppercase-string type.

You can be both LC and UC.
For instance `42` is both.

So we should add both accessory AccessoryLowercaseStringType and AccessoryUppercaseStringType

PR #4348, opened 2025-09-19
Add non regression test for issue 9386

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9386

Cf https://github.com/phpstan/phpstan-src/pull/4168#issuecomment-3285077698


ondrejmirtes:
Thank you.

PR #4349, opened 2025-09-19
add optional strict check for printf parameter types

schlndh:
Closes https://github.com/phpstan/phpstan/issues/13496
Previous PR: https://github.com/phpstan/phpstan-src/pull/3977

Would you accept turning this on by default in phpstan-strict-rules?

### DIFF
--- a/src/Rules/Functions/PrintfParameterTypeRule.php
+++ b/src/Rules/Functions/PrintfParameterTypeRule.php
@@ -42,6 +42,7 @@ final class PrintfParameterTypeRule implements Rule
 		private PrintfHelper $printfHelper,
 		private ReflectionProvider $reflectionProvider,
 		private RuleLevelHelper $ruleLevelHelper,
+		private bool $checkStrictPrintfPlaceholderTypes,
 	)
 	{
 	}
@@ -100,15 +101,23 @@ final class PrintfParameterTypeRule implements Rule
 			new NullType(),
 		);
 		// Type on the left can go to the type on the right, but not vice versa.
-		$allowedTypeNameMap = [
-			'strict-int' => 'int',
-			'int' => 'castable to int',
-			'float' => 'castable to float',
-			// These are here just for completeness. They won't be used because, these types are already enforced by
-			// CallToFunctionParametersRule.
-			'string' => 'castable to string',
-			'mixed' => 'castable to string',
-		];
+		$allowedTypeNameMap = $this->checkStrictPrintfPlaceholderTypes
+			? [
+				'strict-int' => 'int',
+				'int' => 'int',
+				'float' => 'float',
+				'string' => 'int|float|string|Stringable',
+				'mixed' => 'int|float|string|Stringable',
+			]
+			: [
+				'strict-int' => 'int',
+				'int' => 'castable to int',
+				'float' => 'castable to float',
+				// These are here just for completeness. They won't be used because, these types are already enforced by
+				// CallToFunctionParametersRule.
+				'string' => 'castable to string',
+				'mixed' => 'castable to string',
+			];
 
 		for ($i = $formatArgumentPosition + 1, $j = 0; $i < $argsCount; $i++, $j++) {
 			// Some arguments may be skipped entirely.
@@ -117,10 +126,10 @@ final class PrintfParameterTypeRule implements Rule
 					$scope,
 					$args[$i]->value,
 					'',
-					static fn (Type $t) => $placeholder->doesArgumentTypeMatchPlaceholder($t),
+					fn (Type $t) => $placeholder->doesArgumentTypeMatchPlaceholder($t, $this->checkStrictPrintfPlaceholderTypes),
 				)->getType();
 
-				if ($argType instanceof ErrorType || $placeholder->doesArgumentTypeMatchPlaceholder($argType)) {
+				if ($argType instanceof ErrorType || $placeholder->doesArgumentTypeMatchPlaceholder($argType, $this->checkStrictPrintfPlaceholderTypes)) {
 					continue;
 				}
 
--- a/src/Rules/Functions/PrintfParameterTypeRule.php
+++ b/src/Rules/Functions/PrintfParameterTypeRule.php
@@ -106,8 +106,8 @@ final class PrintfParameterTypeRule implements Rule
 				'strict-int' => 'int',
 				'int' => 'int',
 				'float' => 'float',
-				'string' => 'int|float|string|Stringable',
-				'mixed' => 'int|float|string|Stringable',
+				'string' => '__stringandstringable',
+				'mixed' => '__stringandstringable',
 			]
 			: [
 				'strict-int' => 'int',
--- a/tests/PHPStan/Rules/Functions/PrintfParameterTypeRuleTest.php
+++ b/tests/PHPStan/Rules/Functions/PrintfParameterTypeRuleTest.php
@@ -14,6 +14,8 @@ use const PHP_VERSION_ID;
 class PrintfParameterTypeRuleTest extends RuleTestCase
 {
 
+	private bool $checkStrictPrintfPlaceholderTypes = false;
+
 	protected function getRule(): Rule
 	{
 		$reflectionProvider = $this->createReflectionProvider();
@@ -30,6 +32,7 @@ class PrintfParameterTypeRuleTest extends RuleTestCase
 				true,
 				false,
 			),
+			$this->checkStrictPrintfPlaceholderTypes,
 		);
 	}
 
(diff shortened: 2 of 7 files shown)


src/Rules/Functions/PrintfParameterTypeRule.php
```diff
@@ -100,15 +101,23 @@ public function processNode(Node $node, Scope $scope): array
...
-			// CallToFunctionParametersRule.
-			'string' => 'castable to string',
-			'mixed' => 'castable to string',
-		];
+		$allowedTypeNameMap = $this->checkStrictPrintfPlaceholderTypes
+			? [
+				'strict-int' => 'int',
+				'int' => 'int',
+				'float' => 'float',
+				'string' => 'int|float|string|Stringable',
```
schlndh: This seems a bit long and it's not completely accurate either (it accepts classes with `__toString()` even if they don't implement `Stringable`). The other alternative I considered was just `string`, but that might be a bit confusing to users. Is there anything better?

ondrejmirtes: Let's try `__stringandstringable`. I can see you're already using `new StringAlwaysAcceptingObjectWithToStringType` in a related place.

ondrejmirtes:
Yes, please contribute seting thing parameter to true in phpstan-strict-rules once this is released 😊

ondrejmirtes:
Thank you!

ondrejmirtes:
Please also send a docs update for phpstan.org about `checkStrictPrintfPlaceholderTypes`. Thanks!

tests/PHPStan/Rules/Functions/PrintfParameterTypeRuleTest.php
```diff
@@ -111,4 +114,139 @@ public function test(): void
...
+			[
+				'Parameter #2 of function printf is expected to be float by placeholder #1 ("%f"), true given.',
+				44,
+			],
+			[
+				'Parameter #2 of function printf is expected to be float by placeholder #1 ("%f"), SimpleXMLElement given.',
+				45,
+			],
+			[
+				'Parameter #2 of function printf is expected to be __stringandstringable by placeholder #1 ("%s"), null given.',
```
ruudk, line 242: @schlndh @ondrejmirtes After enabling this rule I noticed a lot of errors related to `string|null` types. 

Personally, I think replacing those `sprintf` call arguments with `$value ?? ''` feels like making a computer happy. As it does exactly the same as what `sprintf` does for `%s`.

See https://3v4l.org/TlHU9#v8.4.13

I know this is a strict rule, but wonder if this was really intentional or something that should be reconsidered?

ondrejmirtes: I feel like it's useful that PHPStan lets you know you're potentially passing `null` to `sprintf`. You might want to deal with `null` differently than what `sprintf` does.

schlndh: I intentionally implemented it like that to match my understanding of "strict". However, I encountered the same thing: hundreds of issues like this, most of which I don't care about (though I'm fine with letting them all fall into the baseline). So I would be open to changing it. I'm just not sure what's the best way to go about it.

Strict-rules do allow `$null . $string` and `"{$null}"`, so I guess this wouldn't have to be as strict either (unless it's allowed by omission rather than intentionally). At the very least it's a bit inconsistent.

On the other hand, there are third-party rules (e.g. [shipmonk rules](https://github.com/shipmonk-rnd/phpstan-rules?tab=readme-ov-file#forbidnullininterpolatedstring)) which do prohibit it, so clearly there is also interest in the fully strict version.

Another thing to consider is whether it should be allowed only in `%s` or also other placehoders?

ruudk: > Strict-rules do allow `$null . $string` and `"{$null}"`, so I guess this wouldn't have to be as strict either (unless it's allowed by omission rather than intentionally). At the very least it's a bit inconsistent.

This is an excellent point. 

> Another thing to consider is whether it should be allowed only in `%s` or also other placehoders?

I would say only for `%s`.

ondrejmirtes: An idea: string|null should be reported only on level 8+, basicallh RuleLevelHelper should be utilized.

ruudk: Unfortunately, that won't help me (level 9 🙈)

angeleg: > I would say only for `%s`.

To expand on that point: `%d` + `null` make `0`, which is confusable with, well, passing an actual `0` integer. 😄  So for that reason I wouldn't allow it either.

schlndh: @ondrejmirtes Re: level 8

> 8. report calling methods and accessing properties on nullable types

IMO that's stretching it a little bit. And it wouldn't be consistent with how string + null is handled in other cases (concatenation, interpolation, ...).

Let's consider consistency with phpstan-strict-rules ([playground](https://phpstan.org/r/4c3e52bd-58ab-4f1b-bbb4-37ee5d3e141b)):

- Strict-rules prohibit `bool` and `null` in numerical operations (consistent with `%d` and `%f`).
- Strict-rules allow `numeric-string` in numerical operations (inconsistent with `%f`, `%d` is probably OK since float is excluded as well).
- Strict-rules prohibit general string in numerical operations (consistent with `%d` and `%f`).
- Strict-rules allow `bool` and null in string operations (inconsistent with `%s`).

Idea:
- Let's enable passing `null` to `%s`. If PHPStan later adds a config option to limit type coercions then it could be used here (as well as `$string . $null`, `"{$null}"`, ...). 
- Let's enable passing `numeric-string` to `%f`. I excluded it, because `numeric-string` could lose precision by doing so. But it's probably too strict.
- Let's keep passing `bool` to `%s` prohibited. I could later try to prohibit `bool` in `$string . $bool` and `"{$bool}"` in strict-rules (no promises).

ruudk: Just a small side note, PHPStorm will tell me to remove `(string) $null` for `%s` in a `sprintf`. It's only fine with `$null ?? ''`.

Screenshot 2025-09-26 at 12 42 39@2x

heiglandreas:
I just stumbled upon this and - what shall I say - the usability seems rather limited. Especially when using the same parameter multiple times in the format string.

https://phpstan.org/r/470b0175-9359-418e-87d5-3a37013bccae

Why does an int not need to be cast to string when the format asks for a string? Why does an int not need to be cast to a float? 

I do understand that it can make sense to hint that the formatting parameter will cast a value to something else. But the messages are actually wrong: 

For one thing: `Parameter #2 of function sprintf is expected to be float by placeholder #3 ("%1$0.02f")` - This is placeholder 1 we are talking about. 
And the other thing is that sprintf does not expect the value to be float. It merely will format the value as float. That is a small but important difference! `sprinf` expects anything. But it will make sure that the passed parameter will be cast to the expected format when integrated into the string according to the rules noted in the format. 

And as https://phpstan.org/r/83f7819b-7fef-46dd-a397-c38506bad3b3 and https://phpstan.org/r/ab740815-f7c1-41a8-96ad-2b4d508ca3c9 show it does not apply these rules in every situation. Here I would expect PHPStan to trigger an error whenever something is passed that does not work with the provided type.

In the second example I would also expect PHPStan to understand the function signature that is shown in [the PHP documentation](https://www.php.net/manual/en/function.sprintf.php) (` sprintf(string $format, mixed ...$values): string` especially as `bool` and `null` are nowhere usable in the formatter-strings, so the values should not be allowed at all - according to the logic that only values are allowed that the formatting string actually expects...

ondrejmirtes:
@heiglandreas I think it's fine, you don't need to turn on the strict version if you don't want it?

heiglandreas:
I am more concerned about a false sense of "safety" that this rule implies. 

Even if I want a very strict type-checking, this does - at least in the current way - not add a reliable safety to the `sprintf`-function. 

* hard-coded formatting-strings that are passed as a parameter are not taken into account
* mixed-parameters that require determining the parameter type based on the passed value are not supported
* argumend-numbering - especially with multiple usages of the same argument - are not taken into account.

So enabling that in the strict-checking by default seems ... an interesting choice. Now one has to explicitly **remove** this very opinionated and brittle check. Instead of allowing one to opt-in if that is what I want and when I exactly know what to do.

derrabus:
> So enabling that in the strict-checking by default seems

Strict rules are not enabled by default, are they?

tests/PHPStan/Rules/Functions/PrintfParameterTypeRuleTest.php
derrabus: > Idea:
> 
> * Let's enable passing `null` to `%s`. If PHPStan later adds a config option to limit type coercions then it could be used here (as well as `$string . $null`, `"{$null}"`, ...).

I don't know if that helps, but I've reviewed the new errors that popped up in my current project. The dominant error reported was that we pass something that could be null. And in most of the cases, it would be a problem is the value was actually `null`. It was also old news because we already knew that the code in question is a bit too lax when it comes to handling `null`. So all errors that the new rule added to our baseline were actually just "more of the same".

> * Let's enable passing `numeric-string` to `%f`. I excluded it, because `numeric-string` could lose precision by doing so. But it's probably too strict.

Yes, it is. In fact, you actually might lose precision when casing a numeric string to a float. And besides it doesn't make much sense to cast a string to a float just to format it as a string again.

heiglandreas:
If you enable the strict rules, then this rule is also enabled. As I explained above the rule is not really checking in all circumstances and giving a very skewed result. 

I can see that it might make sense in certain circumstances. But for sure not in all situations where one wants to enable a strict check.

ondrejmirtes:
1) This is in bleeding edge which means things are not set in stone and feedback is very appreciated.
2) strict-rules are by definition opinionated.

heiglandreas:
Opinionated is fine, lulling into a false sense of security is ... arguable? 😁

PR #4351, opened 2025-09-20
False positive "non-empty-array<int, int> might not be a list" when change existing list key

staabm:
closes https://github.com/phpstan/phpstan/issues/13035

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5610,7 +5610,10 @@ final class NodeScopeResolver
 					$varForSetOffsetValue = $var->var;
 				}
 
-				if ($scope->hasExpressionType($var)->yes()) {
+				if (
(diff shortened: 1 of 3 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -5697,11 +5697,24 @@ private function processAssignVar(
...
+						$var->dim,
+						$assignedPropertyExpr,
+					);
+				} else {
+					$assignedPropertyExpr = new SetOffsetValueTypeExpr(
+						$varForSetOffsetValue,
+						$var->dim,
+						$assignedPropertyExpr,
+					);
+				}
```
staabm, line 5629: with this change, we make sure `PropertyAssignNode` contains a proper type.
we already had the right type in scope from a previous fix.

ondrejmirtes:
Thank you!

PR #4352, opened 2025-09-20
Faster code downgrade to speedup CI

staabm:
make use of https://github.com/ondrejmirtes/simple-downgrader/pull/14

downgrading phpstan-src in CI before this PR took ~10min (windows) and ~4m30s (ubuntu)

we are now down to ~15-25 seconds

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,7 +38,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "${{ matrix.php-version }}"
+          php-version: "8.4"
 
       - name: "Downgrade PHPUnit"
         if: matrix.php-version == '7.4' || matrix.php-version == '8.0' || matrix.php-version == '8.1'
@@ -53,6 +53,12 @@ jobs:
           composer install --no-interaction --no-progress --working-dir=compiler
           ./compiler/vendor/bin/simple-downgrade downgrade -c build/downgrade.php ${{ matrix.php-version }}
           composer dump
+
+      - name: "Change to runtime PHP version"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
 
       - name: "Validate Composer"
         run: "composer validate"
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,7 +38,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "8.4"
+          php-version: "${{ matrix.php-version }}"
 
       - name: "Downgrade PHPUnit"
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "${{ matrix.php-version }}"
+          php-version: "8.4"
```
ondrejmirtes: I'd do this a bit differently:

1) Install `${{ matrix.php-version }}`
2) Run the composer steps
3) Only when downgrading: install 8.4
4) Transform source code
5) Only when downgrading: install `${{ matrix.php-version }}` again

You can try playing with https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action to see if we can achieve DRY across different workflows.

ondrejmirtes: What I want to avoid is always start by installing 8.4 which is immediately going to be ovewritten with matrix version.

ondrejmirtes (changes requested):

.github/workflows/lint.yml
staabm: > I'd do this a bit differently:

done

> You can try playing with [docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) to see if we can achieve DRY across different workflows.

I would do that in a followup

staabm: not sure what todo with the "Compile PHAR" - or what the actual problem is.

.github/workflows/static-analysis.yml
```diff
@@ -57,6 +57,14 @@ jobs:
       - name: "Install dependencies"
         run: "composer install --no-interaction --no-progress"
 
+      - name: "Change to simple-downgrade PHP version"
+        uses: "shivammathur/setup-php@v2"
```
ondrejmirtes, line 62: No if condition here?

.github/workflows/lint.yml
ondrejmirtes: Can you link the failure? I don't see anything big.

If you mean "Checksum PHAR" - that's fine. The idea behind that job is that the PHAR checksum should not change when src/ doesn't change.

If we change something about the compilation process then it's fine if PHAR contents change. I suspect that some ConstFetches now use FQN like `\null` so the source slightly changed.

ondrejmirtes (changes requested):

.github/workflows/lint.yml
staabm: > If you mean "Checksum PHAR" - that's fine.

yep, that one. wasn't sure what it means

.github/workflows/static-analysis.yml
staabm: fixed

ondrejmirtes:
Really great! 🎉

PR #4353, opened 2025-09-20
More precise `hrtime()` return type

takaram:
`hrtime(true)` always returns positive-int.
`hrtime(false)` returns an array of `[second, nanosecond]`, where second is positive-int and nanosecond is between 0 and 10⁹-1 (because 10⁹ nanosecond = 1 second)

## Another solution
I tried to remove the return type extension and add a stub like this:
```php
/**
 * @template T of bool
 * @param T $as_number
 * @return (T is true ? __benevolent<float|positive-int> : array{positive-int, int<0, 999999999>})
 */
function hrtime(bool $as_number = false): array|int|float {}
```

but this doesn't work:
```php
dumpType(hrtime()); // array{int<1, max>, int<0, 999999999>}|false
dumpType(hrtime(true)); // float|int<1, max>|false
```

`false` seems to come from phpstan/php-8-stubs.
https://github.com/phpstan/php-8-stubs/blob/0.4.30/stubs/ext/standard/hrtime.php

---

I also tried this stub:
```php
/**
 * @template T of bool
 * @param T $as_number
 * @return (T is true ? __benevolent<float|positive-int> : (T is false ? array{positive-int, int<0, 999999999>} : false))
 */
function hrtime(bool $as_number = false): array|int|float|false {}
```

This works in most cases, but we see `false` when argument is not bool:
```php
// before: array{int, int}|float|int
// after:  array{int<1, max>, int<0, 999999999>}|float|int<1, max>|false
dumpType(hrtime(1));
```

### DIFF
--- a/src/Type/Php/HrtimeFunctionReturnTypeExtension.php
+++ b/src/Type/Php/HrtimeFunctionReturnTypeExtension.php
@@ -12,7 +12,7 @@ use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\FloatType;
-use PHPStan\Type\IntegerType;
+use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use function count;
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
I'm sorry for not looking at this sooner, please rebase and fix the conflict.

takaram:
@ondrejmirtes 
I rebased the branch.

staabm (approved):
makes sense to me.

VincentLanglet (approved):

staabm:
thank you!

PR #4354, opened 2025-09-21
Dynamic return type for DOM::appendChild and similar methods

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/9200


VincentLanglet:
Lint failure is unrelated and comes from https://github.com/phpstan/phpstan-src/commit/4b4836ac4c352e85033b337e55817f71983fdac5

ondrejmirtes:
Thank you!

PR #4359, opened 2025-09-22
Restrict accepted args of php_uname()

dktapps:
This now throws an error in PHP 8.4 when invalid args are given.

See https://www.php.net/manual/en/function.php-uname.php

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -8075,7 +8075,7 @@ return [
 'php_logo_guid' => ['string'],
 'php_sapi_name' => ['__benevolent<non-empty-string|false>'],
 'php_strip_whitespace' => ['string', 'file_name'=>'string'],
-'php_uname' => ['string', 'mode='=>'string'],
+'php_uname' => ['"a"|"s"|"n"|"r"|"v"|"m"', 'mode='=>'string'],
 'php_user_filter::filter' => ['int', 'in'=>'resource', 'out'=>'resource', '&rw_consumed'=>'int', 'closing'=>'bool'],
 'php_user_filter::onClose' => ['void'],
 'php_user_filter::onCreate' => ['bool'],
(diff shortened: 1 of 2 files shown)


resources/functionMap.php
```diff
@@ -8075,7 +8075,7 @@
 'php_logo_guid' => ['string'],
 'php_sapi_name' => ['__benevolent<non-empty-string|false>'],
 'php_strip_whitespace' => ['string', 'file_name'=>'string'],
-'php_uname' => ['string', 'mode='=>'string'],
+'php_uname' => ['"a"|"s"|"n"|"r"|"v"|"m"', 'mode='=>'string'],
```
staabm: looks like your intension was to restrict the parameter. you have adjusted the return-type though

dktapps: Fixed, thanks

thg2k:
I believe that's a regression for phpstan, and that it should go in the 8.4-specific functionMap, so if you are validating for earlier versions it should not error. But that might be negligible.. so wait for the boss review :)

dktapps:
Well, the function still won't work as expected on <8.4 with wrong arguments

resources/functionMap.php
```diff
@@ -8075,7 +8075,7 @@
 'php_logo_guid' => ['string'],
 'php_sapi_name' => ['__benevolent<non-empty-string|false>'],
 'php_strip_whitespace' => ['string', 'file_name'=>'string'],
-'php_uname' => ['string', 'mode='=>'string'],
+'php_uname' => ['string', 'mode='=>'"a"|"s"|"n"|"r"|"v"|"m"'],
```
ondrejmirtes: I don't accept parameter type restrictions outside of bleeding edge. Because this would require users to also narrow down their PHPDocs when they're currently just passing `string` there.

So this change belongs to `resources/functionMap_bleedingEdge.php`.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #4366, opened 2025-09-24
array_column: proper support for private properties

janedbal:
Closes https://github.com/phpstan/phpstan/issues/13573

### DIFF
--- a/src/Type/Php/ArrayColumnHelper.php
+++ b/src/Type/Php/ArrayColumnHelper.php
@@ -147,7 +147,15 @@ final class ArrayColumnHelper
 					continue;
 				}
 
-				$returnTypes[] = $type->getInstanceProperty($propertyName, $scope)->getReadableType();
+				$property = $type->getInstanceProperty($propertyName, $scope);
+				if (!$property->isPublic()) {
+					if (!$type->hasMethod('__isset')->no() && !$type->hasMethod('__get')->no()) {
+						$returnTypes[] = new MixedType();
+					}
+					continue;
+				}
+
+				$returnTypes[] = $property->getReadableType();
 			}
 		}
 
--- a/tests/PHPStan/Analyser/nsrt/array-column.php
+++ b/tests/PHPStan/Analyser/nsrt/array-column.php
@@ -244,11 +244,69 @@ class ArrayColumnTest
 final class Foo
 {
 
+	public int $publicProperty;
+	protected int $protectedProperty;
+	private int $privateProperty;
+
+	/** @param array<int, self> $a */
+	public function doFoo(array $a): void
+	{
+		assertType('array{}', array_column($a, 'nodeName'));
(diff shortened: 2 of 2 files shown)


src/Type/Php/ArrayColumnHelper.php
```diff
@@ -147,7 +147,15 @@ private function getOffsetOrProperty(Type $type, Type $offsetOrProperty, Scope $
 					continue;
 				}
 
-				$returnTypes[] = $type->getInstanceProperty($propertyName, $scope)->getReadableType();
+				$property = $type->getInstanceProperty($propertyName, $scope);
+				if (!$property->isPublic()) {
+					if (!$type->hasMethod('__isset')->no() && !$type->hasMethod('__get')->no()) {
```
staabm, line 152: maybe it would make sense to work with `ClassReflection->allowsDynamicProperties()`. this would also include support for the `#[AllowDynamicProperties]` attribute etc.

src/Type/Php/ArrayColumnHelper.php
```diff
@@ -147,7 +147,15 @@ private function getOffsetOrProperty(Type $type, Type $offsetOrProperty, Scope $
 					continue;
 				}
 
-				$returnTypes[] = $type->getInstanceProperty($propertyName, $scope)->getReadableType();
+				$property = $type->getInstanceProperty($propertyName, $scope);
+				if (!$property->isPublic()) {
```
JanTvrdik, line 151: Why does getInstanceProperty return non-public property? I though the point of the scope parameter was to filter the property by visibility.

ondrejmirtes: It has never worked like that. The main reason why `Type::getInstanceProperty()` asks for Scope is to correctly transform `static` type in property type based on the place it's accessed from.

If a class has private property, this method will still return that. It's useful to provide an error like "Access to private property of class X" instead of "Access to undefined property".

Also Scope is used in `ClassReflection::getProperty()` to decide if it should continute iterating over class reflection extensions, or return the reflection object: https://github.com/phpstan/phpstan-src/blob/ac71c94991d489c60f9fe08be02f26226f89770f/src/Reflection/ClassReflection.php#L735-L755

ondrejmirtes (changes requested):
I have an idea: What if instead we called `Type::toArray()` and then asked for `$arrayType->getOffsetValueType()` with the `array_column` 2nd arg?

ondrejmirtes:
I realize probably not, the returned array there has weird keys. But `ObjectType::toArray()` still has some useful logic that asks `UniversalObjectCratesClassReflectionExtension` and is also interested in whether the class is final etc. We should get inspired about that here.

janedbal:
> We should get inspired about that here.

Can you be more specific please? Which part of current implementation is wrong and need to be adjusted?

ondrejmirtes:
This PR references already closed issue.

PR #4367, opened 2025-09-24
add bcround function signature for php8.4

themasch:
Add signature for bcround so phpstan knows that it accepts & returns `numeric-string` instead of a raw `string`. 

phpdoc link: https://www.php.net/manual/en/function.bcround.php

Related: https://github.com/phpstan/phpstan/discussions/13364

### DIFF
--- a/resources/functionMap_php84delta.php
+++ b/resources/functionMap_php84delta.php
@@ -15,6 +15,7 @@
  */
 return [
 	'new' => [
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

themasch:
Thank you for providing this tool, and guidance on how we can help to improve it.
I deeply appreciate it!

PR #4369, opened 2025-09-25
TableErrorFormatter: Link path in JetBrain IDEs

vrana:
New implementation of #4076.


ondrejmirtes:
Thank you!

EmilMassey:
@vrana 
1. Any idea why it doesn't work in PhpStorm 2025.2.2 using "Reworked 2025" terminal? Path is not linked
2. `editorUrl` [parameter](https://phpstan.org/user-guide/output-format#opening-file-in-an-editor) is now ignored when phpstan is executed in PHPStorm terminal. Was it intentional?

Terminal selection dropdown with Reworked 2025 selected
PHPstan execution output in PHPStorm terminal

PR #4371, opened 2025-09-26
Remove duplicate method word in error message

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12557

I think this is the only text with the issue


ondrejmirtes:
Thank you!

PR #4373, opened 2025-09-26
Fix XDOMPath::query() return type

stefanfisk:
Resolves phpstan/phpstan#13595


stefanfisk:
Removing the stub for `DOMXPath` fixes the issue and all tests pass. But I don't understand if there are any other consequences to this approach compared to providing a fully typed stub.

ondrejmirtes:
Thank you.

PR #4374, opened 2025-09-27
Simplify ClassReflectionExtensions

staabm:
identified in https://github.com/phpstan/phpstan-src/pull/4372

```
------ ------------------------------------------------------------------------------------------------------------------- 
  Line   src/Reflection/Annotations/AnnotationsMethodsClassReflectionExtension.php                                          
 ------ ------------------------------------------------------------------------------------------------------------------- 
  34     Offset string on array<PHPStan\Reflection\ExtendedMethodReflection> in isset() always exists and is not nullable.  
         🪪  isset.offset                                                                                                   
         at src/Reflection/Annotations/AnnotationsMethodsClassReflectionExtension.php:34                                    
 ------ ------------------------------------------------------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------------------------------------- 
  Line   src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php                                         
 ------ --------------------------------------------------------------------------------------------------------------------- 
  28     Offset string on array<PHPStan\Reflection\ExtendedPropertyReflection> in isset() always exists and is not nullable.  
         🪪  isset.offset                                                                                                     
         at src/Reflection/Annotations/AnnotationsPropertiesClassReflectionExtension.php:28                                   
 ------ --------------------------------------------------------------------------------------------------------------------- 
 ```

### DIFF
--- a/src/Reflection/Annotations/AnnotationsMethodsClassReflectionExtension.php
+++ b/src/Reflection/Annotations/AnnotationsMethodsClassReflectionExtension.php
@@ -31,7 +31,7 @@ final class AnnotationsMethodsClassReflectionExtension implements MethodsClassRe
 			$this->methods[$classReflection->getCacheKey()][$methodName] = $method;
 		}
 
-		return isset($this->methods[$classReflection->getCacheKey()][$methodName]);
+		return true;
 	}
 
 	public function getMethod(ClassReflection $classReflection, string $methodName): ExtendedMethodReflection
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Reflection/Annotations/AnnotationsMethodsClassReflectionExtension.php [hunk 1]
```diff
@@ -31,7 +31,7 @@ public function hasMethod(ClassReflection $classReflection, string $methodName):
 			$this->methods[$classReflection->getCacheKey()][$methodName] = $method;
```
staabm, line 31: in case this property is meant to be a cache, I think we could adjust the above IF case, as it effectively means a non-existant method will be checked for over and over again

staabm:
regression test added in https://github.com/phpstan/phpstan-src/pull/4372/commits/5374f423b4a18a83f727c155393884defd948c25

ondrejmirtes:
Thank you!

PR #4376, opened 2025-09-27
Improve ConstantArrayType inference

staabm:
extracted from https://github.com/phpstan/phpstan-src/pull/4372


ondrejmirtes:
Thank you!

PR #4377, opened 2025-09-27
use string instead of __stringandstringable in printf parameter error message

schlndh:
As requested here: https://github.com/phpstan/phpstan/pull/13594#issuecomment-3341869499


ondrejmirtes:
Thank you!

PR #4379, opened 2025-09-27
Rely on RuleLevelHelper in InvalidKeyInArrayItemRule

VincentLanglet:
This change is extracted from https://github.com/phpstan/phpstan-src/pull/4012

I feel like it will help for https://github.com/phpstan/phpstan/issues/13592 @ondrejmirtes by having a consistent behavior in code like
https://phpstan.org/r/43149492-9d7a-4d88-9f19-208b12d262b3 (currently mixed is reported only for one rule but not the other)

Similar to https://github.com/phpstan/phpstan-src/pull/4166


ondrejmirtes:
Thank you!

PR #4380, opened 2025-09-28
Cleanup `instanceof Type` checks

staabm:
extracted from https://github.com/phpstan/phpstan-src/pull/4372

### DIFF
--- a/src/Rules/UnusedFunctionParametersCheck.php
+++ b/src/Rules/UnusedFunctionParametersCheck.php
@@ -9,8 +9,6 @@ use PHPStan\DependencyInjection\AutowiredParameter;
 use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\ShouldNotHappenException;
-use function array_combine;
-use function array_map;
 use function array_merge;
 use function in_array;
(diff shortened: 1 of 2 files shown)


src/Rules/UnusedFunctionParametersCheck.php
```diff
@@ -43,20 +41,18 @@ public function getUnusedParameters(
...
 	{
-		$parameterNames = array_map(static function (Variable $variable): string {
+		$unusedParameters = [];
+		foreach ($parameterVars as $variable) {
 			if (!is_string($variable->name)) {
 				throw new ShouldNotHappenException();
 			}
-			return $variable->name;
-		}, $parameterVars);
-		$unusedParameters = array_combine($parameterNames, $parameterVars);
```
staabm, line 52: rewrite/simplifed this code, as it relied on `array_combine` which can return `false` on PHP7.4, which got caught by the MutatingScope improvement.

> Argument of an invalid type array<string, PhpParser\Node\Expr\Variable>|false supplied for foreach, only iterables are supported.

rewrote it to get rid of the PHP 7.4 specific behaviour so I don't need to add 7.4 only ignores or similar hacks.

ondrejmirtes:
Thank you!

PR #4381, opened 2025-09-28
Fix "Offset (int|string) might not exist on array" after `array_keys($arr)`

staabm:
found while investigating https://github.com/phpstan/phpstan-src/pull/4372

closes https://github.com/phpstan/phpstan/issues/12926

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -6521,6 +6521,23 @@ final class NodeScopeResolver
 			}
 		}
 
+		if (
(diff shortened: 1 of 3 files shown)


staabm:
interessting that issue-bot did not realize the related issue gets fixed

ondrejmirtes:
I think the [phpstanPlayground.configParameter](https://phpstan.org/error-identifiers/phpstanPlayground.configParameter) logic is not in the issue-bot

ondrejmirtes:
Thank you!

PR #4382, opened 2025-09-28
Added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/12115

similar as in https://github.com/phpstan/phpstan-src/pull/4381#issuecomment-3342694313 - issuebot did not realize its fixed

----

covers a few examples of https://github.com/phpstan/phpstan/issues/12574, but not all snippets there have been fixed


ondrejmirtes:
Thank you!

PR #4383, opened 2025-09-28
Fix "array_rand() - offset might not exists"

staabm:
closes https://github.com/phpstan/phpstan/issues/12981

when merging, we should extract https://github.com/phpstan/phpstan/issues/12981#issuecomment-2857342859 into a new issue

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -708,10 +708,9 @@ final class TypeSpecifier
 						&& ($numArg === null || $one->isSuperTypeOf($scope->getType($numArg))->yes())
 					) {
 						$dimFetch = new ArrayDimFetch($arrayArg, $expr->var);
-						$iterableValueType = $arrayType->getFirstIterableValueType();
 
 						return $specifiedTypes->unionWith(
-							$this->create($dimFetch, $iterableValueType, TypeSpecifierContext::createTrue(), $scope),
+							$this->create($dimFetch, $arrayType->getIterableValueType(), TypeSpecifierContext::createTrue(), $scope),
 						);
 					}
 				}
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -694,8 +694,11 @@ final class TypeSpecifier
 					&& in_array($expr->expr->name->toLowerString(), ['array_rand'], true)
 					&& count($expr->expr->getArgs()) >= 1
 				) {
+					$numArg = null;
 					$arrayArg = $expr->expr->getArgs()[0]->value;
-					$numArg = $expr->expr->getArgs()[1]->value;
+					if (count($expr->expr->getArgs()) > 1) {
+						$numArg = $expr->expr->getArgs()[1]->value;
+					}
 					$one = new ConstantIntegerType(1);
 					$arrayType = $scope->getType($arrayArg);
 
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -687,7 +687,7 @@ final class TypeSpecifier
 			if ($context->null()) {
 				$specifiedTypes = $this->specifyTypesInCondition($scope->exitFirstLevelStatements(), $expr->expr, $context)->setRootExpr($expr);
 
-				// infer $arr[$key] after $arr[array_rand($arr)]
+				// infer $arr[$key] after $key = array_rand($arr)
 				if (
 					$expr->expr instanceof FuncCall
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
```diff
@@ -942,6 +942,22 @@ public function testBug12593(): void
 		$this->analyse([__DIR__ . '/data/bug-12593.php'], []);
 	}
 
+	public function testBug12981(): void
+	{
+		$this->reportPossiblyNonexistentGeneralArrayOffset = true;
+
+		$this->analyse([__DIR__ . '/data/bug-12981.php'], [
+			[
+				'Offset array<int, int|string> might not exist on non-empty-array<bool|float|int|string>.',
```
VincentLanglet: I feel like a good improvement would be to not have this error reported because it's already reported as a `Invalid array key type`.

https://phpstan.org/r/b4c81787-7c0b-46b3-b281-dd572ab6f8ac

But that's another issue, I'll a look at it.

tests/PHPStan/Rules/Arrays/data/bug-12981.php
```diff
@@ -0,0 +1,35 @@
...
+	public function sayHello1(array $arr): void
+	{
+		$num = 1;
+		echo $arr[array_rand($arr, $num)];
+		$randIndex = array_rand($arr, $num);
+		echo $arr[$randIndex];
+	}
+
+	/** @param non-empty-array<scalar> $arr */
+	public function sayHello2(array $arr): void
```
VincentLanglet, line 28: You should also test 
```
     /** @param non-empty-array<scalar> $arr */
    public function sayHello3(array $arr, int $num): void
	{
		echo $arr[array_rand($arr, $num)];
		$randIndex = array_rand($arr, $num);
		echo $arr[$randIndex];
	}
```

tests/PHPStan/Rules/Arrays/NonexistentOffsetInArrayDimFetchRuleTest.php
VincentLanglet: WIP https://github.com/phpstan/phpstan-src/pull/4385

tests/PHPStan/Rules/Arrays/data/bug-12981.php
staabm: added, thanks.

src/Analyser/TypeSpecifier.php
```diff
@@ -687,6 +687,35 @@ public function specifyTypesInCondition(
...
+					$one = new ConstantIntegerType(1);
+					$arrayType = $scope->getType($arrayArg);
+
+					if (
+						$arrayType->isArray()->yes()
+						&& $arrayType->isIterableAtLeastOnce()->yes()
+						&& ($numArg === null || $one->isSuperTypeOf($scope->getType($numArg))->yes())
+					) {
+						$dimFetch = new ArrayDimFetch($arrayArg, $expr->var);
+						$iterableValueType = $arrayType->getFirstIterableValueType();
```
staabm: Memo to me: check this line again

VincentLanglet:
I think that if you rebase your PR @staabm your tests won't give the same output

VincentLanglet:
For the record, the fact
```
Offset array<int, int|string> might not exist on non-empty-array<bool|float|int|string>
```
is not reported anymore is expected.

The offset is already reported by the InvalidKey rule.

staabm:
> For the record, the fact
> 
> ```
> Offset array<int, int|string> might not exist on non-empty-array<bool|float|int|string>
> ```

thanks

> The offset is already reported by the InvalidKey rule.

I just put the test-cases additionally into `InvalidKeyInArrayDimFetchRuleTest` so we can see the overall picture.

error on line 39 und 41 still overlap, maybe thats something which can also be improved in a separate PR

VincentLanglet:
> error on line 39 und 41 still overlap, maybe thats something which can also be improved in a separate PR

There are not fully overlapping.
- `Possibly invalid array key type array<int, int|string>|int|string.` is about the invalid array key `array<int, int|string>`
- `Offset array<int, int|string>|int|string might not exist on non-empty-array<bool|float|int|string>.` is about the fact that `int|string might not exist on non-empty-array<bool|float|int|string>` because you enabled reportPossiblyNonexistentGeneralArrayOffset. I assume this error disappear with `reportPossiblyNonexistentGeneralArrayOffset = false`.

ondrejmirtes:
Thank you.

PR #4384, opened 2025-09-28
Inherit PHPDoc return type in child method with narrower native return type

ondrejmirtes:
Closes https://github.com/phpstan/phpstan/issues/12481
Closes https://github.com/phpstan/phpstan/issues/11364
Closes https://github.com/phpstan/phpstan/issues/10600
Closes https://github.com/phpstan/phpstan/issues/13588


ghost:
Thank you for this. It allows us to reduce a lot of rote PHPDoc.

PR #4385, opened 2025-09-28
Do not report non existent offset for invalid ones

VincentLanglet:
Need https://github.com/phpstan/phpstan-src/pull/4386
Need https://github.com/phpstan/phpstan-src/pull/4387

### DIFF
--- a/src/Rules/Arrays/NonexistentOffsetInArrayDimFetchCheck.php
+++ b/src/Rules/Arrays/NonexistentOffsetInArrayDimFetchCheck.php
@@ -12,7 +12,9 @@ use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeUtils;
 use PHPStan\Type\VerbosityLevel;
 use function count;
@@ -60,6 +62,14 @@ final class NonexistentOffsetInArrayDimFetchCheck
 		}
 
 		if ($type->hasOffsetValueType($dimType)->no()) {
+			if ($type->isArray()->yes()) {
+				$validArrayDimType = TypeCombinator::intersect(AllowedArrayKeysTypes::getType(), $dimType);
+				if ($validArrayDimType instanceof NeverType) {
+					// Already reported by InvalidKeyInArrayDimFetchRule
+					return [];
+				}
+			}
+
 			return [
 				RuleErrorBuilder::message(sprintf('Offset %s does not exist on %s.', $dimType->describe(count($dimType->getConstantStrings()) > 0 ? VerbosityLevel::precise() : VerbosityLevel::value()), $type->describe(VerbosityLevel::value())))
 					->identifier('offsetAccess.notFound')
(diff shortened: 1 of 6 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -586,6 +586,9 @@ public function hasOffsetValueType(Type $offsetType): TrinaryLogic
 		if ($offsetArrayKeyType instanceof ErrorType) {
 			$allowedArrayKeys = AllowedArrayKeysTypes::getType();
 			$offsetArrayKeyType = TypeCombinator::intersect($allowedArrayKeys, $offsetType)->toArrayKey();
+			if ($offsetArrayKeyType instanceof NeverType) {
```
VincentLanglet, line 589: I discovered that without this fix, hasOffsetValueType(new ArrayType()) was Yes.

staabm:
Did you double check the issue bot results are intentional?

VincentLanglet:
> Did you double check the issue bot results are intentional?

Yes, the issue bot is misleading, 
Currently there is already 23 errors on PHP 7.4 issue https://phpstan.org/r/f0a1d9d3-a829-4b6f-b21f-38295454a458 cause "mixed" is not a known type. I "add" 3 errors "Offset mixed does not exists" for the same reason.
But the snippet is not changed in PHP 8+

tests/PHPStan/Levels/data/arrayOffsetAccess-3.json
```diff
@@ -3,10 +3,5 @@
         "message": "Invalid array key type DateTimeImmutable.",
         "line": 17,
         "ignorable": true
-    },
-    {
-        "message": "Offset DateTimeImmutable does not exist on array.",
```
VincentLanglet, line 8: As expected, I reduce the duplicates from https://github.com/phpstan/phpstan-src/pull/4387

ondrejmirtes:
Thank you!

src/Rules/Arrays/NonexistentOffsetInArrayDimFetchCheck.php
```diff
@@ -76,28 +86,32 @@ public function check(
 				$flattenedTypes = TypeUtils::flattenTypes($type);
 			}
 
+			$validArrayDimType = TypeCombinator::intersect(AllowedArrayKeysTypes::getType(), $dimType);
+
 			foreach ($flattenedTypes as $innerType) {
+				$dimTypeToCheck = $innerType->isArray()->yes() ? $validArrayDimType : $dimType;
```
VincentLanglet, line 92: This only apply to array because we just have `InvalidKeyInArrayDimFetchRule`.

But you considered maybe extending the rule to String.

I wonder if in another PR we should create something like 
- `isOffsetValueTypeLegal(Type $offsetType, PHPVersion $phpVersion)` (with maybe a `strict` param...)
- `ArrayType::isOffsetValueTypeLegal rely on `AllowedArrayKeysTypes::getType`
- We use isOffsetValueTypeLegal in InvalidKeyInArrayDimFetchRule
- And we use it here to filter the offsetDim

Dunno... this topic does not seems to be easy ^^'

ondrejmirtes:
This PR causes this error to be reported on level < 9 which is a bug: https://phpstan.org/r/bbec48eb-1e0b-4c45-830a-46792cdfba3f

ondrejmirtes:
Reverting as I want to do a release.

VincentLanglet:
> This PR causes this error to be reported on level < 9 which is a bug: [phpstan.org/r/bbec48eb-1e0b-4c45-830a-46792cdfba3f](https://phpstan.org/r/bbec48eb-1e0b-4c45-830a-46792cdfba3f)

Indeed, MixedType shouldn't be intersected because it will transform it into an uniontype and lower the error level.
Fixed in https://github.com/phpstan/phpstan-src/pull/4398

PR #4386, opened 2025-09-28
Improve CombinationHelper phpdoc

VincentLanglet:
Extracted from https://github.com/phpstan/phpstan-src/pull/4385


ondrejmirtes:
Thank you.

PR #4387, opened 2025-09-28
Fix ConstantArrayType::hasOffsetValueType and ArrayType::hasOffsetValueType

VincentLanglet:
Extracted from https://github.com/phpstan/phpstan-src/pull/4385

When you merge this PR, I would recommend to wait for https://github.com/phpstan/phpstan-src/pull/4385 too in order to avoid reporting extra "duplicate" message "offset might not exist"/"offset does not exist" since we're detecting more errors with this fix.

I prefer to test hasOffsetValueType rather with errors reported by NonexistentOffsetInArrayDimFetchRule since
- https://github.com/phpstan/phpstan-src/pull/4385 will remove the message again
- But we should still check  hasOffsetValueType return NO, rather than YES (currently)


ondrejmirtes:
Thank you.

VincentLanglet:
Thanks, I rebased https://github.com/phpstan/phpstan-src/pull/4385

PR #4390, opened 2025-09-29
Skip deep expression when root expression changes while scope generalization

staabm:
implement the idea described in https://github.com/phpstan/phpstan-src/pull/4372#issuecomment-3345672871

it does not fix the problem we see the wrong types we see from https://phpstan.org/r/16197474-84cb-43ad-b983-ad78371fba02 though

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -162,12 +162,14 @@ use function is_string;
 use function ltrim;
 use function md5;
 use function sprintf;
+use function str_contains;
 use function str_decrement;
 use function str_increment;
 use function str_starts_with;
 use function strlen;
 use function strtolower;
 use function substr;
+use function uksort;
 use function usort;
 use const PHP_INT_MAX;
 use const PHP_INT_MIN;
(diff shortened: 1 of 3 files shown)


staabm:
from my POV the problem in the reported types is

```diff
-non-empty-array<int<0, 9>, non-empty-array<0|1, string>>
+non-empty-array<int<0, 9>, non-empty-array<int<0, 9>, string>>
```

so thats why [my idea](https://github.com/phpstan/phpstan-src/pull/4372#issuecomment-3345485920) was that in the deeper nested key-type no generalization happend and thats what missing.

IMO when only one of the involved scopes contains a general or constant array, we don't do deep generalization (recursive) in
- https://github.com/phpstan/phpstan-src/blob/c87456e54fce699905970533c398f8c56422f94d/src/Analyser/MutatingScope.php#L5395
- https://github.com/phpstan/phpstan-src/blob/c87456e54fce699905970533c398f8c56422f94d/src/Analyser/MutatingScope.php#L5436
- https://github.com/phpstan/phpstan-src/blob/c87456e54fce699905970533c398f8c56422f94d/src/Analyser/MutatingScope.php#L5441
- https://github.com/phpstan/phpstan-src/blob/c87456e54fce699905970533c398f8c56422f94d/src/Analyser/MutatingScope.php#L5481

(but I am looking at this stuff for the first time... I might be wrong :))

ondrejmirtes:
Yeah, it's possible the problem is with generalization and not merging scopes, I was just debugging this in my head, not in code 😂

staabm:
after more debugging I have a new idea :)

it seems we update types for expressions, but do not update other types which reference these variables by offset, see e.g.

grafik

I think `$locations[$i]` indirectly depends on `$j` in its key-type of the array

ondrejmirtes:
Yeah, that's basically what I'm talking about in https://github.com/phpstan/phpstan-src/pull/4372#issuecomment-3345672871, this is what I imagine the problem would look like.

After updating `$i`, we need to throw away `$locations[$i]`. After updating `$locations`, we need to throw away `$locations[$i]`.

I think that actually "updating" after assignment works fine, but merging scopes probably doesn't handle this.

staabm:
I finally understood what you are trying to tell me :).

ondrejmirtes:
I'm also trying to figure out the other regression.

staabm:
thank you!

ondrejmirtes:
I don't know yet, will think about it more :) Thank you.

PR #4392, opened 2025-09-30
Configure PromoteParameterRule for issue-bot

staabm:
issue-bot currently misses these errors/fixes, see https://github.com/phpstan/phpstan-src/pull/4381#issuecomment-3342694313

1:1 copy of the config from https://github.com/phpstan/phpstan/blob/2.1.x/playground-runner/playground.neon

### DIFF
--- a/issue-bot/playground.neon
+++ b/issue-bot/playground.neon
@@ -9,73 +9,10 @@ conditionalTags:
 	PHPStan\Rules\Playground\StaticVarWithoutTypeRule:
 		phpstan.rules.rule: %checkImplicitMixed%
 
+parameters:
+	reportPossiblyNonexistentGeneralArrayOffset: true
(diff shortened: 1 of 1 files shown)


staabm:
hmm thinking again.. maybe we should just enable the optional rules in issue bot..?

or maybe even just enable bleeding edge?

staabm:
.. hmm reflecting the playground settings 1:1 seems to be the most intuive fix, which is also reasonable for the end-user (least suprising)

ondrejmirtes:
1) I didn't do this on purpose because I'm afraid that dozens, or maybe even hundreds of comments would spam the issue tracker when this would get enabled.
2) You don't see any changes now because these rules are also filtered in code https://github.com/phpstan/phpstan-src/blob/4972fd927d3cb1097302f3638b938fcc06d647a8/issue-bot/src/Playground/TabCreator.php#L31

PR #4393, opened 2025-10-01
Added regression tests

staabm:
covers fixes we see in https://github.com/phpstan/phpstan-src/actions/runs/18133279748

closes https://github.com/phpstan/phpstan/issues/6173
closes https://github.com/phpstan/phpstan/issues/10640
closes https://github.com/phpstan/phpstan/issues/10025 
closes https://github.com/phpstan/phpstan/issues/12078


staabm:
@ondrejmirtes the nsrt tests reveal some interessting `*ERROR*` types in https://github.com/phpstan/phpstan-src/actions/runs/18154762119/job/51672069236?pr=4393

(I did not yet make sure the assert-types are super accurate, but as soon as I noticed we see some `*ERROR*` types, just added some statements so we can see possible more `*ERROR*`s)

see e.g. nsrt/bug-10640.php

staabm:
ok, scratch the above comment. I somehow mixed it up. should be good to go now.

ondrejmirtes:
Thank you!

PR #4395, opened 2025-10-01
Validate curl_setopt_array parameter array

staabm:
followup to https://github.com/phpstan/phpstan-src/pull/1719

closes https://github.com/phpstan/phpstan/issues/13615

PR #4396, opened 2025-10-01
Support union types in curl_setopt validation

staabm:
while working on https://github.com/phpstan/phpstan-src/pull/4395 I realized there is a low hanging fruit by getting rid of `instanceof ConstantIntegerType`

### DIFF
--- a/src/Reflection/ParametersAcceptorSelector.php
+++ b/src/Reflection/ParametersAcceptorSelector.php
@@ -128,35 +128,30 @@ final class ParametersAcceptorSelector
 			if (count($args) >= 3 && (bool) $args[0]->getAttribute(CurlSetOptArgVisitor::ATTRIBUTE_NAME)) {
 				$optType = $scope->getType($args[1]->value);
 
-				$optValueType = null;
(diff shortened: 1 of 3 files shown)


src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -126,33 +127,51 @@ public static function selectFromArgs(
...
+						$optValueType = null;
+						break;
+					}
+
+					if ($optValueType === null) {
+						$optValueType = $valueType;
+						continue;
 					}
+
+					$optValueType = TypeCombinator::union($optValueType, $valueType);
```
ondrejmirtes: It'd be more performant to call `::union` just once once we gather all the types to union.

ondrejmirtes (changes requested):

src/Reflection/ParametersAcceptorSelector.php
staabm: here you are

ondrejmirtes:
Thank you!

PR #4397, opened 2025-10-02
array_rand() requires a non-empty-array as of php8

staabm:
```
Fatal error: Uncaught ValueError: array_rand(): Argument #1 ($array) must not be empty in /in/sEB2j:3
Stack trace:
#0 /in/sEB2j(3): array_rand(Array)
#1 {main}
  thrown in /in/sEB2j on line 3
```

see https://3v4l.org/sEB2j

---

we did not error before this PR: https://phpstan.org/r/2f650341-3dc9-4849-9cac-2bd267a7fb41


staabm:
it even requires a positve $num arg as of php8, see https://3v4l.org/b53OZ

resources/functionMap_php82delta.php
```diff
@@ -24,6 +24,7 @@
 		'iterator_count' => ['0|positive-int', 'iterator'=>'iterable'],
 		'iterator_to_array' => ['array', 'iterator'=>'iterable', 'use_keys='=>'bool'],
 		'str_split' => ['list<string>', 'str'=>'string', 'split_length='=>'positive-int'],
+		'Random\Randomizer::pickArrayKeys' => ['non-empty-array<int|string>', 'array'=>'non-empty-array', 'num'=>'positive-int'],
```
staabm, line 27: not sure whether/how this signature should be behind bleeding edge but beeing php 8.2+ at the same time.

as the api is pretty recent and throws runtime errors when used with empty array or non-positive-int its acceptable to enforce this more strict signature even in non-bleeding edge?

ondrejmirtes: Yeah, in this case it's okay 👍

ondrejmirtes:
Thank you!

PR #4398, opened 2025-10-02
Do not report non existent offset for invalid ones

VincentLanglet:
This reverts commit b0aa47e.

And add a fix for the error reported in https://github.com/phpstan/phpstan-src/pull/4385#issuecomment-3361967155


ondrejmirtes:
Thank you.

PR #4400, opened 2025-10-03
relax strict printf parameter check

schlndh:
I relaxed the strict printf parameter check as described in [this comment](https://github.com/phpstan/phpstan-src/pull/4349#discussion_r2381919959).

IMO `checkStrictPrintfPlaceholderTypes: true` should mean precisely the level of strictness that's appropriate for phpstan-strict-rules.

Beyond that it may be desirable to configure it to be more/less strict. But for now, those who want it less strict will have to set it to `false` and they will be no worse off than before I introduced it. Those who want it more strict can keep it turned on and they'll be at least somewhat better off than before I introduced it.


ondrejmirtes:
Thank you!

PR #4401, opened 2025-10-03
Add option to prohibit bool to string coercion

schlndh:
I'd like to know your opinion on this approach before I go too far with it.

It's a follow up to [this comment](https://github.com/phpstan/phpstan-src/pull/4349#discussion_r2381919959) (specifically the part about disabling "{$bool}" etc in phpstan-strict-rules).

The idea is to eventually (separate PRs) add all pairs of types which make sense to prohibit (e.g. it makes sense to prohibit coercing `null` to `string` (not necessarily in phpstan-strict-rules), but it probably doesn't make sense to prohibit coercing `int` to `string`). Note that the configuration is only meant to prohibit type coercions (i.e. implicit), not explicit type casts. This could then replace things like:

- `booleansInConditions` and `numericOperandsInArithmeticOperators` from phpstan-strict-rules: the rules would be moved to phpstan and they'd be enabled by `allowedTypeCoercions.XToY`.
- `forbidNullInInterpolatedString` and `forbidNullInBinaryOperations` from [shipmonk/phpstan-rules](https://github.com/shipmonk-rnd/phpstan-rules): it would be handled by existing phpstan rules and enabled by `allowedTypeCoercions.nullToY`.
- `checkStrictPrintfPlaceholderTypes`: it would be removed and instead it would be configured by `allowedTypeCoercions.XToY`. phpstan-strict-rules would set the flags to preserve the current default behavior.

The benefit of this approach is that it would be configured in one place and applied consistently across the rules. The downside is that it would make it harder to enable `XToY` in one rule, but prohibit it in another. Given the discussion in the strict printf parameter check PR, I guess it's safe to assume that there will be users with strong opinions on how it should behave for them. The DI container could pass a different `TypeCoercionRuleHelper` to every rule, so it has a possible solution, but there would have to be a sane way to configure it (ultimately the `RuleLevelHelper` has the same limitation).

### DIFF
--- a/conf/bleedingEdge.neon
+++ b/conf/bleedingEdge.neon
@@ -14,3 +14,4 @@ parameters:
 		rawMessageInBaseline: true
 		reportNestedTooWideType: false
 		assignToByRefForeachExpr: true
+		checkTypeCoercions: true
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -38,6 +38,7 @@ parameters:
 		rawMessageInBaseline: false
 		reportNestedTooWideType: false
 		assignToByRefForeachExpr: false
+		checkTypeCoercions: false
 	fileExtensions:
 		- php
 	checkAdvancedIsset: false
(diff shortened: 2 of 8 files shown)


ondrejmirtes:
I'd like this first https://github.com/phpstan/phpstan/issues/13008 then it would be much easier to build rules for that

PR #4403, opened 2025-10-03
Fix "list type lost in for loop"

staabm:
if the for loop is in the shape of `for($i = 0; $i < count($items); $i++) {` we can set `$items[$i]` as existing expression type.

closes https://github.com/phpstan/phpstan/issues/12807

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1487,6 +1487,8 @@ final class NodeScopeResolver
 				$initScope = $condResult->getScope();
 				$condResultScope = $condResult->getScope();
 
+				// only the last condition expression is relevant whether the loop continues
+				// see https://www.php.net/manual/en/control-structures.for.php
 				if ($condExpr === $lastCondExpr) {
 					$condTruthiness = ($this->treatPhpDocTypesAsCertain ? $condResultScope->getType($condExpr) : $condResultScope->getNativeType($condExpr))->toBoolean();
 					$isIterableAtLeastOnce = $isIterableAtLeastOnce->and($condTruthiness->isTrue());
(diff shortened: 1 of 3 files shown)


staabm:
we could make this work for more involved loops, but I think this will work as a first step

--

other examples
- `for ($i = count($items) - 1; $i >= 0; --$i) {`
- `for ($i = count($items) - 1; $i >= 0; $i--) {`
- `for ($i = count($items) - 1; $i > -1; $i--) {`
- `for ($i = count($disk)-1, $f = 0; $i >= 0; $i--) {`
- `for ($i = count($bytes)-2; $i>=0; $i--)`
- `for ($i = 0; $i < count($items); $i+=4) {`

src/Analyser/NodeScopeResolver.php
```diff
@@ -7116,4 +7120,66 @@ public function getFilteringExprForMatchArm(Expr\Match_ $expr, array $conditions
...
+	{
+		// infer $items[$i] type from for ($i = 0; $i < count($items); $i++) {...}
+
+		if (
+			// $i = 0
+			count($stmt->init) === 1
+			&& $stmt->init[0] instanceof Assign
+			&& $stmt->init[0]->var instanceof Variable
+			&& $stmt->init[0]->expr instanceof Node\Scalar\Int_
+			&& $stmt->init[0]->expr->value === 0
```
staabm, line 7133: I think we could do `>=0`

ondrejmirtes:
Thank you!

staabm:
any opinions on the other given examples? do you think this can work and should be implemented?

ondrejmirtes:
Instead of writing many many conditions, I'd like if this could be expressed more by the type system. That would cover all situations. But I'm not sure yet how it could look.

staabm:
ok thanks. no priority on it.

PR #4406, opened 2025-10-04
Report float and null offset based on PHP version

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13592

We will still need an option to report array key cast, especially for boolean keys.

Partially solves https://github.com/phpstan/phpstan/issues/7864

### DIFF
--- a/src/Rules/Arrays/AllowedArrayKeysTypes.php
+++ b/src/Rules/Arrays/AllowedArrayKeysTypes.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Rules\Arrays;
 
+use PHPStan\Php\PhpVersion;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\Constant\ConstantBooleanType;
@@ -21,15 +22,22 @@ use PHPStan\Type\UnionType;
 final class AllowedArrayKeysTypes
 {
 
-	public static function getType(): Type
+	public static function getType(?PhpVersion $phpVersion = null): Type
 	{
-		return new UnionType([
+		$types = [
 			new IntegerType(),
 			new StringType(),
-			new FloatType(),
 			new BooleanType(),
-			new NullType(),
-		]);
+		];
+
+		if ($phpVersion === null || !$phpVersion->deprecatesImplicitlyFloatConversionToInt()) {
+			$types[] = new FloatType();
+		}
+		if ($phpVersion === null || !$phpVersion->deprecatesNullArrayOffset()) {
+			$types[] = new NullType();
+		}
+
+		return new UnionType($types);
 	}
 
 	public static function narrowOffsetKeyType(Type $varType, Type $keyType): ?Type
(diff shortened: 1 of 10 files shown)


tests/PHPStan/Rules/Arrays/data/invalid-key-array-dim-fetch.php
```diff
@@ -3,18 +3,18 @@
 namespace InvalidKeyArrayDimFetch;
 
 $a = [];
-$foo = $a[null];
```
VincentLanglet, line 6: I moved PHP-related errors to the end for an easier handling in tests.

tests/PHPStan/Rules/Arrays/data/invalid-key-array-dim-fetch.php
```diff
@@ -3,18 +3,18 @@
...
 $foo = $a[new \DateTimeImmutable()];
 $a[[]] = $foo;
 $a[1];
-$a[1.0];
+
 $a['1'];
 $a[true];
 $a[false];
 
-/** @var string|null $stringOrNull */
```
VincentLanglet, line 15: I changed for a valid union here and move the invalid string|null to the end

VincentLanglet:
BaselineNeonErrorFormatterIntegrationTest::testErrorWithTrait failure seems unrelated and doesn't fail locally.

src/Rules/Arrays/AllowedArrayKeysTypes.php
```diff
@@ -21,15 +22,22 @@
 final class AllowedArrayKeysTypes
 {
 
-	public static function getType(): Type
+	public static function getType(?PhpVersion $phpVersion = null): Type
```
ondrejmirtes, line 25: Why optional and nullable?

src/Rules/Arrays/InvalidKeyInArrayItemRule.php
```diff
@@ -22,6 +23,7 @@ final class InvalidKeyInArrayItemRule implements Rule
 
 	public function __construct(
 		private RuleLevelHelper $ruleLevelHelper,
+		private PhpVersion $phpVersion,
```
ondrejmirtes, line 26: More modern approach is to get PhpVersions from Scope (a separate class).

ondrejmirtes (changes requested):

src/Rules/Arrays/AllowedArrayKeysTypes.php
VincentLanglet: It's used in ArrayType and ConstantArrayType
https://github.com/phpstan/phpstan-src/blob/2548dbbffc85f6f310aa506ed3540e1bbb702f8a/src/Type/ArrayType.php#L273

and I don't have a PHPVersion here

src/Rules/Arrays/InvalidKeyInArrayItemRule.php
VincentLanglet: Does it means that all the method from PhpVersion should be deprecated/moved into PhpVersions ?

ondrejmirtes: Not yet

ondrejmirtes:
Thank you!

PR #4407, opened 2025-10-04
Fix oversizeArrayType->isSubTypeOf

VincentLanglet:
Before `oversizeArrayType->isSubTypeOf(array{})` was considered to be YES.

So removing `array{}` to oversizeArray gives Never.

Closes https://github.com/phpstan/phpstan/issues/13509


ondrejmirtes:
Thank you.

PR #4409, opened 2025-10-04
Mark redis method as impure

VincentLanglet:
When looking at Redis methods, there is a lot https://github.com/JetBrains/phpstorm-stubs/blob/d1ee5e570343bd4276a3d5959e6e1c2530b006d0/redis/Redis.php
and I think all of them should be considered as impure/having-side-effect, because they are
- Either getter which rely on the database so should be impure
- Either setter which modify the database so have side effect
- Either are both a setter and a getter

I tried to list the one which should have a return value which might change between two calls, for instance because they rely on one key and another redis instance might change the key value between two calls.

I may have forgotten some but I think it can be considered a good start, and let people update it or report new issues.

If this PR is accepted, I'll can do the same for RedisArray and RedisCluster

Closes https://github.com/phpstan/phpstan/issues/10215

Kinda similar to https://github.com/phpstan/phpstan-src/pull/4344


ondrejmirtes:
Thank you.

PR #4410, opened 2025-10-05
Improve support of FILTER_REQUIRE_ARRAY

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11863

After this merge I'll work on https://github.com/phpstan/phpstan/issues/11485
I think it will require to change `hasFlag` signature to return a TrinaryLogic

cc @herndlm

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/filter-var.php
+++ b/tests/PHPStan/Analyser/nsrt/filter-var.php
@@ -23,10 +23,10 @@ class FilterVar
 		assertType('null', filter_var(17, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
 		assertType('false', filter_var('foo', FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
 		assertType('null', filter_var('foo', FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
-		assertType('array<int>|false', filter_var($mixed, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
-		assertType('array<int>|null', filter_var($mixed, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
-		assertType('array<string, int>|false', filter_var($stringMixedMap, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Analyser/nsrt/filter-var.php
```diff
@@ -23,10 +23,10 @@ public function doFoo($mixed, array $stringMixedMap): void
 		assertType('null', filter_var(17, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
 		assertType('false', filter_var('foo', FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
 		assertType('null', filter_var('foo', FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
-		assertType('array<int>|false', filter_var($mixed, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
-		assertType('array<int>|null', filter_var($mixed, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
-		assertType('array<string, int>|false', filter_var($stringMixedMap, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
-		assertType('array<string, int>|null', filter_var($stringMixedMap, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
+		assertType('array<int|false>|false', filter_var($mixed, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
```
VincentLanglet, line 26: mixed can either be
- array with valid values
- array with invalid values (will be array<false>)
- non-array (will be false)

tests/PHPStan/Analyser/nsrt/filter-var.php
```diff
@@ -23,10 +23,10 @@ public function doFoo($mixed, array $stringMixedMap): void
...
+		assertType('array<int|null>|null', filter_var($mixed, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY|FILTER_NULL_ON_FAILURE]));
+		assertType('array<string, int|false>', filter_var($stringMixedMap, FILTER_VALIDATE_INT, ['flags' => FILTER_REQUIRE_ARRAY]));
```
VincentLanglet, line 28: Here we always have an array so the result cannot be false (but an array with false values)

ondrejmirtes:
Thank you.

PR #4414, opened 2025-10-06
Add ArrayCombineFunctionThrowTypeExtension

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13642

### DIFF
--- a/src/Type/Php/ArrayCombineFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayCombineFunctionReturnTypeExtension.php
@@ -20,7 +20,7 @@ final class ArrayCombineFunctionReturnTypeExtension implements DynamicFunctionRe
 
 	public function __construct(
 		private ArrayCombineHelper $arrayCombineHelper,
-		private PhpVersion $phpVersion
+		private PhpVersion $phpVersion,
 	)
 	{
 	}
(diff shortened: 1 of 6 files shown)


src/Type/Php/ArrayCombineFunctionReturnTypeExtension.php
```diff
@@ -48,119 +40,24 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-				}
-
-				$results[] = $builder->getArray();
-			}
-
-			if ($results !== []) {
-				return TypeCombinator::union(...$results);
-			}
+		[$returnType, $hasValueError] = $this->arrayCombineHelper->getReturnAndThrowType($firstArg, $secondArg, $scope);
+		if ($hasValueError->no()) {
```
VincentLanglet, line 44: If $hasValueError is MAYBE and `throwsValueErrorForInternalFunctions` is FALSE, we have to add `new ConstantBooleanType(false)`.

The mutator think there is an escaped mutant because it only runs on PHP >= 80000

src/Type/Php/ArrayCombineFunctionThrowTypeExtension.php
```diff
@@ -0,0 +1,43 @@
...
+	{
+		if (count($funcCall->getArgs()) < 2) {
+			return $functionReflection->getThrowType();
+		}
+
+		$firstArg = $funcCall->getArgs()[0]->value;
+		$secondArg = $funcCall->getArgs()[1]->value;
+
+		$hasValueError = $this->arrayCombineHelper->getReturnAndThrowType($firstArg, $secondArg, $scope)[1];
+		if (!$hasValueError->no()) {
```
VincentLanglet, line 36: I don't understand the "escaped mutant" since `if ($hasValueError->yes()) {` is failing locally

src/Type/Php/ArrayCombineHelper.php
```diff
@@ -0,0 +1,141 @@
...
+use function is_string;
+
+#[AutowiredService]
+final class ArrayCombineHelper
+{
+
+	/**
+	 * @return array{Type, TrinaryLogic} The return type and if a ValueError may occur on PHP8 (and a warning on PHP7).
+	 */
+	public function getReturnAndThrowType(Expr $firstArg, Expr $secondArg, Scope $scope): array
```
VincentLanglet, line 31: I just basically extracted the existing code from ArrayCombineFunctionReturnTypeExtension ; solving the existing mutants are out of the scope of this PR.

src/Type/Php/ArrayCombineFunctionThrowTypeExtension.php
VincentLanglet: cc @staabm since I think you introduced those mutant checks

staabm: if you are getting a mutant for `TrinaryLogic->yes()`, `!TrinaryLogic->no()` or the inverses of it, it means you do not have a test for `TrinaryLogic->maybe()`.

btw: there can be situations where we need to accept a mutant, because it might be not a valid combination of things or its untestable (or infection is not smart enough to detect a certain situation is logically impossible)

VincentLanglet: > if you are getting a mutant for TrinaryLogic->yes(), !TrinaryLogic->no() or the inverses of it, it means you do not have a test for TrinaryLogic->maybe().

I meant that I when I tried the mutation locally yesterday, the test were failing so I were surprised it was considered as an escaped mutation... Cause `mixed` is MAYBE so it should be already tested.

staabm (approved):
lgtm

PR #4415, opened 2025-10-06
Mark RedisCluster method as impure

VincentLanglet:
Similar to https://github.com/phpstan/phpstan-src/pull/4409


ondrejmirtes:
Would you want to add something like `@phpstan-all-methods-impure` and pure above classes? It'd be more maintainable than this.

It'd still be possible to exclude a method and do the opposite above it.

VincentLanglet:
> Would you want to add something like `@phpstan-all-methods-impure` and pure above classes? It'd be more maintainable than this.
> 
> It'd still be possible to exclude a method and do the opposite above it.

Like this https://github.com/phpstan/phpdoc-parser/pull/277 ?

VincentLanglet:
Closing in favor of https://github.com/phpstan/phpstan-src/pull/4422

PR #4416, opened 2025-10-07
Quality of life improvement, PHPStorm error identifiers

NickSdot:
ref: https://github.com/phpstan/phpstan/discussions/13646

I left the actual source attribute untouched in case someone is using it for any other tools. If Jetbrains ever gets this sorted on their end we can revert the change without affecting anything else.

Edit: failures seem unrelated

### DIFF
--- a/src/Command/ErrorFormatter/CheckstyleErrorFormatter.php
+++ b/src/Command/ErrorFormatter/CheckstyleErrorFormatter.php
@@ -44,15 +44,13 @@ final class CheckstyleErrorFormatter implements ErrorFormatter
 
 			foreach ($errors as $error) {
 
-				$identifier = null !== $error->getIdentifier()
-					? "  // @phpstan-ignore {$error->getIdentifier()}" :
-					'';
+				$identifier = $error->getIdentifier();
 
 				$output->writeRaw(sprintf(
 					'  <error line="%d" column="1" severity="error" message="%s"%s />',
 					$this->escape((string) $error->getLine()),
-					$this->escape($error->getMessage() . $identifier),
-					$error->getIdentifier() !== null ? sprintf(' source="%s"', $this->escape($error->getIdentifier())) : '',
+					$this->escape($error->getMessage() . ($identifier !== null ? sprintf('  // @phpstan-ignore %s', $identifier) : '')),
+					$identifier !== null ? sprintf(' source="%s"', $this->escape($identifier)) : '',
 				));
 				$output->writeLineFormatted('');
 			}
--- a/src/Command/ErrorFormatter/CheckstyleErrorFormatter.php
+++ b/src/Command/ErrorFormatter/CheckstyleErrorFormatter.php
@@ -43,10 +43,15 @@ final class CheckstyleErrorFormatter implements ErrorFormatter
 			$output->writeLineFormatted('');
 
 			foreach ($errors as $error) {
+
+				$identifier = null !== $error->getIdentifier()
+					? "  // @phpstan-ignore {$error->getIdentifier()}" :
+					'';
+
 				$output->writeRaw(sprintf(
 					'  <error line="%d" column="1" severity="error" message="%s"%s />',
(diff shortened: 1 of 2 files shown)


src/Command/ErrorFormatter/CheckstyleErrorFormatter.php
```diff
@@ -43,11 +43,14 @@ public function formatErrors(
...
 			foreach ($errors as $error) {
+
+				$identifier = $error->getIdentifier();
+
 				$output->writeRaw(sprintf(
 					'  <error line="%d" column="1" severity="error" message="%s"%s />',
 					$this->escape((string) $error->getLine()),
-					$this->escape($error->getMessage()),
-					$error->getIdentifier() !== null ? sprintf(' source="%s"', $this->escape($error->getIdentifier())) : '',
+					$this->escape($error->getMessage() . ($identifier !== null ? sprintf('  // @phpstan-ignore %s', $identifier) : '')),
```
ondrejmirtes, line 52: 1) I don't want the error message to contain phpstan-ignore, just the identifier
2) I don't want the error message to change for everyone, that's creating unnecessary noise. Can we detect we're running inside the PhpStorm plugin?

ondrejmirtes (changes requested):

src/Command/ErrorFormatter/CheckstyleErrorFormatter.php
NickSdot: 1) Sure, I can do that. Allow me to add that it is super convinient to be able to just copy&paste the full comment. Your call.

image


2) Not really. I was thinking about that. There is no PHPStorm native env var available. One hacky way to achieve an opt-in like behaviour is the following:

- create phpstorm.php somewhere with content
```php
<?php

putenv('PHPSTAN_IDE=PHP_STORM');
```

- load it as bootstrap file
```neon
parameters:
	bootstrapFiles:
	- ./Foo/phpstorm.php
```

Then I have it available in the CheckstyleErrorFormatter. 

---

But honestly, this feels too hacky. At this point it is probably better to just introduce `CheckstyleErrorFormatterPhpStorm` and then people can opt-in the usual way?

```neon
errorFormatter.checkstyle:
class: Foo\CheckstyleErrorFormatterPhpStorm
```

ondrejmirtes: Honestly I'm not too keen on fixing problems of a different commercial product on my side because they're not giving priority to work on their plugin, creating more work for me.

I'd rather have my own plugin that offers the best experience for PhpStorm users, probably offered as part of PHPStan Pro subscription.

NickSdot: Yes, I am with you. Totally understandable. Also quite disappointing that @jetbrains leaves small issues like this open for such long time. Maybe @pronskiy or @brendt can get something rolling on their end.

Closing here. When official plugin? 🚀

ondrejmirtes: Maybe I'll vibecode it 🪄

staabm: Phpstorm can be detected, see https://github.com/phpstan/phpstan-src/pull/4369/files

NickSdot: Thanks @staabm but that apparenly doesn't work in the inspection; terminal only. Tried, not available.

canvural: Isn't that only for the integrated terminal? PHPStan in PHPStorm works as a plugin.

NickSdot: I wrapped it up as a composer package real quick `composer require nicksdot/phpstan-phpstorm-error-identifiers`

PR #4417, opened 2025-10-07
`!array_key_exists()` should imply `array` for PHP8+

staabm:
closes https://github.com/phpstan/phpstan/issues/6209
closes https://github.com/phpstan/phpstan/issues/13301

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-2001-php8.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-2001-php8.php
@@ -1,6 +1,6 @@
-<?php // lint < 8.0
+<?php // lint >= 8.0
 
-namespace Bug2001;
+namespace Bug2001Php8;
 
 use function PHPStan\Testing\assertType;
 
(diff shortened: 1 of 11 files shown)


tests/PHPStan/Analyser/nsrt/bug-2001-php8.php
```diff
@@ -0,0 +1,51 @@
...
+	{
+		$parsedUrl = parse_url(urldecode($url));
+		assertType('array{scheme?: string, host?: string, port?: int<0, 65535>, user?: string, pass?: string, path?: string, query?: string, fragment?: string}|false', $parsedUrl);
+
+		if (array_key_exists('host', $parsedUrl)) {
+			assertType('array{scheme?: string, host: string, port?: int<0, 65535>, user?: string, pass?: string, path?: string, query?: string, fragment?: string}', $parsedUrl);
+			throw new \RuntimeException('Absolute URLs are prohibited for the redirectTo parameter.');
+		}
+
+		assertType('array{scheme?: string, host?: string, port?: int<0, 65535>, user?: string, pass?: string, path?: string, query?: string, fragment?: string}', $parsedUrl);
```
staabm: in this case we got more preicse, because we no longer have a union `array{...}|false` but just the plain array-shape.
but on the other hand we no longer know that the `'host'` offset cannot exist at this point.

I tried several variants, but wasn't successfull in combining the specified types in a way which turns the mixed into array while also removing the offset.

staabm: I figured it out

staabm:
Memo for me: add test for the issue-bot report

_Edit_: ✅

ondrejmirtes:
Thank you!

PR #4419, opened 2025-10-07
Change return type of apcu_inc & apcu_dec function

LordSimal:
See https://www.php.net/manual/en/function.apcu-inc.php


resources/functionMap.php
```diff
@@ -145,7 +145,7 @@
 'apcu_exists' => ['bool', 'keys'=>'string'],
 'apcu_exists\'1' => ['array', 'keys'=>'string[]'],
 'apcu_fetch' => ['mixed', 'key'=>'string|string[]', '&w_success='=>'bool'],
-'apcu_inc' => ['int', 'key'=>'string', 'step='=>'int', '&w_success='=>'bool', 'ttl='=>'int'],
+'apcu_inc' => ['false|int', 'key'=>'string', 'step='=>'int', '&w_success='=>'bool', 'ttl='=>'int'],
```
staabm, line 148: same problem above in  `apc_dec `

LordSimal:
So... is there something i need to do to get this merged?

staabm:
it looks good to me. ondrej is currently focused on php 8.5 support.

just leave this here as is for now. I think he will come back to it when time allows.

thg2k:
I can confirm this change is correct in all apcu versions.

ondrejmirtes:
Thank you!

PR #4422, opened 2025-10-08
Introduce all-methods-are-(im)pure

VincentLanglet:
Rework of https://github.com/phpstan/phpstan-src/pull/4409
and https://github.com/phpstan/phpstan-src/pull/4415

I added a non regression of https://github.com/phpstan/phpstan/issues/10215 as a proof it works.

I don't like the naming of my methods, suggestion are opened.
Also I'm not fully sure `getDefaultMethodPurity` shouldn't be called somewhere else too...


VincentLanglet:
I dunno which strategy you will prefer @ondrejmirtes between
- doing calls like `isPure() ?? getDefaultMethodPurity()` like https://github.com/phpstan/phpstan-src/pull/4422/commits/846d7bd2d2bc0e5db775f981e958a5c2873aa11a#diff-fb5399c56567ad1b8be9256fe9008a03a43b7686e02156b9f1b3118e76d95e2fR895

- or using the fallback directly in the `isPure` method like https://github.com/phpstan/phpstan-src/pull/4422/files#diff-260042b10e8fde4d3e77fce0c1fc6fcf0851c7e442271dbefb5eded974d607e6R837-R843

The second one seems simpler to me because `isPure` was used multiple times like in `NodeScopeResolver::getPhpDocs` and it would have require more changes.

src/PhpDoc/ResolvedPhpDocBlock.php
```diff
@@ -455,6 +458,16 @@ public function getNullableNameScope(): ?NameScope
 		return $this->nameScope;
 	}
 
+	private function getClassReflection(): ?ClassReflection
+	{
+		$className = $this->nameScope?->getClassName();
```
staabm: The phpstan downgrade process does not support null-safe calls. You need regular if-null checks for php7 compat

ondrejmirtes (changes requested):
I think that ResolvedPhpDocBlock should simply add `isAllMethodsPure` and `isAllMethodsImpure`.

I imagine the new tags should be handled similarly to `isImmutable` / `isReadOnly`.

Above class there can be `@phpstan-immutable` tag which marks all properties as readonly. And above properties you can have `@phpstan-readonly`.

Having `@phpstan-immutable` above class does not change what ResolvedPhpDocBlock says about isReadOnly.

Instead, PhpClassReflectionExtension handles this when creating properties by asking about both tags: https://github.com/phpstan/phpstan-src/blob/d78a5d86c3c00a62790be81d65355d4ca344ff39/src/Reflection/Php/PhpClassReflectionExtension.php#L291

So handling these new tags should be equivalent.

Please do not forget to test that PureMethodRule still correctly reports errors with all possible combinations. This means that these new tags also have to be read in `NodeScopeResolver::getPhpDocs()` so that they're available in PhpMethodFromParserNodeReflection (which is the object used when analysing method body itself).

VincentLanglet:
I think it's ready for a new review :)

tests/PHPStan/Rules/Pure/PureMethodRuleTest.php
```diff
@@ -216,6 +216,37 @@ public function testBug12224(): void
 		]);
 	}
 
+	public function testAllMethodsArePure(): void
+	{
+		$this->treatPhpDocTypesAsCertain = true;
+		$this->analyse([__DIR__ . '/data/all-methods-are-pure.php'], [
+			[
+				'Method AllMethodsArePure\Foo::test() is marked as pure but returns void.',
```
ondrejmirtes: I think that `void` methods (except the constructor) should be excluded from `@phpstan-all-methods-pure` by default.

By looking at logic in PhpMethodReflection::hasSideEffects you can see that PHPDocs don't affect that at all.

ondrejmirtes (changes requested):
Also I'd like some more tests:

1) Calling `all-methods-pure` + nothing method from impure method (should report that impure method has no side effects)
2) Calling `all-methods-pure` + 'impure' method from impure method (no errors)
3) Calling `all-methods-pure` + nothing method from pure method (no errors)
4) Calling `all-methods-pure` + 'impure' method from pure method (should report a side effect in a pure method)

5) Calling `all-methods-impure` + nothing method from pure method (should report a side effect in a pure method)
6) Calling `all-methods-impure` + 'pure' method from pure method (no errors)
7) Calling `all-methods-impure` + nothing method from impure method (no errors)
8) Calling `all-methods-impure` + 'pure' method from impure method (should report that impure method has no side effects)

VincentLanglet:
Lot of test added and void behavior :)

VincentLanglet:
@staabm Do you see a case where 
```
$scope->getFunctionType($node->getReturnType(), false, false)->isVoid()->maybe()
```
is possible or does the mutation testing reports a useless mutant ?

staabm:
I did not look into all details of this PR, but maybe you can do something like

```php

class X {
  public function returnsVoid(): void{}
  public function returnsInt(): int{}
 
  public function doFoo() {
    $method = rand(0,1) ? 'returnsVoid' : 'returnsInt';
    $something = $this->$method();
  } 
}
```

VincentLanglet:
PR is rebased and ready for a new review @staabm @ondrejmirtes

ondrejmirtes (approved):

ondrejmirtes:
Perfect now 😊 

After merging this please send a PR that supersedes https://github.com/phpstan/phpstan-src/pull/4344.

ondrejmirtes:
Thank you.

VincentLanglet:
> Thank you.

And now I need to document this I assume ? ^^

ondrejmirtes:
Can be mentioned on PHPDoc Basics page alongside pure and impure.

thg2k:
> Can be mentioned on PHPDoc Basics page alongside pure and impure.

Would be awesome to have an index page where all "@" notations are listed together, just the directive name in alphabetical order, maybe each one linked to its own documentation page. Sometimes I know what I need but i don't remember the exact name..

ondrejmirtes:
That's basically PHPDocs basics, you can Ctrl-F there. The algolia search widget on the website works well too.

PR #4423, opened 2025-10-08
Fix bitwise on mixed

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8094


tests/PHPStan/Analyser/nsrt/bitwise.php
```diff
@@ -0,0 +1,45 @@
...
+	assertType('int', $int & $int);
+	assertType('*ERROR*', $int & $string);
+	assertType('*ERROR*', $int & $stringOrInt);
+	assertType('int', $int & $mixed);
+	assertType('string', $string & $string);
+	assertType('*ERROR*', $string & $stringOrInt);
+	assertType('*ERROR*', $string & $mixed);
+	assertType('*ERROR*', $stringOrInt & $stringOrInt);
+	assertType('*ERROR*', $stringOrInt & $mixed);
+	assertType('*ERROR*', $mixed & $mixed);
```
VincentLanglet: Current behavior https://phpstan.org/r/4e542067-2b33-4f87-8230-ca7b49937b10

tests/PHPStan/Rules/Functions/data/bug-8094.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Bug8094;
+
+function mask_encode($data, $mask)
+{
+	$mask = substr($mask, 0, strlen($data));
+	$data ^= $mask;
+	return(base64_encode($data));
```
VincentLanglet, line 9: Was reporting an error on level 5 on php 7.4 https://phpstan.org/r/12e63cdd-d516-4db2-a314-51da4f5c3101

tests/PHPStan/Analyser/nsrt/bitwise.php
staabm: Mixed & mixed - should not be a error, right?

VincentLanglet: I proposed a solution to get some feedback.

Things like
String ^ int|string 
Int ^ int|string

Are already considered as an error because int ^ string is an error.

We have
Int ^ mixed considered to be int (even if it could ends up with int ^ string ?)
String ^ mixed considered to be an error (because it could ends up with string ^ int ?)

So I dunno what we want for mixed ^ mixed but it should not be considered as int. I see three solution:
- error
- int|string
- benevolentUnion int|string

I also dunno if we should change string ^ mixed.

I think we will need a decision from @ondrejmirtes unless you have a good suggestion @staabm ? ^^'

staabm: IMO we should not treat something as an error, if it might be valid. `mixed` could be a `int` but just not typed.

therefore I think mixed ^ mixed should be a union of all possible results `^` can have.

but lets wait for ondrej

VincentLanglet: > IMO we should not treat something as an error, if it might be valid. `mixed` could be a `int` but just not typed.
> 
> therefore I think mixed ^ mixed should be a union of all possible results `^` can have.

Currently the InvalidBinaryOperationRule is using the strategy of checking if the result is an error in order to report it. So if you change `int ^ int|string` from Error to `int` this won't be reported anymore by the rule.

As an example, this is the current behavior with `int + int|array` (an error and not an int)
https://phpstan.org/r/7f8dda05-7a6e-49d0-959e-e4bf4f038586

Following the existing strategy, it make sens to have `int ^ int|string` and `string ^ int|string` an error.
I just dunno what is expected for mixed...
Currently the fact `mixed ^ mixed` still allow it to be reported by the rule because on higher level, mixed is transform into strictMixed and `strictMixed ^ strictMixed` is an Error.

So the question might be 
```
int ^ mixed = ? // currently int ; should be Error or string
string ^ mixed = ? // currently Error ; should be Error or string
mixed ^ mixed = ? // currently int ; should be Error or int|string
```

tests/PHPStan/Analyser/Generator/data/gnsr.php
```diff
@@ -110,7 +110,7 @@ public function doBitwiseNot($a, int $b): void
 	public function doBitwiseAnd($a, $b, int $c, int $d): void
 	{
 		assertType('int', $a & $b);
-		assertNativeType('int', $a & $b);
+		assertNativeType('*ERROR*', $a & $b);
```
VincentLanglet: This is equivalent to `mixed & mixed` which is now reported as an error. (we can't be sure it's an int, it could be string & string for instance or worst, int & string)

staabm: if we can't know what it is, why is it an ERROR then? shouldn't it be mixed instead?

VincentLanglet: Current mixed behavior is weird. A good example of the issue is
```
int ^ mixed = ? // currently int ; so no error is reported InvalidBinaryOperationRule
string ^ mixed = ? // currently Error ; an error is reported by InvalidBinaryOperationRule
mixed ^ mixed = ? // currently int ; so no error is reported by InvalidBinaryOperationRule
```

I'm not sure what should be done. Either error for all or something like 
```
int ^ mixed = int
string ^ mixed = string
mixed ^ mixed = (int|string)
```

VincentLanglet: I tried another solution, tell me your opinion about it @staabm :)

ondrejmirtes:
Thank you!

PR #4425, opened 2025-10-10
Infer mixed~array<mixed> | array<string>

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13628
Closes https://github.com/phpstan/phpstan/issues/12930

Original bug was 
```
array<mixed~(0|0.0|''|'0'|array{}|false|null)>|mixed~array<mixed, mixed>
```
resolved into `mixed~non-empty-array`.

Same exists with something simpler like
```
array<string>|mixed~array<mixed, mixed>
```

Because it was converted to `mixed ~ (remove(array<mixed, mixed>, array<string>)`

While it's true that `remove(array<mixed, mixed>, array<string>` is a non-empty-array,
it's untrue that `array<string>|mixed~array<mixed, mixed>` is mixed~non-empty-array.

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -580,7 +580,7 @@ final class TypeCombinator
 		Type $b,
 	): Type
 	{
-		if ($a->getSubtractedType() === null) {
+		if ($a->getSubtractedType() === null || $b instanceof NeverType) {
 			return $a;
 		}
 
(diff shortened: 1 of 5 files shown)


src/Type/TypeCombinator.php
```diff
@@ -621,7 +621,10 @@ private static function intersectWithSubtractedType(
 			} elseif ($isBAlreadySubtracted->yes()) {
 				$subtractedType = self::remove($a->getSubtractedType(), $b);
 
-				if ($subtractedType instanceof NeverType) {
+				if (
+					$subtractedType instanceof NeverType
+					|| !$subtractedType->isSuperTypeOf($b)->no()
```
VincentLanglet, line 626: This case checks if we "didn't remove" enough

ondrejmirtes:
Looks like it fixes https://github.com/phpstan/phpstan/issues/12930 too

ondrejmirtes:
Thank you!

staabm:
nice!

PR #4426, opened 2025-10-10
Do not report dead catch with `treatPhpDocTypesAsCertain: false`

VincentLanglet:
I'd like to move https://github.com/phpstan/phpstan-src/pull/4156 forward, and kinda disagree with your request change from calling `getType` to `getNativeType` inside `BackedEnumFromDynamicStaticMethodThrowTypeExtension`.

Because when using the exception analysis, PHPStan will still ask for a `@throws` tag while I know there is not because I trust my phpdoc (value-of<...>).

Also, when looking at all the existing `DynamicMethodThrowTypeExtension` they are using `getType`.

So I tried to find a solution you will agree with.
- When the throw type extension return NULL
-> If I trust the phpdoc, I consider it doesn't throw an exception.
-> If I don't, I check the throw type extension with the native type
--> If I don't have an exception with the native type I can still return NULL
--> If I have one, I had an implicit ThrowPoint.

The benefit of the implicit throwpoint is that:
- It won't report the catch as a dead catch
- It won't force people to add `@throws` tag on the method

which is a similar behavior of `treatPhpDocTypesAsCertain: false` which allow extra safety check but doesn't enforce them.

I didn't know how to correctly write test about this but the current one I wrote/updated should show the new behavior.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4305,7 +4305,16 @@ final class NodeScopeResolver
 
 				$throwType = $extension->getThrowTypeFromFunctionCall($functionReflection, $normalizedFuncCall, $scope);
 				if ($throwType === null) {
-					return null;
+					if ($this->treatPhpDocTypesAsCertain) {
+						return null;
+					}
+
+					$nativeThrowType = $extension->getThrowTypeFromFunctionCall($functionReflection, $normalizedFuncCall, $scope->doNotTreatPhpDocTypesAsCertain());
+					if ($nativeThrowType === null) {
+						return null;
+					}
+
+					return ThrowPoint::createImplicit($scope, $funcCall);
 				}
 
 				return ThrowPoint::createExplicit($scope, $throwType, $funcCall, false);
@@ -4363,7 +4372,16 @@ final class NodeScopeResolver
 
 				$throwType = $extension->getThrowTypeFromMethodCall($methodReflection, $normalizedMethodCall, $scope);
 				if ($throwType === null) {
-					return null;
+					if ($this->treatPhpDocTypesAsCertain) {
+						return null;
+					}
+
+					$nativeThrowType = $extension->getThrowTypeFromMethodCall($methodReflection, $normalizedMethodCall, $scope->doNotTreatPhpDocTypesAsCertain());
+					if ($nativeThrowType === null) {
+						return null;
+					}
+
+					return ThrowPoint::createImplicit($scope, $methodCall);
 				}
 
 				return ThrowPoint::createExplicit($scope, $throwType, $methodCall, false);
@@ -4407,7 +4425,16 @@ final class NodeScopeResolver
 
 				$throwType = $extension->getThrowTypeFromStaticMethodCall($constructorReflection, $normalizedMethodCall, $scope);
 				if ($throwType === null) {
-					return null;
+					if ($this->treatPhpDocTypesAsCertain) {
+						return null;
+					}
+
+					$nativeThrowType = $extension->getThrowTypeFromStaticMethodCall($constructorReflection, $normalizedMethodCall, $scope->doNotTreatPhpDocTypesAsCertain());
+					if ($nativeThrowType === null) {
+						return null;
+					}
+
+					return ThrowPoint::createImplicit($scope, $new);
 				}
 
 				return ThrowPoint::createExplicit($scope, $throwType, $new, false);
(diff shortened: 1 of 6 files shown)


VincentLanglet:
Damned, seems like an implicit throwPoint has impact on variable certainty...

But maybe that's an ok change when looking at the existing behavior 
https://phpstan.org/r/14eb124e-de59-407e-b803-04468b99e564

The variable certainly will be Maybe only in the try part and only with treatPhpDocTypesAsCertain: false

I'll need your thought on this @ondrejmirtes

VincentLanglet:
> Damned, seems like an implicit throwPoint has impact on variable certainty...
> 
> But maybe that's an ok change when looking at the existing behavior [phpstan.org/r/14eb124e-de59-407e-b803-04468b99e564](https://phpstan.org/r/14eb124e-de59-407e-b803-04468b99e564)
> 
> The variable certainly will be Maybe only in the try part and only with treatPhpDocTypesAsCertain: false

Also, this would be the behavior you're asking for BackedEnumFromDynamicStaticMethodThrowTypeExtension
Since the dead catch is not reported, the certainty of the variable in finally will be maybe.

ondrejmirtes:
The way it'd make sense for this to be implemented is:

* With `treatPhpDocTypesAsCertain: true`, there's a dead catch being reported thanks to calling something that has `@throws void`, or `@throws FooException` (while there only being catch for BarException) or a throw type coming from a dynamic extension.
* By setting to `treatPhpDocTypesAsCertain: false`, these dead catches would no longer be reported.
* With `treatPhpDocTypesAsCertain: false`, some dead catches would still be reported - for example if there's an empty `try {}` block, or if there's a certain `throw new FooException` which doesn't depend on any PHPDoc being interpreted.
* With `treatPhpDocTypesAsCertain: true` and `catch.neverThrown` being reported for situations that would make the error disappear with `treatPhpDocTypesAsCertain: false`, there has to be a tip attached to the error informing about this (like all current errors dependent on this do).

I think all this means that explicit throw points (objects of ThrowPoint class) need to carry whether they're PHPDoc-based or not.

ondrejmirtes:
With `treatPhpDocTypesAsCertain: false`, calling a `@throws void` function should probably create an implicit ThrowPoint.

VincentLanglet:
> The way it'd make sense for this to be implemented is:
> 
> * With `treatPhpDocTypesAsCertain: true`, there's a dead catch being reported thanks to calling something that has `@throws void`, or `@throws FooException` (while there only being catch for BarException) or a throw type coming from a dynamic extension.
> * By setting to `treatPhpDocTypesAsCertain: false`, these dead catches would no longer be reported.
> * With `treatPhpDocTypesAsCertain: false`, some dead catches would still be reported - for example if there's an empty `try {}` block, or if there's a certain `throw new FooException` which doesn't depend on any PHPDoc being interpreted.
> * With `treatPhpDocTypesAsCertain: true` and `catch.neverThrown` being reported for situations that would make the error disappear with `treatPhpDocTypesAsCertain: false`, there has to be a tip attached to the error informing about this (like all current errors dependent on this do).
> 
> I think all this means that explicit throw points (objects of ThrowPoint class) need to carry whether they're PHPDoc-based or not.

Isn't this a separate (big) next step after this draft ?

We could
- In a first time having the same behavior between Dynamic extension, currently we're having
https://phpstan.org/r/254167dd-11ae-4a7d-946a-1379e91a2d67
https://phpstan.org/r/cbc3af21-48a1-4731-b915-40209936ff31

- In a second time handle `@throws` tag correctly with `treatPhpDocTypesAsCertain: false` ?

Looking at https://github.com/phpstan/phpstan-src/blob/844f5693f8caebb1e2c1b57ebe5e288ac94e16f3/src/Analyser/MutatingScope.php#L6246

I feel like to have the exact same behavior, I should just change to
```
$throwType = $extension->getThrowTypeFromFunctionCall($functionReflection, $normalizedFuncCall, $this->treatPhpDocTypesAsCertain ? $scope : $scope->doNotTreatPhpDocTypesAsCertain());
```

ondrejmirtes:
My proposition is in line of what `treatPhpDocTypesAsCertain: false` does in the rest of the codebase.

VincentLanglet:
> My proposition is in line of what `treatPhpDocTypesAsCertain: false` does in the rest of the codebase.

Maybe it's because I don't know enough this part of the codebase but this seems to me to be a too big work just to move forward the PR https://github.com/phpstan/phpstan-src/pull/4156 which seems to get unfortunately randomly blocked by a request change that every other DynamicThrowTypeExtension didn't got 😢 

So I feel like I just have to add the DynamicThrowTypeExtension on my side rather than on PHPStan one

ondrejmirtes:
I don't think it's too much complicated work. You need to look at every `ThrowPoint::createExplicit()` call and decide if they come from a PHPDoc or not. And slightly modify CatchWithUnthrownExceptionNode and CatchWithUnthrownExceptionRule.

Alright, changing the handling of TryCatch in NodeScopeResolver might be necessary and that's 200 lines of code that you'd need to get into but still it's not rocket (or typesystem) science: https://github.com/phpstan/phpstan-src/blob/844f5693f8caebb1e2c1b57ebe5e288ac94e16f3/src/Analyser/NodeScopeResolver.php#L1678-L1895

PR #4427, opened 2025-10-10
Try reportNestedTooWideType again

ondrejmirtes:


staabm:
@ondrejmirtes do you have a reference in which PR we talked about the mentioned problems. I can't remember/find it in which context I tested this before.

ondrejmirtes:
@staabm I don't think so, just this one https://github.com/phpstan/phpstan/issues/13482 but it's not about a nested type.

staabm:
checked 5 different projects and could not find a false positive

ondrejmirtes:
Alright, let's try this again. I'm a little bit worried about this being too annoying still, even if there aren't false positives.

staabm:
just found a false positive in https://github.com/phpstan/phpstan/issues/13676

PR #4429, opened 2025-10-12
NodeCallbackInvoker

ondrejmirtes:


src/Analyser/DirectInternalScopeFactoryFactory.php
```diff
@@ -0,0 +1,65 @@
...
+use PHPStan\DependencyInjection\Type\ExpressionTypeResolverExtensionRegistryProvider;
+use PHPStan\Node\Printer\ExprPrinter;
+use PHPStan\Parser\Parser;
+use PHPStan\Php\PhpVersion;
+use PHPStan\Reflection\AttributeReflectionFactory;
+use PHPStan\Reflection\InitializerExprTypeResolver;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Properties\PropertyReflectionFinder;
+
+final class DirectInternalScopeFactoryFactory implements InternalScopeFactoryFactory
```
staabm, line 16: I am getting java vibes.. where is the FactoryFactoryFactory :-)

src/Rules/Playground/PromoteParameterRule.php
```diff
@@ -87,7 +88,7 @@ private function getOriginalRule(): ?Rule
 		return $this->originalRule = $originalRule;
 	}
 
-	public function processNode(Node $node, Scope $scope): array
+	public function processNode(Node $node, Scope&NodeCallbackInvoker $scope): array
```
staabm, line 91: is this a testing-only change which we need to be cleaned up?

tests/PHPStan/Rules/NodeCallbackInvokerRule.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+class NodeCallbackInvokerRule implements Rule
+{
+
+	public function getNodeType(): string
+	{
+		return Node\Stmt\Echo_::class;
+	}
+
+	public function processNode(Node $node, NodeCallbackInvoker&Scope $scope): array
```
staabm, line 20:
```suggestion
	public function processNode(Node $node, Scope&NodeCallbackInvoker $scope): array
```

src/Rules/Playground/PromoteParameterRule.php
ondrejmirtes: No, we're calling another rule, which might need it. If you revert it, you get PHPStan error.

tests/PHPStan/Rules/NodeCallbackInvokerRule.php
staabm: fixed the downgrader to also support the intersection type when notated in different order:
https://github.com/ondrejmirtes/simple-downgrader/pull/17

src/Analyser/NodeCallbackInvoker.php
```diff
@@ -0,0 +1,37 @@
...
+ * public static create(string $class, mixed ...$args)
+ * ```
+ *
+ * And you'd like to check `Factory::create(Foo::class, 1, 2, 3)` as if it were
+ * `new Foo(1, 2, 3)`.
+ *
+ * You can call `$scope->invokeNodeCallback(new New_(new Name($className), $args))`
+ *
+ * And PHPStan will call all the registered rules for New_, checking as if the instantiation
+ * is actually in the code.
```
canvural, line 28: oh wow, this is awesome! 🚀

ondrejmirtes: Yeah, I vaguely remember you're using FunctionCallParametersCheck somewhere, which is not part of BC promise. This would allow you to write simple and clean code instead 😊

ondrejmirtes: Please note that this PHPDoc is lying a little bit. You don't need to change your processNode signature to take advantage of this. phpstan-src uses native intersection type which is downgraded to PHPDoc because of PHP 7.4+ support. Which means that unless you have your own PHPDoc in your custom rule, PHPStan will implicitly inherit the Rule interface PHPDoc and you can call `$scope->invokeNodeCallback` without changing the signature :)

PR #4434, opened 2025-10-13
Invalidate static expressions when a non-static expression is called

takaram:
Closes phpstan/phpstan#13416

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3656,6 +3656,67 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 		);
 	}
 
+	public function invalidateStaticMembers(string $className): self
+	{
+		if (!$this->reflectionProvider->hasClass($className)) {
+			return $this;
+		}
+
+		$classReflection = $this->reflectionProvider->getClass($className);
+		$classNamesToInvalidate = [strtolower($className)];
+		foreach ($classReflection->getParents() as $parentClass) {
+			$classNamesToInvalidate[] = strtolower($parentClass->getName());
+		}
+
+		$expressionTypes = $this->expressionTypes;
+		$nativeExpressionTypes = $this->nativeExpressionTypes;
+		$invalidated = false;
+		$nodeFinder = new NodeFinder();
+		foreach ($expressionTypes as $exprString => $exprTypeHolder) {
+			$expr = $exprTypeHolder->getExpr();
+			$found = $nodeFinder->findFirst([$expr], static function (Node $node) use ($classNamesToInvalidate): bool {
+				if (!$node instanceof Expr\StaticCall && !$node instanceof Expr\StaticPropertyFetch) {
+					return false;
+				}
+				if (!$node->class instanceof Name || !$node->class->isFullyQualified()) {
+					return false;
+				}
+
+				return in_array($node->class->toLowerString(), $classNamesToInvalidate, true);
+			});
+			if ($found === null) {
+				continue;
+			}
+
+			unset($expressionTypes[$exprString]);
+			unset($nativeExpressionTypes[$exprString]);
+			$invalidated = true;
+		}
+
+		if (!$invalidated) {
+			return $this;
+		}
+
+		return $this->scopeFactory->create(
+			$this->context,
+			$this->isDeclareStrictTypes(),
+			$this->getFunction(),
+			$this->getNamespace(),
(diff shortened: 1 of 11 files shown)


ondrejmirtes:
Please rebase this and fix conflicts, thank yuo.

takaram:
@ondrejmirtes 
I rebased the branch.

src/Analyser/NodeScopeResolver.php
```diff
@@ -3173,9 +3173,16 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context, $sto
 			$scope = $result->getScope();
 
 			if ($methodReflection !== null) {
-				if ($methodReflection->getName() === '__construct' || $methodReflection->hasSideEffects()->yes()) {
+				$hasSideEffects = $methodReflection->hasSideEffects()->yes();
+				if ($hasSideEffects || $methodReflection->getName() === '__construct') {
 					$this->callNodeCallback($nodeCallback, new InvalidateExprNode($normalizedExpr->var), $scope, $storage);
 					$scope = $scope->invalidateExpression($normalizedExpr->var, true);
+					if ($hasSideEffects) {
+						$classNames = $scope->getType($normalizedExpr->var)->getObjectClassNames();
```
staabm: It might make sense to use getObjectClassReflections() here, so you don't need ReflectionProvider handling in the new invalidate-method.

Might also make sense to move the loop into the invalidate method, so you don't need to re-create several scope objects

src/Analyser/MutatingScope.php
```diff
@@ -3608,6 +3608,67 @@ private function invalidateMethodsOnExpression(Expr $expressionToInvalidate): se
...
+		$classReflection = $this->reflectionProvider->getClass($className);
+		$classNamesToInvalidate = [strtolower($className)];
+		foreach ($classReflection->getParents() as $parentClass) {
+			$classNamesToInvalidate[] = strtolower($parentClass->getName());
+		}
+
+		$expressionTypes = $this->expressionTypes;
+		$nativeExpressionTypes = $this->nativeExpressionTypes;
+		$invalidated = false;
+		$nodeFinder = new NodeFinder();
```
staabm, line 3673: Another thing which came to mind is, that we add another ast travsersal here. 

Can this be combined with the already happening ast traversal ?

(Just thinking loud - might not be useful)

src/Analyser/NodeScopeResolver.php
takaram: Fixed

src/Analyser/MutatingScope.php
takaram: Hmm I don't think it is easy to remove NodeFinder here...

staabm (commented):
mutating testing suggests, we are missing a unit test for a call which maybe has side effects.

take inspiration from https://github.com/phpstan/phpstan-src/pull/4422#issuecomment-3616601264 if you need a example on how this can be achieved.

src/Analyser/MutatingScope.php
```diff
@@ -3656,6 +3656,66 @@ private function invalidateMethodsOnExpression(Expr $expressionToInvalidate): se
...
+		$classReflections = $this->getType($var)->getObjectClassReflections();
+		$classNamesToInvalidate = [];
+		foreach ($classReflections as $classReflection) {
+			$classNamesToInvalidate[] = strtolower($classReflection->getName());
+			foreach ($classReflection->getParents() as $parentClass) {
+				$classNamesToInvalidate[] = strtolower($parentClass->getName());
+			}
+		}
+
+		$expressionTypes = $this->expressionTypes;
```
staabm, line 3670: I think we could return early with `return $this;` when `$classNamesToInvalidate` is empty

src/Analyser/NodeScopeResolver.php
```diff
@@ -3215,9 +3217,15 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context, $sto
 			$scope = $result->getScope();
 
 			if ($methodReflection !== null) {
-				if ($methodReflection->getName() === '__construct' || $methodReflection->hasSideEffects()->yes()) {
+				$shouldInvalidateExpr = $this->rememberPossiblyImpureFunctionValues
```
staabm, line 3220: maybe I misread the PR, but it seems the path where `$this->rememberPossiblyImpureFunctionValues` is `false` is not tested ?

PR #4435, opened 2025-10-13
Support int range in bitwise operations

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/7912
Closes https://github.com/phpstan/phpstan/issues/9384

Closes https://github.com/phpstan/phpstan-src/pull/2441

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -976,8 +976,17 @@ final class InitializerExprTypeResolver
 			return $this->getNeverType($leftType, $rightType);
 		}
 
-		$leftTypes = $leftType->getConstantScalarTypes();
-		$rightTypes = $rightType->getConstantScalarTypes();
+		if ($leftType instanceof IntegerRangeType) {
+			$leftTypes = $leftType->getFiniteTypes();
+		} else {
+			$leftTypes = $leftType->getConstantScalarTypes();
+		}
+		if ($rightType instanceof IntegerRangeType) {
+			$rightTypes = $rightType->getFiniteTypes();
+		} else {
+			$rightTypes = $rightType->getConstantScalarTypes();
+		}
+
 		$leftTypesCount = count($leftTypes);
 		$rightTypesCount = count($rightTypes);
 		if ($leftTypesCount > 0 && $rightTypesCount > 0) {
@@ -1062,8 +1071,17 @@ final class InitializerExprTypeResolver
 			return $this->getNeverType($leftType, $rightType);
 		}
 
-		$leftTypes = $leftType->getConstantScalarTypes();
-		$rightTypes = $rightType->getConstantScalarTypes();
+		if ($leftType instanceof IntegerRangeType) {
+			$leftTypes = $leftType->getFiniteTypes();
+		} else {
+			$leftTypes = $leftType->getConstantScalarTypes();
+		}
+		if ($rightType instanceof IntegerRangeType) {
+			$rightTypes = $rightType->getFiniteTypes();
+		} else {
+			$rightTypes = $rightType->getConstantScalarTypes();
+		}
+
 		$leftTypesCount = count($leftTypes);
 		$rightTypesCount = count($rightTypes);
 		if ($leftTypesCount > 0 && $rightTypesCount > 0) {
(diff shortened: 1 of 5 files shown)


VincentLanglet:
Seems like it solves https://github.com/phpstan/phpstan/issues/9384 too, I need to add regression tests

VincentLanglet:
https://github.com/phpstan/phpstan-src/pull/4437 first would be great

ondrejmirtes:
https://github.com/phpstan/phpstan-src/pull/4437 now merged.

tests/PHPStan/Analyser/nsrt/bug-9384.php
```diff
@@ -0,0 +1,22 @@
...
+
+use function PHPStan\Testing\assertType;
+
+class BinaryOr
+{
+	private const TYPE_NONE               = 0;
+	private const TYPE_TRACK_DEPRECATIONS = 1;
+	private const TYPE_TRIGGER_ERROR      = 2;
+
+	/** @var int-mask-of<self::TYPE_*>|null */
```
VincentLanglet: It would work fine with `/** @var int-mask-of<self::TYPE_*>|null */` the issue here is the fact we have an union `int-mask-of<self::TYPE_*>|null` which does not pass the condition `$leftType instanceof IntegerRangeType`.

Any suggestion @staabm ?

Is it ok to call `type->toArrayKey` in order to cast the value to string|int only (?)
The semantic feel weird

staabm: I think we have 2 options.

either add a new method on `Type`, or do something like:

```php
$containsNull = TypeCombinator::containsNull($type);
if ($containsNull) {
  $type = $type=TypeCombinator::removeNull($type);
}

// ... process

if ($containsNull) {
  $type = TypeCombinator::addNull($type);
}
```

VincentLanglet: I'd like to avoid the dedicated behavior with Null, since the same exists with `false` (and other values)
so the same issue exists for `int-mask-of<self::TYPE_*>|false`

I'll try to find another way

staabm: If there is type dependent behavior than a new Type method is the easy way.

VincentLanglet: > If there is type dependent behavior than a new Type method is the easy way.

I feel like lot of IntegerRangeType issues are starting with the fact that `IntegerRangeType::getConstantScalarTypes` returns `[]` so we have to add a `getFiniteTypes` every time with a `instanceof IntegerRangeType` check.

So in reality there is no real need to add a dedicated Type method, but IntegerRangeType::getConstantScalarTypes could be updated. When doing this change and running the tests, every results are
- either more precise
- or similar like `0|1|2` instead of `int<0, 2>`

I don't think the discussion, but I think this change is not wanted. So a dedicated type method to basically do the same feels weird.

staabm: The 2 type methods only work the same for small integer-ranges (for efficiency reasons )

VincentLanglet: I'm not sure to understand, we could still have the same implementation than https://github.com/phpstan/phpstan-src/blob/a4926015eded21a108bb679226e1bc582be5a3a5/src/Type/IntegerRangeType.php#L699-L716 for the getConstantScalarTypes.

Meaning small integerRange will returns a non empty array while big integerRange will return an empty array.

VincentLanglet: Anyway, I think I won't spend more time to support null in this PR.
The issues exists every where we're doing a `instanceof IntegerRangeType` and should be treated in another PR ihmo.

Especially when the value should/could be initialized with 0 rather than Null when doing operation with bitwise operators.

staabm: > Anyway, I think I won't spend more time to support null in this PR.
> The issues exists every where we're doing a `instanceof IntegerRangeType` and should be treated in another PR ihmo.

agree

staabm (approved):
in the future we might move the finite/constant-types logic into a new `PHPStan\Type\TypeUtils` method

```
		if ($leftType instanceof IntegerRangeType) {
			$leftTypes = $leftType->getFiniteTypes();
		} else {
			$leftTypes = $leftType->getConstantScalarTypes();
		}
```

but this is an idea for a followup PR

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1138,8 +1156,17 @@ public function getBitwiseXorTypeFromTypes(Type $leftType, Type $rightType): Typ
...
+		if ($leftType instanceof IntegerRangeType) {
+			$leftTypes = $leftType->getFiniteTypes();
+		} else {
+			$leftTypes = $leftType->getConstantScalarTypes();
+		}
+		if ($rightType instanceof IntegerRangeType) {
+			$rightTypes = $rightType->getFiniteTypes();
+		} else {
+			$rightTypes = $rightType->getConstantScalarTypes();
+		}
```
staabm, line 1168: is this change covered by a test?

staabm:
in a followup we could add similar logic to other operator methods `InitializerExprTypeResolver` which atm only call `getConstantScalarTypes` but not `getFiniteTypes`.

VincentLanglet:
> in a followup we could add similar logic to other operator methods `InitializerExprTypeResolver` which atm only call `getConstantScalarTypes` but not `getFiniteTypes`.

Which operator you have in mind ? For common maths, they end up in `integerRangeMath`

staabm:
> Which operator you have in mind ?

spaceship, mod

staabm (approved):
awesome

PR #4436, opened 2025-10-13
Remove PHP version constraint from the `#[\Override]` attribute rules

johnbillion:
There are two PHPStan rules relating to use of the `#[\Override]` attribute on methods:

1. `method.missingOverride` which detects when a method overrides another but is missing the `#[\Override]` attribute. This rule is operational only when the `checkMissingOverrideMethodAttribute` config option is set to true.
2. `method.override` which detects when `#[\Override]` is used on a method which does not override a method.

Currently these rules are triggered when PHP 8.3 or higher is in use, because that's the version that introduced support for this attribute.

However both of these rules are opt-in and therefore don't need to be constrained by PHP version. The former is explicitly opt-in via the `checkMissingOverrideMethodAttribute` config option and the latter is implicitly opt-in via the existence of an `#[\Override]` attribute on a method.

This PR proposes removing the PHP version constraint from these rules so they can be applied to a codebase which supports versions of PHP older than 8.3 and still benefit from static analysis by PHPStan.

## What's the use case?

I work on codebases which support PHP 7.4 to 8.5. I want to make use of the `#[\Override]` attribute on methods and have PHPStan confirm that their usage is correct. The fact that the codebase is compatible with PHP < 8.3 should be of no concern to these rules.

Currently if the following PHP version range config is in `phpstan.neon` then the `#[\Override]` attribute rules don't apply and PHPStan won't check their correctness.

```neon
phpVersion:
	min: 70400
	max: 80500
```

## Is this a breaking change?

I don't believe this should be considered a breaking change, however there may be codebases that are using `#[\Override]` attributes and a supported version of PHP lower than 8.3 that don't realise these rules are not currently being applied. These rule changes may cause previously passing incorrect use of `#[\Override]` to now fail, which arguably is a bugfix.

### DIFF
--- a/src/Rules/Methods/OverridingMethodRule.php
+++ b/src/Rules/Methods/OverridingMethodRule.php
@@ -87,7 +87,7 @@ final class OverridingMethodRule implements Rule
 				}
 			}
 
-			if ($this->phpVersion->supportsOverrideAttribute() && $this->hasOverrideAttribute($node->getOriginalNode())) {
+			if ($this->hasOverrideAttribute($node->getOriginalNode())) {
 				return [
 					RuleErrorBuilder::message(sprintf(
 						'Method %s::%s() has #[\Override] attribute but does not override any method.',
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Makes total sense! I also did the same thing for properties. Sorry for taking so long.

Thank you!

PR #4437, opened 2025-10-13
Resolve bitwise not on constant integer

VincentLanglet:
While working on https://github.com/phpstan/phpstan-src/pull/4435 I discovered the issue https://github.com/phpstan/phpstan/issues/9384 was almost solve except for the operation with the bitwise operator not `~`.

This one generalize constant integer into integer while we could compute the value.

A comment is saying it depends on PHP_INT_SIZE but we could consider that 
- 99% operation `~` are on small number
- Relying on the PHP_INT_SIZE of the user might be good enough
- It will be consistent with other bitwise operator like `<<` https://phpstan.org/r/58bdfbb7-a047-44ff-8af3-995d976d7bb5

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -2634,8 +2634,11 @@ final class InitializerExprTypeResolver
 
 				return new IntersectionType([new StringType(), ...$accessories]);
 			}
+			if ($type instanceof ConstantIntegerType || $type instanceof ConstantFloatType) {
+				return new ConstantIntegerType(~$type->getValue());
+			}
 			if ($type->isInteger()->yes() || $type->isFloat()->yes()) {
-				return new IntegerType(); //no const types here, result depends on PHP_INT_SIZE
+				return new IntegerType();
 			}
 			return new ErrorType();
 		});
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -2635,7 +2635,7 @@ final class InitializerExprTypeResolver
 				return new IntersectionType([new StringType(), ...$accessories]);
 			}
 			if ($type instanceof ConstantIntegerType || $type instanceof ConstantFloatType) {
-				return new ConstantIntegerType(~$type->getValue());
+				return new ConstantIntegerType(~ (int) $type->getValue());
 			}
 			if ($type->isInteger()->yes() || $type->isFloat()->yes()) {
 				return new IntegerType();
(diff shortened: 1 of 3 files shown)


VincentLanglet:
WDYT of this small fix @staabm @ondrejmirtes ?
I'll need it to fix https://github.com/phpstan/phpstan-src/pull/4435 after

staabm:
> * 99% operation `~` are on small number

I am not sure this argument makes sense, since as far as I understand it `~1` already returns a different result on 64 bit vs. 32 bit PHP.

---

I think we need to either judge whether...
- 32 bit php is still worth supporting (maybe worth dropping with PHPStan 3.x)
- whether this more precise types should be bleeding edge only (or otherwise feature flagged)

VincentLanglet:
> > * 99% operation `~` are on small number
> 
> I am not sure this argument makes sense, since as far as I understand it `~1` already returns a different result on 64 bit vs. 32 bit PHP.

When I tried I got `~n = -(n+1)` and it does not depends on the PHP bit size (for small numbers).
You can try, `~1` returns `-2` in both 32 and 64 bit PHP.

ondrejmirtes:
Thank you.

ondrejmirtes:
Either this or #4435 is a reason for a new error:

```
				$previousErrorHandler = set_error_handler(
					function (int $severity, string $message, string $file, int $line) use ($item, $columnIndex, $rowIndex, $startRowIndex, &$previousErrorHandler): mixed {
						if (($severity & E_WARNING | E_NOTICE) > 0) {
							$this->throwInvalidValueException($item, $this->columns[$columnIndex], $rowIndex - $startRowIndex, $columnIndex);
						} elseif ($previousErrorHandler !== null) {
							return $previousErrorHandler($severity, $message, $file, $line);
						}

						throw new ShouldNotHappenException(sprintf('Unexpected error %s: %s', $severity, $message));
					},
				);
```

```

  117    Comparison operation ">" between 8|9|10 and 0 is always true.  
         🪪  greater.alwaysTrue                                         
  123    Unreachable statement - code above always terminates.          
         🪪  deadCode.unreachable                                       

```

Do you think it's valid or not? I suspect the code should be parenthesized like this `$severity & (E_WARNING | E_NOTICE)` but not sure.

VincentLanglet:
Were does come from this error ? I don't see such code in phpstan codebase.

> Do you think it's valid or not? I suspect the code should be parenthesized like this `$severity & (E_WARNING | E_NOTICE)` but not sure.

This seems valid to me, I think it should be `$severity & (E_WARNING | E_NOTICE)`

ondrejmirtes:
Each commit to PHPStan is analysed on four private projects only I have access to.

staabm:
> I think it should be `$severity & (E_WARNING | E_NOTICE)`

agree

PR #4438, opened 2025-10-14
Implement CompositeRule

staabm:
Allows testing of rules which delegate work to [`NodeCallbackInvoker`](https://github.com/phpstan/phpstan-src/pull/4429)

extracted from https://github.com/phpstan/phpstan-phpunit/pull/238

### DIFF
--- a/src/Testing/CompositeRule.php
+++ b/src/Testing/CompositeRule.php
@@ -21,9 +21,14 @@ final class CompositeRule implements Rule
 
 	private DirectRegistry $registry;
 
-	public function __construct(DirectRegistry $ruleRegistry)
+	/**
+	 * @param array<Rule<Node>> $rules
+	 *
+	 * @api
+	 */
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
The constructor should ask for an array of rules, DirectRegistry should be created internally.

src/Testing/CompositeRule.php
```diff
@@ -0,0 +1,48 @@
...
+ * @implements Rule<Node>
+ *
+ * @api
+ */
+final class CompositeRule implements Rule
+{
+
+	private DirectRegistry $registry;
+
+	public function __construct(DirectRegistry $ruleRegistry)
```
ondrejmirtes: Constructor needs `@api` too

ondrejmirtes (changes requested):
A test please 😊

ondrejmirtes:
Thank you!

PR #4440, opened 2025-10-15
infer `non-empty-list/array` after `array_key_exists($i, $arr)`

staabm:
refs https://github.com/phpstan/phpstan/issues/13674


ondrejmirtes:
Thank you!

PR #4441, opened 2025-10-15
infer `non-empty-list/array` after `isset($arr[$i])`

staabm:
analog https://github.com/phpstan/phpstan-src/pull/4440 but for `isset($arr[$i])`

closes https://github.com/phpstan/phpstan/issues/13674
closes https://github.com/phpstan/phpstan/issues/12482
closes https://github.com/phpstan/phpstan/issues/13675


src/Analyser/TypeSpecifier.php
```diff
@@ -929,6 +929,18 @@ public function specifyTypesInCondition(
 						);
 					} else {
 						$varType = $scope->getType($var->var);
+
+						if ($varType->isArray()->yes() && $dimType->isConstantScalarValue()->no()) {
+							$types = $types->unionWith(
+								$this->create(
+									$var->var,
+									new NonEmptyArrayType(),
```
staabm: hmm interessting.. for some reason this operation makes array loose their template type.

```
3) PHPStan\Rules\Methods\ReturnTypeRuleTest::testBug7511
--- Expected
+++ Actual
@@ @@
-'
+'48: Method Bug7511\HelloWorld::computeForFrontByPosition() should return array<T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface> but returns array<Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface>.
 '

/Users/m.staab/dvl/phpstan-src/src/Testing/RuleTestCase.php:179
/Users/m.staab/dvl/phpstan-src/tests/PHPStan/Rules/Methods/ReturnTypeRuleTest.php:759
```

staabm: I think the root cause of this errors is, that `setExistingOffset` on a `array<TemplateIntersectionType>` leads to a `array<IntersectionType>` in

grafik

staabm: @VincentLanglet do you have an idea how this works?

staabm: @jiripudil may I ask you whether you can make sense of this new error, whether it is correct or how a fix can/should look like? my generics fuu is not enough to reason about the results of the above comments

src/Analyser/TypeSpecifier.php
```diff
@@ -1088,6 +1089,19 @@ public function specifyTypesInCondition(
 								)->setRootExpr($expr),
 							);
 						}
+
+						// should be getType(), but we workaround a generics bug
+						if ($scope->getNativeType($var->var)->isArray()->yes()) {
```
staabm: I have no idea how to fix it, but I think the patch as is already is very valuable.
see details in https://github.com/phpstan/phpstan-src/pull/4441#discussion_r2431684677

I would open a new issue about this open problem, merge it and call it a day.

VincentLanglet: That does not fix the issue for 13674b which should imply some non empty array/list

staabm: I just adjusted the `13674b` test

VincentLanglet: I mean, instead of hiding the bug by adding a native typehint ; it would be better to find why getType doesnt work here.

Ill take a quick look when Ill have a computer if needed( not before monday), but I really dislike this "fix" as it is actually

VincentLanglet (changes requested):
The getNative hack is definetly non acceptable since the bug will occurs if you add native typehint 'array' to the Bug7511.
```
namespace Bug7511;

interface PositionEntityInterface {
	public function getPosition(): int;
}
interface TgEntityInterface {}

abstract class HelloWorld
{
	/**
	 * @phpstan-template T of PositionEntityInterface&TgEntityInterface
	 *
	 * @param array<T> $tgs
	 *
	 * @return array<T>
	 *
	 * @throws \Exception
	 */
	public function computeForFrontByPosition(array $tgs)
	{
		/** @phpstan-var array<T> $res */
		$res = [];

		assertType('T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition(), parameter)', $res[1]);

		foreach ($tgs as $tgItem) {
			$position = $tgItem->getPosition();

			if (!isset($res[$position])) {
				assertType('T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition(), argument)', $tgItem);
				$res[$position] = $tgItem;
			} else {
				assertType('T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition(), argument)', $tgItem);
				assertType('T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition(), parameter)', $res[$position]);
				$tgItemToKeep   = $this->compare($tgItem, $res[$position]);
				assertType('T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition(), parameter)', $tgItemToKeep);
				$res[$position] = $tgItemToKeep;
			}
		}
		ksort($res);

		assertType('array<T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition(), parameter)>', $res);

		return $res;
	}

	/**
	 * @phpstan-template T of PositionEntityInterface&TgEntityInterface
	 *
	 * @param array<T> $tgs
	 *
	 * @return array<T>
	 *
	 * @throws \Exception
	 */
	public function computeForFrontByPosition2(array $tgs)
	{
		/** @phpstan-var array<T> $res */
		$res = [];

		assertType('array<T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition2(), parameter)>', $res);

		foreach ($tgs as $tgItem) {
			$position = $tgItem->getPosition();

			assertType('T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition2(), parameter)', $res[$position]);
			if (isset($res[$position])) {
				assertType('T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition2(), parameter)', $res[$position]);
			}
		}
		assertType('array<T of Bug7511\PositionEntityInterface&Bug7511\TgEntityInterface (method Bug7511\HelloWorld::computeForFrontByPosition2(), parameter)>', $res);

		return $res;
	}

	/**
	 * @phpstan-template S of TgEntityInterface
	 * @phpstan-param S $nextTg
	 * @phpstan-param S $currentTg
	 * @phpstan-return S
	 */
	abstract protected function compare(TgEntityInterface $nextTg, TgEntityInterface $currentTg): TgEntityInterface;
}
```

VincentLanglet:
@staabm with some debugging, I end up with `ArrayDimFetchHandler` calling `IntersectionType::getOffsetValueType` which does `TypeCombinator::intersect(TemplateIntersectionType, MixedType)` which gives a `IntersectionType` instead of a TemplateIntersectionType one.

The bug is created because `Array<TemplateIntersectionType>` is now specified to `Array<TemplateIntersectionType>&NonEmptyArray` (which is an IntersectionType) after your PR.

We could eventually start by fix/improve `TypeCombinator::intersect` with `TemplateIntersectionType`.

I hope it helps you

staabm:
yeah, I agree. its a similar observeration that what we found out in https://github.com/phpstan/phpstan-src/pull/4441#discussion_r2431684677

maybe we need something like https://github.com/phpstan/phpstan-src/pull/3535

staabm:
I think I was able to track it down to

https://github.com/phpstan/phpstan-src/blob/9531b297801d2100af701991a6414c7313c7d5bc/src/Type/TypeCombinator.php#L507-L513

in which a comparison of

`IntersectionType<A>` and `TemplateIntersectionType<A>` is generalized into `IntersectionType<A>` because of `isSuperTypeOf` logic

staabm:
underlying generics problem was separated into https://github.com/phpstan/phpstan/issues/14348

VincentLanglet (approved):

VincentLanglet (approved):

PR #4444, opened 2025-10-16
Fix too wide bool in trivial methods

staabm:
closes https://github.com/phpstan/phpstan/issues/13482

### DIFF
--- a/src/Rules/TooWideTypehints/TooWideTypeCheck.php
+++ b/src/Rules/TooWideTypehints/TooWideTypeCheck.php
@@ -182,21 +182,11 @@ final class TooWideTypeCheck
 		}
 
 		$returnTypes = [];
-		$returnsLiteralTrueFalse = [];
 		foreach ($returnStatements as $returnStatement) {
 			$returnNode = $returnStatement->getReturnNode();
 			if ($returnNode->expr === null) {
 				$returnTypes[] = new VoidType();
 				continue;
-			}
-
-			if (
-				$returnNode->expr instanceof ConstFetch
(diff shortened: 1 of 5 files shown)


src/Rules/TooWideTypehints/TooWideTypeCheck.php
```diff
@@ -176,6 +176,15 @@ public function checkFunctionReturnType(
 		if (count($returnStatements) === 0) {
 			return [];
 		}
+		if (
+			count($returnStatements) === 1
+			&& (
+				$nativeFunctionReturnType->isBoolean()->yes()
+				|| $phpDocFunctionReturnType->isBoolean()->yes()
```
ondrejmirtes: I thought we'd do this only for literal `return true` and `return false`, not for `return $this->foo->doFoo()`.

ondrejmirtes (changes requested):

src/Rules/TooWideTypehints/TooWideTypeCheck.php
staabm: good catch. fixed

ondrejmirtes (changes requested):
You've deleted a bunch of test assertions. Right now there are vast areas of the code that aren't covered by proper tests.

What needs to happen first is to change the current tests so they continue to be reported.

Then you can add your own tests for code that shouldn't be reported. Thanks.

staabm:
@ondrejmirtes great feedback thanks. adjusted the tests accordingly.

ondrejmirtes (changes requested):
Escaped mutants here 😊

ondrejmirtes:
Thank you!

PR #4447, opened 2025-10-16
Add more numeric-string inferences after string concat

VincentLanglet:
Improves https://github.com/phpstan/phpstan-src/pull/3117 cc @staabm

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -588,6 +588,8 @@ final class InitializerExprTypeResolver
 
 		$leftNumericStringNonEmpty = TypeCombinator::remove($leftStringType, new ConstantStringType(''));
 		if ($leftNumericStringNonEmpty->isNumericString()->yes()) {
+			$allowedPattern = $left->isInteger()->yes() ? '#^[0-9.]+$#' : '#^[0-9]+$#';
+
 			$allRightConstantsZeroOrMore = false;
 			foreach ($rightConstantStrings as $rightConstantString) {
 				if ($rightConstantString->getValue() === '') {
@@ -596,7 +598,7 @@ final class InitializerExprTypeResolver
 
 				if (
 					!is_numeric($rightConstantString->getValue())
-					|| Strings::match($rightConstantString->getValue(), '#^[0-9]+$#') === null
+					|| Strings::match($rightConstantString->getValue(), $allowedPattern) === null
 				) {
 					$allRightConstantsZeroOrMore = false;
 					break;
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -117,6 +117,7 @@ use function is_string;
 use function max;
 use function min;
 use function sprintf;
+use function str_starts_with;
 use function strtolower;
 use const INF;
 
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/nsrt/bug-11129.php
```diff
@@ -68,7 +68,7 @@ public function foo(
 		assertType('non-falsy-string&numeric-string&uppercase-string', $float.$positiveInt);
 		assertType('non-falsy-string&uppercase-string', $float.$negativeInt);
 		assertType('non-falsy-string&uppercase-string', $float.$i);
-		assertType('non-falsy-string&uppercase-string', $i.$float); // could be 'non-falsy-string&numeric-string&uppercase-string'
+		assertType('non-falsy-string&uppercase-string', $i.$float);
```
VincentLanglet, line 82: I don't think it `could be 'non-falsy-string&numeric-string&uppercase-string` since we could have a negative float.

So it will give something like `3-3.0` which is not a numeric string.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -506,6 +506,8 @@ public function resolveConcatType(Type $left, Type $right): Type
 
 		$leftNumericStringNonEmpty = TypeCombinator::remove($leftStringType, new ConstantStringType(''));
 		if ($leftNumericStringNonEmpty->isNumericString()->yes()) {
+			$allowedPattern = $left->isInteger()->yes() ? '#^[0-9.]+$#' : '#^[0-9]+$#';
```
staabm:
```suggestion
			$allowedRightPattern = $left->isInteger()->yes() ? '#^[0-9.]+$#' : '#^[0-9]+$#';
```

staabm: I think this code gets more readable when the regex pattern are moved into a private constant, so we can name them properly and intent gets clear.

e.g. 
NON_NEGATIVE_AND_NON_SCIENTIFIC_FLOAT_STRING
NON_NEGATIVE_INTEGER_STRING

tests/PHPStan/Analyser/nsrt/bug-11129.php
staabm: agree. at best put your example in a code comment

src/Reflection/InitializerExprTypeResolver.php
VincentLanglet: Wouldn't it be kinda weird having those two private constant in InitializerExprTypeResolver with 2000+ unrelated lines of code.

staabm: you could also put the information as a comment on the patterns

VincentLanglet: Done

I also discover that `[0-9.]` wasn't ok since it allows 1.1.1.

I reworked the regex and I think we can allow scientific notation too. WDYT ?

staabm: I think we can support scientific. I usually don't support super rare notations to make my life easier :)

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -506,6 +506,10 @@ public function resolveConcatType(Type $left, Type $right): Type
 
 		$leftNumericStringNonEmpty = TypeCombinator::remove($leftStringType, new ConstantStringType(''));
 		if ($leftNumericStringNonEmpty->isNumericString()->yes()) {
+			$allowedRightPattern = $left->isInteger()->yes()
+				? '#^(\d+|\d+.\d+)([eE][+-]?\d+)?$#' // non-negative integer, float or scientific string
```
staabm: Maybe this can even be "doesn't start with minus sign" and is_numeric() so we don't re-implement the logic which defines what a numeric string is?

VincentLanglet: indeed, way better

ondrejmirtes:
Thank you!

PR #4448, opened 2025-10-17
Invalidate native expression types after clearstatcache call

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11484


ondrejmirtes:
Thank you!

PR #4450, opened 2025-10-17
Respect original variable type when using extract on optional keys

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12364

### DIFF
--- a/tests/PHPStan/Rules/Variables/data/bug-12364.php
+++ b/tests/PHPStan/Rules/Variables/data/bug-12364.php
@@ -17,3 +17,24 @@ assertType('string', $x);
 assertType('string|null', $y); // <-- should be: null|string
 var_dump($x);
 var_dump($y); // <-- does exist
+
+/** @return array{xx: string, yy?: string} */
+function foo2(): array {
+	return [ 'xx' => 'foo' ];
+}
+
+function testUndefined()
+{
+	if (rand(0, 1)) {
+		$xx = $yy = 0;
+		assertType('0', $xx);
+		assertType('0', $yy);
+	}
+
+	extract(foo2());
+	assertType('string', $xx);
+
+	if (isset($yy)) {
+		assertType('0|string', $yy);
+	}
+}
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Variables/data/bug-12364.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace Bug12364;
+
+use function PHPStan\Testing\assertType;
+
+/** @return array{x: string, y?: string} */
```
ondrejmirtes: Please add `z?: string` here and test it without assigning first.

I'd say that if `$scope->canAnyVariableExist()` is true and variable certainty is maybe, `$z` has to be `mixed` after extract.

The script can be `include`d from somewhere where `$z` is already defined and is of a different type.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Variables/data/bug-12364.php
VincentLanglet: Done

tests/PHPStan/Rules/Variables/data/bug-12364.php
```diff
@@ -0,0 +1,41 @@
...
+	return [ 'x' => 'foo' ];
+}
+
+$x = $y = null;
+assertType('null', $x);
+assertType('null', $y);
+extract(foo());
+assertType('string', $x);
+assertType('string|null', $y); // <-- should be: null|string
+assertType('mixed', $z);
```
staabm, line 20: I think it would also make sense to `assertVariableCertainty` on the 3 vars

ondrejmirtes: Yup

ondrejmirtes (approved):

tests/PHPStan/Rules/Variables/data/bug-12364.php
```diff
@@ -0,0 +1,46 @@
...
+assertVariableCertainty(TrinaryLogic::createYes(), $x);
+assertVariableCertainty(TrinaryLogic::createYes(), $y); // <-- should be: null|string
```
staabm:
```suggestion
assertVariableCertainty(TrinaryLogic::createYes(), $y);
```

staabm (approved):

PR #4451, opened 2025-10-18
TypeCombinatorTest: turn `return` into `yield`

staabm:
I used rector to transform the test. I think its way more readable to have a single `yield` per test-case instead of a huge `return [[], [], []]` in which its hard to see where test-data from one test starts and where the next one starts etc.

just did the transformation on a single test for now, to discuss the overall idea.

this PR did not transform the `yield from [[], [], []]` into `yield`, as this is [not yet implemented in rector](https://github.com/rectorphp/rector/issues/9443).

wdyt?

### DIFF
--- a/tests/PHPStan/Type/TypeCombinatorTest.php
+++ b/tests/PHPStan/Type/TypeCombinatorTest.php
@@ -66,65 +66,63 @@ use const PHP_VERSION_ID;
 class TypeCombinatorTest extends PHPStanTestCase
 {
 
-	public static function dataAddNull(): array
+	public static function dataAddNull(): Iterator
 	{
-		return [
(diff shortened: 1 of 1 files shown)


staabm:
please do not merge yet, as I need to tripple check the results first. I was just asking for the idea at first

ondrejmirtes:
I'm thinking:

1) I don't want to merge this before we merge and test and release the new PHPUnit rule. I want to see what it does with the current test suite as it has existed for years.
2) I want to eventually do this but with a custom rule here in phpstan-src that uses fixNode() to do this. I don't want to do it once with Rector and then get it wrong again, I want to continuously test it.

PR #4452, opened 2025-10-18
Ignore ApiInstanceofTypeRule in SimultaneousTypeTraverser

VincentLanglet:
Needed for https://github.com/phpstan/phpstan-src/pull/4210

### DIFF
--- a/src/Parser/TypeTraverserInstanceofVisitor.php
+++ b/src/Parser/TypeTraverserInstanceofVisitor.php
@@ -6,12 +6,18 @@ use Override;
 use PhpParser\Node;
 use PhpParser\NodeVisitorAbstract;
 use PHPStan\DependencyInjection\AutowiredService;
+use function in_array;
 
 #[AutowiredService]
 final class TypeTraverserInstanceofVisitor extends NodeVisitorAbstract
 {
 
 	public const ATTRIBUTE_NAME = 'insideTypeTraverserMap';
+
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
I'm not sure about this one yet. It remains to be seen what's the right implementation for a callback in SimultaneousTypeTraverser. Because the implementation of traverseSimultaneously in IntersectionType is not obvious, it might mean that doing `instanceof *Type` in SimultaneousTypeTraverser callback is really not correct.

I'll keep this open, we'll see.

VincentLanglet:
> I'm not sure about this one yet. It remains to be seen what's the right implementation for a callback in SimultaneousTypeTraverser. Because the implementation of traverseSimultaneously in IntersectionType is not obvious, it might mean that doing `instanceof *Type` in SimultaneousTypeTraverser callback is really not correct.
> 
> I'll keep this open, we'll see.

I feel like there is already a flaw ignoring them inside TypeTraverser because while some instanceof might make sens like UnionType or IntersectionType I'm not sure every instanceof make sens.

Personally my main issue was the fact that I replace a TypeTraverser by a simultaneously one in RuleLevelHelper https://github.com/phpstan/phpstan-src/pull/4210/changes#diff-9dace3387909be6cad5cfedf4087102d316daf48f31a033ae258ab119636035bR92 ending with lot of instanceof inside a SimultaneousTypeTraverser.

PR #4454, opened 2025-10-19
Use the same scalar limit for sprintf and concat

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13378


ondrejmirtes:
Thank you!

PR #4455, opened 2025-10-19
Handle impure/throw point of inherited constructors of anonymous classes

takaram:
Improve throw point / impure point handling on anonymous classes when the anonymous class does not declare its own constructor.

Closes phpstan/phpstan#13698


ondrejmirtes:
Perfect, thank you!

PR #4456, opened 2025-10-19
Fix `get_class` return type

VincentLanglet:
Closes https://github.com/phpstan/phpstan-src/pull/2350
Closes https://github.com/phpstan/phpstan/issues/4890

### DIFF
--- a/src/Type/Enum/EnumCaseObjectType.php
+++ b/src/Type/Enum/EnumCaseObjectType.php
@@ -18,6 +18,7 @@ use PHPStan\Type\AcceptsResult;
 use PHPStan\Type\CompoundType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\GeneralizePrecision;
+use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\IsSuperTypeOfResult;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\ObjectType;
@@ -220,6 +221,11 @@ class EnumCaseObjectType extends ObjectType
 	public function getEnumCaseObject(): ?EnumCaseObjectType
 	{
 		return $this;
+	}
+
+	public function getClassStringType(): Type
+	{
+		return new GenericClassStringType(new ObjectType($this->getClassName()));
 	}
 
 	public function toPhpDocNode(): TypeNode
--- a/src/Type/Generic/TemplateMixedType.php
+++ b/src/Type/Generic/TemplateMixedType.php
@@ -63,4 +63,9 @@ final class TemplateMixedType extends MixedType implements TemplateType
 		);
 	}
 
+	public function getClassStringType(): Type
+	{
+		return new GenericClassStringType($this);
+	}
+
 }
--- a/src/Type/Generic/TemplateObjectWithoutClassType.php
+++ b/src/Type/Generic/TemplateObjectWithoutClassType.php
@@ -36,4 +36,9 @@ class TemplateObjectWithoutClassType extends ObjectWithoutClassType implements T
 		$this->default = $default;
 	}
 
+	public function getClassStringType(): Type
+	{
+		return new GenericClassStringType($this);
+	}
+
 }
--- a/src/Type/Php/GetClassDynamicReturnTypeExtension.php
+++ b/src/Type/Php/GetClassDynamicReturnTypeExtension.php
@@ -14,7 +14,6 @@ use PHPStan\Type\Enum\EnumCaseObjectType;
 use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\IntersectionType;
-use PHPStan\Type\ObjectShapeType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\StaticType;
@@ -65,10 +64,6 @@ final class GetClassDynamicReturnTypeExtension implements DynamicFunctionReturnT
 
 				if ($type instanceof EnumCaseObjectType) {
 					return new GenericClassStringType(new ObjectType($type->getClassName()));
-				}
-
-				if ($type instanceof ObjectShapeType) {
-					return new ClassStringType();
 				}
 
 				$isObject = $type->isObject();
--- a/src/Type/Php/GetClassDynamicReturnTypeExtension.php
+++ b/src/Type/Php/GetClassDynamicReturnTypeExtension.php
@@ -10,15 +10,8 @@ use PHPStan\Type\ClassStringType;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
-use PHPStan\Type\Enum\EnumCaseObjectType;
-use PHPStan\Type\Generic\GenericClassStringType;
-use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\IntersectionType;
-use PHPStan\Type\ObjectType;
-use PHPStan\Type\ObjectWithoutClassType;
-use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
-use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeTraverser;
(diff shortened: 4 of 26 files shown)


src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -66,30 +67,35 @@ static function (Type $type, callable $traverse): Type {
...
-					return new UnionType([
-						new ClassStringType(),
-						new ConstantBooleanType(false),
-					]);
-				} elseif ($type instanceof StaticType) {
-					return new GenericClassStringType($type->getStaticObjectType());
-				} elseif ($objectClassNames !== []) {
-					return new GenericClassStringType($type);
-				} elseif ($type instanceof ObjectWithoutClassType) {
+				if ($type instanceof ObjectShapeType) {
```
VincentLanglet: The logic
```
if (!$objectType instanceof TemplateType && $objectType instanceof ObjectWithoutClassType) {
					$classStringType = new ClassStringType();
				} else {
					$classStringType = new GenericClassStringType($objectType);
				}
```
Doesn't work for ObjectShapeType and gaves
```
class-string<object{foo: bar}>
```
which is weird.

src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -66,30 +67,35 @@ static function (Type $type, callable $traverse): Type {
...
 					return new ClassStringType();
 				}
 
-				return new ConstantBooleanType(false);
+				$isObject = $type->isObject();
+				if ($isObject->no()) {
```
VincentLanglet, line 59: Answering https://github.com/phpstan/phpstan-src/pull/2350#discussion_r1443764419

It can be maybe for at least template type or mixed type

src/Type/Php/GetClassDynamicReturnTypeExtension.php
```diff
@@ -66,30 +67,35 @@ static function (Type $type, callable $traverse): Type {
...
+					return new ConstantBooleanType(false);
+				}
+
+				if ($type instanceof StaticType) {
```
VincentLanglet: I use a if/else as suggested (I think) in https://github.com/phpstan/phpstan-src/pull/2350#discussion_r1443764476

ondrejmirtes (changes requested):
I feel like a proper solution would be a new `Type` method. Basically the opposite of `getClassStringObjectType`. That one goes from `class-string<Foo>` to `Foo`.

We need to go from object `Foo` to `class-string<Foo>`.

And personally I don't think there's anything wrong with `class-string<object{foo: Bar}>` because it'd allow for a precise type when doing `new` on it again.

src/Type/Generic/TemplateObjectWithoutClassType.php
```diff
@@ -36,4 +36,9 @@ public function __construct(
 		$this->default = $default;
 	}
 
+	public function getClassStringType(): Type
+	{
+		return new GenericClassStringType($this);
```
VincentLanglet, line 41: Not sure if `$this` should be intersected with ObjectWithoutClassType here.

With the current implementation, `generics.php` assertions are not changed.

But we might prefer having
```
assertType(
		'class-string<object&T (function PHPStan\Generics\FunctionsAssertType\getClassOnTemplateType(), argument)>|false',
		get_class($a)
	);
```
over
```
assertType(
		'class-string<T (function PHPStan\Generics\FunctionsAssertType\getClassOnTemplateType(), argument)>|false',
		get_class($a)
	);
```

If later we're doing `new` on it. WDYT ?

VincentLanglet:
> I feel like a proper solution would be a new `Type` method. Basically the opposite of `getClassStringObjectType`. That one goes from `class-string<Foo>` to `Foo`.
> 
> We need to go from object `Foo` to `class-string<Foo>`.

I introduced `getClassStringType()`, feel free to challenge the name :)

src/Type/Type.php
```diff
@@ -40,6 +40,8 @@ public function getObjectClassNames(): array;
 	 */
 	public function getObjectClassReflections(): array;
 
+	public function getClassStringType(): Type;
```
ondrejmirtes, line 74: Would be nice to document what this is supposed to return similarly to how `getClassStringObjectType` and `getObjectTypeOrClassStringObjectType` are documented.

src/Type/Generic/TemplateObjectWithoutClassType.php
ondrejmirtes: You're asking about this in TemplateObjectWithoutClassType. Intersecting this type with ObjectWithoutClassType is no-op; it's always an object.

ondrejmirtes (changes requested):

src/Type/Generic/TemplateObjectWithoutClassType.php
VincentLanglet: Ok

src/Type/Type.php
VincentLanglet: I added `Return class-string<Foo> for object type Foo.` ; is it enough ?

src/Type/Enum/EnumCaseObjectType.php
```diff
@@ -222,6 +223,11 @@ public function getEnumCaseObject(): ?EnumCaseObjectType
 		return $this;
 	}
 
+	public function getClassStringType(): Type
+	{
+		return new GenericClassStringType(new ObjectType($this->getClassName()));
+	}
+
```
staabm, line 230: from my experience enum-case can generate a separate class of bugs, therefore I think we should have a separate test for it

src/Type/Generic/TemplateMixedType.php
```diff
@@ -63,4 +63,9 @@ public function toStrictMixedType(): TemplateStrictMixedType
 		);
 	}
 
+	public function getClassStringType(): Type
+	{
+		return new GenericClassStringType($this);
+	}
+
```
staabm, line 70: might also be worth testing with templates

src/Type/Enum/EnumCaseObjectType.php
VincentLanglet: I don't understand which test you expect since the bug was about class with `hasProperty` and an enum-case cannot have a property.

staabm: but it could have a method

VincentLanglet: I might be wrong but since an enum is final `method_exists` calls will be either always true or always false on it and the HasMethodType won't be added.

I'm getting
```
enum MyEnum
{
	case CASE1;
	case CASE2;

	public function someMethod(): bool { return true; }
}

class HelloWorld
{
	public function withEnumCase(): void
	{
		$entity = MyEnum::CASE1;
		assertType('class-string<Bug4890PHP8\MyEnum>', get_class($entity));
		assert(method_exists($entity, 'someMethod'));
		assertType('class-string<Bug4890PHP8\MyEnum>', get_class($entity));
	}
}
```

Do you have a test in mind ?

staabm: what about

```php
enum MyEnum
{
	case CASE1;
	case CASE2;

	public function someMethod(): bool { return true; }
}

class HelloWorld
{
	public function withEnumCase(UnitEnum $entity): void
	{
		assertType('class-string<UnitEnum>', get_class($entity));
		assert(method_exists($entity, 'someMethod'));
		assertType('class-string<UnitEnum&hasMethod<someMethod>>', get_class($entity));
	}
}

$h = new HelloWorld();
$h->withEnumCase(MyEnum::CASE1);
```

VincentLanglet: Right, I forgot bout the base enum classes... UnitEnum.

I added the test !

src/Type/Generic/TemplateMixedType.php
VincentLanglet: Also added

staabm (approved):
This PR is so much better than my take in https://github.com/phpstan/phpstan-src/pull/2350

great job!

VincentLanglet:
Since you initially request changes, I need your approval @ondrejmirtes

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #4457, opened 2025-10-19
Introduce `__stringnotstringable`

VincentLanglet:
Naming to debate ? `__stringwithoutstringable` ? `__strictstring` ?

`intval($stringable)` is never allowed, even without strict_types.

The current check in FunctionCallParameterCheck is
```
$this->ruleLevelHelper->accepts($parameterType, $argumentValueType, $scope->isDeclareStrictTypes());
```
either we're doing something like
```
$this->ruleLevelHelper->accepts(
      $parameterType,
      $argumentValueType,
      $scope->isDeclareStrictTypes() || in_array($functionName, ['floatval', 'intval', 'doubleval'], true)
);
```
either we have to introduce the opposite of StringAlwaysAcceptingObjectWithToStringType like I did.

Closes https://github.com/phpstan/phpstan/issues/6560
Closes https://github.com/phpstan/phpstan-src/pull/1005


ondrejmirtes (changes requested):
Look at StringAlwaysAcceptingObjectWithToStringType. Every new Type implementation also has to override isSuperTypeOf to give correct answers.

You can test that in TypeCombinatorTest::testUnion and testIntersect.

tests/PHPStan/Rules/Functions/data/bug-6560.php
```diff
@@ -0,0 +1,98 @@
+<?php
```
staabm, line 1: in https://github.com/phpstan/phpstan/issues/6560#issuecomment-3274299888 there was mentioned this does not depend on strict types.

I think it would be useful to have another test with strict-types enabled
(in case this is not already covered somewhere)

VincentLanglet: test added

staabm (approved):

tests/PHPStan/Rules/Functions/data/bug-6560-strict.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
```
staabm: if we put `declare(strict_types=1);` on line 1 - right behind the starting tag - you might re-use the very same test without duplicating expectations in `CallToFunctionParametersRuleTest` ?

staabm (approved):
lgtm. needs cs-fix though

ondrejmirtes (approved):

PR #4463, opened 2025-10-21
Fix sprintf() types and placeholders

staabm:
fixes all strict errors which can be seen in https://github.com/phpstan/phpstan-src/pull/4461


ondrejmirtes:
Nice, thank you!

PR #4465, opened 2025-10-21
Narrow types

staabm:
at first I thought this would fix a sprintf() PHP 7.4 only error.. but it didn't .. still a good type improvement IMO

### DIFF
--- a/src/Type/Type.php
+++ b/src/Type/Type.php
@@ -28,7 +28,7 @@ interface Type
 {
 
 	/**
-	 * @return list<string>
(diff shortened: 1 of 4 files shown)


src/Type/Type.php
```diff
@@ -28,7 +28,7 @@ interface Type
 {
 
 	/**
-	 * @return list<string>
+	 * @return list<non-empty-string>
```
staabm, line 31: analog `Type->getObjectClassNames()`

ondrejmirtes:
Thank you!

PR #4466, opened 2025-10-21
Fix strtr inferences

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13708

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/strtr.php
+++ b/tests/PHPStan/Analyser/nsrt/strtr.php
@@ -24,4 +24,16 @@ function doFoo(string $s, $nonEmptyString, $nonFalseyString) {
 	assertType('non-empty-string', strtr($nonFalseyString, $s, $nonEmptyString));
 	assertType('non-falsy-string', strtr($nonFalseyString, $nonEmptyString, $nonFalseyString));
 	assertType('non-falsy-string', strtr($nonFalseyString, $nonFalseyString, $nonFalseyString));
+
+	assertType('string', strtr($s, [$s => $nonEmptyString]));
+	assertType('string', strtr($s, [$nonEmptyString => $nonEmptyString]));
+	assertType('string', strtr($s, [$nonFalseyString => $nonFalseyString]));
+
+	assertType('non-empty-string', strtr($nonEmptyString, [$s => $nonEmptyString]));
+	assertType('non-empty-string', strtr($nonEmptyString, [$nonEmptyString => $nonEmptyString]));
+	assertType('non-empty-string', strtr($nonEmptyString, [$nonFalseyString => $nonFalseyString]));
+
+	assertType('non-empty-string', strtr($nonFalseyString, [$s => $nonEmptyString]));
+	assertType('non-falsy-string', strtr($nonFalseyString, [$nonEmptyString => $nonFalseyString]));
+	assertType('non-falsy-string', strtr($nonFalseyString, [$nonFalseyString => $nonFalseyString]));
 }
--- a/tests/PHPStan/Analyser/nsrt/strtr.php
+++ b/tests/PHPStan/Analyser/nsrt/strtr.php
@@ -7,8 +7,9 @@ use function PHPStan\Testing\assertType;
 /**
  * @param non-empty-string $nonEmptyString
  * @param non-falsy-string $nonFalseyString
+ * @param mixed $mixed
  */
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Methods/data/bug-13708.php
```diff
@@ -0,0 +1,23 @@
...
+	 */
+	public function takeNonEmpty(string $s): void
+	{
+		return;
+	}
+
+	public function doStuff(): void
+	{
+		$this->takeNonEmpty(
+			strtr('change {me}', ['{me}' => 'me'])
```
staabm, line 20: Would be good to have an assertType() somewhere so we see the concrete type when all args are constant like here

VincentLanglet: I have added multiple assertion in the existing nsrt file https://github.com/phpstan/phpstan-src/pull/4466/commits/93b42a2e34eaa04095d1d7f1cf8888a03724571f

ondrejmirtes:
Escaped mutants here. This means tests need to be written so that when the code changes like that, there will be some tests that will fail. Possibly the code logic might be wrong too.

VincentLanglet:
> Escaped mutants here. This means tests need to be written so that when the code changes like that, there will be some tests that will fail. Possibly the code logic might be wrong too.

I'm not sure what should be expected here.

The mutant can only occurs if we pass `string|array` as second param and wether it infers non-empty-string or not seems kinda useless to me because it's already reported as an error by PHPStan which should be fixed first https://phpstan.org/r/1b1f1f8a-dbd4-417c-b251-be553d4d64d3.

Also if I try to support "maybe array" with something like
```
TypeCombinator::intersect(new ArrayType(new MixedType(), new MixedType()), $secondArgumentType)
```
I'll still get "false positive" escaped mutant because now `$type->isArray()->maybe()` will be useless and therefor `->yes()` or `! -> no()` are equivalent.

ondrejmirtes:
Maybe array is also for example `mixed` or `array|null`.

> kinda useless to me because it's already reported as an error by PHPStan

Type inference and error reporting are two independent different parts of PHPStan. Type inference should still do something sensible even if the user passes a wrong type there. Especially when the wrong type might not even be reported on a lower rule level.

VincentLanglet:
No mutant escaped now @ondrejmirtes :) 

The lint 8.0 failure seems unrelated/random

VincentLanglet:
> No mutant escaped now @ondrejmirtes :)
> 
> The lint 8.0 failure seems unrelated/random

I also rebased and fixed all my existing PR if you have time for a look https://github.com/phpstan/phpstan-src/pulls/VincentLanglet.
I commented those with failing build, some failure are unrelated/random ; and some mutant are (I think) not fixable.

tests/PHPStan/Analyser/nsrt/strtr.php
```diff
@@ -24,4 +25,27 @@ function doFoo(string $s, $nonEmptyString, $nonFalseyString) {
...
+	assertType('string', strtr($s, [$nonEmptyString => $nonEmptyString]));
+	assertType('string', strtr($s, [$nonFalseyString => $nonFalseyString]));
+
+	assertType('non-empty-string', strtr($nonEmptyString, [$s => $nonEmptyString]));
+	assertType('non-empty-string', strtr($nonEmptyString, [$nonEmptyString => $nonEmptyString]));
+	assertType('non-empty-string', strtr($nonEmptyString, [$nonFalseyString => $nonFalseyString]));
+
+	assertType('non-empty-string', strtr($nonFalseyString, [$s => $nonEmptyString]));
+	assertType('non-falsy-string', strtr($nonFalseyString, [$nonEmptyString => $nonFalseyString]));
+	assertType('non-falsy-string', strtr($nonFalseyString, [$nonFalseyString => $nonFalseyString]));
```
staabm, line 39:
```suggestion
	assertType('non-falsy-string', strtr($nonFalseyString, [$nonFalseyString => $nonFalseyString]));

	// see https://3v4l.org/jCUhW/rfc#vgit.master
	assertType('string', strtr($nonFalseyString, [$nonFalseyString => $s]));
	assertType('string', strtr($nonEmptyString, [$nonEmptyString => $s]));
```

VincentLanglet: Done

staabm (approved):

PR #4467, opened 2025-10-23
Fix crashes in ParametersAcceptorSelector

staabm:
closes https://github.com/phpstan/phpstan/issues/13714

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -1556,6 +1556,19 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug13714(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-13714.php');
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1556,6 +1556,12 @@ public function testBug13310(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug13714(): void
+	{
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-13714.php');
+		$this->assertCount(19, $errors);
```
ondrejmirtes: What are the 19 errors? Can you produce code that makes PHPStan crash but once fixed, it produces no errors?

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
staabm: no errors seems hard todo, but I reduced it to a minimum

ondrejmirtes:
Awesome, thank you!

PR #4468, opened 2025-10-23
Added PHP_BUILD_DATE type (PHP 8.5+)

staabm:
before this PR PHP_BUILD_DATE was inferred as a hardcoded string

see 
- https://3v4l.org/oHPdU/rfc#vgit.master
- https://phpstan.org/r/349be1d8-9544-480a-a747-14364523dafd
- https://php.watch/versions/8.5/PHP_BUILD_DATE

### DIFF
--- a/src/Analyser/ConstantResolver.php
+++ b/src/Analyser/ConstantResolver.php
@@ -159,6 +159,13 @@ final class ConstantResolver
 
 			return $this->createInteger($minVersion, $maxVersion);
 		}
+		// added in PHP 8.5
+		if ($resolvedConstantName === 'PHP_BUILD_DATE') {
+			return new IntersectionType([
+				new StringType(),
+				new AccessoryNonFalsyStringType(),
+			]);
+		}
 		if ($resolvedConstantName === 'PHP_ZTS') {
 			return new UnionType([
 				new ConstantIntegerType(0),
--- a/tests/PHPStan/Analyser/nsrt/predefined-constants-php85.php
+++ b/tests/PHPStan/Analyser/nsrt/predefined-constants-php85.php
@@ -1,0 +1,5 @@
+<?php
+
+use function PHPStan\Testing\assertType;
+
+assertType('non-falsy-string', PHP_BUILD_DATE);


ondrejmirtes:
Thank you.

ondrejmirtes:
More should be updated probably: `PHP_SBINDIR`, `PHP_BUILD_PROVIDER`.

Also there's a big list of `dynamicConstantNames` in the config. Maybe it could be refactored so that the types are in the config as well and not source code.

SanderMuller:
> Maybe it could be refactored so that the types are in the config as well and not source code.

That already works, it is just not used for the built-ins. `dynamicConstantNames` takes a map as well as a list, and `getConfiguredGlobalConstantType()` resolves the value through `TypeStringResolver`. On a constant defined as `'hello'`:

| config | inferred |
| --- | --- |
| no entry | `'hello'` |
| `- MY_DYNAMIC` | `string` |
| `MY_DYNAMIC: 'int<0, 5>'` | `int<0, 5>` |

So it is moving data, not building machinery. All but eight of the branches in `resolvePredefinedConstant()` are static and spell out as a type string.

The eight that cannot move are the four version constants, which come from `phpVersion` and the Composer range, plus `NAN`, `INF`, `PHP_INT_MAX` and `PHP_INT_MIN`. The last two branch on the **analysing** machine's `PHP_INT_SIZE`. A 32-bit analyser tells a 64-bit project that `PHP_INT_MAX` is exactly `2147483647`, which is the same leak this PR fixed for strings.

Two calls for you before I write anything:

1. **Precedence would invert.** `resolveConstant()` asks `resolvePredefinedConstant()` first, so code beats config today. Move the built-ins and a user entry for `PHP_EOL` starts winning. Deliberate, or to be prevented?
2. **`PHP_INT_MAX` / `PHP_INT_MIN`** — worth fixing the 32-bit leak while in there, or leave them alone?

The other two constants you named are in phpstan/phpstan-src#6454.

staabm:
> 1. **Precedence would invert.** `resolveConstant()` asks `resolvePredefinedConstant()` first, so code beats config today. Move the built-ins and a user entry for `PHP_EOL` starts winning. Deliberate, or to be prevented?

IMO we should not allow users to overwrite types for php built-in constant for which phpstan-src defines a type.

the phpstan config feature is primarily meant for userland constant types to be narrowed

> 2\. **`PHP_INT_MAX` / `PHP_INT_MIN`** — worth fixing the 32-bit leak while in there, or leave them alone?

sounds like a separate PR/problem, not sure worth fixing atm

PR #4472, opened 2025-10-23
Avoid false error on is_subclass_of

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13713

Solves https://github.com/phpstan/phpstan/issues/6305#issuecomment-1057180525

### DIFF
--- a/src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
+++ b/src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
@@ -38,13 +38,18 @@ final class IsAFunctionTypeSpecifyingHelper
 
 		return TypeTraverser::map(
 			$classType,
-			static function (Type $type, callable $traverse) use ($objectOrClassTypeClassNames, $allowString, $allowSameClass): Type {
+			static function (Type $type, callable $traverse) use ($objectOrClassType, $objectOrClassTypeClassNames, $allowString, $allowSameClass): Type {
 				if ($type instanceof UnionType || $type instanceof IntersectionType) {
 					return $traverse($type);
 				}
 				if ($type instanceof ConstantStringType) {
 					if (!$allowSameClass && $objectOrClassTypeClassNames === [$type->getValue()]) {
-						return new NeverType();
+						// For objectType we cannot be sure since 'Foo' is used for both
+						// - the Foo class
+						// - a child of foo class
+						if ($objectOrClassType->isString()->yes()) {
+							return new NeverType();
+						}
 					}
 					if ($allowString) {
 						return new UnionType([
--- a/src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
+++ b/src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
@@ -47,14 +47,18 @@ final class IsAFunctionTypeSpecifyingHelper
 				}
 				if ($type instanceof ConstantStringType) {
 					if (!$allowSameClass) {
-						// For objectType we cannot be sure since 'Foo' is used for both
-						// - the Foo class
-						// - a child of foo class
-						if (
-							$objectOrClassTypeClassNames === [$type->getValue()]
-							&& $objectOrClassType->isString()->yes()
-						) {
-							return new NeverType();
+						if ($objectOrClassTypeClassNames === [$type->getValue()]) {
+							$isSameClass = true;
+							foreach ($objectOrClassType->getObjectClassReflections() as $classReflection) {
+								if (!$classReflection->isFinal()) {
+									$isSameClass = false;
+									break;
+								}
+							}
+
+							if ($isSameClass) {
+								return new NeverType();
+							}
 						}
 
 						if (
--- a/src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
+++ b/src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
@@ -16,6 +16,7 @@ use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\UnionType;
 use function array_unique;
 use function array_values;
+use function in_array;
 
 #[AutowiredService]
 final class IsAFunctionTypeSpecifyingHelper
@@ -26,7 +27,7 @@ final class IsAFunctionTypeSpecifyingHelper
 		Type $classType,
 		bool $allowString,
 		bool $allowSameClass,
-	): Type
+	): ?Type
 	{
 		$objectOrClassTypeClassNames = $objectOrClassType->getObjectClassNames();
 		if ($allowString) {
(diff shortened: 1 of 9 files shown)


src/Type/Php/IsAFunctionTypeSpecifyingHelper.php
```diff
@@ -36,15 +37,35 @@ public function determineType(
 			$objectOrClassTypeClassNames = array_values(array_unique($objectOrClassTypeClassNames));
 		}
 
-		return TypeTraverser::map(
+		$isUncertain = $classType->getConstantStrings() === [];
```
VincentLanglet, line 40: From the original check
```
// prevent false-positives in IsAFunctionTypeSpecifyingHelper
		if ($classType->getConstantStrings() === [] && $resultType->isSuperTypeOf($objectOrClassType)->yes()) {
			return new SpecifiedTypes([], []);
		}
```

tests/PHPStan/Rules/Comparison/data/bug-13713.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Bug13713;
+
+function debug(object $object): void {
+	if ($object instanceof \stdClass) {
+		echo var_export(\is_subclass_of($object, \stdClass::class, false), true) . \PHP_EOL;
```
VincentLanglet, line 9: With this PR there is no error here which fix https://github.com/phpstan/phpstan/issues/13713
but I dunno why the bot is not detecting the change...

VincentLanglet:
This issue was more complicated than expected... I'm not against your opinion/review @staabm :)

tests/PHPStan/Analyser/nsrt/bug-6305.php
```diff
@@ -15,5 +15,5 @@ class B extends A {}
 }
 
 if (is_subclass_of($b, B::class)) {
-	assertType('*NEVER*', $b);
+	assertType('Bug6305Types\B', $b); // Could be NEVER
```
staabm: IIRC - as we have `$b = new B();` we can make this `NEVER`, as its implicitly `final`

VincentLanglet: I agree that here it could be never.

But correct me if I'm wrong, currently in the type system we cannot make a difference between
```
$b = new B();
```
and
```
function getB(): B {}

$b = getB();
```
no ?

Therefor, currently I feel like it's better to
- Not be able to detect it's never
- Not having false positive like https://phpstan.org/r/25c06554-9cd6-42ea-bd30-d86dc12d94b6

Rather than the opposite.

staabm: is it implemented in `hasFinalByKeywordOverride` with https://github.com/phpstan/phpstan-src/commit/3019d391130102baa91af881771211caceff2325 ?

VincentLanglet: You rocks @staabm.

Improved with https://github.com/phpstan/phpstan-src/pull/4472/commits/95c28b0e86ec40e26407a9204f4d48220951dc23

tests/PHPStan/Analyser/nsrt/is-subclass-of.php
```diff
@@ -4,7 +4,7 @@
 
 function (Bar $a, Bar $b, Bar $c, Bar $d) {
 	if (is_subclass_of($a, Bar::class)) {
```
staabm, line 6: do we have a test somewhere for use of `is_subclass_of` with a final-class ?

VincentLanglet: Added

VincentLanglet:
I dunno what to do about the webmozart-assert failure because I don't report anymore
```
Call to static method Webmozart\Assert\Assert::implementsInterface() with class-string<WebmozartAssertImpossibleCheck\Bar>|WebmozartAssertImpossibleCheck\Bar and 'WebmozartAssertImpossibleCheck\\Bar' will always evaluate to true.
```

Technically, because a `implementsInterface` check is called before it's indeed always true.
But strictly looking at the type, calling 
```
class-string<WebmozartAssertImpossibleCheck\Bar>|WebmozartAssertImpossibleCheck\Bar
```

And currently, https://github.com/phpstan/phpstan-webmozart-assert/pull/194 is showing the issue

ondrejmirtes:
This needs to be fixed in phpstan-webmozart-assert before we can merge this: https://github.com/phpstan/phpstan-webmozart-assert/blob/b2221d38ee5338531c0bcaa985268ae5bcef452e/src/Type/WebMozartAssert/AssertTypeSpecifyingExtension.php#L401-L419

VincentLanglet:
> This needs to be fixed in phpstan-webmozart-assert before we can merge this: [phpstan/phpstan-webmozart-assert@`b2221d3`/src/Type/WebMozartAssert/AssertTypeSpecifyingExtension.php#L401-L419](https://github.com/phpstan/phpstan-webmozart-assert/blob/b2221d38ee5338531c0bcaa985268ae5bcef452e/src/Type/WebMozartAssert/AssertTypeSpecifyingExtension.php#L401-L419)

It's been a long time I worked on this PR, but the implementation looks fine to me.
PHPStan will still report most of always-true, always-false `implementsInterface` calls.

The only issue is the fact it cannot detect
```
if (\is_subclass_of($object, \stdClass::class, false)) {
      // $object is class-string<\stdClass>|\stdClass
		if (\is_subclass_of($object, \stdClass::class, false)) { // This is not detected as always true
		}
	}
```
Because I dunno currently a way to say it cannot be the class exactly but only a child or for the call to be remembered.
Maybe you know @staabm ?

And so far I thought it was better to have a missing "always true" report rather than a wrong one.

staabm:
> Because I dunno currently a way to say it cannot be the class exactly but only a child or for the call to be remembered.
> Maybe you know @staabm ?

just a guess:

`$a->isSuperTypeOf($b)->yes() && $b->isSuperTypeOf($a)->yes() && !$a->equals($b)`

or maybe `isSubTypeOf()`

> And so far I thought it was better to have a missing "always true" report rather than a wrong one.

absolutely

VincentLanglet:
> > Because I dunno currently a way to say it cannot be the class exactly but only a child or for the call to be remembered.
> > Maybe you know @staabm ?
> 
> just a guess:
> 
> `$a->isSuperTypeOf($b)->yes() && $b->isSuperTypeOf($a)->yes() && !$a->equals($b)`
> 
> or maybe `isSubTypeOf()`

It's more with the type instances after 
```
if (\is_subclass_of($object, \stdClass::class, false)) {
```
how to specify better than `class-string<\stdClass>|\stdClass`

I feel like we cannot do this without a `ChildOfObjectType()` or a boolean in ObjectType to require child only and then we specify to `class-string<child-of<\stdClass>>|child-of<\stdClass>`

So right now @ondrejmirtes, I feel like the best we can is
- Avoid false error on is_subclass_of where there is uncertainty
- Lose some "always true" error reported

And a next step could be to introduce a `childOnly` bool in ObjectType constructor (?)

ondrejmirtes:
So what would you change in phpstan-webmozart-assert after merging this to make the build succeed and make sense?

ondrejmirtes:
Thank you!

VincentLanglet:
> So what would you change in phpstan-webmozart-assert after merging this to make the build succeed and make sense?

This is the only way I have so far to fix the CI on webmozart side https://github.com/phpstan/phpstan-webmozart-assert/pull/208 ; we can't report those always-true check yet without having false positive (like shown in https://github.com/phpstan/phpstan-webmozart-assert/pull/194)

PR #4473, opened 2025-10-24
`array_key_exist` inference is lost when adding a false condition

### LINKED ISSUE #100: AssertArrayHasKey seems to be not supported
Hi @ondrejmirtes, I just got an issue with my tests saying that an offset does not exist in an array.
I tried to add a `assertArrayHasKey` check, without success.

I think this assert is not handle by this plugin, am I wrong ?

I'll try to do a PR in order to add the support, I think I'll be able to copy the way `isArray` works.
https://github.com/phpstan/phpstan-phpunit/blob/master/src/Type/PHPUnit/Assert/AssertTypeSpecifyingExtensionHelper.php#L158

staabm:
closes https://github.com/phpstan/phpstan/issues/11276

fixes https://github.com/phpstan/phpstan-phpunit/issues/100

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -240,6 +240,7 @@ jobs:
         run: |
           php build-infection/bin/infection-config.php \
             --source-directory='build/PHPStan/Build' \
+            --timeout=180 \
           > infection.json5
           cat infection.json5 | jq
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -240,7 +240,6 @@ jobs:
         run: |
           php build-infection/bin/infection-config.php \
             --source-directory='build/PHPStan/Build' \
-            --timeout=300 \
           > infection.json5
           cat infection.json5 | jq
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -240,7 +240,7 @@ jobs:
         run: |
           php build-infection/bin/infection-config.php \
             --source-directory='build/PHPStan/Build' \
-            --timeout=180 \
+            --timeout=300 \
           > infection.json5
           cat infection.json5 | jq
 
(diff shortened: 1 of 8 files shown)


VincentLanglet:
So no issue with this PR ?

Thanks for finding it !

staabm:
I think it works (but did not re-use your previous implementation)

src/Analyser/TypeSpecifier.php
```diff
@@ -663,7 +663,24 @@ public function specifyTypesInCondition(
 			$leftTypes = $this->specifyTypesInCondition($scope, $expr->left, $context)->setRootExpr($expr);
 			$rightScope = $scope->filterByFalseyValue($expr->left);
 			$rightTypes = $this->specifyTypesInCondition($rightScope, $expr->right, $context)->setRootExpr($expr);
-			$types = $context->true() ? $leftTypes->normalize($scope)->intersectWith($rightTypes->normalize($rightScope)) : $leftTypes->unionWith($rightTypes);
+
+			if ($context->true()) {
+				if (
+					$scope->getType($expr->left)->isFalse()->yes()
```
ondrejmirtes: My head hurts 😅 

I think all the `isTrue`/`isFalse` calls should do `toBoolean()` beforehand.

ondrejmirtes (changes requested):

src/Analyser/TypeSpecifier.php
staabm: great catch. fixed.

thank you.

staabm: hmm maybe we should have a `$type->isFalse()->yes()`/`$type->isTrue()->yes()` to `$type->toBoolean()->isFalse()->yes()`/`$type->toBoolean()->isTrue()->yes()` mutator.

it would make us aware of false/falsey coverage

.github/workflows/tests.yml
```diff
@@ -240,6 +240,7 @@ jobs:
         run: |
           php build-infection/bin/infection-config.php \
             --source-directory='build/PHPStan/Build' \
+            --timeout=180 \
```
staabm: in the worst case infection needs enough time to run the full test-suite.
otherwise mutants are considered 'skipped', which we saw in CI:

grafik


this setting means: "how many seconds can infection spent on a single mutant"

staabm: opened a new ticket, as we need more information to proceed.
https://github.com/infection/infection/issues/2510

even after raising the timeout to 300 it still times out.
so at phpstan-src repo we are so slow, that we can't even check 3 mutants in 15minutes CI runtime.

we can merge for now

staabm: lets fix this in a separate PR

ondrejmirtes:
Make a build-infection issue for the new mutator idea 😊

ondrejmirtes:
Thank you!

PR #4474, opened 2025-10-24
feat: isEnum class-string narrowing

SamMousa:
Implements https://github.com/phpstan/phpstan/issues/13712


SamMousa:
It seems many of the failures are not related to my changes....

staabm:
yes, atm a few spurious errors are in CI.

these here are related though: https://github.com/phpstan/phpstan-src/actions/runs/18773626957/job/53563151392?pr=4474

SamMousa:
I'll have another look after the weekend, for now any annotation I add seems to have no effect on the asserted type.

tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -238,6 +238,7 @@ private static function findTestFiles(): iterable
 		yield __DIR__ . '/../Rules/Methods/data/bug-4801.php';
 		yield __DIR__ . '/../Rules/Arrays/data/narrow-superglobal.php';
 		yield __DIR__ . '/../Rules/Methods/data/bug-12927.php';
+		yield __DIR__ . '/data/reflectionclass-isEnum.php';
```
staabm: This line is not required if your tests resides in nsrt/ folder

SamMousa:
It feels like the annotations in the `ReflectionClass.stub` file are ignored.
```php
    /**
     * @return false
     */
    public function isEnum(): bool
    {
    }
```

```
Failed assertions in /projects/phpstan-src/tests/PHPStan/Analyser/nsrt/reflectionclass-isEnum.php:

Line 13:
Expected: false
Actual:   bool
```

Is there some kind of build step for the stubs that i'm missing?

ondrejmirtes:
There's no build step for stubs. I don't know why it does not work for you but I'd argue that:

```php
    /**
     * @return false
     */
    public function isEnum(): bool
    {
    }
```

is wrong: https://3v4l.org/AZfkh

SamMousa:
It is wrong 100%, but my objective here was to test if the annotation is working at all.
The simplest annotation I could think of was `@return false`.... 

And if this already doesn't work I need to figure that part out first before I test my actual more complex annotations...

ondrejmirtes:
Do not forget to also update ReflectionClassWithLazyObjects.stub.

SamMousa:
> Do not forget to also update ReflectionClassWithLazyObjects.stub.

That was the cause, I did not realize there are 2 stubs for this class. Thanks, now I can get to work on the actual implementation

stubs/ReflectionClass.stub
```diff
@@ -48,4 +48,15 @@ class ReflectionClass
     public function getAttributes(?string $name = null, int $flags = 0)
     {
     }
+
+	/**
+	 * @phpstan-assert-if-true class-string<\UnitEnum> $this->name
+	 * @phpstan-assert-if-true class-string<\UnitEnum> $this->getName()
+	 * @phpstan-assert-if-true ReflectionClass<UnitEnum> $this
```
SamMousa, line 55: This assertion does not work currently. Is there anyway to narrow the type `T` for a generic?

I've tried using `@phsptan-assert-if-true UnitEnum T`, but that doesn't work.

SamMousa:
Current implementation works! I'd like to actually narrow the class (template type `T`) of the reflection object as well, but that doesn't seem to work yet.

PR #4478, opened 2025-10-24
Report invalid printf placeholder

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/1889
Closes https://github.com/phpstan/phpstan/issues/8547 (cc @staabm)


tests/PHPStan/Rules/Functions/data/printf.php
```diff
@@ -6,10 +6,10 @@
 sprintf('%s %s', 'foo'); // one parameter missing
 sprintf('foo', 'foo'); // one parameter over
 sprintf('foo %s', 'foo', 'bar'); // one parameter over
-sprintf('%2$s %1$s %% %1$s %%%', 'one'); // one parameter missing
```
VincentLanglet: Adding a second parameter does not solve the error since `%%%` was invalid placeholder.

tests/PHPStan/Rules/Functions/data/printf.php
```diff
@@ -6,10 +6,10 @@
...
+sprintf('%2$s %1$s %% %1$s %%', 'one'); // one parameter missing
 sprintf('%2$s %%'); // two parameters required
 sprintf('%2$s %1$s %1$s %s %s %s %s'); // four parameters required
-sprintf('%2$s %1$s %% %s %s %s %s %%% %%%%', 'one', 'two', 'three', 'four'); // ok
```
VincentLanglet: This was producing an error https://3v4l.org/2dPI6

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -66,7 +66,17 @@ public function processNode(Node $node, Scope $scope): array
 		foreach ($formatArgType->getConstantStrings() as $formatString) {
 			$format = $formatString->getValue();
 
-			$placeHoldersCounts[] = $this->printfHelper->getPrintfPlaceholdersCount($format);
+			$count = $this->printfHelper->getPrintfPlaceholdersCount($format);
+			if ($count === null) {
+				return [
+					RuleErrorBuilder::message(sprintf(
+						'Call to %s contains an invalid placeholder.',
```
ondrejmirtes, line 73: Would be nice to know what the invalid placeholder is and its position. Otherwise it'd be hard to debug with long format strings as `%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%c%`.

ondrejmirtes (changes requested):

ondrejmirtes:
Also maybe this belongs to bleeding edge only?

tests/PHPStan/Rules/Functions/data/printf.php
staabm: we had such test problems in the past. I wonder whether we can somehow detect such bogus tests
(in the general case, not just in (s)printf)

src/Rules/Functions/PrintfArrayParametersRule.php
VincentLanglet: I thought about this and I was wondering if we should iterate and add more details with `acceptsReasonsTip` later because:
- This would require some weird signature for getPrintfPlaceholdersCount (Returning the count + the invalid placeholders/positions)

- This would require to handle error message for
	- Union string `%c%`|`%c%c%` => What would be the message ? 3rd position ? 5th position ?
	- String with multiple wrong placeholder like `%%% %%%` => Should we report all the invalid ones ?

=> maybe you already have an idea about the format of the message @ondrejmirtes ?

VincentLanglet:
> Also maybe this belongs to bleeding edge only?

I would have consider this kinda as a "bugfix". 
Technically the `sprintf('%')` call should have been reported since there is was missing parameter https://3v4l.org/lAIAAs

I can put this behind bleeding edge, but that would mean adding an extra parameters to `getPrintfPlaceholdersCount` methods with something like `getPrintfPlaceholdersCount($format, $reportInvalidPlaceholder)`. It's your call !

src/Rules/Functions/PrintfArrayParametersRule.php
```diff
@@ -66,7 +66,17 @@ public function processNode(Node $node, Scope $scope): array
...
+						$name,
+					))->identifier(sprintf('argument.%s', $name))->build(),
```
ondrejmirtes, line 75: What's this identifier going to be?

src/Rules/Functions/PrintfParametersRule.php
```diff
@@ -86,6 +86,15 @@ public function processNode(Node $node, Scope $scope): array
 				$tempPlaceHoldersCount = $this->printfHelper->getScanfPlaceholdersCount($format);
 			}
 
+			if ($tempPlaceHoldersCount === null) {
+				return [
+					RuleErrorBuilder::message(sprintf(
+						'Call to %s contains an invalid placeholder.',
+						$name,
+					))->identifier(sprintf('argument.%s', $name))->build(),
```
ondrejmirtes, line 94: What's this identifier going to be?

ondrejmirtes (changes requested):

src/Rules/Functions/PrintfArrayParametersRule.php
VincentLanglet: I copied the existing one
https://github.com/phpstan/phpstan-src/blob/43d502614f7a7337f58ddeb1f963a52cfec79e03/src/Rules/Functions/PrintfArrayParametersRule.php#L139

The identifier will be `argument.vprintf` or `argument.vsprintf`.

src/Rules/Functions/PrintfParametersRule.php
VincentLanglet: I copied the existing one
https://github.com/phpstan/phpstan-src/blob/43d502614f7a7337f58ddeb1f963a52cfec79e03/src/Rules/Functions/PrintfParametersRule.php#L113

Identifier will be `argument.printf` and similar

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #4480, opened 2025-10-25
Setup mutation testing

staabm:
analog
- https://github.com/phpstan/phpstan-doctrine/pull/687
- https://github.com/phpstan/phpstan-doctrine/pull/686
- https://github.com/phpstan/phpstan-doctrine/pull/699


ondrejmirtes:
AFAIK the usage of build-infection isn't yet merged in phpstan-doctrine? I'd finish it there first before doing other repos.

staabm:
> AFAIK the usage of build-infection isn't yet merged in phpstan-doctrine?

yes, its waiting in https://github.com/phpstan/phpstan-doctrine/pull/699 for the merge.

.github/workflows/tests.yml
```diff
@@ -222,3 +222,71 @@ jobs:
...
+          - "8.2"
+          - "8.3"
+          - "8.4"
+        operating-system: [ubuntu-latest]
+
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v5
+
+      - name: "Install PHP"
```
ondrejmirtes: Please use the composite action, same way it's used here (the commit first moves "Checkout build-infection" right after "Checkout", then we can use a local path to the composite action): https://github.com/phpstan/phpstan-doctrine/pull/704

.github/workflows/tests.yml
```diff
@@ -222,3 +222,71 @@ jobs:
 
       - name: "Tests"
         run: "make tests"
+
+  mutation-testing:
+    name: "Mutation Testing"
+    runs-on: "ubuntu-latest"
+    needs: ["tests", "tests-levels"]
```
ondrejmirtes: tests-levels? I don't want Infection to run these levels so it doesn't make sense to wait for them.

.github/workflows/tests.yml
```diff
@@ -222,3 +222,71 @@ jobs:
 
       - name: "Tests"
         run: "make tests"
+
+  mutation-testing:
```
ondrejmirtes, line 226: I want this to run only in pull requests. Can be achieved with `if: github.event_name == 'pull_request'`

ondrejmirtes (changes requested):

.github/workflows/tests.yml
ondrejmirtes: Also: what about static analysis? I know it's in another workflow but not warmed-up cache might have a downside or two here?

.github/workflows/tests.yml
```diff
@@ -222,3 +222,68 @@ jobs:
...
+        run: "composer install --no-interaction --no-progress"
+
+      - name: "Configure infection"
+        run: |
+          php build-infection/bin/infection-config.php \
+          > infection.json5
+          cat infection.json5 | jq
+
+      - name: "Cache Result cache"
+        uses: actions/cache@v4
```
ondrejmirtes: This isn't going to save the cache if the build fails, right? Shouldn't we do what's recommended here? https://phpstan.org/user-guide/result-cache#setup-in-github-actions

ondrejmirtes (changes requested):

.github/workflows/tests.yml
staabm: > tests-levels? I don't want Infection to run these levels so it doesn't make sense to wait for them.

I initially added levels to make sure we run late enough for the static analysis step... does no longer make sense as we no longer have the artifact down/upload step.

deleted it.

> Also: what about static analysis? I know it's in another workflow but not warmed-up cache might have a downside or two here?

as we restore the previous cache I think it will be "fresh enough". waiting across workflows will complicate it, as this is only supported in github actions with work-arrounds. I don't think its worth the effort.

infection will start and fail in the "preparation phase" when SA is not green.

ondrejmirtes: I thought we could just run PHPStan again before Infection.

staabm: infection will run it once before it starts it as a process killer

ondrejmirtes:
Thank you! I'm really interested what it will find here 😊

staabm:
```
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/
#StandWithUkraine
Infection - PHP Mutation Testing Framework version 0.31.7
Notice:  You are running Infection with PCOV enabled.
In Assert.php line 1013:
                                                                               
  Expected "PHPStan - PHP Static Analysis Tool dev-648dbd911cef28707338fe5c25  
  875d50e7875391@648dbd9                                                       
  " to be contain a valid SemVer (sub)string value. 
  ```
  
  I will check with the infection team in https://github.com/infection/infection/issues/2491 :)

staabm:
ok future PRs to phpstan-src will now have infection coverage, see a first e2e running build in e.g. https://github.com/phpstan/phpstan-src/actions/runs/18841268167/job/53754293666?pr=4470

//cc @maks-rafalko

maks-rafalko:
Cool! Thanks for working on this @staabm and @ondrejmirtes.

Looking forward to hear the results after some time. So please share them somewhere

PR #4481, opened 2025-10-25
Report unknown parameter for implicit variadic methods

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13719

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-13719.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-13719.php
@@ -1,5 +1,7 @@
 <?php declare(strict_types = 1);
 
+namespace Bug13719;
+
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Functions/data/bug-13719.php
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types = 1);
+
+function non_variadic(string $name, ?string $greeting = null): void {
```
ondrejmirtes, line 5: Please add namespace `Bug13719` to this file

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #4482, opened 2025-10-26
Use faster case-insensitve string comparison

staabm:
this `strtolower` call is showing up in blackfire profiles

grafik

because we invoke it millions of times

---

local benchmarks tells me `strcasecmp` is 2x faster 

```
fec) ✗ hyperfine 'php perf.php'
Benchmark 1: php perf.php
  Time (mean ± σ):     403.3 ms ±   8.6 ms    [User: 392.5 ms, System: 8.0 ms]
  Range (min … max):   385.6 ms … 414.7 ms    10 runs
 
➜  phpstan-src git:(infec) ✗ hyperfine 'php perf.php'
Benchmark 1: php perf.php
  Time (mean ± σ):     195.8 ms ±  10.1 ms    [User: 186.0 ms, System: 7.2 ms]
  Range (min … max):   183.3 ms … 219.4 ms    16 runs
```

running

```
<?php

function generateRandomString($length = 10) {
	return substr(str_shuffle(str_repeat($x='0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ', ceil($length/strlen($x)) )),1,$length);
}

$var1 = generateRandomString();
$var2 = generateRandomString();

$x = false;
for ($i = 0; $i < 10_000_000; $i++) {
	if (strcasecmp($var1, $var2) === 0) {
	//if (strtolower($var1) === strtolower($var2)) {
		$x = true;
	}
}
```

---

working my way thru things showing up in profiles of https://github.com/phpstan/build-infection/pull/11#issuecomment-3448262310

### DIFF
--- a/src/DependencyInjection/AutowiredAttributeServicesExtension.php
+++ b/src/DependencyInjection/AutowiredAttributeServicesExtension.php
@@ -18,7 +18,7 @@ use PHPStan\Rules\LazyRegistry;
 use ReflectionClass;
 use stdClass;
 use function explode;
-use function strtolower;
+use function strcasecmp;
 use function substr;
 
 final class AutowiredAttributeServicesExtension extends CompilerExtension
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

VincentLanglet:
There is ~70 occurences of `strtolower\(.*\) ===` ; should the same be done at other places @staabm ?

staabm:
Other places did not show up in my profilling yet

PR #4483, opened 2025-10-26
PhpDocStringResolver: Remove useless method call

staabm:
working my way thru things showing up in profiles of https://github.com/phpstan/build-infection/pull/11#issuecomment-3448262310


staabm:
Maybe we are just missing a unit test for this line 🧐

ondrejmirtes:
This is just the correct usage of the parser, it's present in both PhpDocStringResolver and TypeStringResolver.

I'm not really sure if we can write a failing test. But also I don't think this is a performance bottleneck.

ondrejmirtes:
Wasn't hard after all: https://github.com/phpstan/phpstan-src/commit/d393a1d7f99cb74c42fced0be94e5baf69c40d81

staabm:
Thx

PR #4484, opened 2025-10-26
Fix DOMElement::getAttributeNode and DOMElement::getAttributeNodeNS

VincentLanglet:
While looking at https://github.com/phpstan/phpstan-src/pull/3714
I discovered that getAttributeNode/getAttributeNodeNS return type was wrong

cf https://www.php.net/manual/en/domelement.getattributenode.php
and https://www.php.net/manual/en/domelement.getattributenodens.php

I fixed it in https://github.com/JetBrains/phpstorm-stubs/pull/1811
but then we will end with false positive since PHPStan does not understand `hasAttribute`,
so I added a benevolent union.

I also added the non-regression from https://github.com/phpstan/phpstan-src/pull/3714 to show the related-issues are solved

Closes https://github.com/phpstan/phpstan-src/pull/3714


ondrejmirtes:
Thank you!

PR #4489, opened 2025-10-27
Faster AutowiredAttributeServicesExtension->processParameters()

staabm:
even after https://github.com/phpstan/phpstan-src/pull/4482 we still see this method call on the hot path.

grafik

---

running PHPStan on the PHPStan-PHPUnit extension tests show this path as the slowest part of the run:

grafik

### DIFF
--- a/src/DependencyInjection/AutowiredAttributeServicesExtension.php
+++ b/src/DependencyInjection/AutowiredAttributeServicesExtension.php
@@ -19,6 +19,7 @@ use ReflectionClass;
 use stdClass;
 use function explode;
 use function strcasecmp;
+use function strtolower;
 use function substr;
 
 final class AutowiredAttributeServicesExtension extends CompilerExtension
@@ -44,8 +45,9 @@ final class AutowiredAttributeServicesExtension extends CompilerExtension
 			if (strcasecmp($parameter->method, '__construct') !== 0) {
 				continue;
 			}
-			$constructorParameters[$parameter->class] ??= [];
-			$constructorParameters[$parameter->class][] = $parameter;
+			$lowerClass = strtolower($parameter->class);
+			$constructorParameters[$lowerClass] ??= [];
+			$constructorParameters[$lowerClass][] = $parameter;
 		}
 
 		foreach (Attributes::findTargetClasses(AutowiredService::class) as $class) {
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/AutowiredAttributeServicesExtension.php
```diff
@@ -135,6 +135,10 @@ private function processParameters(string $className, ServiceDefinition $definit
 	{
 		$builder = $this->getContainerBuilder();
 		foreach ($autowiredParameters as $autowiredParameter) {
+			if ($autowiredParameter->method[0] !== '_') {
```
ondrejmirtes: I don't get it, what is that?

ondrejmirtes (changes requested):
The whole issue of something called a million times could be avoided by restructuring how vendor/attributes.php looks like, so that we don't have to iterate over so many things so many times.

src/DependencyInjection/AutowiredAttributeServicesExtension.php
staabm: it checks whether the 1st character in the string is a `_` (without the need for a function call)

staabm:
> The whole issue of something called a million times could be avoided by restructuring how vendor/attributes.php looks like

I don't want to customize the vendor/attributes.php too much, so we stay as long as possible compatible with the upstream.
(at least not for only 5% gains).

I restructured the loops and the bootleneck no longer exists

grafik

staabm:
before

```
➜  phpstan-phpunit git:(2.0.x) hyperfine vendor/bin/phpunit                                                                    
Benchmark 1: vendor/bin/phpunit
  Time (mean ± σ):      2.636 s ±  0.066 s    [User: 2.032 s, System: 0.183 s]
  Range (min … max):    2.556 s …  2.739 s    10 runs
 ```

with this PR
```
➜  phpstan-phpunit git:(2.0.x) hyperfine vendor/bin/phpunit
Benchmark 1: vendor/bin/phpunit
  Time (mean ± σ):      2.570 s ±  0.051 s    [User: 1.974 s, System: 0.180 s]
  Range (min … max):    2.451 s …  2.625 s    10 runs
```

File: src/DependencyInjection/AutowiredAttributeServicesExtension.php [hunk 2]
```diff
@@ -123,24 +131,18 @@ public function loadConfiguration(): void
...
 		}
 	}
 
 	/**
 	 * @param class-string $className
-	 * @param TargetMethodParameter<AutowiredParameter>[] $autowiredParameters
+	 * @param array<class-string, non-empty-list<TargetMethodParameter<AutowiredParameter>>> $constructorParameters
 	 */
-	private function processParameters(string $className, ServiceDefinition $definition, array $autowiredParameters): void
+	private function processParameters(string $className, ServiceDefinition $definition, array $constructorParameters): void
```
VincentLanglet: Should the method be renamed processConstructorParameters ?

ondrejmirtes:
Thank you!

PR #4492, opened 2025-10-27
Improve return type of base64_decode()

jlherren:
Fix handling of [non-bools as `$strict` parameter](https://phpstan.org/r/3fe3f6c0-70ae-450a-87e9-3615628e0a40) and omit `false` from return type when `$string` is valid base64.

This is useful when constant binary strings are required in code, e.g.:

    $emptyZipFile = base64_decode('UEsFBgAAAAAAAAAAAAAAAAAAAAAAAA==', true);

Related to phpstan/phpstan#7866

### DIFF
--- a/src/Type/Php/Base64DecodeDynamicFunctionReturnTypeExtension.php
+++ b/src/Type/Php/Base64DecodeDynamicFunctionReturnTypeExtension.php
@@ -6,13 +6,15 @@ use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Reflection\FunctionReflection;
+use PHPStan\TrinaryLogic;
 use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
-use PHPStan\Type\MixedType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\UnionType;
+use function base64_decode;
 
 #[AutowiredService]
 final class Base64DecodeDynamicFunctionReturnTypeExtension implements DynamicFunctionReturnTypeExtension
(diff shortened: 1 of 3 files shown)


src/Type/Php/Base64DecodeDynamicFunctionReturnTypeExtension.php
```diff
@@ -29,32 +31,41 @@ public function getTypeFromFunctionCall(
...
-		// second argument could be interpreted as true
-		if (!$isTrueType->no()) {
-			return new UnionType([new StringType(), new ConstantBooleanType(false)]);
+		if ($isStrict->yes() && $isValidBase64->no()) {
+			return new ConstantBooleanType(false);
 		}
-
-		return new StringType();
+		if ($isStrict->maybe() && $isValidBase64->maybe()) {
+			return new BenevolentUnionType([new StringType(), new ConstantBooleanType(false)]);
```
VincentLanglet, line 67: Why a benevolent one ? I think it should be a simple union

jlherren: I figured if the strictness is unclear (which almost never happens), then a benevolent union is better (based on the previous behavior). But of course this can be changed.

VincentLanglet: Oh, didn't know it was the previous behavior.

Indeed https://phpstan.org/r/54560d00-fc90-4e4f-bc72-74a670a8e415

ondrejmirtes:
This improves things a bit too much as it breaks analysis of PHPStan itself 😀 The base64 call inside the codebase meant to obfuscate a string so that we don't need to baseline related errors. But now we can un-obfuscate it and actually baseline the errors.

src/Type/Php/Base64DecodeDynamicFunctionReturnTypeExtension.php
```diff
@@ -30,32 +32,41 @@ public function getTypeFromFunctionCall(
...
-
-		if ($argType instanceof MixedType) {
-			return new BenevolentUnionType([new StringType(), new ConstantBooleanType(false)]);
+		$stringArgNode = $args[0]->value;
+		$constantStrings = $scope->getType($stringArgNode)->getConstantStrings();
+		if ($constantStrings !== []) {
+			$isValidBase64 = TrinaryLogic::lazyExtremeIdentity(
+				$constantStrings,
+				static function (ConstantStringType $constantString): TrinaryLogic {
+					$isValid = base64_decode($constantString->getValue(), true) !== false;
```
staabm, line 45: base64-decoding any string from within the codebase could skyrocket memory needed by phpstan analysis.

e.g. some codebases contain multi-megabyte base64 representations of images

ondrejmirtes: Yeah, I'm inclining not doing this at all.

jlherren: Would a length-limit help?

ondrejmirtes: I'm really not sure why would we want to decode literal strings passed to this method at all.

PR #4418, opened 2025-10-07
Fix filter var on uncertainty flags

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11485

PR #4495, opened 2025-10-30
[PHP 8.5] add support for FILTER_THROW_ON_FAILURE for filter_var

canvural:
Adds support for `FILTER_THROW_ON_FAILURE` for `filter_var`. There are still stuff I need to figure out:

- [ ] How to better test the throw type extension?
- [ ] Should we add the `@throws` tag to function in phpstorm-stubs repo? Or anywhere in PHPStan. Couldn't figure out how to conditionally add it only for PHP 8.5 in phpstorm-stubs repo.
- [ ] Combining `FILTER_THROW_ON_FAILURE` and `FILTER_NULL_ON_FAILURE` is not allowed. Should we have a new rule for that?

If anyone can help me with those questions, that'd be helpful!

### DIFF
--- a/src/Type/Php/FilterFunctionReturnTypeHelper.php
+++ b/src/Type/Php/FilterFunctionReturnTypeHelper.php
@@ -16,6 +16,7 @@ use PHPStan\Type\Constant\ConstantFloatType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\ConstantScalarType;
+use PHPStan\Type\ErrorType;
 use PHPStan\Type\FloatType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
> How to better test the throw type extension?

If the current test fails when the src/ changes are reverted then I'm fine with it 😊 Another approach would be to add a test to the "catch unthrown exception" rule.

> Should we add the @throws tag to function in phpstorm-stubs repo? Or anywhere in PHPStan. Couldn't figure out how to conditionally add it only for PHP 8.5 in phpstorm-stubs repo.

Thanks to the throw type extension, PHPStan doesn't need the PHPDoc. It might be a good idea to add it to phpstorm-stubs if you care about PhpStorm users separately 😊

> Combining FILTER_THROW_ON_FAILURE and FILTER_NULL_ON_FAILURE is not allowed. Should we have a new rule for that?

You can in a new PR if you'd want.

Thank you!

VincentLanglet:
I updated https://github.com/phpstan/phpstan-src/pull/4418 with this new flag.

PR #4499, opened 2025-10-30
add array_first and array_last return type extensions

canvural:
This PR adds return type extensions for `array_first` and `array_last`


src/Type/Php/ArrayFirstDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+use PHPStan\Type\TypeCombinator;
+use function count;
+
+#[AutowiredService]
+final class ArrayFirstDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
+{
+
+	public function isFunctionSupported(FunctionReflection $functionReflection): bool
+	{
+		return $functionReflection->getName() === 'array_first' && $functionReflection->isBuiltin();
```
canvural: Not sure if this is the correct way. Maybe we can check the PHP version. Because I thought there might be userland functions named `array_first` that would also be processed  here.

src/Type/Php/ArrayLastDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\NullType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use function count;
+
+#[AutowiredService]
+final class ArrayLastDynamicReturnTypeExtension implements DynamicFunctionReturnTypeExtension
```
staabm: as this 2 extensions are very similar, you might consider using a single extension class for handling both

src/Type/Php/ArrayFirstDynamicReturnTypeExtension.php
ondrejmirtes: I think it's okay like this, we don't check this anywhere for similar extensions. As a benefit it'd also work for polyfills running on lower PHP versions.

src/Type/Php/ArrayFirstDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		$argType = $scope->getType($args[0]->value);
+		$iterableAtLeastOnce = $argType->isIterableAtLeastOnce();
+
+		if ($iterableAtLeastOnce->no()) {
+			return new NullType();
+		}
+
+		$valueType = $argType->getFirstIterableValueType();
```
ondrejmirtes: When I thought about the implementation, I'd do only `getIterableValueType`. The `getFirstIterable*Type` methods are highly misleading and we should get rid of it. Because array shapes do not enforce the order of the keys. Sorry to put more burden on you, but can you please deprecate `Type::getFirstIterableKeyType`, `getLastIterableKeyType`, `getFirstIterableValueType`, `getLastIterableValueType` and use `getIterableKeyType` and `getIterableValueType` in all places instead? And just update any test assertions because of that. Thank you!

ondrejmirtes (changes requested):

src/Type/Php/ArrayFirstDynamicReturnTypeExtension.php
canvural: So should I remove the `$functionReflection->isBuiltin()` part?

src/Type/Php/ArrayFirstDynamicReturnTypeExtension.php
canvural: Alright. I can do that. But we will lose the precise type for constant arrays that way.

ondrejmirtes: Yeah, that's okay. Ideally we would have two separate implementations for literal arrays and PHPDoc array shapes but we don't.

src/Type/Php/ArrayFirstDynamicReturnTypeExtension.php
ondrejmirtes: Yes 😊

src/Type/Php/ArrayFirstDynamicReturnTypeExtension.php
canvural: Done with 73651b561e9bf8c8c06ea83025d7e03f92595d0a and 5f3237afa1b3eb2759b6a333dcd8c3a27bf32b8c

I didn't really check the assertions, just fixed whatever PHPStan said.

src/Type/Php/ArrayLastDynamicReturnTypeExtension.php
canvural: With the recent refactoring they are now merged into one

ondrejmirtes:
Awesome, looks like it fixed https://github.com/phpstan/phpstan/issues/8270. Please add a regression test.

canvural:
> Awesome, looks like it fixed [phpstan/phpstan#8270](https://github.com/phpstan/phpstan/issues/8270). Please add a regression test.

I'm not sure what got fixed. For PHP 7.3 - 8.2 there are no errors already. And for < 7.3 it doesn't make sense because `array_key_first` doesn't exist there.

staabm:
> I'm not sure what got fixed

see https://github.com/phpstan/phpstan-src/actions/runs/18969820143

looks like it fixed the "If condition is always true." error

canvural:
But that is already not reported https://phpstan.org/r/b63fa8a0-6ca6-4ac4-a8ca-b0a76cb22494 🤔

staabm:
I think it is: https://phpstan.org/r/84f65522-eace-4181-90c1-98c6a0f85206
(strict rules enabled)

canvural:
I don't understand it fully, but I did this 3530dd290af88b4df62d4f139f6ce8edaddbe679 Though it still fails locally 🤷🏽

ondrejmirtes:
Probably not fixed then and we're just misinterpreting the results somehow 😊

canvural:
Reverted then.

ondrejmirtes:
Thank you!

PR #4501, opened 2025-10-31
Fix 25x "Offset X might not exist on..." in FileTypeMapperTest

staabm:


ondrejmirtes:
This one I like, extra PHPUnit asserts are okay.

ondrejmirtes:
Thank you!

PR #4502, opened 2025-10-31
Fix "Offset X might not exist on..." in UnionTypeTest/IntersectionTypeTest

staabm:
fixes 10 errors

### DIFF
--- a/tests/PHPStan/Type/IntersectionTypeTest.php
+++ b/tests/PHPStan/Type/IntersectionTypeTest.php
@@ -5,7 +5,6 @@ namespace PHPStan\Type;
 use DoctrineIntersectionTypeIsSupertypeOf\Collection;
 use Iterator;
 use ObjectTypeEnums\FooEnum;
-use PHPStan\ShouldNotHappenException;
 use PHPStan\Testing\PHPStanTestCase;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
@@ -22,7 +21,6 @@ use PHPUnit\Framework\Attributes\DataProvider;
 use stdClass;
 use Test\ClassWithToString;
 use Traversable;
-use function array_key_exists;
 use function count;
 use function sprintf;
 use const PHP_VERSION_ID;
@@ -400,9 +398,6 @@ class IntersectionTypeTest extends PHPStanTestCase
 		$enumCases = $type->getEnumCases();
 		$this->assertCount(count($expectedEnumCases), $enumCases);
 		foreach ($enumCases as $i => $enumCase) {
-			if (!array_key_exists($i, $expectedEnumCases)) {
-				throw new ShouldNotHappenException();
-			}
 			$expectedEnumCase = $expectedEnumCases[$i];
 			$this->assertTrue($expectedEnumCase->equals($enumCase), sprintf('%s->equals(%s)', $expectedEnumCase->describe(VerbosityLevel::precise()), $enumCase->describe(VerbosityLevel::precise())));
 		}
--- a/tests/PHPStan/Type/IntersectionTypeTest.php
+++ b/tests/PHPStan/Type/IntersectionTypeTest.php
@@ -22,6 +22,7 @@ use PHPUnit\Framework\Attributes\DataProvider;
 use stdClass;
 use Test\ClassWithToString;
 use Traversable;
+use function array_key_exists;
 use function count;
 use function sprintf;
 use const PHP_VERSION_ID;
--- a/tests/PHPStan/Type/IntersectionTypeTest.php
+++ b/tests/PHPStan/Type/IntersectionTypeTest.php
@@ -5,6 +5,7 @@ namespace PHPStan\Type;
 use DoctrineIntersectionTypeIsSupertypeOf\Collection;
 use Iterator;
 use ObjectTypeEnums\FooEnum;
+use PHPStan\ShouldNotHappenException;
 use PHPStan\Testing\PHPStanTestCase;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Type/IntersectionTypeTest.php
```diff
@@ -398,6 +400,9 @@ public function testGetEnumCases(
 		$enumCases = $type->getEnumCases();
 		$this->assertCount(count($expectedEnumCases), $enumCases);
 		foreach ($enumCases as $i => $enumCase) {
+			if (!array_key_exists($i, $expectedEnumCases)) {
+				throw new ShouldNotHappenException();
```
ondrejmirtes: This error would go away after merging the `count === count` PR...

tests/PHPStan/Type/UnionTypeTest.php
```diff
@@ -180,6 +181,12 @@ public static function dataIsSuperTypeOf(): Iterator
 			new IntegerType(),
 			new StringType(),
 		]);
+		if (
+			!isset($unionTypeA->getTypes()[0])
+			|| !isset($unionTypeA->getTypes()[1])
```
ondrejmirtes: This doesn't get reported with `reportPossiblyNonexistentConstantArrayOffset: true`, right? I'm not sure why we'd need these conditions :)

ondrejmirtes (changes requested):

tests/PHPStan/Type/IntersectionTypeTest.php
staabm: Nope I tried that.

Non of the arrays involved is constant therefore we cannot infer the shape of the other one

ondrejmirtes: They are both lists. After doing `$this->assertCount(count($expectedEnumCases), $enumCases);` it's obvious that when `$enumCase[$i]` exists inside the foreach, `$expectedEnumCases[$i]` also must exist in the foreach.

We could carry an additional expression in Scope informing about this.

tests/PHPStan/Type/UnionTypeTest.php
staabm: see

```
 ------ ------------------------------------------------------- 
  Line   UnionTypeTest.php                                      
 ------ ------------------------------------------------------- 
  186    Offset 0 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:186            
  192    Offset 1 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:192            
  272    Offset 0 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:272            
  278    Offset 1 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:278            
  501    Offset 0 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:501            
  507    Offset 1 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:507            
  594    Offset 0 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:594            
  600    Offset 1 might not exist on array<PHPStan\Type\Type>.  
         🪪  offsetAccess.notFound                              
         at tests/PHPStan/Type/UnionTypeTest.php:600            
 ------ ------------------------------------------------------- 
 ```

tests/PHPStan/Type/IntersectionTypeTest.php
staabm: ok, sounds like a task for a different PR

tests/PHPStan/Type/UnionTypeTest.php
ondrejmirtes: When do you see this? What config change causes it?

staabm: `reportPossiblyNonexistentGeneralArrayOffset: true`

ondrejmirtes: Oh, that's not surprising. But I don't know if I ever want to enable that, because of needing to write code like that. It's not a very big difference whether we get "internal error" or "access on an undefined offset" notice when someone runs that code path. It's just that we need more code for the former.

How many errors are currently remaining when you enable that option?

ondrejmirtes: For similar reasons I don't plan to increase to level 9/10. Not enough benefits...

staabm: I am currently looking into how easily we can get the count down. 
I started with ~1500 and after open PRs are merged we are at ~1000

We might e.g. ignore this error in tests/ but not in src/ or similar

ondrejmirtes: The improvements are useful if they don't make the code worse but better.

Also some situations might inspire us to improve the type inference.

But to add a bunch of ShouldNotHappen throws just so we can enable the option isn't great.

Also in tests it's often easier to fix by adding an assert which is okay in tests ("we assume this") but not in src/ where the assumption might actually be wrong.

ondrejmirtes (commented):
I'd much rather refactor this by saving IntegerType into a variable and then reference it in both yields. Instead of these issets and throws.

staabm:
done

ondrejmirtes:
Thank you!

PR #4503, opened 2025-10-31
Create PhpParserExpr.stub

staabm:
fixes "~140 offset might not exist" errors

### DIFF
--- a/src/Reflection/ParametersAcceptorSelector.php
+++ b/src/Reflection/ParametersAcceptorSelector.php
@@ -473,7 +473,9 @@ final class ParametersAcceptorSelector
 		$singleParametersAcceptor = null;
 		if (count($parametersAcceptors) === 1) {
 			if (!array_is_list($args)) {
-				throw new ShouldNotHappenException('args is expected to be a list');
+				// actually $args parameter should be typed to list but we can't atm,
+				// because its a BC break.
+				$args = array_values($args);
 			}
 			$reorderedArgs = ArgumentsNormalizer::reorderArgs($parametersAcceptors[0], $args);
 			$singleParametersAcceptor = $parametersAcceptors[0];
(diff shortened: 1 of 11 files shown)


src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -471,6 +472,9 @@ public static function selectFromArgs(
 		$parameters = null;
 		$singleParametersAcceptor = null;
 		if (count($parametersAcceptors) === 1) {
+			if (!array_is_list($args)) {
```
staabm, line 475: doing a runtime check, to not change the phpdoc on a `@api` class

ondrejmirtes: Well this is worse because suddenly you're throwing an exception for a valid input advertised by the signature.

If it's not a list, let's make it a list.

staabm: Great idea. 


It might even be worth creating a new rule which errors in my wrong code example?

ondrejmirtes: Not sure how to express the logic you'd want...

staabm: Error on conditions which always lead to unchecked exceptions, when the conditions narrows a parameter type.

But you are right, maybe too specific

ondrejmirtes: I'd merge this if you call array_values here instead 😊

staabm: was not at the computer until now. fix incoming ;).

ondrejmirtes: I'm also outside the entire time 😂

ondrejmirtes:
Thank you!

PR #4504, opened 2025-10-31
Utilize array_last()

staabm:
more readable code by using `array_last`. also adds symfony/polyfills required for release builds which support PHP older than 8.5

---

fixes ~9x "offset might not exist" errors on phpstan-src with `reportPossiblyNonexistentGeneralArrayOffset: true`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -109,7 +109,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "8.3"
+          php-version: "8.5"
 
       - uses: "ramsey/composer-install@v3"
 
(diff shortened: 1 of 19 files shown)


ondrejmirtes:
Please also add php84 polyfill to all the places. And once it's merged the bootstrap.php also needs to be modified in phpstan/phpstan.

ondrejmirtes:
The polyfills also need to be mentioned at the bottom of scoper.inc.php (that's what I screwed up last time and shipped it, lead to crashing in some projects).

staabm:
something is fishy with shipmonk/composer-dependency-analyser

ondrejmirtes:
Maybe needs to run on 8.5?

staabm:
seems this did the trick. thank you.

ondrejmirtes:
Thank you!

PR #4505, opened 2025-10-31
make BcMath\Number::$value numeric-string

schlndh:
Fixes https://phpstan.org/r/da494451-a17d-49f9-bcb7-5b2ce11219d2

It's covered by `ReflectionProviderGoldenTest` so I didn't add any other test:

```
1) PHPStan\Reflection\ReflectionProviderGoldenTest::test with data set "PROPERTY BcMath\Number::$value" ('PROPERTY BcMath\Number::$value', 'Readonly\nVisibility: public\...string')
--- Expected
+++ Actual
@@ @@
 'Readonly
 Visibility: public
-Read type: string
-Write type: string'
+Read type: numeric-string
+Write type: numeric-string'
```

### DIFF
--- a/src/PhpDoc/BcMathNumberStubFilesExtension.php
+++ b/src/PhpDoc/BcMathNumberStubFilesExtension.php
@@ -1,0 +1,25 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDoc;
+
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #4506, opened 2025-10-31
Fix "offset might not exist" errors

staabm:


ondrejmirtes:
Conflict here. Do we still need this?

PR #4507, opened 2025-10-31
Improve count on list with greater/smaller-than

staabm:
closes https://github.com/phpstan/phpstan/issues/13747

---

fixes ~55x "offset might not exist" errors on phpstan-src with `reportPossiblyNonexistentGeneralArrayOffset: true`

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-13747.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-13747.php
@@ -1,0 +1,99 @@
+<?php declare(strict_types = 1);
+
+namespace Bug13747;
+
+use function PHPStan\Testing\assertType;
+
+class HelloWorld
+{
+	/** @param list<int> $list */
+	public function count($list): void
+	{
+		if (count($list) === 0) {
+			return;
+		}
+
+		if (count($list) > 2) {
+			assertType('false', array_key_exists(-1, $list));
+			assertType('true', array_key_exists(0, $list));
+			assertType('true', array_key_exists(1, $list));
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/nsrt/bug-13747.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	/** @param list<int> $list */
+	public function doFoo($list): void
+	{
+		if (count($list) === 0) {
+			return;
+		}
+
+		if (count($list) >= 2) {
```
staabm, line 53: Needs another test for a max number of offsets (limit)

staabm: here we go

src/Analyser/LegacyTypeSpecifier.php
```diff
@@ -1189,7 +1192,27 @@ private function specifyTypesForCountFuncCall(
...
+
+						if ($i > self::MAX_ACCESSORIES_LIMIT) {
+							break;
+						}
+
+						$intersection[] = new HasOffsetValueType($offsetType, $valueType);
+						$i++;
+					}
+
+					$resultTypes[] = TypeCombinator::intersect(...$intersection);
```
staabm: could also use a AccessoryArraySize, but this likely would be more involved and require a bigger refactor.

therefore I went with this simple approach which covers most useful use-cases

VincentLanglet (approved):
There is some escaped mutant though.

PR #4515, opened 2025-11-02
`count(non-empty-array, COUNT_RECURSIVE)` is `int<1, max>`

staabm:
Before this PR it was taken as `int<0, max>`

### DIFF
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -686,10 +686,7 @@ class IntersectionType implements CompoundType
 		if ($arraySize instanceof IntegerRangeType) {
 			$knownOffsets = [];
 			foreach ($this->types as $type) {
-				if ($type instanceof HasOffsetValueType) {
-					$knownOffsets[$type->getOffsetType()->getValue()] = true;
-				}
-				if (!($type instanceof HasOffsetType)) {
+				if (!($type instanceof HasOffsetValueType) && !($type instanceof HasOffsetType)) {
 					continue;
 				}
 
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -53,7 +53,6 @@ use function implode;
 use function in_array;
 use function is_int;
 use function ksort;
-use function max;
 use function md5;
 use function sprintf;
 use function strcasecmp;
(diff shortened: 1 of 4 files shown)


src/Type/Php/CountFunctionReturnTypeExtension.php
```diff
@@ -28,18 +30,31 @@ public function getTypeFromFunctionCall(
...
+		$arrayType = $scope->getType($args[0]->value);
+		if (!$this->isNormalCount($functionCall, $arrayType, $scope)->yes()) {
+			if ($arrayType->isIterableAtLeastOnce()->yes()) {
+				return IntegerRangeType::fromInterval(1, null);
 			}
+			return null;
 		}
 
-		return $scope->getType($functionCall->getArgs()[0]->value)->getArraySize();
+		return $scope->getType($args[0]->value)->getArraySize();
```
staabm, line 46: maybe put the COUNT_MODE onto `Type->getArraySize()` as a first parameter

ondrejmirtes (changes requested):
What about the mutants? 😊

staabm:
I cannot think of a test I am missing and which could kill the mutant ~~therefore I simplified the implementation~~

src/Type/IntersectionType.php
```diff
@@ -680,7 +681,22 @@ public function isIterableAtLeastOnce(): TrinaryLogic
 
 	public function getArraySize(): Type
 	{
-		return $this->intersectTypes(static fn (Type $type): Type => $type->getArraySize());
+		$arraySize = $this->intersectTypes(static fn (Type $type): Type => $type->getArraySize());
+
+		if ($arraySize instanceof IntegerRangeType) {
```
ondrejmirtes: What about unions of integer ranges?

ondrejmirtes (changes requested):
And escaped mutants :)

src/Type/IntersectionType.php
staabm: rewrote it to get rid of the instanceof check.

staabm:
@ondrejmirtes mutants killed.

ondrejmirtes:
Thank you!

PR #4516, opened 2025-11-02
infinite loop in never returning function is allowed

canvural:
Doing an infinite loop in a function that never returns is a valid use case.

closes https://github.com/phpstan/phpstan/issues/6458
closes https://github.com/phpstan/phpstan/issues/12813


staabm:
see https://github.com/phpstan/phpstan-src/actions/runs/19016009464 how this change affects other open bug reports

ondrejmirtes:
Thank you!

PR #4519, opened 2025-11-03
Add support for union of constant integer for filter_var

VincentLanglet:
Follow https://github.com/phpstan/phpstan-src/pull/4418


ondrejmirtes:
Thank you!

PR #4520, opened 2025-11-03
Preserve constant array when setting a union of constant scalar keys

ondrejmirtes:


ondrejmirtes:
/cc @VincentLanglet This change solves a few issues, including your own: https://github.com/phpstan/phpstan-src/actions/runs/19044800760

Unfortunately it introduces also some problems which I can't dig into right now. Could you please debug why some existing tests are failing and some errors reappear?

I think some of it might be related to https://github.com/phpstan/phpstan-src/pull/4372 but not sure.

ondrejmirtes:
Also /cc @staabm because you might have touched this more recently 😊

src/Type/Constant/ConstantArrayType.php
```diff
@@ -693,6 +693,20 @@ public function getOffsetValueType(Type $offsetType): Type
...
+				$arrays = [];
+				foreach ($constantScalars as $constantScalar) {
+					$builder = ConstantArrayTypeBuilder::createFromConstantArray($this);
+					$builder->setOffsetValueType($constantScalar, $valueType);
+					$arrays[] = $builder->getArray();
+				}
+
+				return TypeCombinator::union($this, ...$arrays);
+			}
+		}
```
staabm, line 709: I wonder, whether this code should be moved into the impl of `ConstantArrayTypeBuilder->setOffsetValueType()`?

ondrejmirtes: It should be handled there but this code should not be copied, but adapted (it's not so easy to do it there). I just wanted to try a quick PoC first.

src/Type/Constant/ConstantArrayType.php
```diff
@@ -693,6 +693,20 @@ public function getOffsetValueType(Type $offsetType): Type
 
 	public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $unionValues = true): Type
```
staabm, line 694: looking at this `unionValues` parameter, makes me feel we nearly everytime ignore it.

I wonder whether this bool represents the same 2 use-cases as `setOffsetValueType()` vs. `setExistingOffsetValueType()`.

like `unionValues=true` is the same as `setOffsetValueType()`
and `unionValues=false` is the same as `setExistingOffsetValueType()`

ondrejmirtes: The unionValues parameter is some hack I came up with, there's no deeper thought in it and it's likely the proper solution is different.

src/Type/Constant/ConstantArrayType.php
```diff
@@ -693,6 +693,20 @@ public function getOffsetValueType(Type $offsetType): Type
...
+			$constantScalarsCount = count($constantScalars);
+			if ($constantScalarsCount > 1 && $constantScalarsCount < ConstantArrayTypeBuilder::ARRAY_COUNT_LIMIT) {
+				$arrays = [];
+				foreach ($constantScalars as $constantScalar) {
+					$builder = ConstantArrayTypeBuilder::createFromConstantArray($this);
+					$builder->setOffsetValueType($constantScalar, $valueType);
+					$arrays[] = $builder->getArray();
+				}
+
+				return TypeCombinator::union($this, ...$arrays);
```
VincentLanglet: I might be tired but why 
```
				return TypeCombinator::union($this, ...$arrays);
```
and not
```suggestion
				return TypeCombinator::union(...$arrays);
```
?

src/Type/Constant/ConstantArrayType.php
```diff
@@ -693,6 +693,20 @@ public function getOffsetValueType(Type $offsetType): Type
 
 	public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $unionValues = true): Type
 	{
+		if ($offsetType !== null) {
+			$constantScalars = $offsetType->getConstantScalarTypes();
+			$constantScalarsCount = count($constantScalars);
+			if ($constantScalarsCount > 1 && $constantScalarsCount < ConstantArrayTypeBuilder::ARRAY_COUNT_LIMIT) {
```
staabm, line 699: Do we need similar handling for integer ranges with less than array-limit elements?

src/Type/Constant/ConstantArrayType.php
VincentLanglet: > The unionValues parameter is some hack I came up with,

What is the purpose of this hack ?

staabm: I did some research yesterday. it was introduced with https://github.com/phpstan/phpstan-src/commit/5d644831177353d9d44785d461da452f67e58f48

src/Type/Constant/ConstantArrayType.php
ondrejmirtes: I was also tired, it makes sense without the `$this` 😊 Updated that.

src/Type/Constant/ConstantArrayType.php
ondrejmirtes: Yes, that would make sense!

VincentLanglet:
Looking at failure
```
1) PHPStan\Rules\Classes\InstantiationRuleTest::testBug7594
--- Expected
+++ Actual
@@ @@
-'
+'55: Parameter #1 $boolAbilities of class Bug7594\HelloWorld constructor expects array<0|1|2|3|4|5|6|7|8|9|10|11|12|13|14|15|16|17, bool>, array<int<0, max>, bool> given.
 '
```
=> Related to for loop, which generalize the result computed after the second loop.
https://github.com/phpstan/phpstan-src/blob/770b08c052d846ac6b78e947d24f890d5b415915/src/Analyser/NodeScopeResolver.php#L1239-L1241

And looking at all the other issues, I think it's also the fact that for loop generalize result creating less precise ones, except:
```
4) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "tests\PHPStan\Analyser\nsrt\constant-array-type-set.php" ('D:\a\phpstan-src\phpstan-src\...et.php')
Failed assertions in D:\a\phpstan-src\phpstan-src\tests\PHPStan\Analyser\nsrt\constant-array-type-set.php:

Line 24:
Expected: array{bool, bool, bool}
Actual:   array{false, false, true}|array{false, true, false}|array{true, false, false}

Line 36:
Expected: array{0: bool, 1: bool, 2: bool, 3?: true}
Actual:   array{false, false, false, true}|array{false, false, true}|array{false, true, false}|array{true, false, false}

Line 42:
Expected: array{bool, bool, false}
Actual:   array{false, true, false}|array{true, false, false}

Line 104:
Expected: array{bool, bool, false}
Actual:   array{false, true, false}|array{true, false, false}
```
Seems expected to me

staabm:
pick this commit and you get a green build:
https://github.com/staabm/phpstan-src/commit/333d1c4115c528eb41f563314385e606355cba5b

problem is/was: the PR without this fix created unions with lots of elements in the ConstantArray types. this got either turned into OversizedArrayType or generalized while analyzing loops.

staabm:
And pick https://github.com/staabm/phpstan-src/commit/4cbca5ecf291d9de74a715757e560fa05cdf23a8 on top to get `IntegerRangeType` support

staabm:
I decided to [send a new PR](https://github.com/phpstan/phpstan-src/pull/4528), so I can see impact on issue-bot

PR #4527, opened 2025-11-05
Added #[Covers*] attributes to tests

staabm:
goal: speedup the coverage generation process

refs https://github.com/paratestphp/paratest/issues/1044

before this PR
```
XDEBUG_MODE=coverage php tests/vendor/bin/paratest --coverage-xml=tmp/coverage-xml 
...
Time: 07:41.109, Memory: 618.50 MB

Tests: 13325, Assertions: 80593, Skipped: 170.

Generating code coverage report in PHPUnit XML format ... done [00:34.753]
```

after this PR
```
XDEBUG_MODE=coverage php tests/vendor/bin/paratest --coverage-xml=tmp/coverage-xml 
...
Time: 03:37.977, Memory: 668.50 MB

Tests: 13325, Assertions: 80593, Skipped: 170.

Generating code coverage report in PHPUnit XML format ... done [00:32.811]
```

Next: figure out whether we can utilize it and which annotations make sense where

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -11,11 +11,13 @@ use PHPStan\Reflection\SignatureMap\SignatureMapProvider;
 use PHPStan\Testing\PHPStanTestCase;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
+use PHPUnit\Framework\Attributes\CoversNothing;
 use PHPUnit\Framework\Attributes\RequiresPhp;
 use function extension_loaded;
 use function sprintf;
 use const PHP_VERSION_ID;
 
+#[CoversNothing]
 class AnalyserIntegrationTest extends PHPStanTestCase
 {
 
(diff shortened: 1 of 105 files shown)


ondrejmirtes:
A clean way to get rid of static analysis errors would be for simple-downgrader to downgrade the attribute to `@coversNothing`.

staabm:
I still need to figure out a good way forward for this Covers* annotations.

If we use CoversNothing it will make tests beeing skipped for infection.
Only thing which makes sense to me is, that we look for real unit tests and mark them with CoversClass(X).
that way we at least reduce the number of test-classes to make coverage generation a bit faster

(I am still learning this Covers* annotations while beeing in constant conversations with Sebastian and the infection team)

ondrejmirtes:
There are zero real unit tests 😊 I feel like CoversNothing for a few specific classes like AnalyseeIntegrarionTest is the way to go.

There are some classes in this PR that I disagree with but generally I want CoversNothing here 😊

staabm:
maybe it makes sense you look for the test which fit for CoversNothing?
I don't see others besides `AnalyseeIntegrarionTest`. I would just start guessing at this point.

staabm:
closing this, as "POC completed".

Adding the attributes in https://github.com/phpstan/phpstan-src/pull/4535

PR #4528, opened 2025-11-06
Preserve constant array when setting a union of constant scalar keys

staabm:
closes https://github.com/phpstan/phpstan-src/pull/4520

PR #4529, opened 2025-11-06
Simplify ConstantArrayTypeBuilder

staabm:
identified while looking into https://github.com/phpstan/phpstan-src/pull/4528


ondrejmirtes:
Thank you!

PR #4530, opened 2025-11-06
add Type::isError method

canvural:
This one does not add too much value, but I just didn't want to do `instanceof ErrorType` in my extensions 😄

### DIFF
--- a/src/Rules/PhpDoc/UnresolvableTypeHelper.php
+++ b/src/Rules/PhpDoc/UnresolvableTypeHelper.php
@@ -3,7 +3,6 @@
 namespace PHPStan\Rules\PhpDoc;
 
 use PHPStan\DependencyInjection\AutowiredService;
-use PHPStan\Type\ErrorType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeTraverser;
@@ -16,7 +15,7 @@ final class UnresolvableTypeHelper
 	{
 		$containsUnresolvable = false;
 		TypeTraverser::map($type, static function (Type $type, callable $traverse) use (&$containsUnresolvable): Type {
-			if ($type instanceof ErrorType) {
+			if ($type->isError()->yes()) {
 				$containsUnresolvable = true;
 			}
 			if ($type instanceof NeverType && !$type->isExplicit()) {
--- a/src/Type/Generic/TemplateTypeTrait.php
+++ b/src/Type/Generic/TemplateTypeTrait.php
@@ -6,7 +6,6 @@ use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\TypeNode;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\AcceptsResult;
-use PHPStan\Type\ErrorType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\IsSuperTypeOfResult;
 use PHPStan\Type\MixedType;
@@ -18,6 +17,7 @@ use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeUtils;
 use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
+use function is_string;
 use function sprintf;
 
 /**
@@ -74,7 +74,7 @@ trait TemplateTypeTrait
 			$defaultDescription = '';
 			if ($this->default !== null) {
 				$recursionGuard = RecursionGuard::runOnObjectIdentity($this->default, fn () => $this->default->describe($level));
-				if (!$recursionGuard instanceof ErrorType) {
+				if (is_string($recursionGuard)) {
 					$defaultDescription .= sprintf(' = %s', $recursionGuard);
 				}
 			}
--- a/src/Type/MixedType.php
+++ b/src/Type/MixedType.php
@@ -1073,13 +1073,7 @@ class MixedType implements CompoundType, SubtractableType
 
 	public function isError(): TrinaryLogic
 	{
-		if ($this->subtractedType !== null) {
-			if ($this->subtractedType->isSuperTypeOf(new ErrorType())->yes()) {
-				return TrinaryLogic::createNo();
-			}
-		}
-
-		return TrinaryLogic::createMaybe();
+		return TrinaryLogic::createNo();
 	}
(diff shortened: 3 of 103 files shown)


ondrejmirtes (changes requested):
ApiInstanceofTypeRule should be made aware of this 😊

ondrejmirtes:
Do you also feel like doing `isNever()` + `isExplicitNever()`? :)

canvural:
> Do you also feel like doing `isNever()` + `isExplicitNever()`? :)

Didn't ever need that I guess, but sure I can. In a separate PR?

canvural:
The mutations look valid, but I don't know how I can change them in bulk.

Edit: I guess `$type instanceof ErrorType` could be `!$type->isError()->no()` and `!$type instanceof ErrorType` can be `$type->isError()->no()` ?

ondrejmirtes:
I'm not sure how to cover the mutations with tests here. i'm more concerned about the things the issue bot found https://github.com/phpstan/phpstan-src/actions/runs/19215786400, I'm not sure why that's happening.

src/Type/Generic/TemplateTypeTrait.php
```diff
@@ -74,7 +74,7 @@ public function describe(VerbosityLevel $level): string
 			$defaultDescription = '';
 			if ($this->default !== null) {
 				$recursionGuard = RecursionGuard::runOnObjectIdentity($this->default, fn () => $this->default->describe($level));
-				if (!$recursionGuard instanceof ErrorType) {
+				if (is_string($recursionGuard)) {
```
staabm, line 77: this looks fishy

src/Type/ObjectType.php
```diff
@@ -156,7 +156,7 @@ public function hasProperty(string $propertyName): TrinaryLogic
 		}
 
 		$classHasProperty = RecursionGuard::run($this, static fn (): bool => $classReflection->hasProperty($propertyName));
-		if ($classHasProperty === true || $classHasProperty instanceof ErrorType) {
+		if ($classHasProperty !== false) {
```
staabm, line 163: are we sure this does not change existing logic?

ondrejmirtes: Yeah, definitely weird

src/Type/Generic/TemplateTypeTrait.php
ondrejmirtes: Yeah, definitely weird

staabm:
just a gut feeling in case the 2 above comments regarding behaviour changes are not valid: 

another source of problems could be 
- `class ErrorType extends MixedType`
- `class TemplateMixedType extends MixedType`

because `MixedType` - because of substraction logic - might lead to a `Maybe`.

src/Type/ObjectType.php
canvural: Yeah, doesn't change it. It can only be bool or ErrorType.

src/Type/Generic/TemplateTypeTrait.php
canvural: Why? It can only be `string` or `ErrorType` here. So I thought `is_string` is ok instead of `instanceof`

src/Rules/PhpDoc/UnresolvableTypeHelper.php
```diff
@@ -16,7 +15,7 @@ public function containsUnresolvableType(Type $type): bool
 	{
 		$containsUnresolvable = false;
 		TypeTraverser::map($type, static function (Type $type, callable $traverse) use (&$containsUnresolvable): Type {
-			if ($type instanceof ErrorType) {
+			if ($type->isError()->yes()) {
```
canvural, line 18: @staabm @ondrejmirtes 

I debugged a bit and I think the issue bot errors might be valid. It's always an issue with `ValueOfType` and it comes down to this line. When we do `$type->isError()` here `ValueOfType` resolves the underlying value and it resolves to `ErrorType`. We do not get the issue with `instanceof` check because that's always `ValueOfType vs ErrorType`

Does it make sense?

src/Type/MixedType.php
```diff
@@ -1065,6 +1065,17 @@ public function isScalar(): TrinaryLogic
 		return TrinaryLogic::createMaybe();
 	}
 
+	public function isError(): TrinaryLogic
```
VincentLanglet, line 1074: Isn't it just simpler to always return no ?

Cause currently `MixedType instanceof ErrorType` is just false.

And anyway, I don't think there is a `isError()->maybe()` used somewhere...

ErrorType extends MixedType for simplicity, but a "real" MixedType is never an error.
And Mixed~Error is not a Type I often saw...

src/Type/ObjectType.php
VincentLanglet, line 163: Might be easier to understand
```
if ($classHasProperty === true || $classHasProperty->isError()->yes())
```

src/Rules/PhpDoc/UnresolvableTypeHelper.php
VincentLanglet: https://github.com/phpstan/phpstan-src/pull/4548 Might help on this.

VincentLanglet (commented):
In https://github.com/phpstan/phpstan-src/pull/4547
```
$expressionType instanceof NeverType
```
by
```
!$expressionType->isNever()->no()
```

But here you're replacing
```
$type instanceof ErrorType
```
by
```
$type->isError()->yes()
```

How did you choose ? I feel like we should have consistency...

src/Type/MixedType.php
canvural: Yeah makes sense. Did it with 6370bbcfc21204df94554c60b9b182df4d25ea52 Thank you

canvural:
> In #4547
> 
> ```
> $expressionType instanceof NeverType
> ```
> 
> by
> 
> ```
> !$expressionType->isNever()->no()
> ```
> 
> But here you're replacing
> 
> ```
> $type instanceof ErrorType
> ```
> 
> by
> 
> ```
> $type->isError()->yes()
> ```
> 
> How did you choose ? I feel like we should have consistency...

Yeah we should have consistency. In the `isNever` PR I did it that way to see how the mutation testing will behave. But honestly I'd rather have `$type->isError()->yes()` everywhere, it's much easier to understand. So I'll change the other PR later.

staabm:
I am neutral on this PR and would leave the final decision to @ondrejmirtes

PR #4531, opened 2025-11-06
Prevent duplicate work

staabm:


src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -154,7 +154,7 @@ public function findSpecifiedType(
 							foreach ($haystackArrayTypes as $haystackArrayType) {
 								if ($haystackArrayType instanceof ConstantArrayType) {
 									foreach ($haystackArrayType->getValueTypes() as $i => $haystackArrayValueType) {
-										if (count($haystackArrayValueType->getFiniteTypes()) > 1 || $haystackArrayType->isOptionalKey($i)) {
```
staabm, line 157: this line was added with https://github.com/phpstan/phpstan-src/pull/3888 but deleting this line does not make any test fail..

ondrejmirtes:
Thank you!

PR #4534, opened 2025-11-07
Understand always-overwritten arrays in foreach

ondrejmirtes:
I don't understand why I didn't do this years ago 😂 

Closes https://github.com/phpstan/phpstan/issues/2273
Closes https://github.com/phpstan/phpstan/issues/13730

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -91,6 +91,7 @@ use PHPStan\Node\Expr\GetIterableKeyTypeExpr;
 use PHPStan\Node\Expr\GetIterableValueTypeExpr;
 use PHPStan\Node\Expr\GetOffsetValueTypeExpr;
 use PHPStan\Node\Expr\NativeTypeExpr;
+use PHPStan\Node\Expr\OriginalForeachKeyExpr;
 use PHPStan\Node\Expr\OriginalPropertyTypeExpr;
 use PHPStan\Node\Expr\PropertyInitializationExpr;
 use PHPStan\Node\Expr\SetExistingOffsetValueTypeExpr;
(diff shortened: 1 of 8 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -1247,14 +1248,94 @@ private function processStmtNode(
...
 
 			$exprType = $scope->getType($stmt->expr);
+			$hasExpr = $scope->hasExpressionType($stmt->expr);
+			if (
+				count($breakExitPoints) === 0
+				&& count($scopesWithIterableValueType) > 0
+				&& !$continueExitPointHasUnoriginalKeyType
+				&& $stmt->keyVar !== null
+				&& $exprType->isArray()->yes()
+				&& $exprType->isConstantArray()->no()
```
staabm, line 1286: tbh I did not yet understand why this PR works, but I wonder whether we could use the same logic to also infer a better result when the iteratee is a known constant array?

like in 
- https://github.com/phpstan/phpstan/issues/13000
- https://github.com/phpstan/phpstan/issues/2294

ondrejmirtes: Why this works - I look at `$a[$k]` type at the end of the loop and at each `continue` point and if some other conditions are true (like there are zero `break` points), I rewrite the whole array value type with the `$a[$k]` type unioned from all the points I looked at.

ondrejmirtes: When `$foreach->expr` is a constant array I feel like we need a different approach, possibly analysing the foreach separately for each item (but run `$nodeCallback` for the rules still only once).

ondrejmirtes: Also I want to look at `foreach ($a as &$v)` to understand it better that also needs a different approach.

PR #4535, opened 2025-11-08
Add `#[CoversNothing]`

staabm:
In https://github.com/phpstan/phpstan-src/pull/4527 we found out that `#[CoversNothing]` can speedup the coverage generation process.

We can only add this attributes to tests which are not relevant for mutation testing, because infection would not see these tests when they are no longer part of the coverage report.

---

tbh: I cannot believe the results running
(its reproducible though)

```
XDEBUG_MODE=coverage php tests/vendor/bin/paratest \
          --coverage-xml=tmp/coverage/coverage-xml \
          --log-junit=tmp/coverage/junit.xml

ParaTest v7.8.4 upon PHPUnit 11.5.43 by Sebastian Bergmann and contributors.

Warming cache for static analysis ... [00:00.051]
Processes:     14
Runtime:       PHP 8.3.27 with Xdebug 3.4.7
Configuration: /Users/staabm/workspace/phpstan-src/phpunit.xml
Random Seed:   1762589403

...................SS.....................S...S...........S    59 / 13334 (  0%)
...........................................................   118 / 13334 (  0%)
..
```

before this PR:
```
Time: 07:57.258, Memory: 618.50 MB

Tests: 13334, Assertions: 80601, Skipped: 171.

Generating code coverage report in PHPUnit XML format ... done [00:34.829]
```

after this PR:
```
Time: 03:44.550, Memory: 660.50 MB

Tests: 13334, Assertions: 80601, Skipped: 171.

Generating code coverage report in PHPUnit XML format ... done [00:32.989]
```

on macbook m4 pro

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -11,11 +11,13 @@ use PHPStan\Reflection\SignatureMap\SignatureMapProvider;
 use PHPStan\Testing\PHPStanTestCase;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
+use PHPUnit\Framework\Attributes\CoversNothing;
 use PHPUnit\Framework\Attributes\RequiresPhp;
 use function extension_loaded;
 use function sprintf;
 use const PHP_VERSION_ID;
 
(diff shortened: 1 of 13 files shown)


staabm:
In CI the difference is only 20 seconds 🤷‍♂️ .. but still an improvement

ondrejmirtes:
Thank you!

PR #4536, opened 2025-11-08
Infer non-empty-array after array_key_first/last()

staabm:
closes https://github.com/phpstan/phpstan/issues/13546

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -2383,6 +2383,21 @@ final class TypeSpecifier
 			}
 		}
 
+		// array_key_first($a) !== null
+		// array_key_last($a) !== null
+		if (
+			$unwrappedLeftExpr instanceof FuncCall
+			&& $unwrappedLeftExpr->name instanceof Name
+			&& in_array($unwrappedLeftExpr->name->toLowerString(), ['array_key_first', 'array_key_last'], true)
+			&& isset($unwrappedLeftExpr->getArgs()[0])
+			&& $rightType->isNull()->yes()
+		) {
+			$argType = $scope->getType($unwrappedLeftExpr->getArgs()[0]->value);
+			if ($argType->isArray()->yes()) {
+				return $this->create($unwrappedLeftExpr->getArgs()[0]->value, new NonEmptyArrayType(), $context->negate(), $scope)->setRootExpr($expr);
+			}
+		}
+
 		// preg_match($a) === $b
 		if (
 			$context->true()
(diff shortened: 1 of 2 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -2346,6 +2346,21 @@ public function resolveIdentical(Expr\BinaryOp\Identical $expr, Scope $scope, Ty
 			}
 		}
 
+		// array_key_first($a) !== null
```
SplotyCode, line 2386: does this also work for? https://phpstan.org/r/3459cc41-b3a9-4f63-9e69-23b7ed76cf3d

staabm: nope, this needs to be handled separately and I guess can be done in a more generic way

src/Analyser/TypeSpecifier.php
```diff
@@ -2383,6 +2383,21 @@ private function resolveNormalizedIdentical(Expr\BinaryOp\Identical $expr, Scope
...
```
VincentLanglet, line 2386: I thought you forgot `null !== array_key_first($a)` but seems like it works looking at the test.
I'm not familiar with this code to know why at first sight.

src/Analyser/TypeSpecifier.php
```diff
@@ -2383,6 +2383,21 @@ private function resolveNormalizedIdentical(Expr\BinaryOp\Identical $expr, Scope
...
+		// array_key_last($a) !== null
+		if (
+			$unwrappedLeftExpr instanceof FuncCall
+			&& $unwrappedLeftExpr->name instanceof Name
+			&& in_array($unwrappedLeftExpr->name->toLowerString(), ['array_key_first', 'array_key_last'], true)
+			&& isset($unwrappedLeftExpr->getArgs()[0])
+			&& $rightType->isNull()->yes()
```
VincentLanglet, line 2393: You're getting escaped mutant here, but if I understand correctly, it should be supported by
```
if (array_key_first($array) !== $nullOrInt) {
		assertType('list<string>', $array);
	}
```

VincentLanglet (approved):

src/Analyser/TypeSpecifier.php
staabm: we are in `resolveNormalizedIdentical`, which is invoked in both directions here:

https://github.com/phpstan/phpstan-src/blob/eb4e247282a61eaa0b15f8139264e09c183622d8/src/Analyser/TypeSpecifier.php#L2258-L2269

src/Analyser/TypeSpecifier.php
staabm: agree. I tested it locally and the test starts failling when I manually mutate the code.

src/Analyser/TypeSpecifier.php
staabm: extracted into separate feature request: https://github.com/phpstan/phpstan/issues/14081

PR #4538, opened 2025-11-08
Allow offset in fori loop

SplotyCode:


tests/PHPStan/Analyser/nsrt/for-loop-expr.php
```diff
@@ -50,3 +50,24 @@ function getItemsArray(array $items): array
...
 	return $items;
 }
+
+/**
+ * @param array<string> $items
+ */
+function skipFirstElement(array $items): void
+{
+	for ($i = 1; count($items) > $i; ++$i) {
+		$items[$i] = 'hello';
```
staabm: The newly added test miss assertType() calls to assert type results

staabm (changes requested):
the newly added tests do not fail without the change you applied to NodeScopeResolver.

you either need to add more/other assertions or adjust the test-code

ondrejmirtes:
I'm sorry. Please reopen this PR on the latest 2.1.x if you're still interested in making this fix.

PR #4539, opened 2025-11-08
Fix non-existent offset when guarding

SplotyCode:
This pr is trying to fix a false positive.

```
/**
 * @param list<int> $array
 * @param positive-int $index
 */
public function guardNotSafeLowerBound(array $array, int $index) {
    if ($index < count($array)) { // index is within 0..count(array)-1
        return $array[$index]; // phpstan reports a potentially unsafe array access
    }
    return null;
}
```
Note that 1) $array is a list 2) $index is positive (or at least >= 0)

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -307,7 +307,7 @@ final class TypeSpecifier
 							$argType->getIterableValueType(),
 							TypeSpecifierContext::createTrue(),
 							$scope,
-						)->setRootExpr($expr)
+						)
 					);
 				}
 
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -273,7 +273,8 @@ final class TypeSpecifier
 				&& in_array(strtolower((string) $expr->right->name), ['count', 'sizeof'], true)
 				&& $leftType->isInteger()->yes()
 			) {
-				$argType = $scope->getType($expr->right->getArgs()[0]->value);
+				$argExpr = $expr->right->getArgs()[0]->value;
+				$argType = $scope->getType($argExpr);
 
 				if ($leftType instanceof ConstantIntegerType) {
 					if ($orEqual) {
@@ -295,10 +296,11 @@ final class TypeSpecifier
 					$context->true()
 					&& $expr instanceof Node\Expr\BinaryOp\Smaller
 				    && $argType->isList()->yes()
+					&& $argExpr instanceof Expr\Variable
 					&& IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($leftType)->yes()
 				) {
 					$dimFetch = new ArrayDimFetch(
-						$expr->right->getArgs()[0]->value,
+						$argExpr,
 						$expr->left,
 					);
 					$result = $result->unionWith(
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -291,6 +291,25 @@ final class TypeSpecifier
 				if ($specifiedTypes !== null) {
 					$result = $result->unionWith($specifiedTypes);
 				}
+				if (
+					$context->true()
+					&& $expr instanceof Node\Expr\BinaryOp\Smaller
+				    && $argType->isList()->yes()
+					&& IntegerRangeType::fromInterval(0, null)->isSuperTypeOf($leftType)->yes()
+				) {
+					$dimFetch = new ArrayDimFetch(
+						$expr->right->getArgs()[0]->value,
+						$expr->left,
+					);
+					$result = $result->unionWith(
+						$this->create(
+							$dimFetch,
+							$argType->getIterableValueType(),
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Arrays/data/report-possibly-nonexistent-array-offset.php
```diff
@@ -57,4 +57,25 @@ public function nonEmpty(array $a): void
 		echo $a[0];
 	}
 
+	/**
+	 * @param list<int> $array
+	 * @param positive-int $index
```
staabm: Could even include 0

tests/PHPStan/Rules/Arrays/data/report-possibly-nonexistent-array-offset.php
```diff
@@ -57,4 +57,25 @@ public function nonEmpty(array $a): void
...
+		}
+		return null;
+	}
+
+
+	/**
+	 * @param list<int> $array
+	 */
+	public function guardNotSafe(array $array, int $index) {
+		if ($index < count($array)) {
```
staabm, line 76: Another test with <=

tests/PHPStan/Rules/Arrays/data/report-possibly-nonexistent-array-offset.php
```diff
@@ -57,4 +57,25 @@ public function nonEmpty(array $a): void
...
+	public function guard(array $array, int $index) {
+		if ($index < count($array)) {
+			return $array[$index];
+		}
+		return null;
+	}
+
+
+	/**
+	 * @param list<int> $array
```
staabm, line 73: Another test with array instead of list

src/Analyser/TypeSpecifier.php
```diff
@@ -291,6 +291,25 @@ public function specifyTypesInCondition(
...
+					);
+					$result = $result->unionWith(
+						$this->create(
+							$dimFetch,
+							$argType->getIterableValueType(),
+							TypeSpecifierContext::createTrue(),
+							$scope,
+						)
+					);
+				}
```
staabm, line 314: is this PR trying to fix https://github.com/phpstan/phpstan/issues/13773 ?
(since the PR does not at all contain a description.. I can only guess)

if so, you should either adjust the logic added with https://github.com/phpstan/phpstan-src/pull/4403 or maybe even delete it in case you want to re-implement it in a different place

SplotyCode: Thanks for pointing this out! I updated the description of this pr.

I have not opened a issue for this pr. But i will also try and fix https://github.com/phpstan/phpstan/issues/13773 later in a diffrent pr. Here i am trying to fix array access on lists if we explicitly guard the index.

SplotyCode:
There seems to be a regression
```
$var = null;
if ($index < count($array) {
    $var = 2;
}
if ($index < count($array) {
    /* phpstan does not recognize that $var is always 2 now */
}
```
This did work before this pr. I am wondering if this is also true for the other checks in the method i changed.

tests/PHPStan/Analyser/nsrt/bug-10264.php
```diff
@@ -33,7 +33,7 @@ public function sayHello(array $c): void
 		assertType('list<int>', $c);
 		if (count($c) > 0) {
 			$c = array_map(fn() => new stdClass(), $c);
-			assertType('non-empty-list<stdClass>', $c);
+			assertType('non-empty-list<stdClass>&hasOffsetValue(0, stdClass)', $c);
```
staabm, line 36: this type does not make sense, because `non-empty-list<stdClass>` already implies that the list has a offset `0` with value `stdClass`

src/Analyser/TypeSpecifier.php
```diff
@@ -291,6 +292,26 @@ public function specifyTypesInCondition(
 				if ($specifiedTypes !== null) {
 					$result = $result->unionWith($specifiedTypes);
 				}
+				if (
+					$context->true()
+					&& $expr instanceof Node\Expr\BinaryOp\Smaller
+				    && $argType->isList()->yes()
+					&& $argExpr instanceof Expr\Variable
```
staabm, line 299: `instanceof Expr\Variable` is usually a bug as most logic works for other expressions (e.g. a pure function call can also be remembered).

most of the time this works just by dropping this line

src/Analyser/TypeSpecifier.php
```diff
@@ -291,6 +292,26 @@ public function specifyTypesInCondition(
...
+				) {
+					$dimFetch = new ArrayDimFetch(
+						$argExpr,
+						$expr->left,
+					);
+					$result = $result->unionWith(
+						$this->create(
+							$dimFetch,
+							$argType->getIterableValueType(),
+							TypeSpecifierContext::createTrue(),
```
staabm, line 310: usually we pass thru the $context, instead of creating a new one.

ondrejmirtes:
This PR looks abandonded. Feel free to reopen on latest 2.1.x if you still want to fix it.

PR #4541, opened 2025-11-09
Fix `UnionType::getObjectClassReflections`

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13777

The issue is that `UnionType::getObjectClassReflections` was returning `[]` as soon as one of subtypes returns `[]`...

### DIFF
--- a/tests/PHPStan/Rules/PhpDoc/RequireExtendsDefinitionClassRuleTest.php
+++ b/tests/PHPStan/Rules/PhpDoc/RequireExtendsDefinitionClassRuleTest.php
@@ -22,6 +22,7 @@ class RequireExtendsDefinitionClassRuleTest extends RuleTestCase
 		$container = self::getContainer();
 		return new RequireExtendsDefinitionClassRule(
 			new RequireExtendsCheck(
+				$reflectionProvider,
 				new ClassNameCheck(
 					new ClassCaseSensitivityCheck($reflectionProvider, true),
 					new ClassForbiddenNameCheck($container),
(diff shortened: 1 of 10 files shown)


tests/PHPStan/Rules/PhpDoc/RequireExtendsDefinitionClassRuleTest.php
```diff
@@ -85,11 +85,6 @@ public function testRule(): void
 				183,
 				'Learn more at https://phpstan.org/user-guide/discovering-symbols',
 			],
-			[
-				'PHPDoc tag @phpstan-require-extends contains unknown class IncompatibleRequireExtends\SomeClass.',
```
VincentLanglet, line 90: Funny to see that test was added with the bug (since SomeClass was existing)

ondrejmirtes (changes requested):
I think the behaviour of UnionType is correct here. We do pickFromTypes for everything. Once the union type is for example `Foo|null`, both getObjectClassNames and getObjectClassReflections should return an empty array.

But I don't understand how returning an empty array can cause an error message to appear. Looks like something else needs fixing, like the rule.

Looks like RequireExtendsCheck has some complicated logic I didn't write. The logic there using getObjectClassNames and getObjectClassReflections and trying to map something between them is definitely too complicated and suspicious.

ondrejmirtes (changes requested):
src/Rules/PhpDoc/RequireImplementsDefinitionTraitRule.php was changed but not tested

ondrejmirtes:
Thank you!

PR #4543, opened 2025-11-10
Improve EnumSanityRule

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13768

### DIFF
--- a/src/Rules/Classes/EnumSanityRule.php
+++ b/src/Rules/Classes/EnumSanityRule.php
@@ -8,6 +8,7 @@ use PHPStan\DependencyInjection\RegisteredRule;
 use PHPStan\Node\InClassNode;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\IntegerType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\VerbosityLevel;
@@ -16,6 +17,8 @@ use function array_key_exists;
 use function count;
 use function implode;
 use function in_array;
+use function is_int;
+use function is_string;
 use function sprintf;
 
 /**
@@ -176,18 +179,22 @@ final class EnumSanityRule implements Rule
 			}
 
 			$exprType = $scope->getType($stmt->expr);
-			$constantValues = $exprType->getConstantScalarValues();
-			if (count($constantValues) === 1) {
-				$caseValue = $constantValues[0];
-				if (!isset($enumCases[$caseValue])) {
-					$enumCases[$caseValue] = [];
-				}
-
-				$enumCases[$caseValue][] = $caseName;
-			}
-
 			$scalarType = $enumNode->scalarType->toLowerString() === 'int' ? new IntegerType() : new StringType();
 			if ($scalarType->isSuperTypeOf($exprType)->yes()) {
+				$constantValues = $exprType->getConstantScalarValues();
+				if (count($constantValues) === 1) {
+					$caseValue = $constantValues[0];
+					if (!is_string($caseValue) && !is_int($caseValue)) {
+						throw new ShouldNotHappenException();
+					}
+
+					if (!isset($enumCases[$caseValue])) {
+						$enumCases[$caseValue] = [];
+					}
+
+					$enumCases[$caseValue][] = $caseName;
+				}
+
 				continue;
 			}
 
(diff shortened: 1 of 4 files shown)


src/Rules/Classes/EnumSanityRule.php
```diff
@@ -145,27 +145,16 @@ public function processNode(Node $node, Scope $scope): array
...
-
-					if (!isset($enumCases[$caseValue])) {
-						$enumCases[$caseValue] = [];
-					}
-
-					$enumCases[$caseValue][] = $caseName;
-				}
+			if ($enumNode->scalarType === null && $stmt->expr !== null) {
+				$errors[] = RuleErrorBuilder::message(sprintf(
+					'Enum %s is not backed, but case %s has value.',
```
VincentLanglet: The message changed from
```
Enum %s is not backed, but case %s has value %s.
```
to
```
Enum %s is not backed, but case %s has value.
```

I'm not sure you're okay with this or if
- it should be behind bleedingEdge
- it should be behind some extra condition (and have two differents messages ?)

src/Rules/Classes/EnumSanityRule.php
```diff
@@ -187,6 +176,16 @@ public function processNode(Node $node, Scope $scope): array
 			}
 
 			$exprType = $scope->getType($stmt->expr);
+			$constantValues = $exprType->getConstantScalarValues();
+			if (count($constantValues) === 1) {
```
VincentLanglet: By setting enum cases value here, instead of in the condition
```
if ($stmt->expr instanceof Node\Scalar\Int_ || $stmt->expr instanceof Node\Scalar\String_)
```
I also catch other constant values like the one set with a class Constant.

Cf test
```
enum Backed: int {
	case One = Foo::A;
	case Two = Foo::A;
}
```

src/Rules/Classes/EnumSanityRule.php
```diff
@@ -189,6 +181,20 @@ public function processNode(Node $node, Scope $scope): array
 			$exprType = $scope->getType($stmt->expr);
 			$scalarType = $enumNode->scalarType->toLowerString() === 'int' ? new IntegerType() : new StringType();
 			if ($scalarType->isSuperTypeOf($exprType)->yes()) {
+				$constantValues = $exprType->getConstantScalarValues();
```
staabm, line 196: I guess you could support IntegerRanges with getFinitTypes()

VincentLanglet: Do you have a real case scenario ?

EnumCase are support to be constant values. `case Foo = 1`
I'm just adding the support of thing like `case Foo = SomeClass::Constant` by not just looking at scalar int/string but directly the inferred type.

I don't see value of integer range here

staabm: you are right

tests/PHPStan/Rules/Classes/EnumSanityRuleTest.php
```diff
@@ -119,24 +119,28 @@ public function testBug9402(): void
 	}
 
 	#[RequiresPhp('>= 8.1')]
-	public function testBug11592(): void
+	public function testBug13768(): void
```
staabm, line 148: this looks like a bad merge conflict resolution (as a pre-existing test was deleted/overwritten)

VincentLanglet: Indeed, fixed

src/Rules/Classes/EnumSanityRule.php
ondrejmirtes: Why would the message change? If it has value, we can still include it in the error message.

VincentLanglet: The check was
```
if ($stmt->expr instanceof Node\Scalar\Int_ || $stmt->expr instanceof Node\Scalar\String_) {
```
and is changed to
```
if ($enumNode->scalarType === null && $stmt->expr !== null) {
```
so I cannot safely access to `$stmt->value`

I made https://github.com/phpstan/phpstan-src/pull/4543/commits/8e72e2bcfdc8ace4ef8198f362a179fee7cd6b60 to revert this change for Int/String/Float, but the test show the issue with other things like
```
	case E = false;
	case F = Foo::A;
```
where no value is displayed. Is there some helper method to get a value/name from an Expr ? Or is it ok like this ?

ondrejmirtes: You can either use ExprPrinter on `$stmt->expr`, or you can get type of `$stmt->expr` with InitializerExprTypeResolver and do `->describe(VerbosityLevel::value())` on the type. I like the 2nd option more.

ondrejmirtes:
CS build fail

PR #4544, opened 2025-11-10
Fix return type of pcntl_signal_get_handler

marc-mabe:
`pcntl_signal_get_handler` returns the defined callable or `SIG_DFL` / `SIG_IGN` if none has been defined via `pcntl_signal`.

https://www.php.net/manual/en/function.pcntl-signal-get-handler.php
https://github.com/php/php-src/blob/e97c0dc340be7387498b92b182b9356b960a3bd0/ext/pcntl/pcntl.c#L863

https://phpstan.org/r/7632586d-bc04-4b97-865f-3cb5125037a8


ondrejmirtes:
Thank you.

PR #4546, opened 2025-11-11
Add `#[CoversNothing]` to classes which test other test-artifacts

staabm:


ondrejmirtes:
I don't get this at all, you're going to have to explain it to me 😊

staabm:
thesis is, that test-classes adjusted in this PR test other example test-classes which are contained in `phpstan-src/tests`.

e.g. `AnonymousClassNameRuleTest` is a test for `tests/PHPStan/Analyser/AnonymousClassNameRule.php`, which means it does not cover classes from `src/` or only indrectly.

PR #4545, opened 2025-11-11
add Type::isNever and Type::isExplicitNever methods

canvural:
This PR adds `Type::isNever` and `Type::isExplicitNever` methods

PR #4547, opened 2025-11-11
add Type::isNever and Type::isExplicitNever methods

canvural:
Re-opening #4545

I think the issue bot changes are valid and is the same case as I described [here](https://github.com/phpstan/phpstan-src/pull/4530#discussion_r2514168648)

### DIFF
--- a/src/Type/UnionType.php
+++ b/src/Type/UnionType.php
@@ -751,6 +751,16 @@ class UnionType implements CompoundType
 		return $this->unionResults(static fn (Type $type): TrinaryLogic => $type->isScalar());
 	}
 
+	public function isNever(): TrinaryLogic
+	{
+		return $this->unionResults(static fn (Type $type): TrinaryLogic => $type->isNever());
+	}
+
+	public function isExplicitNever(): TrinaryLogic
+	{
+		return $this->unionResults(static fn (Type $type): TrinaryLogic => $type->isExplicitNever());
+	}
+
 	public function looseCompare(Type $type, PhpVersion $phpVersion): BooleanType
 	{
 		return $this->notBenevolentUnionResults(
--- a/src/Type/UnionType.php
+++ b/src/Type/UnionType.php
@@ -243,7 +243,7 @@ class UnionType implements CompoundType
 		if (
 			($otherType instanceof self && !$otherType instanceof TemplateUnionType)
 			|| ($otherType instanceof IterableType && !$otherType instanceof TemplateIterableType)
-			|| $otherType instanceof NeverType
+			|| $otherType->isNever()->yes()
 			|| ($otherType instanceof LateResolvableType && $otherType instanceof CompoundType && !$otherType instanceof TemplateType)
 			|| $otherType instanceof IntegerRangeType
 		) {
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1438,6 +1438,18 @@ parameters:
 			path: src/Type/IterableType.php
 
 		-
+			rawMessage: 'Doing instanceof PHPStan\Type\NeverType is error-prone and deprecated. Use Type::isNever() or Type::isExplicitNever() instead.'
+			identifier: phpstanApi.instanceofType
+			count: 2
+			path: src/Type/NeverType.php
+
+		-
+			rawMessage: 'Doing instanceof PHPStan\Type\NeverType is error-prone and deprecated. Use Type::isNever() or Type::isExplicitNever() instead.'
+			identifier: phpstanApi.instanceofType
+			count: 1
+			path: src/Type/NonAcceptingNeverType.php
+
+		-
 			rawMessage: 'Doing instanceof PHPStan\Type\ConstantScalarType is error-prone and deprecated. Use Type::isConstantScalarValue() or Type::getConstantScalarTypes() or Type::getConstantScalarValues() instead.'
 			identifier: phpstanApi.instanceofType
 			count: 3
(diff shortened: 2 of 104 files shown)


ondrejmirtes:
I disagree about it. There's nothing wrong about this signature:

```php
	/**
	 * @template T of BackedEnum
	 * @param class-string<T> $class
	 * @param value-of<T> $value
	 */
```

I feel like UnresolvableTypeHelper should skip LateResolvableTypes.

canvural:
I think that'd just hide the error. Currently `value-of` with generics and backed enums seems to be broken. And looks like https://github.com/phpstan/phpstan-src/pull/4548 might fix it 👀 

Edit: I copied the snippet you referenced and the fix from #4548 and looks like that fixes it without the need of skipping LateResolvableTypes in UnresolvableTypeHelper

ondrejmirtes:
I'm sorry about sleeping on this. There are now conflicts. Please fix rebase and fix them. At least meanwhile I merged https://github.com/phpstan/phpstan-src/pull/4548.

canvural:
Amazing. Yeah, I was waiting for that PR to be merged first. Will do in the evening 👍🏽

canvural:
@ondrejmirtes Rebased and refactored new usages to methods too.

There is one test failing locally. [Here](https://github.com/phpstan/phpstan-src/blob/349b1ad6e91df5abfd008662698f8dbd564b7e6a/tests/PHPStan/Rules/Methods/data/bug-12219.php#L75) it doesn't report error anymore. @VincentLanglet maybe you have any insights?

src/Type/UnionType.php
```diff
@@ -239,7 +239,7 @@ public function isSuperTypeOf(Type $otherType): IsSuperTypeOfResult
 		if (
 			($otherType instanceof self && !$otherType instanceof TemplateUnionType)
 			|| ($otherType instanceof IterableType && !$otherType instanceof TemplateIterableType)
-			|| $otherType instanceof NeverType
+			|| $otherType->isNever()->yes()
```
VincentLanglet, line 246:
```suggestion
			|| $otherType instanceof CompoundType && $otherType->isNever()->yes()
```
To avoid the error reported maybe ?

VincentLanglet (commented):
When reading the PR, it's unclear to me why sometimes
```
$exprType instanceof NeverType
```
is changed to `!$exprType->isNever()->no()`
and sometimes to `$exprType->isNever()->yes()`

Is `maybe` possible ?
- If not, could we have consistency ? (it seems simpler to use `yes` that `!no()`.
- Also if not, I wonder if the method should return boolean... (Or we will end up with not-caught mutant from mutation testing)

VincentLanglet:
> There is one test failing locally. [Here](https://github.com/phpstan/phpstan-src/blob/349b1ad6e91df5abfd008662698f8dbd564b7e6a/tests/PHPStan/Rules/Methods/data/bug-12219.php#L75) it doesn't report error anymore. @VincentLanglet maybe you have any insights?

No idea like this, the PR is (too) big.
You can start by rebase again since there is conflict, but maybe the right way would be to split the replacement of method into smaller commit in order to check when the CI start to fail ?

canvural:
> > There is one test failing locally. [Here](https://github.com/phpstan/phpstan-src/blob/349b1ad6e91df5abfd008662698f8dbd564b7e6a/tests/PHPStan/Rules/Methods/data/bug-12219.php#L75) it doesn't report error anymore. @VincentLanglet maybe you have any insights?
> 
> No idea like this, the PR is (too) big. You can start by rebase again since there is conflict, but maybe the right way would be to split the replacement of method into smaller commit in order to check when the CI start to fail ?

The issue is that, with this PR an error is not reported whereas the test expects an error. This one https://github.com/phpstan/phpstan-src/blob/8e131a9daafa230b5d9716791ad1b1cbd4e1fbab/tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php#L3713-L3716

Can we say the test is wrong now and remove the error from there?

src/Type/UnionType.php
canvural: I don't know which error you are referring to. This PR did not cause any additional error, it "fixed" one 😄

VincentLanglet:
> Can we say the test is wrong now and remove the error from there?

No. I don't think the test is wrong.
unlike a BackedEnum, `value-of<UnitEnum>` is unresolvable.

src/Type/UnionType.php
VincentLanglet: The one you had to ignore in the baseline https://github.com/phpstan/phpstan-src/pull/4547/changes#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efa

canvural:
78ca602cc491b17db68a5a43beb7d5ba88f34628 fixes it. And doesn't break anything else.

VincentLanglet:
> [78ca602](https://github.com/phpstan/phpstan-src/commit/78ca602cc491b17db68a5a43beb7d5ba88f34628) fixes it. And doesn't break anything else.

Nice, I prefer to let ondrej decide if this is the right way.

(Maybe LateResolvable::isNever should return no (?))

> When reading the PR, it's unclear to me why sometimes
> 
> ```
> $exprType instanceof NeverType
> ```
> 
> is changed to `!$exprType->isNever()->no()` and sometimes to `$exprType->isNever()->yes()`
> 
> Is `maybe` possible ?
> 
> * If not, could we have consistency ? (it seems simpler to use `yes` that `!no()`.
> * Also if not, I wonder if the method should return boolean... (Or we will end up with not-caught mutant from mutation testing)

And about this question ? How did you decide between. `->yes()` and `! ->no()` every time you replace an instanceof check ?

canvural:
> And about this question ? How did you decide between. ->yes() and ! ->no() every time you replace an instanceof check ?

Initially I replaced them with just normal yes and no. But some tests were failing so I changed the conditions to think about maybe too. So currently if you see `!no` instead of `yes` it means it was making a test fail before.

PR #4548, opened 2025-11-11
Resolve template with value of BackEnum

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/12219
Closes https://github.com/phpstan/phpstan/issues/13638
Closes https://github.com/phpstan/phpstan/issues/13282
Closes https://github.com/phpstan/phpstan/issues/13782

### DIFF
--- a/src/Type/ValueOfType.php
+++ b/src/Type/ValueOfType.php
@@ -56,7 +56,7 @@ final class ValueOfType implements CompoundType, LateResolvableType
 			if (
 				$enumCases === []
 				&& $this->type instanceof TemplateType
-				&& (new ObjectType('BackedEnum'))->isSuperTypeOf($this->type->getBound())->yes()
+				&& !(new ObjectType('BackedEnum'))->isSuperTypeOf($this->type->getBound())->no()
 			) {
 				return new UnionType([new IntegerType(), new StringType()]);
 			}
--- a/src/Type/ValueOfType.php
+++ b/src/Type/ValueOfType.php
@@ -56,7 +56,7 @@ final class ValueOfType implements CompoundType, LateResolvableType
 			if (
 				$enumCases === []
 				&& $this->type instanceof TemplateType
-				&& !(new ObjectType('BackedEnum'))->isSuperTypeOf($this->type->getBound())->no()
+				&& (new ObjectType('BackedEnum'))->isSuperTypeOf($this->type->getBound())->yes()
 			) {
 				return new UnionType([new IntegerType(), new StringType()]);
 			}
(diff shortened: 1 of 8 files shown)


staabm:
interessting.. does it also solve https://github.com/phpstan/phpstan/issues/13782 ?

VincentLanglet:
> interessting.. does it also solve [phpstan/phpstan#13782](https://github.com/phpstan/phpstan/issues/13782) ?

You don't appear on the issue bot https://github.com/phpstan/phpstan-src/actions/runs/19269104831?pr=4548
But I can take a look after this one is merged

staabm:
> You don't appear on the issue bot

yes, thats because the issue does not yet have a label like "bug" or "feature request".
issue-bot does not look into unlabelled issues

src/Type/ValueOfType.php
```diff
@@ -50,6 +51,13 @@ public function isResolvable(): bool
 	protected function getResult(): Type
 	{
 		if ($this->type->isEnum()->yes()) {
+			if (
+				$this->type instanceof TemplateType
+				&& $this->type->getBound()->equals(new ObjectType('BackedEnum'))
```
staabm: should this be the other way arround?
```suggestion
				&& new ObjectType('BackedEnum')->isSuperTypeOf($this->type->getBound())->yes()?
```

staabm: just had a look into the code-base. it seems we have spots where a `Bound` is compared using `equals` and sometimes it uses `isSuperTypeOf`.

I am not sure whether thats on purpose or an oversight

VincentLanglet:
> > You don't appear on the issue bot
> 
> yes, thats because the issue does not yet have a label like "bug" or "feature request". issue-bot does not look into unlabelled issues

Oh ok. Test added and seems like it does close https://github.com/phpstan/phpstan/issues/13782.

src/Type/ValueOfType.php
VincentLanglet: isSuperTypeOf is better here indeed, in case someone use 
```
interface SuperBackedEnum extends \BackedEnum
```
and a template of SuperBackedEnum.

test added

VincentLanglet:
Hi @staabm, when I run
```
vendor/bin/phpunit tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php --filter=testBug12219
```
both mutator are caught locally.

Any idea why it could be reported on CI ? Is there some cache ?

VincentLanglet:
> Hi @staabm, when I run
> 
> ```
> vendor/bin/phpunit tests/PHPStan/Rules/Methods/CallMethodsRuleTest.php --filter=testBug12219
> ```
> 
> both mutator are caught locally.
> 
> Any idea why it could be reported on CI ? Is there some cache ?

As shown by https://github.com/phpstan/phpstan-src/pull/4548/commits/6cb52a2049cf9499edc8e25942d5250e1ea6823b

if I tried the mutant reported by https://github.com/phpstan/phpstan-src/actions/runs/19963970402/job/57251263461?pr=4548

it failed.

I got the same issue with the following PR https://github.com/phpstan/phpstan-src/pull/4548

Either I misunderstand something, either it's buggy

staabm:
I had a look but cannot yet make sense of it

ondrejmirtes:
Thank you!

PR #4550, opened 2025-11-11
Fix ltrim and trim on numeric string

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13784


staabm (approved):
nice catch.

VincentLanglet:
> nice catch.

Wasn't easy to understand the issue cause dumpType was displaying `numeric-string` which is the same than the intersection string&numeric-string.

I dunno if some logic exception should be thrown in some case
For instance an UnionType with an `AccessoryType` inside ?
Or a call to TypeCombinator::UnionType which returns an `AccessoryType` ?

staabm:
> I dunno if some logic exception should be thrown in some case
> For instance an UnionType with an `AccessoryType` inside ?
> Or a call to TypeCombinator::UnionType which returns an `AccessoryType` ?

I remember a discussion with ondrey, that we can add a sanity check into TypeCombinator when we can afford a BC break

ondrejmirtes:
Thank you!

PR #4551, opened 2025-11-12
Do not report do-while-false pseudo-loops that are interrupted by a break statement

claudepache:
This enables support of the `do-break-while-false` pattern:

```php
do {
    if ($foo) {
        if ($bar) {
            $a = 'baz';
            break;
        }
     }
     $a = foobar($baz);
} while (false); // no doWhile.alwaysFalse error
```

The error is still reported when the pseudo-loop is not broken with `break`.

### DIFF
--- a/src/Rules/Comparison/DoWhileLoopConstantConditionRule.php
+++ b/src/Rules/Comparison/DoWhileLoopConstantConditionRule.php
@@ -65,6 +65,13 @@ final class DoWhileLoopConstantConditionRule implements Rule
 						return [];
 					}
 				}
+			} else {
+				foreach ($node->getExitPoints() as $exitPoint) {
+					$statement = $exitPoint->getStatement();
+					if ($statement instanceof Break_) {
+						return [];
+					}
+				}
 			}
 
 			$addTip = function (RuleErrorBuilder $ruleErrorBuilder) use ($scope, $node): RuleErrorBuilder {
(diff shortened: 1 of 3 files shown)


claudepache:
(The second commit did not change semantics, only simplified redundant conditions.)

ondrejmirtes:
I rebased the PR. It needs some fixes:


```

Parse error: tests/PHPStan/Rules/Comparison/data/do-while-loop.php:148
    146| 	}
    147| 
  > 148| 	public function doFoo12(array $a)
    149| 	{
    150| 		do {
Cannot redeclare DoWhileLoopConstantCondition\Foo::doFoo12()

```

claudepache:
@ondrejmirtes fixed (and I wonder why it didn’t make the test fail, but anyway...)

ondrejmirtes:
Thank you!

staabm:
@claudepache how did you come up with this construction in PHP?

I am aware of this pattern in languages which have block-scoped variables, but I can't find a use-case in PHP (because variables will be alive even after the block they are assigned/declared in).

no shaming.. I am just curious

ondrejmirtes:
@staabm I'd guess it's another way of basically doing `goto`.

claudepache:
@staabm Nothing to do with variable scope, only with code flow. Here is an example of use:

```php
// $date_in might be: '2.2025', '02 2025', '02-2025', '2025-02',  '2/2025', '022025', ...
do {
    $date_tmp = explode('.', $date_in);
    if (ctype_digit($date_tmp[1] ?? ''))
        break;
    $date_tmp = explode(' ', $date_in);
    if (ctype_digit($date_tmp[1] ?? ''))
        break;
    $date_tmp = explode('-', $date_in);
    if (ctype_digit($date_tmp[1] ?? ''))
        break;
    $date_tmp = explode('/', $date_in);
    if (ctype_digit($date_tmp[1] ?? ''))
        break;
    if ((strlen($date_in) == 4) || (strlen($date_in) == 6)) {
        $date_tmp[0] = substr($date_in, 0, 2);
        $date_tmp[1] = substr($date_in, 2   );
    }
} while (false);
// $date_tmp is assumed to hold date components ($month and $year in either order)
// continue processing...

PR #4552, opened 2025-11-12
add default value for getAttribute call in ContinueBreakInLoopRule

canvural:
closes https://github.com/phpstan/phpstan/issues/13790


staabm (approved):
I had a rough look thru the whole codebase for similar issues and couldn't find any other.

lgtm - thank you

staabm:
@canvural the referenced issue is also talking about `continue`, which likely needs another test and/or fix

canvural:
> @canvural the referenced issue is also talking about `continue`, which likely needs another test and/or fix

Good idea. Added with 969b9d871f7597be832703f201ab2d1ce67252d9

ondrejmirtes:
Nice, thank you!

PR #4553, opened 2025-11-13
[PHP 8.5] add new rule for filter_var

canvural:
Follow up to #4495 

In PHP 8.5 using both `FILTER_THROW_ON_FAILURE` and `FILTER_NULL_ON_FAILURE` at same time is not allowed. https://3v4l.org/Ibd4B/rfc#vgit.master_jit

Not sure about the rule level and the message.

### DIFF
--- a/src/Rules/Functions/FilterVarRule.php
+++ b/src/Rules/Functions/FilterVarRule.php
@@ -1,0 +1,69 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Functions;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Name;
+use PHPStan\Analyser\Scope;
+use PHPStan\DependencyInjection\RegisteredRule;
+use PHPStan\Reflection\ReflectionProvider;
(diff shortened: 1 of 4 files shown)


src/Rules/Functions/FilterVarRule.php
```diff
@@ -0,0 +1,69 @@
...
+		if ($this->reflectionProvider->hasConstant(new Name\FullyQualified('FILTER_THROW_ON_FAILURE'), null)) {
+			if (count($args) < 3) {
+				return [];
+			}
+
+			$flagsType = $scope->getType($args[2]->value);
+
+			if ($this->filterFunctionReturnTypeHelper->hasFlag('FILTER_NULL_ON_FAILURE', $flagsType)
+				->and($this->filterFunctionReturnTypeHelper->hasFlag('FILTER_THROW_ON_FAILURE', $flagsType))
+				->yes()
```
VincentLanglet, line 56: Maybe you should also report maybe if `reportMaybe` is enabled.

But I'm unsure how good those filterFunctionReturnTypeHelper::hasFlag works so far

ondrejmirtes:
Thank you!

PR #4554, opened 2025-11-13
Don't loose known offset-types in array_merge()

staabm:
the result of `array_merge` will always contain all `string` keys we see along the way.
~~we can't reason about the values when not all arrays are constant, but we can remember the keys.~~
since string keys overwrite each other, we know the value of the last string keys in the chain.
of other string based offsets we at least know about their existance.

best [reviewed with whitespaces ignored](https://github.com/phpstan/phpstan-src/pull/4554/files?w=1)

Followup could be doing the same for 
- the `+` operator when merging arrays
- `array_replace`
- spread operator

### DIFF
--- a/src/Type/Accessory/HasOffsetType.php
+++ b/src/Type/Accessory/HasOffsetType.php
@@ -53,10 +53,7 @@ class HasOffsetType implements CompoundType, AccessoryType
 	{
 	}
 
-	/**
-	 * @return ConstantStringType|ConstantIntegerType
-	 */
-	public function getOffsetType(): Type
+	public function getOffsetType(): ConstantStringType|ConstantIntegerType
 	{
 		return $this->offsetType;
 	}
--- a/src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ArrayMergeFunctionDynamicReturnTypeExtension.php
@@ -23,6 +23,7 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeUtils;
 use function array_keys;
+use function array_values;
 use function count;
 use function in_array;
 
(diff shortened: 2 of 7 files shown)


src/Type/Accessory/HasOffsetType.php
```diff
@@ -52,10 +52,7 @@ public function __construct(private ConstantStringType|ConstantIntegerType $offs
 	{
 	}
 
-	/**
-	 * @return ConstantStringType|ConstantIntegerType
-	 */
-	public function getOffsetType(): Type
+	public function getOffsetType(): ConstantStringType|ConstantIntegerType
```
staabm, line 56: analog HasOffsetValueType->getOffsetType()

ondrejmirtes:
I like this 👍 Before looking at the code I thought I'd solve this with HasOffsetValueType as well.

ondrejmirtes (changes requested):
Please do HasOffsetValueType instead in all cases, the result will be more precise.

claudepache:
> the result of `array_merge` will always contain all keys we see along the way.

This is not correct for int keys, as those will be renumbered. (This is the reason `array_replace()` was added.) 

https://3v4l.org/2h5Qn

staabm:
@claudepache thanks - I also realized it. Adjusted the PR accordingly, please have another look

VincentLanglet:
Could it solve https://github.com/phpstan/phpstan/issues/13805 ?

staabm:
> Could it solve [phpstan/phpstan#13805](https://github.com/phpstan/phpstan/issues/13805) ?

there is no `array_merge` call in this issue. but I think we might be able to apply a similar logic for this case.
there are even more use-cases see the PR description (just added spread operator to the list).

staabm:
Looking at this I think we should add `Type->getOffsetTypes()` which returns a array of all known offsets of a type (const-ints/string), no matter its known by keys (const array), or by additional accessories.

This would ease building logic on top of it without the need for checking individual type classes

ondrejmirtes:
Thank you!

PR #4555, opened 2025-11-14
string cast might throw exception

canvural:
closes https://github.com/phpstan/phpstan/issues/13806

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -3752,10 +3752,12 @@ class NodeScopeResolver
 			$exprType = $scope->getType($expr->expr);
 			$toStringMethod = $scope->getMethodReflection($exprType, '__toString');
 			if ($toStringMethod !== null) {
-				if ($toStringMethod->getThrowType() !== null) {
-					$throwPoints[] = InternalThrowPoint::createExplicit($scope, $toStringMethod->getThrowType(), $expr, false);
-				} else {
-					$throwPoints[] = InternalThrowPoint::createImplicit($scope, $expr);
+				if ($this->phpVersion->throwsOnStringCast()) {
+					if ($toStringMethod->getThrowType() !== null) {
+						$throwPoints[] = InternalThrowPoint::createExplicit($scope, $toStringMethod->getThrowType(), $expr, false);
+					} else {
+						$throwPoints[] = InternalThrowPoint::createImplicit($scope, $expr);
+					}
 				}
 
 				if (!$toStringMethod->hasSideEffects()->no()) {
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -3752,6 +3752,12 @@ class NodeScopeResolver
 			$exprType = $scope->getType($expr->expr);
 			$toStringMethod = $scope->getMethodReflection($exprType, '__toString');
 			if ($toStringMethod !== null) {
+				if ($toStringMethod->getThrowType() !== null) {
+					$throwPoints[] = InternalThrowPoint::createExplicit($scope, $toStringMethod->getThrowType(), $expr, false);
+				} else {
+					$throwPoints[] = InternalThrowPoint::createImplicit($scope, $expr);
+				}
+
 				if (!$toStringMethod->hasSideEffects()->no()) {
 					$impurePoints[] = new ImpurePoint(
 						$scope,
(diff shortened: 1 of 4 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -3619,6 +3619,12 @@ static function (): void {
 			$exprType = $scope->getType($expr->expr);
 			$toStringMethod = $scope->getMethodReflection($exprType, '__toString');
 			if ($toStringMethod !== null) {
+				if ($toStringMethod->getThrowType() !== null) {
```
staabm: the rfc is PHP 7.4+, meaning this logic should be gated via `PHPVersion` class.

note though, that NodeScopeResolver is undergoing a major refactoring atm, so this might conflict easily

ondrejmirtes: It won't conflict, I'm not doing any major changes in NodeScopeResolver that I'm not pushing.

But it's possible that the same change also needs to happen in PHPStan\Analyser\Generator namespace. In this case it doesn't because there's no `StringCastHandler` ported yet.

tests/PHPStan/Rules/Exceptions/data/bug-13806.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Bug13806;
```
VincentLanglet, line 3: I feel like this lack of test, 

It should add a test for 
```
function doFoo(Stringable $myVariable): void
```
to check the implicit throw tag

and a test for 
```
class MyStringVoid {
	/** @throws void */
	public function __toString() {
	}
}
```
to check in this case the catch is reported as dead catch

src/Analyser/NodeScopeResolver.php
canvural: I thought the minimum version PHPStan supports was 7.4 Is it 7.1?

staabm: there are 2 things. 

the php version PHPStan can analyze against, configured in the .neon config: https://github.com/phpstan/phpstan-src/blob/5afb0194d994952638226711a493179c3341c451/conf/parametersSchema.neon#L103

and the PHP runtime version you use to run PHPStan itself: 
https://github.com/phpstan/phpstan/blob/5afb019da9f70ae5a2b4247a1830b5103423a325/composer.json#L7

VincentLanglet:
Does it also solve https://github.com/phpstan/phpstan/issues/5952 ?
You could add a test for this issue :)

tests/PHPStan/Rules/Exceptions/CatchWithUnthrownExceptionRuleTest.php
```diff
@@ -643,4 +643,14 @@ public function testPropertyHooks(): void
 		]);
 	}
 
+	public function testBug13806(): void
```
staabm, line 646: I wonder why this test (and/or its result) does not depend on php version

src/Analyser/NodeScopeResolver.php
```diff
@@ -3752,6 +3752,14 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context, $sto
 			$exprType = $scope->getType($expr->expr);
 			$toStringMethod = $scope->getMethodReflection($exprType, '__toString');
 			if ($toStringMethod !== null) {
+				if ($this->phpVersion->throwsOnStringCast()) {
+					if ($toStringMethod->getThrowType() !== null) {
```
ondrejmirtes, line 3756: Nice, almost perfect 😊 But I think we're missing a few things:

1) It should support `@throws void` as a means to mark "this does not throw exceptions".
2) It should respect `$this->implicitThrows`. When throwType is `null`, if this property is true, it means an implicit throw point should be added. If this property is false, it means that `__toString` does not throw an exception.

There's already code in NodeScopeResolver handling these things - `getMethodThrowPoint`. Should be possible to reuse it.

ondrejmirtes (changes requested):

src/Analyser/NodeScopeResolver.php
canvural: 1. It already supports that, no? The test case reports dead catch on line 16.
2. How can I test that? Or just re-using `getMethodThrowPoint` is enough?

VincentLanglet:
Solved by https://github.com/phpstan/phpstan-src/pull/5391

PR #4560, opened 2025-11-17
De-duplicate phpstan.php errors

staabm:
closes https://github.com/phpstan/phpstan/issues/13813

### DIFF
--- a/src/Analyser/FileAnalyser.php
+++ b/src/Analyser/FileAnalyser.php
@@ -29,7 +29,6 @@ use function get_class;
 use function hash;
 use function is_dir;
 use function is_file;
-use function md5;
 use function restore_error_handler;
 use function set_error_handler;
 use function sprintf;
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Thank you!

ondrejmirtes:
Hey, I merged this too soon, there are some problems including CS and failing tests :)

staabm:
Hey,

[cs fix is here](https://github.com/phpstan/phpstan-src/pull/4562) - the test errors seem pre-existing/unrelated to this PR.
seem GNSR related

claudepache:
What should happen when there are two identical errors on the same line?

https://phpstan.org/r/0a335da3-97dc-4ad7-9265-813ce48b1b7a

staabm:
This PR affects only runtime php warnings/errors, which are not very likely to happen at all while running static analysis. 

It does not affect regular phpstan errors. 

I think its fine that we have each runtime warning/error only once per line, as such codes with multiple statements per line is also super rare.

PR #4566, opened 2025-11-19
Prevent global side-effect in Bug13813IntegrationTest

staabm:


tests/PHPStan/Analyser/Bug13813IntegrationTest.php
```diff
@@ -44,6 +46,8 @@ public function testBug13813(): void
 				$analyzerResult->getAllPhpErrors()[1]->getMessage(),
 			);
 		}
+
+		error_reporting($oldReporting);
```
ondrejmirtes: I think the:

```php
		$analyzerResult = $this->runAnalyse([
			__DIR__ . '/data/bug-13813.php',
			__DIR__ . '/Bug13813Rule.php',
		]);
```

should be wrapped in a try-finally block and the restore be done there.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4567, opened 2025-11-20
Fix the handling of sealed concrete classes

stof:
Closes https://github.com/phpstan/phpstan/issues/13734


ondrejmirtes:
Thank you!

PR #4568, opened 2025-11-23
Reduce ContainerFactory::postInitializeContainer() calls

staabm:
prevents duplicate work in data-providers

before this PR
```
➜  phpstan-src git:(pstint) ✗ hyperfine '../phpunit/phpunit --list-suites'
Benchmark 1: ../phpunit/phpunit --list-suites
  Time (mean ± σ):     10.575 s ±  0.160 s    [User: 9.804 s, System: 0.759 s]
  Range (min … max):   10.218 s … 10.711 s    10 runs
```

after this PR:
```
➜  phpstan-src git:(2.1.x) ✗ hyperfine '../phpunit/phpunit --list-suites'
Benchmark 1: ../phpunit/phpunit --list-suites
  Time (mean ± σ):     10.397 s ±  0.151 s    [User: 9.655 s, System: 0.725 s]
  Range (min … max):   10.187 s … 10.623 s    10 runs
```

PR #4572, opened 2025-11-24
Reduce ContainerFactory::postInitializeContainer() calls

staabm:
missed this one in https://github.com/phpstan/phpstan-src/pull/4568


ondrejmirtes:
Thank you!

PR #4574, opened 2025-11-25
[PHP 8.5] ini_get("max_memory_limit") is a string

staabm:


tests/PHPStan/Analyser/nsrt/ini-get.php
```diff
@@ -26,4 +26,11 @@ function doFoo() {
 	}
 	assertType('string|false', ini_get($key));
 	assertType('string|false', ini_get('unknown'));
+
+	if (PHP_VERSION_ID >= 80500) {
+		assertType('string', ini_get("max_memory_limit"));
```
ondrejmirtes, line 31: This works in type inference tests? 😱  Can you point to a piece of code that makes it work? We might want to refactor some tests thanks to this.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/ini-get.php
staabm: I think it works because the constant PHP_VERSION_ID is narrowed via expressionTypes and these are taken into account in `MutatingScope->getPhpVersion()` via https://github.com/phpstan/phpstan-src/blob/2674c0f79ed00d8a89e7d17b7c236cbb7b7a1a94/src/Analyser/MutatingScope.php#L6330

(it works as long as the relevant extensions utilize `$scope->getPhpVersion()` instead of a injected PhpVersion from the DIC)

ondrejmirtes: Yeah, that makes sense! I thought that TypeInferenceTestCase somehow skips the asserts if they are for a different PHP version.

ondrejmirtes: This test fails though https://github.com/phpstan/phpstan-src/actions/runs/19663206996/job/56313831308?pr=4574

staabm: good catch. this was a proof of concept that it actually works ;)

ondrejmirtes:
Thank you!

PR #4575, opened 2025-11-25
Fix verbose output alignment

thg2k:
Fun fact: You only notice the misalignment with 1 cpu core. So... 😆

### DIFF
--- a/src/Parallel/Scheduler.php
+++ b/src/Parallel/Scheduler.php
@@ -65,12 +65,7 @@ final class Scheduler implements DiagnoseExtension
 		[$cpuCores, $filesCount, $jobsCount, $usedNumberOfProcesses] = $this->storedData;
 
 		$output->writeLineFormatted('<info>Parallel processing scheduler:</info>');
-		$output->writeLineFormatted(sprintf(
-			'# of detected CPU %s:  %s%d',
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I don't get it. Please show me a screenshot before/after.

thg2k:
It's really trivial, the second case has the filler space in the wrong spot, it should fill the missing "s", instead it adds it

ondrejmirtes:
Please show me a screenshot before/after.

thg2k:
Before:

image

After:

image

Again.. this only happens with one cpu core. You can easily reproduce it locally by adding a temporary `$cpuCores = 1;` at the beginning of the print method.

thg2k:
Now that I read it.. I think it should actually always say "# of detected CPU cores:" in the plural form.

Should I change the PR accordingly?

Now:

```
# of detected CPU core:  1
# of detected CPU cores:  4
```

After:

```
# of detected CPU cores:  1
# of detected CPU cores:  4
```

What do you think?

thg2k:
Now it's always in the plural form, like the other cases. It makes more sense to use the plural form even in the case of 1.

image

ondrejmirtes:
Thank you!

PR #4577, opened 2025-11-26
Cache files in OptimizedDirectorySourceLocatorFactory

staabm:
shot in the dark for the problem seen in https://github.com/phpstan/phpstan/discussions/13852

---

This PR adds caching when called with a array of files. 

Before this PR the existing caching did only work for directories. 

I tried this PR on the phpunit codebase, which proofed that with this change the numer pf `findSymbols` calls will be reduced to the files actually changed. Unrelated files (like bootstrap files) will no longer be processed for symbols)

---

this PR combined with https://github.com/phpstan/phpstan-src/pull/4576 I can see the phpstan test-suite running with `php tests/vendor/bin/paratest ` in 38-40 seconds

### DIFF
--- a/src/Reflection/BetterReflection/SourceLocator/ComposerJsonAndInstalledJsonSourceLocatorMaker.php
+++ b/src/Reflection/BetterReflection/SourceLocator/ComposerJsonAndInstalledJsonSourceLocatorMaker.php
@@ -129,7 +129,7 @@ final class ComposerJsonAndInstalledJsonSourceLocatorMaker
 		}
 
 		if (count($files) > 0) {
-			$locators[] = $this->optimizedDirectorySourceLocatorFactory->createByFiles($files);
+			$locators[] = $this->optimizedDirectorySourceLocatorFactory->createByFiles($files, 'odsl-installed-files');
 		}
 
 		$binDir = ComposerHelper::getBinDirFromComposerConfig($projectInstallationPath, $composer);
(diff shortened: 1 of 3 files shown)


staabm:
before this PR: running `time php bin/phpstan analyze` on phpstan-src@63ee5de reports `OptimizedDirectorySourceLocatorFactory->findSymbols` beeing invoked..
- 4248x when no cache exists
- 24x when no files changed
- 24x when `src/Cache/CacheItem.php` file is changed

after this PR:
running `time php bin/phpstan analyze` reports `OptimizedDirectorySourceLocatorFactory->findSymbols` beeing invoked..
- 4248x when no cache exists
- 0x when no files changed
- 2x when `src/Cache/CacheItem.php` file is changed

I think this will fix https://github.com/phpstan/phpstan/discussions/13852

---

looking at the windows CI builds it seems those are ~20 seconds faster compared to other currently open PRs

ondrejmirtes (changes requested):
Yeah I think this is an obvious improvement. I also wanted to suggest to share the common code betweet createByFiles and createByDirectory which you already did.

I have an idea for improvement. `createByFiles` is called from a single place. I think we should add a 2nd parameter `string $uniqueCacheIdentifier` which will identify the entry in the cache.

That will allow us for the identifier to become the cache key. Your current approach has a downside - when a file is removed/added, the whole cache has to go. But with a stable identifier, we'll only replace the cached parts which need replacing.

ondrejmirtes:
Thank you!

PR #4578, opened 2025-11-26
Fix uopz_add_function() return type

zonuexe:
Fix my mistake in https://github.com/phpstan/phpstan-src/pull/482.

See <https://www.php.net/uopz_add_function>.


ondrejmirtes:
Thank you!

PR #4579, opened 2025-11-26
Fix subtype check for identical late-resolvable conditional types (bug #10942)

yt-catpaw:
closes https://github.com/phpstan/phpstan/issues/10942

・Summary

Treat identical late-resolvable conditional types as compatible before resolving them.
Short-circuit when the same conditional appears inside a union.
Add regression test for bug #10942.

・Testing

vendor/bin/phpunit tests/PHPStan/Rules/Methods/MethodSignatureRuleTest.php --filter testBug10942

### DIFF
--- a/src/Type/Traits/LateResolvableTypeTrait.php
+++ b/src/Type/Traits/LateResolvableTypeTrait.php
@@ -67,6 +67,18 @@ trait LateResolvableTypeTrait
 			return IsSuperTypeOfResult::createYes();
 		}
 
+		if ($this->equals($type)) {
+			return IsSuperTypeOfResult::createYes();
+		}
+
+		if ($type instanceof UnionType) {
+			foreach ($type->getTypes() as $innerType) {
+				if ($this->equals($innerType)) {
+					return IsSuperTypeOfResult::createYes();
+				}
+			}
+		}
+
 		if ($type instanceof LateResolvableType) {
 			$type = $type->resolve();
 		}
--- a/src/Type/Traits/LateResolvableTypeTrait.php
+++ b/src/Type/Traits/LateResolvableTypeTrait.php
@@ -19,6 +19,7 @@ use PHPStan\Type\IsSuperTypeOfResult;
 use PHPStan\Type\LateResolvableType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
+use PHPStan\Type\UnionType;
 
 trait LateResolvableTypeTrait
 {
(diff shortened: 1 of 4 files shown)


src/Type/Traits/LateResolvableTypeTrait.php
```diff
@@ -570,6 +571,18 @@ public function tryRemove(Type $typeToRemove): ?Type
 
 	public function isSubTypeOf(Type $otherType): IsSuperTypeOfResult
 	{
+		if ($this->equals($otherType)) {
```
ondrejmirtes, line 592: It'd make sense to write some tests that test this method directly. Similar to this: https://github.com/phpstan/phpstan-src/blob/43e8099812359b74ed4e4a89a19bb13639a64a15/tests/PHPStan/Type/UnionTypeTest.php#L174-L473

Also, please review and maybe fix the isSuperTypeOf method as well.

ondrejmirtes (changes requested):

src/Type/Traits/LateResolvableTypeTrait.php
yt-catpaw: Implemented; please take a look

https://github.com/phpstan/phpstan-src/pull/4579/commits/0ba680581efd69de144e250733366d5ddb95920c

tests/PHPStan/Type/LateResolvableTypeTraitTest.php
```diff
@@ -0,0 +1,34 @@
...
+		);
+
+		$this->assertSame('Yes', $conditional->isSuperTypeOf($conditional)->describe());
+
+		$unionWithConditional = new UnionType([
+			new StringType(),
+			$conditional,
+		]);
+
+		$this->assertSame('Yes', $conditional->isSuperTypeOf($unionWithConditional)->describe());
```
ondrejmirtes: Please make a test with a data provider as it looks like in the test I linked. Then I'l try to come up with some data points that break your logic 😊

ondrejmirtes: Also write tests for both isSuperTypeOf and isSubtypeOf. Thanks!

ondrejmirtes (changes requested):

tests/PHPStan/Type/LateResolvableTypeTraitTest.php
yt-catpaw: https://github.com/phpstan/phpstan-src/pull/4579/commits/28bd4ecf2d090ca7b0057a72dcb462c94707f81e

Added data-provider tests for isSuperTypeOf/isSubTypeOf (in tests/PHPStan/Type/LateResolvableTypeTraitTest.php).

VincentLanglet:
Solved by https://github.com/phpstan/phpstan-src/pull/5645

PR #4580, opened 2025-11-27
Fix get_object_vars() return type

Triplkrypl:
I found in crash way, that `get_object_vars` can return `array` with `int` key.

https://onlinephp.io/c/9a628

### DIFF
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -3315,7 +3315,7 @@ return [
 'get_magic_quotes_gpc' => ['false'],
 'get_magic_quotes_runtime' => ['false'],
 'get_meta_tags' => ['array|false', 'filename'=>'string', 'use_include_path='=>'bool'],
-'get_object_vars' => ['array<string, mixed>', 'obj'=>'object'],
+'get_object_vars' => ['array<array-key, mixed>', 'obj'=>'object'],
 'get_parent_class' => ['class-string|false', 'object='=>'mixed'],
 'get_required_files' => ['list<string>'],
 'get_resource_type' => ['string', 'res'=>'resource'],


staabm:
Can numeric keys only occur for objects without a class?

If so it might make sense to use a return type extension instead which can narrow the type depending on the context.

edit: and I think you should target PHPStan 2.1.x

Triplkrypl:
> Can numeric keys only occur for objects without a class?

Not exactly classes extending `\stdClass` or has `#[\AllowDynamicProperties]` attribute.
Fore the classes is this code valid and before deprecation is code valid for every object:

```php
$obj = new \stdClass();
$obj->{1} = 42;
```

If anyone want make dynamic return type detection, condition only for class is not enough.
For now in PHP 8 every object can have only numeric property, it is only deprecation error:
https://onlinephp.io/c/9704f

> and I think you should target PHPStan 2.1.x

I am not sure if you support fixes for function interfaces in phpstan version 1.

VincentLanglet:
> I am not sure if you support fixes for function interfaces in phpstan version 1.

Is there a reason you still use version 1?

Triplkrypl:
> > I am not sure if you support fixes for function interfaces in phpstan version 1.
> 
> Is there a reason you still use version 1?

I not use version 1 what i know and not need it, but my question is. If you make fixes of PHP function interfaces in version 1, if not i will update pull request to version 2.

staabm:
At this point in time 1.x only sees critical fixes. 

Everything else goes into 2.1.x right now

Triplkrypl:
https://github.com/phpstan/phpstan-src/pull/4629

PR #4581, opened 2025-11-28
PHP 8.4 ext-dba function signatures

kassner:
dba_* functions had their signature changed in PHP 8.4 to return  a \Dba\Connection instance.


ondrejmirtes:
Please share phpstan.org/try with an error first.

kassner:
https://phpstan.org/r/6ad959ee-2851-4447-bb1c-90a511f93445

> Parameter #2 $handle of function dba_fetch expects resource, Dba\Connection|false given.

ondrejmirtes:
@kassner I think the right solution is to put `dba_fetch\'1` key into `old` section in `resources/functionMap_php83delta.php`.

1) This signature is deprecated in PHP 8.3.
2) The problem is that PHPStan is thrown off by the extra variant. You could look for the root cause and fix it (the fact that information from phpstan/php-8-stubs is not used there), but the easy fix is to put `dba_fetch\'1` key into `old` section in `resources/functionMap_php83delta.php`.

The problem here is that phpstan/php-8-stubs info is not applied to dba_fetch.

ondrejmirtes:
I just pushed the removal line you added: https://github.com/phpstan/phpstan-src/commit/55d0fd7f3c036da5c28eddd94b5e1462a4cfcbd9

The rest is not necessary. If you still experience errors with dba_ functions, please open an issue.

PR #4585, opened 2025-11-29
Implement BinaryConcatHandler

staabm:


ondrejmirtes:
Thank you!

FYI just noticed that Mod and Div have a special handling in NSR: https://github.com/phpstan/phpstan-src/blob/79aedb0b3cbd9c190ba85194acd9cca0264b3413/src/Analyser/NodeScopeResolver.php#L3569-L3574

So the new handlers should also have that.

PR #4590, opened 2025-11-29
Fix "Array does not have offset" in curl_setopt_array()

staabm:
closes https://github.com/phpstan/phpstan/issues/13862


ondrejmirtes:
Thank you!

PR #4593, opened 2025-11-30
Implement BinaryShiftLeftHandler, BinaryShiftRightHandler

staabm:


ondrejmirtes:
Thank you! BTW `$context->enterDeep()` was missing in all the other handlers as well so I added it. I forgot to add it in the BinaryPlus handler I showed you 😊

PR #4595, opened 2025-11-30
Implement Cast*Handler

staabm:


tests/PHPStan/Analyser/Generator/data/gnsr.php
```diff
@@ -236,6 +236,17 @@ public function doSpaceship($a, $b, string $c, string $d): void
 		assertNativeType('int<-1, 1>', $c <=> $d);
 	}
 
+	function doCast() {
+		$a = '1';
+
+		assertType('1', (int) $a);
+		assertType("array{'1'}", (array) $a);
+		// assertType("array{'1'}", (object) $a);
```
staabm, line 244: left out `(object)` because its more involved in https://github.com/phpstan/phpstan-src/blob/75f223eed4ccfce0eb9022b3db98885aac401365/src/Reflection/InitializerExprTypeResolver.php#L705-L730

do we need to implement `Type->toObject()` first, or is it fine to delegate into `InitializerExprTypeResolver` for `(object)` cast?

staabm: another alternative would be to only have a single CastHandler and just call `InitializerExprTypeResolver->getCastTypeFromType`. I did not go this route, as you already implemented `CastIntHandler` without the use of `InitializerExprTypeResolver`

ondrejmirtes: My thinking: looking at InitializerExprTypeResolver::getCastType(), everything aside from `Cast\Object_` is trivial so we don't need to call InitializerExprTypeResolver from other resolvers.

For `Cast\Object_`, I'd do a new `InitializerExprTypeResolver::getCastObjectType()` which we'd call from the new handler and also from existing place in InitializerExprTypeResolver.

ondrejmirtes (changes requested):
One more thing - these expressions (Bool_, String_, Int_, Double) have some kind of logic in LegacyTypeSpecifier:

```
		} elseif ($expr instanceof Expr\Cast\Bool_) {
			return $this->specifyTypesInCondition(
				$scope,
				new Node\Expr\BinaryOp\Equal($expr->expr, new ConstFetch(new Name\FullyQualified('true'))),
				$context,
			)->setRootExpr($expr);
		} elseif ($expr instanceof Expr\Cast\String_) {
			return $this->specifyTypesInCondition(
				$scope,
				new Node\Expr\BinaryOp\NotEqual($expr->expr, new Node\Scalar\String_('')),
				$context,
			)->setRootExpr($expr);
		} elseif ($expr instanceof Expr\Cast\Int_) {
			return $this->specifyTypesInCondition(
				$scope,
				new Node\Expr\BinaryOp\NotEqual($expr->expr, new Node\Scalar\LNumber(0)),
				$context,
			)->setRootExpr($expr);
		} elseif ($expr instanceof Expr\Cast\Double) {
			return $this->specifyTypesInCondition(
				$scope,
				new Node\Expr\BinaryOp\NotEqual($expr->expr, new Node\Scalar\DNumber(0.0)),
				$context,
			)->setRootExpr($expr);
```

They influence this logic and similar: https://phpstan.org/r/7dbcb4be-6c01-4bf2-b312-450327d887b6 (try it with `mixed` instead of `int` too)

GNSR handlers work a bit differently - ExprAnalysisResult carries $specifiedTruthyTypes / $specifiedFalseyTypes / $specifiedNullTypes. If I take CastBoolHandler, it should do:

```
$specifiedExprResult = yield new ExprAnalysisRequest($stmt, new Node\Expr\BinaryOp\Equal($expr->expr, new ConstFetch(new Name\FullyQualified('true'))), ...)
```

The alternativeNodeCallback of this ExprAnalysisRequest should be `new NoopNodeCallback()`!

And then return `specifiedTruthyTypes: $specifiedExprResult->specifiedTruthyTypes` in its own ExprAnalysisResult.

I didn't verify this idea but I think this is how it should work.

ondrejmirtes:
Oops, I realized this will force you to implement EqualHandler and NotEqualHandler as well 😂

staabm:
ok, will try to implement `EqualHandler` and `NotEqualHandler` in a separate PR, when I am back at a desk with more screens :)

ondrejmirtes:
Thank you!

PR #4596, opened 2025-11-30
Implement CastObjectHandler

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/4595

type-specifying parts will be a followup

---

[reviewed best with whitespaces ignored](https://github.com/phpstan/phpstan-src/pull/4596/files?w=1)


ondrejmirtes:
Thank you!

PR #4598, opened 2025-12-01
Implement UnaryMinusHandler

staabm:
as in
- https://github.com/phpstan/phpstan-src/blob/61b221ca7a74ac55f5df52b3a7a7b6ef263ecc51/src/Reflection/InitializerExprTypeResolver.php#L2523-L2551
- https://github.com/phpstan/phpstan-src/blob/61b221ca7a74ac55f5df52b3a7a7b6ef263ecc51/src/Analyser/NodeScopeResolver.php#L3626-L3640

### DIFF
--- a/src/Analyser/Generator/ExprHandler/UnaryMinusHandler.php
+++ b/src/Analyser/Generator/ExprHandler/UnaryMinusHandler.php
@@ -1,0 +1,51 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Analyser\Generator\ExprHandler;
+
+use Generator;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Stmt;
+use PHPStan\Analyser\ExpressionContext;
+use PHPStan\Analyser\Generator\ExprAnalysisRequest;
+use PHPStan\Analyser\Generator\ExprAnalysisResult;
+use PHPStan\Analyser\Generator\ExprHandler;
+use PHPStan\Analyser\Generator\GeneratorScope;
+use PHPStan\Analyser\SpecifiedTypes;
+use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Reflection\InitializerExprTypeResolver;
+
+/**
+ * @implements ExprHandler<Expr\UnaryMinus>
+ */
+#[AutowiredService]
+final class UnaryMinusHandler implements ExprHandler
+{
+
+	public function __construct(private InitializerExprTypeResolver $initializerExprTypeResolver)
+	{
+	}
+
+	public function supports(Expr $expr): bool
+	{
+		return $expr instanceof Expr\UnaryMinus;
+	}
+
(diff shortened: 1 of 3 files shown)


src/Analyser/Generator/ExprHandler/UnaryMinusHandler.php
```diff
@@ -0,0 +1,62 @@
...
+	public function analyseExpr(Stmt $stmt, Expr $expr, GeneratorScope $scope, ExpressionContext $context, ?callable $alternativeNodeCallback): Generator
+	{
+		$result = yield new ExprAnalysisRequest($stmt, $expr->expr, $scope, $context->enterDeep(), $alternativeNodeCallback);
+
+		$type = $this->initializerExprTypeResolver->getUnaryMinusTypeFromType($expr->expr, $result->type);
+		$nativeType = $this->initializerExprTypeResolver->getUnaryMinusTypeFromType($expr->expr, $result->nativeType);
+		if ($type instanceof IntegerRangeType) {
+			$mulResult = yield new ExprAnalysisRequest($stmt, new Expr\BinaryOp\Mul($expr, new Int_(-1)), $scope, $context->enterDeep(), new NoopNodeCallback());
+			$type = $mulResult->result;
+			$nativeType = $mulResult->nativeType;
```
staabm: not sure why PHPStan reports

```
 ------ ----------------------------------------------------------------------------------------- 
  Line   src/Analyser/Generator/ExprHandler/UnaryMinusHandler.php                                 
 ------ ----------------------------------------------------------------------------------------- 
  44     Access to an undefined property PHPStan\Analyser\Generator\ExprAnalysisResult::$result.  
         🪪  property.notFound                                                                    
         💡  Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property    
         at src/Analyser/Generator/ExprHandler/UnaryMinusHandler.php:44                           
 ------ -----------------------------------------------------------------------------------------
 ```

here but it does not on line 40.

ondrejmirtes: I'm not sure what you mean. The reported lines are sometimes shifted because code is downgraded on lower PHP versions.

But `$mulResult->result` is really an error. It's supposed to be `$mulResult->type`?

staabm: ha, seems it was too early this morning ;).

hopefully we can make improvements so phpstorm autocomplete and type-discovery work better with this fiber code. atm it feels kind of unsupported.

are you aware whether there are already open issues on the phpstorm bugtracker to improve on type stuff with Fibers?

ondrejmirtes: In the handlers it's not about fibers but about generators. I think it's hard and no syntax was yet developed for this. We'd need something for `@return Generator<...` that would tell PHPStan + PhpStorm "when this is yielded, this is sent".

Psalm has `@psalm-yield` annotation but it doesn't make sense to me. Because what is sent depends on context. For the same yielded objects, different things can get sent to it in different generators.

ondrejmirtes:
Thank you!

PR #4599, opened 2025-12-01
Implement UnaryPlusHandler

staabm:
as in
- https://github.com/phpstan/phpstan-src/blob/6cf0ab29ce182d8cbf01d21d5ea886893bdb1a64/src/Analyser/MutatingScope.php#L1129-L1131
- https://github.com/phpstan/phpstan-src/blob/6cf0ab29ce182d8cbf01d21d5ea886893bdb1a64/src/Analyser/NodeScopeResolver.php#L3626-L3640


ondrejmirtes:
Thank you!

PR #4600, opened 2025-12-01
Implement InterpolatedStringHandler

staabm:
as in
- https://github.com/phpstan/phpstan-src/blob/6cf0ab29ce182d8cbf01d21d5ea886893bdb1a64/src/Analyser/MutatingScope.php#L1262-L1276
- https://github.com/phpstan/phpstan-src/blob/6cf0ab29ce182d8cbf01d21d5ea886893bdb1a64/src/Analyser/NodeScopeResolver.php#L3399-L3414

### DIFF
--- a/src/Analyser/Generator/ExprHandler/InterpolatedStringHandler.php
+++ b/src/Analyser/Generator/ExprHandler/InterpolatedStringHandler.php
@@ -1,0 +1,90 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Analyser\Generator\ExprHandler;
+
+use Generator;
+use PhpParser\Node\Expr;
+use PhpParser\Node\InterpolatedStringPart;
+use PhpParser\Node\Scalar\InterpolatedString;
+use PhpParser\Node\Stmt;
+use PHPStan\Analyser\ExpressionContext;
+use PHPStan\Analyser\Generator\ExprAnalysisRequest;
+use PHPStan\Analyser\Generator\ExprAnalysisResult;
+use PHPStan\Analyser\Generator\ExprHandler;
+use PHPStan\Analyser\Generator\GeneratorScope;
+use PHPStan\Analyser\Generator\NoopNodeCallback;
+use PHPStan\Analyser\SpecifiedTypes;
+use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Reflection\InitializerExprTypeResolver;
+use PHPStan\Type\Constant\ConstantStringType;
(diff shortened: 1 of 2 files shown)


src/Analyser/Generator/ExprHandler/InterpolatedStringHandler.php
```diff
@@ -0,0 +1,90 @@
...
+		$resultType = null;
+		$hasYield = false;
+		$throwPoints = [];
+		$impurePoints = [];
+		$isAlwaysTerminating = false;
+		foreach ($expr->parts as $part) {
+			if ($part instanceof InterpolatedStringPart) {
+				$partType = new ConstantStringType($part->value);
+			} else {
+				$result = yield new ExprAnalysisRequest($stmt, $part, $scope, $context->enterDeep(), new NoopNodeCallback());
```
staabm: is noop-correct here?

ondrejmirtes: I don't think so. It'd mean that for `$part` no rule is executed. `NoopNodeCallback` is the equivalent of `static function () {}` when calling `->processExprNode` or `->processStmtNode` in NodeScopeResolver.

So it's used when we want a result of a synthetic node (where we do `new SomeAstNode(...)` because we're interested in the result but otherwise the node is not in the original source so it would not make sense to report errors on it).

Also we'd use `NoopNodeCallback` when we need ExprAnalysisRequest for something that we're sure has already been analysed. Thanks to https://github.com/phpstan/phpstan-src/blob/3bd2d8bf0214813d1f7bef917c2f13056c9591af/src/Analyser/Generator/GeneratorNodeScopeResolver.php#L470-L472 it'd just return the result instead of throwing an exception.

ondrejmirtes (changes requested):

src/Analyser/Generator/ExprHandler/InterpolatedStringHandler.php
```diff
@@ -0,0 +1,90 @@
...
+				continue;
+			}
+
+			$resultType = $this->initializerExprTypeResolver->resolveConcatType($resultType, $partType);
+		}
+
+		$type = $resultType ?? new ConstantStringType('');
+		return new ExprAnalysisResult(
+			$type,
+			$type,
```
ondrejmirtes: What about native type here? :)

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4603, opened 2025-12-01
Narrow types after cast in generator based Handlers

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/4595#pullrequestreview-3521183186

see

https://github.com/phpstan/phpstan-src/blob/d7482afc58d503ca9ae5cccf9497ec078f6b9228/src/Analyser/LegacyTypeSpecifier.php#L198-L221


staabm:
meanwhile implemented in https://github.com/phpstan/phpstan-src/commit/edfabec01eb437c579ef29c75811d7bccf2ff50d

PR #4604, opened 2025-12-01
Implement remaining Magic-Const Handlers

staabm:


ondrejmirtes (changes requested):
An idea - these are exactly, but absolutely exactly the same. We could have a single class that has multiple `instanceof ... || instanceof ...` in `supports`.

ondrejmirtes:
Thank you!

BTW I wouldn't refactor the others :)

PR #4606, opened 2025-12-01
Fix ReflectionClassWithLazyObjects ReflectionClass stub

DaveLiddament:
I think this stub should match: https://github.com/phpstan/phpstan-src/blob/2.1.x/stubs/ReflectionClass.stub#L4C13-L4C23

See this commit for where the lazy object was introduced: https://github.com/phpstan/phpstan-src/commit/25ec5eb2db6474f95a0b9dcac1cca8210ea1a021#diff-34b8ba0664529e6e59d2c54a9a4cf3194fe2937582419c8d71516783474f4dc6

If you agree I think the same is needed for Enum. I can make a PR for that too.

### DIFF
--- a/stubs/ReflectionClassWithLazyObjects.stub
+++ b/stubs/ReflectionClassWithLazyObjects.stub
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * @template T of object
+ * @template-covariant T of object
  */
 class ReflectionClass
 {


ondrejmirtes:
Separate`stubs/ReflectionClassWithLazyObjects.stub` exists exactly because we can't make it `@template-covariant`. 

Because when you do something like that, you get this error:

```
Template type T is declared as covariant, but occurs in contravariant position in parameter object of method Foo::initializeLazyObject().
```

It's explained here: https://phpstan.org/blog/whats-up-with-template-covariant

If you want to see me talk about it, here's a video with a timestamp (https://youtu.be/eRvH9CP4YfY?si=ALy2sEKSqnSYLFbW&t=1818 - 30:18)

DaveLiddament:
I've asked a question about this here: https://github.com/phpstan/phpstan/discussions/13891

PR #4361, opened 2025-09-22
Fix DOMNode::C14N bad definition

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/8096

PR #4608, opened 2025-12-02
Fix return-type for DOMNode::C14N

tvdijen:
Discussion: https://github.com/phpstan/phpstan/discussions/13873
Doc reference: https://www.php.net/manual/en/domnode.c14n.php


tvdijen:
The failing tests seem unrelated to this one-liner change?

ondrejmirtes:
Thank you!

VincentLanglet:
This method is not fully fixed (and C14NFile has the same issue) ; cf https://github.com/phpstan/phpstan-src/pull/4361

PR #4609, opened 2025-12-03
Fix time not being considered an impure function.

mad-briller:
Fixes https://github.com/phpstan/phpstan/issues/13874 by marking `time()` as being impure.


bin/functionMetadata_original.php
```diff
@@ -117,6 +117,7 @@
 	'str_decrement' => ['hasSideEffects' => false],
 	'str_increment' => ['hasSideEffects' => false],
 	'symlink' => ['hasSideEffects' => true],
+	'time' => ['hasSideEffects' => true],
```
staabm: do we need the same for `microtime()` and/or `date()`?

ondrejmirtes:
Thank you!

PR #4610, opened 2025-12-03
Don't report already assigned errors when setting or unsetting an offset on array access objects set as class properties.

mad-briller:
fixes https://github.com/phpstan/phpstan/issues/13856

Took inspiration from https://github.com/phpstan/phpstan-src/pull/3817 when implementing the fix.

Not entirely 100% sure about the proposed changes here, as passing through the `$originalNode` to the `PropertyWrite` constructor feels a little heavy handed, considering i only need one of the 3 types that it could be, and also it has to be null because of the parent constructor case, so open to suggestions there.

### DIFF
--- a/src/Node/ClassPropertiesNode.php
+++ b/src/Node/ClassPropertiesNode.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Node;
 
+use ArrayAccess;
 use Override;
 use PhpParser\Node;
 use PhpParser\Node\Expr\Array_;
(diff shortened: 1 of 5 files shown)


mad-briller:
i have no intuition on how to deal with the escaped mutants mentioned by CI :angel:, if someone could give me a pointer that'd be appreciated

ondrejmirtes (changes requested):
This is good, thank you!

The escaped mutant - you have a test for a subtype of ArrayAccess. You're checking `$type->isSuperTypeOf(ArrayAccess)->yes()` (A bit of pseudocode from my side there).

You need to ask - what type would produce true for `!$type->isSuperTypeOf(ArrayAccess)->no()` that I'm not testing?

And the answer is for example `object` type or `mixed` type.

ondrejmirtes:
Also please rebase the branch, there's a conflict in the test case.

mad-briller:
i've added a test case that makes mutation testing not fail, but the test case i've added i'm unsure about because it's still of the right type in the constructor.

if this is not what you had in mind let me know and i can rework

PR #4611, opened 2025-12-04
Narrow CURLOPT_SHARE accepting type

staabm:
since PHP8.5 there is also `CurlSharePersistentHandle` created via [curl_share_init_persistent](https://www.php.net/curl_share_init_persistent)

### DIFF
--- a/src/Reflection/ParametersAcceptorSelector.php
+++ b/src/Reflection/ParametersAcceptorSelector.php
@@ -58,6 +58,7 @@ use function is_string;
 use function sprintf;
 use const ARRAY_FILTER_USE_BOTH;
 use const ARRAY_FILTER_USE_KEY;
+use const CURLOPT_SHARE;
 use const CURLOPT_SSL_VERIFYHOST;
 
 /**
@@ -1210,6 +1211,14 @@ final class ParametersAcceptorSelector
 			if (defined($constName) && constant($constName) === $curlOpt) {
 				return new ResourceType();
 			}
+		}
+
+		if ($curlOpt === CURLOPT_SHARE) {
+			return new UnionType([
+				new ResourceType(), // PHP 7.x
+				new ObjectType('CurlShareHandle'), // since PHP 8.0
+				new ObjectType('CurlSharePersistentHandle'), // since PHP 8.5
+			]);
 		}
 
 		// unknown constant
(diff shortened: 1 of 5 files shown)


src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -1213,6 +1214,14 @@ private static function getCurlOptValueType(int $curlOpt): ?Type
 			}
 		}
 
+		if ($curlOpt === CURLOPT_SHARE) {
+			return new UnionType([
+				new ResourceType(), // PHP 7.x
+				new ObjectType('CurlShareHandle'), // since PHP 8.0
+				new ObjectType('CurlSharePersistentHandle'), // since PHP 8.5
```
staabm: we don't have access to `PhpVersion` at this point, therefore I hardcoded all possible types.
I think it doesn't really matter to narrow this further, because the values will be created by php-src builtin functions, so cannot be wrongly instantiated at runtime.

src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -1212,6 +1213,14 @@ private static function getCurlOptValueType(int $curlOpt): ?Type
...
```
ondrejmirtes: You can ask PhpVersionStaticAccessor.

ondrejmirtes (changes requested):

src/Reflection/ParametersAcceptorSelector.php
staabm: adjusted, thanks

ondrejmirtes:
Thank you!

src/Reflection/ParametersAcceptorSelector.php
```diff
@@ -1212,6 +1213,21 @@ private static function getCurlOptValueType(int $curlOpt): ?Type
 			}
 		}
 
+		if ($curlOpt === CURLOPT_SHARE) {
```
schlndh, line 1216: Minor issue: this crashes due to `Undefined constant "CURLOPT_SHARE"` when analysing `curl_setopt` without curl extension. It seems that the rest of the method uses `defined($constName) && constant($constName) === $curlOpt`.

E.g. just this is enough `curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);` to trigger a crash if the curl extension is disabled.

staabm: Please send a fix

schlndh: https://github.com/phpstan/phpstan-src/pull/4722

PR #4614, opened 2025-12-04
More precise return type for gatherAssertTypes

staabm:


tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -270,23 +269,24 @@ public function testFile(string $file): void
 		$failures = [];
 
 		foreach ($asserts as $args) {
-			$assertType = array_shift($args);
-			$file = array_shift($args);
+			if ($args[0] === 'type') {
+				$file = $args[1];
 
-			if ($assertType === 'type') {
```
staabm, line 276: phpstan is not able to narrow the array-shape of the tagged union of $args based on `array_shift()` (which is equivalent to `$args[0]`).
rewrote the logic so PHPStan is able to narrow the tagged union

src/Testing/TypeInferenceTestCase.php
```diff
@@ -220,8 +220,13 @@ public function assertFileAsserts(
 	}
 
 	/**
+	 * @return array<string, (
```
staabm, line 223: goal of this PR is, to have a more precise return type here, so downstream projects which use `TypeInferenceTestCase` can properly utilize the PHPUnit Data-Provider validation capabilities on level 8

see e.g. https://github.com/phpstan/phpstan-webmozart-assert/pull/193

staabm:
I think with this change, we can now put level 9 back onto phpstan/webmozarts-assert without errors.

it was a chicken egg problem, which needs adjustments in in both repos, to get finally green again

src/Testing/TypeInferenceTestCase.php
```diff
@@ -390,15 +395,15 @@ public static function gatherAssertTypes(string $file): array
 		}
 
 		if (count($delayedErrors) === 0) {
-			return $asserts;
+			return $asserts; // @phpstan-ignore return.type
```
ondrejmirtes: That's not really a solution I was looking for 😊

ondrejmirtes (changes requested):
I copied the PHAR from here and it makes phpstan-webmozart-assert green on L9 even without removing the PHPDocs there. That's good. But PHPStan shouldn't throw any errors here. We should know when the method returns something bad so ignoring isn't an option.

src/Testing/TypeInferenceTestCase.php
staabm: of course not. but that needs fixing a bug unrelated to the change/goal of this PR here ;)

staabm: ~~to fix this, we need to resolve: https://phpstan.org/r/14f80375-8f26-4720-a551-5db44aaca71c~~

staabm: fixed the wrong return-type

ondrejmirtes:
Awesome, thank you!

PR #4615, opened 2025-12-04
Add ArrayCountValuesDynamicReturnTypeExtension.

mad-briller:
Closes https://github.com/phpstan/phpstan/issues/13816

The issue states that stringable objects are accepted, but it turns out that this is not the case:
https://3v4l.org/07vKm#v8.4.15

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/array-count-values.php
+++ b/tests/PHPStan/Analyser/nsrt/array-count-values.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace ArrayCountValues;
+
+use function PHPStan\Testing\assertType;
+
+$ints = array_count_values([1, 2, 2, 3]);
+
+assertType('array<1|2|3, int<1, max>>', $ints);
+
+$strings = array_count_values(['one', 'two', 'two', 'three']);
+
+assertType('array<\'one\'|\'three\'|\'two\', int<1, max>>', $strings);
+
+$objects = array_count_values([new \stdClass()]);
+
+assertType('array{}', $objects);
+
+/**
+ * @return array<int, string|object>
+ */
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/nsrt/array-count-values.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace ArrayCountValues;
+
+use function PHPStan\Testing\assertType;
+
+$ints = array_count_values([1, 2, 2, 3]);
+
+assertType('array<1|2|3, int<1, max>>', $ints);
```
staabm: could we get more precise?

```suggestion
assertType('non-empty-array<1|2|3, int<1, max>>', $ints);
```

mad-briller: good shout, sorted

mad-briller:
shopware ci failure looks expected as they are calling `array_keys` on `mixed` which will result in `int|string`

ondrejmirtes:
Thank you!

tests/PHPStan/Analyser/nsrt/array-count-values.php
VincentLanglet: We could be even be more precise with one of those
```
array{1: int<1, max>, 2: int<1, max>, 3: int<1, max>}
```

cause `non-empty-array<1|2|3, int<1, max>>` might report 1,2 and 3 offsets as "maybe missing".

Do you think it's worth it ?

VincentLanglet:
I think I found two bugs with the current implementation, I opened https://github.com/phpstan/phpstan-src/pull/4618 about it @ondrejmirtes

PR #4618, opened 2025-12-07
Fix ArrayCountValuesDynamicReturnTypeExtension

VincentLanglet:
`$type->toArrayKey() instanceof ErrorType` is not the right check cause `true, false, 1.0, null, ...` are ignored by the function. It has to be an int or a string.

And numeric string wasn't casted into int when used as array key.


staabm:
Looks like it should be $allowedValues->isSuperTypeOf(...)

VincentLanglet:
> Looks like it should be $allowedValues->isSuperTypeOf(...)

How you deal with something like "int|null" ?
I dont see the benefit with keeping the unionType check and calling filterType on it rather than doing an intersection like I did.

staabm:
I see now, that you also need the normalized result. its not just a check for compatibility. my suggestion did not fit that.

mad-briller:
good catches, didn't think to try other scalar types :+1:

ondrejmirtes:
Thank you!

PR #4626, opened 2025-12-09
Tests: Use Stubs instead of Mocks

staabm:
No expectations were configured for the mock object for XY. You should refactor your test code and use a test stub instead.


ondrejmirtes:
Thank you!

PR #4627, opened 2025-12-11
Remove path normalizing from NodeDependencies

staabm:
in NodeDependencies the same file paths are normalized over and over, as can be seen in blackfire profiles

grafik


ondrejmirtes:
Well I'm sure it's there for a reason. Try to diff the result cache file before/after.

PR #4628, opened 2025-12-11
FNSR ExpressionResultStorage

ondrejmirtes:


staabm:
issue bot results look outstanding  - too good to be real 🥸

ondrejmirtes:
Lots of diffs doesn't mean lots of bugs fixed 😀 The analysis is still too wrong.

ondrejmirtes:
Alright, so this issue-bot report is far more realistic and closer to the end results. Solving issues like https://phpstan.org/r/8d3536cc-d647-455c-8919-300cc35d7abf and https://phpstan.org/r/59b9367b-8f29-4008-8536-91dca00362df is what I had in mind.

ondrejmirtes:
I’m sure it’s used. It overrides an empty method from NodeScopeResolver.


Ondřej Mirtes

src/Analyser/NodeScopeResolver.php
```diff
@@ -3684,23 +3858,24 @@ static function (): void {
...
 				$hasYield = $hasYield || $result->hasYield();
 				$throwPoints = array_merge($throwPoints, $result->getThrowPoints());
 				$impurePoints = array_merge($impurePoints, $result->getImpurePoints());
 				$isAlwaysTerminating = $isAlwaysTerminating || $result->isAlwaysTerminating();
 			} else {
-				$nodeCallback($expr->name, $scope);
+				$this->callNodeCallback($nodeCallback, $expr->name, $scope, $storage);
 			}
 		} elseif ($expr instanceof Expr\Empty_) {
+			$this->processExprNode($stmt, $expr->expr, $scope, $storage, new NoopNodeCallback(), $context->enterDeep());
```
staabm, line 3880: why/which expressions need to be processed twice? I can see this only happens for a few ones

src/Node/DeepNodeCloner.php
```diff
@@ -0,0 +1,29 @@
...
+{
+
+	/**
+	 * @template T of Node
+	 * @param T $node
+	 * @return T
+	 */
+	public function cloneNode(Node $node): Node
+	{
+		$traverser = new NodeTraverser(new CloningVisitor());
```
staabm, line 21: I can see we sometimes clone nodes which have already been cloned.
(so we clone the clone)

grafik

it might make sense to prevent some unnecessary per node work on this area

staabm: it seems occasionally we even clone the cloned clone.

grafik

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Yeah, this was tricky. The problem is that the new system is so annoyingly precise it actually sometimes prevents a rule what it wants to do. When doing `$scope->getType($expr)`, it actually no longer matters what `$scope` is. It just waits for NodeScopeResolver to reach and process `$expr` before giving us the result.

This is best explained on NullsafePropertyFetch and NullsafeMethodCall which also need to do this. NullsafePropertyFetchRule needs to report when doing `$var?->doFoo()` on non-nullable `$expr`. But with this code in NodeScopeResolver:

```php
			$exprResult = $this->processExprNode($stmt, new PropertyFetch(
				$expr->var,
				$expr->name,
				array_merge($expr->getAttributes(), ['virtualNullsafePropertyFetch' => true])
			), $nonNullabilityResult->getScope(), $storage, $nodeCallback, $context);
```

The `$expr->var` is always non-nullable so the rule cannot do its job.

That's why I decided to have different Expr instances for when you have a rule for NullsafePropertyFetch (the original node) and when you have a rule for a synthetic `new PropertyFetch`. So both NullsafePropertyFetch rules (var is still nullable) and PropertyFetch rules  (var is no longer nullable) can do their job.

src/Node/DeepNodeCloner.php
ondrejmirtes: I can imagine this might happen on nested calls like `$this->foo?->bar?->doBar()`. And it's not really preventable.

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Before FNSR this wasn't an issue because the `$scope` passed to a rule had different types so the rules worked fine even when `$var` was the same instance.

ondrejmirtes:
@staabm Alright, I have some performance profiles you could take a look at. 

Here's 2.1.x: https://blackfire.io/profiles/86fcac97-986a-4e41-92e0-b41aaf53fbcc/graph
Here's fnsr-storage without `PHPSTAN_FNSR` env variable: https://blackfire.io/profiles/cfa19ede-14e9-4f37-8961-d45d6f9b570d/graph (the difference should only be the fact we have ExpressionResultStorage with the results)
Here's fnsr-storage with `PHPSTAN_FNSR=1`: https://blackfire.io/profiles/972bb55f-97ad-4870-93be-bfed8f50ad3b/graph (the difference is that we run Fibers and we also store pending Fibers for later when the expr result is available)

Right now it's slower because of these reasons. Maybe some optimizations could make it faster. Feel free to send PRs against fnsr-storage, thanks :)

What I will personally look into is to how take advantage of ExpressionResultStorage to not process AST nodes multiple times, like it's done for BooleanAnd+BooleanOr in MutatingScope, and also for closure bodies.

Also I think we could maybe get rid of these properties in MutatingScope, at least when FiberScope is involved:

```php
	/** @var Type[] */
	private array $resolvedTypes = [];

	/** @var array<string, self> */
	private array $truthyScopes = [];

	/** @var array<string, self> */
	private array $falseyScopes = [];
```

ondrejmirtes:
Here's comparison `2.1-x -> fnsr-storage`: https://blackfire.io/profiles/compare/baf03475-a664-401f-a0df-0f0fd9b1c17b/graph
Here's comparison `fnsr-storage -> PHPSTAN_FNSR=1`: https://blackfire.io/profiles/compare/baf03475-a664-401f-a0df-0f0fd9b1c17b/graph (big increase in memory, presumably because of storing pending Fibers)

ondrejmirtes:
The build now should be green without any changes, just consuming a bit more memory on large files. I'm going to merge this, then I'm going to clean up the Generator namespace which is no longer going to be developed, then we're going to try to address the downsides of the updated implementation.

ondrejmirtes:
Once the build with `PHPSTAN_FNSR=1` is green on tests and PHPStan, I'm going to add more jobs to the matrix that test it too.

staabm:
> Also I think we could maybe get rid of these properties in MutatingScope, at least when FiberScope is involved:

so the idea is, since we already have a caching with `ExpressionResultStorage` in FiberScope context, we no longer need this caching in mutating scope when fibers are in use?

ondrejmirtes:
Yeah, something like that. Of course I could be wrong. For example I was first investigating excessive memory usage because of `ExpressionResultStorage::duplicate()` but it was totally wrong. The problem instead was in many Fibers in `pendingFibers` array and the fix instead was https://github.com/phpstan/phpstan-src/commit/4147d7e0eed5ef65b3ed8701c4a6fefdacb52545.

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Found a different approach https://github.com/phpstan/phpstan-src/commit/d09f75487b1bbb761fdbbdef98b29c58512a53d5

PR #4629, opened 2025-12-11
Fix get_object_vars() return type

Triplkrypl:
I created new pull request for: https://github.com/phpstan/phpstan-src/pull/4580 because i can not update source branch.


staabm:
Please add a test in the tests/PHPStan/Analyser/nsrt
 directory asserting the type of the example of your issue. 

For inspiration see other files in this folder. Just adding a new file with a assertType() should do

ondrejmirtes:
Thank you!

PR #4630, opened 2025-12-12
Constants isDeprecation() depends on Scope->getPhpVersion()

staabm:
tests in https://github.com/phpstan/phpstan-deprecation-rules/pull/170


ondrejmirtes:
Thank you!

PR #4632, opened 2025-12-12
Fix get defined vars return type

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13881

### DIFF
--- a/src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
+++ b/src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
@@ -35,6 +35,10 @@ final class GetDefinedVarsFunctionReturnTypeExtension implements DynamicFunction
 		$typeBuilder = ConstantArrayTypeBuilder::createEmpty();
 
 		foreach ($scope->getDefinedVariables() as $variable) {
+			if ($variable === 'this') {
+				continue;
+			}
+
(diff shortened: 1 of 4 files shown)


src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
```diff
@@ -35,6 +35,10 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 		$typeBuilder = ConstantArrayTypeBuilder::createEmpty();
 
 		foreach ($scope->getDefinedVariables() as $variable) {
+			if ($variable === 'this') {
```
staabm, line 38: just double checked: since regular variables cannot be named '$this' we don't need a inClassScope or similar check

https://3v4l.org/gSRUE#veol

---

its a open feature request that we don't error about such case atm: https://github.com/phpstan/phpstan/issues/3585

staabm (approved):

src/Type/Php/GetDefinedVarsFunctionReturnTypeExtension.php
staabm: I also though about moving this check directly into `$scope->getDefinedVariables()` but I would not do that because there might be extensions out there which depend on the behaviour that '$this' is returned atm

PR #4635, opened 2025-12-15
Refactor ComposerPhpVersionFactory

staabm:
extracts `ComposerPhpVersionParser` to allow re-use of version-string into min/max `PhpVersion` parsing.

I think this will be usefull in https://github.com/phpstan/phpstan-phpunit/pull/269 to build `PhpVersion` min/max for versions in PHPUnits `#[RequiresPhp('5.*')]`-attribute.


ondrejmirtes:
Thank you!

PR #4637, opened 2025-12-19
Prevent double normalization of FuncCall nodes

staabm:
In the profiles of https://github.com/phpstan/phpstan-src/pull/4628#issuecomment-3674032038 we can see `getDynamicFunctionReturnType` as one of the most expensive calls:

grafik

---

in the happy path in which `FuncCall` has a `Name` before this PR we tend to normalize the argument list twice.

this PR remove duplicate arg normalization per FuncCall. I did not send it against `fnsr-storage`-branch, as I think it will not conflict with the PRs changes and makes sense on its own

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -71,7 +71,6 @@ use PHPStan\Reflection\InitializerExprTypeResolver;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\Native\NativeParameterReflection;
 use PHPStan\Reflection\ParameterReflection;
-use PHPStan\Reflection\ParametersAcceptor;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\PassedByReference;
 use PHPStan\Reflection\Php\DummyParameter;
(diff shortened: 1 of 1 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -2410,9 +2410,10 @@ static function (Node $node, Scope $scope) use ($arrowScope, &$arrowFunctionImpu
 					$functionName = $node->name->name;
 				}
 
-				if ($functionName !== null && $this->reflectionProvider->hasFunction($functionName, $this)) {
+				$normalizedNode = ArgumentsNormalizer::reorderFuncArguments($parametersAcceptor, $node);
+				if ($normalizedNode !== null && $functionName !== null && $this->reflectionProvider->hasFunction($functionName, $this)) {
 					$functionReflection = $this->reflectionProvider->getFunction($functionName, $this);
-					$resolvedType = $this->getDynamicFunctionReturnType($parametersAcceptor, $node, $functionReflection);
+					$resolvedType = $this->getDynamicFunctionReturnType($normalizedNode, $functionReflection);
```
staabm, line 2415: in this path we invoked `getDynamicFunctionReturnType` without beeing normalized beforehand.

src/Analyser/MutatingScope.php
```diff
@@ -2471,7 +2472,7 @@ static function (Node $node, Scope $scope) use ($arrowScope, &$arrowFunctionImpu
 
 					return $cloneType;
 				}
-				$resolvedType = $this->getDynamicFunctionReturnType($parametersAcceptor, $normalizedNode, $functionReflection);
+				$resolvedType = $this->getDynamicFunctionReturnType($normalizedNode, $functionReflection);
```
staabm, line 2474: in this path $node was already normalized in line 2449 and before this PR will get normalized again within `getDynamicFunctionReturnType`

ondrejmirtes:
Thank you!

PR #4639, opened 2025-12-19
Update phpstan/phpstan-phpunit to 2.0.11

staabm:
removing distracting perf bottlenecks from the bootstrap process

see https://github.com/phpstan/phpstan-phpunit/pull/273


ondrejmirtes:
Thank you!

PR #4641, opened 2025-12-19
Fix MutatingScope->compareVariableTypeHolders()

staabm:


src/Analyser/MutatingScope.php
```diff
@@ -5641,6 +5641,10 @@ private function compareVariableTypeHolders(array $variableTypeHolders, array $o
 			unset($otherVariableTypeHolders[$variableExprString]);
 		}
 
+		if (count($otherVariableTypeHolders) > 0) {
```
staabm: in the loop this array is reduced one offset at a time, which only makes sense if we check afterwards whether `$otherVariableTypeHolders` contained elements which were not contained in `$variableTypeHolders`

ondrejmirtes: Can you write a failing test against this? Or is it just about performance?

ondrejmirtes (changes requested):

src/Analyser/MutatingScope.php
staabm: yeah, I think I need to come up with a test. looks like a bug to me.

not perf related, just stumbled over it, while looking into https://github.com/phpstan/phpstan-src/pull/4640#issuecomment-3675916706

staabm: > Can you write a failing test against this? Or is it just about performance?

I looked into it and tried. I think a case where $other has more types/conditions than $this cannot happen, because in all cases where we call `MutatingScope->equals()` we do it like `$bodyScope->equals($prevScope)`, meaning we compare a generalized scope (one with reduced information) with one with more information.

this means the scope we get as parameter will always have less information than $this so the theoretical case this PR tried to fix can never happen. but this also means there is dead code in the method, which I now have deleted.

ondrejmirtes:
Thank you!

PR #4643, opened 2025-12-20
Run cs-fixer on PHP 8.5

staabm:


ondrejmirtes:
We don't use PHP-CS-Fixer here so I'm not sure what change made PHP_CodeSniffer realize that `array_last` use is missing. I don't think it's aware of PHP functions added in various PHP versions. Also with PR subject "Run X on PHP 8.5" I'd expect some changes in .github/.

The PHAR is corrupted because PHP-Scoper probably has to be updated to understand that `array_last` should not be prefixed.

staabm:
> We don't use PHP-CS-Fixer here so I'm not sure what change made PHP_CodeSniffer realize that `array_last` use is missing.

oh sorry.. I just used `make cs` and by habit named it "cs fixer" :).

PR #4644, opened 2025-12-20
Remember function return type extensions

staabm:
before this PR:

grafik

```
➜  phpstan-src git:(2.1.x) ✗ PHPSTAN_GNSR=1 hyperfine 'php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php' --runs=5
Benchmark 1: php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php
  Time (mean ± σ):     15.219 s ±  0.105 s    [User: 14.850 s, System: 0.359 s]
  Range (min … max):   15.059 s … 15.339 s    5 runs
```

---

after this PR:
```
➜  phpstan-src git:(remember-fns) ✗ PHPSTAN_GNSR=1 hyperfine 'php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php' --runs=5
Benchmark 1: php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php
  Time (mean ± σ):     15.049 s ±  0.230 s    [User: 14.679 s, System: 0.360 s]
  Range (min … max):   14.659 s … 15.264 s    5 runs
```

### DIFF
--- a/src/Type/DynamicReturnTypeExtensionRegistry.php
+++ b/src/Type/DynamicReturnTypeExtensionRegistry.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type;
 
+use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Reflection\ReflectionProvider;
 use function array_merge;
 use function strtolower;
@@ -14,6 +15,9 @@ final class DynamicReturnTypeExtensionRegistry
 
 	/** @var DynamicStaticMethodReturnTypeExtension[][]|null */
 	private ?array $dynamicStaticMethodReturnTypeExtensionsByClass = null;
+
+	/** @var array<string, list<DynamicFunctionReturnTypeExtension>> */
+	private array $dynamicReturnTypeExtensionsByFunction = [];
 
 	/**
 	 * @param DynamicMethodReturnTypeExtension[] $dynamicMethodReturnTypeExtensions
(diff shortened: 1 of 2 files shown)


staabm:
interessting.. I don't see this errors locally.. will have another look

src/Type/DynamicReturnTypeExtensionRegistry.php
```diff
@@ -88,9 +92,23 @@ private function getDynamicExtensionsForType(array $extensions, string $classNam
...
+		$supportedFunctions = [];
+		foreach ($this->dynamicFunctionReturnTypeExtensions as $dynamicFunctionReturnTypeExtension) {
+			if (!$dynamicFunctionReturnTypeExtension->isFunctionSupported($functionReflection)) {
+				continue;
+			}
+
+			$supportedFunctions[] = $dynamicFunctionReturnTypeExtension;
+		}
+
+		return $this->dynamicReturnTypeExtensionsByFunction[$functionName] ??= $supportedFunctions;
```
ondrejmirtes: I'd say this line doesn't have to use `??=` because of `isset` above. It should just be `=`.

ondrejmirtes (changes requested):

src/Type/DynamicReturnTypeExtensionRegistry.php
staabm: good catch, fixed

staabm:
btw: including this PR and the most recent 10 commits we got considerable faster with the recent changes:

before
```
➜  hyperfine 'php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php' --runs=5
Benchmark 1: php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php
  Time (mean ± σ):     14.478 s ±  0.134 s    [User: 14.118 s, System: 0.353 s]
  Range (min … max):   14.316 s … 14.669 s    5 runs
```
 
now
```
➜  hyperfine 'php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php' --runs=5
Benchmark 1: php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php
  Time (mean ± σ):     13.764 s ±  0.135 s    [User: 13.423 s, System: 0.334 s]
  Range (min … max):   13.642 s … 13.974 s    5 runs
```

ondrejmirtes:
Yeah, great, thank you! A similar thing could be done for methods+static methods.

staabm:
> A similar thing could be done for methods+static methods.

I did not see a profile yet were this showed up. I think the reason is, that these are already key-ed by class-name, which means the list we have to work thru is way shorter than with functions

PR #4645, opened 2025-12-20
Prevent reading & json-decoding composer.json multiple times

staabm:
before this PR, when invoking

```
bin/phpstan analyze foob.php --xdebug --debug
```

you can see `composer.json` beeing read from disc and json decoded 5 times.
after this PR we read and decode `composer.json` only once.

PR #4646, opened 2025-12-21
Prevent reading & neon-decoding multiple times

staabm:
before this PR:

```
➜  phpstan-src git:(2.1.x) ✗ hyperfine 'bin/phpstan analyze foob.php' -i
Benchmark 1: bin/phpstan analyze foob.php
  Time (mean ± σ):      1.114 s ±  0.009 s    [User: 0.752 s, System: 0.267 s]
  Range (min … max):    1.105 s …  1.128 s    10 runs
 
  Warning: Ignoring non-zero exit code.
```

after this PR:

```
➜  phpstan-src git:(neon-cache) ✗ hyperfine 'bin/phpstan analyze foob.php' -i
Benchmark 1: bin/phpstan analyze foob.php
  Time (mean ± σ):     992.7 ms ±  61.0 ms    [User: 679.0 ms, System: 265.5 ms]
  Range (min … max):   938.9 ms … 1077.7 ms    10 runs
 
  Warning: Ignoring non-zero exit code.
```

similar to https://github.com/phpstan/phpstan-src/pull/4645

---

before this PR we read&decoded 46 NEON files in the above example.
after this PR we only need 28.

### DIFF
--- a/src/DependencyInjection/ContainerFactory.php
+++ b/src/DependencyInjection/ContainerFactory.php
@@ -227,7 +227,7 @@ final class ContainerFactory
 		array $loaderParameters,
 	): array
 	{
-		$neonAdapter = new NeonAdapter([]);
+		$neonAdapter = new NeonCachedFileReader([]);
 		$phpAdapter = new PhpAdapter();
 		$allConfigFiles = [];
 		$configArray = [];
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you!

PR #4647, opened 2025-12-21
Use separate cache for CachedParser->parseFile()

staabm:


src/Parser/CachedParser.php
```diff
@@ -30,66 +26,35 @@ public function __construct(
...
 		}
 
-		return $this->cachedNodesByString[$sourceCode];
+		return $this->cachedNodesByFile[$file];
 	}
 
 	/**
 	 * @return Node\Stmt[]
 	 */
 	public function parseString(string $sourceCode): array
```
staabm, line 50: hmm I wonder whether/why we need this `parseString` function. it seems its only used from tests and everywhere else we read a file to pass the contents into `parseString`.

I think we could remove some unnecessary file-read operations by dropping `parseString` and just use `parseFile` everywhere with PHPStan 3.x ... wdyt?

ondrejmirtes:
I have to remember the reason for this but I think there is one. The currently failing test testParseTheSameFileWithDifferentMethod might be showing precisely that.

PR #4648, opened 2025-12-21
Remember resolved types after pushInFunctionCall

staabm:
prevents duplicate work, as MethodCalls will re-calculated one less time.

analysing 

```php
<?php

class Foo
{

	public function doFoo(DateTimeImmutable $d): void
	{
		$a = $d->format('j. n. Y');
	}

}
```

before this PR reaches “if ($node instanceof MethodCall) {“ in MutatingScope.php:2161 4 times.
after the PR only 3 times.

with `PHPSTAN_FNSR=1 php bin/phpstan analyze foo.php --debug` after PR its only 2 times.

---

```
➜  phpstan-src git:(push-resolve) ✗ hyperfine -i 'php bin/phpstan analyze foo.php --debug'
Benchmark 1: php bin/phpstan analyze foo.php --debug
  Time (mean ± σ):     743.8 ms ±   1.8 ms    [User: 578.3 ms, System: 161.6 ms]
  Range (min … max):   741.5 ms … 747.1 ms    10 runs
 
  Warning: Ignoring non-zero exit code.
 
➜  phpstan-src git:(push-resolve) ✗ hyperfine -i 'php bin/phpstan analyze foo.php'        
Benchmark 1: php bin/phpstan analyze foo.php
  Time (mean ± σ):      1.047 s ±  0.005 s    [User: 0.777 s, System: 0.255 s]
  Range (min … max):    1.039 s …  1.054 s    10 runs
 
  Warning: Ignoring non-zero exit code.
  
➜  phpstan-src git:(2.1.x) ✗ hyperfine -i 'php bin/phpstan analyze foo.php --debug'
Benchmark 1: php bin/phpstan analyze foo.php --debug
  Time (mean ± σ):     754.9 ms ±   3.3 ms    [User: 587.2 ms, System: 164.1 ms]
  Range (min … max):   752.6 ms … 761.7 ms    10 runs
 
  Warning: Ignoring non-zero exit code.
 
➜  phpstan-src git:(2.1.x) ✗ hyperfine -i 'php bin/phpstan analyze foo.php'        
Benchmark 1: php bin/phpstan analyze foo.php
  Time (mean ± σ):      1.119 s ±  0.054 s    [User: 0.792 s, System: 0.260 s]
  Range (min … max):    1.055 s …  1.166 s    10 runs
 
  Warning: Ignoring non-zero exit code.

 ```

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2927,7 +2927,7 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 		$stack = $this->inFunctionCallsStack;
 		$stack[] = [$reflection, $parameter];
 
-		return $this->scopeFactory->create(
+		$functionScope = $this->scopeFactory->create(
 			$this->context,
 			$this->isDeclareStrictTypes(),
 			$this->getFunction(),
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2922,7 +2922,7 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 	/**
 	 * @param MethodReflection|FunctionReflection|null $reflection
 	 */
-	public function pushInFunctionCall($reflection, ?ParameterReflection $parameter): self
+	public function pushInFunctionCall($reflection, ?ParameterReflection $parameter, bool $rememberTypes): self
 	{
 		$stack = $this->inFunctionCallsStack;
 		$stack[] = [$reflection, $parameter];
@@ -2945,7 +2945,11 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 			$this->parentScope,
 			$this->nativeTypesPromoted,
 		);
-		$functionScope->resolvedTypes = $this->resolvedTypes;
+
+		if ($rememberTypes) {
+			$functionScope->resolvedTypes = $this->resolvedTypes;
+		}
+
 		return $functionScope;
 	}
 
@@ -2972,7 +2976,9 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 			$this->parentScope,
 			$this->nativeTypesPromoted,
 		);
+
 		$parentScope->resolvedTypes = $this->resolvedTypes;
+
 		return $parentScope;
 	}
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
1) Could be in popInFunctionCall too
2) BUT: `$this->inFunctionCallsStack` is used in resolveType() in `} elseif ($node instanceof Expr\Closure || $node instanceof Expr\ArrowFunction) {` so that could be a problem (the problem should be reproduced in a test first).

ondrejmirtes:
Issue bot points out an example where this breaks https://phpstan.org/r/dfc26675-234f-47b3-8d0f-dc64b66ef5aa

staabm:
> 2\. BUT: `$this->inFunctionCallsStack` is used in resolveType() in `} elseif ($node instanceof Expr\Closure || $node instanceof Expr\ArrowFunction) {` so that could be a problem (the problem should be reproduced in a test first).

I think this problem was caught by existing tests, as the suite failed with the change until I got it right.

ondrejmirtes:
Perfect! Thank you. My idea for preserving `resolvedTypes` in often-called methods like assignVariable, mergeWith, specifyExpressionType is to have something like `$scope->resolvedTypes = $this->preserveResolvedTypes($newExpressionTypes)` before `return $scope;`.

This logic doesn't have to be called in methods like `enterDeclareStrictTypes`, `enterClass`, `enterNamespace` because there are unlikely to be any expressions.

And the implementation of preserveResolvedTypes would look at changes between `$this->expressionTypes` and `$newExpressionTypes`. For example - if the new scope doesn't change anything about `$a`, we can preserve resolved types about `$a`.

This logic is very similar to what `invalidateExpression` does. But of course the risk is that the logic to figure out what cached types to preserve would be more expensive than to just empty the whole cache every time. So even if we have a proper implementation of the idea and all tests pass, we have to measure whether it's any faster.

Do you think you want to look into this?

staabm:
I can have a look

staabm:
> My idea for preserving `resolvedTypes` in often-called methods like assignVariable, mergeWith, specifyExpressionType is to have something like `$scope->resolvedTypes = $this->preserveResolvedTypes($newExpressionTypes)` before `return $scope;`.


I am not sure the memory/perf tradeoff will be good in this direction. a different direction to explore could be trying to implement `TypeTraverser` natively.

it seems bootleneck enough and way less complex than a `TypeCombinator` re-implementation in e.g. rust

see `blackfire run php vendor/bin/phpunit tests/PHPStan/Analyser/NodeScopeResolverTest.php`
grafik

PR #4651, opened 2025-12-22
Immediate unwrap AlwaysRememberedExpr after used on NodeScopeResolver

samsonasik:
@TomasVotruba if this accepted, this will should reduce unnecessary tweak in rector Printer

https://github.com/rectorphp/rector-src/blob/e39c3e1250194f9eaacf34c78c8e951c84f26222/src/PhpParser/Printer/BetterStandardPrinter.php#L131-L146

This PR ensure immediate unwrap AlwaysRememberedExpr after used on NodeScopeResolver so not hide behind deep cloned node.


samsonasik:
@ondrejmirtes Ready to merge 👍

CI notice seems unrelated as happen on latest 2.1.x branch as well.

ondrejmirtes:
Thank you, let's try this. I'll let you know if it breaks anything later.

TomasVotruba:
Thank you @ondrejmirtes  :clap:

samsonasik:
Thank you @ondrejmirtes, I tested phpstan 2.1.x-dev on rector-src and it seems working ok 👍 

- https://github.com/rectorphp/rector-src/pull/7770

PR #4652, opened 2025-12-22
Ignore by identifier instead of regex error pattern

staabm:
regex re-validation/AST-parsing currently is the slowest parts in some tests.

can be seen in 

```
➜  cat foo.php 
<?php

class Foo
{

        public function doFoo(DateTimeImmutable $d): void
        {
                $a = $d->format('j. n. Y');
        }

}

➜  blackfire run /opt/homebrew/bin/php bin/phpstan analyze foo.php --debug 
```

grafik


ondrejmirtes:
Thank you!

PR #4654, opened 2025-12-23
Use self:: over $this-> for PHPUnit assertions

staabm:
with static calls assertion-calls do not affect remembered non-static properties.

in addition assertion methods are declared `static` so we don't need to process ignored errors for these lines

---

before PR
```
➜  phpstan-src git:(2.1.x) ✗ hyperfine 'bin/phpstan analyze tests/PHPStan/Analyser/AnalyserIntegrationTest.php' --runs=5
Benchmark 1: bin/phpstan analyze tests/PHPStan/Analyser/AnalyserIntegrationTest.php
  Time (mean ± σ):      2.537 s ±  0.047 s    [User: 2.224 s, System: 0.279 s]
  Range (min … max):    2.508 s …  2.620 s    5 runs
```

after PR
```
➜  phpstan-src git:(this-self) ✗ hyperfine 'bin/phpstan analyze tests/PHPStan/Analyser/AnalyserIntegrationTest.php' --runs=5
Benchmark 1: bin/phpstan analyze tests/PHPStan/Analyser/AnalyserIntegrationTest.php
  Time (mean ± σ):      2.526 s ±  0.016 s    [User: 2.220 s, System: 0.281 s]
  Range (min … max):    2.513 s …  2.554 s    5 runs
```

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -24,40 +24,40 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 	public function testUndefinedVariableFromAssignErrorHasLine(): void
 	{
 		$errors = $this->runAnalyse(__DIR__ . '/data/undefined-variable-assign.php');
-		$this->assertCount(2, $errors);
+		self::assertCount(2, $errors);
 		$error = $errors[0];
-		$this->assertSame('Undefined variable: $bar', $error->getMessage());
-		$this->assertSame(3, $error->getLine());
+		self::assertSame('Undefined variable: $bar', $error->getMessage());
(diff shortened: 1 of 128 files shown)


ondrejmirtes:
I kind of disagree with this one. PHPUnit documentation allows both, and because of years of muscle memory I still use `$this->` myself.

If anything we should have `ignoreErrors` entry with `reportUnmatched: false` in `phpstan-phpunit` so that people don't have to ignore this on their own. Or an IgnoreErrorsExtension.

talkinnl:
> I kind of disagree with this one. PHPUnit documentation allows both, and because of years of muscle memory I still use $this-> myself.

Of course, this topic is subjective, and it may be a clash of project vs library code style...  but I think this PR helps aligning with the more common test code style in the community, so less friction or surprising. :)

PR #4656, opened 2025-12-23
Use sha256 which is cpu optimized in PHP 8.4+

staabm:
starting with PHP 8.4+ sha256 can be 2-5x faster than sha1, see https://github.com/php/php-src/pull/15152

---

see https://speakerdeck.com/edorian/upcoming-changes-in-php-8-dot-4-property-hooks-improved-performance-and-so-much-more?slide=71

grafik


ondrejmirtes:
Thank you!

PR #4657, opened 2025-12-23
Microoptimize TypeCombinator

staabm:
lets reduce the work beeing done to a minimum in our most hot path

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -201,7 +201,7 @@ final class TypeCombinator
 				unset($types[$i]);
 				continue;
 			}
-			
+
 			if ($types[$i]->isBoolean()->yes()) {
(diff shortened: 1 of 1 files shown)


src/Type/TypeCombinator.php
```diff
@@ -201,24 +201,23 @@ public static function union(Type ...$types): Type
 				unset($types[$i]);
 				continue;
 			}
+			
 			if ($types[$i]->isBoolean()->yes()) {
 				$hasGenericScalarTypes[ConstantBooleanType::class] = true;
-			}
-			if ($types[$i]->isFloat()->yes()) {
+			} elseif ($types[$i]->isFloat()->yes()) {
```
staabm, line 207: something which is a bool yes() cannot be a float yes() at the same time etc.

ondrejmirtes:
Thank you!

PR #4658, opened 2025-12-23
TypeCombinator: Remove unnecessary md5 encoding

staabm:
lets reduce the work beeing done to a minimum in our most hot path


ondrejmirtes:
Thank you!

PR #4659, opened 2025-12-23
Simplify FiberNodeScopeResolver

staabm:
lets see whether you like `goto`, but I think its easier to read that way

[review with whitespaces disabled](https://github.com/phpstan/phpstan-src/pull/4659/files?w=1)


ondrejmirtes:
Thank you, I like it!

PR #4662, opened 2025-12-24
Prevent duplicate super-global-variable detection

staabm:

before this PR:
```
...
Elapsed time: 37 seconds
Used memory: 366.5 MB
```

after this PR:
```
Elapsed time: 36 seconds
Used memory: 354.03 MB
...
```

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -5008,7 +5008,7 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 				if (!$expr->hasAttribute(self::CONTAINS_SUPER_GLOBAL_ATTRIBUTE_NAME)) {
 					$expr->setAttribute(self::CONTAINS_SUPER_GLOBAL_ATTRIBUTE_NAME, $nodeFinder->findFirst($expr, $globalVariableCallback) !== null);
 				}
-				if ($expr->getAttribute(self::CONTAINS_SUPER_GLOBAL_ATTRIBUTE_NAME) !== true) {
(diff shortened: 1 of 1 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -5004,7 +5005,10 @@ private function mergeVariableHolders(array $ourVariableTypeHolders, array $thei
 				$intersectedVariableTypeHolders[$exprString] = $variableTypeHolder->and($theirVariableTypeHolders[$exprString]);
 			} else {
 				$expr = $variableTypeHolder->getExpr();
-				if ($nodeFinder->findFirst($expr, $globalVariableCallback) !== null) {
+				if (!$expr->hasAttribute(self::CONTAINS_SUPER_GLOBAL_ATTRIBUTE_NAME)) {
+					$expr->setAttribute(self::CONTAINS_SUPER_GLOBAL_ATTRIBUTE_NAME, $nodeFinder->findFirst($expr, $globalVariableCallback) !== null);
+				}
+				if ($expr->getAttribute(self::CONTAINS_SUPER_GLOBAL_ATTRIBUTE_NAME) !== null) {
```
staabm: maybe super global variables are rare enough that we could instead remember with a bool property on Scope, whether the current scope contains any super-global variable. 

if not, skip this whole node-finding at this stage.

---

another way could be using a NodeVisitor

ondrejmirtes:
Why did you close this? 😊

staabm:
it contained a bug :) - re-opened but the numbers are not as good as initially thought (updated the description)

ondrejmirtes:
Thank you!

PR #4663, opened 2025-12-25
Prevent Scalar expressions to be specified

staabm:
before this change a call to `match(1)` would specify a expression `$scope->expressionTypes[1]` (integer key). 
we expect this array to be keyed with `string` types though:

https://github.com/phpstan/phpstan-src/blob/21a16027a4b285a06a7e6efe5b3718195b34231b/src/Analyser/MutatingScope.php#L200

In addition literal const scalar values do not make sense to be specified in scope.. it does not provide more information

[found in snippet:](https://phpstan.org/r/2695177e-2bce-4206-9efa-1d3288ad340c)

```php
<?php declare(strict_types = 1); // lint >= 8.1

namespace Bug10418;

use function PHPStan\debugScope;

function (): void {
	$text = '123';
	$result = match(1){
		preg_match('/(\d+)/', $text, $match) => 'matched number: ' . $match[1],
		preg_match('/(\w+)/', $text, $match) => 'matched word: ' . json_encode($match),
		default => 'no matches!'
	};
	debugScope();
};

```

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -24,6 +24,7 @@ use PhpParser\Node\InterpolatedStringPart;
 use PhpParser\Node\Name;
 use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\PropertyHook;
+use PhpParser\Node\Scalar;
 use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Function_;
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
A similar skip could be done in TypeSpecifier::createForExpr as well.

ondrejmirtes:
Thank you!

ondrejmirtes:
The TypeSpecifier idea doesn't work, sometimes we have to report things like `is_numeric('blabla')` is always false.

PR #4664, opened 2025-12-25
Print more precise elapsed time

staabm:
running `time php bin/phpstan analyse src/Analyser/ src/Rules/ -v --debug` takes 37 seconds.
for perf tunning its useful to see more precise timings than a value cut to full seconds


ondrejmirtes:
Thank you!

PR #4669, opened 2025-12-26
ParametersAcceptorSelector: early return type traversal

staabm:
no need to deep traverse maybe complex type structures. stop traversal on first find.

type traversal is one of the slowest parts atm:

grafik


ondrejmirtes:
Thank you!

PR #4670, opened 2025-12-26
VerbosityLevel: early return type traversal

staabm:
analog https://github.com/phpstan/phpstan-src/pull/4669


ondrejmirtes:
Thank you!

PR #4672, opened 2025-12-27
LazyInternalScopeFactory: Prevent expensive repeated getService() calls

staabm:
shows up in profiles

grafik


ondrejmirtes:
Thank you. The getByType calls don't show up?

staabm:
> The getByType calls don't show up?

nope

ondrejmirtes:
Because MemoizingContainer caches getByType but not getService.

PR #4673, opened 2025-12-27
Restrict chr param for PHP 8.5+

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13930

### DIFF
--- a/resources/functionMap_php85delta.php
+++ b/resources/functionMap_php85delta.php
@@ -1,0 +1,23 @@
+<?php // phpcs:ignoreFile
+
+/**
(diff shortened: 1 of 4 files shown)


staabm (approved):
I think this makes sense as php-src 8.5+ reports a deprecation.
hiding it behind bleeding-edge therefore makes no sense for me.

@ondrejmirtes might have a opinion on this one, as it kind-of breaks BC (more precise parameter type)

ondrejmirtes:
I agree here, this is okay. Doing things that only affect upcoming PHP versions without bleeding edge is okay.

It'd be better if this was done before PHP 8.5 release but I'm still okay with merging this.

ondrejmirtes:
Thank you.

PR #4676, opened 2025-12-28
Don't store scalar values in resolvedTypes

staabm:
before this PR we stored types about constant values in scope

grafik

similar to https://github.com/phpstan/phpstan-src/pull/4663 I think this is not useful.

---

this was especially questionable for constant arrays:

grafik


staabm:
lets merge this one only after we see everything works in https://github.com/phpstan/phpstan-src/pull/4677

PR #4677, opened 2025-12-28
Preserve already resolved types after scope re-creation

staabm:
implements the idea described in https://github.com/phpstan/phpstan-src/pull/4648#issuecomment-3682244126

requires https://github.com/phpstan/phpstan-src/pull/4676

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -10,6 +10,7 @@ use PhpParser\Node\Arg;
 use PhpParser\Node\ComplexType;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\ArrayDimFetch;
 use PhpParser\Node\Expr\BinaryOp;
 use PhpParser\Node\Expr\Cast\Unset_;
 use PhpParser\Node\Expr\ConstFetch;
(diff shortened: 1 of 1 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -4406,6 +4406,26 @@ public function assignInitializedProperty(Type $fetchedOnType, string $propertyN
...
+	/**
+	 * @param array<string, ExpressionTypeHolder> $expressionTypes
+	 *
+	 * @return array<string, ExpressionTypeHolder>
+	 */
+	private function preserveResolvedTypes(array $expressionTypes): array
+	{
+		$preservedTypes = $this->resolvedTypes;
+		foreach($preservedTypes as $exprStringToInvalidate => $resolvedType) {
+			foreach ($expressionTypes as $exprString => $exprTypeHolder) {
```
staabm: we cannot use similar logic as in `invalidateExpression` because we don't know the `Expr` which `$this->resolvedTypes` was resolved with

ondrejmirtes: I have a commit that puts ExpressionTypeHolder into resolvedTypes, do you want it?

staabm: Might help. Sure why not :-). 

Will be afk for a few hours now

ondrejmirtes: I pushed it to 2.1.x. If it doesn't help you in the end, or if this PR does not lead to performance improvements, I'll revert it.

staabm:
remaining problem is, that we need to forget resolved type of `$a` when scope specifies `$a['bar']`.
I remember we have such logic arround somewhere, but I am not able to find it atm

staabm:
tried a few more things but did not find something measurably better than what we have atm

PR #4680, opened 2025-12-29
Speedup analysis of big arrays containing closures

staabm:
degrade analysis precision to a callable type

refs https://github.com/phpstan/phpstan/issues/13933

---

before this PR
```
php bin/phpstan analyze ../phpstan-performance-issue-reproduction/src/index_bad.php --debug -vvv 

Elapsed time: 1.44 seconds
Used memory: 82.5 MB
```

after this PR
```
php bin/phpstan analyze ../phpstan-performance-issue-reproduction/src/index_bad.php --debug -vvv 

Elapsed time: 0.67 seconds
Used memory: 70 MB
```

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -171,7 +171,7 @@ class ConstantArrayType implements Type
 
 		$count = count($this->valueTypes);
 		if ($count === 0) {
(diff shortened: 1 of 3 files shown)


staabm:
better fix in https://github.com/phpstan/phpstan-src/pull/4684

PR #4686, opened 2025-12-31
Prevent repetative method calls on ParametersAcceptor

staabm:
as can be seen in 

grafik

`getParameters()` and `getReturnType` is called 3x more than others


ondrejmirtes:
Thank you!

PR #4688, opened 2025-12-31
Improve return type of `range()` function to always yield a `non-empty-list`

FeBe95:
## Description

By definition, the `range()` function will never return an empty array. This PR updates the `RangeFunctionReturnTypeExtension` to follow this definition for all cases.

> [!NOTE]
> PHP 8.3 changed how certain “special” or unusual `range()` calls are handled, but this does not affect the aforementioned definition.
> #### Example:
> ```php
> range(1, 5, -1); // increasing range with a negative step
> ```
> Up to PHP 8.2, the interpreter attempted to infer a sensible result (e.g. by normalizing the step). Starting with PHP 8.3, such calls either emit warnings or throw errors instead.

## Example
```php
range(1, $someInteger);
```
- Before:
  ```md
  list<int>
  ```
- Now:
  ```md
  non-empty-list<int>
  ```
  
## Implementation

Some return types previously did not use `NonEmptyArrayType`. This PR applies `NonEmptyArrayType` to all possible return types of `range()`, ensuring they differ only in their inner element type.

To avoid duplication, a small helper method `getNonEmptyListOfType($type)` was introduced and reused across the 9 different return statements.

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -1416,7 +1416,6 @@ class ConstantArrayType implements Type
 	{
 		$count = count($types);
 		$autoIndexes = range($count - count($this->optionalKeys), $count);
-		assert($autoIndexes !== []);
 
 		if ($this->isList->yes()) {
 			// Optimized version for lists: Assume that if a later key exists, then earlier keys also exist.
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
Thank you!

PR #4690, opened 2026-01-01
Prevent maybe unnecessary work in MutatingScope::filterBySpecifiedTypes()

staabm:


src/Analyser/MutatingScope.php
```diff
@@ -4747,15 +4747,17 @@ public function filterBySpecifiedTypes(SpecifiedTypes $specifiedTypes): self
 			if ($certainty->no()) {
 				unset($scope->expressionTypes[$conditionalExprString]);
 			} else {
-				$type = TypeCombinator::intersect(...array_map(static fn (ConditionalExpressionHolder $holder) => $holder->getTypeHolder()->getType(), $expressions));
```
staabm, line 4750: this type was always calculated but only used in the IF-THEN case

ondrejmirtes:
Thank you!

PR #4691, opened 2026-01-01
Prevent unnecessary node-key calculation

staabm:
node-key calculation is used a lot, therefore sometimes shows up in blackfire profiles

grafik


ondrejmirtes:
Thank you!

PR #4684, opened 2025-12-31
Speedup analysis of arrays containing closures

staabm:
alternative implementation to https://github.com/phpstan/phpstan-src/pull/4680

degrade analysis precision to a callable type

refs https://github.com/phpstan/phpstan/issues/13933

---

before this PR
```
php bin/phpstan analyze ../phpstan-performance-issue-reproduction/src/index_bad.php --debug -vvv 

Elapsed time: 16.15 seconds
Used memory: 174 MB
```

after this PR
```
php bin/phpstan analyze ../phpstan-performance-issue-reproduction/src/index_bad.php --debug -vvv 

Elapsed time: 2.24 seconds
Used memory: 90 MB
```

----

before this PR

```
➜  phpstan-src git:(2.1.x) ✗ time vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php --filter testBug13933
PHPUnit 11.5.46 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.29
Configuration: /Users/staabm/workspace/phpstan-src/phpunit.xml
Random Seed:   1767176262

.                                                                   1 / 1 (100%)

Time: 00:06.510, Memory: 92.00 MB


OK (1 test, 1 assertion)
vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php --filte  6.41s user 0.17s system 99% cpu 6.580 total
```

after this PR

```
➜  phpstan-src git:(2.1.x) ✗ time vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php --filter testBug13933
PHPUnit 11.5.46 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.29
Configuration: /Users/staabm/workspace/phpstan-src/phpunit.xml
Random Seed:   1767176352

.                                                                   1 / 1 (100%)

Time: 00:00.906, Memory: 80.50 MB

OK (1 test, 1 assertion)
vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php --filte  0.81s user 0.16s system 99% cpu 0.979 total
```

PR #4693, opened 2026-01-01
Reapply "Enable FileCacheStorage for OptimizedDirectorySourceLocatorFactory in tests"

staabm:
as discussed in https://github.com/phpstan/phpstan-src/pull/4684#issuecomment-3703666507


ondrejmirtes:
Thank you!

PR #4695, opened 2026-01-02
TypeCombinator: prevent unnecessary work

staabm:
every constant array key needs only be contained once


ondrejmirtes:
Thank you!

PR #4696, opened 2026-01-02
ConstantStringType: Prevent creation of unnecessary constant-object types

staabm:
before this PR most constant-string key types contained the same constant type again, which represents the php auto converted array-key variant.

I don't think we should hold 2 objects per constant string type which is used as array key.

grafik


ondrejmirtes:
Thank you!

PR #4697, opened 2026-01-02
Remove NodeFinder from MutatingScope->restoreThis()

staabm:
less AST traversing of expressions which cannot reference $this


ondrejmirtes:
Thank you!

PR #4698, opened 2026-01-02
Simplify NodeScopeResolver->processImmediatelyCalledCallable()

staabm:
traverse each expression only once, intstead of once per use


ondrejmirtes:
Nice find! Thank you.

PR #4699, opened 2026-01-02
Respect sys_get_temp_dir() from the parent process

dktapps:
Not sure how this should be tested on the CI, but I tested this locally and verified that it fixes https://github.com/phpstan/phpstan/issues/13929.

If I understand the code in `CommandHelper` correctly, `parameters.tmpDir` will override this if set. This is a bit annoying, but since there's no way (that I know of) to tell if an ini setting was set on CLI override or by php.ini, I suppose this is the most predictable behaviour.

It doesn't look too difficult to add a command line option to override the tmpDir from what I can see, although I'm not too clear how that would interact with the dependency injection, so I'm reluctant to mess with it for now.

### DIFF
--- a/src/Process/ProcessHelper.php
+++ b/src/Process/ProcessHelper.php
@@ -32,6 +32,8 @@ final class ProcessHelper
 
 		$processCommandArray = [
 			$phpCmd,
+			'-d',
+			'sys_temp_dir=' . escapeshellarg(sys_get_temp_dir()),
 		];
 
 		if ($input->getOption('memory-limit') === null) {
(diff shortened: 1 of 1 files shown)


dktapps:
Semi related, some of these usages probably should be changed to respect the application config's `tmpDir`:

image

ondrejmirtes:
I'd merge this but please fix CS.

staabm (approved):

ondrejmirtes:
Thank you.

ondrejmirtes:
FYI had to revert this because of https://github.com/phpstan/phpstan/issues/14093

PR #4700, opened 2026-01-03
Fix PHPStorm inspection errors

staabm:


src/Analyser/NodeScopeResolver.php
```diff
@@ -1910,7 +1910,7 @@ private function processStmtNode(
 							continue;
 						}
 
-						foreach ($catchTypes as $catchTypeIndex => $catchTypeItem) {
```
staabm, line 1913: unused variable $catchTypeIndex

src/Type/Php/StatDynamicReturnTypeExtension.php
```diff
@@ -68,7 +69,7 @@ private function getReturnType(): Type
 			'blocks',
 		];
 
-		foreach ($keys as $key) {
```
staabm, line 71: unused variable $key

tests/PHPStan/Rules/Methods/data/uppercase-string.php
```diff
@@ -28,6 +28,6 @@ public function test(
 		$this->acceptUppercaseString($string);
 		$this->acceptUppercaseString($uppercaseString);
 		$this->acceptUppercaseString($numericString);
-		$this->acceptUppercaseString($nonEmptyLowercaseString);
```
staabm, line 31: undefined variable

src/Analyser/NodeScopeResolver.php
ondrejmirtes: Slevomat CS typically catches this, can you figure out why it doesn't in this case, maybe create a failign test for slevomat/coding-standard?

ondrejmirtes (changes requested):

PR #4701, opened 2026-01-03
Use static callback for TypeTraverser::map()

staabm:


ondrejmirtes:
What's the benefit here?

staabm:
not measurable as I can see it.

I came to this change because in php-src they seem to try hard to auto-detect "static callback" as it seems there can be some benefit in it: https://github.com/php/php-src/pull/19941

I am fine with closing in case you don't like it

ondrejmirtes:
It's okay :)

iluuu1994:
Side-note: The caching of static closures has an additional restriction.

> The PR also adds caching for static closures **that don't have any bindings** and don't declare static variables.

The binding disqualifies the closure from being cached, so this won't benefit either way.

staabm:
Thanks for noting!

PR #4705, opened 2026-01-03
Prevent unnecessary work in NodeScopeResolver->processAssignVar()

staabm:
prevents unnecessary work when e.g. assigning variables to non-falsy values.

to my suprise this hardly has an influence on performance, even though I can see `if ($truthyType !== $type) {` is not always `true`.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -6018,15 +6018,16 @@ class NodeScopeResolver
 				}
 
 				$truthySpecifiedTypes = $this->typeSpecifier->specifyTypesInCondition($scope, $assignedExpr, TypeSpecifierContext::createTruthy());
-				$falseySpecifiedTypes = $this->typeSpecifier->specifyTypesInCondition($scope, $assignedExpr, TypeSpecifierContext::createFalsey());
-
 				$truthyType = TypeCombinator::removeFalsey($type);
-				$falseyType = TypeCombinator::intersect($type, StaticTypeFactory::falsey());
-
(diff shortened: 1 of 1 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -6018,15 +6018,16 @@ private function processAssignVar(
...
-
 				$truthyType = TypeCombinator::removeFalsey($type);
-				$falseyType = TypeCombinator::intersect($type, StaticTypeFactory::falsey());
-
 				$conditionalExpressions = $this->processSureTypesForConditionalExpressionsAfterAssign($scope, $var->name, $conditionalExpressions, $truthySpecifiedTypes, $truthyType);
 				$conditionalExpressions = $this->processSureNotTypesForConditionalExpressionsAfterAssign($scope, $var->name, $conditionalExpressions, $truthySpecifiedTypes, $truthyType);
-				$conditionalExpressions = $this->processSureTypesForConditionalExpressionsAfterAssign($scope, $var->name, $conditionalExpressions, $falseySpecifiedTypes, $falseyType);
-				$conditionalExpressions = $this->processSureNotTypesForConditionalExpressionsAfterAssign($scope, $var->name, $conditionalExpressions, $falseySpecifiedTypes, $falseyType);
+
+				if ($truthyType !== $type) {
```
staabm, line 6025: when `$truthyType = TypeCombinator::removeFalsey($type);` returns `$truthyType === $type`, this means `TypeCombinator::intersect($type, StaticTypeFactory::falsey())` will always be a `NeverType`.

staabm: blackfire reports 10.000 less invocations of `TypeCombinator::intersect` with this PR in comparison to 2.1.x@c53546e53a on `bin/phpstan analyse src/Analyser/ src/Rules/ -v --debug`

ondrejmirtes:
Very very nice! Looking forward to what you find next :)

PR #4706, opened 2026-01-03
Prevent unnecessary work in While/Do analysis

staabm:
this type-logic only needs to be executed when `$context->isTopLevel()`


ondrejmirtes:
Thank you!

PR #4707, opened 2026-01-04
Prevent unnecessary work in TypeSpecifier->createForExpr()

staabm:
saves ~1 second on `php bin/phpstan analyse src/Analyser/ src/Rules/ -v --debug`


ondrejmirtes:
Thank you!

PR #4708, opened 2026-01-04
NodeScopeResolver: Move condition out of loop

staabm:


src/Analyser/NodeScopeResolver.php
```diff
@@ -5767,85 +5767,84 @@ private function processArgs(
 
 			$scope = $scope->restoreOriginalScopeAfterClosureBind($originalScope);
 		}
-		foreach ($args as $i => $arg) {
-			if (!isset($parameters) || $parametersAcceptor === null) {
-				continue;
-			}
```
staabm, line 5773: effectively putting these 3 lines arround the loop instead of beeing the first statement which prevents looping thru all args

ondrejmirtes:
Thank you!

PR #4709, opened 2026-01-04
NodeScopeResolver: determine callCallbackImmediately only for callables

staabm:
before this PR $callCallbackImmediately was determined for any expression


ondrejmirtes:
Thank you!

PR #4711, opened 2026-01-04
NodeScopeResolver: Move assignByReference logic out of generic path

staabm:
this logic was executed for every expression, while we only need when `!$argValue instanceof Variable || $argValue->name !== 'this'`.

`processArgs` consumes 17% of overall runtime. lets try to not do unnecessary stuff again and again

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5754,7 +5754,6 @@ class NodeScopeResolver
 
 		if (isset($parameters) && $parametersAcceptor !== null) {
 			foreach ($args as $i => $arg) {
-				$byRefType = new MixedType();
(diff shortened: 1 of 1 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -5754,7 +5754,6 @@ private function processArgs(
 
 		if (isset($parameters) && $parametersAcceptor !== null) {
 			foreach ($args as $i => $arg) {
```
staabm, line 5756: I did not yet understand why we iterate all args twice inside of `processArgs()`, but if possible, I guess we could save quite some time in case we could it do just once

ondrejmirtes:
Thank you!

PR #4712, opened 2026-01-04
NodeScopeResolver: Skip const args processing

staabm:
const args cannot impact the call-site, therefore skip them - except when they are referenced via attributes


staabm:
hmm I don't see a measurable improvement

ondrejmirtes:
I don't get it, we can't skip those. Let's say someone has a rule for `Node\Scalar`. This change would mean the rule is stopped being called.

PR #4713, opened 2026-01-04
More precise Type->getCallableParametersAcceptors() return type

staabm:
we already have [code arround which assumes this method returns a list](https://github.com/search?q=repo%3Aphpstan%2Fphpstan-src%20%22acceptors%5B0%22&type=code)


ondrejmirtes:
Thank you!

PR #4714, opened 2026-01-04
Simplify NodeScopeResolver

staabm:


src/Analyser/NodeScopeResolver.php
```diff
@@ -2752,13 +2752,7 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context, $sto
...
 				$impurePoints = $nameResult->getImpurePoints();
 				$isAlwaysTerminating = $nameResult->isAlwaysTerminating();
-				if ( // phpcs:ignore
-					$nameType->isObject()->yes()
-					&& $nameType->isCallable()->yes()
-					&& (new ObjectType(Closure::class))->isSuperTypeOf($nameType)->no()
-				) {
-					// processed later
-				} elseif ($parametersAcceptor instanceof CallableParametersAcceptor) {
+				if ($parametersAcceptor instanceof CallableParametersAcceptor) {
```
ondrejmirtes, line 2755: This is different behaviour because we don't want this `if` to happen when:

```

					$nameType->isObject()->yes()
					&& $nameType->isCallable()->yes()
					&& (new ObjectType(Closure::class))->isSuperTypeOf($nameType)->no()
```

ondrejmirtes (changes requested):

PR #4716, opened 2026-01-04
Prevent unnecessary work in NodeScopeResolver->processAssignVar()

staabm:
followup to https://github.com/phpstan/phpstan-src/pull/4705


ondrejmirtes:
Thank you!

PR #4719, opened 2026-01-05
Store only variables as expression types with maybe certainty

ondrejmirtes:


src/Analyser/MutatingScope.php
```diff
@@ -5024,6 +5025,10 @@ private function mergeVariableHolders(array $ourVariableTypeHolders, array $thei
 					continue;
 				}
 
+				if (!$expr instanceof Variable && !$expr instanceof VirtualNode) {
```
staabm, line 5028: I think we should do this check earlier.

we don't need todo the ast traversal in line 5022 when we later on skip the expression entirely

src/Analyser/MutatingScope.php
```diff
@@ -5041,6 +5046,10 @@ private function mergeVariableHolders(array $ourVariableTypeHolders, array $thei
 				continue;
 			}
 
+			if (!$expr instanceof Variable && !$expr instanceof VirtualNode) {
```
staabm, line 5049: same as above, in relation to line 5043

src/Analyser/MutatingScope.php
staabm: here we go: https://github.com/phpstan/phpstan-src/pull/4720

PR #4720, opened 2026-01-05
Prevent unnecessary work in MutatingScope->mergeVariableHolders()

staabm:


staabm:
Maybe it might make sense to not even write such expressions into the holders in the first place?

ondrejmirtes:
The problem is here is when they become "maybe". With "yes" in the scope they have value.

ondrejmirtes:
Thank you!

PR #4721, opened 2026-01-05
Update dependency shipmonk/dead-code-detector to ^0.14.0

staabm:


staabm:
@janedbal in case its interessting to you. it seems the newer release of shipmonk/dead-code-detector requires more memory

janedbal:
> the newer release of shipmonk/dead-code-detector requires more memory

Some increase is expected (as we track more member types), but I'll check how much. Thx

ondrejmirtes:
I think memory was a problem even before 0.14 (and after the version we're currently using), but can't find the failing PR right now.

PR #4722, opened 2026-01-05
Prevent crash when analysing curl_setopt

schlndh:
The crash happened when analysing curl_setopt without having the curl extension loaded. It crashed due to Undefined constant "CURLOPT_SHARE".


ondrejmirtes:
Thank you!

PR #4723, opened 2026-01-05
Less expression printing

staabm:
with the recent changes from https://github.com/phpstan/phpstan-src/commit/21d86f0675008ebea2ba52fdee6013a6180a228d we started printing all expressions.

since expression printing shows up in most profiles, we should only print when we really need the string


ondrejmirtes:
It's a good thing you check these changes after me 😊

PR #4724, opened 2026-01-05
Made closure php-src cachable

staabm:
see https://github.com/phpstan/phpstan-src/pull/4701#issuecomment-3707068172

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -73,36 +73,6 @@ parameters:
 			path: src/Analyser/RuleErrorTransformer.php
 
 		-
-			rawMessage: Doing instanceof PHPStan\Type\IntersectionType is error-prone and deprecated.
-			identifier: phpstanApi.instanceofType
-			count: 1
-			path: src/Analyser/Traverser/CloneTypeTraverser.php
-
-		-
-			rawMessage: 'Doing instanceof PHPStan\Type\Constant\ConstantStringType is error-prone and deprecated. Use Type::getConstantStrings() instead.'
-			identifier: phpstanApi.instanceofType
-			count: 1
-			path: src/Analyser/Traverser/InstanceOfClassTypeTraverser.php
-
-		-
-			rawMessage: 'Doing instanceof PHPStan\Type\Generic\GenericClassStringType is error-prone and deprecated. Use Type::isClassStringType() and Type::getClassStringObjectType() instead.'
-			identifier: phpstanApi.instanceofType
-			count: 1
-			path: src/Analyser/Traverser/InstanceOfClassTypeTraverser.php
-
-		-
-			rawMessage: Doing instanceof PHPStan\Type\IntersectionType is error-prone and deprecated.
-			identifier: phpstanApi.instanceofType
-			count: 1
-			path: src/Analyser/Traverser/InstanceOfClassTypeTraverser.php
-
-		-
-			rawMessage: Doing instanceof PHPStan\Type\IntersectionType is error-prone and deprecated.
-			identifier: phpstanApi.instanceofType
-			count: 1
-			path: src/Analyser/Traverser/VoidToNullTraverser.php
-
-		-
 			rawMessage: 'Doing instanceof PHPStan\Type\ConstantScalarType is error-prone and deprecated. Use Type::isConstantScalarValue() or Type::getConstantScalarTypes() or Type::getConstantScalarValues() instead.'
 			identifier: phpstanApi.instanceofType
 			count: 2
@@ -1706,12 +1676,6 @@ parameters:
 			identifier: phpstanApi.instanceofType
 			count: 2
 			path: src/Type/StringType.php
-
-		-
-			rawMessage: 'Doing instanceof PHPStan\Type\CallableType is error-prone and deprecated. Use Type::isCallable() and Type::getCallableParametersAcceptors() instead.'
-			identifier: phpstanApi.instanceofType
-			count: 1
-			path: src/Type/Traverser/LateResolvableTraverser.php
 
 		-
 			rawMessage: 'Doing instanceof PHPStan\Type\ArrayType is error-prone and deprecated. Use Type::isArray() or Type::getArrays() instead.'
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -73,6 +73,36 @@ parameters:
 			path: src/Analyser/RuleErrorTransformer.php
 
 		-
+			rawMessage: Doing instanceof PHPStan\Type\IntersectionType is error-prone and deprecated.
+			identifier: phpstanApi.instanceofType
+			count: 1
+			path: src/Analyser/Traverser/CloneTypeTraverser.php
+
+		-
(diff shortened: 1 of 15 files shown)


staabm:
@iluuu1994 could you have another look at this PR. intention is, that php-src will be able to cache the closure, while it could not before.

does this make sense like that and will it have the expected behaviour?

staabm:
regarding other examples:

```
$classType = TypeTraverser::map($classType, static function (Type $type, callable $traverse) use (&$uncertainty): Type {
						return $traverse($type);
					}
					if ($type->getObjectClassNames() !== []) {
						$uncertainty = true;
						return $type;
					}
					if ($type instanceof GenericClassStringType) {
						$uncertainty = true;
						return $type->getGenericType();
					}
						return new ObjectType($type->getValue());
					}
					return new MixedType();
				});
```

do you think it would make things faster if we transform such closures into e.g. __invoke classes or similar?

Some callsites of `TypeTraverser::map` are easily invoked multi-million times in hot paths

dktapps:
Basically the requirement for being cached is that it has to be stateless. So:

- no `$this`
- no `use`d variables
- no local `static` variables

In some cases it might be worth considering if a plain old function makes more sense

dktapps (approved):
PR changes look good to me wrt. cacheability, although it might be worth considering if a private static function would make more sense than a closure for larger functions

staabm:
@dktapps thank you

> Basically the requirement for being cached is that it has to be stateless

can this be measured somehow?

src/Command/CommandHelper.php
```diff
@@ -99,10 +99,10 @@ public static function begin(
 	{
 		$stdOutput = new SymfonyOutput($output, new SymfonyStyle(new ErrorsConsoleStyle($input, $output)));
 
-		$errorOutput = (static function () use ($input, $output): Output {
+		$errorOutput = (static function ($input, $output): Output {
```
iluuu1994: What's the purpose of this static closure in the first place? The other case has an include (so possibly there to hide local vars?) and does some control flow. Why not just drop the closure entirely?

iluuu1994 (commented):
> do you think it would make things faster if we transform such closures into e.g. __invoke classes or similar?

The static closure patch avoids an allocation each time the closure is used (not called). Allocating an object rather than a closure shouldn't make a significant difference, unless you can avoid allocating the object each time. For closures that have `use`es, that might be easier to achieve with `__invoke`able objects with properties I suppose.

dktapps:
> @dktapps thank you
> 
> > Basically the requirement for being cached is that it has to be stateless
> 
> can this be measured somehow?

Not sure what you mean by this, but for an explicitly static closure it just can't have used vars or local static vars.

The PR you mentioned on php-src is mostly focused on inferring static when it's not explicitly specified.
Inferring static won't guarantee cacheability, because a static closure could still `use` local scope vars or have local static variables.
But inferring static does _potentially_ make it possible to cache closures which aren't explicitly marked with `static`, if they by chance have no uses and no local static vars (a fairly common case seen a lot with array_map callbacks and such).

src/Command/CommandHelper.php
staabm: good point. removed this closure.

thank you.

staabm:
> The static closure patch avoids an allocation each time the closure is used (not called).

does this means, after making a closure cachable we might see a runtime improvement or memory_get_peak improvement?
or is this such a minor thing that changing code to be a cachable closure will kindly be noticable in perf metrics?
(making this more efficient is my primary motivation for such changes atm)

phpstan-baseline.neon
```diff
@@ -72,6 +72,36 @@ parameters:
 			count: 1
 			path: src/Analyser/RuleErrorTransformer.php
 
+		-
+			rawMessage: Doing instanceof PHPStan\Type\IntersectionType is error-prone and deprecated.
```
staabm: @ondrejmirtes I wonder why this errors show up now, but it seems they were not reported when 1:1 the same code was part of MutatingScope class. I could not find ignore-error rules or something which would swallow them either.

---

besides that, I think this PR is ready to merge.
I would do more *Traverser-extractions in separate PRs, as we have too many of them.

dktapps:
> > The static closure patch avoids an allocation each time the closure is used (not called).
> 
> does this means, after making a closure cachable we might see a runtime improvement or memory_get_peak improvement? or is this such a minor thing that changing code to be a cachable closure will kindly be noticable in perf metrics? (making this more efficient is my primary motivation for such changes atm)

It depends on how hot the code is, you might see a performance improvement. I wouldn't expect to see much difference in memory usage

iluuu1994:
> does this means, after making a closure cachable we might see a runtime improvement or memory_get_peak improvement?

That's the idea. It should be beneficial for code that creates closures in a loop. Of course, you could already avoid this performance hit by moving the closure out of the loop, but the idea here is of course is that it happens automatically.

Regarding `__invoke`: Make sure to benchmark. Performance can be unpredictable.

phpstan-baseline.neon
VincentLanglet: The error was ignored inside TypeTraverser::map https://github.com/phpstan/phpstan-src/blob/0e00d4b38443b80765f75719d7781c96a65e310f/src/Rules/Api/ApiInstanceofTypeRule.php#L116 and you move the code inside into specific classes

VincentLanglet: Since all the instanceOfType error were automatically ignored by the rule with the code https://github.com/phpstan/phpstan-src/blob/0e00d4b38443b80765f75719d7781c96a65e310f/src/Rules/Api/ApiInstanceofTypeRule.php#L116 I feel like we should find a similar way to ignore all of them together:
- Either in the ApiInstanceofTypeRule
- Or just in the phpstan config/baseline by ignoring 
```
     identifier: phpstanApi.instanceofType
     path: src/Analayser/Traverser
```

because it's feel odd having to add new error to the baseline when moving this code.

staabm: added a interface for the type-traversers.

that way we can also get rid of magic `__invoke`, which can [be a performance footgun](https://github.com/phpstan/phpstan-src/pull/4724#issuecomment-3715513347)

staabm:
(since this is pure refactoring, I want to ignore mutation at this time)

ondrejmirtes (changes requested):
I generally like this, but we could remove TypeTraverserCallable altogether and have the new classes just have `__invoke`?

staabm:
> I generally like this, but we could remove TypeTraverserCallable altogether and have the new classes just have `__invoke`?

[I did the inverse](https://github.com/phpstan/phpstan-src/pull/4724/commits/eadccaac0b2905dc1a506e78403957d8ec6051ee) as otherwise the ApiInstanceOfRule reported errors, [see discussion thread](https://github.com/phpstan/phpstan-src/pull/4724#discussion_r2664274211)

ondrejmirtes:
Ah, makes sense!

ondrejmirtes (approved):

ondrejmirtes:
Would be really nice to also extract the huge callback in FileAnalyser 😊

ondrejmirtes:
Thank you!

PR #4728, opened 2026-01-06
NodeScopeResolver: cheap checks first

staabm:


src/Analyser/NodeScopeResolver.php
```diff
@@ -946,13 +946,13 @@ private function processStmtNode(
...
 				$nodeCallback($node, $scope);
-				if ($scope->getAnonymousFunctionReflection() !== $currentScope->getAnonymousFunctionReflection()) {
+				if (!$node instanceof VariableAssignNode && !$node instanceof PropertyAssignNode) {
 					return;
 				}
-				if ($scope->getFunction() !== $currentScope->getFunction()) {
+				if ($scope->getAnonymousFunctionReflection() !== $currentScope->getAnonymousFunctionReflection()) {
 					return;
 				}
-				if (!$node instanceof VariableAssignNode && !$node instanceof PropertyAssignNode) {
```
staabm, line 955: re-ordered checks.. instanceof first before method calls

src/Analyser/NodeScopeResolver.php
```diff
@@ -1272,8 +1272,9 @@ private function processStmtNode(
 			}
 
 			$originalStorage = $storage;
-			$storage = $originalStorage->duplicate();
 			if ($context->isTopLevel()) {
+				$storage = $originalStorage->duplicate();
```
staabm, line 1276: duplicate only if we actually change the storage

src/Analyser/NodeScopeResolver.php
```diff
@@ -1337,9 +1338,9 @@ private function processStmtNode(
 				&& count($scopesWithIterableValueType) > 0
 				&& !$continueExitPointHasUnoriginalKeyType
 				&& $stmt->keyVar !== null
+				&& (!$hasExpr->no() || !$stmt->expr instanceof Variable)
```
staabm, line 1341: simple calls before Type system calls

ondrejmirtes:
Thank you!

PR #4729, opened 2026-01-06
NodeScopeResolver: duplicate only if we actually change the storage

staabm:
[review with whitespaces ignored](https://github.com/phpstan/phpstan-src/pull/4729/files?w=1)


ondrejmirtes:
Thank you!

PR #4730, opened 2026-01-07
FileTypeMapper: Don't deep traverse method/function/hook bodies

staabm:
in case we would know that a certain file does not contain any inline functions/class-definitions, we could skip deep traverse method/function/hook bodies.

testing suggests, this would save us 10 MB of memory when running


staabm:
this PR doesn't work as is, since I did not find yet a way how we can cheaply identify whether a file contains inline classes/functions. any idea whether/how this can be checked?

ondrejmirtes:
Can we actually skip it? I think we still need to find all inline `@var` and `@throws`...

But there's a lot of time spent in FileTypeMapper, maybe there are ways to optimize it.

PR #4732, opened 2026-01-08
Added native return types

staabm:
these are the only nodes which used phpdoc return types in this folder.

changed classes are `final`, so I have no BC concerns


ondrejmirtes:
Thank you!

PR #4689, opened 2025-12-31
cs: import `array_last` function

paulbalandan:
I'm on PHP 8.5.1 and running `make cs` currently fails because of unimported `array_last` function.
<details>
<summary>Output of <code>make cs</code></summary>
<code><pre>
git clone https://github.com/phpstan/build-cs.git || true
fatal: destination path 'build-cs' already exists and is not an empty directory.
git -C build-cs fetch origin && git -C build-cs reset --hard origin/2.x
HEAD is now at b37678c Update dependency slevomat/coding-standard to v8.26.0
composer install --working-dir build-cs
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Generating autoload files
3 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
XDEBUG_MODE=off php build-cs/vendor/bin/phpcs
............................................................   60 / 2178 (3%)
............................................................  120 / 2178 (6%)
..................E..............E.E........................  180 / 2178 (8%)
............................................................  240 / 2178 (11%)
............................................................  300 / 2178 (14%)
............................................................  360 / 2178 (17%)
............................................................  420 / 2178 (19%)
...............................E............................  480 / 2178 (22%)
............................................................  540 / 2178 (25%)
..............E.E..........E................................  600 / 2178 (28%)
..............................................E.............  660 / 2178 (30%)
............................................................  720 / 2178 (33%)
............................................................  780 / 2178 (36%)
............................................................  840 / 2178 (39%)
............................................................  900 / 2178 (41%)
............................................................  960 / 2178 (44%)
............................................................ 1020 / 2178 (47%)
...........................................................E 1080 / 2178 (50%)
............................................................ 1140 / 2178 (52%)
............................................................ 1200 / 2178 (55%)
...........................................E................ 1260 / 2178 (58%)
............................................................ 1320 / 2178 (61%)
............................................................ 1380 / 2178 (63%)
...E........................................................ 1440 / 2178 (66%)
......E..................................................... 1500 / 2178 (69%)
........................E................................... 1560 / 2178 (72%)
............................................................ 1620 / 2178 (74%)
............................................................ 1680 / 2178 (77%)
............................................................ 1740 / 2178 (80%)
............................................................ 1800 / 2178 (83%)
............................................................ 1860 / 2178 (85%)
............................................................ 1920 / 2178 (88%)
............................................................ 1980 / 2178 (91%)
............................................................ 2040 / 2178 (94%)
[...]

PR #4734, opened 2026-01-08
Run `make cs-fix` on PHP 8.5

takaram:
This is essentially revival of #4643 and #4689.

On current 2.1.x branch, `make cs` fails with below error only on PHP 8.5.
> Function array_last() should not be referenced via a fallback global name, but via a use statement.

This is because `SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly` sniff uses `function_exists` to decide if it checks the function calls.
On PHP 8.4 or lower, `array_last()` does not exist so the sniff doesn't check the usage of the function.

https://github.com/slevomat/coding-standard/blob/90b688162eba700bebca1f045ef5fe6f1936bc15/SlevomatCodingStandard/Sniffs/Namespaces/ReferenceUsedNamesOnlySniff.php#L207-L209

In this PR those errors are fixed with `make cs-fix`, and PHP_CodeSniffer are now run on CI with PHP 8.5.

With this fix, PHP-Scoper mistakenly prefixes `array_last` because it doesn't know the latest functions of PHP.
(Reported as humbug/php-scoper#1142)
As a workaround, `jetbrains/phpstorm-stubs` was added to `compiler/box/composer.json` so we can use `dev-master` version. The package is used by PHP-Scoper as the list of internal functions.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -77,7 +77,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "8.2"
+          php-version: "8.5"
 
       - name: "Validate Composer"
         run: "composer validate"
--- a/compiler/box/composer.json
+++ b/compiler/box/composer.json
@@ -1,7 +1,8 @@
 {
 	"require": {
 		"humbug/box": "^4.6",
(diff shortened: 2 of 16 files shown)


ondrejmirtes:
Perfect, thank you!

PR #4735, opened 2026-01-09
Optimize keepVoidType

staabm:
followup to https://github.com/phpstan/phpstan-src/commit/ff3922084b90aa4ca53dcc1f4037fb00758cdde9 but also covering NullsafeMethodCall

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-6720b.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-6720b.php
@@ -1,0 +1,15 @@
+<?php declare(strict_types = 1);
+
+namespace Bug6720b;
+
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Functions/data/bug-6720b.php
```diff
@@ -0,0 +1,15 @@
+<?php declare(strict_types = 1);
+
+namespace Bug6720b;
```
staabm, line 3: its not fixing a bug, this test just covers the nullsafe case, which wasn't tested before

ondrejmirtes:
Thank you!

PR #4737, opened 2026-01-09
Find all constant/method types per class at once

staabm:
before this PR we read/ast-traversed every stub file only until we found the single symbol we were searching for.
after something was found, traversal stopped and we returned early.

this means we e.g. see the same file beeing read/php-parsed/ast-traversed from the filesystem multiple times, when e.g. multiple constants of the same class are unkown or a mix of constants and methods are unknown.

see a phpstan-doctrine example run:
```
phar:///Users/staabm/workspace/phpstan-doctrine/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/SignatureMap/../../../vendor/phpstan/php-8-stubs/stubs/ext/pdo/PDO.stub
phar:///Users/staabm/workspace/phpstan-doctrine/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/SignatureMap/../../../vendor/phpstan/php-8-stubs/stubs/ext/pdo/PDO.stub
phar:///Users/staabm/workspace/phpstan-doctrine/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/SignatureMap/../../../vendor/phpstan/php-8-stubs/stubs/ext/pdo/PDO.stub
phar:///Users/staabm/workspace/phpstan-doctrine/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/SignatureMap/../../../vendor/phpstan/php-8-stubs/stubs/ext/pdo/PDO.stub
phar:///Users/staabm/workspace/phpstan-doctrine/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/SignatureMap/../../../vendor/phpstan/php-8-stubs/stubs/ext/pdo/PDO.stub
```

---

after this PR we remember all contained symbols within a stub file on the first read/parse/ast-traversal.
when later requested we can easily lookup whether a stub for a given method or constant exists, without the need to read/parse the stub file over and overr again.

### DIFF
--- a/src/Reflection/SignatureMap/Php8SignatureMapProvider.php
+++ b/src/Reflection/SignatureMap/Php8SignatureMapProvider.php
@@ -44,14 +44,8 @@ final class Php8SignatureMapProvider implements SignatureMapProvider
 	/** @var array<lowercase-string, array<lowercase-string, array{ClassMethod, string}>> */
 	private array $methodNodes = [];
 
-	/** @var array<lowercase-string, array<lowercase-string, array{ClassMethod, string}>> */
-	private array $stubbedMethodNodes = [];
-
 	/** @var array<lowercase-string, array<lowercase-string, Type|null>> */
 	private array $constantTypes = [];
-
-	/** @var array<lowercase-string, array<lowercase-string, Type|null>> */
-	private array $stubbedConstantTypes = [];
 
 	private Php8StubsMap $map;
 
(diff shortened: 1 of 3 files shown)


staabm:
hmm maybe this caching should happen in `FileNodesFetcher` instead, so all consumers will benefit from it.
will give it a try

ondrejmirtes:
No, I don't think so. Please wait until I look at the code.

src/Reflection/SignatureMap/Php8SignatureMapProvider.php
```diff
@@ -41,12 +41,18 @@ final class Php8SignatureMapProvider implements SignatureMapProvider
 
 	private const DIRECTORY = __DIR__ . '/../../../vendor/phpstan/php-8-stubs';
 
-	/** @var array<string, array<string, array{ClassMethod, string}>> */
+	/** @var array<lowercase-string, array<lowercase-string, array{ClassMethod, string}>> */
 	private array $methodNodes = [];
 
-	/** @var array<string, array<string, Type|null>> */
+	/** @var array<lowercase-string, array<lowercase-string, array{ClassMethod, string}>> */
+	private array $stubbedMethodNodes = [];
```
ondrejmirtes: I don't follow, what's the difference between methodNodes and stubbedMethodNodes?

I'm probably slower than usual, you know, Friday morning, heavy snowing outside etc. 😊

ondrejmirtes (changes requested):

src/Reflection/SignatureMap/Php8SignatureMapProvider.php
staabm: I had the impression based on the previous implementation that we should only remember the requested types.

but it seems thats not the case. simplified it now

ondrejmirtes:
Thank you!

PR #4740, opened 2026-01-11
Implement Type->getEnumCaseObject()

staabm:

---

before this PR: 49.23 seconds
after this PR: 41.09 seconds

profile before PR:

grafik

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -209,8 +209,8 @@ final class TypeCombinator
 			} elseif ($types[$i]->isString()->yes() && $types[$i]->isClassString()->no() && TypeUtils::getAccessoryTypes($types[$i]) === []) {
 				$hasGenericScalarTypes[ConstantStringType::class] = true;
 			} else {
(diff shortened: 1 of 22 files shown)


src/Type/TypeCombinator.php
```diff
@@ -209,8 +209,8 @@ public static function union(Type ...$types): Type
 			} elseif ($types[$i]->isString()->yes() && $types[$i]->isClassString()->no() && TypeUtils::getAccessoryTypes($types[$i]) === []) {
 				$hasGenericScalarTypes[ConstantStringType::class] = true;
 			} else {
-				$enumCases = $types[$i]->getEnumCases();
-				if (count($enumCases) === 1) {
+				$enumCase = $types[$i]->getEnumCaseObject();
+				if ($enumCase !== null) {
```
staabm, line 213: essence of this PR is here.

we move this operation into the type classes, in which we can calculate this single object more efficient

ondrejmirtes:
Nice, thank you!

PR #4744, opened 2026-01-12
IntersectionType: remove unnecessary hashing

staabm:

---

before this PR: 39.30 seconds
after this PR: 32.32 seconds

### DIFF
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -54,7 +54,6 @@ use function implode;
 use function in_array;
 use function is_int;
 use function ksort;
-use function md5;
 use function sprintf;
 use function strcasecmp;
 use function strlen;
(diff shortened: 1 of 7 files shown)


src/Type/IntersectionType.php
```diff
@@ -1346,7 +1345,7 @@ public function getFiniteTypes(): array
 		foreach ($this->types as $type) {
 			$oneType = [];
 			foreach ($type->getFiniteTypes() as $finiteType) {
-				$oneType[md5($finiteType->describe(VerbosityLevel::typeOnly()))] = $finiteType;
```
staabm, line 1349: this line fixed the perf problem. changes in the other files just apply the same idea but do not affect the benchmark

ondrejmirtes:
Let's keep the change to just Type implementations please, unless there's a benchmark proving a performance issue.

In anonymous classes I want to keep the original name because it might show up somewhere in baselines and it's more obvious the hashed name is a "fake name" of the class.

For FileTypeMapper also please keep the hash as it would create conflicts with my branch I'm currently working on.

Thanks!

ondrejmirtes:
Thank you!

PR #4746, opened 2026-01-12
Faster IntersectionType->getFiniteTypes()

staabm:

---

requires https://github.com/phpstan/phpstan-src/pull/4744

---

before this PR: 32.32 seconds
after this PR: 8.57 seconds

### DIFF
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -54,6 +54,7 @@ use function implode;
 use function in_array;
 use function is_int;
 use function ksort;
+use function md5;
 use function sprintf;
 use function strcasecmp;
 use function strlen;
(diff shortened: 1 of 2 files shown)


src/Type/IntersectionType.php
```diff
@@ -1345,6 +1345,10 @@ public function getFiniteTypes(): array
 		foreach ($this->types as $type) {
 			$oneType = [];
 			foreach ($type->getFiniteTypes() as $finiteType) {
+				if ($finiteType instanceof EnumCaseObjectType) {
+					$oneType[$finiteType->getClassName() . '::' . $finiteType->getEnumCaseName()] = $finiteType;
+					continue;
+				}
```
staabm, line 1351: using the same type description as in https://github.com/phpstan/phpstan-src/blob/7c5aa631fe293c2975f43c5875557e4fe18085e1/src/Type/IntersectionType.php#L1027

ondrejmirtes:
Thank you!

staabm:
@ondrejmirtes since the repro from https://github.com/phpstan/phpstan/issues/11968 is now below 9 seconds, it might be worth adding it as a AnalyserIntegrationTest (before we only had a stripped down version of it in the test-suite, because the original case was still too slow with 49 seconds)

wdyt?

ondrejmirtes:
I feel like in the longterm, we need a different approach to performance tests, not just shove them into AnalyserIntegrationTest. I don't want to slow down the main test suite because of them, I think that fixing bugs != performance work. I think that out of those 200 tests in AnalyserIntegrationTest, we should pick the ones about performance and have a separate job for them in GitHub Actions, maybe with phpbench, that would reliably detect when it gets slower.

PR #4750, opened 2026-01-13
MemoizingReflectionProvider->hasClass() is case-insensitive

staabm:
since `MemoizingReflectionProvider->getClass()`  and `MemoizingReflectionProvider->getClassName()` work on a lowercased string, I think `MemoizingReflectionProvider->hasClass()` should work the same way to provide consistent results

### DIFF
--- a/src/Reflection/ReflectionProvider/MemoizingReflectionProvider.php
+++ b/src/Reflection/ReflectionProvider/MemoizingReflectionProvider.php
@@ -14,13 +14,13 @@ use function strtolower;
 final class MemoizingReflectionProvider implements ReflectionProvider
 {
 
-	/** @var array<string, bool> */
+	/** @var array<lowercase-string, bool> */
 	private array $hasClasses = [];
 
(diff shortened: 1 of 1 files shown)


staabm:
the out-of-memory errors are unrelated to this change.

I had a look into it, and it seems most of the memory regression we currently see on 2.1.x is related to a increase in resultCache.php size

grafik

the biggest part of the cache might not even be related to phpstan-src itself: https://github.com/shipmonk-rnd/dead-code-detector/issues/287 (removing dead-code-detector shrinks the resultCache.php by ~65%)

https://blackfire.io/profiles/90337c83-aba6-4c80-853d-547f0ac566f2/graph

staabm:
the above comment only holds for the main process alone.

running with `--debug` and therefore without workers the memory profile looks like:

grafik

https://blackfire.io/profiles/25009815-b33e-41f1-87d0-3cd594971d8e/graph

ondrejmirtes:
Thank you!

PR #4754, opened 2026-01-13
Implement rule for `unserialize`

eliashaeussler:
This PR implements a new rule for [`unserialize`](https://www.php.net/manual/en/function.unserialize.php). Following behaviors will trigger an error:

* Parameter $2 options is NOT set (only with parameter `checkInsecureUnserialize: true`)
* Parameter $2 options has an invalid array key (neither `allowed_classes` nor `max_depth`)
* Parameter $2 options has invalid type for `allowed_classes`
* Parameter $2 options has `'allowed_classes' => true` (only with parameter `checkInsecureUnserialize: true`)
* Parameter $2 options has invalid array item for `allowed_classes`
* Parameter $2 options has `max_depth` key on PHP < 7.4
* Parameter $2 options has invalid type for `max_depth`
* Parameter $2 options is set, but does not have `allowed_classes` configured (only with parameter `checkInsecureUnserialize: true`)

### DIFF
--- a/tests/PHPStan/Rules/Functions/UnserializeRuleTest.php
+++ b/tests/PHPStan/Rules/Functions/UnserializeRuleTest.php
@@ -1,0 +1,75 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Functions;
+
+use PHPStan\Php\PhpVersion;
+use PHPStan\Rules\Rule;
+use PHPStan\Testing\RuleTestCase;
+use PHPUnit\Framework\Attributes\RequiresPhp;
+use const PHP_VERSION_ID;
+
+/**
+ * @extends RuleTestCase<UnserializeRule>
+ */
+class UnserializeRuleTest extends RuleTestCase
+{
+
+	protected function getRule(): Rule
+	{
+		return new UnserializeRule(new PhpVersion(PHP_VERSION_ID), self::createReflectionProvider(), true);
+	}
+
+	public function testFile(): void
+	{
+		$expectedErrors = [
+			[
+				'Parameter #2 $options to function unserialize contains an invalid value for "allowed_classes" item #1.',
+				5,
+			],
+			[
+				'Parameter #2 $options to function unserialize contains an invalid value null for "allowed_classes".',
+				7,
+			],
+			[
+				'Parameter #2 $options to function unserialize contains an invalid value null for "max_depth".',
+				9,
+			],
+			[
+				'Parameter #2 $options to function unserialize must be present with "allowed_classes" set to false or a list of allowed class names.',
+				9,
+			],
+			[
+				'Parameter #2 $options to function unserialize contains unsupported option "foo".',
+				11,
+			],
+			[
+				'Parameter #2 $options to function unserialize must be present with "allowed_classes" set to false or a list of allowed class names.',
+				11,
+			],
+			[
+				'Parameter #2 $options to function unserialize must either be false or a list of allowed class names.',
+				13,
+			],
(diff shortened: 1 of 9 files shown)


tests/PHPStan/Rules/Functions/UnserializeRuleTest.php
```diff
@@ -0,0 +1,75 @@
...
+			[
+				'Calling unserialize() without parameter $2 options and "allowed_classes" set to false or a list of allowed class names is insecure.',
+				15,
+			],
+		];
+
+		$this->analyse([__DIR__ . '/data/unserialize.php'], $expectedErrors);
+	}
+
+	#[RequiresPhp('< 7.4')]
```
eliashaeussler, line 62: I'm not sure whether to keep this test case, since it does not seem like PHP < 7.4 is actually included in the test matrix.

ondrejmirtes (changes requested):
Two things:

1) This is a new rule, should only be enabled with bleeding edge using conditionalTags in config.
2) Some things this rule reports should be achieved by changing resources/functionMap files (the deltas) - this will make existing rule to pick up the errors. I'm thinking about:

* Parameter #2 $options to function unserialize contains an invalid value null for "max_depth"
* Parameter #2 $options to function unserialize contains unsupported option "foo".
* Parameter #2 $options to function unserialize contains an option "max_depth" which is not supported by this PHP version.

eliashaeussler:
> 1. This is a new rule, should only be enabled with bleeding edge using conditionalTags in config.

Done – I hope I did it the correct way. Is it okay to have the new rule in level 5? 

> 2. Some things this rule reports should be achieved by changing resources/functionMap files (the deltas) - this will make existing rule to pick up the errors. I'm thinking about:
> 
> * Parameter #2 $options to function unserialize contains an invalid value null for "max_depth"
> * Parameter #2 $options to function unserialize contains unsupported option "foo".
> * Parameter #2 $options to function unserialize contains an option "max_depth" which is not supported by this PHP version.

I'm afraid I need a little guidance here. What do you mean with "changing resources/functionMap files (the deltas)"?

eliashaeussler:
I guess I know what you mean. There's already a definition for `unserialize` in the `functionMap` file:

https://github.com/phpstan/phpstan-src/blob/2.1.39/resources/functionMap.php#L11280

However, I think it should be slightly adapted:

```diff
diff --git a/resources/functionMap.php b/resources/functionMap.php
index 88f32cc38..a661c7521 100644
--- a/resources/functionMap.php
+++ b/resources/functionMap.php
@@ -11277,7 +11277,7 @@ return [
 'unlink' => ['bool', 'filename'=>'string', 'context='=>'resource'],
 'unpack' => ['array|false', 'format'=>'string', 'data'=>'string', 'offset='=>'int'],
 'unregister_tick_function' => ['void', 'function_name'=>'callable'],
-'unserialize' => ['mixed', 'variable_representation'=>'string', 'allowed_classes='=>'array{allowed_classes?:string[]|bool}'],
+'unserialize' => ['mixed', 'data'=>'string', 'options='=>'array{allowed_classes?:string[]|bool,max_depth?:non-negative-int}'],
 'untaint' => ['bool', '&rw_string'=>'string', '&...rw_strings='=>'string'],
 'uopz_add_function' => ['bool', 'class'=>'string', 'function'=>'string', 'handler'=>'Closure', '$flags'=>'int', '$all'=>'bool'],
 'uopz_add_function\'1' => ['bool', 'function'=>'string', 'handler'=>'Closure', '$flags'=>'int'],
```

PR #4758, opened 2026-01-14
FileAnalyser: Move repetitive work out of hot loop

staabm:


src/Analyser/FileAnalyser.php
```diff
@@ -171,7 +172,7 @@ public function analyseFile(
 							$error = $this->ruleErrorTransformer->transform($ruleError, $scope, $parserNodes, $node);
 
 							if ($error->canBeIgnored()) {
-								foreach ($this->ignoreErrorExtensionProvider->getExtensions() as $ignoreErrorExtension) {
```
staabm, line 174: this method was invoked once for every error found

ondrejmirtes, line 174: Could make sense to move it completely out of the callback and just `use` its variable? Maybe measure the memory difference.

ondrejmirtes (changes requested):

src/Analyser/FileAnalyser.php
staabm: I cannot see a measurable improvement, even though this loop callback is invoked 800.000 times

ondrejmirtes:
Thank you!

PR #4759, opened 2026-01-15
Split MutatingScope->resolveType() into smaller methods

staabm:
to get better blackfire profiles, I have splitted the biggest portions of resolveType() method into separate methods.

no functional changes

before this PR profiles showed we are spending most of the time in resolveType() but not where exactly:
grafik

---

my thesis is, that time spent on resolving types highly depends on the specific expression. to ease finding meaningfull bottlenecks code is separated into more named and independent building blocks (methods), which show up separately in profiles (after PR):

grafik

(see less tiny "down arrows" below the resolveType() box, which means we see better how time distributes from there.

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -1653,59 +1653,9 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 		}
 
 		if ($node instanceof Expr\StaticCall) {
-			if ($node->name instanceof Node\Identifier) {
-				if ($this->nativeTypesPromoted) {
-					if ($node->class instanceof Name) {
-						$staticMethodCalledOnType = $this->resolveTypeByNameWithLateStaticBinding($node->class, $node->name);
-					} else {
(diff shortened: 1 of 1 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -2221,61 +1653,11 @@ static function (Node $node, Scope $scope) use ($arrowScope, &$arrowFunctionImpu
...
-				if ($callType === null) {
-					$callType = new ErrorType();
-				}
-
-				if ($node->class instanceof Expr) {
-					return $this->getNullsafeShortCircuitingType($node->class, $callType);
-				}
-
+			$callType = $this->getStaticCallType($node);
+			if ($callType !== null) {
```
staabm, line 1657: we could prevent this null-check pattern, but for now I translated the code 1:1.
it did not always return in this cases before but let it fall thru to the below `return new Mixed` at the very end of the method

ondrejmirtes:
Perfect, thank you!

PR #4760, opened 2026-01-15
Split MutatingScope->resolveType() into smaller methods - part 2

staabm:
continuation of https://github.com/phpstan/phpstan-src/pull/4759


ondrejmirtes:
Thank you!

PR #4761, opened 2026-01-15
Cache ClassReflections

staabm:
saves a small bit of runtime 0,3-0,5 seconds and also 1% memory usage reduction


ondrejmirtes:
Thank you!

PR #4764, opened 2026-01-15
Debug-Rules: cheap checks first

staabm:
prevent calls into reflection when $args does not contain the expected number of elements


ondrejmirtes:
Thank you!

PR #4765, opened 2026-01-15
UnionType: inline array_map() call

staabm:
the method is called often enough that this change leads to a 0,5% memory decrease 

before this PR
```
Used memory: 396.59 MB
```

after this PR
```
Used memory: 394.59 MB
```


ondrejmirtes:
Thank you!

PR #4766, opened 2026-01-15
Intersection of array&hasOffset is accepted by non-empty-array

staabm:
closes https://github.com/phpstan/phpstan/issues/13964

### DIFF
--- a/src/Type/Accessory/NonEmptyArrayType.php
+++ b/src/Type/Accessory/NonEmptyArrayType.php
@@ -76,10 +76,6 @@ class NonEmptyArrayType implements CompoundType, AccessoryType
 
 	public function accepts(Type $type, bool $strictTypes): AcceptsResult
 	{
-		if ($type instanceof CompoundType) {
-			return $type->isAcceptedBy($this, $strictTypes);
-		}
-
 		$isArray = $type->isArray();
 		$isIterableAtLeastOnce = $type->isIterableAtLeastOnce();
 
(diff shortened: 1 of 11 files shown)


staabm:
~~hmm alternatively it could be the other way arround. adding non-empty-array to the array when `array_key_exists()`~~

staabm:
~~stumbled over https://github.com/phpstan/phpstan-src/pull/3937 today .. need to have another look~~

ok, I think its fine.

src/Type/Accessory/NonEmptyArrayType.php
```diff
@@ -76,10 +76,6 @@ public function getConstantStrings(): array
 
 	public function accepts(Type $type, bool $strictTypes): AcceptsResult
 	{
-		if ($type instanceof CompoundType) {
-			return $type->isAcceptedBy($this, $strictTypes);
```
ondrejmirtes, line 80: I don't understand this. All Accessory types do this. So it has to stay.

But you can do another condition before this one, like AccessoryNonEmptyStringType does:

```php
	{
		if ($type->isNonEmptyString()->yes()) {
			return AcceptsResult::createYes();
		}
		if ($type instanceof CompoundType) {
			return $type->isAcceptedBy($this, $strictTypes);
		}

		return new AcceptsResult($type->isNonEmptyString(), []);
	}
```

You can also try doing the same thing (call related method before `instanceof CompoundType`) in other Accessory types so they all still look the same..

ondrejmirtes (changes requested):

src/Type/Accessory/NonEmptyArrayType.php
staabm: fixed, thanks

staabm: ~~I will look into other accessories in a separate PR~~

done

ondrejmirtes:
Thank you!

PR #4767, opened 2026-01-16
More optimizations

ondrejmirtes:


staabm:
applying this diff:

```diff
diff --git a/src/Analyser/LazyInternalScopeFactory.php b/src/Analyser/LazyInternalScopeFactory.php
index 611c11f598..64503833a5 100644
--- a/src/Analyser/LazyInternalScopeFactory.php
+++ b/src/Analyser/LazyInternalScopeFactory.php
@@ -59,6 +59,14 @@ final class LazyInternalScopeFactory implements InternalScopeFactory
                bool $nativeTypesPromoted = false,
        ): MutatingScope
        {
+               static $i = 0;
+               if ($i === 0) {
+                       register_shutdown_function(function () use (&$i) {
+                               var_dump($i);
+                       });
+               }
+               $i++;
```

I wonder why we only save very few scope creations after the `MutatingScope->mergeWith(...)` change.

before `MutatingScope->mergeWith(...)`: 598562
on https://github.com/phpstan/phpstan-src/pull/4767/commits/4634af2c097ca861c5e55231a750fc77b137fe9d : 597376

this might also be the reason why it does not really improve performance

staabm:
additionally I see nearly no `TypeCombinator::union` from within `ExpressionTypeHolder->and` with more than 2 types

ondrejmirtes:
I see no meaningful performance improvements here, only bugs :) I'll keep it open, you can keep looking into it. I am also surprised it doesn't improve anything. Maybe the current code is already crazily well optimized.

I suspect the line `if ($variableTypeHolder === $theirVariableTypeHolders[$exprString]) {` does a lot of heavy lifting here, but not sure.

staabm:
> Maybe the current code is already crazily well optimized.

I have a similar feeling that blackfire cannot point out a specific bootleneck anymore because we got most of the code equal fast

ondrejmirtes:
I'd expect the `TypeCombinator::union()` to be called with three types for this code:

```php
if (rand(0, 1)) {
    $a = 1;
} elseif (rand(0, 1)) {
    $a = 2;
} else {
    $a = 3;
}
```

If it's not the case then the optimization is not working as intended.

staabm:
its invoked 4 times

ondrejmirtes:
Abandoning this idea :)

PR #4768, opened 2026-01-16
Drop MutatingScope->rememberConstructorScope()

staabm:
the scope when processing methods already contains remembered expressions.
this means we don't need to re-remember expressions and can drop this unnecessary logic.

refs https://github.com/phpstan/phpstan-src/pull/3930


ondrejmirtes:
Thank you!

ondrejmirtes:
FYI I reverted this but added https://github.com/phpstan/phpstan-src/commit/78d946dacc0d8aede1508e5691f501436011c9d7, it makes more sense to me to do the logic once instead of each method.

PR #4769, opened 2026-01-16
Simplify TypeCombinator->processArrayTypes()

staabm:
removes some unnecessary work from the hot path

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -1109,12 +1109,12 @@ final class TypeCombinator
 
 	public static function intersect(Type ...$types): Type
 	{
-		$types = array_values($types);
-
 		$typesCount = count($types);
 		if ($typesCount === 0) {
 			return new NeverType();
 		}
+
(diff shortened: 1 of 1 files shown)


src/Type/TypeCombinator.php
```diff
@@ -1106,12 +1109,12 @@ private static function reduceArrays(array $constantArrays, bool $preserveTagged
...
 	public static function intersect(Type ...$types): Type
 	{
-		$types = array_values($types);
-
 		$typesCount = count($types);
 		if ($typesCount === 0) {
 			return new NeverType();
 		}
+
+		$types = array_values($types);
```
staabm, line 1117: hmm .. maybe we could even tag `TypeCombinator::union()` and `TypeCombinator::intersect()` as `no-named` and drop the `array_values` call?

ondrejmirtes: Yes, please do that next 👍

ondrejmirtes:
Thank you!

PR #4772, opened 2026-01-16
More precise types

staabm:
first found of fixes to get TypeCombinator::union() / intersect() ready for no-named-arguments

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -686,7 +686,7 @@ final class TypeCombinator
 
 	/**
 	 * @param Type[] $arrayTypes
-	 * @return Type[]
+	 * @return list<Type>
 	 */
 	private static function processArrayAccessoryTypes(array $arrayTypes): array
 	{
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -381,14 +381,14 @@ final class TypeCombinator
 
 			if ($tempTypes === []) {
 				if ($benevolentUnionObject instanceof TemplateBenevolentUnionType) {
-					return $benevolentUnionObject->withTypes($types);
+					return $benevolentUnionObject->withTypes(array_values($types));
 				}
 
-				return new BenevolentUnionType($types, true);
+				return new BenevolentUnionType(array_values($types), true);
 			}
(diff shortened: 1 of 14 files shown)


src/Type/UnionTypeHelper.php
```diff
@@ -16,8 +16,9 @@ final class UnionTypeHelper
 {
 
 	/**
-	 * @param Type[] $types
-	 * @return Type[]
+	 * @template T of Type
+	 * @param list<T>|array<T> $types
```
ondrejmirtes: `list<T>|array<T>` does not make sense, should be just `array<T>`. The return type condition would still work.

But could we instead always make sure that list is passed here?

ondrejmirtes (changes requested):

src/Type/UnionTypeHelper.php
staabm: > But could we instead always make sure that list is passed here?

to make that work I either need to ignore some errors, or narrow the constructor parameters of e.g. UnionType, IntersectionType etc. (which you might not want because BC break)

staabm: did the change, please have a look whether you like it

src/Type/TypeCombinator.php
```diff
@@ -380,14 +380,14 @@ public static function union(Type ...$types): Type
...
 				}
 
-				return new BenevolentUnionType($types, true);
+				return new BenevolentUnionType(array_values($types), true);
 			}
 		}
 
-		return new UnionType($types, true);
+		return new UnionType(array_values($types), true);
 	}
```
staabm, line 392: we might phpdoc-cast this parameter instead of doing a runtime `array_values` to reduce overhead

src/Type/UnionTypeHelper.php
ondrejmirtes: > or narrow the constructor parameters of e.g. UnionType, IntersectionType etc. (which you might not want because BC break)

I think it's okay because mostly people should be using TypeCombinator. It's only okay to do `new UnionType()` if you have the exact array and you know the types are already normalized.

ondrejmirtes:
Thank you!

PR #4773, opened 2026-01-17
TypeCombinator: Prevent unnecessary work

staabm:
micro optimization in a very hot path

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -494,54 +494,50 @@ final class TypeCombinator
 		if (
 			$a instanceof ConstantStringType
 		) {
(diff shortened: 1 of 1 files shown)


src/Type/TypeCombinator.php
```diff
@@ -494,54 +494,50 @@ private static function compareTypesInUnion(Type $a, Type $b): ?array
 		if (
 			$a instanceof ConstantStringType
 		) {
-			$description = $b->describe(VerbosityLevel::value());
```
staabm, line 497: describe $b only when `$a == ''`.

before we described $b no matter which constant-value $a had

ondrejmirtes:
Thank you!

PR #4775, opened 2026-01-17
test remembered types in inherited scopes

staabm:
re-add tests from https://github.com/phpstan/phpstan-src/pull/4768


ondrejmirtes:
Sorry, and thank you 😊

PR #4776, opened 2026-01-17
IntersectionType: Extract describeType() method

staabm:


src/Type/IntersectionType.php
```diff
@@ -304,50 +304,53 @@ public function equals(Type $type): bool
 	public function describe(VerbosityLevel $level): string
 	{
 		return $level->handle(
-			function () use ($level): string {
```
staabm, line 307: extract inline callable into separate method

ondrejmirtes:
Thank you!

PR #4777, opened 2026-01-17
Faster ExpressionTypeHolder->equals()

staabm:
since we have this [`return $this` shortcut in `and`](https://github.com/phpstan/phpstan-src/pull/4777/files#diff-2767b4b5b27bcfe03991183ee9839ced3a7b8c65a86899cf03e7b56605d6ad65R44), we can utilize it in the `equals` case


ondrejmirtes:
Thank you!

PR #4779, opened 2026-01-18
MutatingScope: remove always-true condition

staabm:
It seems this is unused


src/Analyser/MutatingScope.php
```diff
@@ -3928,9 +3928,6 @@ public function mergeWith(?self $otherScope): self
 			if (!$theirNativeExpressionTypes[$exprString]->equals($theirExpressionTypes[$exprString])) {
 				continue;
 			}
-			if (!array_key_exists($exprString, $mergedExpressionTypes)) {
-				continue;
-			}
```
ondrejmirtes, line 3933: To be it's not obvious that `$mergedExpressionTypes[$exprString]` exists. Is it obvious?

ondrejmirtes (changes requested):

PR #4780, opened 2026-01-18
MutatingScope: Remove unnecessary union

staabm:
I think this is the 3rd time this week, which made me think that I need https://github.com/phpstan/phpstan-src/pull/4766 to implement some other stuff.

Lets see how you feel about it :-)

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3343,16 +3343,7 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 			if ($dimType->isInteger()->yes() || $dimType->isString()->yes()) {
 				$exprVarType = $scope->getType($expr->var);
 				if (!$exprVarType instanceof MixedType && !$exprVarType->isArray()->no()) {
-					$types = [
-						new ArrayType(new MixedType(), new MixedType()),
-						new ObjectType(ArrayAccess::class),
-						new NullType(),
-					];
-					if ($dimType->isInteger()->yes()) {
-						$types[] = new StringType();
-					}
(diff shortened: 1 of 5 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -3343,16 +3343,7 @@ public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType,
...
-					$types = [
-						new ArrayType(new MixedType(), new MixedType()),
-						new ObjectType(ArrayAccess::class),
-						new NullType(),
-					];
-					if ($dimType->isInteger()->yes()) {
-						$types[] = new StringType();
-					}
-					$offsetValueType = TypeCombinator::intersect($exprVarType, TypeCombinator::union(...$types));
-
```
staabm, line 3355: the actual change is here.

I am not sure why this no longer makes a difference.. maybe related to Type->setExistingOffset() and friends

src/Type/Accessory/NonEmptyArrayType.php
```diff
@@ -76,10 +76,6 @@ public function getConstantStrings(): array
 
 	public function accepts(Type $type, bool $strictTypes): AcceptsResult
 	{
-		if ($type instanceof CompoundType) {
-			return $type->isAcceptedBy($this, $strictTypes);
-		}
-
```
staabm, line 90: this is https://github.com/phpstan/phpstan-src/pull/4766 which makes sense on its own

staabm:
ok - lets see what you think about it. in case you are fine with the change I can add regression tests for the issue-bot reports

PR #4781, opened 2026-01-18
Fix PHP7.4 build

staabm:
PHP7.4 does not support named arguments, therefore is triggering 

> Parameter... of array_values is already a list, call has no effect

which does not happen on PHP8+


ondrejmirtes:
Thank you!

PR #4782, opened 2026-01-18
Fix PHP8.0 build

staabm:
in https://github.com/phpstan/phpstan-src/pull/4781 I put the ignore into the wrong baseline file.

it made php8.x fail ;)


ondrejmirtes:
Thank you!

PR #4784, opened 2026-01-19
TrinaryLogic: Prevent unnecessary work

staabm:
prevent unnecessary method calls in a hot path

---

in `php bin/phpstan analyse src/Analyser/ src/Rules/ -v --debug`

`lazyMaxMin` calls `yes()` 1.1 million times

grafik


ondrejmirtes:
Thank you!

PR #4788, opened 2026-01-20
RuleTestCase: Use stable error order

staabm:
Sort errors, so tests don't need to be adjusted when run with FNSR or NSR, which might change order of errors.


ondrejmirtes:
I've got a better version locally and I'm working on it 😊

PR #4789, opened 2026-01-20
SplObjectStorage iterates over objects

staabm:
closes https://github.com/phpstan/phpstan/issues/13985
closes https://github.com/phpstan/phpstan/issues/14046

partly reverts https://github.com/phpstan/phpstan-src/pull/4761

### DIFF
--- a/src/Type/ObjectType.php
+++ b/src/Type/ObjectType.php
@@ -1735,21 +1735,20 @@ class ObjectType implements TypeWithClassName, SubtractableType
 
 	public function getClassReflection(): ?ClassReflection
 	{
-		if ($this->classReflection !== null) {
-			return $this->classReflection;
+		if ($this->classReflection === null) {
+			$reflectionProvider = ReflectionProviderStaticAccessor::getInstance();
+			if (!$reflectionProvider->hasClass($this->className)) {
+				return null;
+			}
+
+			$this->classReflection = $reflectionProvider->getClass($this->className);
 		}
 
-		$reflectionProvider = ReflectionProviderStaticAccessor::getInstance();
-		if (!$reflectionProvider->hasClass($this->className)) {
-			return null;
+		if ($this->classReflection->isGeneric()) {
+			return $this->classReflection->withTypes(array_values($this->classReflection->getTemplateTypeMap()->map(static fn (): Type => new ErrorType())->getTypes()));
 		}
 
-		$classReflection = $reflectionProvider->getClass($this->className);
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
I don't get it, why shouldn't it be cached?

staabm:
ClassReflection created via `withTypes` has properties like `isDeprecated`, `isGeneric`, `isInternal` set to `null` 

grafik

while the one created via `getClass` contains more information:

grafik


so by caching the `withTypes` created one, we have less information at hand later on when we respond to a `getClassReflection()` call with a cached object

staabm:
Hmm no. This doesn't make sense. Will have another look tomorrow

staabm:
had another look but still I am not yet sure about the causation.

the difference between caching and not caching is, that 
- when caching generics ClassReflection for `SplObjectStorage` the template-type map will be `ErrorType, ErrorType` which causes lots of `if instanceof ErrorType` logic to happen and fallbacks are getting evaluated
- when not caching generics ClassReflection for `SplObjectStorage` the template-type map will reflect the [unresolved types of the phpdoc stub](https://github.com/phpstan/phpstan-src/blob/9dc265fc2987e9f01330c264c4018ebd294e7869/stubs/SplObjectStorage.stub#L4-L11)

cs278 (approved):
Tested - fixes https://github.com/phpstan/phpstan/issues/14046 for me

src/Type/ObjectType.php
```diff
@@ -1746,7 +1746,7 @@ public function getClassReflection(): ?ClassReflection
 
 		$classReflection = $reflectionProvider->getClass($this->className);
```
VincentLanglet: Should we still cache the classReflection before the withType call ?

Maybe the code could be
```
if ($this->classReflection === null) {
      $reflectionProvider = ReflectionProviderStaticAccessor::getInstance();
		if (!$reflectionProvider->hasClass($this->className)) {
			return null;
		}
		
		$this->classReflection = $reflectionProvider->getClass($this->className);
}

if ($this->classReflection->isGeneric()) {
     return $this->classReflection->withTypes(array_values($classReflection->getTemplateTypeMap()->map(static fn (): Type => new ErrorType())->getTypes()));
}

return $this->classReflection;
```

staabm: good point. I think your suggestions makes the intent clearer.

VincentLanglet (approved):

PR #4791, opened 2026-01-20
Fix named argument parameter matching in getFunctionCallStackWithParameters

janedbal:
When processing function calls with named arguments, the parameter was incorrectly matched using positional index instead of argument name. This caused Scope::getFunctionCallStackWithParameters() to return the wrong parameter when:

1. Arguments were passed out of order using named syntax
2. A named argument was passed for an optional parameter while a required parameter was missing

The fix matches parameters by name for named arguments before falling back to positional matching.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -5542,14 +5542,26 @@ class NodeScopeResolver
 			$parameterType = null;
 			$parameterNativeType = null;
 			if ($parameters !== null) {
-				if (isset($parameters[$i])) {
-					$assignByReference = $parameters[$i]->passedByReference()->createsNewVariable();
-					$parameterType = $parameters[$i]->getType();
-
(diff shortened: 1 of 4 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -5542,14 +5542,26 @@ private function processArgs(
 			$parameterType = null;
 			$parameterNativeType = null;
 			if ($parameters !== null) {
-				if (isset($parameters[$i])) {
-					$assignByReference = $parameters[$i]->passedByReference()->createsNewVariable();
-					$parameterType = $parameters[$i]->getType();
+				$matchedParameter = null;
```
ondrejmirtes, line 5545: This fix is in a wrong place. `processArgs` gets `$normalizedExpr` which is output of ArgumentsNormalizer. The fix should be there.

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4792, opened 2026-01-20
Prevent unnecessary work around parametersAcceptor->getParameters() call-sites

staabm:
depending on the used reflection or acceptor type `getParameters()` can be expensive.
don't invoke it when we might return without using the result.
also use variables instead of calling it twice


ondrejmirtes:
Thank you!

PR #4794, opened 2026-01-20
Propose stricter types por SplFileInfo

devnix:
Those methods can return an empty string, but only if the instance is created with an empty string.

I propose these changes in bleeding edge, as it would break the compatibility in the constructor.

### DIFF
--- a/resources/functionMap_bleedingEdge.php
+++ b/resources/functionMap_bleedingEdge.php
@@ -3,6 +3,12 @@
 return [
 	'new' => [
 		'php_uname' => ['string', 'mode='=>'"a"|"s"|"n"|"r"|"v"|"m"'],
+		'SplFileInfo::__construct' => ['void', 'filename'=>'non-empty-string'],
+		'SplFileInfo::__toString' => ['non-empty-string'],
+		'SplFileInfo::getBasename' => ['non-empty-string', 'suffix='=>'string'],
+		'SplFileInfo::getFilename' => ['non-empty-string'],
+		'SplFileInfo::getPathname' => ['non-empty-string'],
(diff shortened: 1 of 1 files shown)


VincentLanglet:
> Those methods can return an empty string, but only if the instance is created with an empty string.

Technically that doesn't forbid passing a empty string to the constructor no (?)

devnix:
> Technically that doesn't forbid passing a empty string to the constructor no (?)

Yeah, that's my dichotomy: if static analysis "forbids" or "prevents" you from calling the constructor with an empty string, we could consider "safe" to infer that the functions should return a `non-empty-string` too? 

That's how it works in userland anyway: if you ignore the SAT warning in the constructor, the getters will have an invalid value at runtime.

VincentLanglet:
> That's how it works in userland anyway: if you ignore the SAT warning in the constructor, the getters will have an invalid value at runtime.

Which warning ?

I dont get one with `new SplFileInfo('')`.
https://onlinephp.io/c/386d9

devnix:
I mean the error message that PHPStan would throw because of this line: https://github.com/phpstan/phpstan-src/pull/4794/changes#diff-cf5b76484540219ae620db261e215bda6b16a613fdb72eede0befb5c55fec466R6

VincentLanglet:
> I mean the error message that PHPStan would throw because of this line: [#4794 (changes)](https://github.com/phpstan/phpstan-src/pull/4794/changes#diff-cf5b76484540219ae620db261e215bda6b16a613fdb72eede0befb5c55fec466R6)

Yes why would PHPStan introduce such error when it's permitted by PHP itself ?

PR #4458, opened 2025-10-19
Revert "Cleanup `instanceof Type` checks"

takaram:
This partially reverts 3b72271648184ebc0cc3fa215ac355e6315f10cc (#4380).

Closes phpstan/phpstan#13694

PR #4796, opened 2026-01-21
Added regression test

staabm:
closes https://github.com/phpstan/phpstan/issues/13694
closes https://github.com/phpstan/phpstan-src/pull/4458


ondrejmirtes:
What commit fixed it? Does the test fail without the commit?

staabm:
the error starts to disappear with  https://github.com/phpstan/phpstan-src/commit/70dcdfaf6ead7f9939490e0ebb4edffa1d004147 (but the test doesn't fail when reverting this commit; likely because the actual change no longer exist in the codebase in the same form)

ondrejmirtes:
Thank you!

PR #4797, opened 2026-01-21
Fix invokable objects incorrectly labeled as instances of `Closure`

HypeMC:
Closes https://github.com/phpstan/phpstan/issues/13975

### DIFF
--- a/src/Type/Accessory/HasMethodType.php
+++ b/src/Type/Accessory/HasMethodType.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type\Accessory;
 
+use Closure;
 use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\TypeNode;
 use PHPStan\Reflection\ClassMemberAccessAnswerer;
@@ -17,6 +18,7 @@ use PHPStan\Type\Enum\EnumCaseObjectType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\IsSuperTypeOfResult;
+use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Traits\NonGeneralizableTypeTrait;
 use PHPStan\Type\Traits\NonGenericTypeTrait;
@@ -83,7 +85,11 @@ class HasMethodType implements AccessoryType, CompoundType
 			return $otherType->isSuperTypeOf($this);
 		}
 
-		if ($this->isCallable()->yes() && $otherType->isCallable()->yes()) {
+		if (
+			$this->isCallable()->yes()
+			&& $otherType->isCallable()->yes()
+			&& !(new ObjectType(Closure::class))->isSuperTypeOf($otherType)->yes()
+		) {
 			return IsSuperTypeOfResult::createYes();
 		}
 
--- a/tests/PHPStan/Rules/Classes/ImpossibleInstanceOfRuleTest.php
+++ b/tests/PHPStan/Rules/Classes/ImpossibleInstanceOfRuleTest.php
@@ -569,4 +569,17 @@ class ImpossibleInstanceOfRuleTest extends RuleTestCase
 		]);
 	}
 
+	public static function dataBug13975(): iterable
+	{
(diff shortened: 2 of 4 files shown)


src/Type/Accessory/HasMethodType.php
```diff
@@ -83,7 +84,11 @@ public function isSubTypeOf(Type $otherType): IsSuperTypeOfResult
 			return $otherType->isSuperTypeOf($this);
 		}
 
-		if ($this->isCallable()->yes() && $otherType->isCallable()->yes()) {
+		if (
+			$this->isCallable()->yes()
+			&& $otherType->isCallable()->yes()
+			&& !($otherType->isObject()->yes() && $otherType->isInstanceOf(Closure::class)->yes())
```
HypeMC: @ondrejmirtes Hi, I'm not sure how to proceed here. I tried using `instanceof`, but PHPStan said I should use `isObject()` instead. Now I'm getting this error. Any advice?

staabm: see other similar spots in the codebase.
it need to look like `(new ObjectType(Closure::class))->isSuperTypeOf(...)`

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -1625,6 +1625,12 @@ public function testBug13980(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testBug13975(): void
```
staabm: your test needs to be moved into a *RuleTest which relates to the error message you no longer want to have.

AnalyzerIntegrationTest is reserved for performance and crash-tests

src/Type/Accessory/HasMethodType.php
HypeMC: @staabm Thank you for the help, it seems ok now.

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
HypeMC: Done

staabm: I guess an additional test like https://phpstan.org/r/3cdd0efb-d240-49dd-9073-d1ada4d9843e might kill the mutant

staabm: it seems you lost the original test-case while adding the new one.

HypeMC: Oh sorry, I misunderstood you, fixed

tests/PHPStan/Rules/Classes/data/bug-13975-2.php
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types=1);
+
+namespace Bug139752;
```
staabm: we should not use a bug number of a different problem

staabm (approved):
one nit - LGTM otherwise.

thank you

tests/PHPStan/Rules/Classes/data/bug-13975-2.php
ondrejmirtes: This number seems high 😀

tests/PHPStan/Rules/Classes/data/bug-13975-1.php
```diff
@@ -0,0 +1,18 @@
+<?php declare(strict_types=1);
+
+namespace Bug139751;
```
staabm: to make sure we don't clash with a future bug number

```suggestion
namespace Bug13975a;
```

tests/PHPStan/Rules/Classes/data/bug-13975-2.php
staabm: ha, you are right. now I see it :)

```suggestion
namespace Bug13975b;
```

HypeMC: :man_facepalming: Of course, [fixed](https://github.com/phpstan/phpstan-src/compare/17063fb7033299cb9c49c5a50773ec3bc3112b32..bb2f85270ded51053ec229d70a24a0bf9835bd67).

staabm (approved):
thanks

VincentLanglet (approved):

HypeMC:
@staabm Thank you for the help :smile:

PR #4798, opened 2026-01-21
Fix broken `array_count_values` array key inference

staabm:
closes https://github.com/phpstan/phpstan/issues/13996

### DIFF
--- a/src/Type/Php/ArrayCountValuesDynamicReturnTypeExtension.php
+++ b/src/Type/Php/ArrayCountValuesDynamicReturnTypeExtension.php
@@ -54,6 +54,12 @@ final class ArrayCountValuesDynamicReturnTypeExtension implements DynamicFunctio
 				continue;
 			}
 
+			if (!$itemType->isString()->no() && !$itemType->isConstantScalarValue()->yes()) {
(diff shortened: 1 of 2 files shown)


src/Type/Php/ArrayCountValuesDynamicReturnTypeExtension.php
```diff
@@ -54,6 +54,12 @@ public function getTypeFromFunctionCall(
 				continue;
 			}
 
+			if (!$itemType->isString()->no() && !$itemType->isConstantScalarValue()->yes()) {
```
VincentLanglet, line 57: I feel like you're creating lot of weird inferences:
- `'foo'|int will be resume to `string|int`
- 'non-empty-string|int' lose the non-empty accessory (same for non-falsy)

Same for `'foo'|positive-int` or 'non-falsy-string|positive-int' ; if later casted to array they both are non-falsy while it's not the case if they are resumed to a simple int|string.

Then I'm unsure if this fix is doing more good than bad and shouldn't wait for a more general bugfix (?), cf https://github.com/phpstan/phpstan/issues/13996#issuecomment-3786955472

PR #4799, opened 2026-01-21
Prevent repetitive calls to getArgs()

staabm:
getArgs() currently takes 1% of cpu time

Its invoked a lot

grafik


ondrejmirtes:
Thank you!

PR #4801, opened 2026-01-22
ExpressionTypeHolder: re-use objects more often

staabm:

before this PR:
```
Elapsed time: 28.18 seconds
Used memory: 443.66 MB
```
after this PR:
```
Elapsed time: 28.09 seconds
Used memory: 434.81 MB
```

=> saves 9 MB memory


ondrejmirtes:
Thank you!

PR #4802, opened 2026-01-22
TrinaryLogic: detect logic errors

staabm:
allows detecting of dead code like

```
if ($flagState->yes()) {
	if ($flagState->no()) {
	}
}
```

### DIFF
--- a/src/TrinaryLogic.php
+++ b/src/TrinaryLogic.php
@@ -53,16 +53,28 @@ final class TrinaryLogic
 		return self::$registry[$value];
 	}
 
+	/**
+	 * @phpstan-assert-if-true =false $this->no()
+	 * @phpstan-assert-if-true =false $this->maybe()
+	 */
 	public function yes(): bool
 	{
 		return $this->value === self::YES;
 	}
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Oh nice! Thank you.

staabm:
I think we could narrow the types even more, but atm phpdoc cannot express it well enough I guess.

e.g.

`TrinaryLogic->createFromBoolean()` can only return TrinaryLogic::YES, TrinaryLogic::NO
`TrinaryLogic->createYes()` can only return TrinaryLogic::YES
`TrinaryLogic->createNo()` can only return TrinaryLogic::NO
`TrinaryLogic->createMaybe()` can only return TrinaryLogic::MAYBE
`TrinaryLogic->negate()` will leave MAYBE untouched, but invert YES/NO

and conditional types like
when `$trinary->yes()` is false, this means either `$trinary->no()|$trinary->maybe()` will be `true`

might be easier to express in case we can turn TrinaryLogic into a real `enum` so we could subtract enum-cases

PR #4804, opened 2026-01-22
Fix ignore comment handling for multiple traits in a single file

takaram:
`$linesToIgnore` data was not correctly gathered when:

- more than one traits are defined in a same file
- those traits are used in a same class

Closes phpstan/phpstan#13945

### DIFF
--- a/src/Analyser/FileAnalyser.php
+++ b/src/Analyser/FileAnalyser.php
@@ -118,7 +118,9 @@ final class FileAnalyser
 					}
 					if ($node instanceof InTraitNode) {
 						$traitNode = $node->getOriginalNode();
-						$linesToIgnore[$scope->getFileDescription()] = $this->getLinesToIgnoreFromTokens([$traitNode]);
+						$fileDescription = $scope->getFileDescription();
+						$linesToIgnore[$fileDescription] ??= [];
+						$linesToIgnore[$fileDescription] += $this->getLinesToIgnoreFromTokens([$traitNode]);
 
 						$traitFileName = $node->getTraitReflection()->getFileName();
 						if ($traitFileName !== null) {
--- a/tests/PHPStan/Analyser/data/bug-13945.php
+++ b/tests/PHPStan/Analyser/data/bug-13945.php
@@ -1,0 +1,18 @@
+<?php declare(strict_types = 1);
+
+namespace Bug13945;
+
+trait Foo {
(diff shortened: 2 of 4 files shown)


tests/PHPStan/Analyser/data/bug-13945.php
```diff
@@ -0,0 +1,18 @@
...
+	}
+}
+
+trait Baz {
+}
+
+class HelloWorld
+{
+	use Foo;
+	use Baz;
```
staabm, line 17: in https://github.com/phpstan/phpstan/issues/13945 its reported that the errors emitted depend on the order of this use-statements.

I think it would be good than, to additional test this with `use` in different order

src/Analyser/FileAnalyser.php
```diff
@@ -118,7 +118,9 @@ public function analyseFile(
 					}
 					if ($node instanceof InTraitNode) {
 						$traitNode = $node->getOriginalNode();
-						$linesToIgnore[$scope->getFileDescription()] = $this->getLinesToIgnoreFromTokens([$traitNode]);
+						$fileDescription = $scope->getFileDescription();
+						$linesToIgnore[$fileDescription] ??= [];
+						$linesToIgnore[$fileDescription] += $this->getLinesToIgnoreFromTokens([$traitNode]);
```
ondrejmirtes, line 123: Can you please explain what's happening here? `$scope->getFileDescription()` should be unique per trait usage.

ondrejmirtes (changes requested):

src/Analyser/FileAnalyser.php
takaram: The value of `$scope->getFileDescription()` contains:

- File path where the trait is defined
- Name of the class which `use`s the trait

It does not contain trait name, so is not necessarily unique when a file has two or more traits.

tests/PHPStan/Analyser/data/bug-13945.php
takaram: Added a test case

src/Analyser/FileAnalyser.php
ondrejmirtes: Ohh, right!

ondrejmirtes (approved):

ondrejmirtes:
Thank you very much!

PR #4805, opened 2026-01-22
FileTypeMapper - optimize traits

ondrejmirtes:


VincentLanglet:
Seems like the StaticAnalysis is failing for memory issue on every PR.
https://github.com/phpstan/phpstan-src/actions/runs/21264781157/job/61201092694?pr=4805

Maybe I miss a discussion with staabm ; are you aware why ?
Not sure if it's related but the error message talk about `shipmonk/dead-code-detector`.

ondrejmirtes:
Yeah, we're discussing it. Right now I'm trying to make build of Shopware faster, it gets really slow near the end because of traits in testcases and I'm solving the causes one by one.

It's possible the problems with memory will go away once this gets optimized too.

We have ideas how to make result cache take less memory when loaded.

PR #4808, opened 2026-01-23
FileCacheStorage: Don't use sprintf() on maybe huge strings

staabm:
this seems to be the most expensive use of `sprintf` - memory wise

grafik


ondrejmirtes:
Thank you!

PR #4809, opened 2026-01-23
NodeDependencies: prevent unnecessary FileHelper::normalizePath() calls

staabm:
in shopware this is invoked a lot, and I can see on my machine that the path oftentimes already matches even before normalization

grafik

----

its in the top 5 used calls when running `php vendor/bin/phpstan analyze tests/integration/Core/System --debug`

grafik


ondrejmirtes:
Nice, thank you!

PR #4813, opened 2026-01-23
fix: nested trait `@use`, use original class file name to get trait data

calebdw:
Hello!

Git bisect shows that commit https://github.com/phpstan/phpstan-src/commit/d8ec11f91ff26eea594c80d469820a7700e96189 introduced a regression in nested trait `@use` resolution, e.g., the following no longer passes the template to the base trait:

```php
/** @template TBuilder of Builder<*> */
trait HasBuilder
{
    /** @use BaseHasBuilder<TBuilder> */
	use BaseHasBuilder;
    // other methods
}
```

This bug only seems to show up when the classes are in different files and namespaces, so it was impossible to replicate on the playground (see https://phpstan.org/r/985be0cb-f0ba-47e9-8f3c-3f7f63dccdfa). In the e2e test, I also tried to just have a `data/` directory with `framework.php` and a `src.php` files that had the respective namespace and classes in them but that passed and did not replicate the failure. So sorry for the 6 individual classes / files but this was the only way that I could replicate.

> [!NOTE]
> I'm not sure if this is the best fix, this was just suggested by the AI agent that was helping me diagnose the cause of this regression. This does however, fix the issue I was having...

> [!WARNING]
> There seems to be some additional caching going on that's not being cleared by `phpstan clear-result-cache`. Running the following command before and after the commit that fixes the issue does not result in a different output unless the `tmp/` directory is completely cleared in between runs (i.e., if ran before the fix and then ran after the fix the command still results in an error when it should be passing; and the same backwards, running after the fix and then before the fix results in a passing command which should be failing):
>  ```bash
>    # this does not completely clear all state between runs
>    cd e2e/nested-trait-use && ../../bin/phpstan clear-result-cache && ../../bin/phpstan analyze
>    # instead need to manually clear the entire cache between runs with
>    cd e2e/nested-trait-use && rm -rf ./tmp/cache && ../../bin/phpstan analyze
>    ```

Thanks!

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -295,7 +295,6 @@ jobs:
               ../bashunit -a contains 'Method StubFiles\Foo::doFoo() has no return type specified.' "$OUTPUT"
           - script: |
               cd e2e/nested-trait-use
-              rm -rf ./tmp/cache
               ../../bin/phpstan analyze
 
     steps:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -293,6 +293,10 @@ jobs:
               OUTPUT=$(../bashunit -a exit_code "1" "../../bin/phpstan")
               echo "$OUTPUT"
               ../bashunit -a contains 'Method StubFiles\Foo::doFoo() has no return type specified.' "$OUTPUT"
+          - script: |
+              cd e2e/nested-trait-use
+              rm -rf ./tmp/cache
+              ../../bin/phpstan analyze
 
     steps:
       - name: "Checkout"
--- a/e2e/nested-trait-use/framework/Builder.php
+++ b/e2e/nested-trait-use/framework/Builder.php
@@ -1,0 +1,5 @@
+<?php
+
+namespace NestedTraitUse\Framework;
+
+class Builder {}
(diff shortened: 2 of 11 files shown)


ondrejmirtes:
> There seems to be some additional caching going on that's not being cleared by phpstan clear-result-cache. Running the following command before and after the commit

That's what variable cache key (2nd parameter of the load() method is for. When the cache becomes stale, you need to bump the number in the source code.

calebdw:
I suppose it needs to be bumped then as part of this fix to clear out all the bad caches?

.github/workflows/e2e-tests.yml
```diff
@@ -293,6 +293,10 @@ jobs:
               OUTPUT=$(../bashunit -a exit_code "1" "../../bin/phpstan")
               echo "$OUTPUT"
               ../bashunit -a contains 'Method StubFiles\Foo::doFoo() has no return type specified.' "$OUTPUT"
+          - script: |
+              cd e2e/nested-trait-use
+              rm -rf ./tmp/cache
```
ondrejmirtes:
```suggestion
```

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you!

PR #4814, opened 2026-01-24
Fix escaped hash in regex extended modefixed escaped hash in regex extended mode

dg:
When using the `x` (extended) modifier, `\#` was incorrectly treated as the start of a comment. According to PCRE specification, `\#` is an escaped literal hash character and should not start a comment.

### Example

```php
preg_match('/^ ([\#.]) $/x', $s, $matches);
```

Before this fix, PHPStan would strip `\#.]) $/x` as a comment, breaking the pattern parsing.

### Changes

- `src/Type/Regex/RegexGroupParser.php`: Changed the comment-stripping regex from `/(?<!\?)#.*/` to `/(?<!\\)#.*/` to respect escaped hashes
- Added test case `testExtendedSyntaxEscapedHash` in `tests/PHPStan/Analyser/nsrt/preg_match_shapes.php`

### Note

This fix does not address `#` inside character classes like `[#.]` (which is also not a comment in PCRE). However, users can use the workaround `[\#.]` which now works correctly with this fix.

### DIFF
--- a/src/Type/Regex/RegexGroupParser.php
+++ b/src/Type/Regex/RegexGroupParser.php
@@ -85,7 +85,8 @@ final class RegexGroupParser
 
 			if (str_contains($modifiers, 'x')) {
 				// in freespacing mode the # character starts a comment and runs until the end of the line
-				$regex = preg_replace('/(?<!\?)#.*/', '', $regex) ?? '';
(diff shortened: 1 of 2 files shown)


staabm (approved):
I was not aware of such a regex feature, but implementation wise it looks good to me.

thank you

ondrejmirtes:
Thank you!

PR #4815, opened 2026-01-24
NodeDependencies: cache normalized filenames

staabm:
in shopware, in which we have a lot of shared traits, the same paths are over and over again normalized.

grafik

we can see `FileHelper::normalizePath` take ~41 seconds, from which 95% are called from `NodeDependencies::getFileDependencies`.

---

even running this PR on phpstan-src, I can see `FileHelper::normalizePath` dropping from 175.059 to 137.958

### DIFF
--- a/src/Dependency/NodeDependencies.php
+++ b/src/Dependency/NodeDependencies.php
@@ -9,6 +9,10 @@ use function array_values;
 
 final class NodeDependencies
 {
+	/**
+	 * @var array<string, string>
+	 */
+	private static array $normalizedFileNames = [];
 
 	/**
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Can you experiment with having the cache in FileHelper directly?

staabm:
great idea. moving the cache to FileHelper makes the number of non-cached invocations drop to 11.623 in phpstan-src

staabm:
just double checked: after this PR `FileHelper::normalizePath` does not show up in the profile anymore

grafik

ondrejmirtes:
Thank you!

PR #4816, opened 2026-01-24
ImpossibleCheckTypeHelper: Remove unnecessary late-resolving

staabm:
late-resolvable return types are already resolved within the parameter acceptor

### DIFF
--- a/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
+++ b/src/Rules/Comparison/ImpossibleCheckTypeHelper.php
@@ -429,7 +429,7 @@ final class ImpossibleCheckTypeHelper
 				$parametersAcceptor = ParametersAcceptorSelector::selectFromArgs($scope, $node->getArgs(), $functionReflection->getVariants(), $functionReflection->getNamedArgumentsVariants());
 				$returnType = $parametersAcceptor->getReturnType();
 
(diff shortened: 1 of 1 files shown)


src/Rules/Comparison/ImpossibleCheckTypeHelper.php
```diff
@@ -428,7 +427,7 @@ private function determineContext(Scope $scope, Expr $node): TypeSpecifierContex
 			if ($this->reflectionProvider->hasFunction($node->name, $scope)) {
 				$functionReflection = $this->reflectionProvider->getFunction($node->name, $scope);
 				$parametersAcceptor = ParametersAcceptorSelector::selectFromArgs($scope, $node->getArgs(), $functionReflection->getVariants(), $functionReflection->getNamedArgumentsVariants());
-				$returnType = TypeUtils::resolveLateResolvableTypes($parametersAcceptor->getReturnType());
+				$returnType = $parametersAcceptor->getReturnType();
```
staabm, line 430: my initial idea which made me visit this code was, whether we can cache the `TypeSpecifierContext` per `CallLike`, in case its a native non-conditional return type of a single-variant callable.

turns out I cannot detect at this level whether the return type is conditional, because its already resolved.

any idea how I could do it instead?

update: found a way https://github.com/phpstan/phpstan-src/pull/4817

ondrejmirtes:
Thank you!

PR #4817, opened 2026-01-24
ImpossibleCheckTypeHelper: Cache context determination

staabm:
cache the `TypeSpecifierContext` per `CallLike`, in case its a native non-conditional return type of a single-variant callable.


staabm:
the method is not slow enough to add caching for it

PR #4819, opened 2026-01-25
Support StaticMethodParameterClosureTypeExtension for New_ expressions

ruudk:
See https://github.com/phpstan/phpstan/discussions/14014#discussioncomment-15590392

Decided to add support to the current extensions to make it easier to merge this instead of proposing the change to:
- https://github.com/phpstan/phpstan-src/pull/4290

/cc @ondrejmirtes @calebdw

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -4816,6 +4816,17 @@ class NodeScopeResolver
 						return $staticMethodParameterClosureTypeExtension->getTypeFromStaticMethodCall($calleeReflection, $callLike, $parameter, $scope);
 					}
 				}
+			} elseif ($callLike instanceof New_ && $callLike->class instanceof Name) {
+				$staticCall = new StaticCall(
+					$callLike->class,
+					new Identifier('__construct'),
+					$callLike->getArgs(),
+				);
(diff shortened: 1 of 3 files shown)


ruudk:
It seems these failures are not related to this PR but coming from the main branch.

ondrejmirtes:
Thank you!

src/Analyser/NodeScopeResolver.php
```diff
@@ -5874,6 +4816,17 @@ private function getParameterTypeFromParameterClosureTypeExtension(CallLike $cal
 						return $staticMethodParameterClosureTypeExtension->getTypeFromStaticMethodCall($calleeReflection, $callLike, $parameter, $scope);
 					}
 				}
+			} elseif ($callLike instanceof New_ && $callLike->class instanceof Name) {
+				$staticCall = new StaticCall(
+					$callLike->class,
+					new Identifier('__construct'),
+					$callLike->getArgs(),
```
samsonasik, line 4823: We use `NodeScopeResolver` in rector, I think `getArgs()` should be replaced with `args` here, or check `isFirstClassCallable()` early here to avoid error assert error, see:

https://github.com/nikic/PHP-Parser/blob/8c360e27327c8bd29e1c57721574709d0d706118/lib/PhpParser/Node/Expr/CallLike.php#L32

ondrejmirtes: `$callLike` here cannot be a first class callable.

PR #4823, opened 2026-01-26
Don't error with `checkMissingCallableSignature: true` about degraded closures

staabm:
closes https://github.com/phpstan/phpstan/issues/14012

---

_Updated Description_

after this PR we still have errors when callables have been degraded, but as we degrade less often/later this should be less of a problem.

I also checked the reproducers of https://github.com/phpstan/phpstan-src/pull/4684 so they are equal fast.

In addition we no longer degrade callables when defined in a phpdoc array shape.


staabm:
We could raise the threshold to degenerate later.. need to test this in the origin slow snippets though

ondrejmirtes:
An idea - we could degrade it only for literal array in code, but we wouldn't need to degrade it for array shapes? Meaning ConstantArrayTypeBuilder in TypeNodeResolver would be created in a different mode.

Also maybe the limit for degrading closures could be higher?

staabm:
> An idea - we could degrade it only for literal array in code, but we wouldn't need to degrade it for array shapes? Meaning ConstantArrayTypeBuilder in TypeNodeResolver would be created in a different mode.

I don't see how this is possible, because the worst case scenario is built up over one assign per line
see https://github.com/GrzegorzDrozd/phpstan-performance-issue-reproduction/blob/master/src/index_bad.php

staabm:
I had to discard the previous fix because it discarded all perf benefits we got previosuly with https://github.com/phpstan/phpstan-src/pull/4684

tests/PHPStan/Rules/Properties/data/bug-14012.php
```diff
@@ -0,0 +1,40 @@
...
+	 *   isNonEmptyString: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isNull: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isNumeric: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isObject: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isPositiveInt: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isResource: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isSameAs: \Closure(Scope, Node\Arg, Node\Arg): Node\Expr,
+	 *   isScalar: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isString: \Closure(Scope, Node\Arg): Node\Expr,
+	 *   isTrue: \Closure(Scope, Node\Arg): Node\Expr,
```
ondrejmirtes, line 37: Please also assert what's the type of the property here.

ondrejmirtes (changes requested):

tests/PHPStan/Rules/Properties/data/bug-14012.php
staabm: done

tests/PHPStan/Rules/Properties/data/bug-14012.php
```diff
@@ -0,0 +1,43 @@
...
+	 * }
+	 */
+	public static array $resolvers = [];
+}
+
+assertType("non-empty-array<'hasMethod'|'hasProperty'|'isArray'|'isBool'|'isCallable'|'isCountable'|'isFalse'|'isFloat'|'isInstanceOf'|'isInt'|'isIterable'|'isList'|'isMap'|'isNaturalInt'|'isNegativeInt'|'isNonEmptyString'|'isNull'|'isNumeric'|'isObject'|'isPositiveInt'|'isResource'|'isSameAs'|'isScalar'|'isString'|'isTrue', callable(): mixed>&oversized-array", ExpectationMethodResolver::$resolvers);
```
paulbalandan: This would consequently cause an `argument.type` error on consumers as it's now `mixed` instead of `Node\Expr`.

staabm: I think this problem is inevitable (producing more precise types will slow down the analysis).
I raised the limit to 32 now, so it is less likely to happen

staabm:
sorry for flooding this PR with commits/pushes

paulbalandan:
> I don't see how this is possible, because the worst case scenario is built up over one assign per line see https://github.com/GrzegorzDrozd/phpstan-performance-issue-reproduction/blob/master/src/index_bad.php

How about this one:

```diff
diff --git a/src/PhpDoc/TypeNodeResolver.php b/src/PhpDoc/TypeNodeResolver.php
index 4df978002..979977508 100644
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -1050,6 +1050,13 @@ final class TypeNodeResolver
 			$builder->degradeToGeneralArray(true);
 		}
 
+		foreach ($typeNode->items as $itemNode) {
+			if ($itemNode->valueType instanceof CallableTypeNode) {
+				$builder->degradeClosures(false); // explicit opt-out of closure degradation
+				break;
+			}
+		}
+
 		foreach ($typeNode->items as $itemNode) {
 			$offsetType = $this->resolveArrayShapeOffsetType($itemNode, $nameScope);
 			$builder->setOffsetValueType($offsetType, $this->resolve($itemNode->valueType, $nameScope), $itemNode->optional);
diff --git a/src/Type/Constant/ConstantArrayTypeBuilder.php b/src/Type/Constant/ConstantArrayTypeBuilder.php
index 9396fec39..3d7005e93 100644
--- a/src/Type/Constant/ConstantArrayTypeBuilder.php
+++ b/src/Type/Constant/ConstantArrayTypeBuilder.php
@@ -35,7 +35,7 @@ final class ConstantArrayTypeBuilder
 
 	private bool $degradeToGeneralArray = false;
 
-	private bool $degradeClosures = false;
+	private ?bool $degradeClosures = false;
 
 	private bool $oversized = false;
 
@@ -94,7 +94,7 @@ final class ConstantArrayTypeBuilder
 					$numClosures++;
 				}
 
-				if ($numClosures >= self::CLOSURES_COUNT_LIMIT) {
+				if ($numClosures >= self::CLOSURES_COUNT_LIMIT && $this->degradeClosures !== false) {
 					$this->degradeClosures = true;
 					$this->degradeToGeneralArray = true;
 					$this->oversized = true;
@@ -300,6 +300,15 @@ final class ConstantArrayTypeBuilder
 		$this->oversized = $this->oversized || $oversized;
 	}
 
+	/**
+	 * @param bool|null $degrade Use `null` to auto-detect based on closures count;
+	 *                           Use boolean to explicitly enable/disable closure degradation.
+	 */
+	public function degradeClosures(?bool $degrade = false): void
+	{
+		$this->degradeClosures = $degrade;
+	}
+
 	public function getArray(): Type
 	{
 		$keyTypesCount = count($this->keyTypes);

```

On my laptop, it is a steady ~3.02 seconds before/after this patch.

staabm:
> On my laptop, it is a steady ~3.02 seconds before/after this patch.

`time vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php --filter testBug13933` takes 6 seconds after this change and 1.8 seconds before.

edit: found the typo in your suggestion :)

staabm:
should be good to go. thank you both

ondrejmirtes:
Nice, I imagined something like that 👍

ondrejmirtes:
Maybe we could do something like this for the Builder's `ARRAY_COUNT_LIMIT` limit in PHPDocs as well?

ondrejmirtes (approved):

PR #4824, opened 2026-01-26
Add regression test

takaram:
phpstan/phpstan#11598 was already resolved probably somewhere in 2.0.0.
This PR adds a regression test for the issue.

Closes phpstan/phpstan#11598

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-11598.php
+++ b/tests/PHPStan/Analyser/data/bug-11598.php
@@ -1,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace Bug11598;
+
+use function unserialize;
+
+function(): void {
+	[ // @phpstan-ignore offsetAccess.nonArray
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Analyser/data/bug-11598.php
```diff
@@ -0,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace Bug11598;
+
+use function unserialize;
+
+function(): void {
+	[ // @phpstan-ignore offsetAccess.nonArray
+		'bar' => $foo,
+	] = '';
```
takaram, line 10: In the original code rhs was `unserialize('')`, but it requires level 9 (checkExplicitMixed: true) while AnalyserIntegrationTest runs analysis with level 8.
So the code is modified to make rhs not mixed.

ondrejmirtes:
Thank you!

PR #4825, opened 2026-01-26
Speed up CombinationHelper

kaja47:
This change swaps loops to reduce number of intermediate arrays allocated and flips direction in which results are constructed to eliminate explicit append loop.

For 3 arrays with 10 elements each this leads to 5x speedup.

### DIFF
--- a/src/Internal/CombinationsHelper.php
+++ b/src/Internal/CombinationsHelper.php
@@ -2,7 +2,7 @@
 
 namespace PHPStan\Internal;
 
-use function array_shift;
+use function array_pop;
 
 final class CombinationsHelper
 {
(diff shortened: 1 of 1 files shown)


staabm:
> to 5x speedup.

how did you measure the speedup?

ondrejmirtes:
Thank you very much!

kaja47:
> how did you measure the speedup?

PHP 8.4.16, JIT turned off, CPU i5-4570 and timed this loop

```php
$ass = array_fill(0,3, range(1,10));
$sink = 0;
for ($i = 0; $i < 20_000; $i++) {
  $sink += count(iterator_to_array(CombinationsHelper::combinations($ass)));
}
var_dump($sink);
```

PR #4826, opened 2026-01-27
Don't lose known offset-types in array_replace()

staabm:
analog https://github.com/phpstan/phpstan-src/pull/4554

### DIFF
--- a/src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
+++ b/src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
@@ -8,6 +8,8 @@ use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
+use PHPStan\Type\Accessory\HasOffsetType;
+use PHPStan\Type\Accessory\HasOffsetValueType;
 use PHPStan\Type\Accessory\NonEmptyArrayType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantArrayType;
@@ -15,9 +17,11 @@ use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\TypeUtils;
 use function array_keys;
 use function count;
 use function in_array;
@@ -97,6 +101,45 @@ final class ArrayReplaceFunctionReturnTypeExtension implements DynamicFunctionRe
 			return $newArrayBuilder->getArray();
 		}
 
+		$offsetTypes = [];
+		foreach ($argTypes as $argType) {
+			$constArrays = $argType->getConstantArrays();
+			if ($constArrays !== []) {
+				foreach ($constArrays as $constantArray) {
+					foreach ($constantArray->getKeyTypes() as $keyType) {
+						$hasOffsetValue = TrinaryLogic::createFromBoolean($argType->hasOffsetValueType($keyType)->yes());
+						$offsetTypes[$keyType->getValue()] = [
+							$hasOffsetValue,
+							$argType->getOffsetValueType($keyType),
+						];
+					}
+				}
+			} else {
+				foreach ($offsetTypes as $key => [$hasOffsetValue, $offsetValueType]) {
+					$offsetTypes[$key] = [
+						$hasOffsetValue->and(TrinaryLogic::createMaybe()),
+						new MixedType(),
+					];
+				}
+			}
+
+			foreach (TypeUtils::getAccessoryTypes($argType) as $accessoryType) {
+				if (
+					!($accessoryType instanceof HasOffsetType)
+					&& !($accessoryType instanceof HasOffsetValueType)
+				) {
+					continue;
+				}
+
+				$offsetType = $accessoryType->getOffsetType();
+				$offsetTypes[$offsetType->getValue()] = [
+					TrinaryLogic::createYes(),
+					$argType->getOffsetValueType($offsetType),
+				];
+			}
+		}
+
+
 		$keyTypes = [];
 		$valueTypes = [];
 		$nonEmpty = false;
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Analyser/nsrt/array-replace.php
```diff
@@ -76,11 +76,11 @@ public function arrayReplaceUnionTypeArrayShapes($array1, $array2): void
...
 	{
-		assertType("non-empty-array<string, '1'|'2'|int>", array_replace($array1, $array2));
-		assertType("non-empty-array<string, '1'|'2'|int>", array_replace($array2, $array1));
+		assertType("non-empty-array<string, '1'|'2'|int>&hasOffset('bar')&hasOffset('foo')", array_replace($array1, $array2));
+		assertType("non-empty-array<string, '1'|'2'|int>&hasOffsetValue('bar', '2')&hasOffsetValue('foo', '1')", array_replace($array2, $array1));
 
-		assertType("non-empty-array<'bar'|'foo'|int, string>", array_replace($array1, $array3));
-		assertType("non-empty-array<'bar'|'foo'|int, string>", array_replace($array3, $array1));
+		assertType("non-empty-array<'bar'|'foo'|int, string>&hasOffset('bar')&hasOffset('foo')", array_replace($array1, $array3));
+		assertType("non-empty-array<'bar'|'foo'|int, string>&hasOffsetValue('bar', '2')&hasOffsetValue('foo', '1')", array_replace($array3, $array1));
```
staabm, line 83: this kind of overlaps. should we do something about it, or is it fine?

staabm:
@claudepache @VincentLanglet you might be interessted in reviewing this one

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
```diff
@@ -97,6 +102,44 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
 		}
 
+		$offsetTypes = [];
+		foreach ($argTypes as $argType) {
+			$constArrays = $argType->getConstantArrays();
+			if ($constArrays !== []) {
+				foreach ($constArrays as $constantArray) {
+					foreach ($constantArray->getKeyTypes() as $keyType) {
+						$hasOffsetValue = TrinaryLogic::createFromBoolean($argType->hasOffsetValueType($keyType)->yes());
+						$offsetTypes[$keyType->getValue()] = [
```
VincentLanglet, line 112: If multiple constantArray has the same Key, like `array{1: int, 2: string}|array{1: int, 3: string}` you will call `hasOffsetValueType` and `getOffsetValueType` two times to set the same value ; not sure if we could avoid the extra calls.

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
```diff
@@ -97,6 +102,44 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+							$hasOffsetValue,
+							$argType->getOffsetValueType($keyType),
+						];
+					}
+				}
+			} else {
+				foreach ($offsetTypes as $key => [$hasOffsetValue, $offsetValueType]) {
+					$offsetTypes[$key] = [
+						$hasOffsetValue->and(TrinaryLogic::createMaybe()),
+						new MixedType(),
```
VincentLanglet, line 124: Technically we could provide better than Mixed here.
Maybe a comment to explain there is no need to provide `Union($offsetValueType, $argType->getArrayValues)` since the value won't be used since we're only instantiating `HasOffsetType`

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
```diff
@@ -97,6 +102,44 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 			return $newArrayBuilder->getArray();
 		}
 
+		$offsetTypes = [];
+		foreach ($argTypes as $argType) {
+			$constArrays = $argType->getConstantArrays();
+			if ($constArrays !== []) {
+				foreach ($constArrays as $constantArray) {
+					foreach ($constantArray->getKeyTypes() as $keyType) {
+						$hasOffsetValue = TrinaryLogic::createFromBoolean($argType->hasOffsetValueType($keyType)->yes());
```
VincentLanglet, line 111: It's unclear to me reading this why you're doing
```
TrinaryLogic::createFromBoolean($argType->hasOffsetValueType($keyType)->yes())
```
and not directly
```
$argType->hasOffsetValueType($keyType)
```
You're transforming the Maybe into a No ?

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
```diff
@@ -133,6 +176,31 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
 		if ($isList) {
 			$arrayType = TypeCombinator::intersect($arrayType, new AccessoryArrayListType());
 		}
+		if ($offsetTypes !== []) {
+			$knownOffsetValues = [];
+			foreach ($offsetTypes as $key => [$hasOffsetValue, $offsetType]) {
```
VincentLanglet, line 183: Eventually, looking at the usage you could have use `array<Type|null>` as offsetTypes:
- a Type means `HasOffsetValueType` (the yes case)
- null means `HasOffsetType` (the maybe case)

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
staabm: correct. I need the trinary, as I later on combine it with `$hasOffsetValue->and(TrinaryLogic::createMaybe())` for non constant arrays (else case)

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
staabm: as long as this is not identified as a bottleneck I don't want to complicate it further

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
staabm: added a comment

src/Type/Php/ArrayReplaceFunctionReturnTypeExtension.php
staabm: agree we could transform it. not sure this will lead to logic which is easier to follow?

VincentLanglet (approved):

tests/PHPStan/Analyser/nsrt/array-replace.php
VincentLanglet: You mean the non-empty-array and the hasOffset(Value) ?
I think it's more clear this way, by being explicit.

I dunno if it's currently possible to get a dump 
```
array<...>&hasOffset(..)
```
It would be weird to be able to get both `array` or `non-empty-array` in this situation.

VincentLanglet (approved):

ondrejmirtes:
Thank you!

tests/PHPStan/Analyser/nsrt/array-replace.php
staabm: the type is
`non-empty-array<'bar'|'foo'|int, string>&hasOffsetValue('bar', '2')&hasOffsetValue('foo', '1')`

while some keys in
`non-empty-array<'bar'|'foo'|int, string>`
are redundant with the accessories in
`hasOffsetValue('bar', '2')&hasOffsetValue('foo', '1')`

PR #4827, opened 2026-01-27
Make sure function scope is set in closure in constant expression

ruudk:
See https://github.com/phpstan/phpstan/discussions/14014#discussioncomment-15620096

Let's see what the tests say.

### DIFF
--- a/tests/PHPStan/Analyser/data/closure-type-in-constant-expression-php85.php
+++ b/tests/PHPStan/Analyser/data/closure-type-in-constant-expression-php85.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace ClosureTypeInConstantExpressionPhp85;
+
+use Attribute;
+use PhpParser\Node\Expr\StaticCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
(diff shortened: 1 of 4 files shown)


ruudk:
@ondrejmirtes I rebased with main, then tried to see if I could get a test that uses StaticMethodParameterClosureTypeExtension.

But I feel like this is not the right extension point.

tests/PHPStan/Analyser/data/closure-type-in-constant-expression-php85.php
```diff
@@ -0,0 +1,83 @@
...
+			],
+			new MixedType()
+		);
+
+		// What we need here, is a way to find out that this is being called from `__invoke`
+		// using $scope->getFunctionName()
+		// Then we would want to find out the 1st parameter of __invoke
+		// It would be SomeCommand.
+		// Then we would want to return that the callable signature is `static function (SomeCommand): string`
+		// But I'm not sure if this is the correct extension point at all...
```
ondrejmirtes, line 53: Why do you feel that? Did you try to implement the logic here along with the change to NodeScopeResolver to see if my suggestion worked?

ondrejmirtes (changes requested):

PR #4831, opened 2026-01-29
MutatingScope: Faster createConditionalExpressions

staabm:
this PR optimizes one of the top 3 methods regarding cost:

grafik

---

blackfire diff of `php bin/phpstan analyze src/Rules src/Analyser --debug -v`

grafik


ondrejmirtes:
Nice, thank you!

PR #4833, opened 2026-01-29
speed up TrinaryLogic

kaja47:
- Split variadic arguments for and() and or(). Common case of single argument doesn't allocate an array.
- Add static fields mirroring $registry to skip one level of indirection.
- Numeric values chosen that & and | is the same as min and max. and()/or() optimized accordingly.
- Inline trivial create() method.
- Remove one unnecessary access to registry in method create().

this loop runs 1.8x faster (php 8.4.16, jit off)

$t = TrinaryLogic::createYes();
for ($i = 10_000_000; $i--;) {
	$t = $t->and(TrinaryLogic::createYes());
}

before: 1.125 s
now:    0.616 s

### DIFF
--- a/src/TrinaryLogic.php
+++ b/src/TrinaryLogic.php
@@ -38,13 +38,17 @@ use function min;
  */
 final class TrinaryLogic
 {
+	private const YES = 3;
+	private const MAYBE = 1;
+	private const NO = 0;
 
-	private const YES = 1;
-	private const MAYBE = 0;
-	private const NO = -1;
 
 	/** @var self[] */
 	private static array $registry = [];
+
+	private static self $YES;
+	private static self $MAYBE;
+	private static self $NO;
 
 	private function __construct(private int $value)
 	{
@@ -52,17 +56,17 @@ final class TrinaryLogic
 
 	public static function createYes(): self
 	{
-		return self::$registry[self::YES] ??= new self(self::YES);
+		return self::$YES ??= (self::$registry[self::YES] ??= new self(self::YES));
 	}
 
 	public static function createNo(): self
 	{
-		return self::$registry[self::NO] ??= new self(self::NO);
+		return self::$NO ??= (self::$registry[self::NO] ??= new self(self::NO));
 	}
 
 	public static function createMaybe(): self
 	{
-		return self::$registry[self::MAYBE] ??= new self(self::MAYBE);
+		return self::$MAYBE ??= (self::$registry[self::MAYBE] ??= new self(self::MAYBE));
 	}
 
 	public static function createFromBoolean(bool $value): self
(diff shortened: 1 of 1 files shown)


src/TrinaryLogic.php
```diff
@@ -89,17 +92,13 @@ public function toBooleanType(): BooleanType
 		return new ConstantBooleanType($this->value === self::YES);
 	}
 
-	public function and(self ...$operands): self
+	public function and(self $operand, self ...$rest): self
```
staabm: I think this needs to be 
`public function and(?self $operand = null, self ...$rest): self`

because previously `$operands` could be a empty array (so no args given to `and` when invoked via splat
`$trinary->and(...$moreTrinaries)`)

staabm:
one bug in `negate` otherwise this looking really good already. awesome.

```

1) PHPStan\TrinaryLogicTest::testNegate with data set #0 (PHPStan\TrinaryLogic Object (...), PHPStan\TrinaryLogic Object (...))

/Users/m.staab/dvl/phpstan-src/tests/PHPStan/TrinaryLogicTest.php:105

2) PHPStan\TrinaryLogicTest::testNegate with data set #2 (PHPStan\TrinaryLogic Object (...), PHPStan\TrinaryLogic Object (...))

/Users/m.staab/dvl/phpstan-src/tests/PHPStan/TrinaryLogicTest.php:105

3) PHPStan\TrinaryLogicTest::testNegate with data set #1 (PHPStan\TrinaryLogic Object (...), PHPStan\TrinaryLogic Object (...))

/Users/m.staab/dvl/phpstan-src/tests/PHPStan/TrinaryLogicTest.php:105
```

thg2k:
Sorry if I miss something obvious, but what's the point of keeping the $registry static with the addition of the new $YES $NO $MAYBE? Isn't it fully redundant now?

ondrejmirtes:
This is really nice! I welcome your low-level knowledge like this here 😊

staabm:
@kaja47 are you still working on this, or should I finish it?

thank you

staabm:
verified the benchmark locally

```
➜  phpstan-src git:(pr/4833) ✗ cat loop.php
<?php

use PHPStan\TrinaryLogic;

require_once __DIR__ . '/vendor/autoload.php';

$t = TrinaryLogic::createYes();
for ($i = 10_000_000; $i--;) {
        $t = $t->and(TrinaryLogic::createYes());
}
```

----

this PR

```
➜  phpstan-src git:(pr/4833) ✗ hyperfine 'php loop.php'
Benchmark 1: php loop.php
  Time (mean ± σ):     880.5 ms ±   4.1 ms    [User: 866.7 ms, System: 10.8 ms]
  Range (min … max):   874.6 ms … 886.4 ms    10 runs
```

vs. 2.1.x

```
➜  phpstan-src git:(2.1.x) ✗ hyperfine 'php loop.php'
Benchmark 1: php loop.php
  Time (mean ± σ):      1.323 s ±  0.025 s    [User: 1.297 s, System: 0.018 s]
  Range (min … max):    1.291 s …  1.360 s    10 runs
```

ondrejmirtes:
Thank you!

staabm:
Thank you @kaja47

PR #4835, opened 2026-01-29
IntersectionType cannot be void

staabm:
before this PR we checked IntersectionTypes of up to 38 inner-elements whether these are void.

by definition a IntersectionType cannot be void


ondrejmirtes:
Thank you!

PR #4836, opened 2026-01-30
Cache IntersectionType results

staabm:
cache results of most invoked `IntersectionType`-methods because `IntersectionType->intersectResults` shows up in profiles:

grafik

---

diff

grafik


ondrejmirtes:
Nice, thank you!

PR #4839, opened 2026-01-30
Implement bidirectional array type narrowing in foreach loops

takeokunn:
## Summary

Closes phpstan/phpstan#13959

When a foreach value variable is narrowed via `instanceof`, the source array's item type is now also narrowed accordingly.

## Example

```php
/** @param list<Animal|string> $animals */
function example(array $animals): void {
    foreach ($animals as $animal) {
        if ($animal instanceof Dog) {
            // $animals is now list<Dog>
        }
    }
}
```

## Implementation

- Add ForeachSourceTracking class to track foreach value → array relationships
- Add narrowItemType() method to Type interface (47 implementations)
- Extend TypeSpecifier with foreach narrowing propagation logic
- Preserve list markers (list<T> stays list<T>, not array<int, T>)

## Limitations

- Only narrows on instanceof checks (conservative approach)
- Variable array offsets do not trigger narrowing
- By-reference foreach does not narrow (mutation safety)

## Commits

| # | Hash        | Message                                                               |
|---|-------------|-----------------------------------------------------------------------|
| 1 | 853982d78 | Type narrowing: Implement bidirectional array type narrowing (#13959) |
| 2 | b2ff92760 | Refactor: Optimize and document foreach type narrowing implementation |

### DIFF
--- a/build/phpstan.neon
+++ b/build/phpstan.neon
@@ -91,6 +91,9 @@ parameters:
 				- ../tests/PHPStan/Fixture
 			reportUnmatched: false # constants on enums, not reported on PHP8-
 		-
+			identifier: shipmonk.deadMethod
+			path: ../src/Type/Traits/MaybeOffsetAccessibleTypeTrait.php
+		-
 			message: '''
 				#^Access to constant on deprecated class DeprecatedAnnotations\\DeprecatedFoo\:
 				in 1\.0\.0\.$#
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1380,7 +1380,7 @@ parameters:
 		-
 			rawMessage: 'Doing instanceof PHPStan\Type\ArrayType is error-prone and deprecated. Use Type::isArray() or Type::getArrays() instead.'
 			identifier: phpstanApi.instanceofType
-			count: 4
+			count: 5
 			path: src/Type/IntersectionType.php
 
 		-
--- a/src/Analyser/ForeachSourceTracking.php
+++ b/src/Analyser/ForeachSourceTracking.php
@@ -1,0 +1,24 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Analyser;
+
+use PhpParser\Node\Expr;
+use PHPStan\Type\Type;
+
+/**
+ * Tracks the source of a foreach loop value variable
+ *
+ * This class stores information about where a foreach value variable comes from,
+ * enabling bidirectional type narrowing between the array and its values.
+ */
+class ForeachSourceTracking
+{
+
+	public function __construct(
+		public readonly string $valueVarName,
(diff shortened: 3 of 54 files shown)


thg2k:
```php
/** @param list<Animal|string> $animals */
function example(array $animals): void {
    foreach ($animals as $animal) {
        if ($animal instanceof Dog) {
            // $animals is now list<Dog>
        }
    }
}
```

I do not understand how `$animals` is `list<Dog>` in your example. I only see
that one specific item in the array `Dog`, but how does that imply that all
the elements are?

Maybe you meant something like the following?

```php
/** @param list<Animal|string> $animals */
function example(array $animals): void {
    foreach ($animals as $animal) {
        if (!$animal instanceof Dog)
            throw \Exception("not Dog");
    }
    // $animals is now list<Dog>
}
```

VincentLanglet:
> ```
> /** @param list<Animal|string> $animals */
> function example(array $animals): void {
>     foreach ($animals as $animal) {
>         if ($animal instanceof Dog) {
>             // $animals is now list<Dog>
>         }
>     }
> }
> ```

This is wrong
- without something like an exception when it's not the expected instanceof
- in the foreach loop

Without exceptions there no reason to restrict the types
```
$animals = [$dog1, $dog2, $cat];
foreach ($animals as $animal) {
     if ($animal instanceof Dog) {
          // $animals is still not a list<Dog>
     }
}
```
And with exception, you should not restrict the types too early
```
$animals = [$dog1, $dog2, $cat];
foreach ($animals as $animal) {
     if ($animal instanceof Dog) {
          // $animals is still not a list<Dog>
          // cause we didn't check all the value yet
     } else {
          throw new \Exception();
     }
}
// But now, $animals is a list<Dog>
```

Also, no need for `@internal` tag ; everything is internal by default, except when `@api` tag is used.

takeokunn:
Thank you @thg2k and @VincentLanglet for the thorough review — you're both absolutely right.

Narrowing the entire source array's item type inside a single foreach iteration is unsound. `$animal instanceof Dog` only proves the **current** element is a `Dog`, not that all elements are:

```php
$animals = [new Dog(), new Cat()];
foreach ($animals as $animal) {
    if ($animal instanceof Dog) {
        // $animals is NOT list<Dog> — Cat is still in the array
    }
}
```

As @VincentLanglet correctly pointed out, the valid narrowing should happen **after** the loop, when all elements have been validated through early termination:

```php
foreach ($animals as $animal) {
    if (!$animal instanceof Dog) {
        throw new \Exception();
    }
}
// Now $animals is list<Dog> — this is sound
```

Regarding `@internal`: understood, I'll drop those tags since everything is internal by default unless marked with `@api`.

I'm closing this PR and will reconsider the approach to address the original issue (#13959) correctly — focusing on after-loop narrowing via exception/return elimination rather than in-loop narrowing.

PR #4842, opened 2026-01-31
NodeScopeResolver: faster produceArrayDimFetchAssignValueToWrite()

staabm:
~50% faster

before PR

grafik

----

after PR

grafik


ondrejmirtes:
Nice, thank you!

PR #4844, opened 2026-02-01
Fix array types after `array_splice()`

takaram:
Closes phpstan/phpstan#14037

### DIFF
--- a/src/Type/ArrayType.php
+++ b/src/Type/ArrayType.php
@@ -521,13 +521,31 @@ class ArrayType implements Type
 			return new ConstantArrayType([], []);
 		}
 
+		$isIntegerKey = $this->keyType->isInteger();
+		if ($isIntegerKey->yes()) {
+			$keyType = IntegerRangeType::createAllGreaterThanOrEqualTo(0);
+		} elseif ($isIntegerKey->maybe()) {
+			$keyType = TypeCombinator::union(
+				TypeCombinator::remove($this->getIterableKeyType(), new IntegerType()),
+				IntegerRangeType::createAllGreaterThanOrEqualTo(0),
+			);
+		} else {
+			$keyType = TypeCombinator::union($this->getIterableKeyType(), $replacementArrayType->getKeysArray()->getIterableKeyType());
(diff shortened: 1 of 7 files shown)


src/Type/ArrayType.php
```diff
@@ -521,13 +521,31 @@ public function spliceArray(Type $offsetType, Type $lengthType, Type $replacemen
 			return new ConstantArrayType([], []);
 		}
 
+		$isIntegerKey = $this->keyType->isInteger();
+		if ($isIntegerKey->yes()) {
+			$keyType = IntegerRangeType::createAllGreaterThanOrEqualTo(0);
+		} elseif ($isIntegerKey->maybe()) {
+			$keyType = TypeCombinator::union(
```
VincentLanglet, line 528: I think all the existing `'non-empty-array'`  are transformed into `non-empty-array<int<0, max>|string, mixed>` because of now benevolent union (array key) is transformed into a real union.

I dunno if we want to keep a benevolent union here...

tests/PHPStan/Analyser/nsrt/bug-14037.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Bug14037;
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @param array<10|20|30|'a', mixed> $a
+ */
+function splice(array $a): void {
+	array_splice($a, 0, 0);
+	assertType("array<'a'|int<0, max>, mixed>", $a);
```
VincentLanglet, line 12: Should be `'a'|int<0, 2>`

You're doing `$keyType = IntegerRangeType::createAllGreaterThanOrEqualTo(0);`

but maybe you should check `getConstantScalarTypes`.
and if `if (count($constantScalars) > 0)`, you're doing `$keyType = IntegerRangeType::fromInterval(0, $count -1);` ?

VincentLanglet:
Thanks, I merged your PR with https://github.com/phpstan/phpstan-src/pull/5139 which provides a slight better fix

PR #4845, opened 2026-02-01
Regression test for #14041

longwave:
Regression coverage for https://github.com/phpstan/phpstan/issues/14041

If `pdo_mysql` is not installed, this fails today, but should pass on Tuesday when the JetBrains stubs are updated.


ondrejmirtes:
I can trigger the update manually.

ondrejmirtes:
Oh, the script is failing on something and has to be fixed it seems: https://github.com/phpstan/phpstan-src/actions/workflows/update-phpstorm-stubs.yml

So this requires manual intervention and a fix to the script.

ondrejmirtes:
Thank you.

PR #4848, opened 2026-02-02
MutatingScope: Prevent unnecessary scope re-creation

staabm:
~~this does not work yet and I don't know why.~~
~~leaving it here as someone might have ideas~~

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3412,8 +3412,6 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 			);
 		}
 
-
-
 		if ($expr instanceof AlwaysRememberedExpr) {
 			return $scope->specifyExpressionType($expr->expr, $type, $nativeType, $certainty);
 		}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3386,24 +3386,33 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 		$nativeTypes = $scope->nativeExpressionTypes;
 		$nativeTypes[$exprString] = new ExpressionTypeHolder($expr, $nativeType, $certainty);
 
-		$scope = $this->scopeFactory->create(
-			$this->context,
-			$this->isDeclareStrictTypes(),
-			$this->getFunction(),
-			$this->getNamespace(),
-			$expressionTypes,
-			$nativeTypes,
-			$this->conditionalExpressions,
-			$this->inClosureBindScopeClasses,
-			$this->anonymousFunctionReflection,
-			$this->inFirstLevelStatement,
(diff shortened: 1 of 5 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -3382,9 +3382,19 @@ public function specifyExpressionType(Expr $expr, Type $type, Type $nativeType,
...
+		$newNativeHolder = new ExpressionTypeHolder($expr, $nativeType, $certainty);
+
+		if (
+			!isset($scope->expressionTypes[$exprString])
+			|| !isset($scope->nativeExpressionTypes[$exprString])
+			|| !$newHolder->equals($scope->expressionTypes[$exprString])
+			|| !$newNativeHolder->equals($scope->nativeExpressionTypes[$exprString])
+		) {
+			$expressionTypes[$exprString] = $newHolder;
+			$nativeTypes[$exprString] = $newNativeHolder;
```
staabm, line 3396: I get the same test-errors no matter whether I re-use the whole $scope, or just the `ExpressionTypeHolder`. need to debug it further.

interessting test errors are:
```
PHPStan\Rules\Classes\ImpossibleInstanceOfRuleTest::testNewIsAlwaysFinalClass
--- Expected
+++ Actual
@@ @@
 '17: Instanceof between ImpossibleInstanceofNewIsAlwaysFinal\Bar and ImpossibleInstanceofNewIsAlwaysFinal\Foo will always evaluate to false.
-33: Instanceof between ImpossibleInstanceofNewIsAlwaysFinal\Bar and ImpossibleInstanceofNewIsAlwaysFinal\Foo will always evaluate to false.
-43: Instanceof between ImpossibleInstanceofNewIsAlwaysFinal\Bar and ImpossibleInstanceofNewIsAlwaysFinal\Foo will always evaluate to false.
-53: Instanceof between ImpossibleInstanceofNewIsAlwaysFinal\Bar and ImpossibleInstanceofNewIsAlwaysFinal\Foo will always evaluate to false.
-63: Instanceof between ImpossibleInstanceofNewIsAlwaysFinal\Bar and ImpossibleInstanceofNewIsAlwaysFinal\Foo will always evaluate to false.
 73: Instanceof between ImpossibleInstanceofNewIsAlwaysFinal\Bar|null and ImpossibleInstanceofNewIsAlwaysFinal\Foo will always evaluate to false.
 88: Instanceof between ImpossibleInstanceofNewIsAlwaysFinal\Bar|null and ImpossibleInstanceofNewIsAlwaysFinal\Baz will always evaluate to false.
 '

/Users/staabm/workspace/phpstan-src/src/Testing/RuleTestCase.php:236
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Rules/Classes/ImpossibleInstanceOfRuleTest.php:529

```

PR #4849, opened 2026-02-02
mb_str_pad is an impure function

longwave:
Reasoning for this change is in https://github.com/JetBrains/phpstorm-stubs/pull/1793 - I think this is the correct change to make on the PHPStan side.

Helps #4845 and in turn https://github.com/phpstan/phpstan/issues/14041


resources/functionMetadata.php
```diff
@@ -1496,13 +1496,16 @@
...
 	'mb_list_encodings' => ['hasSideEffects' => false],
+	'mb_ltrim' => ['hasSideEffects' => false],
 	'mb_ord' => ['hasSideEffects' => false],
 	'mb_output_handler' => ['hasSideEffects' => false],
 	'mb_preferred_mime_name' => ['hasSideEffects' => false],
+	'mb_rtrim' => ['hasSideEffects' => false],
 	'mb_scrub' => ['hasSideEffects' => false],
 	'mb_split' => ['hasSideEffects' => false],
-	'mb_str_pad' => ['hasSideEffects' => false],
+	'mb_str_pad' => ['hasSideEffects' => true],
```
ondrejmirtes, line 1508: What side effect does this function have?

ondrejmirtes (changes requested):

resources/functionMetadata.php
longwave:
```
 mb_str_pad(
    string $string,
    int $length,
    string $pad_string = " ",
    int $pad_type = STR_PAD_RIGHT,
    ?string $encoding = null
): string
```
If `$encoding` is null then the internal encoding is used, which can be changed by `mb_internal_encoding()`. This is not a side effect but means it's not entirely a pure function?

ondrejmirtes: Alright, good enough :) Thank you.

PR #4852, opened 2026-02-03
IterableType: Remove unnecessary work in toCoercedArgumentType()

staabm:


src/Type/IterableType.php
```diff
@@ -241,13 +241,7 @@ public function toCoercedArgumentType(bool $strictTypes): Type
 		return TypeCombinator::union(
 			$this,
 			new ArrayType(
-				TypeCombinator::intersect(
-					$this->keyType->toArrayKey(),
```
staabm, line 245: my thinking is that `$this->keyType->toArrayKey()` should already by definition return a subtype of `int|string`

ondrejmirtes: Can you check that things like MixedType, StrictMixedType and TemplateMixedType do that?

My worry is also that some types return ErrorType so keeping this would make sense.

staabm: StrictMixedType returns a ErrorType

https://github.com/phpstan/phpstan-src/blob/13a2f02edb0f763049973ae926dc10b6899a7cff/src/Type/StrictMixedType.php#L426

PR #4856, opened 2026-02-03
ConstantArrayTypeBuilder: Raise ARRAY_COUNT_LIMIT to 512

staabm:
closes https://github.com/phpstan/phpstan/issues/8636

running `make tests` before and after this PR results in ~30 seconds runtime on my machine, so it seems we are fast enough to allow bigger arrays

---

I changed the `tests/PHPStan/Analyser/data/bug-5081.php` test to accomodate 1 element more as we allow.
all other tests have just been adjusted to reflect the new reality.

`tests/PHPStan/Rules/Methods/data/bug-8636.php` reported `Method Config::huge() should return string but returns array<int, string>|string.` before this PR and does not longer after this PR.


staabm:
//cc @gnutix please try the .phar file from the build artifacts of this PR and check it on your problematic codebase from https://github.com/phpstan/phpstan/issues/8636

tests/PHPStan/Analyser/nsrt/bug-11703.php
```diff
@@ -90,7 +90,7 @@ function test(int|string|null $duration, string $foo): ?bool
 		];
 	}
 
-	assertType('int<0, max>', count($alerts));
```
VincentLanglet, line 93: Most of those tests were about oversizedArray.
Like this one https://github.com/phpstan/phpstan/issues/11703 reporting an oversized array was considered as always non-empty.

With this "test update", the regression is not covered anymore since we are not testing oversizedArray anymore but just big constantArray.

If you change ARRAY_COUNT_LIMIT, IMHO all the oversized array in the tests need to be updated to stay oversized array.
Which is some works...

VincentLanglet (commented):
I feel like https://github.com/phpstan/phpstan/issues/8636#issuecomment-1374807798 would be a great way to solve such issue rather than updating the limit for every projects

staabm:
> I feel like [phpstan/phpstan#8636 (comment)](https://github.com/phpstan/phpstan/issues/8636#issuecomment-1374807798) would be a great way to solve such issue rather than updating the limit for every projects

see https://github.com/phpstan/phpstan-src/pull/4850#issuecomment-3836219009

tests/PHPStan/Analyser/nsrt/bug-11703.php
staabm: I guess I should adjust that some tests still use the 256 limit

VincentLanglet: > I guess I should adjust that some tests still use the 256 limit

That could be a great way indeed ;

And if the limit is configurable for tests (only cf https://github.com/phpstan/phpstan-src/pull/4856#issuecomment-3846559665 indeed) it could give us easier test writing of oversize array by lowering the limit when needed (?)

src/DependencyInjection/ContainerFactory.php
```diff
@@ -199,6 +200,7 @@ public static function postInitializeContainer(Container $container): void
 		$container->getService('typeSpecifier');
 
 		BleedingEdgeToggle::setBleedingEdge($container->getParameter('featureToggles')['bleedingEdge']);
+		ConstantArrayTypeBuilder::setArrayCountLimit(256);
```
VincentLanglet: Should it be the opposite ?
Letting 512 as default, and manually calling `setArrayCountLimit(256)` for test which need it ?

Maybe it's because I'm not familiar with this code but setting 256 in `postInitializeContainer` won't also set it for user of phpstan (and not only for tests) ?

staabm:
tried a few different things but was not yet successful to make the setting have a different value for certain test cases, as I don't want to introduce a new container parameter (which would allow people to define the setting at app level).

gnutix:
I ran benchmarks on our codebase (~4min total PHPStan runtime) by patching the ARRAY_COUNT_LIMIT constant directly in the 2.1.38 PHAR. Cold cache for each run.

| Limit | phpstan-main | phpstan-legacy | Total |                                                                     
|---|---|---|---|                                        
| **256** (current) | 127.93s | 112.90s | **240.83s** |                                                                                     
| **512** (this PR) | 141.98s | 99.42s | **241.40s** |                                                                                      
| **1024** (because why not) | 126.47s | 99.38s | **225.85s** | 

Single run each (no iterations), so variance across rows is just noise. The totals are essentially identical — no measurable performance impact from doubling or even quadrupling the limit on this codebase.

Given that there's no performance regression (or at least measurable on this codebase, I know), would it make sense to remove the limit entirely? or make it a configurable parameter? I can't say I understand the strategy of "just doubling it", it feels arbitrary if we can't measure differences in benchmarks.

staabm:
> Given that there's no performance regression (or at least measurable on this codebase, I know), would it make sense to remove the limit entirely? or make it a configurable parameter? I can't say I understand the strategy of "just doubling it", it feels arbitrary if we can't measure differences in benchmarks.

I am adjusting the implementation to fix the reported reproducible problem in https://github.com/phpstan/phpstan/issues/8636

anything else I would do is plain guesswork. I cannot judge your case because there is no reproducer.

VincentLanglet:
This seems unneeded with https://github.com/phpstan/phpstan-src/pull/5555

PR #4857, opened 2026-02-03
Comparison with strtolower() etc. leads lower/upper-case-string

staabm:
closes https://github.com/phpstan/phpstan/issues/14047

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/non-empty-string-strcasing-specifying.php
+++ b/tests/PHPStan/Analyser/nsrt/non-empty-string-strcasing-specifying.php
@@ -8,11 +8,11 @@ class Foo {
 	public function strtolower(string $s): void
 	{
 		if (strtolower($s) === 'hallo') {
-			assertType('lowercase-string&non-falsy-string', $s);
+			assertType('non-falsy-string', $s);
 		}
 		assertType('string', $s);
 		if ('hallo' === strtolower($s)) {
-			assertType('lowercase-string&non-falsy-string', $s);
+			assertType('non-falsy-string', $s);
 		}
 		assertType('string', $s);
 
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/nsrt/non-empty-string-strcasing-specifying.php
```diff
@@ -48,11 +48,11 @@ public function strtoupper(string $s): void
 	public function mb_strtoupper(string $s): void
 	{
 		if (mb_strtoupper($s) === 'HA') {
-			assertType('non-falsy-string', $s);
+			assertType('non-falsy-string&uppercase-string', $s);
 		}
 		assertType('string', $s);
 		if ('hallo' === mb_strtoupper($s)) {
-			assertType('non-falsy-string', $s);
```
staabm, line 55: if we wanne be super correct, this should be `*NEVER*`... a task for a future PR

edit: reported as https://github.com/phpstan/phpstan/issues/14058

staabm:
@VincentLanglet would love your feedback, as you implemented lowser/upper case handling IIRC

tests/PHPStan/Analyser/nsrt/non-empty-string-strcasing-specifying.php
```diff
@@ -8,11 +8,11 @@ class Foo {
 	public function strtolower(string $s): void
 	{
 		if (strtolower($s) === 'hallo') {
-			assertType('non-falsy-string', $s);
+			assertType('lowercase-string&non-falsy-string', $s);
```
VincentLanglet: This is wrong, it's not because strtolower($s) is a lowerstring than $s is a lowercase string.

It's only the case if `$s === strtolower($something)`

VincentLanglet (changes requested):

tests/PHPStan/Analyser/nsrt/non-empty-string-strcasing-specifying.php
staabm: you are right. I mixed up the operands 🫣

VincentLanglet (approved):

PR #4861, opened 2026-02-09
Refactor InitializerExprTypeResolver

staabm:
After looking again at https://github.com/phpstan/phpstan-src/pull/4435 I think this minor simplification makes sense

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -976,39 +976,10 @@ final class InitializerExprTypeResolver
 			return $this->getNeverType($leftType, $rightType);
 		}
 
-		$leftTypes = $this->getFiniteOrConstantScalarTypes($leftType);
-		$rightTypes = $this->getFiniteOrConstantScalarTypes($rightType);
-
-		$leftTypesCount = count($leftTypes);
-		$rightTypesCount = count($rightTypes);
-		if ($leftTypesCount > 0 && $rightTypesCount > 0) {
-			$resultTypes = [];
-			$generalize = $leftTypesCount * $rightTypesCount > self::CALCULATE_SCALARS_LIMIT;
-			if (!$generalize) {
-				foreach ($leftTypes as $leftTypeInner) {
-					foreach ($rightTypes as $rightTypeInner) {
-						if ($leftTypeInner instanceof ConstantStringType && $rightTypeInner instanceof ConstantStringType) {
-							$resultType = $this->getTypeFromValue($leftTypeInner->getValue() & $rightTypeInner->getValue());
-						} else {
-							$leftNumberType = $leftTypeInner->toNumber();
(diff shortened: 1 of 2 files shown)


VincentLanglet (approved):
I was wondering if this simplification was interesting enough or if we should/could include the whole `if ($leftTypesCount > 0 ...)` given the fact the only changed line were
```
$resultType = $this->getTypeFromValue($leftTypeInner->getValue() | $rightTypeInner->getValue());
$resultType = $this->getTypeFromValue($leftTypeInner->getValue() & $rightTypeInner->getValue());
$resultType = $this->getTypeFromValue($leftTypeInner->getValue() ^ $rightTypeInner->getValue());
```

and could be resolve with a callback ; WDYT ?

staabm:
> and could be resolve with a callback ; WDYT ?

good idea - did it like that now.

my initial idea was to turn the pattern

```
		if ($leftType instanceof IntegerRangeType) {
			$leftTypes = $leftType->getFiniteTypes();
		} else {
			$leftTypes = $leftType->getConstantScalarTypes();
		}
```

into a `TypeUtils` method, so we can re-use it even outside of `InitializerExprTypeResolver`.
but for the `InitializerExprTypeResolver`-local refactoring it makes sense to go into your idea first.

thank you

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1150,12 +1076,16 @@ public function getBitwiseXorType(Expr $left, Expr $right, callable $getTypeCall
...
-		}
+	private const IS_SCALAR_TYPE = 1;
+	private const IS_UNKNOWN = 2;
 
+	/**
+	 * @param callable(bool|float|int|string|null, bool|float|int|string|null):string $operationCallable
+	 *
+	 * @return self::IS_UNKNOWN|self::IS_SCALAR_TYPE|Type
+	 */
+	private function getFiniteOrConstantScalarTypes(Type $leftType, Type $rightType, callable $operationCallable): int|Type
```
VincentLanglet, line 1087: The method might be easier to read with early return ?

````
if ($leftTypesCount === 0 || $rightTypesCount === 0) {
     return self::IS_UNKNOWN;
}
...
if ($generalize) {
     return self::IS_SCALAR_TYPE;
}
...
return TypeCombinator::union(...$resultTypes);
```

I also wonder if a Type|bool isn't enough (and avoid introducing two constant). True when it need to be generalized.
Another solution could be including the generalize call inside this method and returning `[$leftType, $rightType, $result]` but I'm not sure it should be preferred.

staabm: > The method might be easier to read with early return ?

done.

> I also wonder if a Type|bool isn't enough (and avoid introducing two constant).

I had it like that before I introduced constants. was not happy with the client-side code of this method as it was pretty unreadable.

VincentLanglet (approved):

PR #4863, opened 2026-02-09
Narrow PHP_VERSION_ID on version_compare()

staabm:
closes https://github.com/phpstan/phpstan/issues/13904


src/Type/Php/VersionCompareFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,141 @@
...
+			$version1 instanceof ConstFetch
+			&& $version1->name->name === 'PHP_VERSION'
+			&& $version2 instanceof String_
+		) {
+			$integerVersionRange = $this->getVersionCompareType($version2->value, isset($args[2]) ? $args[2]->value : null, $scope);
+
+			if ($integerVersionRange !== null) {
+				$narrowedVersion = TypeCombinator::intersect($scope->getPhpVersion()->getType(), $integerVersionRange);
+				return $this->typeSpecifier->create(
+					new ConstFetch(new Name('\\PHP_VERSION_ID')),
```
staabm, line 68: I feel I am hitting a bug here, as depending on how I name the constant here
`PHP_VERSION_ID` vs. `\\PHP_VERSION_ID` the inference does not work as expected

(and it seems it depends on whether the code beeing analyzed is namespaced or not)

PR #4864, opened 2026-02-10
Do not degrade array-shapes in phpdocs

staabm:
as requested in https://github.com/phpstan/phpstan-src/pull/4823#issuecomment-3876240786


src/PhpDoc/TypeNodeResolver.php
```diff
@@ -1062,6 +1058,7 @@ private function resolveArrayShapeNode(ArrayShapeNode $typeNode, NameScope $name
 			$builder->setOffsetValueType($offsetType, $this->resolve($itemNode->valueType, $nameScope), $itemNode->optional);
 		}
 
+		$builder->disableArrayDegradation();
```
ondrejmirtes: I'd say this method should be called right after build creation because inevitably it influences what happens inside setOffset method too.

I think we need another property to track this setting.

ondrejmirtes (changes requested):
Also the performance should be observed with special care. Does anything gets slower because of this?

staabm:
> Does anything gets slower because of this?

before PR:
```
➜  phpstan-src git:(2.1.x) ✗ hyperfine 'vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php' --runs=5 
Benchmark 1: vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php
  Time (mean ± σ):     17.367 s ±  0.437 s    [User: 15.445 s, System: 0.599 s]
  Range (min … max):   16.896 s … 18.083 s    5 runs
```

with this PR https://github.com/phpstan/phpstan-src/pull/4864/commits/8c8b65a5e0504a308917b0cd2dbfe33658897564:
```
➜  phpstan-src git:(dis-arr) ✗ hyperfine 'vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php' --runs=5 
Benchmark 1: vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php
  Time (mean ± σ):     17.080 s ±  0.285 s    [User: 14.936 s, System: 0.631 s]
  Range (min … max):   16.744 s … 17.378 s    5 runs
  ```

I think its within the margin of error (because my company laptop has varying performance because of anti-virus stuff)

**Comment by ondrejmirtes** [comment]
I'm still worried this can break stuff. I imagine that some `$this->degradeToGeneralArray = true;` assignments should still be allowed to happen but some should be guarded with `!$this->disableArrayDegradation`. Instead of your current way of changing conditions into `if (!$this->degradeToGeneralArray || $this->disableArrayDegradation) {`.

One example: we want `$this->degradeToGeneralArray = true;` to happen and be respected at the bottom of `setOffsetValueType` method. Because when someone assigns general `IntegerType`, it doesn't make sense to keep the array shape.

**Comment by staabm** [comment]
ok, how do you feel about it now?

with this PR https://github.com/phpstan/phpstan-src/pull/4864/commits/71303cc4855c8721629389d94d941e02fe972a08:

```
➜  phpstan-src git:(dis-arr) ✗ hyperfine 'vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php' --runs=5 
Benchmark 1: vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php
  Time (mean ± σ):     16.712 s ±  0.157 s    [User: 15.144 s, System: 0.609 s]
  Range (min … max):   16.459 s … 16.848 s    5 runs
```

ondrejmirtes:
This makes sense to me now :) Thank you!

ondrejmirtes (approved):

PR #4867, opened 2026-02-10
Fix GMP::toNumber

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13775

(In my understanding

Closes https://github.com/phpstan/phpstan/issues/12123

(Improved later with https://github.com/phpstan/phpstan-src/pull/4873)

### DIFF
--- a/tests/PHPStan/Rules/Functions/data/bug-13775.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-13775.php
@@ -30,5 +30,8 @@ echo 'array, invalid cast: ' . array_product($a) . "\n\n";
 $a = [new stdClass()];
 echo 'class that does not auto-cast, invalid cast: ' . array_product($a) . "\n\n";
 
+$a = [rand(0, 1) ? new stdClass() : gmp_init(42)];
+echo 'possibly invalid cast: ' . array_product($a) . "\n\n";
+
 $a = [gmp_init(42)];
 echo 'gmp, valid cast: ' . array_product($a) . "\n\n";
--- a/tests/PHPStan/Rules/Functions/data/bug-13775.php
+++ b/tests/PHPStan/Rules/Functions/data/bug-13775.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+$a = [null];
+echo 'null, valid cast: ' . array_product($a) . "\n\n";
+
+$a = [false];
+echo 'false, valid cast: ' . array_product($a) . "\n\n";
+
+$a = [true];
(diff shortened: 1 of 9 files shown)


staabm:
this seems to also affect https://github.com/phpstan/phpstan/issues/12123 in which ondrej suggested to use a [operator overloading extension](https://github.com/phpstan/phpstan/issues/12123#issuecomment-2500814285).

tests/PHPStan/Rules/Functions/data/bug-13775.php
```diff
@@ -0,0 +1,34 @@
+<?php declare(strict_types = 1);
```
staabm, line 1: issue 13775 also talks about `array_sum` which is not yet covered in tests

staabm:
in https://github.com/phpstan/phpstan/issues/13775#issuecomment-3503627982 there is a php version difference mentioned. should this also be taken into account here?

tests/PHPStan/Rules/Functions/data/param-castable-to-number-functions.php
```diff
@@ -40,6 +40,6 @@ function wrongNumberOfArguments(): void
 
 function validUsages(): void
 {
-	var_dump(array_sum(['5.5', false, true, new \SimpleXMLElement('<a>7.7</a>'), 5, 5.5, null]));
-	var_dump(array_product(['5.5', false, true, new \SimpleXMLElement('<a>7.7</a>'), 5, 5.5, null]));
```
VincentLanglet, line 44: This wasn't a valid usage, see 
https://3v4l.org/GNJiY#v8.2.30

The SimpleXMLElement is silently ignored

tests/PHPStan/Rules/Functions/data/bug-13775.php
VincentLanglet: Done

staabm (approved):

Firehed:
Thanks for contributing this! I'm the original author of https://github.com/phpstan/phpstan/issues/12123, and after re-analyzing code with this change in place (PHPStan 2.1.40) it's now reporting more errors than the original, despite being clearly operationally correct at runtime.

https://phpstan.org/r/6063122e-1a41-4286-a664-78484599152a

I wasn't quite sure how to interpret the comment about the extension, e.g. whether that was in reference to PHPStan-internal code or actually setting up a (third-party?) extension, or maybe it needs to handle other operators? I'll take a look at some follow-ups here.

ondrejmirtes:
Please do not comment on merged month-old PRs. Open a new bug report insteas. Thank you.

PR #4871, opened 2026-02-11
Stop ignoring offsetAccess.invalidOffset

VincentLanglet:
In this case we're just using true/false as key, so they can be converted to int safely.

### DIFF
--- a/src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
+++ b/src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
@@ -120,7 +120,7 @@ final class VersionCompareFunctionDynamicReturnTypeExtension implements DynamicF
 						}
 
 						$value = version_compare($version1String->getValue(), $version2String->getValue(), $operatorValue);
-						$types[$value] = new ConstantBooleanType($value); // @phpstan-ignore offsetAccess.invalidOffset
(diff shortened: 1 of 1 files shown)


src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
```diff
@@ -120,7 +120,7 @@ public function getTypeFromFunctionCall(
 						}
 
 						$value = version_compare($version1String->getValue(), $version2String->getValue(), $operatorValue);
-						$types[$value] = new ConstantBooleanType($value); // @phpstan-ignore offsetAccess.invalidOffset
+						$types[(int) $value] = new ConstantBooleanType($value);
 					}
 				} else {
 					$value = version_compare($version1String->getValue(), $version2String->getValue());
```
staabm, line 126: interessting that line 127 does not have the same problem

staabm (approved):

src/Type/Php/VersionCompareFunctionDynamicReturnTypeExtension.php
staabm: ohh I see its

> When using the optional operator argument, the function will return [true](https://www.php.net/manual/en/reserved.constants.php#constant.true) if the relationship is the one specified by the operator, [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) otherwise.

PR #4872, opened 2026-02-11
Extract FileAnalyserCallback

staabm:
As requested in https://github.com/phpstan/phpstan-src/pull/4724#issuecomment-3883153574


ondrejmirtes:
This is so nice! Thank you!

PR #4873, opened 2026-02-11
Respect phpstan level in ParameterCastableToNumberRule

VincentLanglet:
Complete https://github.com/phpstan/phpstan-src/pull/4867

### DIFF
--- a/src/Rules/Functions/ParameterCastableToNumberRule.php
+++ b/src/Rules/Functions/ParameterCastableToNumberRule.php
@@ -11,6 +11,7 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\ParameterCastableToStringCheck;
 use PHPStan\Rules\Rule;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use function count;
 use function in_array;
@@ -69,7 +70,7 @@ final class ParameterCastableToNumberRule implements Rule
 
 		$castFn = $this->phpVersion->supportsObjectsInArraySumProduct()
 			? static fn (Type $t) => $t->toNumber()
-			: static fn (Type $t) => !$t->isObject()->no() ? new ErrorType() : $t->toNumber();
+			: static fn (Type $t) => !$t instanceof MixedType && !$t->isObject()->no() ? new ErrorType() : $t->toNumber();
 
 		$error = $this->parameterCastableToStringCheck->checkParameter(
 			$origArgs[0],
--- a/src/Rules/ParameterCastableToStringCheck.php
+++ b/src/Rules/ParameterCastableToStringCheck.php
@@ -6,6 +6,7 @@ use PhpParser\Node\Arg;
 use PHPStan\Analyser\ArgumentsNormalizer;
 use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Node\Expr\TypeExpr;
 use PHPStan\Reflection\ParameterReflection;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\Type;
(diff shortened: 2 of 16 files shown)


src/Rules/ParameterCastableToStringCheck.php
```diff
@@ -34,22 +35,31 @@ public function checkParameter(
 			return null;
 		}
 
-		$typeResult = $this->ruleLevelHelper->findTypeToCheck(
+		$arrayTypeResult = $this->ruleLevelHelper->findTypeToCheck(
 			$scope,
 			$parameter->value,
 			'',
-			static fn (Type $type): bool => $type->isArray()->yes() && !$castFn($type->getIterableValueType()) instanceof ErrorType,
```
VincentLanglet, line 41: The following change follows the same idea than https://github.com/phpstan/phpstan-src/pull/4166/changes for instance.

`&&` and `||` doesn't work always very well in UnionTypeCriteriaCallback it's sometimes better to split the check into multiple findTypeToCheck calls.

Here this allows to transform the iterable value type `$type->getIterableValueType()` into a StrictMixed when it's a Mixed for instance or `array<mixed>` is never reported at any level.

src/Rules/Functions/ParameterCastableToNumberRule.php
```diff
@@ -69,7 +70,7 @@ public function processNode(Node $node, Scope $scope): array
 
 		$castFn = $this->phpVersion->supportsObjectsInArraySumProduct()
 			? static fn (Type $t) => $t->toNumber()
-			: static fn (Type $t) => !$t->isObject()->no() ? new ErrorType() : $t->toNumber();
+			: static fn (Type $t) => $t instanceof MixedType || $t->isObject()->no() ? $t->toNumber() : new ErrorType();
```
VincentLanglet: MixedType should not be transform automatically to ErrorType, we will rely on the ruleLevelHelper inside the parameterCastableToStringCheck to decide.

VincentLanglet:
Some errors are new on [Compile PHAR / integration-tests / PMMP Tests (pull_request)](https://github.com/phpstan/phpstan-src/actions/runs/21917103294/job/63287441594?pr=4873) I need to check

Edit: This is expected ! They are using the level 9
https://github.com/pmmp/PocketMine-MP/blob/e55b272280ba45f2d363167b586e628327fd43a0/phpstan.neon.dist#L22

staabm (commented):
I would prefer @ondrejmirtes reviewing this one, as my experience with RuleLevelHelper is very limited

src/Rules/ParameterCastableToStringCheck.php
```diff
@@ -34,22 +35,31 @@ public function checkParameter(
...
+			new TypeExpr($arrayType->getIterableValueType()),
+			'',
+			static fn (Type $type): bool => !$castFn($type) instanceof ErrorType,
 		);
 
-		if (
-			! $typeResult->getType()->isArray()->yes()
-			|| !$castFn($typeResult->getType()->getIterableValueType()) instanceof ErrorType
-		) {
+		if (!$castFn($typeResult->getType()) instanceof ErrorType) {
```
ondrejmirtes, line 57: If you're not sure about how to work with RuleLevelHelper::findTypeToCheck, write thorough test for LevelsIntegrationTest. (The test regenerates the JSON files with expectations so 2nd run should always succeed.)

Before level 7, it should only complain about a type that is never castable to number.
On level 7+, it should complain about unions where only some types are castable to number.
On level 8+, it should complain about a nullable type (but maybe since null is castable to number, there will be no difference?)
On level 9+, it should complain about explicit mixed.
On level 10, it should complain about implicit mixed (without typehint).

Also I suspect we'll need also a separate test file for PHP 8.3+ because of supportsObjectsInArraySumProduct.

ondrejmirtes (changes requested):

src/Rules/ParameterCastableToStringCheck.php
VincentLanglet: I added tests in LevelsIntegrationTest to show the result.

I don't think it's worth using separate test file for PHP 8.3+ or - ; because the only difference is for some specific object type which are already tested.

ondrejmirtes (changes requested):
It's okay, just the mutation needs sorting out :)

VincentLanglet:
> It's okay, just the mutation needs sorting out :)

The first mutation
```
$castFn = $this->phpVersion->supportsObjectsInArraySumProduct()
 			? static fn (Type $t) => $t->toNumber()
-			: static fn (Type $t) => !$t instanceof MixedType && !$t->isObject()->no() ? new ErrorType() : $t->toNumber();
+			: static fn (Type $t) => !$t instanceof MixedType && $t->isObject()->yes() ? new ErrorType() : $t->toNumber();
```
is reported because `supportsObjectsInArraySumProduct` is always true in PHP 8.3+ so the second path is not used during mutation testing (running only on 8.3 and 8.4).

The second mutation is now handled.

Seems like some new mutations are reported by the Ci but on totally unrelated code.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #4891, opened 2026-02-12
Loadable classes not recognized since 2.1.34 release

staabm:
closes https://github.com/phpstan/phpstan/issues/14036
closes https://github.com/phpstan/phpstan/issues/14099

fixes regression introduced in https://github.com/phpstan/phpstan-src/commit/752bed7

the regression test only reproduces on linux - not on macOs/Windows


staabm:
(failling tests are unrelated.. also fail on 2.1.x)

VincentLanglet (approved):

xHeaven:
Thanks @staabm!

PR #4922, opened 2026-02-13
MutatingScope: prevent unnecessary work in removeTypeFromExpression()

staabm:
removes a few unnecessary `MutatingScope->getType()` calls


ondrejmirtes:
Thank you!

PR #4923, opened 2026-02-13
DependencyResolver: Reduce duplicate work

staabm:
before 

grafik

after

grafik


staabm:
seems I am not able to find a cache key which works for all e2e tests

ondrejmirtes:
Thank you!

PR #4924, opened 2026-02-14
Don't invalidate container cache on user-interactions in the shell

staabm:
refs https://github.com/phpstan/phpstan/issues/14072#issuecomment-3901284202

---

ENV variables used for building the container key sometimes change, based on user interactions in the shell.

Image

it happens after you `cd` between folders and run PHPStan again.

see https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html

> OLDPWD
> The previous working directory as set by the cd builtin.

---

I see another case: the container invalidates as soon as you re-size the shell window, because a ENV var exists which defines the LINES and COLUMNS in the window

Image

---

while I think this is just a workaround for a more general problem - I see no way atm to find out which ENV vars are relevant for us and which are not


ondrejmirtes:
Thank you!

PR #4925, opened 2026-02-14
Make sure invocations via blackfire use the same container

staabm:
analog https://github.com/phpstan/phpstan-src/pull/4924

refs https://github.com/phpstan/phpstan/issues/14072

blackfire adds a ENV var

grafik


ondrejmirtes:
Thank you!

PR #4926, opened 2026-02-14
ClassNameCheck: Retrieve extensions just once

staabm:


src/Rules/ClassNameCheck.php
```diff
@@ -12,13 +12,17 @@
...
+	private array $extensions;
+
 	public function __construct(
 		private ClassCaseSensitivityCheck $classCaseSensitivityCheck,
 		private ClassForbiddenNameCheck $classForbiddenNameCheck,
 		private ReflectionProvider $reflectionProvider,
 		private Container $container,
 	)
 	{
+		$this->extensions = $this->container->getServicesByTag(RestrictedClassNameUsageExtension::CLASS_NAME_EXTENSION_TAG);
```
ondrejmirtes: I don't want to do stuff in constructor, you're risking circular references. It has to be done lazily with `??=` and nullable property.

ondrejmirtes (changes requested):

src/Rules/ClassNameCheck.php
staabm: ahh right. we had this discussion already. I forgot.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #4927, opened 2026-02-14
Fix inconsistent type inference

xHeaven:
Fixes https://github.com/phpstan/phpstan/issues/14105.


xHeaven:
Welp, I've accidentally branched from 2.2.x. Closing this.

PR #4928, opened 2026-02-14
Fix inconsistent type inference

xHeaven:
Fixes https://github.com/phpstan/phpstan/issues/14105.

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2225,10 +2225,17 @@ class NodeScopeResolver
 				if ($scope->getClassReflection() === null) {
 					throw new ShouldNotHappenException();
 				}
+				$constType = $scope->getType($const->value);
+				$constNativeType = $scope->getNativeType($const->value);
 				$scope = $scope->assignExpression(
 					new Expr\ClassConstFetch(new Name\FullyQualified($scope->getClassReflection()->getName()), $const->name),
-					$scope->getType($const->value),
-					$scope->getNativeType($const->value),
+					$constType,
+					$constNativeType,
+				);
+				$scope = $scope->assignExpression(
+					new Expr\ClassConstFetch(new Name('self'), $const->name),
+					$constType,
+					$constNativeType,
 				);
 			}
 		} elseif ($stmt instanceof Node\Stmt\EnumCase) {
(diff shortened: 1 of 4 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -1062,6 +1062,19 @@ private function processStmtNode(
 			});
 
 			$this->processStmtNodesInternal($stmt, $classLikeStatements, $classScope, $storage, $classStatementsGatherer, $context);
+			foreach ($stmt->stmts as $classLikeStmt) {
+				if (!$classLikeStmt instanceof Node\Stmt\ClassConst || !$classLikeStmt->isPublic()) {
```
staabm, line 1066: Please also add a test when the constant is private:

https://3v4l.org/qhTGS#veol

I wonder why this works before the PR already

https://phpstan.org/r/5e8b0fd5-fa52-4787-bcd5-cb25eed12508

src/Analyser/NodeScopeResolver.php
```diff
@@ -1062,6 +1062,19 @@ private function processStmtNode(
...
+					continue;
+				}
+
+				foreach ($classLikeStmt->consts as $const) {
+					$scope = $scope->assignExpression(
```
staabm, line 1071: I wonder why we don't know the full closure type, whereever it is assigned/calculated before this PR

src/Analyser/NodeScopeResolver.php
xHeaven: I'm gonna add a few more tests with fixes. Also, I think I've accidentally fixed this: https://github.com/phpstan/phpstan-src/blob/5abeb3cceedf31dc51dc49a3fd7851dbf36de957/tests/PHPStan/Analyser/nsrt/initializer-expr-type-resolver.php#L17

I'm not sure if this was supposed to be fixed? It would make sense, though. Now it correctly resolves to `42` and the `*ERROR*` assertion fails.

> I wonder why this works before the PR already
> [phpstan.org/r/5e8b0fd5-fa52-4787-bcd5-cb25eed12508](https://phpstan.org/r/5e8b0fd5-fa52-4787-bcd5-cb25eed12508)

It doesn't look like working to me. It prints `Dumped type: Closure(int): array`, it should print `Dumped type: Closure(int): array{num: int}`.

src/Analyser/NodeScopeResolver.php
xHeaven: We already inferred it, but scope transitions dropped `ClassConstFetch`/`self::CONST` info, so lookup fell back to reflection and lost shape precision.

xHeaven:
Pushed some more tests and changes.
<hr>

Oh well, it seems this broke some things in the legacy resolver...

PR #4930, opened 2026-02-14
Cache PHPStorm stubs

staabm:
refs https://github.com/phpstan/phpstan/issues/14072#issuecomment-3901567221

requires https://github.com/ondrejmirtes/BetterReflection/pull/40

---

with caching we see a ~5-6% perf improvement when result-cache is warmed. (single process with `--debug`)

grafik

---

I guess the caching is even more effective in parallel runs because it removes PHP-Parser::parse() of PHPStorm stubs on every worker-process creation.


src/Reflection/BetterReflection/SourceStubber/CachedPhpStormStubsSourceStubber.php
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	/**
+	 * @return array{non-empty-string, string}
+	 */
+	private function getCacheKeys(): array
+	{
+		$stubsVersion = ComposerHelper::getPhpStormStubsVersion();
+		$cacheKey = sprintf('phpstorm-stubs-%s', $stubsVersion);
+		$variableCacheKey = sprintf('v1-%s-%s', ComposerHelper::getBetterReflectionVersion(), $this->phpVersion->getVersionString());
```
staabm: not sure whether this keys are correct this way.
whats the difference between `$cacheKey` and `$variableCacheKey`?

ondrejmirtes (changes requested):
Personally I'd rather cache the same things the current locators cache. If you cache the "stubs", they are just PHP code that still needs to be parsed by the locator. So rather I'd have a caching locator around PhpInternalSourceLocator that would cache reflection objects which already have `importFromCache`/`exportToCache`.

src/Reflection/BetterReflection/SourceLocator/FileCachedSourceLocator.php
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	private function storeOnShutdown(): void
+	{
+		if ($this->storeOnShutdown) {
+			return;
+		}
+
+		$this->storeOnShutdown = true;
+		register_shutdown_function([$this, 'storeCache']);
```
ondrejmirtes: Why does this need such a hack? Could it store things to cache like other things do? When it's requested and computed for the first time?

ondrejmirtes (changes requested):

src/Reflection/BetterReflection/SourceLocator/FileCachedSourceLocator.php
```diff
@@ -0,0 +1,165 @@
...
+			}
+			$restored['functions'][$class] = ReflectionFunction::importFromCache($reflector, $cachedReflection);
+		}
+		foreach ($cached['constants'] ?? [] as $constantName => $cachedReflection) {
+			if ($cachedReflection === null) {
+				$restored['constants'][$constantName] = null;
+				continue;
+			}
+
+			$restored['constants'][$constantName] = ReflectionConstant::importFromCache($reflector, $cachedReflection);
```
staabm: ReflectionConstant/ReflectionParameter->importFromCache() is expensive.
it is dominated by PHPParser overhead


Image

looks like the parser instantiation happens multiple times

src/Reflection/BetterReflection/SourceLocator/FileCachedSourceLocator.php
staabm: my feeling is calling `store` on every diff could result in a really slow first analysis when no/most cache files are not present yet (or when phpstorm stubs are updated)

src/Reflection/BetterReflection/SourceLocator/FileCachedSourceLocator.php
staabm: fix in https://github.com/ondrejmirtes/BetterReflection/pull/41

ondrejmirtes:
For example one downside of this approach is that it loads into memory symbols it might not need in the current run.

src/Internal/ComposerHelper.php
```diff
@@ -124,6 +126,21 @@ public static function getBetterReflectionVersion(): string
...
+	public static function getPhpStormStubsVersion(): string
+	{
+		if (self::$phpstomStubsVersion !== null) {
+			return self::$phpstomStubsVersion;
+		}
+
+		$installed = self::getInstalled();
+		$rootPackage = $installed['versions']['jetbrains/phpstorm-stubs'] ?? null;
+		if ($rootPackage === null) {
+			return self::$phpstomStubsVersion = self::UNKNOWN_VERSION;
```
ondrejmirtes: typo `phpstom`

src/Reflection/BetterReflection/SourceLocator/FileCachedSourceLocator.php
```diff
@@ -0,0 +1,167 @@
...
+					&& !$reflection instanceof ReflectionConstant
+				) {
+					throw new ShouldNotHappenException();
+				}
+
+				$exported[$type][$name] = $reflection->exportToCache();
+			}
+		}
+
+		$this->cache->save($this->cacheKey, $variableCacheKey, $exported);
```
ondrejmirtes: Why it tries to have a single cache for everything? It could have a separate cache for each symbol.

ondrejmirtes (changes requested):

src/Reflection/BetterReflection/SourceLocator/FileCachedSourceLocator.php
ondrejmirtes: Feeling is a friend of premature optimization, premature optimization is evil :) We don't need this technique anywhere else so I assume we don't need it here either. So please try to do a normal save/load for each requested symbol at the right point. You can measure before/after time to see if it makes a difference.

staabm:
at the current state I cannot see any perf improvement anymore.

hopefully it gets better after https://github.com/ondrejmirtes/BetterReflection/pull/41 is merged

staabm:
I think we have the same thing with turbo, so no need for the PR anymre

PR #4931, opened 2026-02-14
Prevent duplicate path normalization

staabm:
refs https://github.com/phpstan/phpstan/issues/14072

grafik

### DIFF
--- a/src/File/FileFinder.php
+++ b/src/File/FileFinder.php
@@ -53,9 +53,7 @@ final class FileFinder
 		sort($files);
 
 		return new FileFinderResult(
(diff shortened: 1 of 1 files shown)


src/File/FileFinder.php
```diff
@@ -51,7 +52,10 @@ public function findFiles(array $paths): FileFinderResult
```
staabm, line 52: `isExcludedFromAnalysing` is already normalizing (so doesn't require a normalized $files arg), therefore lets normalize only after filtering to reduce the duplicate work

VincentLanglet (approved):

PR #4932, opened 2026-02-14
Prevent unnecessary work

staabm:
Skip obvious non-callable arrays early (without calling into `Scope->getType()`)


staabm:
path not hot enough.. not worth fiddling

PR #4938, opened 2026-02-15
feat: adds AI agent detection to auto-switch to JSON output

pushpak1300:
This pull request adds AI agent detection that automatically switches PHPStan to JSON error format when executed within **Claude Code**, **Cursor**, **Gemini CLI**, **Codex**, **OpenCode**, **Augment**, **Replit**, or **Devin**.

**Why:**

The default PHPStan output is designed for humans - colorful tables, progress bars with animations, and formatted summaries. AI agents need something different: structured JSON they can reliably parse, unambiguous error details with file paths and line numbers, minimal output that saves context window and reduces token cost, and no progress bar animations that waste tokens on redraw noise.

**How it works:**

1. When Claude Code runs PHPStan, it sets the `CLAUDECODE` / `CLAUDE_CODE` environment variable
2. When Cursor runs PHPStan, it sets `CURSOR_TRACE_ID` / `CURSOR_AGENT`
3. Other agents set their own env vars (`GEMINI_CLI`, `CODEX_SANDBOX`, `AI_AGENT`, etc.)
4. PHPStan detects these via `shipfastlabs/agent-detector` and automatically switches to JSON format + suppresses progress bar animations - no flags needed

Output:

```json
{"totals":{"errors":0,"file_errors":1},"files":{"/path/to/file.php":{"errors":1,"messages":[{"message":"Error message","line":10,"ignorable":true,"identifier":"error.id"}]}},"errors":[]}
```

Explicit `--error-format=table` or config `errorFormat: table` always takes priority over agent detection.

related https://github.com/laravel/pint/pull/415

PR #4940, opened 2026-02-15
Switch agent output from JSON to TOON for ~19% token reduction

pushpak1300:
This pull request switches the AI agent output format from JSON to [TOON](https://github.com/HelgeSverre/toon-php) (Token-Oriented Object Notation) — a compact, machine-parseable format optimized for LLM contexts.

Follows up on #4938.

**Why:**

TOON saves ~15% tokens compared to JSON while keeping the same data structure fully machine-parseable. It achieves this by declaring field names once as a header (`messages[9]{message,line,ignorable,identifier}:`) and encoding each error as a single comma-separated line — no repeated keys, braces, or brackets per entry.

**Token comparison (9 real errors, measured with `o200k_base` tokenizer):**

| Format | Characters | Tokens | vs JSON |
|--------|-----------|--------|---------|
| table | 2,667 | 379 | -4% |
| json | 1,705 | 394 | baseline |
| **toon** | **1,377** | **334** | **-15%** |

**How detection works:**

1. When Claude Code runs PHPStan, it sets the `CLAUDECODE` / `CLAUDE_CODE` environment variable
2. When Cursor runs PHPStan, it sets `CURSOR_TRACE_ID` / `CURSOR_AGENT`
3. Other agents set their own env vars (`GEMINI_CLI`, `CODEX_SANDBOX`, `AI_AGENT`, `AUGMENT_AGENT`, `OPENCODE`, `REPL_ID`, etc.)
4. PHPStan detects these via `shipfastlabs/agent-detector` and automatically switches to TOON format + suppresses progress bar animations — no flags needed

**What changed:**

1. Added `helgesverre/toon` dependency for TOON encoding
2. Created `ToonErrorFormatter` — builds the same error data structure as `JsonErrorFormatter` but encodes as TOON
3. Agent detection now switches to `toon` format instead of `json`
4. TOON format is also available directly via `--error-format=toon`

**Output:**

```
totals:
  errors: 0
  file_errors: 1
files:
  "/path/to/file.php":
    errors: 1
    messages[1]{message,line,ignorable,identifier}:
      "Error message",10,true,error.id
errors[0]:
```

Explicit `--error-format=table` or config `errorFormat: table` always takes priority over agent detection.

related https://github.com/laravel/pint/pull/415

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
 		"nette/utils": "^3.2.5",
 		"nikic/php-parser": "^5.7.0",
 		"ondram/ci-detector": "^4.0",
+		"shipfastlabs/agent-detector": "^1.0",
 		"ondrejmirtes/better-reflection": "6.65.0.9",
 		"ondrejmirtes/composer-attribute-collector": "^1.1.1",
 		"ondrejmirtes/php-merge": "^4.1",
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
 		"nette/utils": "^3.2.5",
 		"nikic/php-parser": "^5.7.0",
 		"ondram/ci-detector": "^4.0",
+		"helgesverre/toon": "^1.0",
 		"shipfastlabs/agent-detector": "^1.0",
 		"ondrejmirtes/better-reflection": "6.65.0.9",
 		"ondrejmirtes/composer-attribute-collector": "^1.1.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "497c2b3e1f40e830554ddfc28db64bf3",
+    "content-hash": "b5e59905b1e998eeaab3c3ab52d1c959",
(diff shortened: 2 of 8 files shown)


composer.json
```diff
@@ -24,6 +24,8 @@
 		"nette/utils": "^3.2.5",
 		"nikic/php-parser": "^5.7.0",
 		"ondram/ci-detector": "^4.0",
+		"helgesverre/toon": "^1.0",
+		"shipfastlabs/agent-detector": "^1.0",
```
VincentLanglet, line 28: I wonder what's your policy about added library in PHPStan @ondrejmirtes ?
Adding library with low number of star and/or low number of contributors/maintainers is always a risk (especially for security reason).
I always have [this story in mind](https://snyk.io/fr/blog/malicious-code-found-in-npm-package-event-stream/)

ondrejmirtes: Not worth it here...

ondrejmirtes:
I said you should try `raw` format instead 😊

VincentLanglet:
It could still be exposed as a standalone extension `pushpak1300/phpstan-toon-formatter` maybe ?

I'm not sure about it but I think PHPStan expose enough extension point to add custom formatter, no ?

PR #4941, opened 2026-02-15
Allow to dump more than one thing

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/6419

### DIFF
--- a/src/dumpType.php
+++ b/src/dumpType.php
@@ -9,7 +9,7 @@ namespace PHPStan;
  *
  * @throws void
  */
-function dumpType($value) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
+function dumpType(...$value) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
 {
 	return null;
 }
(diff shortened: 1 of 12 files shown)


src/dumpType.php
```diff
@@ -9,7 +9,7 @@
  *
  * @throws void
  */
-function dumpType($value) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
+function dumpType(...$value) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
```
staabm: we should make sure at least one arg is given

```suggestion
function dumpType($value, ...$values) // phpcs:ignore Squiz.Functions.GlobalFunction.Found
```

and test it errors on zero args

VincentLanglet: Added

staabm (approved):

staabm:
Thank you

ondrejmirtes (commented):
Awesome!

staabm:
Do we need to update docs for this new signature?

Or maybe the identifier article at https://phpstan.org/error-identifiers/phpstan.dumpType

PR #4943, opened 2026-02-15
Add param-out for getmxrr

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/10704

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -377,3 +377,12 @@ function register_tick_function(callable $callback, mixed ...$args): bool {}
  * @return resource|false
  */
 function proc_open($command, array $descriptor_spec, &$pipes, ?string $cwd = null, ?array $env_vars = null, ?array $options = null) {}
+
+/**
+ * @param mixed &$hosts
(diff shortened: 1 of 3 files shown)


stubs/core.stub
```diff
@@ -377,3 +377,12 @@ function register_tick_function(callable $callback, mixed ...$args): bool {}
  * @return resource|false
  */
 function proc_open($command, array $descriptor_spec, &$pipes, ?string $cwd = null, ?array $env_vars = null, ?array $options = null) {}
+
+/**
+ * @param mixed &$hosts
+ * @param mixed &$weights
+ *
+ * @param-out array<string> $hosts
```
staabm: [per php.net docs](https://www.php.net/manual/en/function.getmxrr.php), its a list

> 

    A list of the MX records found is placed into the array hosts.


```suggestion
 * @param-out list<string> $hosts
```

VincentLanglet: Done

staabm (approved):

staabm:
thank you

PR #4944, opened 2026-02-15
Fix array + array inference

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/13561

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -1499,18 +1499,45 @@ final class InitializerExprTypeResolver
 				$keyType = TypeCombinator::union(...$keyTypes);
 			}
 
+			$leftIterableValueType = $leftType->getIterableValueType();
 			$arrayType = new ArrayType(
 				$keyType,
-				TypeCombinator::union($leftType->getIterableValueType(), $rightType->getIterableValueType()),
+				TypeCombinator::union($leftIterableValueType, $rightType->getIterableValueType()),
 			);
 
 			$accessories = [];
-			foreach ($leftType->getConstantArrays() as $type) {
-				foreach ($type->getKeyTypes() as $i => $offsetType) {
-					if ($type->isOptionalKey($i)) {
+			if ($leftCount > 0) {
+				// Use the first constant array as a reference to list potential offsets.
+				// We only need to check the first array because we're looking for offsets that exist in ALL arrays.
+				$constantArray = $leftConstantArrays[0];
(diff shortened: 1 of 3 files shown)


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1499,18 +1499,37 @@ public function getPlusTypeFromTypes(Expr $left, Expr $right, Type $leftType, Ty
...
 			);
 
 			$accessories = [];
-			foreach ($leftType->getConstantArrays() as $type) {
-				foreach ($type->getKeyTypes() as $i => $offsetType) {
-					if ($type->isOptionalKey($i)) {
+			if ($leftCount > 0) {
+				// Use the first constant array as a reference to list potential offsets.
+				// We only need to check the first array because we're looking for offsets that exist in ALL arrays.
+				$constantArray = $leftConstantArrays[0];
```
staabm, line 1512: I am not sure why this only works on index `0`.

[in `array_merge` we iterate over all left arrays](https://github.com/phpstan/phpstan-src/pull/4554/files)

VincentLanglet: This is something you could try for array_merge too I think.

Since we're looking only for certainty of `hasOffsetValueType` YES the key has to be in all the constant arrays, so checking only one constant array is enough.
- only keys in the first constant array can have YES certainty
- key which are not in the first contant array will at most have the MAYBE certainty

staabm: do you think it would make sense to adjust `array_merge()` and `array_replace()` with this PR, so we have it in sync over all implementations?

VincentLanglet: Honestly I'm not sure about it.

Those implementations seems more complicated, especially because there is an iteration
```
foreach ($argTypes as $argType) {
```

Also the logic seems to be different since there is some looking at code like
```
TrinaryLogic::createFromBoolean($argType->hasOffsetValueType($keyType)->yes());
```
or
```
if ($hasOffsetValue->yes()) {
					// the last string-keyed offset will overwrite previous values
					$hasOffsetType = new HasOffsetValueType(
						$keyType,
						$offsetType,
					);
				} elseif ($hasOffsetValue->maybe()) {
					$hasOffsetType = new HasOffsetType(
						$keyType,
					);
				} else {
					continue;
				}
```
which I don't have here.

So I would prefer avoid to delay with PR (with the risk of introducing a regression) and fixing the existing bug, and we could try a dedicated PR with the refacto for array_merge/replace ?

staabm (approved):
thank you

PR #4945, opened 2026-02-15
Differentiate signaturePrototype and inheritancePrototype

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/11067

Closes https://github.com/phpstan/phpstan/issues/12830

Closes https://github.com/phpstan/phpstan/issues/12272

### DIFF
--- a/src/Rules/Methods/MethodPrototypeFinder.php
+++ b/src/Rules/Methods/MethodPrototypeFinder.php
@@ -29,6 +29,7 @@ final class MethodPrototypeFinder
 	 * Returns two prototypes with different purposes:
 	 * - Signature prototype: Used for validating method signature (parameters, return type, ...).
 	 * - Inheritance prototype: Used for validating inheritance rules (final keyword, override attribute, ...).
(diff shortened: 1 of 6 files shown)


src/Rules/Methods/MethodPrototypeFinder.php
```diff
@@ -25,14 +25,19 @@ public function __construct(
 	}
 
 	/**
-	 * @return array{ExtendedMethodReflection, ClassReflection, bool}|null
+	 * Finds the prototype method that a class method should be validated against.
```
staabm, line 28: since we started describing the return type, I think it would be usefull to also describe what the bool means

staabm (commented):
I can't really judge the fix, but fixing 3 bugs at once seems to prove it looks good :)

src/Rules/Methods/MethodPrototypeFinder.php
VincentLanglet: I added one line about it

staabm (approved):

ondrejmirtes:
Thank you!

PR #4946, opened 2026-02-16
Add workflow for Claude to react to PR comments and reviews

ondrejmirtes:
Automatically triggers Claude Code when comments or reviews are left on
PRs created by phpstan-bot, without requiring @claude mentions.

- Handles issue comments, review submissions, and inline review comments
- Filters to only phpstan-bot PRs, skips bot's own comments
- Adds 👀 reaction to acknowledge the comment
- Collects full PR comment history and inline review context for Claude
- Limits to 5 reaction commits per PR to prevent runaway loops
- Skips closed/merged PRs
- Uses concurrency groups to avoid conflicting runs

https://claude.ai/code/session_01UfUUMUBGgCttD9VMTf6FvJ

### DIFF
--- a/.github/workflows/claude-pr-reactions.yml
+++ b/.github/workflows/claude-pr-reactions.yml
@@ -21,18 +21,6 @@ concurrency:
 jobs:
   react:
     name: "React to PR feedback"
-    if: >-
-      (
-        github.event_name == 'issue_comment'
-        && github.event.issue.pull_request
-        && github.event.issue.user.login == 'phpstan-bot'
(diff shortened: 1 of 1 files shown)


staabm:
do we need to make sure the phpstan-bot (claude) can only be triggered by phpstan-src maintainers (or at least a hardcoded list of people) to prevent misuse?

ondrejmirtes:
@staabm My theory is that it only works for people with write access but you can try it maybe from your other account 😊

PR #4952, opened 2026-02-16
Support integer range keys in constant arrays

staabm:
## Summary
When assigning to a constant array with a union of constant string keys (e.g., `'a'|'b'`) or a finite integer range (e.g., `int<1,5>`), PHPStan previously degraded to a general `ArrayType`, losing the constant array shape information. This change produces a union of constant arrays instead, preserving precise type information.

For example, given `@param array{foo: int} $a` and `@param int<1,5> $intRange`:
```php
$a[$intRange] = 256;
// Before: non-empty-array<'foo'|int<1, 5>, int>
// After:  array{foo: int, 1: 256}|array{foo: int, 2: 256}|array{foo: int, 3: 256}|array{foo: int, 4: 256}|array{foo: int, 5: 256}
```

## Changes
- `src/Type/Constant/ConstantArrayType.php`:
  - Modified `setOffsetValueType()` to detect when the offset type can be expanded to a finite set of constant keys with at least one new key, and produce a union of constant arrays
  - Added `resolveFiniteScalarKeyTypes()` private helper method that resolves offset types to individual constant key types from string unions and integer ranges
  - Limited expansion to `CHUNK_FINITE_TYPES_LIMIT` (5) new keys to avoid combinatorial explosion
  - Only expands string constant unions and `IntegerRangeType` (not integer constant unions like `0|1` which typically come from loop fixpoint analysis)
- `tests/PHPStan/Analyser/nsrt/set-constant-union-offset-on-constant-array.php`: New regression test covering string union keys, integer range keys, infinite ranges, and existing keys
- `tests/PHPStan/Analyser/nsrt/constant-array-type-set.php`: Updated test expectation for `int<0, 4>` offset on 3-element array to reflect more precise union result
- `CLAUDE.md`: Added documentation about the new union expansion behavior

## Root cause
`ConstantArrayTypeBuilder::setOffsetValueType()` handles non-constant offset types by expanding them to constant scalars and checking if all match existing keys. When some keys were new (not in the array), it fell through to a degradation path that produced a general `ArrayType`. The fix intercepts this at the `ConstantArrayType::setOffsetValueType()` level, creating a union of arrays for each possible key before the builder can degrade.

The expansion is carefully limited to avoid regressions:
- Only string constant unions and `IntegerRangeType` trigger expansion (not integer constant unions which are common in loop analysis)
- At most `CHUNK_FINITE_TYPES_LIMIT` (5) new keys are expanded
- When all keys already exist in the array, the builder handles it directly without expansion

## Test
Added `tests/PHPStan/Analyser/nsrt/set-constant-union-offset-on-constant-array.php` with four test cases:
1. `doFoo`: String union offset `'a'|'b'` on `array{foo: int}` → union of two constant arrays
2. `doBar`: Integer range `int<1,5>` on `array{foo: int}` → union of five constant arrays
3. `doInfiniteRange`: Infinite range `int<0, max>` → falls back to general array (no expansion)
4. `doExistingKeys`: Range `int<0,1>` on `array{0: 'a', 1: 'b'}` → handled by builder directly

Fixes phpstan/phpstan#14129

Closes https://github.com/phpstan/phpstan/issues/9907


src/Type/Constant/ConstantArrayType.php
```diff
@@ -699,6 +700,42 @@ public function getOffsetValueType(Type $offsetType): Type
...
+					$arrayTypes = [];
+					foreach ($scalarKeyTypes as $scalarKeyType) {
+						$builder = ConstantArrayTypeBuilder::createFromConstantArray($this);
+						$builder->setOffsetValueType($scalarKeyType, $valueType);
+						$arrayTypes[] = $builder->getArray();
+					}
+
+					return TypeCombinator::union(...$arrayTypes);
+				}
+			}
```
staabm: in a perfect world, I think this logic should be part of `ConstantArrayTypeBuilder->setOffsetValueType`

src/Type/Constant/ConstantArrayType.php
```diff
@@ -699,6 +700,57 @@ public function getOffsetValueType(Type $offsetType): Type
 
 	public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $unionValues = true): Type
 	{
+		if ($offsetType !== null) {
+			$scalarKeyTypes = $offsetType->toArrayKey()->getConstantStrings();
+			if (count($scalarKeyTypes) === 0) {
+				$integerRanges = TypeUtils::getIntegerRanges($offsetType);
```
VincentLanglet, line 706: You're calling getIntegerRanges on offsetType but previously you worked with
```
$offsetType->toArrayKey()
```

Shouldn't you work every where with `$offsetType->toArrayKey()` ?

that could have impact for things like `int<0, 2>|'3'` maybe ?

src/Type/Constant/ConstantArrayType.php
```diff
@@ -699,6 +700,57 @@ public function getOffsetValueType(Type $offsetType): Type
...
+
+						foreach ($finiteTypes as $finiteType) {
+							$scalarKeyTypes[] = $finiteType;
+						}
+					}
+				}
+			}
+
+			// turn into tagged union for more precise results
+			if (
```
VincentLanglet, line 722: It's unclear to me why
- We're working only with constantStrings (getConstantStrings)
- We're adding int range only when there is no string

What about `@param 1|2 $key` ?
And what about `@param '1'|2 $key` ?

Also the `$existingKeyType->getValue() === $scalarKeyType->getValue()` check might miss some `1 === '1'`.

src/Type/Constant/ConstantArrayType.php
staabm: `TypeUtils::getIntegerRanges` only returns integer ranges. as soon as something else is contained it returns a empty-array.

there is not type which can be converted into a integer range using `toArrayKey`.

the logic here pretty much resembles whats [already in `ConstantArrayTypeBuilder`](https://github.com/phpstan/phpstan-src/blob/0bfbaf9e1c2fb8d41a4dfca848a3788fa807493f/src/Type/Constant/ConstantArrayTypeBuilder.php#L243-L280).

staabm: in case I use `getConstantScalarTypes` instead of `getConstantStrings` I get failling tests and I don't know yet why.

staabm:
@phpstan-bot refactor the changes in ConstantArrayType into ConstantArrayTypeBuilder

staabm:
@phpstan-bot refactor the changes in ConstantArrayType into ConstantArrayTypeBuilder

staabm:
I will close now, as ondrej will likely rework array semantics as a whole and we can have a look at issue again, after that work landed

PR #4954, opened 2026-02-16
Add sodium_crypto_secretbox_keygen function stub

thePanz:
According to the docs, the function returns a non-empty string: https://www.php.net/manual/en/function.sodium-crypto-secretbox-keygen.php

### DIFF
--- a/stubs/core.stub
+++ b/stubs/core.stub
@@ -53,6 +53,11 @@ function sodium_bin2base64(string $string, int $id): string {}
 function sodium_bin2hex(string $string): string {}
 
 /**
+ * @return non-empty-string
+ */
+function sodium_crypto_secretbox_keygen(): string {}
+
+/**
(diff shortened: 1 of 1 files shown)


staabm (approved):
Looks correct

https://github.com/php/php-src/blob/99bf26be0b81a08f2fef4f96f8983075c81eacae/ext/sodium/libsodium.c#L3133

VincentLanglet (commented):
Can it return `'0'` or is it non-falsy ?

staabm:
most of this 'encoding' functions just use non-empty-string atm. I think thats good enough.

VincentLanglet (approved):

staabm:
Thank you @thePanz

thePanz:
Thank you @staabm @VincentLanglet 
I will try to find time and add another PR for the other `sodium_ ... _keygen()` functions

PR #4955, opened 2026-02-16
Cleanup ConstantArrayTypeBuilder

staabm:


VincentLanglet (approved):
I feel like for offset like `1|true` it will avoid doing the foreach two times.

PR #4957, opened 2026-02-16
Fix BcMath\Number increment/decrement

mpesari:
Partial fix for https://github.com/phpstan/phpstan/issues/13965

This adds support for BcMath\Number increment/decrement operations on PHP >= 8.4.

```
php > $x = new BcMath\Number('3.14');
php > print $x;
3.14
php > $x++;
php > print $x;
4.14
```

### DIFF
--- a/tests/PHPStan/Rules/Operators/InvalidIncDecOperationRuleTest.php
+++ b/tests/PHPStan/Rules/Operators/InvalidIncDecOperationRuleTest.php
@@ -234,4 +234,10 @@ class InvalidIncDecOperationRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/inc-non-numeric-string.php'], $errors);
 	}
 
+	#[RequiresPhp('>= 8.4')]
+	public function testBcMathNumber(): void
+	{
+		$this->analyse([__DIR__ . '/data/inc-dec-bcmath-number.php'], []);
+	}
+
 }
--- a/tests/PHPStan/Rules/Operators/data/inc-dec-bcmath-number.php
+++ b/tests/PHPStan/Rules/Operators/data/inc-dec-bcmath-number.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace IncDecBcMathNumber;
+
+use BcMath\Number;
+
+function testPreInc(Number $x): void {
+	++$x;
+}
+
+function testPostInc(Number $x): void {
+	$x++;
+}
+
+function testPreDec(Number $x): void {
+	--$x;
+}
+
+function testPostDec(Number $x): void {
+	$x--;
+}
--- a/src/Rules/Operators/InvalidIncDecOperationRule.php
+++ b/src/Rules/Operators/InvalidIncDecOperationRule.php
@@ -132,6 +132,10 @@ final class InvalidIncDecOperationRule implements Rule
 			$deprecatedBool = true;
 		}
 
+		if ($this->phpVersion->supportsBcMathNumberOperatorOverloading()) {
+			$allowedTypes[] = new ObjectType('BcMath\Number');
+		}
+
 		$allowedTypes = new UnionType($allowedTypes);
 
 		$varType = $this->ruleLevelHelper->findTypeToCheck(
--- a/tests/PHPStan/Analyser/nsrt/bcmath-number.php
+++ b/tests/PHPStan/Analyser/nsrt/bcmath-number.php
@@ -405,4 +405,12 @@ class Foo
 			assertType('bool', $a or $b);
 		}
 	}
+
+	public function bcIncDec(Number $a): void
+	{
+		assertType('BcMath\Number', ++$a);
+		assertType('BcMath\Number', $a++);
+		assertType('BcMath\Number', --$a);
+		assertType('BcMath\Number', $a--);
+	}
 }


staabm (approved):

VincentLanglet (changes requested):
You need to target 2.1.x

mpesari:
Rebased and targeted to `2.1.x`

mpesari:
Hmm.. I wanted to make sure that we report errors on PHP < 8.4 when the polyfill is installed, but looks like the test doesn't report any errors.

VincentLanglet (approved):

tests/PHPStan/Rules/Operators/data/inc-dec-bcmath-number.php
```diff
@@ -0,0 +1,13 @@
...
+
+use BcMath\Number;
+
+function test(Number $x): void
+{
+	++$x;
+	$x++;
+	$x--;
+	--$x;
+}
```
VincentLanglet: The first line `++$x` transform into an Error then the other doesn't fail.

I recommend you to split the test
```
function test1(Number $x): void
{
	++$x;
}

function test2(Number $x): void
{
	$x++;
}
```
and so on for a better understanding.

VincentLanglet (changes requested):

tests/PHPStan/Rules/Operators/data/inc-dec-bcmath-number.php
mpesari: Thanks, it looks more reasonable now.

I also removed the PHP version check in the test. Not sure if I should add `#[RequiresPhp('>= 8.4')]` there?

I could also add a first commit that adds initial tests demonstrating the problems, and which the second commit then fixes. I like it that way, but seems like it's not commonly used in this repo.

VincentLanglet: I get 
```
Parameter $x of function IncDecNonNumericString\testPreInc() has invalid type BcMath\Number.
```
https://phpstan.org/r/d9804c7a-e0b4-4a60-89a5-889f733dbbb8

Since the class does not exist before 8.4 it make sens to not have invalid increment/decrement error.

```
#[RequiresPhp('>= 8.4')]
```
makes sens.

mpesari: Added

VincentLanglet (approved):

VincentLanglet:
Please take a new look @staabm with the changes

staabm:
I like the PR so far.

per https://www.php.net/manual/en/class.bcmath-number.php

> A class for an arbitrary precision number. These objects support overloaded [arithmetic](https://www.php.net/manual/en/language.operators.arithmetic.php) and [comparison](https://www.php.net/manual/en/language.operators.comparison.php) operators.

this class also has special comparison semantics. I think it would make sense to cover both aspects in this PR.
@mpesari could you look into comparisons?

VincentLanglet:
> I like the PR so far.
> 
> per [php.net/manual/en/class.bcmath-number.php](https://www.php.net/manual/en/class.bcmath-number.php)
> 
> > A class for an arbitrary precision number. These objects support overloaded [arithmetic](https://www.php.net/manual/en/language.operators.arithmetic.php) and [comparison](https://www.php.net/manual/en/language.operators.comparison.php) operators.
> 
> this class also has special comparison semantics. I think it would make sense to cover both aspects in this PR. @mpesari could you look into comparisons?

Isn't already handled by https://github.com/phpstan/phpstan-src/blob/c2da2e89e611749d0d455133c1ec482ab936efc6/src/Type/Php/BcMathNumberOperatorTypeSpecifyingExtension.php#L17 ?

staabm:
you are right. its covered in tests/PHPStan/Analyser/nsrt/bcmath-number.php

staabm:
thank you @mpesari

tests/PHPStan/Rules/Operators/InvalidIncDecOperationRuleTest.php
```diff
@@ -234,4 +234,10 @@ public function testIncNonNumericString(): void
 		$this->analyse([__DIR__ . '/data/inc-non-numeric-string.php'], $errors);
 	}
 
+	#[RequiresPhp('>= 8.4')]
+	public function testBcMathNumber(): void
+	{
+		$this->analyse([__DIR__ . '/data/inc-dec-bcmath-number.php'], []);
```
ondrejmirtes, line 240: Would be nice to run this before 8.4 too, and set the expected errors.

VincentLanglet: According to the author, the test was reporting no error for version before 8.4 (cf https://github.com/phpstan/phpstan-src/pull/4957#issuecomment-3909788759) 

Looking at the playground https://phpstan.org/r/bf941228-962c-42af-a859-b909918634a7, I assume that the type is "ERROR" and no error is reported when doing increment/decrement on ErrorType.

So I thought this wasn't a relevant test to add

ondrejmirtes: Oh it's a new class. I get it, thanks.

mpesari: BTW in that playground enabling strict-rules shows more errors. I'll take a look tomorrow.

PR #4980, opened 2026-02-17
Fix BcMath\Number unary plus/minus/bitwise-not

mpesari:
Fixes https://github.com/phpstan/phpstan/issues/13965

Returning `ObjectType('BcMath\Number')` from `toNumber()` fixed `+$x` and `-$x` but caused other tests to fail in these operations (https://github.com/phpstan/phpstan-src/blob/2.1.x/tests/PHPStan/Analyser/nsrt/bcmath-number.php)

* `$a & $b`
* `$a | $b`
* `$a ^ $b`
* `$a << $b`
* `$a >> $b`
 
This is fixed by checking for `BcMath\Number` early and returning `ErrorType` before calling `toNumber()` when handling those bitwise operations (`getBitwiseAndTypeFromTypes()` etc.)

I'm not sure if the `*ERROR*` -> `*NEVER*` changes in tests fix something or break something?

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -993,6 +993,10 @@ final class InitializerExprTypeResolver
 			return $this->getNeverType($leftType, $rightType);
 		}
 
+		if ((new ObjectType('BcMath\Number'))->isSuperTypeOf($leftType)->yes()) {
+			return new ErrorType();
+		}
+
 		$result = $this->getFiniteOrConstantScalarTypes($leftType, $rightType, static fn ($a, $b) => $a & $b);
 		if ($result instanceof Type) {
 			return $result;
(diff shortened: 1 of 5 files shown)


VincentLanglet:
I'm not sure it's the right way to do it see
https://github.com/phpstan/phpstan/issues/13965#issuecomment-3914508563

Given the fact we have a `BcMathNumberOperatorTypeSpecifyingExtension` I don't see why we would have the unary operator hardcoded.

So maybe the best way to do it the same way is to
- introduce UnaryOperatorTypeSpecifyingExtension 
- implements it for BCMathNumber
- Introduce UnaryOperatorTypeSpecifyingExtensionRegistry and provider (or add a param to the construct of OperatorTypeSpecifyingExtensionRegistry)
- Use it in the InitializerExprTypeResolver the same way it's already used.

But I'd like the confirmation from someone else

staabm:
I did not verify it, but I think we have support for [GMP which also supports working with objects on operators](https://wiki.php.net/rfc/operator_overloading_gmp).

please see whether this can be used as inspiration how BcMath\Number could/should work

staabm:
see e.g. https://github.com/phpstan/phpstan-src/blob/e696ed0dd045e0d0d25b365ae8ef401b46272f6f/src/Type/OperatorTypeSpecifyingExtension.php#L6-L10

VincentLanglet:
You should be able to rework your PR with UnaryOperatorTypeSpecifyingExtension  https://github.com/phpstan/phpstan-src/pull/5284

VincentLanglet:
Closing since https://github.com/phpstan/phpstan/issues/13965 is solved

PR #4982, opened 2026-02-17
Filter unsupported OpenSSL cipher methods on PHP 8.0-8.4

VincentLanglet:
## Summary

On PHP 8.0-8.4, `openssl_get_cipher_methods()` reports cipher algorithms that are not actually supported by OpenSSL functions like `openssl_cipher_iv_length()` and `openssl_cipher_key_length()`. This is due to a PHP bug (https://github.com/php/php-src/issues/19994) where `openssl_get_cipher_methods()` uses a different source of algorithms than the actual cipher functions, causing it to list algorithms that OpenSSL 3.0 has disabled by default.

This caused PHPStan to incorrectly refine the return type of these functions — for example, `openssl_cipher_iv_length('aes-128-cbc-cts')` was narrowed to `int` when it actually returns `false` on PHP 8.0-8.4.

## Changes

- Added `src/Type/Php/OpenSslCipherMethodsProvider.php` — a shared service that filters the cipher methods list by actually testing each algorithm with `openssl_cipher_iv_length()`, excluding those that return `false` (i.e. are not truly supported)
- Updated `src/Type/Php/OpensslCipherFunctionsReturnTypeExtension.php` to use the new shared provider instead of directly querying `openssl_get_cipher_methods()`
- Updated `src/Type/Php/OpenSslEncryptParameterOutTypeExtension.php` to use the same shared provider for consistent behavior

## Root cause

`openssl_get_cipher_methods()` on PHP 8.0-8.4 uses `EVP_CIPHER_do_all_sorted()` to enumerate algorithms, while functions like `openssl_cipher_iv_length()` use `EVP_get_cipherbyname()` to resolve them. With OpenSSL 3.0+, the former returns algorithms from all providers (including disabled ones), while the latter only returns algorithms from loaded providers. This discrepancy was fixed in PHP 8.5 via a refactoring of the OpenSSL implementation.

The fix filters the algorithm list at PHPStan's analysis time by calling `openssl_cipher_iv_length()` for each algorithm and removing those that fail, ensuring only truly supported algorithms are used for type refinement.

## Test

Added `tests/PHPStan/Analyser/nsrt/bug-13692.php` — verifies that `openssl_cipher_iv_length('aes-128-cbc-cts')` is correctly refined to `false` (not `int`) on PHP 8.4, while `aes-128-cbc` still correctly refines to `int`.

Fixes phpstan/phpstan#13692


tests/PHPStan/Analyser/NodeScopeResolverTest.php
```diff
@@ -39,6 +39,10 @@ private static function findTestFiles(): iterable
 			yield __DIR__ . '/data/enum-reflection-backed.php';
 		}
 
+		if (PHP_VERSION_ID >= 80000 && PHP_VERSION_ID < 80500) {
+			yield __DIR__ . '/data/bug-13692.php';
+		}
```
staabm, line 44: do we need separate assertions for this test for other php versions?

staabm (approved):

staabm (approved):

PR #4983, opened 2026-02-17
Fix isset() falsey branch not narrowing array type for dim fetches

VincentLanglet:
## Summary
When using `isset($a['key'])` on a union type array, the falsey branch did not narrow the array variable's type. This caused the original union members to leak through after the if-block, even though they were handled by the truthy branch.

For example:
```php
$a = [];
if (rand() % 2) {
    $a = ['bar' => 'string'];
}
// $a: array{}|array{bar: 'string'}

if (isset($a['bar'])) {
    $a['bar'] = 1;
}
// Expected: array{}|array{bar: 1}
// Actual:   array{}|array{bar: 'string'}|array{bar: 1}
```

The `array{bar: 'string'}` type leaked through because the falsey branch of `isset` for `ArrayDimFetch` expressions returned empty `SpecifiedTypes()`, leaving the array variable's type unchanged. When merged with the truthy branch result, the old type persisted.

## Changes
- `src/Analyser/TypeSpecifier.php`: In the falsey `isset` handler for non-variable expressions, added narrowing logic for `ArrayDimFetch` with constant string/integer dims. The logic iterates over union members of the array variable and removes those where the offset is definitely present (`hasOffsetValueType().yes()`) and the value is definitely non-null (`getOffsetValueType().isNull().no()`). When the array variable has uncertain existence (`maybe` certainty), an `IssetExpr` specification is included to preserve the `maybe` certainty.
- `tests/PHPStan/Analyser/nsrt/bug-9908.php`: New regression test reproducing the original issue.
- `tests/PHPStan/Analyser/nsrt/tagged-unions.php`: Updated test expectation at line 58 to reflect the improved narrowing precision — `!isset($foo['C'])` now correctly excludes the `array{A: string, C: 1}` member that definitely has key `C` with a non-null value.
- `phpstan-baseline.neon`: Updated `instanceof ConstantStringType` count for `TypeSpecifier.php` (4→5).

## Root cause
The `TypeSpecifier::specifyTypesInCondition` method's falsey `isset` handler had a code path for non-variable `ArrayDimFetch` expressions (line ~923) that returned empty `SpecifiedTypes()` when the offset value was not nullable and the isset check result was uncertain. This meant the array variable received no type narrowing in the falsey branch, causing its original type (including members where isset would be true) to persist and leak through during scope merging after the if-block.

The fix carefully avoids two pitfalls:
1. Using `HasOffsetType` removal directly (via `TypeCombinator::remove`) would be too aggressive for optional keys, where `tryRemove` unsets the offset entirely.
2. Any type narrowing through `SpecifiedTypes` sets variable certainty to `yes`, which would be wrong when the variable might not be defined. The fix uses `IssetExpr` to restore `maybe` certainty when needed.

## Test
The regression test `tests/PHPStan/Analyser/nsrt/bug-9908.php` includes:
- `test()`: Asserts that after `isset($a['bar'])` + `$a['bar'] = 1`, the type is `array{}|array{bar: 1}` (not including the original `array{bar: 'string'}`)
- `test2()`: Verifies the same code passes when used as an argument to a method expecting `array{bar?: int}`

Fixes phpstan/phpstan#9908

Closes phpstan/phpstan#10544
Closes phpstan/phpstan#8724
Closes phpstan/phpstan#5128
Closes phpstan/phpstan#12401

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -929,19 +929,20 @@ final class TypeSpecifier
 						$dimType = $scope->getType($issetExpr->dim);
 
 						if ($dimType instanceof ConstantIntegerType || $dimType instanceof ConstantStringType) {
-							$types = $varType instanceof UnionType ? $varType->getTypes() : [$varType];
+							$constantArrays = $varType->getConstantArrays();
 							$typesToRemove = [];
-							foreach ($types as $innerType) {
-								$hasOffset = $innerType->hasOffsetValueType($dimType);
-								if (!$hasOffset->yes() || !$innerType->getOffsetValueType($dimType)->isNull()->no()) {
+							foreach ($constantArrays as $constantArray) {
+								$hasOffset = $constantArray->hasOffsetValueType($dimType);
+								if (!$hasOffset->yes() || !$constantArray->getOffsetValueType($dimType)->isNull()->no()) {
 									continue;
 								}
 
-								$typesToRemove[] = $innerType;
+								$typesToRemove[] = $constantArray;
 							}
 
 							if ($typesToRemove !== []) {
 								$typeToRemove = TypeCombinator::union(...$typesToRemove);
+
 								$result = $this->create(
 									$issetExpr->var,
 									$typeToRemove,
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -920,6 +920,52 @@ final class TypeSpecifier
 					return $exprType;
 				}
 
+				if (
+					$issetExpr instanceof ArrayDimFetch
+					&& $issetExpr->dim !== null
+				) {
+					$varType = $scope->getType($issetExpr->var);
+					if (!$varType instanceof MixedType) {
+						$dimType = $scope->getType($issetExpr->dim);
+
+						if ($dimType instanceof ConstantIntegerType || $dimType instanceof ConstantStringType) {
+							$types = $varType instanceof UnionType ? $varType->getTypes() : [$varType];
+							$typesToRemove = [];
+							foreach ($types as $innerType) {
+								$hasOffset = $innerType->hasOffsetValueType($dimType);
+								if (!$hasOffset->yes() || !$innerType->getOffsetValueType($dimType)->isNull()->no()) {
+									continue;
+								}
+
+								$typesToRemove[] = $innerType;
+							}
+
+							if ($typesToRemove !== []) {
+								$typeToRemove = TypeCombinator::union(...$typesToRemove);
+								$result = $this->create(
+									$issetExpr->var,
+									$typeToRemove,
+									TypeSpecifierContext::createFalse(),
+									$scope,
+								)->setRootExpr($expr);
+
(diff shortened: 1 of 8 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -920,6 +920,52 @@ public function specifyTypesInCondition(
...
+				if (
+					$issetExpr instanceof ArrayDimFetch
+					&& $issetExpr->dim !== null
+				) {
+					$varType = $scope->getType($issetExpr->var);
+					if (!$varType instanceof MixedType) {
+						$dimType = $scope->getType($issetExpr->dim);
+
+						if ($dimType instanceof ConstantIntegerType || $dimType instanceof ConstantStringType) {
+							$types = $varType instanceof UnionType ? $varType->getTypes() : [$varType];
```
VincentLanglet: TODO: I'm not sure it's useful to do this if it's not an unionType.

src/Analyser/TypeSpecifier.php
```diff
@@ -920,6 +920,52 @@ public function specifyTypesInCondition(
...
+							}
+
+							if ($typesToRemove !== []) {
+								$typeToRemove = TypeCombinator::union(...$typesToRemove);
+								$result = $this->create(
+									$issetExpr->var,
+									$typeToRemove,
+									TypeSpecifierContext::createFalse(),
+									$scope,
+								)->setRootExpr($expr);
```
VincentLanglet: This need to be reworked cf the test
```
// Could be "array{a: null, c: string}|array{d: string}"
assertType("array{a: string|null, c: string}|array{a?: string, d: string}", $a);
```

it would be great to refining the constant array without totally removing them
```
array{a: string|null, c: string} => array{a: null, c: string}
array{a?: string, d: string} => array{c: string}
```
Do you see an easy way to do such thing @staabm ?

staabm: would it work to iterate over `getConstantArrays()` instead?

and re-union the results?

VincentLanglet: I tried lot of different thing without success.
I'm not sure it's easy to remove a constant array `{a: string, b: string}` from another `{a?: string, b: string}`.

I used `getConstantArrays` instead of UnionType, and I think we could merge it like that as a first step to close 5 issues.

src/Analyser/TypeSpecifier.php
```diff
@@ -920,6 +920,53 @@ public function specifyTypesInCondition(
 					return $exprType;
 				}
 
+				if (
```
staabm, line 923: looking at 

https://github.com/phpstan/phpstan-src/blob/a10a9a264cc1d2abe25adb82adc013a5f45b4eae/src/Analyser/TypeSpecifier.php#L961-L970

I wonder whether we could implement this directly in the type-system. maybe in `ConstantArrayType->tryRemove()` or similar

staabm (approved):
I think its fine.

src/Analyser/TypeSpecifier.php
staabm: scratch that.

PR #4984, opened 2026-02-18
Fix typo

vrana:


staabm:
it seems we don't have a single test with this error message. would you be willing to cover this change with a test?

vrana:
Nah.

staabm:
thanks

PR #4985, opened 2026-02-18
Fix false positive dead catch for ReflectionMethod::invoke/invokeArgs

staabm:
## Summary

PHPStan incorrectly reported "Dead catch - RuntimeException is never thrown in the try block" when catching exceptions around `ReflectionMethod::invoke()` and `ReflectionMethod::invokeArgs()` calls. These methods execute arbitrary user code via reflection and can throw any exception, not just `ReflectionException`.

## Changes

- Added `src/Type/Php/ReflectionMethodInvokeMethodThrowTypeExtension.php` — a `DynamicMethodThrowTypeExtension` that declares `ReflectionMethod::invoke()`, `ReflectionMethod::invokeArgs()`, `ReflectionFunction::invoke()`, and `ReflectionFunction::invokeArgs()` can throw `Throwable`
- Added regression test `tests/PHPStan/Rules/Exceptions/data/bug-7719.php` covering all four methods
- Added test method `testBug7719()` to `CatchWithUnthrownExceptionRuleTest.php`
- Updated `phpstan-baseline.neon` with one new entry for the extension file

## Root cause

The BetterReflection library's adapter class (`vendor/ondrejmirtes/better-reflection/src/Reflection/Adapter/ReflectionMethod.php`) has `@throws ReflectionException` annotations on `invoke()` and `invokeArgs()`. PHPStan reads these throw type annotations and concludes that only `ReflectionException` can be thrown by these methods. However, in real PHP, `ReflectionMethod::invoke()` and `invokeArgs()` execute arbitrary user code that can throw any exception. The fix adds a dynamic throw type extension that overrides this to return `Throwable`, correctly indicating that any exception type is possible.

## Test

The regression test in `tests/PHPStan/Rules/Exceptions/data/bug-7719.php` creates four scenarios:
1. `ReflectionMethod::invokeArgs()` with a dynamic method name — catch RuntimeException should not be reported as dead
2. `ReflectionMethod::invoke()` with a dynamic method name — same
3. `ReflectionFunction::invokeArgs()` with a dynamic function name — same
4. `ReflectionFunction::invoke()` with a dynamic function name — same

All four expect no errors (empty expected errors array).

Fixes phpstan/phpstan#7719

Closes https://github.com/phpstan/phpstan/issues/9267


src/Analyser/NodeScopeResolver.php
```diff
@@ -4765,6 +4767,14 @@ private function getFunctionThrowPoint(
 
 	private function getMethodThrowPoint(MethodReflection $methodReflection, ParametersAcceptor $parametersAcceptor, MethodCall $methodCall, MutatingScope $scope): ?InternalThrowPoint
 	{
+		if (
+			$this->implicitThrows
```
VincentLanglet: There are multiple implicit throwPoint which are not behind this config option

cf 
```
if (!(new ObjectType(Traversable::class))->isSuperTypeOf($scope->getType($stmt->expr))->no()) {
				$throwPoints[] = InternalThrowPoint::createImplicit($scope, $stmt->expr);
			}
```
or
```
if ($this->phpVersion->supportsPropertyHooks()) {
					$throwPoints[] = InternalThrowPoint::createImplicit($scope, $expr);
				}
```
or
```
elseif ($expr instanceof Expr\Include_) {
			$result = $this->processExprNode($stmt, $expr->expr, $scope, $storage, $nodeCallback, $context->enterDeep());
			$throwPoints = $result->getThrowPoints();
			$throwPoints[] = InternalThrowPoint::createImplicit($scope, $expr);
```
same for Eval_ and YieldFrom

So I wonder wheither this implicitThrowPoint should be behind the flag or not. No strong opinion about it...

src/Analyser/NodeScopeResolver.php
```diff
@@ -4765,6 +4767,14 @@ private function getFunctionThrowPoint(
 
 	private function getMethodThrowPoint(MethodReflection $methodReflection, ParametersAcceptor $parametersAcceptor, MethodCall $methodCall, MutatingScope $scope): ?InternalThrowPoint
 	{
+		if (
```
VincentLanglet: This condition should occur later in the function, in case someone implement a `getDynamicMethodThrowTypeExtensions` for those method/function.

VincentLanglet (commented):
Maybe it's better to have ondrej opinion about something hardcoded like this ? WDYT ?

VincentLanglet (approved):

staabm:
> Maybe it's better to have ondrej opinion about something hardcoded like this ? WDYT ?

since there is currently no way to differentiate implicity throw points via extension I think we can hardcode it (similar to how we hardcode things into TypeSpecifier which are not achievable via extenions).

VincentLanglet:
> since there is currently no way to differentiate implicity throw points via extension I think we can hardcode it (similar to how we hardcode things into TypeSpecifier which are not achievable via extenions).

The other solution would be a dynamicThrowTypeExtention with explicitPoint which resolve the method called (so ReflectionMethod/Function needs to be generic of the name method/function) AND the args, in order to know whether it will throw an exception or not ; and which one.

But that's way more work.

PR #4986, opened 2026-02-18
Fix test lint-comment

staabm:
fix existing tests and claude instructions.

added a exception when the first line contains the `lint`-substring but at the wrong position

### DIFF
--- a/.github/workflows/tests-levels-matrix.php
+++ b/.github/workflows/tests-levels-matrix.php
@@ -2,9 +2,15 @@
 
 shell_exec('php vendor/bin/phpunit --group levels --list-tests-xml test-list.xml');
 
+libxml_use_internal_errors(true);
 $simpleXml = simplexml_load_file('test-list.xml');
 if ($simpleXml === false) {
-	throw new RuntimeException('Error loading test-list.xml');
+	$errors = [];
+	foreach (libxml_get_errors() as $error) {
+		$errors[] = $error->message;
+	}
+
+	throw new RuntimeException('Error loading test-list.xml: ' . implode(', ', $errors));
 }
 
 $testFilters = [];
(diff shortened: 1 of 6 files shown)


src/Testing/TypeInferenceTestCase.php
```diff
@@ -467,6 +473,8 @@ private static function isFileLintSkipped(string $file): bool
 
 			if (preg_match('~<?php\\s*\\/\\/\s*lint\s*([^\d\s]+)\s*([^\s]+)\s*~i', $firstLine, $m) === 1) {
```
staabm, line 474: here we can see that for `TypeInferenceTestCase` to work properly, we expect `<?php // lint` at the very start

.github/workflows/tests-levels-matrix.php
```diff
@@ -1,10 +1,19 @@
...
+libxml_use_internal_errors(true);
 $simpleXml = simplexml_load_file('test-list.xml');
 if ($simpleXml === false) {
-	throw new RuntimeException('Error loading test-list.xml');
+	$errors = [];
+	foreach (libxml_get_errors() as $error) {
+		$errors[] = $error->message;
+	}
+
+	throw new RuntimeException('Error loading test-list.xml: ' . implode(', ', $errors));
```
staabm, line 16: make sure when a underlying error occurs we see a meaningful error with hints for the root-cause

src/Testing/TypeInferenceTestCase.php
```diff
@@ -467,6 +473,8 @@ private static function isFileLintSkipped(string $file): bool
...
 				return version_compare(PHP_VERSION, $m[2], $m[1]) === false;
+			} elseif (str_contains($firstLine, 'lint')) {
+				throw new LogicException(sprintf('// lint comment must immediately follow the php starting tag in %s', $file));
```
staabm: the new error which will be triggered

VincentLanglet (approved):

PR #4988, opened 2026-02-18
Fix #12574: Erroneous Offset might not exist

staabm:
Fixes phpstan/phpstan#12574

Closes https://github.com/phpstan/phpstan/issues/10040
Closes https://github.com/phpstan/phpstan/issues/11102
Closes https://github.com/phpstan/phpstan/issues/11870
Closes https://github.com/phpstan/phpstan/issues/6991
Closes https://github.com/phpstan/phpstan/issues/7716

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -6811,7 +6811,6 @@ class NodeScopeResolver
 			} elseif (
 				!$expr instanceof PropertyFetch
 				&& !$expr instanceof ArrayDimFetch
-				&& !$expr instanceof StaticPropertyFetch
 			) {
 				continue;
 			}
@@ -6850,7 +6849,6 @@ class NodeScopeResolver
 			} elseif (
 				!$expr instanceof PropertyFetch
 				&& !$expr instanceof ArrayDimFetch
-				&& !$expr instanceof StaticPropertyFetch
 			) {
 				continue;
 			}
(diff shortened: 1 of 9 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -6800,14 +6800,18 @@ private function unwrapAssign(Expr $expr): Expr
...
+					continue;
+				}
 
-			if ($expr->name === $variableName) {
+				if ($expr->name === $variableName) {
+					continue;
+				}
+			} elseif (
+				!$expr instanceof PropertyFetch
+				&& !$expr instanceof ArrayDimFetch
```
staabm, line 6813: I started thinking about whether we need to verify the `$expr`'s `*Fetch` does not reference the `$variableName` - similar to how it happens above for `$expr instanceof Variable`.

did not yet find a case in which it makes a difference

VincentLanglet (commented):
The fix is correct, I just wonder if we can fix more situation.
I added test suggestion if you think they are worth it.

(You let this PR in draft ?)

staabm:
> The fix is correct, I just wonder if we can fix more situation.

I guess AI will find a fix as soon as someone reports a example :-)

VincentLanglet:
You added a lot of non regression test related to some issues ; you need to update the PR description to tell which ones are fixed, no ?

staabm:
Thanks

PR #4991, opened 2026-02-18
Fix error reported on the wrong line

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/14150


VincentLanglet:
The failure on phpstan-doctrine seems legit since now it's report on `getQuery` and not on the start of the line
https://github.com/phpstan/phpstan-doctrine/blob/734ef36c2709b51943f04aacadddb76f239562d3/tests/Rules/Doctrine/ORM/data/query-builder-dynamic.php#L43

But maybe this behavior (and the initial PR https://github.com/phpstan/phpstan-src/pull/4967) should be behind bleedingEdge ?

tests/PHPStan/Rules/Properties/data/bug-14150.php
```diff
@@ -0,0 +1,29 @@
...
+	{
+		$this
+			->x = null;
+	}
+
+	public function test4(): void
+	{
+		$this
+			->x()
+			->x = null;
```
staabm, line 27: Please also test nullsafe property fetch

src/Rules/FunctionCallParametersCheck.php
```diff
@@ -89,11 +90,7 @@ public function check(
 		string $namedArgumentMessage,
 	): array
 	{
-		if ($funcCall instanceof Node\Expr\MethodCall || $funcCall instanceof Node\Expr\StaticCall || $funcCall instanceof Node\Expr\FuncCall) {
-			$funcCallLine = $funcCall->name->getStartLine();
-		} else {
-			$funcCallLine = $funcCall->getStartLine();
-		}
+		$funcCallLine = RuleErrorTransformer::getStartLineFromNode($funcCall);
```
staabm: Do we still need this line?

tests/PHPStan/Rules/Properties/data/bug-14150.php
```diff
@@ -0,0 +1,29 @@
...
+	 * @return $this
+	 */
+	public function x()
+	{
+		return $this;
+	}
+
+	public function test3(): void
+	{
+		$this
```
staabm, line 19: It just came to mind that we should also test pipe operator

src/Rules/FunctionCallParametersCheck.php
VincentLanglet: Yes, because
- I compute the line from the node only if it's not set and this Check set the line.
- If I remove the `->line(...)` calls, since the AttributesCheck calls the functionCallParametersCheck with another Node than the one from the ClassAttributesRule it makes a test from ClassAttributesRuleTest to fail.

tests/PHPStan/Rules/Properties/data/bug-14150.php
VincentLanglet: Do we have a phpstan reproducer with a dedicated rule ? 

Since the pipe operator is an operator I dont see why the issue would exist for this.
It's the same than `+` on multiline for instance.

tests/PHPStan/Rules/Properties/data/bug-14150.php
VincentLanglet: I added test for NullsafePropertyFetch and NullsafeMethodCall.

It should cover most of the issue.
If @mvorisek finds more after this PR we can still iterate.

VincentLanglet: NB: this is not perfect since if someone writes

```
$this->
     x()?->
     y()
```
this might be considered to be reported on the wrong line (on the `y` line)

But when I dump the NullsafeMethodCall I get
```
object(PhpParser\Node\Expr\NullsafeMethodCall)#3004 (4) {
  ["attributes":protected]=>
  array(7) {
    ["startLine"]=>
    int(19)
    ["startTokenPos"]=>
    int(60)
    ["startFilePos"]=>
    int(225)
    ["endLine"]=>
    int(21)
    ["endTokenPos"]=>
    int(70)
    ["endFilePos"]=>
    int(248)
    ["phpstan_cache_printer"]=>
    string(16) "$this->x()?->y()"
  }
  ["var"]=>
  object(PhpParser\Node\Expr\MethodCall)#3005 (4) {
    ["attributes":protected]=>
    array(7) {
      ["startLine"]=>
      int(19)
      ["startTokenPos"]=>
      int(60)
      ["startFilePos"]=>
      int(225)
      ["endLine"]=>
      int(20)
      ["endTokenPos"]=>
      int(65)
      ["endFilePos"]=>
      int(238)
      ["phpstan_cache_printer"]=>
      string(10) "$this->x()"
    }
    ["var"]=>
    object(PhpParser\Node\Expr\Variable)#3002 (2) {
      ["attributes":protected]=>
      array(7) {
        ["startLine"]=>
        int(19)
        ["startTokenPos"]=>
        int(60)
        ["startFilePos"]=>
        int(225)
        ["endLine"]=>
        int(19)
        ["endTokenPos"]=>
        int(60)
        ["endFilePos"]=>
        int(229)
        ["phpstan_cache_printer"]=>
        string(5) "$this"
      }
      ["name"]=>
      string(4) "this"
    }
    ["name"]=>
    object(PhpParser\Node\Identifier)#2996 (2) {
      ["attributes":protected]=>
      array(6) {
        ["startLine"]=>
        int(20)
        ["startTokenPos"]=>
        int(63)
        ["startFilePos"]=>
        int(236)
        ["endLine"]=>
        int(20)
        ["endTokenPos"]=>
        int(63)
        ["endFilePos"]=>
        int(236)
      }
      ["name"]=>
      string(1) "x"
    }
    ["args"]=>
    array(0) {
    }
  }
  ["name"]=>
  object(PhpParser\Node\Identifier)#3007 (2) {
    ["attributes":protected]=>
    array(6) {
      ["startLine"]=>
      int(21)
      ["startTokenPos"]=>
      int(68)
      ["startFilePos"]=>
      int(246)
      ["endLine"]=>
      int(21)
      ["endTokenPos"]=>
      int(68)
      ["endFilePos"]=>
      int(246)
    }
    ["name"]=>
    string(1) "y"
  }
  ["args"]=>
  array(0) {
  }
}
```
with no `startLine` of 20 I could use.

And I never saw somone writing his code this way anyway

VincentLanglet: Also added static calls.

tests/PHPStan/Rules/Properties/data/bug-14150.php
staabm: see https://phpstan.org/blog/phpstan-fully-supports-php-8-5#pipe-operator

> What’s interesting is that most of PHPStan doesn’t actually see the pipe operator at all. The AST gets rewritten on-the-fly to traditional function calls. This is pretty neat because it means all existing rules — type checking, argument validation, return type inference — just work with pipe expressions for free. No need to go and teach every single rule about a new syntax. The transformation handles FuncCall, MethodCall, and StaticCall with first-class callables, so $x |> $obj->method(...) simply becomes $obj->method($x).

I am not sure whether pipe operator regarding this rules is more a operator or more a chain of method calls ;)

src/Analyser/RuleErrorTransformer.php
```diff
@@ -75,6 +75,8 @@ public function transform(
 			&& $ruleError->getLine() !== -1
 		) {
 			$line = $ruleError->getLine();
+		} else {
+			$line = self::getStartLineFromNode($node);
```
staabm: I just had a look thru the codebase.

we have already `PhpDocLineHelper::detectLine()` which is called like [`line(PhpDocLineHelper::detectLine($node, $phpDocTag))`](https://github.com/phpstan/phpstan-src/blob/a10a9a264cc1d2abe25adb82adc013a5f45b4eae/src/Rules/PhpDoc/InvalidPHPStanDocTagRule.php#L116) .

this makes me think we should not have this special fallback case here, but call something like `NodeChainLineHelper::detectLine($node)` right at the ErrorBuilder call-sites where `->line()` is invoked.

tests/PHPStan/Rules/Properties/data/bug-14150.php
staabm: :+1:

src/Analyser/RuleErrorTransformer.php
VincentLanglet: > this makes me think we should not have this special fallback case here, but call something like NodeChainLineHelper::detectLine($node) right at the ErrorBuilder call-sites where ->line() is invoked.

Lot of existing rules are not calling `line()` and rely on the defaut fallback which was already implemented in RuleErrortransformer `$node->getStartLine()`.

There is tons of rule to fix, so it would be easier to modify the fallback.

Even binary operator are wrong
https://phpstan.org/r/090b54c3-0255-4027-9934-491d8ab962b7

And yes PipeOperator has the same issue too https://phpstan.org/r/7ab46b15-18c9-4d5b-bfc9-27bc78efe3d4

But there is so much to fix, I feel like we should first:
- validate a strategy with ondrej 
- Fix/Implement it, step by step

I start to feel like we should fix rules one by one...

staabm: agree - thanks for looking into it

VincentLanglet: I tried another strategy with https://github.com/phpstan/phpstan-src/pull/4991/commits/af0f9843c99015787ff5ec3bca47dce71578d2dc

I feel like it's safer to avoid a general strategy in RuleTransformer and instead to fix rules one by one. WDYT ?

staabm: yes, thats what I had in [mind with my comment](https://github.com/phpstan/phpstan-src/pull/4991#discussion_r2826034693)

VincentLanglet:
I think it can be merged like this.

I focused on method and properties.

staabm (approved):
I feel this looks really good

ondrejmirtes:
Nice! I'd say this is a bugfix and it won't impact too much code I think.

But I'd say in some cases making the line number more precise can be considered a BC break. For example some time ago I set more precise lines for multiline calls:

```
doFoo(
    $arg1,
    $arg2,
    $arg3,
);
```

Previously all errors were reported on the `doFoo` line but I made sure that error about parameter 2 started to be reported next to `$arg2`. But because there's a lot of code like this out there with lots of people ignoring it, I only did this change for bleeding edge.

So please keep this in mind, we can't always make the line more precise because of backward compatibility.

PR #4994, opened 2026-02-19
More precise array_key_first/array_key_last inference

staabm:
refs https://github.com/phpstan/phpstan/issues/14081#issuecomment-3925672432


staabm:
I will look into a similar optimization for other hardcoded functions in a separate PR

tests/PHPStan/Analyser/nsrt/bug-14081.php
```diff
@@ -0,0 +1,68 @@
...
+	assertType('0|1|null', $key);
+	assertType("'one'|'two'", $array[$key]);
+}
+
+/** @param list<string> $array */
+function firstNotNull(array $array): mixed
+{
+	if (($key = array_key_first($array)) !== null) {
+		assertType('int<0, max>', $key); // could be int<1, max>
+		assertType('list<string>', $array); // could be non-empty-list<string>
```
VincentLanglet: Why could it be `int<1, max>` ?

You're checking `$key !== null` The 0 key could be in the if.

It could be non-empty indeed.

tests/PHPStan/Analyser/nsrt/bug-14081.php
```diff
@@ -0,0 +1,68 @@
...
+		return $array[$key];
+	}
+	return null;
+}
+
+/** @param list<string> $array */
+function lastNotNull(array $array): mixed
+{
+	if (($key = array_key_last($array)) !== null) {
+		assertType('int<0, max>', $key); // could be int<1, max>
```
VincentLanglet: Same

src/Analyser/TypeSpecifier.php
```diff
@@ -725,7 +725,44 @@ public function specifyTypesInCondition(
...
+							$this->create($arrayArg, new NonEmptyArrayType(), TypeSpecifierContext::createTrue(), $scope),
+						);
+						$isNonEmpty = true;
+					} else {
+						$isNonEmpty = $arrayType->isIterableAtLeastOnce()->yes();
+					}
+
+					if ($isNonEmpty) {
+						$dimFetch = new ArrayDimFetch($arrayArg, $expr->var);
+						$iterableValueType = $expr->expr->name->toLowerString() === 'array_key_first'
```
VincentLanglet:
```
$iterableValueType = $expr->expr->name->toLowerString() === 'array_key_first'
							? $arrayType->getIterableValueType()
							: $arrayType->getIterableValueType();
```
is weird ; you're calling `$arrayType->getIterableValueType()` in both case no ?

staabm: ha, great find.

VincentLanglet (approved):
Does it works if `$key = array_key_last($array)` is done before the if ?

staabm:
added a test which I think covers your question

staabm:
thank you

PR #5000, opened 2026-02-19
Fix #5020: phpstan could handle $class::someMethod syntax better

staabm:
inspired by https://github.com/phpstan/phpstan-src/pull/4911

Fixes phpstan/phpstan#5020

### DIFF
--- a/tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php
+++ b/tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php
@@ -43,10 +43,6 @@ class CallToStaticMethodStatementWithoutSideEffectsRuleTest extends RuleTestCase
 				12,
 			],
 			[
-				'Call to static method DateTimeImmutable::createFromFormat() on a separate line has no effect.',
-				13,
-			],
-			[
 				'Call to method DateTime::format() on a separate line has no effect.',
 				23,
 			],
--- a/tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php
+++ b/tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php
@@ -43,6 +43,10 @@ class CallToStaticMethodStatementWithoutSideEffectsRuleTest extends RuleTestCase
 				12,
 			],
 			[
+				'Call to static method DateTimeImmutable::createFromFormat() on a separate line has no effect.',
+				13,
+			],
+			[
 				'Call to method DateTime::format() on a separate line has no effect.',
 				23,
 			],
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -2044,19 +2044,7 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 		return new ObjectType($originalClass);
 	}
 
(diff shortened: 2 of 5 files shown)


tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php
```diff
@@ -42,6 +42,10 @@ public function testRulePhp7(): void
 				'Call to static method DateTimeImmutable::createFromFormat() on a separate line has no effect.',
 				12,
 			],
+			[
+				'Call to static method DateTimeImmutable::createFromFormat() on a separate line has no effect.',
+				13,
+			],
```
staabm: I think it makes sense, that we see the error on line 13. I feel this one was missing before

https://github.com/phpstan/phpstan-src/blob/0bfbaf9e1c2fb8d41a4dfca848a3788fa807493f/tests/PHPStan/Rules/Methods/data/static-method-call-statement-no-side-effects.php#L10-L14

VincentLanglet: Isn't related to the fact you can extend \DateTimeImmutable and override createFromFormat with impure method ? 

Maybe not for DateTimeImmutable (dunno if it's final), but it seems possible for any non final class no ?
So we should add a test with a random non final class (?)

staabm: I think you are right. 

the code atm is understood as "static call to DateTimeImmutable::createFromFormat()" but not as a "late static binding call to $x::createFromFormat()".

and `DateTimeImmutable::createFromFormat()` did not throw on PHP7.x - therefore we get the test-error only on PHP 7.4

staabm: https://github.com/phpstan/phpstan-src/pull/5009 adds a test, which emits the following unexpected errors within this PR here:

```
➜  phpstan-src git:(tests) ✗ vendor/bin/phpunit tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php
Xdebug: [Step Debug] Could not connect to debugging client. Tried: 127.0.0.1:9003 (through xdebug.client_host/xdebug.client_port).

Runtime:       PHP 8.3.30
Configuration: /Users/staabm/workspace/phpstan-src/phpunit.xml
Random Seed:   1771604476

FSS......                                                           9 / 9 (100%)

Time: 00:00.594, Memory: 86.50 MB

There was 1 failure:

1) PHPStan\Rules\Methods\CallToStaticMethodStatementWithoutSideEffectsRuleTest::testDynamicStaticCall
--- Expected
+++ Actual
@@ @@
-'
+'34: Call to static method DynamicStaticCall\Foo::doFoo() on a separate line has no effect.
+35: Call to static method DynamicStaticCall\FinalFoo::doFoo() on a separate line has no effect.
+36: Call to static method DynamicStaticCall\Bar::finalFoo() on a separate line has no effect.
 '

/Users/staabm/workspace/phpstan-src/src/Testing/RuleTestCase.php:236
/Users/staabm/workspace/phpstan-src/tests/PHPStan/Rules/Methods/CallToStaticMethodStatementWithoutSideEffectsRuleTest.php:126

FAILURES!
Tests: 9, Assertions: 7, Failures: 1, Skipped: 2.
```

PR #5001, opened 2026-02-19
Introduce ExtendedMethodReflection::isDummy

VincentLanglet:
Similar to ExtendedPropertyReflection::isDummy


VincentLanglet:
I'm not sure it's the right fix.

PR #5003, opened 2026-02-19
False Positive on Match Arm Comparison with Incremented Bounded Integer

VincentLanglet:
Inspired from https://github.com/phpstan/phpstan-src/pull/4968

Closes https://github.com/phpstan/phpstan/issues/11310

### DIFF
--- a/tests/PHPStan/Rules/Comparison/data/bug-11310.php
+++ b/tests/PHPStan/Rules/Comparison/data/bug-11310.php
@@ -1,0 +1,28 @@
+<?php // lint >= 8.0
+
+namespace Bug11310;
+
+/** @param int<0, max> $i */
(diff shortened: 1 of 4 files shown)


tests/PHPStan/Rules/Comparison/data/bug-11310.php
```diff
@@ -0,0 +1,28 @@
...
+
+/** @param int<0, max> $i */
+function foo(int $i): void {
+	echo match ($i++) {
+		0 => 'zero',
+		default => 'default',
+	};
+}
+
+/** @param int<0, max> $i */
```
staabm, line 13: Please add another test with "union of int-range" and another one with "union of scalar ints"

staabm (approved):

PR #5004, opened 2026-02-19
Remove callable.nonNativeMethod

VincentLanglet:
As suggested by https://github.com/phpstan/phpstan/issues/13596#issuecomment-3929332346

Closes https://github.com/phpstan/phpstan/issues/13596

cf https://onlinephp.io/c/5de57

### DIFF
--- a/src/Rules/Methods/MethodCallableRule.php
+++ b/src/Rules/Methods/MethodCallableRule.php
@@ -5,12 +5,10 @@ namespace PHPStan\Rules\Methods;
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\RegisteredRule;
-use PHPStan\Internal\SprintfHelper;
 use PHPStan\Node\MethodCallableNode;
 use PHPStan\Php\PhpVersion;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
-use function sprintf;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Rules/Methods/MethodCallableRule.php
```diff
@@ -46,23 +44,7 @@ public function processNode(Node $node, Scope $scope): array
...
-		}
-
-		$messagesMethodName = SprintfHelper::escapeFormatString($declaringClass->getDisplayName() . '::' . $methodReflection->getName() . '()');
-
-		$errors[] = RuleErrorBuilder::message(sprintf('Creating callable from a non-native method %s.', $messagesMethodName))
-			->identifier('callable.nonNativeMethod')
-			->build();
-
-		return $errors;
+		return $this->methodCallCheck->check($scope, $methodNameName, $node->getVar(), $node->getName())[0];
```
ondrejmirtes, line 47: I think the returning of the array shape with the method reflection might have been introduced exactly because of this. It might be possible to simplify the return type and let it just returns `errors` instead of the tuple.

VincentLanglet: I looked and the shape seems to always have existed, and the methodReflection is used somewhere else
https://github.com/phpstan/phpstan-src/blob/dc7d857cad10564cd5bacfc3e1d7df496cd2e609/src/Rules/Methods/CallMethodsRule.php#L65

ondrejmirtes:
Thank you!

PR #5006, opened 2026-02-20
Fix #14124: Reassigning to `array<string, list<T>>` out parameter does not see the `list<T>` property during assignment

staabm:
inspired by https://github.com/phpstan/phpstan-src/pull/4999

Fixes phpstan/phpstan#14124


VincentLanglet (approved):
There is two tests and three place you modify the code ; does every change is covered ?

PR #5007, opened 2026-02-20
Fix stub @param types not applied inside method bodies (regression since 2.1.38)

mglaman:
## Summary

- Fixes `NodeScopeResolver::getPhpDocs()` to consult `StubPhpDocProvider` before falling back to `FileTypeMapper` when resolving parameter types inside method bodies
- Adds `StubPhpDocProvider` as a constructor parameter to `NodeScopeResolver`
- Updates all direct `NodeScopeResolver` constructor call sites in test infrastructure to include the new parameter

## Problem

Since 2.1.38 (PR #4829 — "Rework phpDoc inheritance to resolve through reflection instead of re-walking the hierarchy"), stub `@param` type overrides were silently ignored inside method bodies. External callers of a stubbed method still saw the correct stub types, but the method's own body would use the source file's `@param` type instead.

**Root cause**: The removed `PhpDocBlock::docBlockToResolvedDocBlock()` helper previously called `StubPhpDocProvider::findMethodPhpDoc()` first and fell back to `FileTypeMapper` only if no stub was found. The rewritten path in `getPhpDocs()` only consulted `FileTypeMapper`, never checking stubs.

Fixes https://github.com/phpstan/phpstan/issues/14118

## Test plan

- [ ] New `StubMethodBodyTypesTest` verifies stub `@param callable-string` overrides source `@param string` inside the method body
- [ ] New test also confirms external callers still see stub `@return non-empty-string`
- [ ] Full analyser + rules test suites pass (7752 tests, 0 failures)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mglaman:
Superseded by #5008 which targets 1.11.x (2.1.x releases) per bot feedback.

PR #5009, opened 2026-02-20
Cover dynamic static calls in tests

staabm:
lets add coverage for this scenario with more basic classes which do not depend on phpversion (like DateTimeImmutable via phpstorm stubs)

### DIFF
--- a/tests/PHPStan/Rules/Methods/data/dynamic-static-call.php
+++ b/tests/PHPStan/Rules/Methods/data/dynamic-static-call.php
@@ -29,8 +29,8 @@ class Bar {
 
 class Baz {
 	function doBaz(Foo $foo, FinalFoo $finalFoo, Bar $bar):void {
-		$foo::doFoo();
-		$finalFoo::doFoo();
-		$bar::finalFoo();
+		$foo::doFoo(); // no error, subclass could override static method with impure impl
+		$finalFoo::doFoo(); // could error, because final class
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Rules/Methods/data/dynamic-static-call.php
```diff
@@ -0,0 +1,36 @@
...
+		return 42;
+	}
+}
+
+
+class Baz {
+	function doBaz(Foo $foo, FinalFoo $finalFoo, Bar $bar):void {
+		$foo::doFoo(); // no error, subclass could override static method with impure impl
+		$finalFoo::doFoo(); // could be "Call to static method .. on a separate line has no effect", because final class
+		$bar::finalFoo(); // could be "Call to static method .. on a separate line has no effect", because final method
```
VincentLanglet, line 34: I'm not sure to understand the PR.
Is it worth covering this with a test with missing expectation ?

Or do you plan fixing this in the next PR ? (Then should it fix in the same ?)

staabm: the PR is asserting the current state of things so we see when we change existing behaviour.

I think the possible 
> "Call to static method .. on a separate line has no effect"

are hardly useful, but they would not be wrong when they start to appear. (for 2 of the 3 examples)

VincentLanglet (approved):

PR #5018, opened 2026-02-21
TypeSpecifier: Prevent duplicate expression printing

staabm:
its invoked a lot

grafik

only very minor improvements

---

before/after PR diff

`blackfire run --ignore-exit-status php bin/phpstan analyse src/Analyser/ -v --debug`

grafik

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -2017,18 +2017,18 @@ final class TypeSpecifier
 		$sureNotTypes = [];
 		if ($context->false()) {
 			$exprString = $this->exprPrinter->printExpr($expr);
-			$originalExprString = $this->exprPrinter->printExpr($originalExpr);
-
 			$sureNotTypes[$exprString] = [$expr, $type];
-			if ($exprString !== $originalExprString) {
+
+			if ($expr !== $originalExpr) {
+				$originalExprString = $this->exprPrinter->printExpr($originalExpr);
 				$sureNotTypes[$originalExprString] = [$originalExpr, $type];
 			}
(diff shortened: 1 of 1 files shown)


src/Analyser/TypeSpecifier.php
```diff
@@ -2017,18 +2017,18 @@ private function createForExpr(
 		$sureNotTypes = [];
 		if ($context->false()) {
 			$exprString = $this->exprPrinter->printExpr($expr);
-			$originalExprString = $this->exprPrinter->printExpr($originalExpr);
-
 			$sureNotTypes[$exprString] = [$expr, $type];
-			if ($exprString !== $originalExprString) {
+
+			if ($expr !== $originalExpr) {
```
VincentLanglet, line 2022: Is it really saving calls ? The code was comparing the string but now you are comparing the expr.

Also should you still check string after in case two expr has the same string print ? (Dunno if its possible)

staabm: updated the PR description with a diff

staabm: the 2 expressions involved can be compared this way, because they originate from a assignment a few lines above

https://github.com/phpstan/phpstan-src/pull/5018/files#diff-1076e38598f300a69c9dc998dae31374052f95bd69d0d71c11e9d0a4c188976dL1933

staabm: also note that even if we theoretically would preduce the same string again, we would overwrite the previous existing value with the very same touple again

VincentLanglet (approved):

PR #5019, opened 2026-02-21
Simplify MemoizingContainer

staabm:


VincentLanglet (approved):
Technically it's not the same if the value memoize is NULL, but I agree it shouldn't be the case here

PR #5020, opened 2026-02-21
Fix SimplXmlElement and GMP casts

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/14136

I tried https://3v4l.org/BtQlq#v8.4.18

Fun fact, I discovered that GMP is not castable to bool for PHP < 8.4
See https://wiki.php.net/rfc/fix_up_bcmath_number_class

But:
- Currently ObjectType::toBool was returning ConstantBoolean(true), so returning a BooleanType is still better (cause for PHP >= 8.4 it can be false !)
- `boolval` is defined as accepting `mixed` in the resourceMap
- There is currently no way to do a condition about the PHPVersion inside ObjectType::toBool
- Nobody reported the issue yet

So I think it's a good enough improvement yet.


src/Type/ObjectType.php
```diff
@@ -785,15 +788,21 @@ public function toInteger(): Type
 
 	public function toFloat(): Type
 	{
-		if ($this->isInstanceOf('SimpleXMLElement')->yes()) {
+		if (
+			$this->isInstanceOf('SimpleXMLElement')->yes()
```
mvorisek, line 792:
```suggestion
			$this->isInstanceOf(\SimpleXMLElement::class)->yes()
```

staabm: I think using `\SimpleXMLElement::class` might lead to problems when php is compiled with `--disable-simplexml`

mvorisek: It won't, try `\BlaBla::class`, this is only refactorable way to write the string, no autoload is involved.

VincentLanglet: I don't really see a reason to change the existing check ; especially when it was done this way in a dedicated commit
https://github.com/phpstan/phpstan-src/commit/2ac2d25f4c3975d6f3f847fdce145cf07583e6ba

staabm:
> * There is currently no way to do a condition about the PHPVersion inside ObjectType::toBool

in case you would want the PhpVersion at this point, there is also `PhpVersionStaticAccessor`.
(but I agree its a edge case maybe not worth caring).

In my over 20 years of PHP I have never seen a app which uses GMP. therefore I would not spent too much time on it and spent time elsewhere (priority wise)

staabm:
(sorry, wrong button)

tests/PHPStan/Analyser/nsrt/bug-14136.php
```diff
@@ -0,0 +1,30 @@
...
+
+use function PHPStan\Testing\assertType;
+
+$xml = <<<'EOT'
+<?xml version="1.0"?><item>3</item>
+EOT;
+
+$xml = simplexml_load_string($xml);
+assertType('int', intval($xml));
+assertType('int', intval(gmp_init(42)));
```
staabm, line 13: the mutations make me think we should also test unions like

```
intval(rand(0,1) ? gmp_init(42) : 'abc');
intval(rand(0,1) ? gmp_init(42) : []);
intval(rand(0,1) ? gmp_init(42) : '123');
```

staabm (approved):

PR #5022, opened 2026-02-21
Add ClassConstantReflection::isFinalByKeyword() method

samsonasik:
Closes https://github.com/phpstan/phpstan/issues/14171


samsonasik:
@ondrejmirtes @VincentLanglet ready 👍

tests/PHPStan/Reflection/data/ClassWithConstants.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace SomeNamespace;
```
VincentLanglet: I just wonder if a "more specific" namespace should be use to avoid later possible conflict...

VincentLanglet (approved):

tests/PHPStan/Reflection/data/ClassWithConstants.php
samsonasik: sure, any suggestion?

VincentLanglet: For Bug, I always use BugNumberOfTheIssue ; but seems like it's not that used for feature request (while it could have been used, like FeatureNumberOfTheIssue or IssueNumberOfTheIssue).

Also, I would have say that namespace are often related to the name of the file so the file could have been named `final-class-constant.php` with a namespace `FinalClassConstant` but on the other side I found 3 class with the namespace `AttributeReflectionTest` and three `FunctionReflectionDocTest` so maybe `ClassConstantReflectionTest` is the best choice to be consistant ?

samsonasik: ok, updated namespace to `ClassConstantReflectionTest` 👍 https://github.com/phpstan/phpstan-src/pull/5022/changes/bc3cb6d522587043fd3c45dbf76c4d51c81cc18f

ondrejmirtes:
Thank you!

PR #5023, opened 2026-02-22
Assert ExprCacheHelper->import() will not mix up same expr-string based attributes

staabm:
test when attributes which represent the same value expression don't inherit data from a previous call to `ExprCacheHelper::import()` because of a poisened cache in the underlying `CachedParser`.

this test makes sure we don't regress after starting to re-use the `CachedParser` in https://github.com/ondrejmirtes/BetterReflection/pull/41.

with this PR in place, we can be sure that we don't get back the bug which was fixed by https://github.com/ondrejmirtes/BetterReflection/commit/e124747f5e8097089d6a2eb9c4d11ad978de656d when reverting said commit.


ondrejmirtes:
Thank you!

PR #5027, opened 2026-02-22
ClassNameCheck: Retrieve extensions just once

staabm:
followup to https://github.com/phpstan/phpstan-src/pull/4926 - missed this one call-site with my search regex on the first pass


ondrejmirtes:
Thank you!

PR #5029, opened 2026-02-22
Unsetting optional offset might still give a list.

VincentLanglet:
When doing `array_key_exists` on `list{0: string, 1: string, 2?: string, 3?: string}`,
the ArrayKeyExistsFunctionTypeSpecifyingExtension is specify `HasOffsetType` and in the else condition we're calling `tryRemove` on this `HasOffsetType`.

ConstantArrayType::tryRemove do
```
		if ($typeToRemove instanceof HasOffsetType) {
			return $this->unsetOffset($typeToRemove->getOffsetType());
		}
```
but the new array created is always with `$this->isList = Ternary::no`.

The issue is that the intersection with List gives Never ; see
https://phpstan.org/r/c8177c3c-a643-43dc-b62f-b207740d75f8

Technically, 
- removing an optional key could still be a list
- removing the last non optional key (which is not the easiest check since they might not be ordered) could still be a list

This is fully related to https://github.com/phpstan/phpstan/issues/11628
see https://phpstan.org/r/fb18be32-9fba-42ba-8815-b767734cac6a

This is kinda tricky as changing `$newIsList = TrinaryLogic::createMaybe()` to maybe or true will then creates an issue if I append an element later (cause PHPStan won't detect it's not a list anymore).

Also, some extra thought: when we're inside the `else` part of an array_key_exists call we're not really calling unset on the array. So maybe a simpler way to solve this, is to
- Ignore the issue with `unset`
- Solving the array_key_exists issue by
  - Adding an optional param to ConstantArrayType::unsetOffset like something `$newIsList` or preserve list
  - Changing the call inside ConstantArrayType::tryRemove to override the behavior of isList when trying to remove HasOffsetType and HasOffsetValueType.
  
Closes https://github.com/phpstan/phpstan/issues/14177

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -741,7 +741,21 @@ class ConstantArrayType implements Type
 					$k++;
 				}
 
-				return new self($newKeyTypes, $newValueTypes, $this->nextAutoIndexes, $newOptionalKeys, TrinaryLogic::createNo());
+				$newIsList = TrinaryLogic::createNo();
+				if (!$this->isList->no() && in_array($i, $this->optionalKeys, true)) {
+					$preserveIsList = true;
+					foreach ($newKeyTypes as $k2 => $newKeyType2) {
+						if (!$newKeyType2 instanceof ConstantIntegerType || $newKeyType2->getValue() !== $k2) {
+							$preserveIsList = false;
+							break;
+						}
+					}
+					if ($preserveIsList) {
+						$newIsList = $this->isList;
+					}
+				}
+
+				return new self($newKeyTypes, $newValueTypes, $this->nextAutoIndexes, $newOptionalKeys, $newIsList);
 			}
 
 			return $this;
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -742,7 +742,10 @@ class ConstantArrayType implements Type
 				}
 
 				$newIsList = TrinaryLogic::createNo();
-				if (!$this->isList->no()) {
+				// We're unsetting something that might not be on the array,
+				// so it might still be a list (with PHPStan definition)
+				// because the nextAutoIndexes will not change.
+				if (!$this->isList->no() && in_array($i, $this->optionalKeys, true)) {
 					$preserveIsList = true;
 					$isListOnlyIfKeysAreOptional = false;
 					foreach ($newKeyTypes as $k2 => $newKeyType2) {
@@ -765,7 +768,7 @@ class ConstantArrayType implements Type
 					}
 
 					if ($preserveIsList) {
-						$newIsList = $this->isList;
+						$newIsList = TrinaryLogic::createMaybe();
 					}
 				}
 
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -821,6 +821,9 @@ class ConstantArrayType implements Type
 	/**
 	 * When we're unsetting something not on the array, it will be untouched,
 	 * So the nextAutoIndexes won't change, and the array might still be a list even with PHPStan definition.
+	 *
+	 * @param list<ConstantIntegerType|ConstantStringType> $newKeyTypes
+	 * @param int[] 									   $newOptionalKeys
 	 */
 	private static function isListAfterUnset(array $newKeyTypes, array $newOptionalKeys, TrinaryLogic $arrayIsList, bool $unsetOptionalKey): TrinaryLogic
 	{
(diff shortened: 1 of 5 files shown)


src/Type/Constant/ConstantArrayType.php
```diff
@@ -741,7 +741,12 @@ public function unsetOffset(Type $offsetType): Type
 					$k++;
 				}
 
-				return new self($newKeyTypes, $newValueTypes, $this->nextAutoIndexes, $newOptionalKeys, TrinaryLogic::createNo());
+				$newIsList = TrinaryLogic::createNo();
+				if (!$this->isList->no() && in_array($i, $this->optionalKeys, true)) {
+					$newIsList = TrinaryLogic::createMaybe();
```
ondrejmirtes: Not always, right? The new type is a list only if we're removing an optional key and:

* The key is the last one
* Or all of the keys after it are also optional

If you're going to implement this logic, add failing tests first 😊

ondrejmirtes (changes requested):

src/Type/Constant/ConstantArrayType.php
staabm: Per https://github.com/phpstan/phpstan/issues/12768 even unsetting the last item makes it no longer a list because of internal pointer and append with [] will create a hole

ondrejmirtes: Yeah but the point is that we're unsetting something that might not be on the array. So it might still be a list.

src/Type/Constant/ConstantArrayType.php
```diff
@@ -770,21 +779,77 @@ public function unsetOffset(Type $offsetType): Type
...
+
+		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, $newIsList);
+	}
+
+	/**
+	 * When we're unsetting something not on the array, it will be untouched,
+	 * So the nextAutoIndexes won't change, and the array might still be a list even with PHPStan definition.
+	 *
+	 * @param list<ConstantIntegerType|ConstantStringType> $newKeyTypes
+	 * @param int[] 									   $newOptionalKeys
```
ondrejmirtes: I don't like these alignments. There should be just a single space between the type and variable.

tests/PHPStan/Analyser/nsrt/bug-14177.php
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	/**
+	 * @param list{0: string, 1: string, 2?: string, 3?: string} $b
+	 */
+	public function testUnset1OnList(array $b): void
+	{
+		assertType('true', array_is_list($b));
+		unset($b[1]);
+		assertType('false', array_is_list($b)); // Could be true
```
ondrejmirtes: It could not. When you're unsetting an existing key, adding a new one will create a gap.

ondrejmirtes (changes requested):

tests/PHPStan/Analyser/nsrt/bug-14177.php
VincentLanglet: Cf https://3v4l.org/4GBOm#v8.3.30

It's still a list, and it change when we add back a new element
So technically it could be true, and PHPStan should be aware it's false as soon as we add back an element.

But I agree it's harder.

I removed the comments.

ondrejmirtes: The problem is that the logic is unsound when passing the array around to function arguments.

When we accept a list, it must mean that after appending a new item, it's still a list.

VincentLanglet: Yes. Maybe the right fix will be to introduce a DynamicReturnTypeExtension for array_is_list to handle this special case.

Anyway, this PR should be ok to fix the issue 14177 so far.

src/Type/Constant/ConstantArrayType.php
```diff
@@ -823,7 +823,7 @@ public function unsetOffset(Type $offsetType): Type
 	 * So the nextAutoIndexes won't change, and the array might still be a list even with PHPStan definition.
 	 *
 	 * @param list<ConstantIntegerType|ConstantStringType> $newKeyTypes
-	 * @param int[] 									   $newOptionalKeys
+	 * @param int[] $newOptionalKeys
 	 */
 	private static function isListAfterUnset(array $newKeyTypes, array $newOptionalKeys, TrinaryLogic $arrayIsList, bool $unsetOptionalKey): TrinaryLogic
```
staabm, line 828: you might be able to kill the mutation with somethig like

https://phpstan.org/r/53bdb42a-d790-4576-aa47-62e33eb40865

VincentLanglet: I don't think so, it will be an union of two ConstantArrayType ; and each one will be handled separatly by this method.

Also, the tests are failing locally with the mutation ; dunno why it's reported as not caught then.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #5031, opened 2026-02-22
Tip about adding `@phpstan-impure` where applicable

ondrejmirtes:
Closes https://github.com/phpstan/phpstan/issues/14183

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -815,7 +815,7 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 			// narrowing affected the type (the cached value was narrowed).
 			assert($found instanceof Expr);
 			$scopeType = $this->getType($found);
-			$declaredReturnType = $holderExpr->getDeclaredReturnType();
+			$declaredReturnType = $holder->getType();
 			if ($declaredReturnType->isSuperTypeOf($scopeType)->yes() && $scopeType->isSuperTypeOf($declaredReturnType)->yes()) {
 				continue;
 			}
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -46,6 +46,7 @@ use PHPStan\Node\Expr\NativeTypeExpr;
 use PHPStan\Node\Expr\OriginalForeachKeyExpr;
 use PHPStan\Node\Expr\OriginalPropertyTypeExpr;
 use PHPStan\Node\Expr\ParameterVariableOriginalValueExpr;
+use PHPStan\Node\Expr\PossiblyImpureCallExpr;
 use PHPStan\Node\Expr\PropertyInitializationExpr;
 use PHPStan\Node\Expr\SetExistingOffsetValueTypeExpr;
 use PHPStan\Node\Expr\SetOffsetValueTypeExpr;
@@ -153,6 +154,7 @@ use function array_map;
 use function array_merge;
 use function array_pop;
 use function array_slice;
+use function array_unique;
 use function array_values;
 use function count;
 use function explode;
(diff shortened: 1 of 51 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -2918,6 +2921,20 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context, $sto
...
+			if (
+				$functionReflection !== null
+				&& $this->rememberPossiblyImpureFunctionValues
+				&& $functionReflection->hasSideEffects()->maybe()
+				&& !$functionReflection->isBuiltin()
+				&& $parametersAcceptor !== null
+			) {
+				$scope = $scope->assignExpression(
+					new PossiblyImpureCallExpr($normalizedExpr, $normalizedExpr, sprintf('%s()', $functionReflection->getName()), $parametersAcceptor->getReturnType()),
+					new MixedType(),
```
ondrejmirtes: Instead of the fourth PossiblyImpureCallExpr constructor parameter, this type could be carried as the expression type instead of `new MixedType()`.

src/Rules/Comparison/MatchExpressionRule.php
```diff
@@ -112,11 +115,14 @@ public function processNode(Node $node, Scope $scope): array
...
 					$armConditionScope->getType($armCondition->getCondition())->describe(VerbosityLevel::value()),
 				);
-				$errors[] = RuleErrorBuilder::message($message)
+				$errorBuilder = RuleErrorBuilder::message($message)
 					->line($armLine)
-					->identifier('match.alwaysTrue')
-					->tip('Remove remaining cases below this one and this error will disappear too.')
-					->build();
+					->identifier('match.alwaysTrue');
+				if (!$errorBuilder->hasTips()) {
```
ondrejmirtes: It's a new error builder, it's not going to have any tips on this line.

src/Rules/Comparison/PossiblyImpureTipHelper.php
```diff
@@ -0,0 +1,47 @@
...
+		Scope $scope,
+		Expr $conditionExpr,
+		RuleErrorBuilder $ruleErrorBuilder,
+	): RuleErrorBuilder
+	{
+		if (!$this->possiblyImpureTip) {
+			return $ruleErrorBuilder;
+		}
+		if ($ruleErrorBuilder->hasTips()) {
+			return $ruleErrorBuilder;
```
ondrejmirtes: We don't need this condition, it's okay to have more than one tip. I suspect we could remove the `hasTips` method declaration altogether.

ondrejmirtes (commented):
@phpstan-bot fix them please

ondrejmirtes (commented):
@phpstan-bot re-run `make cs`, `make tests` and `make phpstan` and fix the failures

src/Analyser/NodeScopeResolver.php
```diff
@@ -2918,6 +2921,20 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context, $sto
 				$scope = $scope->invalidateExpression(new Variable('this'), true);
 			}
 
+			if (
+				$functionReflection !== null
+				&& $this->rememberPossiblyImpureFunctionValues
+				&& $functionReflection->hasSideEffects()->maybe()
+				&& !$functionReflection->isBuiltin()
+				&& $parametersAcceptor !== null
```
staabm: this condition should be checked before `$functionReflection->hasSideEffects()->maybe()`

src/Analyser/NodeScopeResolver.php
```diff
@@ -3426,6 +3449,20 @@ function (MutatingScope $scope) use ($stmt, $expr, $nodeCallback, $context, $sto
 				&& $scope->getClassReflection()->is($methodReflection->getDeclaringClass()->getName())
 			) {
 				$scope = $scope->invalidateExpression(new Variable('this'), true, $methodReflection->getDeclaringClass());
+			} elseif (
+				$methodReflection !== null
+				&& $this->rememberPossiblyImpureFunctionValues
+				&& $methodReflection->hasSideEffects()->maybe()
+				&& !$methodReflection->getDeclaringClass()->isBuiltin()
+				&& $parametersAcceptor !== null
+				&& $scope->isInClass()
```
staabm, line 3473: these 2 lines should be moved above `$methodReflection->hasSideEffects()->maybe()`

ondrejmirtes:
@phpstan-bot ^^ fix what @staabm says

Also instead of changing the expectations in native-types-81.php, could you fix the closure cache key collision?

src/Analyser/MutatingScope.php
```diff
@@ -1050,6 +1050,10 @@ private function getClosureScopeCacheKey(): string
 			$parts[] = sprintf(',%s', $parameter->getType()->describe(VerbosityLevel::cache()));
 		}
 
+		if ($this->nativeTypesPromoted) {
```
ondrejmirtes: I don't think this fix is right. I think instead it should go over `$this->nativeExpressionTypes` same way it currently does over `$this->expressionTypes`.

ondrejmirtes (commented):
@phpstan-bot fix

staabm:
this should save us a lot of issue-duplicates and enduser support - awesome :-)

ondrejmirtes:
Last time I did this with `@template-covariant` it helped a lot. But some people will still miss the tip, which is okay, we can just point it out 😊

PR #5032, opened 2026-02-23
feat: add option to report ignores without comments

calebdw:
Hello!

Supersedes #3976 

Closes https://github.com/phpstan/phpstan/issues/11340

This PR adds the following option (false by default)
- `reportIgnoresWithoutComments`: reports `@phpstan-ignore` usages without comments

Thanks!

### DIFF
--- a/conf/config.neon
+++ b/conf/config.neon
@@ -118,6 +118,7 @@ parameters:
 		resolvedPhpDocBlockCacheCountMax: 2048
 		nameScopeMapMemoryCacheCountMax: 2048
 	reportUnmatchedIgnoredErrors: true
+	reportIgnoresWithoutComments: false
 	typeAliases: []
 	universalObjectCratesClasses:
 		- stdClass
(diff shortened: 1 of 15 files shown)


conf/config.neon
```diff
@@ -226,6 +227,7 @@ parameters:
 		- [parameters, errorFormat]
 		- [parameters, ignoreErrors]
 		- [parameters, reportUnmatchedIgnoredErrors]
+		- [parameters, reportIgnoresWithoutComments]
```
ondrejmirtes: You have to remove this

ondrejmirtes (changes requested):
Did you change any tests from the original PR, showing that it previously did not work? I'd like a special e2e/ test anyway (referenced from e2e-tests.yml, expecting an error).

calebdw:
> Did you change any tests from the original PR, showing that it previously did not work?

I added a new `ReportIgnoresWithoutCommentsIntegrationTest` that was failing with the original code and is now passing. The problem was that matched ignores were being removed before I was processing them and the original test did not account for this


I can add an e2e test

ondrejmirtes:
Yes, the E2E test is much better and shorter.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #5034, opened 2026-02-23
feat: add AnalysisResult::withFileSpecificErrors() to public API

janedbal:
Custom ErrorFormatter implementations that need to filter or modify file-specific errors currently have to call the `AnalysisResult` constructor directly, which is not part of the public API.

This adds an `@api` `withFileSpecificErrors()` method that returns a copy of the result with different file-specific errors, allowing extensions to create modified results without relying on internal constructor.

See example usecase: https://github.com/shipmonk-rnd/dead-code-detector/pull/292


ondrejmirtes:
Thank you!

PR #5035, opened 2026-02-23
Add non regression test

VincentLanglet:
Closes https://github.com/phpstan/phpstan/issues/1501

Dunno when it was fixed.


staabm (approved):
thanks

PR #5037, opened 2026-02-24
debug parallel-lint on windows

staabm:


staabm:
PR proved the thesis, [that our `make lint` command does not work at windows because too many arguments](https://github.com/ondrejmirtes/simple-downgrader/pull/22#issuecomment-3951083875)

PR #5038, opened 2026-02-24
Fix phpstan/phpstan#8744: Fix deprecated string interpolation syntax

codebymikey:
Adds support for flagging deprecated string interpolation syntaxes, as well as fixing them.

It involves fetching the token from the parser, and using that to infer the original syntax used in the source code.

Fixes phpstan/phpstan#8744

### DIFF
--- a/src/Rules/String/InterpolatedStringRule.php
+++ b/src/Rules/String/InterpolatedStringRule.php
@@ -8,10 +8,10 @@ use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Scalar\InterpolatedString;
 use PHPStan\Analyser\Scope;
 use PHPStan\DependencyInjection\RegisteredRule;
+use PHPStan\Parser\DeprecatedInterpolatedStringVisitor;
 use PHPStan\Php\PhpVersion;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
-use function assert;
 use function is_string;
 
 /**
@@ -38,17 +38,12 @@ final class InterpolatedStringRule implements Rule
 			return [];
 		}
 
-		$sourceTokens = $scope->getTokens();
+		if ($node->getAttribute(DeprecatedInterpolatedStringVisitor::ATTRIBUTE_NAME) !== true) {
+			return [];
+		}
+
 		foreach ($node->parts as $part) {
 			if (!$part instanceof Variable && !($part instanceof ArrayDimFetch && $part->var instanceof Variable)) {
-				continue;
-			}
-			$startTokenPos = $part->getStartTokenPos();
-			if (!isset($sourceTokens[$startTokenPos])) {
-				continue;
-			}
-			$startToken = (string) $sourceTokens[$startTokenPos];
-			if ($startToken !== '${') {
 				continue;
 			}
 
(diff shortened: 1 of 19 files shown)


ondrejmirtes:
I'm not fan of this approach of polluting all the APIs with parser tokens.

src/Rules/String/InterpolatedStringRule.php
```diff
@@ -0,0 +1,73 @@
...
+			}
+
+			if ($part instanceof ArrayDimFetch || ($part instanceof Variable && is_string($part->name))) {
+				$deprecatedMessage = 'Using ${var} in strings is deprecated in PHP 8.2. Use {$var} instead.';
+			} else {
+				$deprecatedMessage = 'Using ${expr} (variable variables) in strings is deprecated in PHP 8.2. Use {${expr}} instead.';
+			}
+
+			return [
+				RuleErrorBuilder::message($deprecatedMessage)
```
VincentLanglet, line 57: If it's a deprecation, and not a real error, should it be in deprecation-rules instead ?

codebymikey:
> I'm not fan of this approach of polluting all the APIs with parser tokens.

Wasn't sure of a better way to expose the tokens received by the parser since the parser interface is currently pretty simple.

And tokens are the only way I can think of to address it (that's how Rector [fixes it too](https://github.com/rectorphp/rector/blob/main/rules/Php82/Rector/Encapsed/VariableInStringInterpolationFixerRector.php))

Having access to the token as additional context seems like a general improvement for rules that might need it in the future.

Any thoughts on how to work around it without the API pollution? 

Unless only specific parsers have to implement a new `TokenAwareParser` interface. But I'm not familiar enough with the codebase to know which ones would require that additional interface and which ones should be left untouched, so just updated all the ones I was aware of.

src/Rules/String/InterpolatedStringRule.php
codebymikey: Potentially, my finding was that the rule needs access to the parser tokens in some form to check the original syntax.

So as long as that API is exposed in the main phpstan library, then the rule could be moved into deprecation-rules if that's the most sensible part for it.

src/Rules/String/InterpolatedStringRule.php
```diff
@@ -0,0 +1,73 @@
...
+		return InterpolatedString::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (!$this->phpVersion->deprecatesStringInterpolation()) {
+			return [];
+		}
+
+		$sourceTokens = $scope->getTokens();
```
staabm: could we instead add a new `NodeVisitor` which inspects the tokens at Ast-Parsing time and add necessary attributes onto the `InterpolatedString`-node, so we can just work on e.g. `$interpolatedString->getAttribute(...)` at rule level?

see other visitors in `src/Parser` which do similar things

codebymikey: That's a good idea, but I believe we only have access to the same Node object in the AST as we do in the custom rules.

edit: I see what you mean.

codebymikey:
Added a `TokenAwareVisitor` interface so that node visitors may now access tokens as necessary.

ondrejmirtes (changes requested):
Just marking a review so that I have a say in this before merge.

PR #5043, opened 2026-02-25
Faster LazyInternalScopeFactory->create()

staabm:
moving millions of method calls out of the hot path

grafik

---

similar to https://github.com/phpstan/phpstan-src/pull/4672

### DIFF
--- a/src/Analyser/LazyInternalScopeFactory.php
+++ b/src/Analyser/LazyInternalScopeFactory.php
@@ -25,9 +25,9 @@ final class LazyInternalScopeFactory implements InternalScopeFactory
 {
 
 	/** @var int|array{min: int, max: int}|null */
-	private readonly int|array|null $phpVersion;
+	private int|array|null $phpVersion;
 
-	private ?Parser $currentSimpleVersionParser = null;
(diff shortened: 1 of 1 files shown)


src/Analyser/LazyInternalScopeFactory.php
```diff
@@ -38,6 +64,21 @@ public function __construct(
 	{
 		$this->phpVersion = $this->container->getParameter('phpVersion');
 		$this->currentSimpleVersionParser = $this->container->getService('currentPhpVersionSimpleParser');
+
+		$this->reflectionProvider = $this->container->getByType(ReflectionProvider::class);
```
ondrejmirtes: Again, a little bit worried about doing this in the constructor

ondrejmirtes (changes requested):

src/Analyser/LazyInternalScopeFactory.php
staabm: args, sorry :-/

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #5044, opened 2026-02-25
Refactor claude comment trigger logic

takaram:
This PR removes `check-trigger` job in `claude-react-on-comment` workflow.

Precondition check is done by `if` expression of `react` job.
This will reduce the overhead of the check-trigger job.


ondrejmirtes:
Hi, thank you, but I need to completely rework this because of security, I'm not sure if this workflow will even be bere in this form.

PR #5046, opened 2026-02-25
Do not parse PR body to find issues to close

takaram:
Extracting linked issues from PR body is very hard as there are variety of syntaxes to link an issue. 
https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Especially some PRs use `phpstan/phpstan#<issue number>` format instead of the issue URL.

We should use `closingIssuesReferences` instead.

### DIFF
--- a/.github/workflows/close-issues-on-merge.yml
+++ b/.github/workflows/close-issues-on-merge.yml
@@ -27,18 +27,17 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.PHPSTAN_BOT_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I wouldn't be doing this if it wasn't necessary.  I'm aware of `closingIssuesReferences` but it doesn't work when being merged on non-default branch. Even the API returns an empty array.

takaram:
> it doesn't work when being merged on non-default branch. Even the API returns an empty array.

Ah, I misunderstood. Sorry!

PR #5047, opened 2026-02-26
More precise array-item types in loops

staabm:


src/Analyser/MutatingScope.php
```diff
@@ -1853,6 +1853,10 @@ private function getTypeFromArrayDimFetch(
 			throw new ShouldNotHappenException();
 		}
 
+		if ($offsetAccessibleType instanceof NeverType) {
+			return $offsetAccessibleType;
+		}
+
 		if (!$offsetAccessibleType->isArray()->yes() && (new ObjectType(ArrayAccess::class))->isSuperTypeOf($offsetAccessibleType)->yes()) {
```
staabm, line 1860: before this PR we ran into this `offsetGet`-method-call logic which turned the `Never` into a `Error`.

staabm:
regarding issue-bot errors: 

the first error is related to sealed-arrays, which we do not yet support.

and I think the remaining ones are related to explicit `mixed` parameter type which is not valid on PHP7.x and therefore the errors look a bit weird.

ondrejmirtes:
Thank you!

PR #5049, opened 2026-02-26
Fix slow analysis on return with a big OR condition

staabm:
Closes https://github.com/phpstan/phpstan/issues/14207

Regression test before PR:
```
➜  phpstan-src git:(2.1.x) ✗ time vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php  

Runtime:       PHP 8.3.30
Configuration: /Users/m.staab/dvl/phpstan-src/phpunit.xml
Random Seed:   1772116653

...............................................................  63 / 212 ( 29%)
............................................................... 126 / 212 ( 59%)
............................................................... 189 / 212 ( 89%)
....S..................                                         212 / 212 (100%)

Time: 00:25.512, Memory: 330.50 MB
```

After PR (equal fast as `AnalyserIntegrationTest` on 2.1.x):
```
➜  phpstan-src git:(bug-14207-2) ✗ time vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php  

Runtime:       PHP 8.3.30
Configuration: /Users/m.staab/dvl/phpstan-src/phpunit.xml
Random Seed:   1772116718

...............................................................  63 / 211 ( 29%)
............................................................... 126 / 211 ( 59%)
......S........................................................ 189 / 211 ( 89%)
......................                                          211 / 211 (100%)

Time: 00:17.420, Memory: 314.50 MB
```

----

Wordpress `phpstan analyse --debug src/wp-includes/html-api/class-wp-html-processor.php -v` before PR:
```
Elapsed time: 8.61 seconds
```

after PR:
```
Elapsed time: 1.69 seconds
```

### DIFF
--- a/tests/PHPStan/Analyser/data/bug-14207.php
+++ b/tests/PHPStan/Analyser/data/bug-14207.php
@@ -31,6 +31,8 @@ class HelloWorld
 				: "{$tag_name->namespace} {$tag_name->node_name}";
 		}
 
+		assertType('non-falsy-string|uppercase-string', $tag_name);
+
 		$x = (
 			'ADDRESS' === $tag_name ||
 			'APPLET' === $tag_name ||
@@ -130,7 +132,7 @@ class HelloWorld
 			'svg TITLE' === $tag_name
 		);
 
-		assertType('non-falsy-string|uppercase-string', $tag_name);
+		assertType('bool', $x);
 
 		return $x;
 	}
--- a/tests/PHPStan/Analyser/data/bug-14207.php
+++ b/tests/PHPStan/Analyser/data/bug-14207.php
@@ -1,0 +1,137 @@
+<?php declare(strict_types = 1);
+
+namespace Bug14207;
+
+use function assert;
+use function PHPStan\Testing\assertType;
+
+class WP_HTML_Token {
+	public string $namespace;
+	public string $node_name;
+}
+
+class HelloWorld
+{
+	/**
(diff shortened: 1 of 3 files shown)


ondrejmirtes (changes requested):
Reuse it for nested BooleanAnd as well

staabm:
> Reuse it for nested BooleanAnd as well

I can add a test for bool-AND but I don't think we need the fix for AND, because AND will narrow the type, not widen as OR does (so it will not produce a big union)

ondrejmirtes:
Yeah, only if you produce an example with BooleanAnd that’s slow.

tests/PHPStan/Analyser/data/bug-14207.php
```diff
@@ -0,0 +1,139 @@
...
+	public static function is_special( $tag_name ): bool {
+		if ( is_string( $tag_name ) ) {
+			$tag_name = strtoupper( $tag_name );
+		} else {
+			$tag_name = 'html' === $tag_name->namespace
+				? strtoupper( $tag_name->node_name )
+				: "{$tag_name->namespace} {$tag_name->node_name}";
+		}
+
+		assertType('non-falsy-string|uppercase-string', $tag_name);
```
staabm, line 34: this type is not correct, but its a pre-existing unrelated problem: [created a separate issue for that](https://github.com/phpstan/phpstan/issues/14208)

staabm:
the failling tests make me feel the fix is wrong

PR #5052, opened 2026-02-26
POC force list in ConstantArrayTypeBuilder

VincentLanglet:
I feel like there is some inconsistency about how constant list are handled.
In someplace we're generating a ConstantArray(isList = maybe) and we're adding the AccessoryArrayListType and in some other place we're generating a ConstantArray(isList = true) without the accessory.

The second one is better, since the logic is well handled in the constantArray while the first one can gives weird behavior because we're not handling similar logic in the UnionType (like the fact that a constant array with optional keys and being a list implying some conditions).

The ConstantArrayTypeBuilder, when building the array is often loosing the certainty of the list when adding optional keys (as soon as the second one is added), so I feel like we should have a way to re-force it.

This could be used here too https://github.com/phpstan/phpstan-src/blob/93e6978b46e8c5be3a75210b6bfecc7d56d4f2eb/src/Type/Php/RegexArrayShapeMatcher.php#L298-L304 and certainly at some others place.

But first I'd like your opinion on the idea + how the ConstantArrayBuilder API should change:
- A new method ?
- A new param to getArray ?

But looking at the tests I have an issue:
`list{0?: string, 1?: string, 2?: string, 3?: string}` is `ConstantArray(isList = true)&AccessoryIsList` which is simplified to `ConstantArray(isList = true)` by TypeCombinator::intersect (because the accessory is superTypeOf ConstantArray), so it's displayed `array{0?: string, 1?: string, 2?: string, 3?: string}`.

I tried:
- modifying ConstantArray(isList = true)::describe but at least ~400 tests are failing because it changes lot of error message (with array like `array{1, 2, 3}` becoming `list{1, 2, 3}`).

- modifying TypeCombinator::intersect to keep AccessoryArrayListType, less tests are failing, but it's still hard to avoid changing existing error messages ; I didn't find a way yet.

I dunno if it inspired you @ondrejmirtes @staabm or if I just abandon this for now

### DIFF
--- a/src/PhpDoc/TypeNodeResolver.php
+++ b/src/PhpDoc/TypeNodeResolver.php
@@ -1060,13 +1060,15 @@ final class TypeNodeResolver
 			$builder->setOffsetValueType($offsetType, $this->resolve($itemNode->valueType, $nameScope), $itemNode->optional);
 		}
 
-		$arrayType = $builder->getArray();
-
-		$accessories = [];
-		if (in_array($typeNode->kind, [
+		$isList = in_array($typeNode->kind, [
 			ArrayShapeNode::KIND_LIST,
 			ArrayShapeNode::KIND_NON_EMPTY_LIST,
-		], true)) {
+		], true);
+
+		$arrayType = $isList ? $builder->getList() : $builder->getArray();
+
+		$accessories = [];
+		if ($isList) {
 			$accessories[] = new AccessoryArrayListType();
 		}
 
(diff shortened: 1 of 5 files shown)


VincentLanglet:
A good example about why it's hard to improve the situation without changing the description of the arrays are
https://phpstan.org/r/7e003700-8a20-4a5e-b166-4a966cf9949f

Currently there is a list 
```
list{0: string, 1: string, 2?: string}
```
can also be named
```
array{0: string, 1: string, 2?: string}
```
while still being a list, depending of the way it's listed.

PR #5053, opened 2026-02-27
Fix slow analysis on return with a big OR condition

staabm:
Closes https://github.com/phpstan/phpstan/issues/14207

Regression test before PR:
```
➜  phpstan-src git:(2.1.x) ✗ time vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php  

Runtime:       PHP 8.3.30
Configuration: /Users/m.staab/dvl/phpstan-src/phpunit.xml
Random Seed:   1772116653

...............................................................  63 / 212 ( 29%)
............................................................... 126 / 212 ( 59%)
............................................................... 189 / 212 ( 89%)
....S..................                                         212 / 212 (100%)

Time: 00:27.512, Memory: 334.50 MB
```

After PR (equal fast as `AnalyserIntegrationTest` on 2.1.x):
```
➜  phpstan-src git:(bug-14207-2) ✗ time vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php  

Runtime:       PHP 8.3.30
Configuration: /Users/m.staab/dvl/phpstan-src/phpunit.xml
Random Seed:   1772116718

...............................................................  63 / 211 ( 29%)
............................................................... 126 / 211 ( 59%)
......S........................................................ 189 / 211 ( 89%)
......................                                          211 / 211 (100%)

Time: 00:17.676, Memory: 326.50 MB
```

----

Wordpress `phpstan analyse --debug src/wp-includes/html-api/class-wp-html-processor.php -v` before PR:
```
Elapsed time: 8.61 seconds
```

after PR:
```
Elapsed time: 3.89 seconds
```

### DIFF
--- a/src/Analyser/SpecifiedTypes.php
+++ b/src/Analyser/SpecifiedTypes.php
@@ -3,15 +3,16 @@
 namespace PHPStan\Analyser;
 
 use PhpParser\Node\Expr;
-use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
 use function array_key_exists;
 use function array_merge;
+use function count;
 
 final class SpecifiedTypes
 {
+
 	private const MAX_SURE_NOT_TYPES = 8;
 
 	private bool $overwrite = false;
--- a/src/Analyser/SpecifiedTypes.php
+++ b/src/Analyser/SpecifiedTypes.php
@@ -13,7 +13,7 @@ use function count;
 final class SpecifiedTypes
 {
 
-	private const MAX_SURE_NOT_TYPES = 8;
+	private const MAX_ADDED_TYPES = 8;
 
 	private bool $overwrite = false;
 
@@ -172,6 +172,10 @@ final class SpecifiedTypes
 				continue;
 			}
 
+			if ($type instanceof UnionType && count($type->getTypes()) >= self::MAX_ADDED_TYPES) {
+				continue;
+			}
+
 			$sureTypeUnion[$exprString] = [
 				$exprNode,
 				TypeCombinator::intersect($type, $other->sureTypes[$exprString][1]),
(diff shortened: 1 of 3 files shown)


staabm:
what I like about this fix
- its ~60% faster than before
- its a simple fix

what I don't like about this fix
- the repro still in the 3-4 second range of analysis time for a single file (since the file has ~6600 lines of code, this might be acceptable though)

staabm:
the issue-bot result seems outdated, as I get the same diff when I [open the playground link](https://phpstan.org/r/5efc781a-6381-4d1b-ac96-cd96f373ece4) and click "Show the latest result" (so the diff the bot reports with this PR, is already on 2.1.x)

src/Analyser/SpecifiedTypes.php
```diff
@@ -179,6 +183,10 @@ public function unionWith(SpecifiedTypes $other): self
 				continue;
 			}
 
+			if ($type instanceof UnionType && count($type->getTypes()) >= self::MAX_SURE_NOT_TYPES) {
```
ondrejmirtes: I'd limit the sureTypes the same way, just to be consistent and find possible issues with this approach

ondrejmirtes (changes requested):

src/Analyser/SpecifiedTypes.php
staabm: so you mean we should generalize a sure-type when this limit is hit?
not beeing precise about what we not know has less impact on the overall analysis as generalizing what we know though.

ondrejmirtes: I just meant to copy:

```php

			if ($type instanceof UnionType && count($type->getTypes()) >= self::MAX_SURE_NOT_TYPES) { 
				continue;
			}
```

in the `foreach ($this->sureTypes...` above.

staabm: ok - did just that. I don't understand why so ;-)

ondrejmirtes: The two arrays have the same importance. I feel like what we do to one we should do to the other.

If some tests start failing, it means the fix is wrong for both `sureTypes` and `sureNotTypes`.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/pull/5076

PR #5054, opened 2026-02-27
Prevent unnecessary work in IntersectionType

staabm:
return early when possible.

before this PR we built the whole result, merging part results and interpreting the combined overall result in the end.
we can save unnecessary work exiting early in case one of the part results are YES (similar how its done in `TrinaryLogic::maxMin`).

---

Wordpress repro

2.1.x
```
➜  wordpress-develop git:(trunk) hyperfine 'php ../../bin/phpstan analyse --debug src/wp-includes/html-api/class-wp-html-processor.php -v'
Benchmark 1: php ../../bin/phpstan analyse --debug src/wp-includes/html-api/class-wp-html-processor.php -v
  Time (mean ± σ):      8.962 s ±  0.061 s    [User: 8.224 s, System: 0.366 s]
  Range (min … max):    8.880 s …  9.060 s    10 runs
```

this PR
```
➜  wordpress-develop git:(trunk) hyperfine 'php ../../bin/phpstan analyse --debug src/wp-includes/html-api/class-wp-html-processor.php -v'
Benchmark 1: php ../../bin/phpstan analyse --debug src/wp-includes/html-api/class-wp-html-processor.php -v
  Time (mean ± σ):      8.647 s ±  0.112 s    [User: 7.864 s, System: 0.368 s]
  Range (min … max):    8.548 s …  8.932 s    10 runs
```

----

I think similar things can be done for union

### DIFF
--- a/src/Type/AcceptsResult.php
+++ b/src/Type/AcceptsResult.php
@@ -152,4 +152,25 @@ final class AcceptsResult
 		return new self(TrinaryLogic::maxMin(...$results), array_values(array_unique($reasons)));
 	}
 
+	/**
+	 * @template T
+	 * @param T[] $objects
+	 * @param callable(T): self $callback
+	 */
+	public static function lazyMaxMin(
+		array $objects,
+		callable $callback,
+	): self
+	{
+		$results = [];
+		foreach ($objects as $object) {
+			$isAcceptedBy = $callback($object);
+			if ($isAcceptedBy->result->yes()) {
+				return $isAcceptedBy;
+			}
+			$results[] = $isAcceptedBy;
+		}
+		return self::maxMin(...$results);
+	}
+
 }
--- a/src/Type/AcceptsResult.php
+++ b/src/Type/AcceptsResult.php
@@ -163,14 +163,26 @@ final class AcceptsResult
 	): self
 	{
 		$results = [];
+		$reasons = [];
+		$hasNo = false;
 		foreach ($objects as $object) {
 			$isAcceptedBy = $callback($object);
 			if ($isAcceptedBy->result->yes()) {
 				return $isAcceptedBy;
+			} elseif ($isAcceptedBy->result->no()) {
+				$hasNo = true;
 			}
 			$results[] = $isAcceptedBy;
+
+			foreach ($isAcceptedBy->reasons as $reason) {
+				$reasons[] = $reason;
+			}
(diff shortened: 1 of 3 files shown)


VincentLanglet (commented):
TrinaryLogic has lazyMaxMin, isn't the right moment to introduce IsSuperTypeOfResult::lazy method ? (And maybe the same for AcceptsResult ?)

staabm:
> TrinaryLogic has lazyMaxMin, isn't the right moment to introduce IsSuperTypeOfResult::lazy method ? (And maybe the same for AcceptsResult ?)

good idea, adjusted accordingly.
(I did not do it in the first place, because there is only a single use-case for maxMin() and this one is now the only use case for lazyMaxMin)

src/Type/AcceptsResult.php
```diff
@@ -152,4 +152,25 @@ public static function maxMin(self ...$operands): self
...
+	{
+		$results = [];
+		foreach ($objects as $object) {
+			$isAcceptedBy = $callback($object);
+			if ($isAcceptedBy->result->yes()) {
+				return $isAcceptedBy;
+			}
+			$results[] = $isAcceptedBy;
+		}
+		return self::maxMin(...$results);
```
VincentLanglet: This add another foreach.
Looking at `TrinaryLogic::lazyMaxMin` it does not call self::maxMin.

Should we implement it this way
```
public static function lazyMaxMin(
    array $objects,
    callable $callback,
): self
{
    if ($objects === []) {
    }

    $results = [];
    $reasons = [];
    foreach ($objects as $object) {
        $isAcceptedBy = $callback($object);
        if ($isAcceptedBy->result->yes()) {
            return $isAcceptedBy;
        }
        $results[] = $isAcceptedBy->result;
        foreach ($isAcceptedBy->reasons as $reason) {
            $reasons[] = $reason;
        }
    }

    return new self(TrinaryLogic::maxMin(...$results), array_values(array_unique($reasons)));
}
```
instead ?

(I also wonder if we should -- now or later -- avoid calling `TrinaryLogic::maxMin` to remove an extra foreach and doing it in the existing one)

staabm: ok - unrolled all nested loops now

VincentLanglet (approved):

staabm:
Thank you!

PR #5067, opened 2026-02-27
Improve intersection of ConstantArray and AccessoryIsList

VincentLanglet:
Proof of the gain in the next PR https://github.com/phpstan/phpstan-src/pull/5069

### DIFF
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -715,6 +715,14 @@ class ConstantArrayType implements Type
 		return $builder->getArray();
 	}
 
+	/**
+	 * Removes or marks as optional the key(s) matching the given offset type from this constant array.
+	 *
+	 * By default, the method assumes an actual `unset()` call was made, which actively modifies the
+	 * array and weakens its list certainty to "maybe". However, in some contexts, such as the else
+	 * branch of an array_key_exists() check, the key is statically known to be absent without any
+	 * modification, so list certainty should be preserved as-is.
+	 */
 	public function unsetOffset(Type $offsetType, bool $preserveListCertainty = false): Type
 	{
 		$offsetType = $offsetType->toArrayKey();
--- a/src/Type/Constant/ConstantArrayType.php
+++ b/src/Type/Constant/ConstantArrayType.php
@@ -748,8 +748,12 @@ class ConstantArrayType implements Type
 					$newOptionalKeys,
 					$this->isList,
 					in_array($i, $this->optionalKeys, true),
-					$preserveListCertainty,
 				);
+				if (!$preserveListCertainty) {
+					$newIsList = $newIsList->and(TrinaryLogic::createMaybe());
+				} elseif ($this->isList->yes() && $newIsList->no()) {
+					return new NeverType();
+				}
 
 				return new self($newKeyTypes, $newValueTypes, $this->nextAutoIndexes, $newOptionalKeys, $newIsList);
 			}
@@ -791,8 +795,12 @@ class ConstantArrayType implements Type
 				$optionalKeys,
 				$this->isList,
 				count($optionalKeys) === count($this->optionalKeys),
-				$preserveListCertainty,
 			);
+			if (!$preserveListCertainty) {
+				$newIsList = $newIsList->and(TrinaryLogic::createMaybe());
+			} elseif ($this->isList->yes() && $newIsList->no()) {
+				return new NeverType();
+			}
 
 			return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, $newIsList);
 		}
@@ -817,8 +825,12 @@ class ConstantArrayType implements Type
 			$optionalKeys,
 			$this->isList,
 			count($optionalKeys) === count($this->optionalKeys),
-			$preserveListCertainty,
 		);
+		if (!$preserveListCertainty) {
+			$newIsList = $newIsList->and(TrinaryLogic::createMaybe());
+		} elseif ($this->isList->yes() && $newIsList->no()) {
+			return new NeverType();
+		}
 
 		return new self($this->keyTypes, $this->valueTypes, $this->nextAutoIndexes, $optionalKeys, $newIsList);
 	}
@@ -830,7 +842,7 @@ class ConstantArrayType implements Type
 	 * @param list<ConstantIntegerType|ConstantStringType> $newKeyTypes
 	 * @param int[] $newOptionalKeys
 	 */
-	private static function isListAfterUnset(array $newKeyTypes, array $newOptionalKeys, TrinaryLogic $arrayIsList, bool $unsetOptionalKey, bool $preserveListCertainty): TrinaryLogic
+	private static function isListAfterUnset(array $newKeyTypes, array $newOptionalKeys, TrinaryLogic $arrayIsList, bool $unsetOptionalKey): TrinaryLogic
 	{
 		if (!$unsetOptionalKey || $arrayIsList->no()) {
 			return TrinaryLogic::createNo();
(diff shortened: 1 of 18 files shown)


src/Type/TypeCombinator.php
```diff
@@ -1331,6 +1331,20 @@ public static function intersect(Type ...$types): Type
 						continue 2;
 					}
 
+					if ($types[$i] instanceof ConstantArrayType && $types[$j] instanceof AccessoryArrayListType) {
```
VincentLanglet, line 1334: The main idea of this PR.

Before, we had sometime a ConstantArray(isList = true) and sometimes ConstantArray(isList=maybe)&AccessoryArrayListType ; now I merged them to always have ConstantArray(isList = true).

src/Type/Constant/ConstantArrayType.php
```diff
@@ -1570,18 +1575,36 @@ public function describe(VerbosityLevel $level): string
...
 		};
 		return $level->handle(
-			fn (): string => $this->isIterableAtLeastOnce()->no() ? 'array' : sprintf('array<%s, %s>', $this->getIterableKeyType()->describe($level), $this->getIterableValueType()->describe($level)),
+			fn (): string => $this->isIterableAtLeastOnce()->no() ? $arrayName : sprintf('%s<%s, %s>', $arrayName, $this->getIterableKeyType()->describe($level), $this->getIterableValueType()->describe($level)),
 			static fn (): string => $describeValue(true),
 			static fn (): string => $describeValue(false),
 		);
 	}
 
+	private function shouldBeDescribedAsAList(): bool
```
VincentLanglet, line 1611: Since ConstantArray&AccessoryArrayIsList is simplified into ConstantArray, I always get a description like
```
array{...}
````

This is annoying when having too much optional keys.

I tried to find the formula to change as few description as possible:
- array like `array{0: string, 1: string}` without optional keys are untouched because they are _always_ a list
- same for array like `array{0: string, 1: string, 2?: string}` where the optional keys is the last one
- `list{0: string, 1?: string, 2: string}` should stay a list (in the next PR I'll make the offset 1 required)
- `list{0: string, 1?: string, 2?: string}` with 2+ optional keys should stay a list.

So we cannot only rely on IntersectionType to add the `list` keyWord.

File: src/Type/IntersectionType.php [hunk 3]
```diff
@@ -448,7 +449,8 @@ private function describeItself(VerbosityLevel $level, bool $skipAccessoryTypes)
 					continue;
 				} elseif ($type instanceof ConstantArrayType) {
 					$description = $type->describe($level);
-					$descriptionWithoutKind = substr($description, strlen('array'));
+					$kind = str_starts_with($description, 'list') ? 'list' : 'array';
+					$descriptionWithoutKind = substr($description, strlen($kind));
```
VincentLanglet, line 453: Now constant array is not always a `array{...}` and strlen of `list` and `array` are different so I need to check it.

tests/PHPStan/Type/TypeToPhpDocNodeTest.php
```diff
@@ -379,25 +380,14 @@ public static function dataToPhpDocNode(): iterable
...
-			]),
-			'list{0: string, 1: string, 2?: string, 3?: string}',
-		];
+		], [3], [2, 3], TrinaryLogic::createYes());
 
 		yield [
-			new IntersectionType([
-				$constantArrayWithOptionalKeys,
-				new AccessoryArrayListType(),
-			]),
```
VincentLanglet, line 396: This is technically not a valid intersection anymore as it will be simplified into a constant array with `isList = Yes` by TypeCombinator::intersect.

So the test was failing because
- this type was transformed into `list{0: string, 1: string, 2?: string, 3?: string}`.
- this string was parsed into a ConstantArray (isList = Yes) without the need of an intersectionType.

I updated the test to reproduce the "new real case".

src/Type/Constant/ConstantArrayType.php
```diff
@@ -715,7 +715,7 @@ public function setExistingOffsetValueType(Type $offsetType, Type $valueType): T
 		return $builder->getArray();
 	}
 
-	public function unsetOffset(Type $offsetType): Type
+	public function unsetOffset(Type $offsetType, bool $preserveListCertainty = false): Type
```
VincentLanglet, line 726: Kinda related to https://github.com/phpstan/phpstan-src/pull/5029 ; this is an idea I already had in mind. Without this fix tests are failing because we don't have the AccessoryIsList anymore (already merged into the constant array).

This method is used in two different context:
- when unsetting the value 
- during something like
```
if (array_key_exist(...)) {

} else {
     // Here we're trying to remove the array key, and calling unsetOffset
}
```

For the second example, even if the key is "unset", shouldn't touch the list certainty since we didn't really modified the array. That's why tryRemove will call with 'true'.
```
if ($typeToRemove instanceof HasOffsetType) {
			return $this->unsetOffset($typeToRemove->getOffsetType(), true);
		}
```

VincentLanglet:
This solve https://github.com/phpstan/phpstan/issues/11602#issuecomment-3968119733 but not the whole issue.

src/Type/TypeCombinator.php
staabm: I wonder why these 2 types of "isList" exist in the first place. could this always be using `AccessoryArrayListType`?

VincentLanglet: ConstantArray need to implement `isList`, if you return `Maybe` for this method, you'll end up with too many test failure ; having to intersect with AccessoryArrayListType every time, even when doing a basic `$a = []`.

Also, there is some optimisation in the ConstantArray possible when we know it's a list. It would be harder to do such thing in the intersection type. And that's the whole purpose of this PR ; currently some optimisation are ignored because the array doesn't know it's a list. And some other optimisation are possible.

src/Type/Constant/ConstantArrayType.php
staabm: I think this needs a phpdoc for this parameter, because intuitively unsetting on a list will always destroy the list

VincentLanglet: I added
``` 
/**
	 * Removes or marks as optional the key(s) matching the given offset type from this constant array.
	 *
	 * By default, the method assumes an actual `unset()` call was made, which actively modifies the
	 * array and weakens its list certainty to "maybe". However, in some contexts, such as the else
	 * branch of an array_key_exists() check, the key is statically known to be absent without any
	 * modification, so list certainty should be preserved as-is.
	 */
```

tests/PHPStan/Analyser/nsrt/array-chunk.php
```diff
@@ -49,9 +49,9 @@ public function constantArraysWithOptionalKeys(array $arr): void
 	 */
 	public function chunkUnionTypeLength(array $arr, $positiveRange, $positiveUnion) {
 		/** @var array{a: 0, b?: 1, c: 2} $arr */
-		assertType('array{0: array{0: 0, 1?: 1|2, 2?: 2}, 1?: array{0?: 2}}', array_chunk($arr, $positiveRange));
+		assertType('array{0: list{0: 0, 1?: 1|2, 2?: 2}, 1?: array{0?: 2}}', array_chunk($arr, $positiveRange));
```
VincentLanglet, line 52: This is some bonus of this PR.

Before a ConstantArray(isList = true) was described as `array` in every occasion.

Now I describe it as list in some conditions, which gives a meaningful type.

As an example here array_is_list would have a return true for `array{0: 0, 1?: 1|2, 2?: 2}` but this would not have been clear for the user why.
See https://phpstan.org/r/a82c93cb-e834-4fd9-8c01-6bd06446ec04

staabm (approved):
I think this makes sense. Ondrej should have another look though.

VincentLanglet:
> I think this makes sense. Ondrej should have another look though.

I agree ; please take a look @ondrejmirtes

ondrejmirtes:
Thank you! This is how I'd approach it too 👍

PR #5094, opened 2026-02-28
Fix seemingly same type reports as invalid/incompatible.

VincentLanglet:
Rework of https://github.com/phpstan/phpstan-src/pull/5073

Closes https://github.com/phpstan/phpstan/issues/13440
Closes https://github.com/phpstan/phpstan/issues/12688

The issue was that one closure was changed into Closure(template-strict-mixed), while the other stayed Closure(template-mixed) and therefor they were not equal.

Instead of just transforming common in return type, we should transform them in parameter and return types.
Best way to do it it's to transform common type on the closure. (Cf commit 2)

And then when I thought again about it, I think there is some loss on calling transformCommonType inside the `TypeTraverser::map` call since we might call in multiple times when traversing the types. The first one will transform the mixed to strict mixed, but other call will be useless.
So it's seems better to just call it once on the whole accepted type at the end.

### DIFF
--- a/src/Rules/RuleLevelHelper.php
+++ b/src/Rules/RuleLevelHelper.php
@@ -93,15 +93,15 @@ final class RuleLevelHelper
 					return $acceptedType;
 				}
 
-				return new CallableType(
+				return $this->transformCommonType(new CallableType(
 					$acceptedType->getParameters(),
-					$traverse($this->transformCommonType($acceptedType->getReturnType())),
+					$traverse($acceptedType->getReturnType()),
 					$acceptedType->isVariadic(),
 					$acceptedType->getTemplateTypeMap(),
 					$acceptedType->getResolvedTemplateTypeMap(),
 					$acceptedType->getTemplateTags(),
 					$acceptedType->isPure(),
-				);
+				));
 			}
 
 			if ($acceptedType instanceof ClosureType) {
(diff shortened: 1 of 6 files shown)


src/Rules/RuleLevelHelper.php
```diff
@@ -95,7 +95,7 @@ private function transformAcceptedType(Type $acceptingType, Type $acceptedType):
 
 				return new CallableType(
 					$acceptedType->getParameters(),
-					$traverse($this->transformCommonType($acceptedType->getReturnType())),
```
staabm, line 98: Do we need another test for callable?

staabm:
> Instead of just transforming common in return type, we should transform them in parameter and return types.

What about throws type?

VincentLanglet:
> > Instead of just transforming common in return type, we should transform them in parameter and return types.
> 
> What about throws type?

ClosureType::traverse only touch param and return type.

And it kinda make sens here since we're interesting removing null or transforming mixed into strict ; two things which does not make sens for throw type.

src/Rules/RuleLevelHelper.php
VincentLanglet: Added

staabm:
this seems to also fix https://github.com/phpstan/phpstan/issues/12688

staabm (approved):

PR #5095, opened 2026-02-28
AcceptsResult: cache Yes/No/Maybe

VincentLanglet:
I found some useless variable in lazyMaxMin, and then I found that you cached Yes/no/maybe in SuperTypeOf but not AcceptResult @staabm. I think we can do the same.

### DIFF
--- a/src/Type/AcceptsResult.php
+++ b/src/Type/AcceptsResult.php
@@ -28,6 +28,12 @@ use function array_values;
 final class AcceptsResult
 {
 
+	private static self $YES;
+
+	private static self $MAYBE;
+
+	private static self $NO;
+
 	/**
 	 * @api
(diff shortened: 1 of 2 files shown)


staabm (approved):
Thanks

src/Type/AcceptsResult.php
```diff
@@ -162,7 +174,6 @@ public static function lazyMaxMin(
 		callable $callback,
 	): self
 	{
-		$results = [];
```
staabm, line 165: Would be nice phpstan would detect local variables which are only written but never read

ondrejmirtes: There are feature requests for that. It'd be very easy to implement too I think. We'd need virtual expr nodes like PropertyInitializationExpr to track things in scope.

PR #5110, opened 2026-02-28
Add non regression tests

VincentLanglet:
Related to https://github.com/phpstan/phpstan-src/pull/5105

### DIFF
--- a/tests/PHPStan/Rules/Exceptions/data/pr-5105.php
+++ b/tests/PHPStan/Rules/Exceptions/data/pr-5105.php
@@ -1,0 +1,32 @@
+<?php // lint >= 8.3
+
+declare(strict_types = 1);
+
+namespace Pr5105;
+
+enum ReactionType: string
+{
+	case EMOJI_HEART = '❤️';
+
+	public static function tryFromName1(string $name): ?self
+	{
+		try {
+			return ReactionType::{$name};
+		} catch (\RuntimeException) {
+			return null;
+		}
+	}
+
+	public static function tryFromName2(string $name): ?self
+	{
+		if ($name !== 'EMOJI_HEART') {
+			return null;
+		}
+
+		try {
+			return ReactionType::{$name};
+		} catch (\Error) {
+			return null;
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Rules/Exceptions/data/pr-5105.php
```diff
@@ -0,0 +1,32 @@
...
+
+enum ReactionType: string
+{
+	case EMOJI_HEART = '❤️';
+
+	public static function tryFromName1(string $name): ?self
+	{
+		try {
+			return ReactionType::{$name};
+		} catch (\RuntimeException) {
```
staabm, line 15:
```suggestion
		} catch (\RuntimeException) { // will throw ERROR on invalid $name
```

tests/PHPStan/Rules/Exceptions/data/pr-5105.php
```diff
@@ -0,0 +1,32 @@
...
+	{
+		if ($name !== 'EMOJI_HEART') {
+			return null;
+		}
+
+		try {
+			return ReactionType::{$name};
+		} catch (\Error) {
+			return null;
+		}
```
staabm, line 30: `\Error` can still be thrown, see 
- https://3v4l.org/tKMaZ#veol
- https://3v4l.org/OsPNS#veol

VincentLanglet: Not after the condition line 22 ?

staabm: you are right

staabm: you are right.

---

maybe we should add a test for

```
	public static function tryFromInvalid(string $name): ?self
	{
		try {
			return ReactionType::XYZ;
		} catch (\Error $e) {
			return null;
		}
	}
```

VincentLanglet: Currently it's not supported
https://phpstan.org/r/2b645c07-0bae-4dec-b0f1-1ce3012b0e8e

And it's like any method call
```
try {
			return self::acceptsString(true);
		} catch (\Error $e) {
			return null;
		}
```
we're currently reporting the method call as an error but also reporting the dead catch (if implictThrow = false)

staabm (approved):

PR #5114, opened 2026-03-01
Fix phpstan/phpstan#13526: False positive with array_key_exists and union types

staabm:
Fixes phpstan/phpstan#13526

Related to https://github.com/phpstan/phpstan-src/pull/3933

### DIFF
--- a/src/Analyser/MutatingScope.php
+++ b/src/Analyser/MutatingScope.php
@@ -3680,10 +3680,6 @@ class MutatingScope implements Scope, NodeCallbackInvoker
 
 		if ($originalExprType->equals($nativeType)) {
 			$newType = TypeCombinator::intersect($type, $originalExprType);
-			if ($newType->isConstantScalarValue()->yes() && $newType->equals($originalExprType)) {
-				// don't add the same type over and over again to improve performance
-				return $this;
-			}
 			return $this->specifyExpressionType($expr, $newType, $newType, TrinaryLogic::createYes());
 		}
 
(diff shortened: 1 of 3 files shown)


src/Analyser/MutatingScope.php
```diff
@@ -3680,10 +3680,6 @@ public function addTypeToExpression(Expr $expr, Type $type): self
 
 		if ($originalExprType->equals($nativeType)) {
 			$newType = TypeCombinator::intersect($type, $originalExprType);
-			if ($newType->isConstantScalarValue()->yes() && $newType->equals($originalExprType)) {
```
staabm, line 3683: this code was added in the past to improve performance.

after measuring again it seems we no longer need this logic, as it seems we fixed the bottleneck somewhere else.

before this PR
```
➜  phpstan-src git:(2.1.x) hyperfine 'vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php' --runs=3 --warmup=1
Benchmark 1: vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php
  Time (mean ± σ):     15.854 s ±  0.095 s    [User: 15.462 s, System: 0.383 s]
  Range (min … max):   15.749 s … 15.934 s    3 runs
```

with deleted logic
```
➜  phpstan-src git:(bug13526) hyperfine 'vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php' --runs=3 --warmup=1
Benchmark 1: vendor/bin/phpunit tests/PHPStan/Analyser/AnalyserIntegrationTest.php
  Time (mean ± σ):     15.840 s ±  0.213 s    [User: 15.455 s, System: 0.375 s]
  Range (min … max):   15.642 s … 16.065 s    3 runs
```

VincentLanglet (approved):

PR #5117, opened 2026-03-01
Consider ArrayAccess side effects for dead catch analysis

VincentLanglet:
- Synthesize offsetGet throw points in ArrayDimFetch handler for ArrayAccess objects
- Synthesize offsetUnset throw points in Unset_ handler for ArrayAccess objects
- Synthesize offsetExists throw points in Isset_ handler for ArrayAccess objects
- offsetSet was already handled; this brings the other three operations to parity
- New regression test in tests/PHPStan/Rules/Exceptions/data/bug-11427.php

From https://github.com/phpstan/phpstan-src/pull/4910

Closes https://github.com/phpstan/phpstan/issues/11427

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -2094,7 +2094,7 @@ class NodeScopeResolver
 					if (!$varType->isArray()->yes() && !(new ObjectType(ArrayAccess::class))->isSuperTypeOf($varType)->no()) {
 						$throwPoints = array_merge($throwPoints, $this->processExprNode(
 							$stmt,
-							new MethodCall($this->deepNodeCloner->cloneNode($var->var), 'offsetUnset'),
+							new MethodCall($var->var, 'offsetUnset'),
 							$scope,
 							$storage,
 							new NoopNodeCallback(),
@@ -3629,11 +3629,11 @@ class NodeScopeResolver
 			if (!$varType->isArray()->yes() && !(new ObjectType(ArrayAccess::class))->isSuperTypeOf($varType)->no()) {
 				$throwPoints = array_merge($throwPoints, $this->processExprNode(
 					$stmt,
-					new MethodCall($this->deepNodeCloner->cloneNode($expr->var), 'offsetGet'),
+					new MethodCall($expr->var, 'offsetGet'),
 					$scope,
 					$storage,
 					new NoopNodeCallback(),
-					ExpressionContext::createDeep(),
+					$context,
 				)->getThrowPoints());
 			}
 		} elseif ($expr instanceof Array_) {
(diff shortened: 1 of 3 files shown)


src/Analyser/NodeScopeResolver.php
```diff
@@ -2090,6 +2090,18 @@ private function processStmtNode(
...
+					$varType = $scope->getType($var->var);
+					if (!$varType->isArray()->yes() && !(new ObjectType(ArrayAccess::class))->isSuperTypeOf($varType)->no()) {
+						$throwPoints = array_merge($throwPoints, $this->processExprNode(
+							$stmt,
+							new MethodCall($this->deepNodeCloner->cloneNode($var->var), 'offsetUnset'),
+							$scope,
+							$storage,
+							new NoopNodeCallback(),
+							ExpressionContext::createDeep(),
+						)->getThrowPoints());
```
staabm, line 2103: we have similar but different code already in 

https://github.com/phpstan/phpstan-src/blob/94731ccd4d617f48a36aa8c721a0bdf806c2570c/src/Analyser/NodeScopeResolver.php#L6428-L6435

I wonder why
- we need the cloning here
- why we process a method call on `$var->var` and not `$var`
- whether we need deep-context

VincentLanglet: > $var->var and not $var

The place `$var` is used we're doing 
```
while ($var instanceof ArrayDimFetch) {
     $var = $var->var;
```
before

src/Analyser/NodeScopeResolver.php
```diff
@@ -2090,6 +2091,18 @@ private function processStmtNode(
...
 				if ($var instanceof ArrayDimFetch && $var->dim !== null) {
+					$varType = $scope->getType($var->var);
+					if (!$varType->isArray()->yes() && !(new ObjectType(ArrayAccess::class))->isSuperTypeOf($varType)->no()) {
+						$throwPoints = array_merge($throwPoints, $this->processExprNode(
+							$stmt,
+							new MethodCall($var->var, 'offsetUnset'),
+							$scope,
+							$storage,
+							new NoopNodeCallback(),
+							ExpressionContext::createDeep(),
```
staabm, line 2102: this is the only occurrence where you still use a deep-context. intentional?

src/Analyser/NodeScopeResolver.php
```diff
@@ -6372,10 +6415,15 @@ private function processAssignVar(
 				$scope = $scope->assignExpression($expr, $type, $nativeType);
 			}
 
-			if (!$varType->isArray()->yes() && !(new ObjectType(ArrayAccess::class))->isSuperTypeOf($varType)->no()) {
+			$setVarType = $scope->getType($originalVar->var);
+			if (
+				!$setVarType instanceof ErrorType
```
staabm, line 6420: why do you need the `instanceof ErrorType` check only here?

staabm (approved):

src/Analyser/NodeScopeResolver.php
VincentLanglet: The `$context` is not an ExpressionContext but a StatementContext here, so I have to create a context (like it's done some line before)

src/Analyser/NodeScopeResolver.php
VincentLanglet: Maybe we need somewhere else, but this is the only place I need it to have green tests.
And I'm not sure to find more tests to write.

Notice this is also the only place we dont access $varType, because with
```
$foo[$a][$b] = 2
```
we have to call offsetSet on `$foo[$a]`;

PR #5118, opened 2026-03-01
Fix while-true-if-break: Variable might not be defined

VincentLanglet:
Cf https://github.com/phpstan/phpstan-src/pull/4959

Closes https://github.com/phpstan/phpstan/issues/11545
Closes https://github.com/phpstan/phpstan/issues/10245
Closes https://github.com/phpstan/phpstan/issues/11984
Closes https://github.com/phpstan/phpstan/issues/5477
Closes https://github.com/phpstan/phpstan/issues/5919
Closes https://github.com/phpstan/phpstan/issues/9023

### DIFF
--- a/src/Analyser/NodeScopeResolver.php
+++ b/src/Analyser/NodeScopeResolver.php
@@ -1522,8 +1522,16 @@ class NodeScopeResolver
 			}
 
 			$breakExitPoints = $finalScopeResult->getExitPointsByType(Break_::class);
-			foreach ($breakExitPoints as $breakExitPoint) {
-				$finalScope = $finalScope->mergeWith($breakExitPoint->getScope());
+			if ($alwaysIterates && count($breakExitPoints) > 0) {
+				$breakScope = null;
+				foreach ($breakExitPoints as $breakExitPoint) {
+					$breakScope = $breakScope === null ? $breakExitPoint->getScope() : $breakScope->mergeWith($breakExitPoint->getScope());
+				}
+				$finalScope = $breakScope;
+			} else {
+				foreach ($breakExitPoints as $breakExitPoint) {
+					$finalScope = $finalScope->mergeWith($breakExitPoint->getScope());
+				}
 			}
 
 			$isIterableAtLeastOnce = $beforeCondBooleanType->isTrue()->yes();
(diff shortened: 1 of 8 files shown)


tests/PHPStan/Rules/Variables/DefinedVariableRuleTest.php
```diff
@@ -1257,4 +1257,64 @@ public function testBug12944(): void
...
+	}
+
+	public function testBug11545(): void
+	{
+		$this->cliArgumentsVariablesRegistered = true;
+		$this->polluteScopeWithLoopInitialAssignments = false;
+		$this->checkMaybeUndefinedVariables = true;
+		$this->polluteScopeWithAlwaysIterableForeach = true;
+
+		$this->analyse([__DIR__ . '/data/bug-11545.php'], []);
```
VincentLanglet: This currently does not work because of 
https://phpstan.org/r/cd248edd-5252-4bba-ada5-26cce593585b

src/Analyser/NodeScopeResolver.php
```diff
@@ -1522,8 +1522,16 @@ private function processStmtNode(
...
+			if ($alwaysIterates && count($breakExitPoints) > 0) {
+				$breakScope = null;
+				foreach ($breakExitPoints as $breakExitPoint) {
+					$breakScope = $breakScope === null ? $breakExitPoint->getScope() : $breakScope->mergeWith($breakExitPoint->getScope());
+				}
+				$finalScope = $breakScope;
+			} else {
+				foreach ($breakExitPoints as $breakExitPoint) {
+					$finalScope = $finalScope->mergeWith($breakExitPoint->getScope());
+				}
```
staabm: looking at this, I think the only difference is whether we start at the `$breakScope` (and override `$finalScope`) or we always merge into `$finalScope`).

if I am right, I think this means this code can be simplified to

```suggestion
			if (count($breakExitPoints) > 0) {
				$breakScope = null;
				if (!$alwaysIterates) {
					$breakScope = $finalScope;
				}
				foreach ($breakExitPoints as $breakExitPoint) {
					$breakScope = $breakScope === null ? $breakExitPoint->getScope() : $breakScope->mergeWith($breakExitPoint->getScope());
				}
				$finalScope = $breakScope;
			}
```

staabm (approved):

PR #5131, opened 2026-03-04
Pass CLI `--error-format` option into the DIC container parameter

janedbal:
The errorFormat DIC parameter was not updated when the `--error-format` CLI option was used - it always held the config file value (default null). This meant reading the parameter from the container would not reflect the actually-used value.

### DIFF
--- a/src/Command/AnalyseCommand.php
+++ b/src/Command/AnalyseCommand.php
@@ -161,6 +161,7 @@ final class AnalyseCommand extends Command
 
 		$tmpFile = $input->getOption('tmp-file');
 		$insteadOfFile = $input->getOption('instead-of');
+		$errorFormat = $input->getOption('error-format');
 
 		if (
 			!is_array($paths)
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
If you want to have a non-default error formatter and you don't want to execute PHPStan each time with the CLI option, you can use the configuration parameter.

The configuration parameter was never meant to be used to query the current error formatter. What is your usecase, why are you interested in it?

janedbal:
> What is your usecase, why are you interested in it?

I want to be able to detect used formatter from within DCD. I need to enable some memory-intensive tracking for [autoremoval of dead code](https://github.com/shipmonk-rnd/dead-code-detector?tab=readme-ov-file#automatic-removal-of-dead-code) - and that does not make sense to do for regular executions. Thus, my idea was to check if errorFormat matches ours and we can perform some extra work.

ondrejmirtes:
I'm skeptical about this. For one, `errorFormat` is included in the list of parameters not invalidating result cache.

So it's not really okay to inject it into rules and collectors because the result cache might become stale (except for CollectedDataNode rule but it's brittle and hard to control).

Could you instead do the expensive work inside the special "fixing" error formatter altogether? Maybe somehow mark the output of rules and collectors with "hey, this needs some extra work" which could the error formatter pick up and finish before removing dead code?

janedbal:
> Could you instead do the expensive work inside the special "fixing" error formatter altogether?

Not really or very hardly. The concrete usecase is full removal neverRead property - not just definition, but all remaining writes to be deleted. So during main DCD analysis, I need to collect those and send them thru error.metadata to the formatter.

janedbal:
> So it's not really okay to inject it into rules and collectors because the result cache might become stale

The same issue is already there for all parameters not invalidating result cache. The only problem with this one is that it may not hold actual value so it is unreliable.

PR #5132, opened 2026-03-04
Update PDOStatement stub

VincentLanglet:
Rework of https://github.com/phpstan/phpstan-src/pull/5128

PDOStatement can either returns array, string or object based on the fetch mode
see https://github.com/phpstan/phpstan/issues/14206#issuecomment-3996154805,
so retricting the type to array is wrong ; rather than scalar|object|array, I think mixed will create less issues.

Fixes phpstan/phpstan#14206
Fixes https://github.com/phpstan/phpstan/issues/9120

Still ok with https://github.com/phpstan/phpstan/issues/8886 since it just asked about Iterator.

### DIFF
--- a/tests/PHPStan/Rules/PhpDoc/data/bug-14206.php
+++ b/tests/PHPStan/Rules/PhpDoc/data/bug-14206.php
@@ -14,4 +14,19 @@ class Test
 		$statement->setFetchMode(PDO::FETCH_COLUMN, 0);
 		$statement->execute();
 	}
+
+	public function test2(PDO $db): void
+	{
+		/** @var PDOStatement<int,array<string>> */
+		$statement = $db->prepare('SELECT foo FROM bar');
+		$statement->execute();
+	}
+
+	public function test3(PDO $db): void
+	{
+		/** @var PDOStatement<int,object> */
+		$statement = $db->prepare('SELECT foo FROM bar');
+		$statement->setFetchMode(PDO::FETCH_OBJ, 0);
+		$statement->execute();
+	}
 }
--- a/tests/PHPStan/Rules/PhpDoc/data/bug-14206.php
+++ b/tests/PHPStan/Rules/PhpDoc/data/bug-14206.php
@@ -1,0 +1,17 @@
+<?php declare(strict_types = 1);
+
+namespace Bug14206;
+
+use PDO;
+use PDOStatement;
+
+class Test
+{
+	public function test(PDO $db): void
+	{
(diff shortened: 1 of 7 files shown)


tests/PHPStan/Rules/PhpDoc/data/bug-14206.php
```diff
@@ -0,0 +1,17 @@
...
+use PDOStatement;
+
+class Test
+{
+	public function test(PDO $db): void
+	{
+		/** @var PDOStatement<int,string> */
+		$statement = $db->prepare('SELECT foo FROM bar');
+		$statement->setFetchMode(PDO::FETCH_COLUMN, 0);
+		$statement->execute();
```
staabm, line 15: for better real world use-case coverage, I think it would be good to have another similar example for the `PDOStatement<int,array>` case - implementation wise I think its good

staabm (approved):

MacGritsch:
> PDOStatement can either returns array, string or object based on the fetch mode see https://github.com/phpstan/phpstan/issues/14206#issuecomment-3996154805,

it can also return int,float or null. At least the mysql-driver does NOT return bool.

but `Iterator<mixed>` works fine (for me).

staabm (approved):

tests/PHPStan/Rules/PhpDoc/data/bug-14206.php
```diff
@@ -0,0 +1,32 @@
...
+		$statement = $db->prepare('SELECT foo FROM bar');
+		$statement->execute();
+	}
+
+	public function test3(PDO $db): void
+	{
+		/** @var PDOStatement<int,object> */
+		$statement = $db->prepare('SELECT foo FROM bar');
+		$statement->setFetchMode(PDO::FETCH_OBJ, 0);
+		$statement->execute();
```
staabm, line 30: this looks similar to 
https://github.com/phpstan/phpstan/issues/9120
https://phpstan.org/r/bddfabcb-164f-48e7-bf2e-b8f67939f655

(see issue-bot)

staabm: maybe it would make sense to add another test [with a object shape](https://phpstan.org/r/fb42db15-01aa-4188-837b-15d278049f2a)

VincentLanglet: I feel like it's ok for now

PR #5134, opened 2026-03-05
Implement array_all type specifying for arrow functions

uekann:
This PR partially resolves https://github.com/phpstan/phpstan/issues/12939 by specifying array element types when array_all returns true.

To keep this initial PR focused, it currently supports only ArrowFunction (e.g., `array_all($arr, fn($v, $k) => is_int($v) && is_string($k))`). It correctly recognizes union and intersection types when the arrow function uses `||` (or) and `&&` (and).

The following features are left as future scope:

Standard Closures: (e.g., `function($v) { ... }`)

String and First-class callables: (e.g., `'is_int'` or `is_int(...)`). Note that these currently throw an ArgumentCountError in PHP anyway, as `array_all()` passes both value and key, but functions like is_int expect exactly one argument (see: https://3v4l.org/2DQ9a#veol).

Custom assertions: Support for `@phpstan-assert-if-true` and similar annotations is not included in this PR, as mapping variables to types in those contexts is currently difficult and requires further work.

### DIFF
--- a/src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
@@ -18,7 +18,6 @@ use PHPStan\Type\FunctionTypeSpecifyingExtension;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 use function array_find;
-use function assert;
 use function count;
 use function is_string;
 use function strtolower;
@@ -38,7 +37,7 @@ final class ArrayAllFunctionTypeSpecifyingExtension implements FunctionTypeSpeci
 	public function specifyTypes(FunctionReflection $functionReflection, FuncCall $node, Scope $scope, TypeSpecifierContext $context): SpecifiedTypes
 	{
 		$args = $node->getArgs();
-		if (!$context->true() || count($args) < 2) {
+		if (!$context->truthy() || count($args) < 2) {
 			return new SpecifiedTypes();
 		}
 
@@ -49,14 +48,13 @@ final class ArrayAllFunctionTypeSpecifyingExtension implements FunctionTypeSpeci
 		} elseif (
 			$callable instanceof Expr\Closure &&
 			count($callable->stmts) === 1 &&
-			$callable->stmts[0] instanceof Stmt\Return_
+			$callable->stmts[0] instanceof Stmt\Return_ &&
+			isset($callable->stmts[0]->expr)
 		) {
 			$callableExpr = $callable->stmts[0]->expr;
 		} else {
 			return new SpecifiedTypes();
 		}
-
-		assert(isset($callableExpr));
 
 		$callableParams = $callable->params;
 		$specifiedTypesInFuncCall = $this->typeSpecifier->specifyTypesInCondition($scope, $callableExpr, $context)->getSureTypes();
--- a/src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
+++ b/src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
@@ -4,12 +4,14 @@ namespace PHPStan\Type\Php;
 
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\Variable;
 use PHPStan\Analyser\Scope;
 use PHPStan\Analyser\SpecifiedTypes;
 use PHPStan\Analyser\TypeSpecifier;
 use PHPStan\Analyser\TypeSpecifierAwareExtension;
 use PHPStan\Analyser\TypeSpecifierContext;
 use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Node\Expr\TypeExpr;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\ShouldNotHappenException;
@@ -38,12 +40,39 @@ final class ArrayAllFunctionTypeSpecifyingExtension implements FunctionTypeSpeci
 		}
 
 		$array = $args[0]->value;
+		$arrayArgType = $scope->getType($array);
+		$arrayTypes = $arrayArgType->getArrays();
+
+		if(count($arrayTypes) === 0) {
+			return new SpecifiedTypes();
+		}
+
 		$callable = $args[1]->value;
 
 
-		if ($callable instanceof Expr\ArrowFunction && $callable->expr instanceof Expr\FuncCall) {
+		// if ($callable instanceof Expr\ArrowFunction && $callable->expr instanceof Expr\FuncCall) {
+		if ($callable instanceof Expr\ArrowFunction) {
 
 			$callableParms = $callable->params;
+
+			// foreach($arrayTypes as $arrayType) {
+			// 	$keyType = $arrayType->getKeyType();
+			// 	$itemType = $arrayType->getItemType();
+			//
+			// 	$variables = [];
+			//
+			// 	if (count($callableParms) < 1) {
+			// 		continue;
+			// 	}
+			//
+			// 	if ($callableParms[0] ?? null instanceof Expr\Variable) {
+			// 		$variables[$callableParms[0]->name] = $itemType;
+			// 	}
+			//
+			// 	if ($callableParms[1] ?? null instanceof Expr\Variable) {
+			// 		$variables[$callableParms[1]->name] = $keyType;
+			// 	}
+			// }
 			$specifiedTypesInFuncCall = $this->typeSpecifier->specifyTypesInCondition($scope, $callable->expr, $context)->getSureTypes();
 
 			if(count($callableParms) >= 1 && $callableParms[0] instanceof Expr\Variable) {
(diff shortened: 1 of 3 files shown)


uekann:
I'm out this weekend and will get to any feedback on Monday.

tests/PHPStan/Analyser/TypeSpecifierTest.php
```diff
@@ -1320,6 +1320,139 @@ public static function dataCondition(): iterable
 				],
 				[],
 			],
+			[
+				new FuncCall(
```
staabm: AST based tests are unreadable. better would be to add new NodeScopeResolver tests in the `nsrt` folder and using `assertType` to define assertions.

feel free to look into other already merged PRs. most of them have tests like this

uekann: Thanks for the review! I've removed the AST-based tests and added the `NodeScopeResolver` tests using `assertType()` in the `nsrt` folder instead.

tests/PHPStan/Analyser/nsrt/array-all.php
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	/**
+	 * @param array<mixed> $array
+	 */
+	public function test8($array) {
+		if (array_all($array, fn ($value) => $value instanceof DateTime || $value instanceof DateTimeImmutable)) {
+			assertType("array<DateTime|DateTimeImmutable>", $array);
+		}
+	}
```
staabm, line 154: please add also tests for `list<mixed>` and `non-empty-array<mixed>`

src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,109 @@
...
+		$arrayArgType = $scope->getType($array);
+		$arrayTypes = $arrayArgType->getArrays();
+
+		if (count($arrayTypes) === 0) {
+			return new SpecifiedTypes();
+		}
+
+		$callable = $args[1]->value;
+
+		if ($callable instanceof Expr\ArrowFunction) {
```
staabm: when handling `Expr\ArrowFunction` you usually also need to think about `Expr\Closure`

tests/PHPStan/Analyser/nsrt/array-all.php
```diff
@@ -0,0 +1,82 @@
...
+use function PHPStan\Testing\assertType;
+
+class Foo {
+	/**
+	 * @param array<mixed> $array
+	 */
+	public function test1($array) {
+		if (array_all($array, fn ($value) => is_int($value))) {
+			assertType("array<int>", $array);
+		}
```
staabm, line 20: please also assert `$array` in the ELSE-branch and after the IF

uekann:
@staabm 
Thank you for the review.
I made the following changes:

- Added the test cases you suggested
- Added support for `Closure`

I also considered implementing support so that passing other functions annotated with `@phpstan-assert-if-true` to `array_all()` would enable proper type narrowing. However, when such a function is passed to `array_all()` as a first-class callable, for example as `f(...)`, it is treated as a `Closure`, and in PHPStan a `Closure` currently does not retain information equivalent to assert-if-true. For that reason, I have not implemented this at this time. (See: https://github.com/phpstan/phpstan/issues/14249)

src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,112 @@
...
+	public function isFunctionSupported(FunctionReflection $functionReflection, FuncCall $node, TypeSpecifierContext $context): bool
+	{
+		return strtolower($functionReflection->getName()) === 'array_all'
+			&& !$context->null();
+	}
+
+	public function specifyTypes(FunctionReflection $functionReflection, FuncCall $node, Scope $scope, TypeSpecifierContext $context): SpecifiedTypes
+	{
+		$args = $node->getArgs();
+		if (!$context->truthy() || count($args) < 2) {
```
staabm, line 40: I never know whether its `$context->truthy()` or `$context->true()`.

please add a test with a `=== true` comparison
(the escaped mutant at least tells us, its not covered by tests)

src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,112 @@
...
+
+		$array = $args[0]->value;
+		$callable = $args[1]->value;
+		if ($callable instanceof Expr\ArrowFunction) {
+			$callableExpr = $callable->expr;
+		} elseif (
+			$callable instanceof Expr\Closure &&
+			count($callable->stmts) === 1 &&
+			$callable->stmts[0] instanceof Stmt\Return_ &&
+			isset($callable->stmts[0]->expr)
```
staabm: I think you don't need this `Return_` related logic. if you drop this lines the logic should also work for multi-expr closures.

tests/PHPStan/Analyser/nsrt/array-all.php
```diff
@@ -0,0 +1,179 @@
+<?php // lint >= 8.4
```
staabm, line 1: does this test really depend on php 8.4? I think this can be lowered without problems?

src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,112 @@
...
+		) {
+			$callableExpr = $callable->stmts[0]->expr;
```
staabm, line 54: I don't think this is the right approach, but don't have a better idea atm either

tests/PHPStan/Analyser/nsrt/array-all.php
uekann: I understand that `array_all` is a function added in PHP 8.4. Therefore, the tests are set for 8.4 and later.
I ran it on my machine and it passed on 8.3 (I'm not sure why this is), so should I remove the mention of 8.4 and later?

staabm: good point. it works on 8.3 because we have symfony/polyfills.

leave it as is than, thanks.

src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
uekann: I initially considered performing type specification for arbitrary closures, but I concluded that this would require tracking the result of executing statements, which does not seem realistic. I also do not think that would align well with PHPStan’s design.

For that reason, I implemented it so that type specification is applied only to closures that contain nothing but a `return` statement.

Another option would be to apply type specification only to closures produced by converting a function to a first-class callable. I considered this because assertions written in PHPDoc can exist on the original function. However, at the time of implementation, that assertion information was lost when the function was converted to a closure, so this did not seem realistic either.

(The issue I opened about that, [https://github.com/phpstan/phpstan/issues/14249](https://github.com/phpstan/phpstan/issues/14249), has since been resolved, so this may become a viable option in the future. However, at least for now, I also cannot think of a better alternative, so I think this approach is acceptable as a temporary solution.)

src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
uekann: I added a test for the `=== true` comparison, but for a function returning bool it does not distinguish between `$context->true()` and `$context->truthy()`: both lead to the same narrowing on the true branch. So the test covers the syntax, but it does not really prove which context check is the correct one here.

zonuexe:
@ondrejmirtes 
Is there a possibility that this PR will be merged? While there are still some less-than-ideal aspects to the closure analysis, it seems to cover the realistic cases.

src/Type/Php/ArrayAllFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,112 @@
...
+		$array = $args[0]->value;
+		$callable = $args[1]->value;
+		if ($callable instanceof Expr\ArrowFunction) {
+			$callableExpr = $callable->expr;
+		} elseif (
+			$callable instanceof Expr\Closure &&
+			count($callable->stmts) === 1 &&
+			$callable->stmts[0] instanceof Stmt\Return_ &&
+			isset($callable->stmts[0]->expr)
+		) {
```
VincentLanglet, line 53: I agree with @staabm that this might not be satisfying.

The first thing I thought then was to check how it was done for array_filter (in ArrayFilterFunctionReturnTypeHelper) and they seems to do something similar
```
if ($callbackArg instanceof Closure && count($callbackArg->stmts) === 1 && count($callbackArg->params) > 0) {
			$statement = $callbackArg->stmts[0];
			if ($statement instanceof Return_ && $statement->expr !== null) {
```

But it handles also
```
 elseif (
			($callbackArg instanceof FuncCall || $callbackArg instanceof MethodCall || $callbackArg instanceof StaticCall)
			&& $callbackArg->isFirstClassCallable()
		)
```
and
```
else {
			$constantStrings = $scope->getType($callbackArg)->getConstantStrings();
			if (count($constantStrings) > 0) {
```
So I think we could have more case to handle.

PR #5136, opened 2026-03-05
TableErrorFormatter: Limit number of errors

staabm:
closes https://github.com/phpstan/phpstan/issues/14232

"mago benchmark" running PHPStan with warmed cache on macbook m4 pro with PHP 8.3.34
before this PR: 20-21 seconds
after this PR: 3-4 seconds

----

running on wordpress with warmed result cache

note the difference in "Elapsed time"

## before the PR

grafik

grafik

## after the PR

`../../bin/phpstan -v` 

grafik

----

`PHPSTAN_ERRORS_LIMIT=0 ../../bin/phpstan -v ` 

grafik

----

`PHPSTAN_ERRORS_LIMIT=50 ../../bin/phpstan -v `

grafik

### DIFF
--- a/src/Command/ErrorFormatter/TableErrorFormatter.php
+++ b/src/Command/ErrorFormatter/TableErrorFormatter.php
@@ -12,6 +12,7 @@ use PHPStan\File\RelativePathHelper;
 use PHPStan\File\SimpleRelativePathHelper;
 use Symfony\Component\Console\Formatter\OutputFormatter;
 use function array_map;
+use function array_slice;
 use function count;
 use function explode;
 use function getenv;
@@ -159,12 +160,15 @@ final class TableErrorFormatter implements ErrorFormatter
 				];
 			}
 
-			$style->table(['Line', $this->relativePathHelper->getRelativePath($file)], $rows);
 			$printedErrors += count($rows);
-
 			if ($errorsBudget > 0 && $printedErrors > $errorsBudget) {
+				$rows = array_slice($rows, 0, $errorsBudget - ($printedErrors - count($rows)));
+
+				$style->table(['Line', $this->relativePathHelper->getRelativePath($file)], $rows);
 				break;
 			}
+
+			$style->table(['Line', $this->relativePathHelper->getRelativePath($file)], $rows);
 		}
 
 		if (count($analysisResult->getNotFileSpecificErrors()) > 0) {
--- a/src/Command/ErrorFormatter/TableErrorFormatter.php
+++ b/src/Command/ErrorFormatter/TableErrorFormatter.php
@@ -5,6 +5,7 @@ namespace PHPStan\Command\ErrorFormatter;
 use PHPStan\Analyser\Error;
 use PHPStan\Command\AnalyseCommand;
 use PHPStan\Command\AnalysisResult;
+use PHPStan\Command\CommandHelper;
 use PHPStan\Command\Output;
 use PHPStan\DependencyInjection\AutowiredParameter;
 use PHPStan\DependencyInjection\AutowiredService;
@@ -44,7 +45,7 @@ final class TableErrorFormatter implements ErrorFormatter
 		#[AutowiredParameter]
 		private ?string $editorUrlTitle,
 		#[AutowiredParameter]
-		private ?int $level,
+		private string $usedLevel,
 	)
 	{
 	}
@@ -189,7 +190,7 @@ final class TableErrorFormatter implements ErrorFormatter
 
 			$note = [];
 			$note[] = sprintf('Result is limited to the first %d errors', $errorsBudget);
-			if ($this->level > 0) {
+			if ($this->usedLevel !== CommandHelper::DEFAULT_LEVEL) {
 				$note[] = '- Consider lowering the PHPStan level';
 			}
 			$note[] = sprintf('- Pass %s=1 environment variable to show all errors', self::FORCE_SHOW_ALL_ERRORS);
--- a/src/Command/ErrorFormatter/TableErrorFormatter.php
+++ b/src/Command/ErrorFormatter/TableErrorFormatter.php
@@ -29,7 +29,7 @@ use function str_replace;
 final class TableErrorFormatter implements ErrorFormatter
 {
 
-	private const MAX_ERRORS_TO_SHOW = 1000;
+	private const DEFAULT_MAX_ERRORS_TO_SHOW = 1000;
 
 	public function __construct(
 		private RelativePathHelper $relativePathHelper,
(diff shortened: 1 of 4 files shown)


VincentLanglet (commented):
A (big) file with looooot of errors will still display way more than 1000 errors since you break between files.

Also I find weird that something like `PHPSTAN_ERRORS_LIMIT=1` will display more than 1 error (if the first file has more than 1 error).

Should we break inside the file foreach too ?

staabm:
> A (big) file with looooot of errors will still display way more than 1000 errors since you break between files.

yeah, I did it intentionally like this.
adjusted it now, as I found an easy way to fit the limit

VincentLanglet (approved):

src/Command/ErrorFormatter/TableErrorFormatter.php
```diff
@@ -172,6 +191,15 @@ public function formatErrors(
 			$style->warning($finalMessage);
 		}
 
+		if ($errorsBudget > 0 && $printedErrors > $errorsBudget) {
+			$note = [];
+			$note[] = sprintf('Result is limited to the first %d errors', $errorsBudget);
+			$note[] = '- Consider lowering the PHPStan level';
```
ondrejmirtes: Only output this on rule level > 0.

src/Command/ErrorFormatter/TableErrorFormatter.php
ondrejmirtes: On this line show "1000+ errors" when applicable. Yeah, it's meant to be annoying :)

src/Command/ErrorFormatter/TableErrorFormatter.php
```diff
@@ -172,6 +191,15 @@ public function formatErrors(
...
+			$note[] = '- Consider using PHPStan Pro for more comfortable error browsing';
```
ondrejmirtes, line 202: Add: "Learn more: https://www.phpstan.com/"

ondrejmirtes (changes requested):
Instead of `PHPSTAN_ERRORS_LIMIT` variable where user can shoot themselves in the foot with a non-sense value like `3`, I'd more like a boolean-like var with the name of `PHPSTAN_TABLE_ERROR_FORMATTER_FORCE_SHOW_ALL_ERRORS`. The default (undefined or `0`) would only show the first 1000 errors, setting to `1` would show everything and hide the note.

staabm:
latest result

grafik

src/Command/ErrorFormatter/TableErrorFormatter.php
```diff
@@ -84,6 +90,14 @@ public function formatErrors(
 			$fileErrors[$fileSpecificError->getFile()][] = $fileSpecificError;
 		}
 
+		$forceShowAll = getenv('PHPSTAN_TABLE_ERROR_FORMATTER_FORCE_SHOW_ALL_ERRORS');
```
ondrejmirtes: Would be nice to have this name here in a private constant, it's referred to twice in the source code.

src/Command/ErrorFormatter/TableErrorFormatter.php
```diff
@@ -84,6 +90,14 @@ public function formatErrors(
...
+		if (in_array($forceShowAll, [false, '0'], true)) {
+			$errorsBudget = self::ERRORS_LIMIT;
+		} else {
+			$errorsBudget = 0;
```
ondrejmirtes: Null would be nicer?

src/Command/ErrorFormatter/TableErrorFormatter.php
```diff
@@ -38,6 +42,8 @@ public function __construct(
 		private ?string $editorUrl,
 		#[AutowiredParameter]
 		private ?string $editorUrlTitle,
+		#[AutowiredParameter]
+		private ?int $level,
```
ondrejmirtes: `usedLevel` might work a bit better, that's what's used across the codebase for this.

ondrejmirtes (changes requested):

staabm:
Memo to me: check whether the new limit also affects the baseline generation

ondrejmirtes:
>  check whether the new limit also affects the baseline generation

It won't, you just touched TableErrorFormatter, baseline is generated with a different one.

src/Command/ErrorFormatter/TableErrorFormatter.php
```diff
@@ -149,6 +165,14 @@ public function formatErrors(
 				];
 			}
 
+			$printedErrors += count($rows);
+			if ($errorsBudget !== null && $printedErrors > $errorsBudget) {
+				$rows = array_slice($rows, 0, $errorsBudget - ($printedErrors - count($rows)));
```
ondrejmirtes, line 175: Last thing - I'm not sure whether this line works correctly, a lot is going on there.

For example it's easy to make a mistake in `$errorsBudget - ($printedErrors - count($rows))` based on whether `$printedErrors` already includes current `count($rows)` or if the `+=` operation happened after this computation.

I'm thinking this should really have some unit tests.

My suggestion - constructor of TableErrorFormatter should have a new optional nullable parameter. If it's null, then the autodetection based on the env variable will happen. If it's not null then that will become the error limit.

The unit test will then construct TableErrorFormatter with a specific value, can be 5. And we should test the output for < 5, exactly 5 and > 5.

ondrejmirtes (changes requested):

tests/PHPStan/Command/ErrorFormatter/TableErrorFormatterTest.php
```diff
@@ -293,6 +295,191 @@ public function testFormatErrors(
...
+	#[DataProvider('dataErrorLimit')]
+	public function testErrorLimit(
+		?int $errorsBudget,
+		string $usedLevel,
+		bool $showAllErrors,
+		string $expected,
+		int $generateErrorsCount = 4,
+	): void
+	{
+		putenv('COLUMNS=200');
```
staabm: took me a while to realize that I need this env var in the test, to get stable results across local workstation and CI :-)

staabm:
@ondrejmirtes finally the tests "work" cross plattform. had to skip windows because of slight line-break differences enforced by symfony/console. but I think that doesn't matter for the tests we do here.

staabm:
remaining errors also happen on 2.1.x - so seem unrelated

mvorisek:
Based on the issue description the performance issue should be given by symfony/console. Why not fix the issue there and possibly patch it until released?

I personally prefer all errors shown as very often I do ctrl+f and want to know how many occurences of something are present.

ondrejmirtes (approved):

ondrejmirtes:
Love it, thank you!

ondrejmirtes:
@mvorisek For your use-case, see the error message that's printed when there are 1000+ errors.

staabm:
opened a upstream issue, so symfony knows about the performance and might improve it in the future:
https://github.com/symfony/symfony/issues/63630

PR #5145, opened 2026-03-08
Fix spurious error nullCoalesce.offset

staabm:
before this PR we only remembered maybe/yes expr certainty when restoring expressions.
this lead to that previously not-existing expressions (no-certainy) where restored into a yes-certainty instead of the state before.

closes https://github.com/phpstan/phpstan/issues/13921
refs https://github.com/phpstan/phpstan/issues/10305#issuecomment-1885217867 (fixes a issue mentioned in comments but not the issue itself)

### DIFF
--- a/src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
+++ b/src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
@@ -42,7 +42,7 @@ final class NonNullabilityHelper
 				$originalNativeType = $originalScope->getNativeType($exprToSpecify);
 
 				return new EnsuredNonNullabilityResult($scope, [
-					new EnsuredNonNullabilityResultExpression($exprToSpecify, $originalExprType, $originalNativeType, $certainty),
+					new EnsuredNonNullabilityResultExpression($exprToSpecify, $originalExprType, $originalNativeType, $hasExpressionType),
 				]);
 			}
 			return new EnsuredNonNullabilityResult($scope, []);
@@ -55,16 +55,13 @@ final class NonNullabilityHelper
 		// To properly revert this, we must also save and restore the parent expression's type.
 		if ($exprToSpecify instanceof Expr\ArrayDimFetch && $exprToSpecify->dim !== null) {
 			$parentExpr = $exprToSpecify->var;
-			$parentCertainty = TrinaryLogic::createYes();
 			$hasParentExpressionType = $originalScope->hasExpressionType($parentExpr);
-			if (!$hasParentExpressionType->no()) {
-				$parentCertainty = $hasParentExpressionType;
-			}
+
 			$specifiedExpressions[] = new EnsuredNonNullabilityResultExpression(
 				$parentExpr,
 				$scope->getType($parentExpr),
 				$scope->getNativeType($parentExpr),
-				$parentCertainty,
+				$hasParentExpressionType,
 			);
 		}
 
@@ -104,6 +101,10 @@ final class NonNullabilityHelper
 	public function revertNonNullability(MutatingScope $scope, array $specifiedExpressions): MutatingScope
 	{
 		foreach ($specifiedExpressions as $specifiedExpressionResult) {
+			if ($specifiedExpressionResult->getCertainty()->no()) {
+				$scope = $scope->invalidateExpression($specifiedExpressionResult->getExpression());
+				continue;
+			}
 			$scope = $scope->specifyExpressionType(
 				$specifiedExpressionResult->getExpression(),
 				$specifiedExpressionResult->getOriginalType(),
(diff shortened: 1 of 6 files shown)


tests/PHPStan/Analyser/nsrt/has-offset-type-bug.php
```diff
@@ -105,7 +105,7 @@ public function doFoo2($mixed)
 	public function doBar(\SimpleXMLElement $xml)
 	{
 		if (isset($xml['foo'])) {
-			assertType('SimpleXMLElement', $xml['foo']);
+			assertType("SimpleXMLElement&hasOffset('foo')", $xml['foo']);
```
VincentLanglet: This seems wrong.

`xml` has offset foo,
but `xml['foo']` does not have offset foo (we have nothing about `xml['foo']['foo']`

VincentLanglet (changes requested):

tests/PHPStan/Analyser/nsrt/has-offset-type-bug.php
staabm: great catch, thanks!

src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
```diff
@@ -32,7 +32,10 @@ public function ensureShallowNonNullability(MutatingScope $scope, Scope $origina
 		// keep certainty
 		$certainty = TrinaryLogic::createYes();
 		$hasExpressionType = $originalScope->hasExpressionType($exprToSpecify);
-		if (!$hasExpressionType->no() || $exprToSpecify instanceof ArrayDimFetch) {
+		if (
+			!$hasExpressionType->no()
+			|| ($exprType->isArray()->yes() && $exprToSpecify instanceof ArrayDimFetch)
```
VincentLanglet: Could you explain to me why the `$exprType->isArray()->yes()` solve the bug about the XmlElement ?

I mean why won't we have the bug with the array
```
public function doBar(array $array)
	{
		if (isset($array['foo'])) {
			assertType("array&hasOffset('foo')", $array['foo']);
			assertType("array&hasOffset('foo')", $array);
		}
	}
```
?

staabm: thanks to your comment, I found more test-cases which were broken.

.. and after lots of tries finally found a way to make all test pass :-)

staabm:
remaining errors also happen on 2.1.x - so seem unrelated

src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
```diff
@@ -42,7 +42,7 @@ public function ensureShallowNonNullability(MutatingScope $scope, Scope $origina
 				$originalNativeType = $originalScope->getNativeType($exprToSpecify);
 
 				return new EnsuredNonNullabilityResult($scope, [
-					new EnsuredNonNullabilityResultExpression($exprToSpecify, $originalExprType, $originalNativeType, $certainty),
+					new EnsuredNonNullabilityResultExpression($exprToSpecify, $originalExprType, $originalNativeType, $hasExpressionType),
```
VincentLanglet, line 45: Do we still need `$certainty` var ?

src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
```diff
@@ -55,16 +55,13 @@ public function ensureShallowNonNullability(MutatingScope $scope, Scope $origina
...
-			if (!$hasParentExpressionType->no()) {
-				$parentCertainty = $hasParentExpressionType;
-			}
+
 			$specifiedExpressions[] = new EnsuredNonNullabilityResultExpression(
 				$parentExpr,
 				$scope->getType($parentExpr),
 				$scope->getNativeType($parentExpr),
-				$parentCertainty,
+				$hasParentExpressionType,
```
VincentLanglet: Do we still need `$parentCertainty` var ?

src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
staabm: yes - its used a few lines below

src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
staabm: inlined

src/Analyser/ExprHandler/Helper/NonNullabilityHelper.php
VincentLanglet: We could maybe move the block
```
// keep certainty
		$certainty = TrinaryLogic::createYes();
		$hasExpressionType = $originalScope->hasExpressionType($exprToSpecify);
		if (!$hasExpressionType->no()) {
			$certainty = $hasExpressionType;
		}
```
where it's used in order to avoid those check when they are not needed, but it doesn't cost a lot i think.

VincentLanglet (approved):

staabm:
> We could maybe move the block

we need `$hasExpressionType` and `$certainty` in 2 independent branches. and `$certainty` depends on `$hasExpressionType`.

I think its already as easy as it can get

thank you.

staabm:
aahh I finally understood what you mean :-)

https://github.com/phpstan/phpstan-src/pull/5158

PR #5157, opened 2026-03-09
Support `chop()` function as an alias of `rtrim()`

zonuexe:
It's now a minor alias, but can occasionally be found in legacy code.

See <https://www.php.net/chop>.


staabm (approved):
thank you

VincentLanglet (approved):

PR #5135, opened 2026-03-05
Upgrade Dead Code Detector to 0.15

janedbal:
- https://github.com/shipmonk-rnd/dead-code-detector/releases/tag/0.15.0

PR #5160, opened 2026-03-09
Upgrade Dead Code Detector to 0.15.x

staabm:
https://github.com/phpstan/phpstan-src/pull/5135 but with conflicts resolved

### DIFF
--- a/build/phpstan.neon
+++ b/build/phpstan.neon
@@ -91,6 +91,12 @@ parameters:
 				- ../tests/PHPStan/Fixture
 			reportUnmatched: false # constants on enums, not reported on PHP8-
 		-
+			identifier: shipmonk.deadEnumCase
+			paths:
+				- ../tests/PHPStan/Fixture
+		-
+			rawMessage: 'Property PHPStan\Command\CommandHelper::$reservedMemory is never read' # intentional
+		-
 			message: '''
 				#^Access to constant on deprecated class DeprecatedAnnotations\\DeprecatedFoo\:
 				in 1\.0\.0\.$#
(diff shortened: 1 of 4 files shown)


staabm:
@janedbal fyi, it seem the new version requires more memory than the old one.

a job which worked on 0.14.x now runs into a memory limit:
https://github.com/phpstan/phpstan-src/actions/runs/22850713236/job/66278297491?pr=5160

VincentLanglet (approved):

ondrejmirtes:
We have planned improvements to the result cache which should help. I'd rather soon do the improvements and then upgrade this package.

janedbal:
I analysed this on friday. See https://github.com/shipmonk-rnd/dead-code-detector/pull/310. Using `--memory-limit` is pretty bad way how to keep memory in shape. PHPStan limits memory per process, which limits how much work can DCD do here. But the peak of consumed memory by PHPStan is the sum of worker processes near their exit:

image



And that is FAR beyond what DCD takes (becase that runs when child processes are long dead). So I'd say it is ok for collector-based rules to take MUCH more memory (below that sum) as it does not increase overall memory consumption which you typically care about (CI limits). But thanks to `--memory-limit`, I cannot.

------

Also, regular projects are not affected by DCD memory at all (not even when you look at the main process, e.g. ShipMonk codebase main process takes about 4GB and DCD is only about 800 MB of that) - phpstan-src is an exception for some reason.

-----

Also, I can "fix" memory consumtion here in PHPStan by omitting [that cache](https://github.com/shipmonk-rnd/dead-code-detector/pull/310), but it makes no sense in real world. I can also make it configurable for you, but again - it makes no sense imo.

staabm:
thanks everyone.

lets revisit after https://github.com/phpstan/phpstan/issues/14074 was resolved

PR #5183, opened 2026-03-10
Fix fatal error when `path` has array value in ignoreErrors

janedbal:
Closes phpstan/phpstan#14231

### DIFF
--- a/src/DependencyInjection/ValidateIgnoredErrorsExtension.php
+++ b/src/DependencyInjection/ValidateIgnoredErrorsExtension.php
@@ -38,10 +38,12 @@ use PHPStan\Type\TypeAliasResolver;
 use function array_keys;
 use function array_map;
 use function count;
+use function gettype;
 use function implode;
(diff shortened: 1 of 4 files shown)


staabm:
why is this parameter not validated [against the schema](https://github.com/phpstan/phpstan-src/blob/5e75c86187ab9f827027430a75be9483ac05907c/conf/parametersSchema.neon#L134)?

janedbal:
The `ValidateIgnoredErrorsExtension::loadConfiguration()` runs during DI container compilation, before the schema validation. So the schema (`path: string()`) never gets a chance to catch the wrong type.

VincentLanglet (approved):

VincentLanglet:
@ondrejmirtes you might want to give a word about the error wording ?

staabm (approved):

VincentLanglet:
Let's merge this, ondrej can still change the wording later ; it won't be a BC break.

PR #5192, opened 2026-03-10
Narrow DOMDocument::createElement() return type and throw type for valid constant names

janedbal:
Fixes phpstan/phpstan#13792


src/Type/Php/DomDocumentCreateElementDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,63 @@
...
+
+		foreach ($argType->getConstantStrings() as $constantString) {
+			try {
+				$doc->createElement($constantString->getValue());
+			} catch (DOMException) {
+				return null;
+			}
+
+			$argType = TypeCombinator::remove($argType, $constantString);
+		}
```
staabm, line 52: this feels like a expensive operation. could you measure whether it would make sense to establish some kind of caching so we don't need to create elements for same named strings we find while analysing over and over?

src/Type/Php/DomDocumentCreateElementDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,63 @@
...
+{
+
+	public function getClass(): string
+	{
+		return DOMDocument::class;
+	}
+
+	public function isMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return extension_loaded('dom') && $methodReflection->getName() === 'createElement';
```
staabm, line 30: I am not sure whether this kind of optional runtime dependency is a good thing to have.
would it make sense to hard-code a list of valid elements instead?

janedbal: This logic was basically copied from SimpleXMLElementXpathMethodReturnTypeExtension, so I just followed that one.

src/Type/Php/DomDocumentCreateElementDynamicReturnTypeExtension.php
janedbal: This logic was basically copied from SimpleXMLElementXpathMethodReturnTypeExtension, so I just followed that one.

janedbal: > some kind of caching so we don't need to create elements for same named strings we find while analysing over and over

I dislike microoptimizations without seeing some SPX profile first. But another option might be to do some regex check for safe names, like a-z

staabm: > without seeing some SPX profile first

absolutely. thats what I had in mind as I said "could you measure whether.." :-)

might also make sense to hard-code the most common strings - but as you said, would require some profilling first

janedbal: Just measured difference between DOM and regex over 100k iterations:

- Regex (preg_match):                             15.19 ms
- new DOMDocument + createElement:                43.34 ms

Imo not worth it, wdyt?

staabm: ok fine than, thanks for looking into it

tests/PHPStan/Analyser/nsrt/dom-document-create-element.php
```diff
@@ -0,0 +1,48 @@
+<?php declare(strict_types = 1);
+
+namespace DomDocumentCreateElement;
+
+use DOMDocument;
+use function PHPStan\Testing\assertType;
+
+class Foo
```
staabm, line 8: running this file outside unit tests results in

```
  13     Method DomDocumentCreateElement\Foo::dynamicName() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.           
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:13                                                                                                            
  18     Method DomDocumentCreateElement\Foo::validConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.    
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:18                                                                                                            
  19     Method DomDocumentCreateElement\Foo::validConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.    
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:19                                                                                                            
  20     Method DomDocumentCreateElement\Foo::validConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.    
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:20                                                                                                            
  21     Method DomDocumentCreateElement\Foo::validConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.    
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:21                                                                                                            
  22     Method DomDocumentCreateElement\Foo::validConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.    
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:22                                                                                                            
  27     Method DomDocumentCreateElement\Foo::invalidConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.  
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:27                                                                                                            
  28     Method DomDocumentCreateElement\Foo::invalidConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.  
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:28                                                                                                            
  29     Method DomDocumentCreateElement\Foo::invalidConstantNames() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.  
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:29                                                                                                            
  38     Method DomDocumentCreateElement\Foo::unions() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.                
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:38                                                                                                            
  39     Method DomDocumentCreateElement\Foo::unions() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.                
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:39                                                                                                            
  45     Method DomDocumentCreateElement\Foo::localVariable() throws checked exception DOMException but it's missing from the PHPDoc @throws tag.         
         🪪  missingType.checkedException                                                                                                                 
         at dom-document-create-element.php:45   
```

which makes me think whether it would make sense to also add a `DynamicMethodThrowTypeExtension` which removes the `DOMException` throw-type for `$doc->createElement()` when we know elements can be created without errors

janedbal: Good point, added.

staabm (approved):

tests/PHPStan/Rules/Exceptions/MissingCheckedExceptionInMethodThrowsRuleTest.php
```diff
@@ -67,6 +67,10 @@ public function testRule(): void
 				122,
 			];
 		}
+		$errors[] = [
```
VincentLanglet: I think it's better to have a dedicated test
```
testBug13792
```
method rather modifying the existing one

Especially because there is already some 
```
if (PHP_VERSION_ID >=
```
condition which complexify it.

VincentLanglet (changes requested):

tests/PHPStan/Rules/Exceptions/MissingCheckedExceptionInMethodThrowsRuleTest.php
janedbal: Separated.

tests/PHPStan/Rules/Exceptions/MissingCheckedExceptionInMethodThrowsRuleTest.php
```diff
@@ -67,10 +67,6 @@ public function testRule(): void
 				122,
 			];
 		}
-		$errors[] = [
```
VincentLanglet: I'm sorry I wasn't clear.

I wanted to split the data file, but it can still be in this TestCase by writing a new test here
```
public function testBug13792(): void
	{
		$this->analyse([__DIR__ . '/data/bug-13792.php'], [
			[
				'Method Bug13792\Foo::dynamicName() throws checked exception DOMException but it\'s missing from the PHPDoc @throws tag.',
				20,
			],
			[
				'Method Bug13792\Foo::invalidConstantName() throws checked exception DOMException but it\'s missing from the PHPDoc @throws tag.',
				25,
			],
			[
				'Method Bug13792\Foo::unions() throws checked exception DOMException but it\'s missing from the PHPDoc @throws tag.',
				35,
			],
		]);
	}
```

janedbal: fixed

VincentLanglet (approved):

staabm (approved):
will merge after todays release

VincentLanglet:
Release is done

staabm:
since we had a crash bug I guess we will have another release ... maybe today.. will merge after we stabilized it

staabm:
thank you!

janedbal:
See real-world impact in this cleanup: https://github.com/shipmonk-rnd/coverage-guard/pull/78

PR #5200, opened 2026-03-11
Introduce PHPBench and move performance tests from AnalyserIntegrationTest to RegressionBench

ondrejmirtes:
Each testBug method now has a comment on its first line describing
the nature of the original issue: crash, performance, infinite run,
false positive, or false negative.

https://claude.ai/code/session_017EKgzxmzYFa1qUT8bcuohj


phpbench.json
```diff
@@ -0,0 +1,7 @@
+{
+	"$schema":"./vendor/phpbench/phpbench/phpbench.schema.json",
+    "runner.bootstrap": "vendor/autoload.php",
+    "runner.path": "tests/bench",
+	"runner.file_pattern": "*Bench.php",
+	"storage.xml_storage_path": "tests/bench/storage"
```
staabm: mix of tabs and spaces

PR #5156, opened 2026-03-09
Fix more PHP 7.4 errors

staabm:
after this PR 2 errors are left:

```
1) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "tests\PHPStan\Analyser\nsrt\range-function.php" ('D:\a\phpstan-src\phpstan-src\...on.php')
Failed assertions in D:\a\phpstan-src\phpstan-src\tests\PHPStan\Analyser\nsrt\range-function.php:

Line 17:
Expected: array{2, 3, 4, 5}
Actual:   array{2.0, 3.0, 4.0, 5.0}

D:\a\phpstan-src\phpstan-src\tests\PHPStan\Analyser\NodeScopeResolverTest.php:320
D:\a\phpstan-src\phpstan-src\tests\vendor\phpunit\phpunit\src\TextUI\Command.php:146
```

this seems to be a recent regression

----

```
2) PHPStan\Analyser\NodeScopeResolverTest::testFile with data set "tests\PHPStan\Analyser\nsrt\properties.php" ('D:\a\phpstan-src\phpstan-src\...es.php')
Failed assertions in D:\a\phpstan-src\phpstan-src\tests\PHPStan\Analyser\nsrt\properties.php:

Line 169:
Expected: DOMElement|null
Actual:   string
```

I wonder why this test does not fail starting with PHP 8.2+ - not sure whether its expected or not

PR #5202, opened 2026-03-11
Fix remaining errors after LegacyNodeScopeResolverTest refactoring

staabm:
fix remaining errors mentioned in https://github.com/phpstan/phpstan-src/pull/5156#issue-4044206472

see https://github.com/phpstan/phpstan-src/blob/5930232b544b357d3ece6a5248c7e99fc88dd22c/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php#L1844

see https://github.com/phpstan/phpstan-src/blob/5930232b544b357d3ece6a5248c7e99fc88dd22c/tests/PHPStan/Analyser/LegacyNodeScopeResolverTest.php#L5589


ondrejmirtes:
👍

PR #5210, opened 2026-03-12
array_column should not extract non-public properties from objects

janedbal:
Fixes https://github.com/phpstan/phpstan/issues/13573

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/array-column-php82.php
+++ b/tests/PHPStan/Analyser/nsrt/array-column-php82.php
@@ -267,7 +267,7 @@ class ArrayColumnVisibilityNonFinalTest
 	public function testNonFinalConstant(array $objects): void
 	{
 		assertType('array{int}', array_column($objects, 'pub'));
-		assertType('list<int>', array_column($objects, 'prot'));
+		assertType('list<int>', array_column($objects, 'prot')); // Could be array{}|array{int}
 		assertType('list', array_column($objects, 'priv'));
 	}
 
--- a/tests/PHPStan/Analyser/nsrt/array-column-php82.php
+++ b/tests/PHPStan/Analyser/nsrt/array-column-php82.php
@@ -237,3 +237,150 @@ final class Foo
 	}
 
 }
+
+class NonFinalObjectWithVisibility
+{
+	public int $pub = 1;
+	protected int $prot = 2;
+	private int $priv = 3;
+}
+
+final class FinalObjectWithVisibility
+{
+	public int $pub = 1;
+	protected int $prot = 2;
+	private int $priv = 3;
+}
+
+class ArrayColumnVisibilityNonFinalTest
+{
+
+	/** @param array<int, NonFinalObjectWithVisibility> $objects */
+	public function testNonFinal(array $objects): void
+	{
+		assertType('list<int>', array_column($objects, 'pub'));
+		assertType('list<int>', array_column($objects, 'prot'));
+		assertType('list', array_column($objects, 'priv'));
+	}
+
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Analyser/nsrt/array-column-php82.php
```diff
@@ -237,3 +237,122 @@ public function doFoo(array $a): void
...
+
+class ArrayColumnVisibilityTest
+{
+
+	/** @param array<int, ObjectWithVisibility> $objects */
+	public function testNonPublicProperties(array $objects): void
+	{
+		assertType('list<int>', array_column($objects, 'pub'));
+		assertType('array{}', array_column($objects, 'prot'));
+		assertType('array{}', array_column($objects, 'priv'));
```
VincentLanglet, line 284: This is not true with
```
class ParentObj extends ObjectWithVisibility {
    public int $prot = 2;
	public int $priv = 3;
}
```
Or even
```
class ParentObj extends ObjectWithVisibility {
	public string $priv = 'foo';
}
```

VincentLanglet (changes requested):

tests/PHPStan/Analyser/nsrt/array-column-php82.php
janedbal: What do you suggest? Only infer for final classes?

VincentLanglet: I would expect something like
```
assertType('list<int>', array_column($nonFinalObjects, 'pub'));
assertType('list<int>', array_column($finalObjects, 'pub'));

assertType('list<int>', array_column($nonFinalObjects, 'prot')); // Since the child need to respect the type.
assertType('array{}', array_column($finalObjects, 'prot'));

assertType('list<mixed>', array_column($nonFinalObjects, 'priv'));
assertType('array{}', array_column($finalObjects, 'priv'));
``

Comment (staabm) [reply tests/PHPStan/Analyser/nsrt/array-column-php82.php:284]: just came to mind.. it can be also made a difference between

```
	/** @param array<int, ObjectWithVisibility> $objects */
	public function testNonPublicProperties(array $objects): void
	{
```

and implicit final

```
	public function testNonPublicProperties(): void
	{
		$objects = [new ObjectWithVisibility()];
```

Comment (janedbal) [reply tests/PHPStan/Analyser/nsrt/array-column-php82.php:284]: Should be adjusted as you wish :)

**Review by staabm — COMMENTED** [review]
I think we also need to assert expectations for classes which define properties via phpdoc like

```
/**
 * @property int $id
 * @property string $name
 */
class ObjectWithIsset {}
```

File: tests/PHPStan/Analyser/nsrt/array-column-php82.php [hunk 2]
```diff
@@ -237,3 +237,163 @@ public function doFoo(array $a): void
...
+		assertType('list<int>', array_column($objects, 'pub'));
+		assertType('list<int>', array_column($objects, 'prot'));
+		assertType('list', array_column($objects, 'priv'));
+	}
+
+	/** @param array{NonFinalObjectWithVisibility} $objects */
+	public function testNonFinalConstant(array $objects): void
+	{
+		assertType('array{int}', array_column($objects, 'pub'));
+		assertType('list<int>', array_column($objects, 'prot'));
```
VincentLanglet: Might be useful to add a comment to explain the result (and how to improve it later)
```suggestion
		assertType('list<int>', array_column($objects, 'prot')); // Could be array{}|array{int}
```

VincentLanglet (approved):
Seems fine to me

staabm:
lets give it a go.

thank you

PR #5211, opened 2026-03-12
Fix false positive nullsafe.neverNull on left side of ??

kemo:
## Problem

`IssetCheck` unconditionally reports `?->` property access as unnecessary (`nullsafe.neverNull`) when it appears on the left side of `??`, `isset()`, or `empty()`.

This is correct for `isset()`/`empty()` — PHP handles null objects natively in those constructs without throwing. But for `??`, it is a **false positive**: the null coalescing operator does **not** catch `TypeError` from property access on a null object.

```php
class Foo {
    public function getBar(): ?Bar { /* ... */ }
}

// PHPStan reports: Using nullsafe property access "?->baz" on left side of ?? is unnecessary.
// But removing ?-> would cause TypeError at runtime when getBar() returns null:
$foo->getBar()?->baz ?? 'default';
//              ^^^
//  Without ?->: null->baz throws TypeError (not caught by ??)
//  With ?->:    null?->baz evaluates to null, ?? returns 'default'
```

## Root cause

In `IssetCheck::check()`, the `NullsafePropertyFetch` block (line 254) fires unconditionally regardless of context. It doesn't distinguish between `??` (where `?->` prevents `TypeError`) and `isset()`/`empty()` (where PHP already handles null objects).

## Fix

When the identifier is `nullCoalesce`, check whether the object being accessed (`$expr->var`) has a nullable type. If it does, the nullsafe operator is genuinely needed and no error should be reported.

`isset()` and `empty()` behavior is unchanged — `?->` remains correctly flagged as unnecessary there since PHP handles null objects natively in those constructs.

## Test changes

- Removed 3 false positive expectations from `NullCoalesceRuleTest::testBug7109()` (lines 17, 28, 66 of the test data file)
- Added new regression test `testNullsafeCoalesceNullableObject()` covering:
  - Chained nullable method calls with `?->` on left of `??` (no error expected)
  - Single nullable return with `?->` (no error expected)
  - Non-nullable object with `?->` (correctly reports "Expression not nullable")

### DIFF
--- a/tests/PHPStan/Rules/Variables/NullCoalesceRuleTest.php
+++ b/tests/PHPStan/Rules/Variables/NullCoalesceRuleTest.php
@@ -254,20 +254,8 @@ class NullCoalesceRuleTest extends RuleTestCase
 	{
 		$this->analyse([__DIR__ . '/../Properties/data/bug-7109.php'], [
 			[
-				'Using nullsafe property access "?->aaa" on left side of ?? is unnecessary. Use -> instead.',
-				17,
-			],
-			[
-				'Using nullsafe property access "?->aaa" on left side of ?? is unnecessary. Use -> instead.',
-				28,
-			],
-			[
 				'Expression on left side of ?? is not nullable.',
 				40,
-			],
-			[
-				'Using nullsafe property access "?->aaa" on left side of ?? is unnecessary. Use -> instead.',
-				66,
 			],
 			[
 				'Expression on left side of ?? is not nullable.',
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Rules/Variables/NullCoalesceRuleTest.php
```diff
@@ -253,22 +253,10 @@ public function testBug5933(): void
 	public function testBug7109(): void
 	{
 		$this->analyse([__DIR__ . '/../Properties/data/bug-7109.php'], [
-			[
-				'Using nullsafe property access "?->aaa" on left side of ?? is unnecessary. Use -> instead.',
```
VincentLanglet, line 257: I have no error with
https://3v4l.org/5TbIf#v8.3.30

so the ?-> seems useless.

VincentLanglet (commented):
Please provides a php snippet with leads to an error.

kemo:
Closing this — after testing actual PHP behavior, I confirmed that `??` does handle `null->property` access via `FETCH_OBJ_IS` opcode, so the nullsafe on property access is genuinely unnecessary on the left side of `??`. The distinction is that `null->method()` throws (method calls always execute), but `null->prop` is handled by `??`. PHPStan's current behavior is correct. Apologies for the noise.

PR #5212, opened 2026-03-12
Fix list description with offset required

VincentLanglet:
Now https://github.com/phpstan/phpstan-src/pull/5067 is merged, it allows this.


tests/PHPStan/Analyser/nsrt/bug-14177.php
```diff
@@ -12,7 +12,7 @@ class HelloWorld
 	public function testList(array $b): void
 	{
 		if (array_key_exists(3, $b)) {
-			assertType('list{0: string, 1: string, 2?: string, 3: string}', $b);
```
VincentLanglet, line 15: Since 3 is required, 2 shouldn't be described as optional.

staabm (approved):

PR #5220, opened 2026-03-14
Revert d7ba1e3

staabm:
reverts https://github.com/phpstan/phpstan-src/pull/5047

closes https://github.com/phpstan/phpstan/issues/14279


staabm:
I think we need to revert this for now, because the real fix for https://github.com/phpstan/phpstan/issues/14281 will require to change types expected behaviour when instantiation of a generic class without a template

https://phpstan.org/r/63aaf2ad-5bdb-4929-a32b-11b2b6332460

as it currently leads to Shopware\Tests\Unit\Core\Framework\Struct\TestCollection<*NEVER*>

ondrejmirtes:
Thank you!

PR #5223, opened 2026-03-15
Implement GMP operator type specifying extension

Firehed:
## Summary
- Add `GmpOperatorTypeSpecifyingExtension` to properly infer return types for GMP operator overloads
- GMP supports arithmetic (+, -, *, /, %, **), bitwise (&, |, ^, ~, <<, >>), and comparison operators
- Extension only claims support when both operands are GMP-compatible (GMP, int, or numeric-string)
- Update `InitializerExprTypeResolver` to call operator extensions early for object types

Fixes phpstan/phpstan#14288

## Test plan
- [x] Added comprehensive test coverage in `tests/PHPStan/Analyser/nsrt/gmp-operators.php`
- [x] Tests cover GMP on both left and right sides of operators
- [x] Tests cover both operator overloads and corresponding `gmp_*` functions
- [x] All existing tests pass (11627 tests, 78892 assertions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Type/Php/GmpOperatorTypeSpecifyingExtension.php
+++ b/src/Type/Php/GmpOperatorTypeSpecifyingExtension.php
@@ -4,6 +4,7 @@ namespace PHPStan\Type\Php;
 
 use PHPStan\DependencyInjection\AutowiredService;
 use PHPStan\Type\BooleanType;
+use PHPStan\Type\ErrorType;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\ObjectType;
@@ -21,41 +22,21 @@ final class GmpOperatorTypeSpecifyingExtension implements OperatorTypeSpecifying
 			return false;
 		}
 
-		if (!in_array($operatorSigil, ['+', '-', '*', '/', '**', '%', '&', '|', '^', '<<', '>>', '<', '<=', '>', '>=', '==', '!=', '<=>'], true)) {
-			return false;
-		}
-
 		$gmpType = new ObjectType('GMP');
-		$leftIsGmp = $gmpType->isSuperTypeOf($leftSide)->yes();
-		$rightIsGmp = $gmpType->isSuperTypeOf($rightSide)->yes();
 
-		// At least one side must be GMP
-		if (!$leftIsGmp && !$rightIsGmp) {
-			return false;
-		}
-
-		// The other side must be GMP-compatible (GMP, int, or numeric-string)
-		// GMP operations with incompatible types (like stdClass) will error at runtime
-		return $this->isGmpCompatible($leftSide, $gmpType) && $this->isGmpCompatible($rightSide, $gmpType);
-	}
-
-	private function isGmpCompatible(Type $type, ObjectType $gmpType): bool
-	{
-		if ($gmpType->isSuperTypeOf($type)->yes()) {
-			return true;
-		}
-		if ($type->isInteger()->yes()) {
-			return true;
-		}
-		if ($type->isNumericString()->yes()) {
-			return true;
-		}
-		return false;
+		return in_array($operatorSigil, ['+', '-', '*', '/', '**', '%', '&', '|', '^', '<<', '>>', '<', '<=', '>', '>=', '==', '!=', '<=>'], true)
+			&& (
+				$gmpType->isSuperTypeOf($leftSide)->yes()
+				|| $gmpType->isSuperTypeOf($rightSide)->yes()
+			);
 	}
 
 	public function specifyType(string $operatorSigil, Type $leftSide, Type $rightSide): Type
 	{
 		$gmpType = new ObjectType('GMP');
+		$otherSide = $gmpType->isSuperTypeOf($leftSide)->yes()
+			? $rightSide
+			: $leftSide;
 
 		// Comparison operators return bool or int (for spaceship)
 		if (in_array($operatorSigil, ['<', '<=', '>', '>=', '==', '!='], true)) {
@@ -66,9 +47,16 @@ final class GmpOperatorTypeSpecifyingExtension implements OperatorTypeSpecifying
 			return IntegerRangeType::fromInterval(-1, 1);
 		}
 
-		// All arithmetic and bitwise operations on GMP return GMP
 		// GMP can operate with: GMP, int, or numeric-string
-		return $gmpType;
+		if (
+			$otherSide->isInteger()->yes()
+			|| $otherSide->isNumericString()->yes()
+			|| $gmpType->isSuperTypeOf($otherSide)->yes()
+		) {
+			return $gmpType;
+		}
+
+		return new ErrorType();
 	}
 
 }
--- a/src/Type/Php/GmpOperatorTypeSpecifyingExtension.php
+++ b/src/Type/Php/GmpOperatorTypeSpecifyingExtension.php
@@ -1,0 +1,74 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Type\BooleanType;
+use PHPStan\Type\IntegerRangeType;
+use PHPStan\Type\NeverType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\OperatorTypeSpecifyingExtension;
+use PHPStan\Type\Type;
+use function in_array;
+
+#[AutowiredService]
+final class GmpOperatorTypeSpecifyingExtension implements OperatorTypeSpecifyingExtension
+{
+
+	public function isOperatorSupported(string $operatorSigil, Type $leftSide, Type $rightSide): bool
+	{
+		if ($leftSide instanceof NeverType || $rightSide instanceof NeverType) {
+			return false;
+		}
+
+		if (!in_array($operatorSigil, ['+', '-', '*', '/', '**', '%', '&', '|', '^', '<<', '>>', '<', '<=', '>', '>=', '==', '!=', '<=>'], true)) {
+			return false;
+		}
+
+		$gmpType = new ObjectType('GMP');
+		$leftIsGmp = $gmpType->isSuperTypeOf($leftSide)->yes();
+		$rightIsGmp = $gmpType->isSuperTypeOf($rightSide)->yes();
+
+		// At least one side must be GMP
+		if (!$leftIsGmp && !$rightIsGmp) {
+			return false;
+		}
+
+		// The other side must be GMP-compatible (GMP, int, or numeric-string)
+		// GMP operations with incompatible types (like stdClass) will error at runtime
+		return $this->isGmpCompatible($leftSide, $gmpType) && $this->isGmpCompatible($rightSide, $gmpType);
+	}
+
+	private function isGmpCompatible(Type $type, ObjectType $gmpType): bool
+	{
+		if ($gmpType->isSuperTypeOf($type)->yes()) {
+			return true;
+		}
+		if ($type->isInteger()->yes()) {
+			return true;
+		}
+		if ($type->isNumericString()->yes()) {
+			return true;
+		}
+		return false;
+	}
+
+	public function specifyType(string $operatorSigil, Type $leftSide, Type $rightSide): Type
+	{
+		$gmpType = new ObjectType('GMP');
+
+		// Comparison operators return bool or int (for spaceship)
+		if (in_array($operatorSigil, ['<', '<=', '>', '>=', '==', '!='], true)) {
+			return new BooleanType();
+		}
+
+		if ($operatorSigil === '<=>') {
+			return IntegerRangeType::fromInterval(-1, 1);
+		}
+
+		// All arithmetic and bitwise operations on GMP return GMP
+		// GMP can operate with: GMP, int, or numeric-string
+		return $gmpType;
+	}
+
+}
--- a/src/Type/Php/GmpUnaryOperatorTypeSpecifyingExtension.php
+++ b/src/Type/Php/GmpUnaryOperatorTypeSpecifyingExtension.php
@@ -1,0 +1,35 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Php;
+
+use PHPStan\DependencyInjection\AutowiredService;
+use PHPStan\Type\NeverType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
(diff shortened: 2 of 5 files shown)


src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -2034,6 +2058,17 @@ private function resolveConstantArrayTypeComparison(ConstantArrayType $leftType,
...
 	{
+		// Check operator type specifying extensions first for object types
+		// This allows extensions like GmpOperatorTypeSpecifyingExtension to
+		// handle operator overloading before integer range optimizations kick in
+		if ($leftType->isObject()->yes() || $rightType->isObject()->yes()) {
+			$specifiedTypes = $this->operatorTypeSpecifyingExtensionRegistryProvider->getRegistry()
+				->callOperatorTypeSpecifyingExtensions($expr, $leftType, $rightType);
+			if ($specifiedTypes !== null) {
+				return $specifiedTypes;
+			}
```
VincentLanglet: This become a duplicate with the check later 
```
$specifiedTypes = $this->operatorTypeSpecifyingExtensionRegistryProvider->getRegistry()
			->callOperatorTypeSpecifyingExtensions($expr, $leftType, $rightType);
		if ($specifiedTypes !== null) {
			return $specifiedTypes;
		}
```

I would either:
- remove this and rely on the later call
- or just move the existing call first with condition on object.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1092,6 +1108,14 @@ public function getBitwiseXorType(Expr $left, Expr $right, callable $getTypeCall
 		$leftType = $getTypeCallback($left);
 		$rightType = $getTypeCallback($right);
 
+		if ($leftType->isObject()->yes() || $rightType->isObject()->yes()) {
```
VincentLanglet: This shouldn't check for Object and be done unconditionally.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -1044,6 +1052,14 @@ public function getBitwiseOrType(Expr $left, Expr $right, callable $getTypeCallb
...
```
VincentLanglet: This shouldn't check for Object and be done unconditionally.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -986,6 +986,14 @@ public function getBitwiseAndType(Expr $left, Expr $right, callable $getTypeCall
...
```
VincentLanglet: This shouldn't check for Object and be done unconditionally.

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -2581,6 +2616,11 @@ public function getUnaryMinusType(Expr $expr, callable $getTypeCallback): Type
 	{
 		$type = $getTypeCallback($expr);
 
+		// GMP supports unary minus and returns GMP
+		if ($type->isObject()->yes() && (new ObjectType('GMP'))->isSuperTypeOf($type)->yes()) {
+			return new ObjectType('GMP');
+		}
```
VincentLanglet: Rather than an hardcoded check for this, I feel like it we should introduce an UnaryOperatorTypeSpecifyingExtension as explained here: https://github.com/phpstan/phpstan-src/pull/4980#issuecomment-3914609451

This is maybe better in a dedicated PR ?

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -2622,6 +2662,11 @@ public function getBitwiseNotType(Expr $expr, callable $getTypeCallback): Type
 	{
 		$exprType = $getTypeCallback($expr);
 
+		// GMP supports bitwise not and returns GMP
+		if ($exprType->isObject()->yes() && (new ObjectType('GMP'))->isSuperTypeOf($exprType)->yes()) {
```
VincentLanglet: Same for UnaryOperatorTypeSpecifyingExtension

src/Type/Php/GmpOperatorTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,74 @@
...
+use PHPStan\Type\Type;
+use function in_array;
+
+#[AutowiredService]
+final class GmpOperatorTypeSpecifyingExtension implements OperatorTypeSpecifyingExtension
+{
+
+	public function isOperatorSupported(string $operatorSigil, Type $leftSide, Type $rightSide): bool
+	{
+		if ($leftSide instanceof NeverType || $rightSide instanceof NeverType) {
```
VincentLanglet, line 21: The BC math operator type specifying extension has a check
```
$this->phpVersion->supportsBcMathNumberOperatorOverloading() 
```

Do we need something similar here or does it work on any php version ?

Firehed: Anything since 5.6, according to the RFC that added it (linked in issue). I'm not sure how far back PHPStan support goes and by extension whether that'd be necessary.

Firehed:
Updated to target 2.1 instead of 2.2 per the bot's suggestion. The two failing-as-of-writing tests appear to be failing on other PRs so I don't think they're specific to this change.

I'll check out the other feedback shortly. Thank you for the very quick review!

ondrejmirtes:
Please introduce the changes to InitializerExprTypeResolver in a separate PR - this will force you to create synthetic extensions just for test purposes, and review these changes separately, which I want you to do.

Without this, if we once removed the GMP extensions then this feature would become untested.

After that is merged, then you can take advantage of it with new GMP extensions in a new PR.

Firehed:
Can do. Thanks for the feedback!

Firehed:
Split per feedback: #5226 contains the `InitializerExprTypeResolver` changes with synthetic test extensions. Once that's merged, I'll rebase this PR to only contain the GMP-specific extension (or new separate one if you prefer)

VincentLanglet (commented):
Nice.

Now please again, split this PR in two:
- One with the GmpOperatorTypeSpecifyingExtension 
- One with the unary minus, unary plus and bitwise not operator

Because for the second one, I think it will be a better API to introduce an UnaryOperatorTypeSpecifyingExtension

Firehed:
Gotcha, will do. Missed your message before continuing.

Firehed:
@VincentLanglet  https://github.com/phpstan/phpstan-src/pull/5284 sets up the UnaryOperatorTypeSpecifyingExtension; once that's looking good I'll update the GMP code to build on it instead.

VincentLanglet:
> @VincentLanglet #5284 sets up the UnaryOperatorTypeSpecifyingExtension; once that's looking good I'll update the GMP code to build on it instead.

It's merged, you can now rebase/update this PR :)

Firehed:
Alright @VincentLanglet I've done the rebase again, and now this is purely the GMP stuff leveraging the type overloading stuff from the previous two split-off PRs. It's still pretty much entirely Claude going at it, but it's now feeling pretty tightly-scoped from my perspective. LMK if there's anything else here :)

src/Type/Php/GmpOperatorTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,74 @@
...
+		$rightIsGmp = $gmpType->isSuperTypeOf($rightSide)->yes();
+
+		// At least one side must be GMP
+		if (!$leftIsGmp && !$rightIsGmp) {
+			return false;
+		}
+
+		// The other side must be GMP-compatible (GMP, int, or numeric-string)
+		// GMP operations with incompatible types (like stdClass) will error at runtime
+		return $this->isGmpCompatible($leftSide, $gmpType) && $this->isGmpCompatible($rightSide, $gmpType);
```
VincentLanglet: Rather than not supporting when it's not compatible you could return an ErrorType.

I feel like you could do like BcMathNumberOperatorTypeSpecifyingExtension and
- Not having a isGmpCompatible method
- adding some checks in specifyType and returning an ErrorType

Except this feedback, this PR seems ok to me :)

VincentLanglet (approved):

VincentLanglet:
You just need to fix the build

Firehed:
Ok, I can't for the life of me figure out the mutation testing issues, even after going through setting it up locally and ensuring it passed there. Perhaps it's somehow specific to 8.3/8.4 vs my 8.5? I will probably need an assist here to get it over the line.

I don't think the Larastan issues are related to this PR based on the errors, though I also don't see it failing on the upstream branch. It's a bit tricky to confirm since CI generates an enormous number of jobs (how do you all get such good parallelism btw? I see my own job matrixes cap out _way_ lower, including for OSS projects)

ondrejmirtes:
@Firehed PHPStan org pays for the enterprise plan 😊

VincentLanglet:
There is sometimes false positives about mutation testing and larastan is unrelated indeed

so I would say it's ready @Firehed

Firehed:
Alright @VincentLanglet - moved out of draft state. LMK if you want any other edits.

Thanks for the info on CI @ondrejmirtes. And thanks to both of you for the support and feedback getting this wrapped up!

src/Type/Php/GmpUnaryOperatorTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,35 @@
...
+		if ($operand instanceof NeverType) {
+			return false;
+		}
+
+		if (!in_array($operatorSigil, ['-', '+', '~'], true)) {
+			return false;
+		}
+
+		$gmpType = new ObjectType('GMP');
+		return $gmpType->isSuperTypeOf($operand)->yes();
```
staabm, line 27: the 

```
- return $gmpType->isSuperTypeOf($operand)->yes(); 
+ return !$gmpType->isSuperTypeOf($operand)->no();
```

mutation can be killed by testing with a "maybe GMP operand", like

```php
function maybeGmp(GMP $a, GMP $b) {
	if (rand(0,1)) {
		$a = 5;
	}
	assertType(..., $a + $b));
}
```

---

same for the other extension

staabm:
running PHPStan on gmp code with this PR leads errors like

```
  27     Only numeric types are allowed in *, GMP given on the left side.                                     
         🪪  mul.leftNonNumeric                                                                               
         at gmp-operators.php:27                                                                              
  28     Only numeric types are allowed in /, GMP given on the left side.                                     
         🪪  div.leftNonNumeric                                                                               
         at gmp-operators.php:28                                                                              
  29     Only numeric types are allowed in %, GMP given on the left side.                                     
         🪪  mod.leftNonNumeric                                                                               
         at gmp-operators.php:29                                                                              
  30     Only numeric types are allowed in **, GMP given on the left side.                                    
         🪪  pow.leftNonNumeric                                                                               
         at gmp-operators.php:30                                                                              
  33     Only numeric types are allowed in +, GMP given on the right side.                                    
         🪪  plus.rightNonNumeric                                                                             
         at gmp-operators.php:33                                        

  81     Loose comparison via "!=" between GMP and int is not allowed.                                        
         🪪  notEqual.notAllowed                                                                              
         💡  Use strict comparison via "!==" instead.                                                         
         at gmp-operators.php:81                                                                              
  89     Loose comparison via "==" between int and GMP is not allowed.                                        
         🪪  equal.notAllowed                                                                                 
         💡  Use strict comparison via "===" instead.           
         
  105    Only numeric types are allowed in *, GMP given on the left side.                                     
         🪪  mul.leftNonNumeric                                                                               
         at gmp-operators.php:105                                                                             
  181    Only numeric types are allowed in *, GMP given on the left side.                                     
         🪪  mul.leftNonNumeric                                                                               
         at gmp-operators.php:181
```

repro
```
cp tests/PHPStan/Analyser/nsrt/gmp-operators.php .
php bin/phpstan analyze gmp-operators.php --debug
```

VincentLanglet:
> running PHPStan on gmp code with this PR leads errors like

The error are coming from the strict rules no ?
There is a similar PR https://github.com/phpstan/phpstan-strict-rules/pull/305

I would say that the strict rules will need to be updated once the operator type specifying extension is merged

staabm:
> I would say that the strict rules will need to be updated once the operator type specifying extension is merged

I think we should already prepare the strict-rules PR - before merging here.
(so we don't get a release blocker)

VincentLanglet:
> > I would say that the strict rules will need to be updated once the operator type specifying extension is merged
> 
> I think we should already prepare the strict-rules PR - before merging here. (so we don't get a release blocker)

I don't think it's a release blocker considering this:
https://phpstan.org/r/f47eab0a-5160-4995-b2a6-108a480fb852

The `Only numeric types are allowed in +, GMP given on the left side.` already exists
The inference is wrong

This PR is fixing the second point without touching the first one.

Firehed:
Hm, that actually prompts a question I should have asked earlier - would it make sense to take the whole block of code that prompted me to file the issue as a test case (or, indirectly, is there a preferred way to run this branch on my codebase)? In theory all of the ignores [here](https://github.com/Firehed/webauthn-php/blob/1e403f080a10085657a4f32118e545bed7f54dbb/src/PublicKey/EllipticCurve.php#L143-L165) should be able to go away following this change, but the strict rules may still raise warnings depending on how opinionated they get. I believe it's fully sound and strict mode shouldn't complain, but I recognize it takes a more opinionated stance on things.

I'm happy to throw the robots at whatever makes the most sense here, though my own bandwidth is limited.

VincentLanglet:
> Hm, that actually prompts a question I should have asked earlier - would it make sense to take the whole block of code that prompted me to file the issue as a test case (or, indirectly, is there a preferred way to run this branch on my codebase)? In theory all of the ignores [here](https://github.com/Firehed/webauthn-php/blob/1e403f080a10085657a4f32118e545bed7f54dbb/src/PublicKey/EllipticCurve.php#L143-L165) should be able to go away following this change, but the strict rules may still raise warnings depending on how opinionated they get. I believe it's fully sound and strict mode shouldn't complain, but I recognize it takes a more opinionated stance on things.
> 
> I'm happy to throw the robots at whatever makes the most sense here, though my own bandwidth is limited.

I feel like we have enough test in this PR.
But you can start a PR on phpstan/strict-rules.

cc @staabm I insist on the fact I think we should merge this.
If you look at https://phpstan.org/r/d8777fba-cdd5-4e11-9750-d96b6d460880 there is 
- 4 inferences errors
- 7 strict-rule errors "Only numeric types are allowed"

This PR fixes the 4 inferences errors and does not introduce new error, so it's not a release blocker.

staabm (approved):

staabm:
Thank you @Firehed

PR #5225, opened 2026-03-15
Fix method call on unioned class type analysis is broken

VincentLanglet:
From https://github.com/phpstan/phpstan-src/pull/4960

Fixes phpstan/phpstan#11687
Closes https://github.com/phpstan/phpstan/issues/12562
Fixes https://github.com/phpstan/phpstan/issues/14203

### DIFF
--- a/tests/PHPStan/Analyser/nsrt/bug-14203.php
+++ b/tests/PHPStan/Analyser/nsrt/bug-14203.php
@@ -1,0 +1,90 @@
+<?php // lint >= 8.1
+
+declare(strict_types = 1);
+
+namespace Bug14203;
+
+use function PHPStan\Testing\assertType;
+
+/**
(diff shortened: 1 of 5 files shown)


VincentLanglet:
I think this test is better than https://github.com/phpstan/phpstan-src/pull/5299

tests/PHPStan/Analyser/nsrt/bug-14203.php
```diff
@@ -0,0 +1,90 @@
...
+
+use function PHPStan\Testing\assertType;
+
+/**
+ * @template TKey of array-key
+ * @template TValue
+ */
+class Collection
+{
+    /**
```
staabm, line 15: this file mixes tabs and spaces for indent

staabm (approved):

PR #5226, opened 2026-03-15
Call operator type specifying extensions for bitwise and arithmetic operators

Firehed:
## Summary

This PR ensures `OperatorTypeSpecifyingExtension` implementations are called consistently for all supported binary operators:

- Add extension calls to `getBitwiseAndType`, `getBitwiseOrType`, `getBitwiseXorType`
- Move extension call to the top of `resolveCommonMath` (before integer range optimization)
- Remove the duplicate extension call that was later in `resolveCommonMath`

This is prerequisite work for phpstan/phpstan#14288 (GMP operator type inference), split out per review feedback on #5223.

## Testing approach

Per @ondrejmirtes' feedback:
> "Please introduce the changes to InitializerExprTypeResolver in a separate PR - this will force you to create synthetic extensions just for test purposes... Without this, if we once removed the GMP extensions then this feature would become untested."

This PR creates **synthetic test extensions** independent of any specific type (like GMP):

1. **`TestBitwiseOperand`** - A dummy fixture class
2. **`TestBitwiseOperatorTypeSpecifyingExtension`** - Returns `TestBitwiseOperand` for `&`, `|`, `^` operators
3. Uses existing **`TestDecimal`** + **`TestDecimalOperatorTypeSpecifyingExtension`** for arithmetic operators

The test verifies the resolver correctly invokes extensions for both bitwise and arithmetic operators. This ensures the resolver infrastructure remains tested even if specific extensions (like a future GMP extension) are removed.

## Test plan

- [x] Added `OperatorTypeSpecifyingExtensionTypeInferenceTest` with 8 assertions
- [x] All existing tests pass (11634 tests, 78899 assertions)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Reflection/InitializerExprTypeResolver.php
+++ b/src/Reflection/InitializerExprTypeResolver.php
@@ -986,6 +986,12 @@ final class InitializerExprTypeResolver
 		$leftType = $getTypeCallback($left);
 		$rightType = $getTypeCallback($right);
 
+		$specifiedTypes = $this->operatorTypeSpecifyingExtensionRegistryProvider->getRegistry()
+			->callOperatorTypeSpecifyingExtensions(new BinaryOp\BitwiseAnd($left, $right), $leftType, $rightType);
+		if ($specifiedTypes !== null) {
+			return $specifiedTypes;
+		}
+
 		return $this->getBitwiseAndTypeFromTypes($leftType, $rightType);
 	}
 
@@ -1044,6 +1050,12 @@ final class InitializerExprTypeResolver
 		$leftType = $getTypeCallback($left);
 		$rightType = $getTypeCallback($right);
 
+		$specifiedTypes = $this->operatorTypeSpecifyingExtensionRegistryProvider->getRegistry()
+			->callOperatorTypeSpecifyingExtensions(new BinaryOp\BitwiseOr($left, $right), $leftType, $rightType);
+		if ($specifiedTypes !== null) {
+			return $specifiedTypes;
+		}
+
 		return $this->getBitwiseOrTypeFromTypes($leftType, $rightType);
 	}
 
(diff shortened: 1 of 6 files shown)


Firehed:
I believe the two failing tests are unrelated to this change based on other PRs I'm seeing and their associated error messages.

VincentLanglet (approved):

src/Reflection/InitializerExprTypeResolver.php
```diff
@@ -986,6 +986,12 @@ public function getBitwiseAndType(Expr $left, Expr $right, callable $getTypeCall
 		$leftType = $getTypeCallback($left);
 		$rightType = $getTypeCallback($right);
 
+		$specifiedTypes = $this->operatorTypeSpecifyingExtensionRegistryProvider->getRegistry()
```
staabm, line 989: this change only makes sense when `getBitwiseAndTypeFromTypes` is no longer `public` because otherwise you can by-pass extensions.

but it seems like `getBitwiseAndTypeFromTypes` is not called from anywhere outside this class and since `InitializerExprTypeResolver` is not `@api`-annotated, I think we can reduce visibility for such methods to `private`.

affects multiple methods in this class

tests/PHPStan/Analyser/data/operator-type-specifying-extension.php
```diff
@@ -0,0 +1,55 @@
+<?php declare(strict_types = 1);
+
+namespace OperatorExtensionTest;
+
+use PHPStan\Fixture\TestBitwiseOperand;
+use PHPStan\Fixture\TestDecimal;
+use function PHPStan\Testing\assertType;
+
+// =============================================================================
```
staabm, line 9: misses tests for shift left/right

src/Reflection/InitializerExprTypeResolver.php
ondrejmirtes: Please don’t change this, I’m pretty sure I use these in my refactoring at https://github.com/phpstan/phpstan-src/pull/5224

Firehed:
I can get some extra tests added for bit-shifting.

@staabm & @ondrejmirtes Just to make sure I'm following your comments correctly, should the logic in InitializerExprTypeResolver stay as-is? Sounds like changing method visibility per the original suggestion is out.

staabm:
Leave the visibility as is , right

staabm (approved):

VincentLanglet:
I think we can merge this, since ondrej only request is respected. He will still be able to review this before the release.

VincentLanglet:
Thanks @Firehed you can now rebase and work on https://github.com/phpstan/phpstan-src/pull/5223

Firehed:
Sounds great! I'm not sure if I can get to it tomorrow but hopefully fairly soon if not. TY for the feedback on everything.

PR #5243, opened 2026-03-18
IntersectionType: Prevent duplicate work in isOffsetAccessible()

staabm:
~29% faster in https://github.com/phpstan/phpstan/issues/14319

grafik

----

after this change we are not yet fast, but its a low hanging fruit along the way

### DIFF
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -86,6 +86,8 @@ class IntersectionType implements CompoundType
 
 	private ?TrinaryLogic $isOversizedArray = null;
 
+	private ?TrinaryLogic $isOffsetAccessible = null;
+
 	private ?TrinaryLogic $isIterableAtLeastOnce = null;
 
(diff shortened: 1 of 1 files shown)


VincentLanglet (approved):

VincentLanglet:
I wonder @staabm 

- If this strategy should be done for some other methods of the IntersectionType ? (We could easily do all the TrinaryLogic methods with a refactor)
- Why it's only memoized on IntersectionType (should some other type memoize too ? Either just UnionType which seems similar to Intersection or even all the types....) 
- Should we have a general memoize strategy with a trait ?

WDYT ?

staabm:
this small fix was guided by a blackfire profile which showed this slow path.
I usually add this caching only after we have evidence - and a real world case - that it is slow.

I did not change anything else, because this minimal fix was enough to get the case at hand faster.
IMO adding caching without the proof of a slow path makes things more complex, and might lead to copy/paste of caching into even more layers without a need for it.

profile before this FIX was

grafik

PR #5250, opened 2026-03-19
Makefile: added phpbench target

staabm:
allow running the benchmark locally.

this seems useful, as I recently got 'red' benchmark github action jobs, and want to verify/reproduce locally 

---

adds 2 new makefile targets

`make phpbench-baseline` (re)-generates the baseline
`make phpbench` verifies a recent enough baseline exists and runs the tests, compares them against the baseline

to achieve this we add the phpbench composer dependency at the start and later on remove it again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,10 +195,10 @@ phpbench:
 	@find tests/bench/storage/local-baseline.xml -mtime -60m | grep . || (echo "PHPBench baseline file does not exist or is too old. Regenerate it using 'make phpbench-baseline'." && exit 1)
 	composer require --dev phpbench/phpbench:^1.2.15
 	XDEBUG_MODE=off tests/vendor/bin/phpbench run --file=tests/bench/storage/local-baseline.xml --report=aggregate
-	composer remove --dev phpbench/phpbench --no-interaction
+	composer remove --dev phpbench/phpbench --no-interaction -q
 
 .PHONY: phpbench-baseline
 phpbench-baseline:
 	composer require --dev phpbench/phpbench:^1.2.15
 	XDEBUG_MODE=off tests/vendor/bin/phpbench run --dump-file=tests/bench/storage/local-baseline.xml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -189,3 +189,9 @@ infection:
 		--skip-initial-tests \
 		--ignore-msi-with-no-mutations \
 		--logger-text=php://stdout
+
+.PHONY: phpbench
+phpbench:
+	composer require --dev phpbench/phpbench:^1.2.15
```
VincentLanglet: Why not always adding it as dev dependency ?

staabm: I guess ondrej did it that way because PHP version support for 7.2 in CI would not work.

VincentLanglet (approved):

Makefile
ondrejmirtes: phpbench is already a dependency in tests/composer.json. No need to add it here.

Makefile
```diff
@@ -189,3 +189,16 @@ infection:
...
+.PHONY: phpbench
+phpbench:
+	@find tests/bench/storage/local-baseline.xml -mtime -60m | grep . || (echo "PHPBench baseline file does not exist or is too old. Regenerate it using 'make phpbench-baseline'." && exit 1)
+	composer require --dev phpbench/phpbench:^1.2.15 -q
+	XDEBUG_MODE=off tests/vendor/bin/phpbench run --file=tests/bench/storage/local-baseline.xml --report=aggregate
+	composer remove --dev phpbench/phpbench --no-interaction -q
+
+.PHONY: phpbench-baseline
+phpbench-baseline:
+	composer require --dev phpbench/phpbench:^1.2.15 -q
```
ondrejmirtes, line 202: This should not be here. Use the phpbench from tests/.

Makefile
staabm: good point. will be simplified in https://github.com/phpstan/phpstan-src/pull/5251

PR #5252, opened 2026-03-19
Add non regression test

VincentLanglet:
Both https://github.com/phpstan/phpstan-src/pull/5203 and https://github.com/phpstan/phpstan-src/pull/5189 seems to break the following snippet.

It might be simpler to
- merge this regression test
- close the PRs
- reopen them with the bot

WDYT @staabm ?


staabm:
I had a similar idea. I think we can merge this test.

I would make it smaller though: https://phpstan.org/r/50723b9b-06bc-4d7d-b1e1-5692ca02ca84

(and maybe rebase open PRs instead of starting from square-1 is enough)

staabm:
thanks

PR #5256, opened 2026-03-20
Check constants in parameters

ondrejmirtes:
I prototyped a new pretty useful feature in a couple of hours.

In the end I want this to be gated behind the bleeding edge, but for now it always checks, so we can see the impact on issue-bot and integration tests.

Can you please review this and bring it over the finish line @staabm and @VincentLanglet?

Closes https://github.com/phpstan/phpstan/issues/12850


VincentLanglet (commented):
Pretty cool, I didn't have the time to fully look at it ; you finished it quickly

ondrejmirtes:
You can still go through it to see if you find something to change

resources/constantToFunctionParameterMap.php
```diff
@@ -0,0 +1,2451 @@
...
+ *   'constants'  => list of constant names valid for this parameter
+ *   'exclusiveGroups' => (optional, bitmask only) groups of constants that are mutually exclusive
+ */
+return [
+
+	// ————————————————————————————————————————————
+	// JSON
+	// ————————————————————————————————————————————
+
+	'json_encode' => [
```
staabm, line 20: how was this map generated? is it AI based? is it generated on other metadata (e.g. php8stubs or similar)?

is this something which can/could be typed via phpdoc instead (e.g. in stub files)?

tests/PHPStan/Rules/Functions/data/constant-parameter-check-call-user-func.php
```diff
@@ -0,0 +1,9 @@
+<?php // lint >= 8.0
+
+namespace ConstantParameterCheckCallUserFunc;
+
+// call_user_func with correct constant
+call_user_func('json_encode', [], JSON_PRETTY_PRINT);
```
staabm, line 6: looking at this, made me think that we could also support splat operator invocations:

https://phpstan.org/r/ed2c6e70-fef6-4fc1-93c9-3a3608223d9b

https://3v4l.org/8otIb

resources/constantToFunctionParameterMap.php
ondrejmirtes: It’s Claude looking at PHP documentation and some relevant metadata in phpstorm-stubs describing this. The parameter names are proper from php-8-stubs, that’s why it’s behind “named arguments” support in PhpVersions.

Right now it’s only for built-in functions/methods/constants. If we find a use case for PHPDocs we can add it.

tests/PHPStan/Rules/Functions/data/constant-parameter-check-call-user-func.php
ondrejmirtes: This feature is purely AST based.

staabm:
it seems starting with this PR on 2.2.x a test started failling: https://github.com/phpstan/phpstan-src/actions/runs/23346042957/job/67911556548

PR #5258, opened 2026-03-20
Fix incorrect narrowing of nested array after assignment

staabm:
closes https://github.com/phpstan/phpstan/issues/13857
closes https://github.com/phpstan/phpstan/issues/10089

### DIFF
--- a/src/Analyser/ExprHandler/AssignHandler.php
+++ b/src/Analyser/ExprHandler/AssignHandler.php
@@ -938,7 +938,7 @@ final class AssignHandler implements ExprHandler
 
 	/**
 	 * @param list<ArrayDimFetch> $dimFetchStack
-	 * @param list<array{Type|null, ArrayDimFetch}> $offsetTypes
+	 * @param non-empty-list<array{Type|null, ArrayDimFetch}> $offsetTypes
 	 *
 	 * @return array{Type, list<array{Expr, Type}>}
 	 */
--- a/src/Analyser/ExprHandler/AssignHandler.php
+++ b/src/Analyser/ExprHandler/AssignHandler.php
@@ -947,11 +947,11 @@ final class AssignHandler implements ExprHandler
 		$originalValueToWrite = $valueToWrite;
 
 		$offsetValueTypeStack = [$offsetValueType];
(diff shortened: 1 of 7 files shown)


tests/PHPStan/Analyser/nsrt/pr-4390.php
```diff
@@ -13,6 +13,6 @@ function (string $s): void {
 		}
 	}
 
-	assertType('non-empty-array<int<0, 9>, non-empty-array<int<0, 9>, string>>', $locations);
```
staabm, line 16: try to fix the implementation so we don't need to adjust test expectations in this file

src/Analyser/ExprHandler/AssignHandler.php
```diff
@@ -1010,7 +1013,22 @@ private function produceArrayDimFetchAssignValueToWrite(array $dimFetchStack, ar
 				}
 
 			} else {
-				$valueToWrite = $offsetValueType->setOffsetValueType($offsetType, $valueToWrite, $i === 0);
+				$unionValues = false;
+				if ($i === 0) {
+					$unionValues = true;
+				} elseif (
+					$overwritesExistingOffset === true
```
VincentLanglet: `$overwritesExistingOffset === true` can be simplified to `$overwritesExistingOffset`

VincentLanglet:
The condition 
```
if ($i === 0) {
					$unionValues = true;
				} elseif (
					$overwritesExistingOffset === true
					&& $i === count($offsetTypes) - 1
					&&
						(
							$originalValueToWrite->isConstantScalarValue()->yes()
							|| !$offsetValueType->getIterableValueType()->isSuperTypeOf($valueToWrite)->yes()
						)
				) {
					$unionValues = true;
				}
```
is unclear to me @staabm I feel like a comment would be nice ; do you want to write one (since you worked on the PR) or let the bot generate one ?

VincentLanglet (approved):
The renaming makes the code clearer to me

staabm:
I am 100% sure there is a better way to fix this.. but I spent already days on this (and related phpstan-bot PRs).
I will try to generalize and improve the logic in a future PR, but will merge it now as it still improves the status-quo

PR #5268, opened 2026-03-21
Revert "Fix phpstan/phpstan#13629: False report for 2.1.30 release "arrayValues.list" ("Parameter X of array_values is already a list, call has no effect") (#5264)"

staabm:
This reverts commit cd904bba55b4345d2054b0e743f717e954b31290 which broke the build


staabm:
I think this fix got influenced/had bad interaction with the other fix in https://github.com/phpstan/phpstan-src/pull/5263 which got merged a few minutes before

PR #5271, opened 2026-03-21
Fix "Parameter X of array_values is already a list, call has no effect"-false positive

staabm:
closes https://github.com/phpstan/phpstan/issues/13629

### DIFF
--- a/src/Type/IntersectionType.php
+++ b/src/Type/IntersectionType.php
@@ -984,7 +984,7 @@ class IntersectionType implements CompoundType
 		if (
 			$this->isList()->yes()
 			&& $offsetType !== null
-			&& $offsetType->isInteger()->yes()
+			&& $offsetType->toArrayKey()->isInteger()->yes()
 			&& $this->getIterableValueType()->isArray()->yes()
 		) {
 			$result = TypeCombinator::intersect($result, new AccessoryArrayListType());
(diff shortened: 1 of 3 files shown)


staabm:
integration test failures are unrelated -> https://github.com/phpstan/phpstan/pull/14334

src/Type/IntersectionType.php
```diff
@@ -981,7 +981,12 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
 			}
 		}
 
-		if ($this->isList()->yes() && $this->getIterableValueType()->isArray()->yes()) {
+		if (
```
staabm, line 984: In the long term, this IF should not be here I think.

it feels like a workaround from a time in which `setExistingOffsetValueType` did not yet exist.

src/Type/IntersectionType.php
```diff
@@ -981,7 +981,12 @@ public function setOffsetValueType(?Type $offsetType, Type $valueType, bool $uni
...
+			$this->isList()->yes()
+			&& $offsetType !== null
+			&& $offsetType->isInteger()->yes()
```
VincentLanglet: Should it be toArrayKey() ?

src/Type/IntersectionType.php
VincentLanglet: Should we try removing this if and find the right place to fix ?

Cause setting an Int key to a List shouldnt always keep the list

staabm: I don't want to change too many array related stuff atm - I don't want make ondrejs' work harder than necessary.

VincentLanglet (approved):

PR #5279, opened 2026-03-23
decimal-int-string and non-decimal-int-string

ondrejmirtes:


src/Type/Accessory/AccessoryDecimalIntegerStringType.php
```diff
@@ -0,0 +1,446 @@
...
+ * These are guaranteed to be cast to an integer in an array key.
+ * Examples of constant values covered by this type: "0", "1", "1234"
+ *
+ * When inverse=true, this represents strings containing non-decimal integers and other text.
+ * These are guaranteed to stay as string in an array key.
+ * Examples of constant values covered by this type: "+1", "00", "18E+3", "1.2", "1,3", "foo"
+ *
+ * @api
+ */
+class AccessoryDecimalIntegerStringType implements CompoundType, AccessoryType
```
staabm, line 50: in a followup we might use this new accessory for regex type inference for `[0-9]`, `:digit:` or `\d`

ondrejmirtes: I realized this includes negative numbers as well.

PR #5286, opened 2026-03-23
Support call_user_func_array

VincentLanglet:
Extracted from https://github.com/phpstan/phpstan-src/pull/5282

This copy the existing behavior of `call_user_func` but for `call_user_func_array`.

Closes https://github.com/phpstan/phpstan/issues/5942
Closes https://github.com/phpstan/phpstan/issues/5934

### DIFF
--- a/src/Analyser/ArgumentsNormalizer.php
+++ b/src/Analyser/ArgumentsNormalizer.php
@@ -22,7 +22,9 @@ use function array_key_exists;
 use function array_keys;
 use function array_values;
 use function count;
+use function key;
 use function ksort;
+use function is_string;
 use function max;
 use function sprintf;
 
@@ -120,16 +122,17 @@ final class ArgumentsNormalizer
 				}
 			}
 
-			if ($argsArrayArg === null) {
-				if ($arg->name === null && $i === 1) {
-					$argsArrayArg = $arg;
-					continue;
-				}
-				if ($arg->name !== null && $arg->name->toString() === 'args') {
-					$argsArrayArg = $arg;
-					continue;
-				}
+			if ($argsArrayArg !== null) {
+				continue;
 			}
+			if ($arg->name === null && $i === 1) {
+				$argsArrayArg = $arg;
+				continue;
+			}
+			if ($arg->name === null || $arg->name->toString() !== 'args') {
+				continue;
+			}
+			$argsArrayArg = $arg;
 		}
 
 		if ($callbackArg === null || $argsArrayArg === null) {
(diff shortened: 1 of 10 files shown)


src/Analyser/ArgumentsNormalizer.php
```diff
@@ -89,6 +95,102 @@ public static function reorderCallUserFuncArguments(
...
+		}
+
+		if (!$argsArrayArg->value instanceof Array_) {
+			return null;
+		}
+
+		$calledOnType = $scope->getType($callbackArg->value);
+		if (!$calledOnType->isCallable()->yes()) {
+			return null;
+		}
```
staabm: since the following loop might return early, we could move this 4 lines nearer to the usage in line 174 to prevent unnecessary work

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
```diff
@@ -1128,7 +1128,7 @@ public function testCallUserFuncArray(): void
 				],
 			];
 		}
-		$this->analyse([__DIR__ . '/data/call-user-func-array.php'], $errors);
+		$this->analyse([__DIR__ . '/data/call-user-func-array-named-args.php'], $errors);
 	}
```
staabm, line 1132: is `call-user-func-array.php` no longer tested in this rule?

staabm:
issue-bot shows a new error with https://github.com/phpstan/phpstan/issues/8214 .. might indicate we miss splat operator support...?

VincentLanglet:
> issue-bot shows a new error with [phpstan/phpstan#8214](https://github.com/phpstan/phpstan/issues/8214) .. might indicate we miss splat operator support...?

The phpdoc of the snippet is
```
@param Closure(mixed...):T $createState Create the state
@phpstan-param Closure(mixed,mixed,mixed,mixed,mixed,mixed):T $createState
```
so the error is legit

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
VincentLanglet: I just renamed the initial `call-user-func-array.php` file

staabm: interessting.. the PR does not include a file which got renamed?

staabm (approved):

tests/PHPStan/Rules/Functions/CallToFunctionParametersRuleTest.php
VincentLanglet: image

I renamed `call-user-func-array` and recreated a new `call-user-func-array`.

Git consider `call-user-func-array-named-args` new, and the other as modified

PR #5294, opened 2026-03-24
Display the correct number of processes used, when using '--debug'.

esthezia:


ondrejmirtes:
That’s not how I’d do it. The scheduler should only be called with !$debug. Meaning that the DiagnoseExtension should not print anything at all.

Also add e2e tests into e2e-tests.yml that asserts the output (or lack thereof).

esthezia:
> The scheduler should only be called with !$debug. Meaning that the DiagnoseExtension should not print anything at all.

@ondrejmirtes What do you mean? From what I can see, the DiagnoseExtension prints this information if `-vvv` is used, whether `--debug` is given or not.

Added e2e tests.

.github/workflows/e2e-tests.yml
```diff
@@ -436,6 +436,16 @@ jobs:
               cd e2e/bug-14036
               composer install
               ../../bin/phpstan analyze
+          - script: |
+              cd e2e/vvv-with-debug
+              OUTPUT=$(../../bin/phpstan analyze -l 0 -vvv --debug test.php 2>&1)
+              echo "$OUTPUT"
+              ../bashunit -a contains '# of spawned processes:    1' "$OUTPUT"
```
ondrejmirtes: As I said in my earlier preview, this should be missing completely when running with `--debug`. Scheduler should not be called at all.

ondrejmirtes (changes requested):

.github/workflows/e2e-tests.yml
esthezia: I don't understand. Why shouldn't the Scheduler be called? Because I _can_ run PHPStan with `--debug` _and_ `-vvv`.
Granted, I am new to PHPStan's codebase, but I see that you _can_ run PHPStan with `--debug` and `-vvv` and they're independent. `--debug` does its job, and `-vvv` does _its_ job, independent of each other. Am I missing something?

ondrejmirtes: This is roughly the kind of change I want you to make in AnalyserRunner:

```diff
diff --git a/src/Command/AnalyserRunner.php b/src/Command/AnalyserRunner.php
index c452ba4976..c44b14ba6a 100644
--- a/src/Command/AnalyserRunner.php
+++ b/src/Command/AnalyserRunner.php
@@ -71,6 +71,7 @@ final class AnalyserRunner
 			);
 		}
 
+		if (!$debug && $allowParallel) {
 			$schedule = $this->scheduler->scheduleWork($this->cpuCoreCounter->getNumberOfCpuCores(), $files);
 			$mainScript = null;
 			if (isset($_SERVER['argv'][0]) && is_file($_SERVER['argv'][0])) {
@@ -78,9 +79,7 @@ final class AnalyserRunner
 			}
 
 			if (
-			!$debug
-			&& $allowParallel
-			&& function_exists('proc_open')
+				function_exists('proc_open')
 				&& $mainScript !== null
 				&& $schedule->getNumberOfProcesses() > 0
 			) {
@@ -96,6 +95,7 @@ final class AnalyserRunner
 				}
 				return $result;
 			}
+		}
 
 		return $this->analyser->analyse(
 			$this->switchTmpFile($files, $insteadOfFile, $tmpFile),

```

esthezia: I see now. So you want to not display the parallel processing information at all when using `--debug`, because it's not relevant in this case. True. Made the necessary changes.

.github/workflows/e2e-tests.yml
```diff
@@ -436,6 +436,16 @@ jobs:
...
+          - script: |
+              cd e2e/vvv-with-debug
+              OUTPUT=$(../../bin/phpstan analyze -l 0 -vvv --debug test.php 2>&1)
+              echo "$OUTPUT"
+              ../bashunit -a not_contains 'Parallel processing scheduler' "$OUTPUT"
+          - script: |
+              cd e2e/vvv-with-debug
+              OUTPUT=$(../../bin/phpstan analyze -l 0 -vvv --debug --configuration parallel.neon test.php 2>&1)
+              echo "$OUTPUT"
+              ../bashunit -a not_contains 'Parallel processing scheduler' "$OUTPUT"
```
ondrejmirtes, line 448: Would be nice to also have a test with `contains` when you run without `--debug`

src/Command/AnalyserRunner.php
```diff
@@ -71,30 +71,27 @@ public function runAnalyser(
...
-			$loop = new StreamSelectLoop();
-			$result = null;
-			$promise = $this->parallelAnalyser->analyse($loop, $schedule, $mainScript, $postFileCallback, $projectConfigFile, $tmpFile, $insteadOfFile, $input, null);
-			$promise->then(static function (AnalyserResult $tmp) use (&$result): void {
-				$result = $tmp;
-			});
-			$loop->run();
-			if ($result === null) {
-				throw new ShouldNotHappenException();
+			if (($mainScript !== null) && ($schedule->getNumberOfProcesses() > 0)) {
```
ondrejmirtes: Too many parentheses


```suggestion
			if ($mainScript !== null && $schedule->getNumberOfProcesses() > 0) {
```

ondrejmirtes (changes requested):

.github/workflows/e2e-tests.yml
esthezia: Correct. I added it in.

src/Command/AnalyserRunner.php
esthezia: Generally speaking, it's best practice, and recommended, to enforce the order of precedence with parentheses, to ensure no bugs can sneak in. But I'm sure you're aware of it when choosing this coding standard.

Made the change.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

esthezia:
Perfect! Thank _you_! 🙏

esthezia:
I want to add, for documentation purposes, that eventually, the output about parallel processing was removed completely when using `--debug`, as it's not relevant in this case, because `--debug` disables parallel processing. It will only be displayed when using `-vvv` without `--debug`.

PR #5296, opened 2026-03-25
Fix false positive match.alwaysTrue with consecutive match(true) expressions

gnutix:
Fixes https://github.com/phpstan/phpstan/issues/14368

## Summary

- After an exhaustive `match(true)` with a throwing default arm, the post-match scope carried conditional expressions encoding relationships between arm-narrowed types (e.g. "if `equals(A)` is false, then `equals(B)` is true")
- When a subsequent `match(true)` filtered by falsey for the first arm, these conditional expressions activated and incorrectly narrowed the second arm's condition to `true`, causing a false positive `match.alwaysTrue`
- The fix restores the pre-match conditional expressions after the arm body scope merge, preventing match-internal type relationships from leaking into the post-match scope

## Root cause

PR #5066 (fix #9601) changed `MatchHandler` to exclude always-terminating arms (e.g. `default => throw ...`) from the arm body scope merge. This is correct for type resolution, but the `mergeWith` call also creates `conditionalExpressions` that encode relationships between the narrowed types of different arms. These conditional expressions survived into the post-match scope and were activated by subsequent code that narrowed the same expressions.

### DIFF
--- a/src/Analyser/ExprHandler/MatchHandler.php
+++ b/src/Analyser/ExprHandler/MatchHandler.php
@@ -454,11 +454,18 @@ final class MatchHandler implements ExprHandler
 		}
 
 		if ($isExhaustive) {
+			$preMatchConditionalExpressions = $scope->getConditionalExpressions();
 			$armBodyFinalScope = null;
 			foreach ($armBodyScopes as $armBodyScope) {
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
No fix is needed, this is expected behaviour.

PR #4752, opened 2026-01-13
Fix type alias resolution in template bound

ondrejmirtes:
Closes https://github.com/phpstan/phpstan/issues/11314 Closes https://github.com/phpstan/phpstan/issues/7152 Closes https://github.com/phpstan/phpstan/issues/13332

PR #5304, opened 2026-03-26
Resolve generic parameter defaults that reference other template parameters

JanTvrdik:
## Summary

When a template parameter has a default referencing another template parameter (e.g. `@template DO of EI = EI`), the default was left as an unresolved TemplateType instead of being substituted with the concrete type.

### Root cause

`TemplateTypeHelper::resolveTemplateTypes` returns resolved values (`$newType`) without re-invoking the callback. When the standins map has `DO → TemplateType(EI)` and `EI → MoneyValue`, resolving `DO` yields `TemplateType(EI)` — but that is never resolved further to `MoneyValue`. This is because TypeTraverser's `$traverse` traverses **inner types** (bound, default) but does not re-invoke the callback on the type itself, so `$traverse($newType)` wouldn't help either.

### Fix

Single-point fix in `ClassReflection::typeMapFromList` — the gateway where all type arguments enter the standins map (active template type map). After building the map, any entry that is a TemplateType referencing another entry in the same class scope is resolved to that entry's concrete type.

**Why `typeMapFromList` and not elsewhere:**
- It's the **single point** through which all standins flow (via `GenericObjectType::getClassReflection()` → `withTypes()`)
- No changes needed in `TypeNodeResolver`, `resolveTemplateTypes`, or `FileTypeMapper`
- Minimal, 8-line change — just a map lookup, no TypeTraverser needed
- No risk to existing `resolveTemplateTypes` behavior

### Example

```php
/**
 * @template-contravariant DI
 * @template-contravariant EI
 * @template-covariant DO of EI = EI
 * @template-covariant EO of DI = DI
 */
interface Codec {
    /** @param DI $data @return DO */
    public function decode(mixed $data): mixed;
    /** @param EI $data @return EO */
    public function encode(mixed $data): mixed;
}

/** @param Codec<array{currency: string, cents: int}, MoneyValue> $codec */
function test(Codec $codec): void {
    $codec->decode([...]); // Before: EI (unresolved), After: MoneyValue
    $codec->encode(new MoneyValue(...)); // Before: DI (unresolved), After: array{currency: string, cents: int}
}
```

## Test plan

- [x] New test `tests/PHPStan/Analyser/nsrt/template-default-referring-other.php` passes
- [x] All existing `template-default` tests pass
- [x] Full NodeScopeResolverTest (1480 tests) — no regressions
- [x] Rules tests (2917 tests) — no regressions
- [x] Generics rules tests (41 tests) — all pass

Co-Authored-By: Claude Code

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -40,6 +40,7 @@ use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\FileTypeMapper;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\Generic\TemplateTypeFactory;
 use PHPStan\Type\Generic\TemplateTypeHelper;
 use PHPStan\Type\Generic\TemplateTypeMap;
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
I don't love this logic, I feel like this should be coming out clean out of FileTypeMapper/PhpDocNodeResolver (the correct default type being in TemplateTag).

I have a couple of PRs that tried to achieve something similar (https://github.com/phpstan/phpstan-src/pull/4753 + https://github.com/phpstan/phpstan-src/pull/4752) but they're not finished.

JanTvrdik:
@ondrejmirtes Claude has considered your comment and has chosen much simpler approach.

ondrejmirtes:
Perfect 👍

PR #5349, opened 2026-03-30
Fix throw on failure with require/force array

VincentLanglet:
Some findings when working on https://github.com/phpstan/phpstan-src/pull/5315


src/Type/Php/FilterFunctionReturnTypeHelper.php
```diff
@@ -197,9 +197,14 @@ public function getType(Type $inputType, ?Type $filterType, ?Type $flagsType): T
 			}
 		}
 
+		$throwOnFailure = $this->hasFlag('FILTER_THROW_ON_FAILURE', $flagsType)->yes();
```
staabm: this condition is used a few lines above another time.
I think this can also use `if ($throwOnFailure) {`

staabm (approved):

PR #5364, opened 2026-03-31
Add external file dependency tracking for incremental cache invalidation

janedbal:
## Summary

- Introduces `ExternalFileDependencyRegistrar`, a new `@api` service that allows extensions to declare that an analyzed file depends on an external (non-analyzed) file
- When that external file changes, only the dependent analyzed files are re-analyzed instead of triggering full cache invalidation
- This is an alternative to `ResultCacheMetaExtension` for cases where external data changes should cause surgical re-analysis of affected files only, not full invalidation

## Motivation

`ResultCacheMetaExtension` is an all-or-nothing mechanism — when any extension's hash changes, the entire result cache is invalidated. phpstan-symfony uses this to track Symfony DI container changes, but any service definition change (even trivially adding a dependency via autowiring) causes full re-analysis of the entire project. On large codebases this is devastating.

This PR adds a granular alternative: extensions register per-file dependencies on external files during analysis. When those files change, only the affected analyzed files (and their dependents) are re-analyzed.

See companion PR: phpstan/phpstan-symfony#478

## Changes

- New `ExternalFileDependencyRegistrar` service (`@api`) with `add(string $externalFilePath)` method
- `FileAnalyserResult` / `AnalyserResult` carry external dependencies through the pipeline
- `WorkerCommand` / `ParallelAnalyser` serialize and aggregate external deps across parallel workers
- `ResultCacheManager`:
  - `restore()`: checks external file hashes, adds dependent files to incremental re-analysis list
  - `save()`: inverts and stores external dependencies with file hashes
  - `process()`: merges cached + fresh external dependencies

## Test plan

- [x] All 11,719 existing tests pass
- [ ] Add dedicated tests for external dependency cache invalidation

Co-Authored-By: Claude Code

### DIFF
--- a/src/Analyser/Analyser.php
+++ b/src/Analyser/Analyser.php
@@ -75,6 +75,7 @@ final class Analyser
 		$dependencies = [];
 		$usedTraitDependencies = [];
 		$exportedNodes = [];
+		$externalFileDependencies = [];
 		foreach ($files as $file) {
 			if ($preFileCallback !== null) {
 				$preFileCallback($file);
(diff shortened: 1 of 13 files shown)


ondrejmirtes:
Smells a lot of like vibecoding, I don't like the API. I think a special Collector, especially now when we have CollectedDataEmitter on 2.2.x, might work better.

janedbal:
> I think a special Collector might work better

Do you mean the phpstan-symfony wont call any `addExternalDependency(xml)` upon `Container::get()` calls, but we rather create separate special collector that finds those usages and add those to the dependency graph somehow?

ondrejmirtes:
I don't know. I posted https://github.com/phpstan/phpstan-symfony/issues/455#issuecomment-4161012853 after some thinking, this needs deeper analysis and a different idea.

PR #5366, opened 2026-03-31
Fix phpstan/phpstan#14413: Narrowing generic union via ::class comparison discards type parameters

mhert:
## Summary

Narrowing a generic union like `Cat<string>|Dog<string>` via `$a::class === Cat::class` discarded the generic type parameters — the inferred type became plain `Cat` instead of `Cat<string>`. This caused downstream type errors when accessing generic methods on the narrowed variable.

## Motivation

When using `::class` comparisons to narrow generic union types, PHPStan's `TypeSpecifier::resolveNormalizedIdentical()` constructed a plain `ObjectType` without consulting the variable's known type for generic information. This meant that code like:

```php
/** @param Cat<string>|Dog<string> $a */
function foo(Animal $a): void {
    if ($a::class === Cat::class) {
        $a->value(); // was inferred as mixed instead of string
    }
}
```

lost all generic type information after narrowing.

## Changes

- **`src/Analyser/TypeSpecifier.php`** — Extracted the duplicated `$a::class === 'Foo'` / `'Foo' === $a::class` handling into a single `resolveClassStringComparison()` helper. Added generic-aware narrowing that infers template type parameters from the variable's current type via `getAncestorWithClassName()` + `inferTemplateTypes()`.

## Test Cases

- **`tests/PHPStan/Analyser/nsrt/class-string-generic-narrowing.php`** — Multiple scenarios:
  - Union match preserves generics (`Cat<string>|Dog<string>` → `Cat<string>` in match arm)
  - If/else `::class` comparison preserves generics
  - Mirror case (`Cat::class === $a::class`) preserves generics
  - Match with method call on narrowed generic type
  - Non-matching class comparison yields `*NEVER*`

Fixes https://github.com/phpstan/phpstan/issues/14413

Co-Authored-By: Claude Code

### DIFF
--- a/src/Analyser/TypeSpecifier.php
+++ b/src/Analyser/TypeSpecifier.php
@@ -58,6 +58,7 @@ use PHPStan\Type\ConstantScalarType;
 use PHPStan\Type\FloatType;
 use PHPStan\Type\FunctionTypeSpecifyingExtension;
 use PHPStan\Type\Generic\GenericClassStringType;
+use PHPStan\Type\Generic\GenericObjectType;
 use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\Generic\TemplateTypeHelper;
 use PHPStan\Type\Generic\TemplateTypeVariance;
@@ -87,6 +88,7 @@ use function array_map;
 use function array_merge;
 use function array_reverse;
 use function array_shift;
+use function array_values;
 use function count;
 use function in_array;
 use function is_string;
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Please open an issue with a reproducing code example first.

mhert:
Friendly ping for review when you get a chance 😊

staabm:
running this PR on the [snippet given in the bug report](https://phpstan.org/r/d12de998-b3cd-49ec-992c-bf34b1601e8a) I don't see any error beeing fixed.

the test added with this PR is already green without the changes in src/

mhert:
I'm sorry, that was my misstake. I fixed it, so you can check again.

Thanks for looking into!

PR #5367, opened 2026-03-31
fix: generic sealed class-string match exhaustiveness for ::class comparisons

mhert:
## Summary

`GenericClassStringType::tryRemove()` did not correctly handle `@phpstan-sealed` hierarchies when the sealed class has generic type parameters (`@template`). Match expressions on `$foo::class` falsely reported "Match expression does not handle remaining values" even when all allowed subtypes were covered.

Follow-up to https://github.com/phpstan/phpstan-src/pull/5305

## Motivation

The fix in #5305 passed `$generic` (e.g. `FooCov<string>`) directly to `TypeCombinator::remove()`, which couldn't correctly subtract sealed subtypes from a generic type. This meant sealed exhaustiveness still failed when the sealed class had template parameters.

## Changes

- **`src/Type/Generic/GenericClassStringType.php`** — In the sealed-aware removal path of `tryRemove()`, construct a plain `ObjectType` (with the existing subtracted type carried over) instead of passing the generic type directly to `TypeCombinator::remove()`. This allows sealed subtraction to work correctly regardless of template parameters.

## Test Cases

- **`tests/PHPStan/Rules/Comparison/data/match-generic-sealed-class-string.php`** — Two scenarios:
  - Covariant template sealed hierarchy (`FooCov<T>` sealed to `BarCov|BazCov`) — exhaustive match produces no error
  - Invariant template with covariant param (`FooInv<T>` sealed to `BarInv|BazInv`) — exhaustive match produces no error
- **`tests/PHPStan/Rules/Comparison/MatchExpressionRuleTest.php`** — Added `testGenericSealedClassStringMatch()` rule test method.

### DIFF
--- a/src/Type/Generic/GenericClassStringType.php
+++ b/src/Type/Generic/GenericClassStringType.php
@@ -18,6 +18,7 @@ use PHPStan\Type\ObjectType;
 use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\StaticType;
 use PHPStan\Type\StringType;
+use PHPStan\Type\SubtractableType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\UnionType;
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Please open a bug report first.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-src/pull/5369

PR #5325, opened 2026-03-29
Add fast path to TypeCombinator::union() for trivial 2-type cases

SanderMuller:
## Problem

`TypeCombinator::union()` is one of the most called methods in PHPStan. Profiling shows 30k-66k calls per file on heavy benchmarks. Every call goes through the full normalization pipeline: flattening nested unions, classifying scalars/arrays/enums, sorting integer ranges, O(n^2) deduplication, and array processing.

Many of these calls are trivial 2-type cases where the result is immediately known:
- `union(NeverType, X)` — result is `X` (never is the identity element)
- `union(X, MixedType)` — result is `MixedType` (mixed absorbs everything)
- `union(X, X)` — result is `X` (same object passed twice)

These patterns are common in type narrowing, scope merging, and conditional type resolution.

## Solution

Add a fast path before the main normalization loop that handles exactly these three cases when `$typesCount === 2`. Each check is a cheap `instanceof` or identity comparison.

The checks mirror logic that already exists deeper in the method (never removal at line 199, mixed short-circuit at line 191, dedup at line 350) but avoids all the setup work.

## Impact

Benchmarked on the 5 slowest files from `tests/bench/data/`, measured in isolation (without other optimizations):

**9,363ms → 7,893ms (-15.7%)**

All files benefit roughly equally since `union()` is called pervasively.

PR #5374, opened 2026-04-01
Merge NeverType pre-scan into existing flattening loop

SanderMuller:
Follow-up to #5325 and fe1df1aa0eb90711dfdf97bbb40245e89c8dcf45.

The N>2 pre-scan and the flattening loop both iterate all types checking for MixedType and NeverType. This removes the pre-scan and counts NeverTypes during the existing flattening loop instead. After the loop, a single bulk filter removes them all, replacing the per-element `array_splice` that was there before.

The `$hasUnionOrBenevolent` guard is no longer needed. UnionTypes get flattened first, then the post-loop filter removes any NeverTypes from the result.

The escaped mutant on `!$types[0]->isArray()->yes()` vs `$types[0]->isArray()->no()` is pre-existing — that line is unchanged from before this PR.

### DIFF
--- a/src/Type/TypeCombinator.php
+++ b/src/Type/TypeCombinator.php
@@ -188,58 +188,12 @@ final class TypeCombinator
 			}
 		}
 
-		// Fast path for N>2: strip implicit NeverTypes and short-circuit on mixed
-		if ($typesCount > 2) {
-			$neverCount = 0;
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
It’s nice, thank you!

PR #5390, opened 2026-04-02
Respect sys_temp_dir from the parent process

theodorejb:
Addresses phpstan/phpstan#13929 without the bug that caused #4699 to be reverted.

Closes phpstan/phpstan#14093

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -451,6 +451,17 @@ jobs:
               OUTPUT=$(../../bin/phpstan analyze -l 0 -vvv test.php 2>&1)
               echo "$OUTPUT"
               ../bashunit -a contains 'Parallel processing scheduler' "$OUTPUT"
+          - script: |
+              cd e2e/bug-14093
+              # Use a relative sys_temp_dir so the value starts with a letter (not /), just like Windows
+              # paths starting with a drive letter. If the sys_temp_dir value isn't correctly quoted,
+              # PHP evaluates ~ as a bitwise NOT rather than a literal character, truncating the path.
+              mkdir tmp~1
+              php -d "sys_temp_dir='tmp~1'" ../../bin/phpstan analyze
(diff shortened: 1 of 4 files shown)


staabm:
Can we e2e test the regression bug which gets fixed?

theodorejb:
@staabm I added an E2E test now, which I confirmed fails with the reverted code in #4699, but passes with this PR. Let me know if there's anything I need to change, since this is my first time contributing to PHPStan!

staabm:
@dktapps could you double check this works for you?

staabm (approved):
verified the test. it works for me.

thank you

VincentLanglet (approved):

.github/workflows/e2e-tests.yml
```diff
@@ -451,6 +451,17 @@ jobs:
               OUTPUT=$(../../bin/phpstan analyze -l 0 -vvv test.php 2>&1)
               echo "$OUTPUT"
               ../bashunit -a contains 'Parallel processing scheduler' "$OUTPUT"
+          - script: |
+              cd e2e/bug-14093
+              # Use a relative sys_temp_dir so the value starts with a letter (not /), just like Windows
+              # paths starting with a drive letter. If the sys_temp_dir value isn't correctly quoted,
+              # PHP evaluates ~ as a bitwise NOT rather than a literal character, truncating the path.
+              mkdir tmp~1
```
mvorisek, line 459: Might we worth testing also characters like `"`, `'`, `$`, `(` and ` `.

Also, I think this E2E test will run on linux only, so E2E test for Windows might be helpful too.

theodorejb: The bug isn't Windows-specific (just much more likely to manifest on Windows due to `sys_temp_dir` starting with a drive letter by default). But I can also add a Windows E2E test if that's desired.

staabm:
Thank you!

PR #5396, opened 2026-04-04
Do fast check first

VincentLanglet:
While working on https://github.com/phpstan/phpstan-src/pull/5394 I found this

Also, this helper rely a lot on IsSuperTypeOf result but then transform checks into boolean and lose the reason.
I feel like it would be great to keep the reasons to add them as tips of the errors.


staabm (commented):
Can we come up with a test showing a missing tip?

VincentLanglet:
> Can we come up with a test showing a missing tip?

This will be in another PR, here I don't add the tip yet and just improve the perf with a fast check

staabm (approved):

PR #5399, opened 2026-04-04
Fix ObjectType::equals

VincentLanglet:
https://github.com/phpstan/phpstan-src/pull/5398 shows the behavior before

Closes https://github.com/phpstan/phpstan/issues/3028
Closes https://github.com/phpstan/phpstan/issues/14429
Closes https://github.com/phpstan/phpstan/issues/5298

Failures:
- [x] nextra/orm https://github.com/phpstan/phpstan-src/actions/runs/23979232001/job/69941215879?pr=5399 => This seems justified to me, as proven by the non regression test added ; Prior to this fix we had an inconsistency in the error reported https://github.com/phpstan/phpstan-src/pull/5398

- [x] rector/downgrade https://github.com/phpstan/phpstan-src/actions/runs/23979232001/job/69941215795?pr=5399 => Explain by https://github.com/phpstan/phpstan-src/pull/5399#issuecomment-4187117573

- [x] rector https://github.com/phpstan/phpstan-src/actions/runs/23979232001/job/69941215804?pr=5399 => Explain by https://github.com/phpstan/phpstan-src/pull/5399#issuecomment-4187117573

- [x] rector/src https://github.com/phpstan/phpstan-src/actions/runs/23979232001/job/69941215874?pr=5399 => I think `'#Provide more specific return type "Iterator" over abstract one#'` is reported because before, `new ObjectType(Iterator)` was equals to `new GenericObject(Iterator)`, which not now. I would say it need to be fix on rector codebase

I reported the issue on rector side https://github.com/rectorphp/rector/issues/9729

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1474,12 +1474,6 @@ parameters:
 			path: src/Type/ObjectShapeType.php
 
 		-
-			rawMessage: 'Doing instanceof PHPStan\Type\Enum\EnumCaseObjectType is error-prone and deprecated. Use Type::getEnumCases() instead.'
-			identifier: phpstanApi.instanceofType
-			count: 1
-			path: src/Type/ObjectType.php
-
-		-
 			rawMessage: Doing instanceof PHPStan\Type\Generic\GenericObjectType is error-prone and deprecated.
 			identifier: phpstanApi.instanceofType
 			count: 1
@@ -1488,7 +1482,7 @@ parameters:
 		-
 			rawMessage: 'Doing instanceof PHPStan\Type\ObjectType is error-prone and deprecated. Use Type::isObject() or Type::getObjectClassNames() instead.'
 			identifier: phpstanApi.instanceofType
-			count: 3
+			count: 2
 			path: src/Type/ObjectType.php
 
 		-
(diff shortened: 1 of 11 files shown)


VincentLanglet:
I'm breaking rector because they have multiple class extending the ObjectType like https://github.com/rectorphp/rector/blob/8e733dbcc4fe72493e533d354e195d77c7baee07/src/StaticTypeMapper/ValueObject/Type/FullyQualifiedObjectType.php#L15 

Which rely on the ObjectType::equals method.

And some rules from rector are relying on the fact that ObjectType::equals(FullyQualifiedObjectType) is true for instance.

ondrejmirtes:
Yes, don't worry about it

VincentLanglet:
> Yes, don't worry about it

@ondrejmirtes Does it means we prefer the "right" fix `get_class($type) !== static::class`, and we'll ask rector to update their implementation
Or should we choose the `instanceof` checks (https://github.com/phpstan/phpstan-src/pull/5399/changes/aa95f94942948bcefb02ef47df161861a611ce21) to imply as few change as possible in rector codebase.

I do understand you prefer the `get_class($type) !== static::class` ?

VincentLanglet:
TODO:
- Adding non regression tests https://github.com/phpstan/phpstan-src/actions/runs/23980124209?pr=5399
- Adding more Unit test on ObjectType::equals

ondrejmirtes:
Yes, I prefer `get_class($type) !== static::class`. Rector's code is just wrong.

staabm (approved):
I think we did a similar fix in MixedType/ErrorType in the past.

https://github.com/phpstan/phpstan-src/commit/ed7eb4722b70188e2072ed0b983910187cbfed4a

PR #5400, opened 2026-04-04
Add hint in VarTagTypeRuleHelper

VincentLanglet:
This is now interesting after https://github.com/phpstan/phpstan-src/pull/5399


staabm (approved):

staabm:
Should we do something about the integration tests? Do we need to update a baseline or something?

VincentLanglet:
> Should we do something about the integration tests? Do we need to update a baseline or something?

I already merge the baseline update ; it should be better now

PR #5420, opened 2026-04-07
exit function is not internal, change isInternal method to return 'No'

spaze:
See discussion https://github.com/phpstan/phpstan/discussions/14438


ondrejmirtes:
Thank you!

PR #5423, opened 2026-04-07
Fix infinite recursion in ConstantArrayType::isCallable() via RecursionGuard

MizouziE:
## Summary

When PHPStan analyses code that uses CarbonPeriod (from the nesbot/carbon package), it enters a recursive type resolution loop and runs out of memory.                                                                                                                              

**Root cause**: ConstantArrayType::isCallable() had no recursion guard. When checking if [ClassName::class, 'method'] is callable, it calls getMethod('method') to verify the method is static (PHP 8.0+ requirement). That triggers full method resolution — getVariants() → getReturnType() → TypehintHelper::decideType(). If the method's native return type is callable, decideType() checks $phpDocType->isCallable(), which calls ConstantArrayType::isCallable() again, creating an infinite loop.

**Fix**: Wrapped ConstantArrayType::isCallable() in RecursionGuard::run(), identical to the existing pattern in ObjectType::isCallable() (line 1558). When re-entry is detected, ErrorType is returned and converted to TrinaryLogic::createNo().

### Changes:                                                                                                                                
  - src/Type/Constant/ConstantArrayType.php: Added RecursionGuard import and wrapping                                                     
  - tests/PHPStan/Analyser/data/bug-constant-array-callable-recursion.php: Regression test data with the self-referential callable pattern  
  - tests/PHPStan/Analyser/AnalyserIntegrationTest.php: Test verifying analysis completes without OOM

### DIFF
--- a/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
+++ b/tests/PHPStan/Analyser/AnalyserIntegrationTest.php
@@ -93,6 +93,14 @@ class AnalyserIntegrationTest extends PHPStanTestCase
 		$this->assertNoErrors($errors);
 	}
 
+	public function testConstantArrayCallableDoesNotCauseInfiniteRecursion(): void
+	{
+		// Previously caused infinite recursion / OOM via ConstantArrayType::isCallable()
+		// resolving getMethod() which triggered return type resolution that called isCallable() again
+		$errors = $this->runAnalyse(__DIR__ . '/data/bug-constant-array-callable-recursion.php');
+		$this->assertCount(3, $errors);
+	}
+
 	public function testClassThatExtendsUnknownClassIn3rdPartyPropertyTypeShouldNotCauseAutoloading(): void
 	{
 		// no error about PHPStan\Tests\Baz not being able to be autoloaded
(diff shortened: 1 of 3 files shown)


staabm (approved):
I am able to reproduce the inifinite loop without the fix, and can also confirm the fix adresses it.

thank you

MizouziE:
I noticed the php 7.4 tests don't like the test accompanying this. Could I use the same pattern I found in just a couple of other bug fix tests like the below?

```diff
      $errors = $this->runAnalyse(__DIR__ . '/data/bug-constant-array-callable-recursion.php');                                    
+    if (PHP_VERSION_ID < 80000) {                                                                                                
+      $this->assertCount(1, $errors);                                                                                            
+      return;                                                                                                                    
+    }                                                                                                                            
     $this->assertCount(3, $errors); 
```

tests/PHPStan/Analyser/AnalyserIntegrationTest.php
```diff
@@ -93,6 +93,14 @@ public function testInfiniteRecursionWithCallable(): void
 		$this->assertNoErrors($errors);
 	}
 
+	public function testConstantArrayCallableDoesNotCauseInfiniteRecursion(): void
```
staabm, line 97:
```suggestion
	#[RequiresPhp('>= 8.0')]
	public function testConstantArrayCallableDoesNotCauseInfiniteRecursion(): void
```

VincentLanglet (approved):

PR #5424, opened 2026-04-07
More concrete return type for `opcache_get_configuration`

devnix:
Array shape based on https://github.com/php/php-src/blob/master/ext/opcache/zend_accelerator_module.c#L800-L904


VincentLanglet:
Not sure such precision is needed ; wasn't https://github.com/phpstan/phpstan-src/pull/5422/changes enough ?
Dunno what you think about it @staabm ; this would make a huge constant array (which are often simplified)

staabm:
what the real world use-case/motivation for this change?

devnix:
Expecting a shape, avoiding unnecessary checks of array keys, reporting typos in array keys...

staabm:
I think we can land the change after we added a separate [NodeScopeResolverTest in nsrt/](https://github.com/phpstan/phpstan-src/tree/2.1.x/tests/PHPStan/Analyser/nsrt).

usually we don't enforce nsrt/ tests for changes in `resources/functionMap.php` but I think in this case it would be useful, because we add a very long line full of escaping etc. and its pretty easy to get something wrong in this case.

does the suggested return-type array-shape also work for PHP7?
is it something we need to make php version specific?

devnix:
@staabm I'm going to carefully review the shape between versions.

Just one quick question: if there is any difference, which version should be in `functionMap.php`, a `7.3 <=` version and then the different version in the deltas?

staabm:
> Just one quick question: if there is any difference, which version should be in `functionMap.php`, a `7.3 <=` version and then the different version in the deltas?

yes. we only need it in delta-files when there are differences.
since we are adding a more precise return type, `functionMap.php` is fine.
if we would e.g. add a more precise parameter type (which is a BC break), we would likely used `functionMap_bleedingEdge.php` instead.

if php version differences exist, we need multiple nsrt/ tests

devnix:
I think I've got it @staabm, sorry for the delay!

tests/PHPStan/Analyser/nsrt/opcache-get-configuration.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace OpcacheGetConfiguration;
+
+use function opcache_get_configuration;
+use function PHPStan\Testing\assertType;
+
+$config = opcache_get_configuration();
+
+if (PHP_VERSION_ID >= 80300) {
```
VincentLanglet: It doesn't work this way.

You should duplicate the files in Analyser/data/ and load then conditionally in NodeScopeResolverTest like it's already for
```
if (PHP_VERSION_ID < 80200 && PHP_VERSION_ID >= 80100) {
			yield __DIR__ . '/data/enum-reflection-php81.php';
		}

		if (PHP_VERSION_ID >= 80100 && PHP_VERSION_ID < 80400) {
			yield __DIR__ . '/data/enum-reflection-backed.php';
		}
```
for instance

devnix: I tried a couple of ways, thanks for the feedback, and sorry for the noise

VincentLanglet (approved):
My only worry is for how long the type from PHP8.3 will stay the same...

staabm (approved):

staabm:
Thank you!

PR #5434, opened 2026-04-09
bisect: advertise gzip,deflate encoding when supported

staabm:
time to fetch the commits..

... before this PR
```
➜  phpstan-src git:(bisect-gzip) ✗ time php bin/phpstan bisect foo.php  --good=2.1.0 --bad=2.1.46

Fetching commits between 2.1.0 and 2.1.46...
--------------------------------------------

php bin/phpstan bisect foo.php --good=2.1.0 --bad=2.1.46  0.27s user 0.17s system 0% cpu 54.959 total
```

... after this PR
```
➜  phpstan-src git:(bisect-gzip) ✗ time php bin/phpstan bisect foo.php  --good=2.1.0 --bad=2.1.46

Fetching commits between 2.1.0 and 2.1.46...
--------------------------------------------

php bin/phpstan bisect foo.php --good=2.1.0 --bad=2.1.46  0.22s user 0.08s system 0% cpu 44.516 total
```

-> saves ~10 seconds when downloading big responses

### DIFF
--- a/src/Command/BisectCommand.php
+++ b/src/Command/BisectCommand.php
@@ -9,6 +9,7 @@ use Nette\Utils\Json;
 use Override;
 use PHPStan\Command\Bisect\BinarySearch;
 use PHPStan\File\FileReader;
+use PHPStan\Internal\HttpClientFactory;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputArgument;
@@ -23,7 +24,6 @@ use function array_values;
 use function chmod;
 use function count;
 use function escapeshellarg;
-use function extension_loaded;
 use function getenv;
 use function implode;
 use function is_array;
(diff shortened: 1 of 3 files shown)


src/Command/BisectCommand.php
```diff
@@ -106,13 +107,18 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			return 1;
 		}
 
+		$headers = [
```
staabm: we might instead create a PHPStan namespaced HttpClient and re-use the gzip support in all places we currently directly use GuzzleHttp\Client

ondrejmirtes: Not a namespaces child class, just a ClientFactory

VincentLanglet (approved):

src/Internal/HttpClientFactory.php
```diff
@@ -0,0 +1,28 @@
...
+
+final class HttpClientFactory
+{
+
+	/**
+	 * @param array<mixed> $config
+	 *
+	 * @see \GuzzleHttp\RequestOptions
+	 */
+	public static function createClient(array $config): Client
```
ondrejmirtes: Should be a service, not a static function. So we can pass options from config parameters

ondrejmirtes (changes requested):

src/Command/BisectCommand.php
```diff
@@ -106,9 +107,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			return 1;
 		}
 
-		$client = new Client([
-			RequestOptions::TIMEOUT => 30,
-			RequestOptions::CONNECT_TIMEOUT => 10,
+		$client = (new HttpClientFactory())->createClient([
```
staabm, line 110: to make use of this service via DI in this command we would need to call `CommandHelpe::begin`, and this might require more changes.

I feel its not the right time todo this now. as this command is using cli options to pass it into a subshell, which `CommandHelpe::begin` seems to expect to be used for the main-process

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

